PR #1, opened 2021-05-28
Kinesis

psFried:
Adds a `kinesis` connector executable that implements the airbyte protocol.
This does not yet build a docker image, but I thought it would be easier to review without that part anyway. I'll create a separate PR for building the image.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1)


jgraettinger (changes requested):
> Reviewed 8 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/1)* status: 8 of 11 files reviewed, 19 unresolved discussions (waiting on @psFried)

___
*[go/kinesis/capture.go, line 15 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb7sPcG7iQ-6V3_FLU5:-Mb7sPcG7iQ-6V3_FLU6:b-mmt7x3) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L15)):*
> ```Go
> type kinesisCapture struct {
> 	client             *kinesis.Kinesis
> 	ctx                context.Context
> ```

See https://github.com/golang/go/wiki/CodeReviewComments#contexts

Generally contexts shouldn't be included in structs. I've broken this rule also, though, but  typically only when the struct represents a specific scoped lifetime.
___
*[go/kinesis/capture.go, line 28 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb7sxNO4T8TSwpy8Wje:-Mb7sxNO4T8TSwpy8Wjf:brv2ulu) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L28)):*
> ```Go
> }
> 
> type kinesisSource struct {
> ```

optional nit: these names stutter.  For example`kinesis.kenesisSource`
___
*[go/kinesis/capture.go, line 50 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb7tS2a3BCNXGwj1oJ8:-Mb7tS2a3BCNXGwj1oJ9:bxs3jvi) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L50)):*
> ```Go
> 		shardSequences: state,
> 	}
> 	var err = kc.startReadindStream()
> ```

nit: typo
___
*[go/kinesis/capture.go, line 79 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb7tsUd5rNBdLd6UGUv:-Mb7tsUd5rNBdLd6UGUw:b3kre) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L79)):*
> ```Go
> 	// Start the background goroutine that will buffer data and send control messages to notify the
> 	// Flow consumer when data is available.
> 	//go kc.startMessagePump()
> ```

?
___
*[go/kinesis/capture.go, line 200 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb81F2Q3AXXJYMllF5x:-Mb81F2Q3AXXJYMllF5y:b-tk6e9) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L200)):*
> ```Go
> 	}
> 
> 	var err error
> ```

nit: i'm getting confused as to whether these variables being extracted outside of the loop is meaningful or not (as opposed to initializing them with getShardIterator()). At first I thought it was meaningful, but it doesn't appear to be?
___
*[go/kinesis/capture.go, line 205 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb80rvgDr5wV5SFuHEd:-Mb80rvgDr5wV5SFuHEe:b-kk7nfk) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L205)):*
> ```Go
> 	for {
> 		shardIter, err = kc.getShardIterator(shardID, shardReader.lastSequenceID)
> 		if err != nil {
> ```

nit: collapse cases into a single-depth if statement? a bit tough to follow. `if err != nil && isRetryable(err)`, or start with the `if err == nil` case and remaining `else if` can handle error variants.
___
*[go/kinesis/capture.go, line 213 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb80UIg7cszRhN360Ma:-Mb80UIg7cszRhN360Mb:bk4bm4a) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L213)):*
> ```Go
> 				case <-kc.ctx.Done():
> 					err = kc.ctx.Err()
> 					break ReadLoop
> ```

just return? does this need to be labeled?

Also it doesn't appear that anything happens with this set `err`, right?
___
*[go/kinesis/capture.go, line 219 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb7ygbr57ziU8RvyP30:-Mb7ygbr57ziU8RvyP31:b-p4ddqg) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L219)):*
> ```Go
> 				// sometime after we started trying to read it. This is probably not indicative of
> 				// any problem, but rather just peculiar timing where we start reading the shard
> 				// right before the last record in the shard expires. This should only really be
> ```

nit: trailed off here.
___
*[go/kinesis/capture.go, line 238 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb89y5p-2wNnD6gxMZL:-Mb89y5p-2wNnD6gxMZM:b951u53) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L238)):*
> ```Go
> 			}
> 		} else {
> 			err = shardReader.readShardIterator(shardIter)
> ```

nit: this can also be collapsed to single depth as well:

` else if err = ....readShardIterator(); err != nil `
___
*[go/kinesis/capture.go, line 246 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8AulKBciVgUgOqJ3M:-Mb8AulKBciVgUgOqJ3N:b-2smnhn) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L246)):*
> ```Go
> 				// Don't wait before retrying, since the previous failure was from GetRecords and
> 				// the next call will be to GetShardIterator, which have separate rate limits.
> 				logEntry.WithField("error", err).Warn("reading kinesis shardIterator returned error (will retry)")
> ```

these are likely to be rate limit errors?
___
*[go/kinesis/capture.go, line 300 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8CKWuCza8VulVRy5W:-Mb8CKWuCza8VulVRy5X:b-9nrj5e) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L300)):*
> ```Go
> 		getRecordsResp, err = r.parent.client.GetRecordsWithContext(r.parent.ctx, &getRecordsReq)
> 		if err != nil {
> 			if isRetryable(err) {
> ```

nit: this can be collapsed as well
___
*[go/kinesis/capture.go, line 307 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8BopM8oXa5jTsUmGI:-Mb8BopM8oXa5jTsUmGJ:b-o42sdz) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L307)):*
> ```Go
> 				case <-r.parent.ctx.Done():
> 					err = r.parent.ctx.Err()
> 					return
> ```

nit: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

| Naked returns are okay if the function is a handful of lines. Once it's a medium sized function, be explicit with your return values.
___
*[go/kinesis/capture.go, line 328 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8Cmsd17Gzn1DAi_0-:-Mb8Cmsd17Gzn1DAi_00:b2ryg0v) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L328)):*
> ```Go
> 			r.updateRecordLimit(getRecordsResp)
> 
> 			var parsed []map[string]interface{}
> ```

I think we can just skip parsing altogether. Go reflection, used here, is notoriously slow. An alternative is to model record data as `json.RawMessage`. The `encoding/json`  package will actually tokenize and match the `RawMessage` to the encoded indentation  (e.g. stripping whitespace, if present), and will error if the tokenization fails.
___
*[go/kinesis/connection.go, line 157 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8F8LE9GeOzBszSVrZ:-Mb8F8LE9GeOzBszSVr_:b2dbedd) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/connection.go#L157)):*
> ```Go
> 
> type backoff struct {
> 	initialMillis int64
> ```

why not have these be `time.Duration` ? Also an int64, and then the unit doesn't leak into the name.
___
*[go/kinesis/connection.go, line 163 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8FbhwEJbHF9Y8iVbJ:-Mb8FbhwEJbHF9Y8iVbK:bm8m11h) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/connection.go#L163)):*
> ```Go
> }
> 
> func (b *backoff) nextBackoff() <-chan time.Time {
> ```

I'm unclear from this interface how this is supposed to change over time (initialMillis is directly increased? or the multiplier is? when and why?)

Also, could https://pkg.go.dev/golang.org/x/time/rate serve here (for example, initialize it for 5 TPS?)
___
*[go/kinesis/main.go, line 119 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8PH6E3ouxOKaBgt4Y:-Mb8PH6E3ouxOKaBgt4Z:b-8gg50k) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/main.go#L119)):*
> ```Go
> 	}
> 	var catalog protocol.ConfiguredCatalog
> 	err = args.CatalogFile.Parse(&catalog)
> ```

nit: it's fewer lines and more idiomatic to do:

```
if err := args.CatalogFile.Parse(&catalog); err != nil {
    return fmt.Errorf(...)
```

throughout.

What's here is correct, but my brain keeps wanting to read into the fact that it's _not_ this way.
___
*[go/kinesis/main.go, line 169 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8P1-LAocUsFUxZ0mA:-Mb8P1-LAocUsFUxZ0mB:bozmg1q) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/main.go#L169)):*
> ```Go
> 			cancelFunc()
> 			return next.Error
> 		} else {
> ```

nit: don't need to indent this (returned an error in prior block)
___
*[go/protocol/args.go, line 1 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8Qez_4TinWabxMZ6h:-Mb8Qez_4TinWabxMZ6i:b-t1faqz) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/protocol/args.go#L1)):*
> ```Go
> package protocol
> ```

go/airbyte-source ?

There will be other protocols (such as the materialization protocol).
___
*[go/protocol/partition_range.go, line 29 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8R6oF489Ocoxx4LfB:-Mb8R6oF489Ocoxx4LfC:bvl7s5q) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/protocol/partition_range.go#L29)):*
> ```Go
> // TODO: this is gross, but I'm not sure if there's a better way
> func (pr *PartitionRange) UnmarshalJSON(bytes []byte) error {
> 	// This map will be the whole object if
> ```

You can create an anonymous struct with string fields, and unmarshal into it:

```
var tmp struct{ Begin *string; End *string }
```

jgraettinger (commented):
Drat, there's an issue here in the interaction of sharding / `IncludesHwHash` and the nature of hand-off from parent to child. How does a capture member discover it should read a child shard it owns, when it doesn't own the parent?

> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/1#-:-MbBZPa8-0nWB-e3sZM3:b-eycy9p)* status: 8 of 11 files reviewed, 19 unresolved discussions (waiting on @psFried)

psFried (commented):
aw crap, you're right.

> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/1#-:-MbBqhOK1K6GHSL7H5gk:b2kmh34)* status: 1 of 15 files reviewed, 10 unresolved discussions (waiting on @jgraettinger)

___
*[go/kinesis/capture.go, line 15 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb7sPcG7iQ-6V3_FLU5:-MbD9hWmA0wm36bCHVFn:b7fsdyl) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L15)):*

See https://github.com/golang/go/wiki/CodeReviewComments#contexts

Generally contexts shouldn't be included in structs. I've broken this rule also, though, but  typically only when the struct represents a specific scoped lifetime.
</blockquote></details>

In this case, the context does indeed correspond to the lifetime of the capture. The struct isn't used outside of this file, but exists mostly to group the numerous parameters that were previously getting passed via function parameters or closure scopes. My gut take is that it's probably OK as it is, and that perhaps it should be more clear that `readStream` is the entrypoint and everything else is a private implementation detail. But maybe that's just my bias from other languages?
___
*[go/kinesis/capture.go, line 28 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb7sxNO4T8TSwpy8Wje:-MbIxWvO-f4LgW2Eqf45:b-5zyeri) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L28)):*

optional nit: these names stutter.  For example`kinesis.kenesisSource`
</blockquote></details>

yea `recordSource` is a much better name anyway
___
*[go/kinesis/capture.go, line 79 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb7tsUd5rNBdLd6UGUv:-MbIxlTl-1EFaGf1bsK6:b8he9j2) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L79)):*

?
</blockquote></details>

How'd that sneak in there? ;) Deleted
___
*[go/kinesis/capture.go, line 200 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb81F2Q3AXXJYMllF5x:-MbIyi2Q5ePCOSd9Efbb:b-k1yfz) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L200)):*

nit: i'm getting confused as to whether these variables being extracted outside of the loop is meaningful or not (as opposed to initializing them with getShardIterator()). At first I thought it was meaningful, but it doesn't appear to be?
</blockquote></details>

yeah they didn't need to be up there. Moved them down into the loop.
___
*[go/kinesis/capture.go, line 213 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb80UIg7cszRhN360Ma:-MbIzYyiCCdV_unSGAPW:b-3f5fvd) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L213)):*

just return? does this need to be labeled?

Also it doesn't appear that anything happens with this set `err`, right?
</blockquote></details>

The only reason for break instead of return was just the logging at the end. The only reason it needed to be labeled was because of the `break` being inside the `select` block. I refactored this function to log from a `defer` and to use `return` instead of `break`. Also got rid of that err assignment.
___
*[go/kinesis/capture.go, line 246 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8AulKBciVgUgOqJ3M:-MbJ0uqA9vAybzR0-R_T:bpe3ux9) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L246)):*

these are likely to be rate limit errors?
</blockquote></details>

No, rate limit errors are handled within `readShardIterator`, as are any other errors that are considered retryable. So we only loop around if reading the existing shard iterator failed with some non-retryable error, such as an `ExpiredIteratorException`. That scenario (which I think is probably the most likely) could be caused by our application taking too long to process all the records that were returned, causing the iterator to expire in the meantime. In that case, or in the case of any other 4xx error reading the iterator, I don't see any reason to wait before acquiring a new iterator.
___
*[go/kinesis/capture.go, line 307 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8BopM8oXa5jTsUmGI:-MbN08ux0KyTebuitFo-:bz5x4ub) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L307)):*

nit: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

| Naked returns are okay if the function is a handful of lines. Once it's a medium sized function, be explicit with your return values.
</blockquote></details>

Yea the naked return was more of a holdover from a previous factoring anyway. I got rid of it, and agree that it's more readable.
___
*[go/kinesis/capture.go, line 328 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8Cmsd17Gzn1DAi_0-:-MbN0T_W3-oY7gDIBQHL:bn0k7g5) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/capture.go#L328)):*

I think we can just skip parsing altogether. Go reflection, used here, is notoriously slow. An alternative is to model record data as `json.RawMessage`. The `encoding/json`  package will actually tokenize and match the `RawMessage` to the encoded indentation  (e.g. stripping whitespace, if present), and will error if the tokenization fails.
</blockquote></details>

TIL about `RawMessage`. That's exactly the behavior I wanted here, but didn't know how else to get it without writing a bunch of bespoke code. I changed it to use `RawMessage` and avoid parsing altogether.
___
*[go/kinesis/connection.go, line 157 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8F8LE9GeOzBszSVrZ:-MbNAnUR7dJAghW6JumP:b-fxa88g) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/connection.go#L157)):*

why not have these be `time.Duration` ? Also an int64, and then the unit doesn't leak into the name.
</blockquote></details>

This backoff struct may not stick around for very long. My hope is to be able to use the `Retryer` interface from the aws sdk, but it'll take a bit of experimentation because their docs are crap. Now that this is in a separate repo, there's no reason not to use the `v2` sdk, which might have more clarity on error handling. I was keeping with the v1 sdk just to avoid linking to multiple versions in the flow repo. TLDR, I'd like to leave this quick and dirty backoff here for now and just get rid of it later.
___
*[go/kinesis/connection.go, line 163 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8FbhwEJbHF9Y8iVbJ:-MbNDaDM-wfmIMP9WSX1:b-wfbdfd) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/connection.go#L163)):*

I'm unclear from this interface how this is supposed to change over time (initialMillis is directly increased? or the multiplier is? when and why?)

Also, could https://pkg.go.dev/golang.org/x/time/rate serve here (for example, initialize it for 5 TPS?)
</blockquote></details>

`currentMillis` is what gets mutated here. See the above comment on backoff, but I do think we want some sort of backoff for most of these cases, not a rate limiter, since we want the backoff to increase as we encounter subsequent errors. We could use a rate limiter around the loop that's reading the shard iterator, though, which would help avoid getting rate limit errors in the first place.
___
*[go/kinesis/main.go, line 119 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8PH6E3ouxOKaBgt4Y:-MbN4mdq57QK_6dWCdNK:b2z1max) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/kinesis/main.go#L119)):*

nit: it's fewer lines and more idiomatic to do:

```
if err := args.CatalogFile.Parse(&catalog); err != nil {
    return fmt.Errorf(...)
```

throughout.

What's here is correct, but my brain keeps wanting to read into the fact that it's _not_ this way.
</blockquote></details>

Yeah, sorry, I know you've mentioned this before. TBH it's one of those go idioms that I find pretty distasteful, but I'd rather follow a distasteful pattern if it's the one everyone is used to. I'll try to collapse them into the conditional all by default.
___
*[go/protocol/partition_range.go, line 29 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8R6oF489Ocoxx4LfB:-MbN7PDIEeNMyiBo7TNG:b-kb8v6d) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/protocol/partition_range.go#L29)):*

You can create an anonymous struct with string fields, and unmarshal into it:

```
var tmp struct{ Begin *string; End *string }
```
</blockquote></details>

Cool, thanks
___
*[go/protocol/args.go, line 1 at r1](https://reviewable.io/reviews/estuary/connectors/1#-Mb8Qez_4TinWabxMZ6h:-MbN9BPcEeFgKGkKiaTv:b-3ajjci) ([raw file](https://github.com/estuary/connectors/blob/d5bd8eb6c5120620171b5b80efb2234d1ae2c430/go/protocol/args.go#L1)):*

go/airbyte-source ?

There will be other protocols (such as the materialization protocol).
</blockquote></details>

Right on. I went with just plain `airbyte`, since technically that's inclusive of both the source and destination protocols, and doesn't have a hyphen.

jgraettinger (changes requested):
just minor comments, otherwise :lgtm_strong: 

> Reviewed 13 of 14 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/1#-:-McBUr2P3UaYDlBrNDTW:b-51zbok)* status: all files reviewed, 5 unresolved discussions (waiting on @psFried)

___
*[go/kinesis/capture.go, line 480 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBMiRF7NErN7B-hEF6:-McBMiRF7NErN7B-hEF7:b3kucnf) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/capture.go#L480)):*
> ```Go
> // The AWS SDK will wrap context.Canceled errors, sometimes in multiple layers,
> // which is what this function is meant to deal with.
> func isContextCanceled(err error) bool {
> ```

nit: could this use https://golang.org/pkg/errors/#Is ? (does awserr.Error implement Unwrap()? )
___
*[go/kinesis/connection.go, line 18 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBNHotC1Q5L_n7kQTn:-McBNHouCJmWi_heyGia:bmyciti) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/connection.go#L18)):*
> ```Go
> // It matches the `KinesisConfig` struct in `crates/sources/src/specs.rs`
> type Config struct {
> 	ShardRange         *shardrange.Range `json:"shardRange"`
> ```

No action:

One question I have is, does Airbyte supported nested configuration objects? This _may_ break their UI for configuring connectors (I don't know).

In any case, in a future PR I think we'll want to move owned key ranges into the configured catalog, rather than the configuration portion of the connector spec.
___
*[go/kinesis/main.go, line 85 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBOWkP5gLUwlcmE2bT:-McBOWkQ6yV6pkEji2Zh:bp2u8p9) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/main.go#L85)):*
> ```Go
> }
> 
> func put(state *airbyte.State, source *recordSource, sequenceNumber string) {
> ```

nit: updateState ? `put` naming confused me.
___
*[go/kinesis/main.go, line 131 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBPnQX3sKKDLP2bFzD:-McBPnQX3sKKDLP2bFzE:bv9o5og) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/main.go#L131)):*
> ```Go
> 		Type: airbyte.MessageTypeState,
> 		State: &airbyte.State{
> 			Data: make(map[string]interface{}),
> ```

How about making `Data` an `interface{}`, and then instantiating the exact structure shape you want here (eg `make(map[string]map[string]string)` ?).

Or, you could have `Data` be `map[string]json.RawMessage` and defer parsing into the `map[string]string` ? Not sure if that structure works better. In any case the manual runtime type validation above is a smell that could be handled by the `encoding/json` package.
___
*[go/kinesis/shard_range.go, line 14 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBT13I9yuIUSZ-0rG5:-McBT13I9yuIUSZ-0rG6:b25pvfy) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/shard_range.go#L14)):*
> ```Go
> // Applies the same hash operation that kinesis uses to determine which shard a given record should
> // map to, and then truncates the lower 96 bits to translate it into the uint32 hashed key space.
> func hashPartitionKey(key *string) uint32 {
> ```

nit: why is `key` a pointer? it's not valid for it to be nil (and `string` is passed by reference already)

psFried (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/1)* status: all files reviewed, 2 unresolved discussions (waiting on @jgraettinger)

___
*[go/kinesis/capture.go, line 480 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBMiRF7NErN7B-hEF6:-McEvE9R5MGf8HD0ma7_:b-aoboqn) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/capture.go#L480)):*

nit: could this use https://golang.org/pkg/errors/#Is ? (does awserr.Error implement Unwrap()? )
</blockquote></details>

Unfortunately, `awserr.Error` calls it `OrigErr()` instead of `Unwrap()`, which I think means that the `errors` functions won't work :(
___
*[go/kinesis/connection.go, line 18 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBNHotC1Q5L_n7kQTn:-McExNZ8AfoLMKiGHU03:bawy1si) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/connection.go#L18)):*

No action:

One question I have is, does Airbyte supported nested configuration objects? This _may_ break their UI for configuring connectors (I don't know).

In any case, in a future PR I think we'll want to move owned key ranges into the configured catalog, rather than the configuration portion of the connector spec.
</blockquote></details>

Yeah good point about nested configuration. That would be good to know. And agreed that moving the key ranges into the configured catalog seems like the way to go.
___
*[go/kinesis/main.go, line 131 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBPnQX3sKKDLP2bFzD:-McEyXn_DD3qQjPim4Hr:b-skovp2) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/main.go#L131)):*

How about making `Data` an `interface{}`, and then instantiating the exact structure shape you want here (eg `make(map[string]map[string]string)` ?).

Or, you could have `Data` be `map[string]json.RawMessage` and defer parsing into the `map[string]string` ? Not sure if that structure works better. In any case the manual runtime type validation above is a smell that could be handled by the `encoding/json` package.
</blockquote></details>

The airbyte schema specifies that this is `type: object`, which is why I went for the map type. If this was Rust, then I'd want to use the `serde_json::Map` type where the values are `serde_json::Value`s. I think the golang types corresponding to `serde_json::Value` would be either `interface{}` or `json.RawMessage`, unless you can think of another option. If it's just between those two, then I'd prefer `interface{}` because using `json.RawMessage` would require re-encoding the value every time it's updated. That re-encoding wouldn't make a difference for this kinesis connector, but I can imagine other connectors wanting to accumulate multiple state updates before they write the State message, which seems like it would get awkward.
That said, I'm 100% with you on the bad smell here. I'm just not sure what, if anything, to change. I thought about maybe defining our own `JsonValue` interface, but it seemed like it would add a lot of code, and would still require type assertions in order to update the map. LMK if you have other ideas, or if there's something I'm missing WRT `json.RawMessage`.
___
*[go/kinesis/shard_range.go, line 14 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBT13I9yuIUSZ-0rG5:-McF1D6k7jCM3CUrCeVS:b-ifc5sk) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/shard_range.go#L14)):*

nit: why is `key` a pointer? it's not valid for it to be nil (and `string` is passed by reference already)
</blockquote></details>

Oh this was a relic from when this function was defined within the kinesis capture file, since the aws sdk uses pointers everywhere for "reasons". I just updated this to take a plain `string`, which is for sure what we want.

jgraettinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/1)* status: 11 of 15 files reviewed, 1 unresolved discussion (waiting on @jgraettinger and @psFried)

___
*[go/kinesis/capture.go, line 480 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBMiRF7NErN7B-hEF6:-McF5kPu3N4bkHzHqrvb:b-50bjr) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/capture.go#L480)):*
<details><summary><i>Previously, psFried (Phil Fried) wrote…</i></summary><blockquote>

Unfortunately, `awserr.Error` calls it `OrigErr()` instead of `Unwrap()`, which I think means that the `errors` functions won't work :(
</blockquote></details>

lol wat. okay then
___
*[go/kinesis/main.go, line 131 at r3](https://reviewable.io/reviews/estuary/connectors/1#-McBPnQX3sKKDLP2bFzD:-McF5sUmDrynVqiDSMKE:bi09q4t) ([raw file](https://github.com/estuary/connectors/blob/e0268cf4af488eafdb9cce2d883bd5487f9ecd55/go/kinesis/main.go#L131)):*
<details><summary><i>Previously, psFried (Phil Fried) wrote…</i></summary><blockquote>

The airbyte schema specifies that this is `type: object`, which is why I went for the map type. If this was Rust, then I'd want to use the `serde_json::Map` type where the values are `serde_json::Value`s. I think the golang types corresponding to `serde_json::Value` would be either `interface{}` or `json.RawMessage`, unless you can think of another option. If it's just between those two, then I'd prefer `interface{}` because using `json.RawMessage` would require re-encoding the value every time it's updated. That re-encoding wouldn't make a difference for this kinesis connector, but I can imagine other connectors wanting to accumulate multiple state updates before they write the State message, which seems like it would get awkward.
That said, I'm 100% with you on the bad smell here. I'm just not sure what, if anything, to change. I thought about maybe defining our own `JsonValue` interface, but it seemed like it would add a lot of code, and would still require type assertions in order to update the map. LMK if you have other ideas, or if there's something I'm missing WRT `json.RawMessage`.
</blockquote></details>

If the definition were `Data interface{}`, and the _value_ of `Data` is pre-instantiated to a map type, my understanding is that the `json` package will parse into that runtime value and return an error if there's a type mismatch. So, if `Data` were instantiated with map[string]map[string]string, for example, then it would require a two-level map of strings (and nothing else). I _think_ this is how it works, but I'm not 100%.

So, a way to structure this would be to have an explicit, typed variable `foo` holding that `map[string]map[string]string` or whatever, then set `thing.Data = foo`, and then continue to use `foo` directly while encoding `stateMessage` whenever needed. Since maps are a reference type, it will encode the present value of `foo`.

PR #5, opened 2021-06-25
S3 source

psFried:


source-s3/capture.go
```diff
@@ -0,0 +1,209 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"regexp"
+	//"strings"
```
jgraettinger, line 11: nit: cleanup

source-s3/capture.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+	}
+	log.WithField("observedObjects", objectCount).Info("completed processing stream")
+	return nil
+}
+
+func (c *stream) importObject(ctx context.Context, obj *s3Object, onResult func(readResult) error) error {
+	var fullKey = obj.fullKey()
+
+	// Does this key hash into our shard's assigned range?
```
jgraettinger, line 96: As I recall the plan was to hoist and consolidate these checks plus the regex check into a single place ?

source-s3/capture.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	if parseConfig.AddValues == nil {
+		parseConfig.AddValues = make(map[parser.JsonPointer]interface{})
+	}
+	parseConfig.AddValues[sourceFilenameLocation] = relativeKey
+	return &parseConfig
+}
+
+const sourceFilenameLocation = "/_meta/sourceFile"
+const defaultRecordOffsetLocation = "/_meta/recordOffset"
```
jgraettinger, line 195: "/_meta/source/record" and "/_meta/source/file" ? Consolidate source attributes into a well-known property seems a good idea.

go-types/parser/run.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os/exec"
+	"syscall"
+
+	"github.com/estuary/connectors/go-types/firsterror"
```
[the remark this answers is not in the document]
psFried: Yea, I went back and forth on this one a bit, and ultimately put them in separate packages just for lack of unifying package name that seemed reasonable. I don't think there's a way to keep the types private, though, since they need to be usable from other packages. Am I missing something there? The alternative would I think be exposing more of a high-level public API like `func RunProgram(...)` that just accepts the callback parameters and a `[]string` for the program/args. But I was hesitant to implement that since it's not clear to me that both the captures runtime and the connectors themselves would actually want the same API.
Ultimately, I'm fine with grouping them into a single package, but I think my preference would be to default to keeping them separate. WDYT?

go-types/firsterror/firsterror.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Collector is used to accumulate at most one error, which will be the first one that occurs.
+type Collector struct {
+	first error
+	mutex sync.Mutex
+}
+
+// OnError sets the first error, if it is not already set, and if e is not nil. If the first error
+// has already been set, then e is logged and ignored.
+func (c *Collector) OnError(e error) {
```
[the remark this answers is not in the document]
psFried: yea I just went with `SetIfNil`

parser/src/format/json.rs
```diff
@@ -0,0 +1,19 @@
+//! Parser for the json format. This is similar to jsonl, but more lax because it allows any amount
+//! of whitespace (including newlines) within and in between records. The source offset for this
+//! parser is in terms of bytes, not lines, unlike jsonl.
```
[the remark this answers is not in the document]
psFried: whoops! fixed it

go-types/parser/run.go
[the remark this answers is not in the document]
jgraettinger: What we're building here is fundamentally a Go development kit for building Airbyte-spec connectors, similar to the python CDK that Airbyte has published. Since this is intended as public facing, the bias must be that we're only making things public when reasonably confident it's the right API -- we should be delaying exporting things as long as possible, because they can't be taken back. For that reason I think the bias is to lump it together in a package as private implementation. I also agree that a `RunProgram` factoring is probably what it will evolve into, but we don't need to figure that out quite yet.

source-s3/capture.go
psFried: yea that was the plan, and apparently I dropped the ball there. Fixed now

source-s3/capture.go
psFried: Yeah, agreed :+1:

PR #6, opened 2021-06-30
Use json.RawMessage & extend ConfiguredCatalog

jgraettinger:


go-types/airbyte/catalog.go
```diff
@@ -74,6 +76,8 @@ type Catalog struct {
 
 type ConfiguredCatalog struct {
 	Streams []ConfiguredStream `json:"streams"`
+	Block   bool               `json:"block"`
```
psFried, line 79: This is maybe a bit of a pedantic nit, but I think the term "continuous" might be better than "block". Subjectively, it seems to better convey the meaning that the connector should never exit unless it encounters an error or receives a signal.
I agree that this is the right place to put this, though, and "block" is definitely not terrible, so I'm good with however you want to resolve this.

psFried (approved):
One suggestion, but otherwise LGTM

go-types/airbyte/catalog.go
jgraettinger: How about `tail` ?

PR #7, opened 2021-07-02
Connector configuration cleanups

psFried:
This PR addresses a few different issues:

- Add `projections` to `ConfiguredStream` and pass them through to the parser configuration in the S3 connector
- Namespace `ConfiguredStream.projections` and `ConfiguredCatalog.tail | range` with `estuary.dev/` prefix, but still accept the un-prefixed field names as well. The intent will be to do away with the prefix if and when these are merged into the airbyte spec.
- Fixes a few issues I found with the `spec` subcommands panicing due to invalid json.
- Add missing pre-requisites to connector build makefile tasks

### DIFF
--- a/go-types/airbyte/catalog.go
+++ b/go-types/airbyte/catalog.go
@@ -53,6 +53,36 @@ type ConfiguredStream struct {
 	DestinationSyncMode DestinationSyncMode `json:"destination_sync_mode"`
 	CursorField         []string            `json:"cursor_field,omitempty"`
 	PrimaryKey          []string            `json:"primary_key,omitempty"`
+	Projections         map[string]string   `json:"estuary.dev/projections"`
+}
+
+// This impl exists solely so that we can allow deserializing either the namespaced or
+// non-namepsaced version of Projections, for the purpose of compatibility.
+func (s *ConfiguredStream) UnmarshalJSON(b []byte) error {
+	var tmp = struct {
+		Stream              Stream              `json:"stream"`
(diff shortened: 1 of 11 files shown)


go-types/airbyte/catalog.go
```diff
@@ -53,6 +53,36 @@ type ConfiguredStream struct {
 	DestinationSyncMode DestinationSyncMode `json:"destination_sync_mode"`
 	CursorField         []string            `json:"cursor_field,omitempty"`
 	PrimaryKey          []string            `json:"primary_key,omitempty"`
+	Projections         map[string]string   `json:"estuary.dev/projections"`
+}
+
+// This impl exists solely so that we can allow deserializing either the namespaced or
```
jgraettinger, line 59: groaaaannnn. I don't have anything better to offer :(

go-types/shardrange/partition_range.go
```diff
@@ -25,6 +25,11 @@ func NewFullRange() Range {
 	}
 }
 
+// IsZeroed returns true if the begin and end are both 0
+func (r Range) IsZeroed() bool {
```
jgraettinger, line 29: nit: traditionally this is named `IsZero` ([example](https://golang.org/pkg/time/#Time.IsZero))

jgraettinger (approved):
LGTM % nit. Sorry, I missed that this was sitting in my inbox! Believe this also fixes the JSON schema schema errors I was seeing.

PR #11, opened 2021-07-14
factor out `filesource` package and add GCS connector

jgraettinger:
I have a bunch of unit-test work to do in filesource, but wanted to get this in front of you.

I've manually tested both S3 and GCS, including manually turning on monotonic mode (code change currently) and incremental states. I've also done testing of the E2E `flowctl discover` workflow from S3 => sqlite.


go-types/filesource/filesource.go
```diff
@@ -0,0 +1,563 @@
...
+
+	cfg.Filename = obj.Path
+	if obj.ContentType != "" {
+		cfg.ContentType = obj.ContentType
+	}
+	if obj.ContentEncoding != "" {
+		cfg.ContentEncoding = obj.ContentEncoding
+	}
+	// If the user supplied a location for this, then we'll use that. Otherwise, use the default
+	if cfg.AddRecordOffset == "" {
```
psFried, line 483: I'm realizing that if the user supplies a different location for this, then the document schema and SourceDefinedPrimaryKey will no longer be correct. I don't know if that really matters, though, since the user can change the schema, and I don't know if we really do anything with SourceDefinedPrimaryKey anyway.

go-types/filesource/filesource.go
```diff
@@ -0,0 +1,563 @@
...
+	}
+	if cfg.AddValues == nil {
+		cfg.AddValues = make(map[parser.JsonPointer]interface{})
+	}
+	if cfg.Schema == nil {
+		cfg.Schema = r.schema
+	}
+	cfg.Projections = r.projections
+
+	cfg.AddValues[metaFileLocation] = obj.Path
```
psFried, line 494: It seems a little weird to me that the file path and record offset are handled differently. It feels like they should either both be user-configurable or neither should be. Personally, my instinct at this point would be to say that neither needs to be user-configurable, but given our direct embedding of the parser json schema that may be impractical. But one idea that comes to mind would be to unconditionally overwrite `AddRecordOffset` with the hard-coded default. This is based on the assumption that our strategy of directly embedding the parser schema will soon be replaced by a bespoke schema that includes only a subset of the fields exposed by the parser config. I can't really think of a reason why users would need to customize either of these, though.

go-types/filesource/filesource.go
```diff
@@ -0,0 +1,563 @@
...
+				return fmt.Errorf("during listing %q: %w", prefix, err)
+			} else if entry.IsPrefix {
+				stack = append(stack, entry.Path)
+			} else {
+				hasObjects = true
+			}
+		}
+
+		// Make streams of non-empty directories and the discovery root,
+		// even if the latter is empty.
```
psFried, line 179: It seems a little weird for the stream name to be an empty string, although I don't see anything that explicitly disallows it. Just thinking out loud, we could use `"/"` as a sentinel value for the root. It might be better for users to type `stream: /` in their flow.yaml as opposed to `stream: ""`.

source-s3/main.go
```diff
@@ -4,226 +4,280 @@ import (
...
+		for len(l.output.Contents) != 0 {
+			var obj = l.output.Contents[0]
+			l.output.Contents = l.output.Contents[1:]
 
-		// Should we emit a state checkpoint?
-		if chunker.shouldEmitState(&next) {
-			if err := writeState(&stateMessage, stateMap, encoder); err != nil {
-				cancelFunc()
-				return fmt.Errorf("failed to write state: %w", err)
+			// returns true if the object is one that may be readable. This function filters out any objects
```
psFried, line 162: nit: comment could use updated since this is no longer a separate function

psFried (commented):
I think this is really good, especially the clever way that state is managed and kept small. The comments I left are all things that felt like are worth talking about, but I'm honestly not sure whether any of them need to be changed in this PR. Happy to talk through any of it.

go-types/filesource/filesource.go
```diff
@@ -0,0 +1,563 @@
...
+
+		var hasObjects bool
+		for i := 0; i != discoverListLimit; i++ {
+			var entry, err = listing.Next()
+			if err == io.EOF {
+				break
+			} else if err != nil {
+				return fmt.Errorf("during listing %q: %w", prefix, err)
+			} else if entry.IsPrefix {
+				stack = append(stack, entry.Path)
```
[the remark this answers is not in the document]
jgraettinger: Done.

go-types/filesource/filesource.go
```diff
@@ -0,0 +1,563 @@
...
+
+	if err := r.shared.enc.Encode(stateWrapper); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+const discoverListLimit = 256
+const metaFileLocation = "/_meta/file"
```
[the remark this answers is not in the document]
jgraettinger: Yea... I started playing with it through the flowctl discover workflow, and also added in a minimal amount of schema into the connector in order to schematize it, and just found it really gross to have that much nesting. Can discuss in person, but while I think `/_meta/source` is nice and explicit, it fails the sniff test of initial UX.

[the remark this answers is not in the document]
jgraettinger: Addendum: I'm trying `offset` rather than `record` as I believe that's a more easily re-used name for other types of connectors (e.g. Kafka). It's hand-wavy-enough to be general purpose while being specific enough to be reasonable in the context of a given connector.

go-types/filesource/filesource.go
jgraettinger: We do use SourceDefinedPrimaryKey -- it's plumbed through to the `key` of the collection.

Yea, it will break currently. It would require dynamically parsing the pointer and adjusting the schema. Right now I'd rather just say "you can't change this at all".

go-types/filesource/filesource.go
jgraettinger: Agreed. I don't believe I changed this behavior from how the `source-s3` connector previously did it.

My preference would also be to yank these out as things which can be configured.

go-types/filesource/filesource.go
jgraettinger: it's actually `stream: "the-bucket:"`

go-types/filesource/filesource.go
psFried: Yeah, I had that thought that I probably wrote it this way originally. Just didn't think that through at the time.

jgraettinger: To do this properly would require that rust, and not Go, defines `metaOffsetLocation`, right?

Then it could be removed as a parser configurable.

go-types/filesource/filesource.go
psFried: Ah, gotcha! That makes more sense. Though now I'm waffling a bit on the `:` vs `/` separating the bucket and prefix. It's looking like most other tools use a `/` to separate them, and `stream: the-bucket/` seems a little more intuitive than `stream: "the-bucket:"`. I hadn't considered before how it might look in the yaml when you want to capture everything from a bucket.

jgraettinger: Sounds good, I'll change back to '/' to separate bucket and key.

jgraettinger:
I think this is ready to go now, if you wouldn't mind looking at new commits. Thanks!

psFried (approved):
I think this looks good. I'm still not 100% comfortable with all the edge cases around state management and last-modified timestamps... I feel like it's better than how it used to be, with an entry for every single file, and I also can't identify any cases that seem like they would be very common in practice. So overall, I feel like this is good to merge. But just thinking out loud, I'm wondering if maybe we should try to keep some time-bounded history of files, in order to have a well-defined limit of how much temporal "weirdness" we can tolerate without missing a file? I'd be down to talk through ideas, but I also feel like we could just merge this as it is, and we likely wouldn't have any complaints.

go-types/filesource/state.go
```diff
@@ -0,0 +1,218 @@
...
+	// Base path which is currently being processed, or was last processed (if Complete).
+	Path string `json:"path,omitempty"`
+	// Number of records from the file at |Path| which have been emitted.
+	Records int `json:"records,omitempty"`
+	// Whether the file at |Path| is complete.
+	Complete bool `json:"complete,omitempty"`
+
+	// skip is used for crash recovery. It's the number of records of the current
+	// Path which we'll skip, to seek to the point of prior maximum progres.
+	skip int `json:"-"`
```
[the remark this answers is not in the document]
jgraettinger: That's right. We don't need to serialize it, as it will be re-computed from Records on a future recovery (if there is one).

go-types/filesource/state.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// State of a captured file prefix.
+type State struct {
+	// Exclusive lower bound modification time of files processed in this sweep.
+	MinBound time.Time `json:"minBound,omitempty"`
+	// MaxBound is the exclusive upper-bound of files being processed in this sweep.
+	// If MaxBound is nil, then this state checkpoint was generated after the completion
+	// of a prior sweep and before the commencement of the next one.
+	MaxBound *time.Time `json:"maxBound,omitempty"`
```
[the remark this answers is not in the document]
jgraettinger: MaxBound isn't drawn from file modification times (that's MaxMod). MaxBound is derived from time.Now.

State doesn't care if there are 1 or N files that have a particular ModTime. Can you elaborate on which case you're worried about?

jgraettinger:
You know, what checkpoints really need is schemas and reduction annotations. With that it becomes reasonable for checkpoints to be much larger, and to track a lot more, with that cost only being paid on initial startup of a connector. Incremental, fine-grain updates can be small and lightweight.

But that means breaking compatibility w/ the airbyte spec... :(

go-types/filesource/state.go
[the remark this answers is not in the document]
psFried: Suppose we list and write out `fileA` with las-modified time of `x`, then write the state, then crash. When we start back up, we see `fileB` with last-modified time also of `x`. Would we process `fileB` or skip it?
I really don't have a specific concern here, but just thinking out loud about possible edge cases, and checking for understanding.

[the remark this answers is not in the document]
jgraettinger: We'd process `fileB`. The MinBound only increases *between* sweeps, not after recovering from a crashed / partial sweep. So it starts up using the exact same MinBound / MaxBound the pre-crash sweep was using.

We'd also continue to accumulate MaxMod, so if fileA is processed before a crash and fileB after, then MaxMod will still have the greater of the two modTimes.

PR #12, opened 2021-07-15
Parser handles compressed files

psFried:
Parser now detects and decompresses input that is compressed. Currently,
this only supports gzip compression and zip archives, though surely more
compression formats will follow.

Adds the `contentEncoding` field to the parser config, to allow
detection of compression based on a `Content-Encoding` header, if known.


jgraettinger (approved):
LGTM! The manual ZipArchive reference management is a bit of a smell but I don't have a better suggestion.

psFried:
> The manual ZipArchive reference management is a bit of a smell but I don't have a better suggestion.
Sigh, yea :/

PR #13, opened 2021-07-16
Minor fixes to source-s3

psFried:
- Use AnonymousCredentials when no credentials are provided in the config. This allows ingestion of data from public buckets.
- Include the object Path in the error message when parsing fails, so that it's easier to track down parse errors.

### DIFF
--- a/go-types/filesource/filesource.go
+++ b/go-types/filesource/filesource.go
@@ -229,12 +229,13 @@ func (src Source) Read(args airbyte.ReadCmd) error {
 	if err != nil {
 		return err
 	}
-	var catalog = airbyte.ConfiguredCatalog{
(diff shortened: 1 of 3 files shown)


go-types/filesource/filesource.go
```diff
@@ -401,7 +401,7 @@ func (r *reader) processObject(ctx context.Context, obj ObjectInfo) error {
 		return r.emit(lines)
 	})
 	if err != nil {
-		return err
+		return fmt.Errorf("Failed to parse object %q: %w", obj.Path, err)
```
jgraettinger, line 404: nit: errors should begin w/ lower-case (linter will flag this)

jgraettinger (approved):
LGTM!

PR #18, opened 2021-08-03
update source-s3 spec to make keys optional

psFried:
Resolves points 2 and 3 in #181

### DIFF
--- a/source-s3/main.go
+++ b/source-s3/main.go
@@ -209,20 +209,20 @@ func main() {
 		"title":   "S3 Source Spec",
 		"type":    "object",
 		"required": [
-			"awsAccessKeyId",
-			"awsSecretAccessKey"
+			"bucket",
+			"region"
 		],
 		"properties": {
 			"awsAccessKeyId": {
 				"type":        "string",
 				"title":       "AWS Access Key ID",
-				"description": "Part of the AWS credentials that will be used to connect to S3",
+				"description": "Part of the AWS credentials that will be used to connect to S3. Required unless the bucket is public and allows anonymous listings and reads.",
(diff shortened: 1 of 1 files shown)


source-s3/main.go
```diff
@@ -255,7 +255,7 @@ func main() {
 			"region": {
 				"type":        "string",
 				"title":       "AWS Region",
-				"description": "The name of the AWS region where the S3 stream is located",
+				"description": "The name of the AWS region where the S3 stream is located. \"us-east-1\" is a popular default you can try, if you're unsure what to put here.",
```
saterus, line 258: Is "S3 stream" the way _we're_ choosing to talk about it here? Or would this make more sense to the user if we just said "S3 bucket"?

source-s3/main.go
```diff
@@ -209,8 +209,8 @@ func main() {
 		"title":   "S3 Source Spec",
 		"type":    "object",
 		"required": [
-			"awsAccessKeyId",
-			"awsSecretAccessKey"
```
saterus, line 213: Should we add anything else to the `description` fields for these? Omitting these fields is only really viable if it's completely public, right?

saterus (approved):
Looks good. Small questions, but not blockers.

source-s3/main.go
psFried: lol I didn't even notice that. I'm assuming it's an artifact of copypasta from source-kinesis. Changed it to "bucket".

source-s3/main.go
psFried: Yeah good idea. I added a bit of clarification.

PR #19, opened 2021-08-06
Filter filesource prefixes

psFried:


saterus (approved):
Looks good.

Are there any other sorts of files we should automatically exclude? Didn't we talk about skipping hidden files?

psFried:
> Looks good.
> 
> Are there any other sorts of files we should automatically exclude? Didn't we talk about skipping hidden files?

The hidden files discussion was specifically related to files within zip archives. It's certainly arguable that we should also exclude s3 objects if the filename starts with a `.`, but TBH it's not a situation I've seen before, and the `matchKeys` regex might be a better approach for that. AFAIK there's not any other files we should exclude (we already filter out objects in s3 Glacier).

PR #25, opened 2021-09-02
Build Connectors inside Docker

saterus:
I've got CI working for all the new Docker builds. A fully cached CI run takes just over 4 minutes. :fireworks: Cache misses will obviously make that take longer, but it's a nice improvement from the 10-11 minute builds we have without any caching whatsoever.

The resulting images are also nice and small. :tada: 

```
❯ docker images | head -n 5
REPOSITORY                                                    TAG                         IMAGE ID       CREATED             SIZE
ghcr.io/estuary/source-s3                                     6eaa000                     db3db1e68d53   14 minutes ago      40.3MB
ghcr.io/estuary/source-kinesis                                6eaa000                     784cc91c8450   15 minutes ago      30.1MB
ghcr.io/estuary/source-hello-world                            6eaa000                     e5d456d8a46c   15 minutes ago      22.1MB
ghcr.io/estuary/source-gcs                                    6eaa000                     33a33c727b0f   22 minutes ago      45.4MB
```


source-kinesis/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# Run the unit tests
+RUN go test -v /builder/source-kinesis/...
+
+# Build the connector
+RUN go build -o /builder/connector -v /builder/source-kinesis/...
+
+
+# Runtime Stage
+################################################################################
+FROM gcr.io/distroless/base-debian10:debug
```
saterus, line 31: This is a neat trick I wanted to point out. You can attach the `:debug` tag to any `distroless` image and it'll give you a busybox shell that you can `docker exec` into. :tada:

.dockerignore
```diff
@@ -1,9 +1,16 @@
-# Ignore everything except for the specific files we actually care about. This is to avoid sending a
-# massive build context to the docker daemon, and also allows us to avoid copying everything we
-# might need into a temp directory.
-*
```
saterus, line 4: I did a little digging here. With this `.dockerignore` file as it stands now, I've got a ~10MB build context sent to Docker.

That said, one option we could try avoid such micromanagement of this `.dockerignore` would be to use [BuildKit](https://docs.docker.com/develop/develop-images/build_enhancements/). It apparently scans the Dockerfile for which files are _actually_ referenced by that Dockerfile and only sends those to the context. From my tests, this results in the same ~10MB build context when given an empty `.dockerignore` file.

If you'd like to run your own test, you can set `DOCKER_BUILDKIT=1 docker build ...`.

Take this with a grain of salt. I haven't used BuildKit before and I'm not sure why it isn't the default builder yet.

jgraettinger: > we could try avoid such micromanagement of this .dockerignore

We could also just .dockerignore `parser/target` and not worry about the rest

source-kinesis/Dockerfile
```diff
@@ -0,0 +1,42 @@
+# Build Stage
+################################################################################
+FROM golang:1.17-buster as builder
+
+WORKDIR /builder
+
+# Compile-time dependencies
```
jgraettinger, line 7: This is intended for removal ?

source-kinesis/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# RUN apt-get update \
+#  && apt-get install -y \
+#       ca-certificates \
+#  && apt-get clean
+
+# Download & compile dependencies. Doing this separately allows caching this
+# layer when no dependencies are updated.
+COPY go.* /builder/
```
jgraettinger, line 15: nit: /builder/ is already the WORKDIR.

`COPY go.* ./` here, and relative paths elsewhere in this Dockerfile ?

source-kinesis/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+WORKDIR /builder
+
+# Compile-time dependencies
+# RUN apt-get update \
+#  && apt-get install -y \
+#       ca-certificates \
+#  && apt-get clean
+
+# Download & compile dependencies. Doing this separately allows caching this
+# layer when no dependencies are updated.
```
jgraettinger, line 14: nit: comment isn't quite right. It won't cache in between CI runs having equal dependencies, but will cache between different built images of the _same_ CI run.

source-kinesis/Dockerfile
jgraettinger: Is this something we should do all the time, or something that can be used in a local checkout when debugging a connector ?

jgraettinger (commented):
approach LGTM, some pedantic nits

source-kinesis/Dockerfile
saterus: No, not in the final release images. I just wanted to point it out to people here since I found out about it. It seemed like a good opportunity to share knowledge after I pushed this with the `:debug` tag still in there from when I was poking around. :upside_down_face:

source-kinesis/Dockerfile
saterus: Yeah, it wasn't intended to live on in any final Dockerfiles that didn't need any dependencies. I just had it as a placeholder.

source-kinesis/Dockerfile
saterus: Yeah, I've been looking into CI layer caching too. It looks like it's possible with GitHub Actions' caching APIs. I'll be trying that out when I start working with CI.

.github/workflows/ci.yaml
```diff
@@ -7,64 +7,120 @@ on:
     branches: [main]
 
 jobs:
-  build:
-    services:
-      localstack:
-        # Provides local ephemeral versions of AWS services to use during tests.
```
saterus, line 13: I've removed this for now. After talking with Phil, I confirmed that our S3 connector tests are hitting the real S3 endpoints and this wasn't being used at all.

.github/workflows/ci.yaml
```diff
@@ -7,64 +7,120 @@ on:
...
+
+      - name: Build ${{ matrix.connector }} Docker Image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ${{ matrix.connector }}/Dockerfile
+          load: true
+          tags: ghcr.io/estuary/${{ matrix.connector }}:test
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
saterus, line 108: `cache-to`/`cache-from` are using the built-in GitHub Actions Cache API.

`mode=max` tells the cache to also cache layers from intermediate builds of a multi-stage Dockerfile. This allows us to cache the builder stage, and thus, all the specific library dependencies of each connector.

.github/workflows/ci.yaml
```diff
@@ -7,64 +7,120 @@ on:
...
+
+      - name: ${{ matrix.connector }} integration tests
+        if: matrix.connector != 'source-hello-world'
```
saterus, line 111: We can probably add integration tests for `source-hello-world`, but they don't easily slot into the skeleton we have right now. Given that we're also changing the test runner to handle materializations, I think we can come back to this.

.github/workflows/ci.yaml
```diff
@@ -7,64 +7,120 @@ on:
...
         env:
           GCP_SERVICE_ACCOUNT_KEY: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           DEFAULT_AWS_REGION: ${{ secrets.DEFAULT_AWS_REGION }}
```
saterus, line 116: One thing I'm not sure about is how to avoid a proliferation of environment variables being set. This will end up being the superset of all env vars needed by all connectors. Not something I'm super worried about right now, but it's not something I've found a simple answer to yet.

.github/workflows/ci.yaml
```diff
@@ -7,64 +7,120 @@ on:
...
-        shell: bash
-      - run: make push-all
-
+      - name: Push ${{ matrix.connector }} image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ${{ matrix.connector }}/Dockerfile
+          push: true
+          tags: ghcr.io/estuary/${{ matrix.connector }}:${{ steps.prep.outputs.tag }}
```
saterus, line 126: We tag any successful connector build with the git sha and push it. This *does not* push a `latest`/`dev`/whatever tag on `main` builds, but we were not doing that before either. 

Whenever we decide that we want to do that, I think we'd want to make that a distinct separate job that depends on `build_connectors`.

tests/run.sh
```diff
@@ -19,9 +19,9 @@ function bail() {
 test -n "$CONNECTOR" || bail "must specify CONNECTOR env variable"
 test -n "$VERSION" || bail "must specify VERSION env variable"
 
-FLOW_IMAGE="quay.io/estuary/flow:v0.1.0-376-g273cd3d"
+FLOW_IMAGE="ghcr.io/estuary/flow:v0.1.0-406-gebef86c"
 # the connector image needs to be available to envsubst
-export CONNECTOR_IMAGE="ghcr.io/estuary/${CONNECTOR}:${VERSION}" 
+export CONNECTOR_IMAGE="ghcr.io/estuary/${CONNECTOR}:${VERSION}"
```
saterus, line 24: One thing we might consider doing is replacing the pair of `CONNECTOR` & `VERSION` variables with directly expecting `CONNECTOR_IMAGE` to be set. I've found myself doing local docker builds without the `ghcr.io/estuary` prefixes (eg. `source-s3:dev` rather than `ghcr.io/estuary/source-s3:dev`), but then I can't actually run the connector tests using those local images.

Makefile
```diff
@@ -23,6 +23,9 @@ parser_cargo_args = --release --manifest-path parser/Cargo.toml --target x86_64-
 $(parser): parser
 	cargo build $(parser_cargo_args)
 
+# TODO: talk to Phil about make again
```
jgraettinger, line 26: nit: still TODO ?

tests/run.sh
jgraettinger: I'm 👍 to `CONNECTOR_IMAGE` (why have two variables when one will do?). Not in this PR though (let's 🚢 it!).

jgraettinger (approved):
LGTM! Would you please land when ready? I'd like to rebase on it.

Makefile
saterus: Yes, it was. Good catch. I just spoke with him about this. I wasn't sure about using `parser_cli` as an "alias task" of sorts.

As we spoke, we decided that since CI is handling running _all_ of the things entirely outside the Makefile, and running any _one_ of the tasks defined within was fairly easy "docker build" or "go test", the Makefile was probably no longer necessary. We can always bring it back if we want more coordination.

PR #26, opened 2021-09-03
Update go parser config to match rust declaration

psFried:
The go struct definition for the parser config had deviated from the definition on the rust side. This is to make them match again. This fixes a bug that prevented the csv quote character from being set in the capture config.


jgraettinger (approved):

saterus:
As a general question, should this be a protobuf definition? It sounds like the issue was a mismatch in data structures between the Rust and Go, which is generally a problem we tackle with protobufs, right?

psFried:
> As a general question, should this be a protobuf definition? It sounds like the issue was a mismatch in data structures between the Rust and Go, which is generally a problem we tackle with protobufs, right?

Maybe? I'm not really sure of the best way to handle this. I think perhaps the ideal would be to use something that can generate the Go definitions from the JSON schema. Protobuf would complicate the build for both go and rust, and getting the JSON serde working on the Rust side was (last I checked) a total pain. But even with all that, using protobuf might still be better than the current manual system. I think either way we should merge this as is and take that on in a follow up PR, since it's a big enough chunk of work that I'd like to clear out the rest of what I'm working on before starting that.

PR #28, opened 2021-09-08
materializations: factor Postgres, Snowflake, and Webhook from Flow

jgraettinger:
Add `materialize-boilerplate` and `testsupport` for implementing and
testing materialization connectors.

Add postgres, snowflake, and webhooks -- each extracted from the Flow
repo -- as connectors managed and built by this repository.

Issue #173


jgraettinger:
Any feedback on changes here ?

(As per our VC today, the build failure is due to docker caching races in the CI builder and not an actual failure)

jgraettinger:
Went ahead and merged. The failure is related to the PR-specific cache key -- will see if it also fails on `main` but I suspect not.

PR #30, opened 2021-09-09
Make run.sh failure less cryptic when running a second time

willdonnelly:
The rationale for not automatically deleting the directory makes
good sense, but giving actionable instructions and identifying the
specific path that needs to be deleted makes it less painful for
interactive use.


jgraettinger (approved):
LGTM

PR #33, opened 2021-09-10
Explicitly scope Docker layer caching

saterus:
It looks like we're running into intermittent problems with concurrent jobs all writing to the same cache entry. The recommended advice for now appears to be "parameterize the scope with your matrix name". I don't know if this is going to be fixed upstream at some point, but I'm curious if this fixes it for us.

I've tested this with `johnny/materializations` merged in. It seems to fix the issue, at least anecdotally. More runs will be the real test. I've had 3 successive, successful builds.


jgraettinger (approved):
LGTM

PR #35, opened 2021-09-15
Use "local" Docker layer caching strategy

saterus:
The native integration with the GitHub Actions Cache appears to be pretty flaky. There are several (possibly unrelated) issues with buildx that appear to prevent it from working reliably.

Let's try using the GitHub "cache" action to save/restore the layers as normal files.

Once again, the testing is a bit anecdotal. I built this, re-ran the workflow manually several times, pushed a cache-busting change, and then reverted it. Each time, the build seems to have passed as expected.


jgraettinger:
We're not doing this, in favor of just not caching for now right ?

PR #36, opened 2021-09-16
Remove Docker layer caching

saterus:
This has proved to be less reliable than initially hoped. Let's see how long the builds are now that they are parallelized. The job parallelism may prove to be fast enough in practice.


jgraettinger (approved):
Lol these build times are totally fine right now 🚀. LGTM

jgraettinger:
can this merge now?

PR #38, opened 2021-09-20
snowflake: add timeout logging and aborted attempt to drop snowsql

jgraettinger:
Improve timeout-related error messages as per estuary/flow#238

Attempt to switch to driver PUT support, but back out with comments due
to gosnowflake implementation bug.


snowzach:
LGTM

PR #41, opened 2021-09-23
Skip Malformed CSV Rows

saterus:
Adds the ability to configure an `errorThreshold` as part of the Parser's configuration. This allows it to be threaded through from the Capture's configuration within the Flow catalog. The Parser will absorb the parsing errors it encounters until the threshold is exceeded, at which point it returns all the errors it as seen thus far.

### DIFF
--- a/parser/src/config/mod.rs
+++ b/parser/src/config/mod.rs
@@ -235,6 +235,76 @@ impl schemars::JsonSchema for Compression {
     }
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+#[serde(transparent)]
+pub struct ErrorThreshold {
+    pub max_percent: u8,
+}
+
+impl ErrorThreshold {
+    pub fn new(percent: u64) -> Result<Self, ConfigError> {
+        if percent <= 100 {
+            Ok(Self {
+                max_percent: percent as u8,
+            })
+        } else {
+            Err(ConfigError::InvalidErrorThreshold(percent))
+        }
+    }
+
+    pub fn exceeded(&self, test_percent: u8) -> bool {
+        test_percent >= self.max_percent
+    }
+}
+
+impl Default for ErrorThreshold {
+    fn default() -> Self {
+        Self { max_percent: 0 }
+    }
+}
+
+impl<'de> Deserialize<'de> for ErrorThreshold {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        struct Visitor;
+
+        impl<'v> serde::de::Visitor<'v> for Visitor {
+            type Value = ErrorThreshold;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("expected a percentage value between 0 and 100")
+            }
+
+            fn visit_u64<E>(self, v: u64) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                ErrorThreshold::new(v).map_err(|e| E::custom(e))
+            }
+        }
+
+        deserializer.deserialize_u64(Visitor)
+    }
+}
+
+impl schemars::JsonSchema for ErrorThreshold {
+    fn schema_name() -> String {
+        "errorThreshold".to_string()
+    }
+    fn json_schema(_: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        serde_json::from_value(serde_json::json!({
+            "type": ["integer"],
+            "title": "errorThreshold",
+            "description": "The percentage of malformed rows which can be encountered without halting the parsing process. Only the most recent 1000 rows are used to calculate the error rate.",
+            "minimum": 0,
+            "maximum": 100,
+        }))
+        .unwrap()
+    }
+}
+
 #[derive(Clone, PartialEq, Debug, Serialize, Deserialize, schemars::JsonSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct ParseConfig {
@@ -303,6 +373,11 @@ pub struct ParseConfig {
     /// Configures handling of tab-separated values (TSV) format.
     #[serde(default)]
     pub tsv: Option<CharacterSeparatedConfig>,
+
+    /// Allows a percentage of errors to be ignored without failing the entire
+    /// parsing process. When this limit is exceeded, parsing halts.
+    #[serde(default)]
+    pub error_threshold: Option<ErrorThreshold>,
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -312,6 +387,9 @@ pub enum ConfigError {
 
     #[error("failed to parse config: {0}")]
     Parse(#[from] serde_json::Error),
+
+    #[error("ErrorThreshold cannot be greater than 100%: {0}%")]
+    InvalidErrorThreshold(u64),
 }
 
 impl ParseConfig {
@@ -384,6 +462,10 @@ impl ParseConfig {
             }
         }
 
+        if other.error_threshold.is_some() {
+            self.error_threshold = other.error_threshold.clone();
+        }
+
         self
     }
 
@@ -409,6 +491,7 @@ impl Default for ParseConfig {
             content_type_mappings: default_content_type_mappings(),
             csv: None,
             tsv: None,
+            error_threshold: None,
         }
     }
 }
--- a/parser/src/format/csv.rs
+++ b/parser/src/format/csv.rs
@@ -1,16 +1,19 @@
 //! Parsers for character-separated formats like csv.
-use super::{Format, Output, ParseError, Parser};
-use crate::config::{
-    csv::{CharacterSeparatedConfig, LineEnding},
-    ParseConfig,
-};
+use super::{Format, Output, ParseError, ParseResult, Parser};
 use crate::format::projection::{build_projections, collate, TypeInfo};
 use crate::input::{detect_encoding, Input};
+use crate::{
+    config::{
+        csv::{CharacterSeparatedConfig, LineEnding},
+        ParseConfig,
+    },
+    ErrorThreshold,
+};
 use csv::{Reader, StringRecord, Terminator};
 use doc::Pointer;
 use json::schema::types;
 use serde_json::Value;
-use std::io;
+use std::{collections::VecDeque, io};
 
 struct CsvParser {
     /// The specific format associated with this parser. Used to lookup the correct configuration
@@ -160,12 +163,15 @@ impl Parser for CsvParser {
         }
         tracing::info!("Resolved column headers: {:?}", columns);
 
-        Ok(Box::new(CsvOutput {
+        let csv_output = CsvOutput {
             headers: columns,
             reader,
             current_row: csv::StringRecord::new(),
             row_num: 0,
-        }))
+        };
+        let iterator =
+            ParseErrorBuffer::new(csv_output, config.error_threshold.unwrap_or_default());
+        Ok(Box::new(iterator))
     }
 }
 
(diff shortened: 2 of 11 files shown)


parser/src/format/mod.rs
```diff
@@ -253,6 +255,102 @@ fn compression_from_filename(filename: &str) -> Option<Compression> {
...
+                Ok(value) => return Some(Ok(value)),
+                Err(e) => {
+                    self.absorb(e);
+
+                    if self.is_full() || self.in_bad_region() {
+                        let limit = std::mem::take(&mut self.threshold);
+                        let errors = std::mem::take(&mut self.absorbed);
+                        return Some(Err(ParseError::ErrorLimitExceeded(limit, errors)));
+                    } else {
+                        tracing::error!(record_count = self.total_records, "failed to parse row");
```
psFried, line 345: I'm thinking this log should be at the `warn` level, and should include some more information in the fields. Eventually, the connector can forward those logs to Flow, which can publish them to the logs collection. The ideal end state being that there's a warning log in their logs collection that has the filename, line number, and the actual content of any line that's triggering the error. I'm not suggesting that you need to do all that in this PR, but if you at least get the errors logged with sufficient information, then we can work on the other pieces.

psFried (changes requested):
I had lots to say about this one, and I'm a little unsure whether it's all 100% coherent in the end. Happy to discuss if you think that'd be helpful.

parser/src/config/mod.rs
```diff
@@ -235,6 +235,76 @@ impl schemars::JsonSchema for Compression {
...
+    }
+}
+
+impl schemars::JsonSchema for ErrorThreshold {
+    fn schema_name() -> String {
+        "errorThreshold".to_string()
+    }
+    fn json_schema(_: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        serde_json::from_value(serde_json::json!({
+            "type": ["integer", "null"],
```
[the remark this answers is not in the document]
saterus: Agreed. I think I just misunderstood how to specifiy that this field was optional. I'll take a closer look.

parser/src/format/mod.rs
```diff
@@ -253,6 +255,102 @@ fn compression_from_filename(filename: &str) -> Option<Compression> {
...
+impl<I: Iterator> ParseErrorBuffer<I> {
+    /// Consumes the next item out of the inner iterator.
+    fn advance(&mut self) -> Option<I::Item> {
+        // How many of the recent records to consider when trying to decide if
+        // we've entered a new region of bad data of the file.
+        const ERROR_BUFFER_WINDOW_SIZE: usize = 1000;
+
+        self.total_records += 1;
+
+        if self.total_records > ERROR_BUFFER_WINDOW_SIZE {
```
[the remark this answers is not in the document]
saterus: Hmm, yeah. I got too clever. :facepalm: Plus my test cases hid the problem. 

I'll take a crack at the rate limiter approach. I was trying to avoid having to store data per row processed. I think this will be a pretty similar to how it is implemented now, but hopefully less broken.

parser/src/format/mod.rs
```diff
@@ -253,6 +255,102 @@ fn compression_from_filename(filename: &str) -> Option<Compression> {
...
+    }
+
+    /// Returns true if we've exceeded the error threshold for the entire file.
+    fn is_full(&self) -> bool {
+        self.exceeded_limit(self.absorbed.len())
+    }
+
+    /// Returns true if we've exceeded the error threshold for the most recent records.
+    fn in_bad_region(&self) -> bool {
+        self.exceeded_limit(self.errors_in_latest_window)
```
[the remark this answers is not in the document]
saterus: Yeah, 100%. I think I introduced this bug late in the game when I did a bit of refactoring.

parser/src/format/mod.rs
```diff
@@ -140,15 +147,10 @@ fn format_output(
...
-            tracing::warn!(
-                record_count = record_count,
-                "parsing failed after {} records",
-                record_count
-            );
-            e
-        })?;
+    let error_threshold = config.error_threshold.unwrap_or_default();
+
+    for result in ParseErrorBuffer::new(output, error_threshold) {
```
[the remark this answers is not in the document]
saterus: Okay, yeah. That should be doable. I think my initial thought was that malformed records are a problem for any file type. Depending on exactly what it is, there might be different error parsing characteristics, but the possibility of bad data remains.

Ex. Given a `jsonl` file where a row is improperly formatted by being pretty printed into the stream:
```jsonl
{"id": 0, "color": "green"}
{
  "id": 1,
  "color": "blue"
}
{"id": 2, "color": "red"}
```
We'd encounter line 0, parse it correctly. Then line 1,2,3,4 would all fail to parse. Then line 5 would succeed. We'd count that "one" malformed document as 4 `ParseError`s, but we *would* be able to skip over it.

My worry here isn't so much that serde is left in an invalid state, but that deeply nested documents might _parse_ correctly but on an inner subset document (though I suspect that would probably fail schema validation).

If we're trying to parse a file as a _single_ json document and that document is malformed, there's no meaningful usage for an error threshold greater than 0. If _the document_ fails to parse, it fails.

Now, all that said, I don't know if we know how many other file formats should handle malformed records yet. I'm fine with moving the error buffer into the csv module.

> it seems like error threshold that really only makes sense for [...] certain error types. A decompression error should always be treated as terminal, for example.

I wasn't sure about treating all types of `ParseError`s as recoverable/skippable, but I didn't try to differeniate them yet. Your decompression example is a perfect case. I can add a check against a whitelist of the `ParseError`s we allow to be skipped.

parser/src/format/mod.rs
saterus: I waffled back and forth about `warn` vs `error`. It *is* an error, but we're skipping it. :upside_down_face: 

I wasn't sure how much info we'd want to output in the event we're skipping the row. I can log the entire error which includes specific span information. I can also add a bit more tracing context elsewhere so that this includes more information (the file name).

parser/src/format/mod.rs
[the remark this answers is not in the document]
saterus: Upon further reflection, the token bucket approach makes understanding the error detection quite opaque. The algorithm ended up being very dependent upon initial conditions, which makes it rather difficult to understand if it was working correctly.

As discussed, now we just track the exact number of errors within the window at any given point and fail if we have seen too many errors.

parser/src/format/mod.rs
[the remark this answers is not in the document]
saterus: >  I can add a check against a whitelist of the `ParseError`s we allow to be skipped.

Upon further inspection, I punted on this. I believe that `ParseError::Parse(_)` is the only variant that we _can_ skip and it's on the only variant we can even produce at this stage of parsing. It could be worth revisiting if we want to differentiate finer-grain parsing problems, but for now, I think we're probably okay.

parser/src/format/mod.rs
saterus: I looked into this a bit more. It looks like we are already including the file name in the span information. 

Here's a sample of the log output for a malformed row:

```json
{
  "timestamp": "Sep 30 15:27:26.815",
  "level": "WARN",
  "fields": {
    "message": "failed to parse row",
    "item": "Parse(InvalidContent(Error(UnequalLengths { pos: Some(Position { byte: 664, line: 62, record: 61 }), expected_len: 2, len: 3 })))"
  },
  "target": "parser::format::csv",
  "span": {
    "parse_args": "ParseArgs { config_file: Some(\"/tmp/jsonl-parser-test.JcgEhXN5qFwT/config.json\"), file: \"-\" }",
    "name": "do_parse"
  }
}
```

If we pass the file as stdin (as with tests) you end up getting `"file": "-"`, but it should already be there.

Anyway, I think we can probably do a separate round of enriching this log data so that it's easier for an end user to consume, but we probably want to talk more about that.

saterus:
@psFried Whenever you get a chance, I think this is ready for another look. Thanks!

parser/src/format/csv.rs
```diff
@@ -367,6 +373,110 @@ impl Iterator for CsvOutput {
...
+    pub fn new(inner: I, threshold: ErrorThreshold) -> Self {
+        let mut buf = Self {
+            inner,
+            threshold,
+            total_records: 0,
+            errors_in_buffer: 0,
+            buffer: VecDeque::with_capacity(ERROR_BUFFER_WINDOW_SIZE),
+        };
+
+        buf.prefill_buffer();
```
psFried, line 407: This feels a little tricky to be within a `new` function, since we generally expect iterators to be lazy. It seems preferable to make it lazily fill the buffer if needed. But it might also work to just rename `new` to something like `fill_buffer`.

parser/src/format/csv.rs
```diff
@@ -367,6 +373,110 @@ impl Iterator for CsvOutput {
...
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            if self.exceeded() {
+                return Some(Err(ParseError::ErrorLimitExceeded(self.threshold)));
+            } else {
+                let item = self.advance()?;
+                if item.is_ok() {
+                    return Some(item);
+                } else {
+                    tracing::warn!(item=?item.unwrap_err(), "failed to parse row");
```
psFried, line 472: nit: Feels like the field in the logs should be called `error` instead of `item`.

parser/src/main.rs
```diff
@@ -112,7 +112,9 @@ where
         match self {
             Ok(t) => t,
             Err(e) => {
-                tracing::debug!("Error details: {:#?}", e);
+                // QUESTION: Since we're expecting to parse this output as logs,
```
psFried, line 115: Yeah, I think this seems like a good idea. At a minimum, it would help to standardize as much as possible on field names for things like filename and line number in the error representations.

parser/tests/csv_error_absorption_test.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    });
+    run_test(&config, csv.as_input()).assert_success(800);
+
+    // Increase the error rate to 100%. 200 errors are in the buffer. 250 will
+    // cross the threshold. As we write these Bad rows, Good rows from the first
+    // 1000 will fall out of the other side of the buffer, but so will some of
+    // the old Bad rows. If we want to end up with 25% of 1000 inside the
+    // window, we'll need to account for 1/5 of the old errors dropping out.
+    // 200 + 60 - 12 = 248 + 3 = 251. 63 new Bad rows put us over the threshold.
```
psFried, line 77: These were really helpful comments :+1:

psFried (approved):
Looks great! I left a few comments, but should be good to merge once those are addressed.

parser/src/format/csv.rs
saterus: Yeah, I went back and forth on this a bit. 

On one hand, it doesn't really operate as expected if you forget to call a `fill_buffer` function after `new` but before `next`. The struct is technically initialized, but not _really_ properly setup. 

On the other, as you pointed out, iterators are generally lazy. This one is a bit weird, in that even if we did `prefill_buffer` at the top of `next`, we'd be lazily triggering the read 1000 items. Kind of a mismatch of expectation there too. At some point, I had `prefill_buffer` there, but I moved it since once the buffer was filled the first time, it always preserves that filled buffer until the inner iterator is empty and it starts draining.

On balance, I think I'm leaning towards your point about laziness. That should be easy to swap.

parser/src/main.rs
saterus: I've moved the error formatting to tracing's field syntax, rather than directly interpolating them into the message. 

<details>
  <summary>Examples of tracing output:</summary>

```jsonl
{
  "timestamp": "Oct 01 11:22:56.814",
  "level": "WARN",
  "fields": {
    "message": "failed to parse row",
    "error": "Parse(InvalidContent(Error(UnequalLengths { pos: Some(Position { byte: 664, line: 62, record: 61 }), expected_len: 2, len: 3 })))"
  },
  "target": "parser::format::csv",
  "span": {
    "parse_args": "ParseArgs { config_file: Some(\"/tmp/jsonl-parser-test.SaI2L21nhmbp/config.json\"), file: \"-\"}",
    "name": "do_parse"
  },
  "spans": [
    {
      "parse_args": "ParseArgs { config_file: Some(\"/tmp/jsonl-parser-test.SaI2L21nhmbp/config.json\"), file: \"-\" }",
      "name": "do_parse"
    }
  ]
}
{
  "timestamp": "Oct 01 11:22:56.815",
  "level": "DEBUG",
  "fields": {
    "error_details": "ErrorLimitExceeded(ErrorThreshold { max_percent: 25 })",
    "message": "parsing failed"
  },
  "target": "parser",
  "span": {
    "parse_args": "ParseArgs { config_file: Some(\"/tmp/jsonl-parser-test.SaI2L21nhmbp/config.json\"), file: \"-\" }",
    "name": "do_parse"
  },
  "spans": [
    {
      "parse_args": "ParseArgs { config_file: Some(\"/tmp/jsonl-parser-test.SaI2L21nhmbp/config.json\"), file: \"-\" }",
      "name": "do_parse"
    }
  ]
}
{
  "timestamp": "Oct 01 11:22:56.815",
  "level": "ERROR",
  "fields": {
    "error": "error limit exceeded",
    "message": "parsing failed"
  },
  "target": "parser",
  "span": {
    "parse_args": "ParseArgs { config_file: Some(\"/tmp/jsonl-parser-test.SaI2L21nhmbp/config.json\"), file: \"-\" }",
    "name": "do_parse"
  },
  "spans": [
    {
      "parse_args": "ParseArgs { config_file: Some(\"/tmp/jsonl-parser-test.SaI2L21nhmbp/config.json\"), file: \"-\" }",
      "name": "do_parse"
    }
  ]
}
```
</details>

This is the minimal amount of json formatting. The errors themselves are still the default `Debug` format for Rust structs/enums. If we want to make these errors even easier to parse, I think we'll want to go back and implement `Serialize` on our error types.

parser/tests/csv_error_absorption_test.rs
saterus: Haha, good to hear it. I wasn't sure if anyone else was going to be able to follow it or not. :sweat_smile: Writing these out was critical for me to trust that it was working correctly.

PR #46, opened 2021-10-06
Refactor due to changes in protocol repo to make Endpoint an interface.

snowzach:
The protocols/materialize repo was refactored to make Endpoint an interface. This allows for making SQLish databases that do not have support for sql.DB drivers compatible with the existing SQL generation code. This updates the Postgres and Snowflake drivers to support this new Endpoint interface.


snowzach:
I've done checks locally against Postgres. If there's anything else I should be checking (besides what's done in CI here) let me know and I will do that as well.

PR #60, opened 2021-10-29
Materialization docs

snowzach:


docs/materialize/README.md
```diff
@@ -0,0 +1,288 @@
...
+Flow Checkpoints are used to delineate pieces of work. Operations are completed using these delineations
+such that if there is a failure, it will start processing again at the last completed checkpoint of data. 
+Flow indicates to a connector that it is starting a particular checkpoint and once the connector confirms the
+checkpoint is complete/committed, Flow will register and consider that a completed piece of work.
+
+Driver Checkpoints are optional values that the connector/driver can register with Flow for tracking state
+information. Flow can provide this information back to the connector if there is a failure once it is restarted.
+Read below for more information about the Driver Checkpoint values.
+
+## RPCs
```
oliviamiannone, line 75: I want to clarify some word order/comma related stuff. So, the 3 RPCs are Spec, Validate, and Apply. And this paragraph refers to checks, verifications, setup, streaming. Looking at ordering I want to assume that spec=checks, verification+setup=validate, and streaming=apply? But that doesn't seem right

oliviamiannone: (This is the only thing, from a factual/wording standpoint, that I was confused about in this draft. Will follow up with plans for how I'd like to move forward tomorrow!)

snowzach: The Spec is basically like running the "help" command. 
Validate really does checks/verification
Apply does "setup"
And streaming is when it is actually sending the data which is the last step that goes on while there is still data. 
Hopefully that helps.

PR #62, opened 2021-11-01
Log all parser output to connector stderr

psFried:
Changes the code for invoking the parser from Go so that the parser logs
will be printed to the stderr of the connector. This allows the
logs from both the connector and the parser to appear in the logs
collections of captures.

This also updates the parser to log with flattened fields, so that log
the forwarded logs will look a little nicer in Flow.


jixij (approved):
:+1:

PR #65, opened 2021-11-05
Create `materialize-rockset`

saterus:
I think the majority of the churn is behind me, so I wanted to put this up for a bit of review. I've left a bunch of TODO comments throughout where I have questions, unresolved issues, or things I'd like to come back to.

As always, reviewing commit-by-commit is probably going to be easiest. 

Please let me know when I'm doing something strange in go land - it all still feels pretty strange. :upside_down_face:

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/65)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -68,8 +68,9 @@ jobs:
           - materialize-bigquery
           - materialize-elasticsearch
           - materialize-postgres
-          - materialize-snowflake
+          - materialize-rockset
           - materialize-s3-parquet
+          - materialize-snowflake
           - materialize-webhook
 
     steps:
@@ -144,6 +145,8 @@ jobs:
           file: ${{ matrix.connector }}/Dockerfile
           load: true
           tags: ghcr.io/estuary/${{ matrix.connector }}:test
+          secrets: |
+            "rockset_api_key=${{ secrets.ROCKSET_API_KEY }}"
 
       - name: Start Dockerized test infrastructure
         if: matrix.connector == 'source-kafka'
@@ -166,7 +169,7 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           DEFAULT_AWS_REGION: ${{ secrets.DEFAULT_AWS_REGION }}
           AWS_DEFAULT_OUTPUT: json
-
+          ROCKSET_API_KEY: ${{ secrets.ROCKSET_API_KEY }}
           # Compare to: https://www.postgresql.org/docs/current/libpq-envars.html
           PGDATABASE: flow
           PGHOST: localhost
--- a/flowsim/matsim/basic.go
+++ b/flowsim/matsim/basic.go
@@ -39,13 +39,13 @@ func (c *BasicConfig) Execute(args []string) error {
 	if err != nil {
 		return fmt.Errorf("test catalog: %v", err)
 	}
-	catalog, err := testcat.BuildCatalog(c.ctx, testCatalog)
+	materializationSpecs, err := testcat.BuildMaterializationSpecs(c.ctx, testCatalog)
 	if err != nil {
-		return fmt.Errorf("build catalog: %v", err)
+		return fmt.Errorf("build materialization specs: %v", err)
 	}
 
 	// There will only be one materialization for this test.
-	materialization := &catalog.Materializations[0]
+	materialization := materializationSpecs[0]
 
 	// Perform the connector setup (spec, validate, apply) and open the transactions stream.
 	stream, err := SetupConnectorOpenTransactions(c.ctx, materialization, materialize.AdaptServerToClient(c.driverServer), false)
--- a/flowsim/matsim/basic.go
+++ b/flowsim/matsim/basic.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/estuary/connectors/flowsim/testcat"
 	"github.com/estuary/connectors/flowsim/testdata"
-	"github.com/estuary/flow/go/materialize"
 	pm "github.com/estuary/protocols/materialize"
 	log "github.com/sirupsen/logrus"
 )
@@ -48,7 +47,7 @@ func (c *BasicConfig) Execute(args []string) error {
 	materialization := materializationSpecs[0]
 
 	// Perform the connector setup (spec, validate, apply) and open the transactions stream.
-	stream, err := SetupConnectorOpenTransactions(c.ctx, materialization, materialize.AdaptServerToClient(c.driverServer), false)
+	stream, err := SetupConnectorOpenTransactions(c.ctx, materialization, pm.AdaptServerToClient(c.driverServer), false)
 	if err != nil {
 		return fmt.Errorf("setup connector: %v", err)
 	}
@@ -118,6 +117,9 @@ func (c *BasicConfig) Execute(args []string) error {
 
 		// Wait until previous commit has completed if it's still running.
 		commitWait.Wait()
+		if err := Acknowledge(stream); err != nil {
+			return err
+		}
 
 		// Previously committed passStore is now available to put into the store again as the commit on it is complete.
 		store.Push((<-returnToStore).Range()...)
--- a/flowsim/matsim/delta.go
+++ b/flowsim/matsim/delta.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/estuary/connectors/flowsim/testcat"
 	"github.com/estuary/connectors/flowsim/testdata"
-	"github.com/estuary/flow/go/materialize"
 	pm "github.com/estuary/protocols/materialize"
 	log "github.com/sirupsen/logrus"
 )
@@ -47,7 +46,7 @@ func (c *DeltaConfig) Execute(args []string) error {
 	materialization := materializationSpecs[0]
 
 	// Perform the setup and open the transactions stream
-	stream, err := SetupConnectorOpenTransactions(c.ctx, materialization, materialize.AdaptServerToClient(c.driverServer), true)
+	stream, err := SetupConnectorOpenTransactions(c.ctx, materialization, pm.AdaptServerToClient(c.driverServer), true)
 	if err != nil {
 		return err
 	}
@@ -96,7 +95,6 @@ func (c *DeltaConfig) Execute(args []string) error {
 
 	// Main prepare/store/commit loop.
 	for loop := 0; loop < c.Loops; loop++ {
-
 		// The number of documents we will store this pass.
 		docCount := (1 << loop) * c.Start
 
@@ -104,6 +102,9 @@ func (c *DeltaConfig) Execute(args []string) error {
 
 		// Wait until previous commit has completed if it's still running.
 		commitWait.Wait()
+		if err := Acknowledge(stream); err != nil {
+			return err
+		}
 
 		// Write the Prepared request with the checkpoint information.
 		var checkpoint = fmt.Sprintf(`"checkpoint loop %d"`, loop)
@@ -170,7 +171,6 @@ func (c *DeltaConfig) Execute(args []string) error {
 			// sit in this channel until the next loading cycle has completed.
 			returnToStore <- store
 		}(loop, passStore)
-
 	}
 
 	// Wait until the final commit has completed.
@@ -191,5 +191,4 @@ func (c *DeltaConfig) Execute(args []string) error {
 	log.Infof("validation completed: duration: %v", time.Since(start))
 
 	return nil
-
 }
--- a/flowsim/matsim/delta.go
+++ b/flowsim/matsim/delta.go
@@ -38,12 +38,13 @@ func (c *DeltaConfig) Execute(args []string) error {
 	if err != nil {
 		return fmt.Errorf("test catalog: %v", err)
 	}
-	catalog, err := testcat.BuildCatalog(c.ctx, testCatalog)
+	materializationSpecs, err := testcat.BuildMaterializationSpecs(c.ctx, testCatalog)
 	if err != nil {
-		return fmt.Errorf("build catalog: %v", err)
+		return fmt.Errorf("build materialization specs: %v", err)
 	}
 
-	materialization := &catalog.Materializations[0]
+	// There will only be one materialization for this test.
+	materialization := materializationSpecs[0]
 
 	// Perform the setup and open the transactions stream
 	stream, err := SetupConnectorOpenTransactions(c.ctx, materialization, materialize.AdaptServerToClient(c.driverServer), true)
--- a/flowsim/matsim/delta.go
+++ b/flowsim/matsim/delta.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math/rand"
 	"os"
 	"sync"
 	"time"
@@ -69,7 +70,13 @@ func (c *DeltaConfig) Execute(args []string) error {
 	}
 	log.WithFields(log.Fields{"checkpoint": string(opened.Opened.FlowCheckpoint)}).Info("connection opened")
 
+	// The store will contain our documents for caching and comparison at the end.
+	var store = testdata.NewStore()
 	var commitWait sync.WaitGroup
+
+	// After a loop is committed, this channel will return committed items to the store.
+	var returnToStore = make(chan *testdata.Store, 1)
+	returnToStore <- testdata.NewStore() // preload with an empty store for the first loop.
 
 	start := time.Now()
 
@@ -113,18 +120,30 @@ func (c *DeltaConfig) Execute(args []string) error {
 		}
 		log.Infof("prepared: checkpoint: %s", string(prepared.DriverCheckpointJson))
 
-		// Generate documents to store.
-		var store = testdata.NewStore()
+		store.Push((<-returnToStore).Range()...)
+		// Build a passStore of documents to load and then store, some of them will exist and others won't.
+		var passStore = testdata.NewStore()
 		var exists []bool
+
 		for i := 0; i < docCount; i++ {
-			store.Push(c.newTestData())
-			exists = append(exists, false)
+			// If nothing is in the store or at random pick a get TestData to update in the batch.
+			if store.Len() == 0 || rand.Intn(2) == 0 {
+				item := c.newTestData()
+				passStore.Push(item)
+				exists = append(exists, false)
+			} else {
+				// Pop an element from the store to the batch to update.
+				item := store.Pop()
+				item.UpdateValues()
+				passStore.Push(item)
+				exists = append(exists, true)
+			}
 		}
 
 		storeStart := time.Now()
 
 		// Send the Store requests.
-		if err = SendStore(stream, store.Range(), exists, c.BatchSize); err != nil {
+		if err = SendStore(stream, passStore.Range(), exists, c.BatchSize); err != nil {
 			return fmt.Errorf("store error: %v", err)
 		}
 		log.Infof("store: duration: %v", time.Since(storeStart))
@@ -146,16 +165,23 @@ func (c *DeltaConfig) Execute(args []string) error {
 			}
 			log.Infof("committed: loop: %d duration: %v pass duration: %v", loop, time.Since(commitStart), time.Since(passStart))
 
-			// Once committed, encode to the output location
-			for _, item := range store.Range() {
-				outputEncoder.Encode(item)
-			}
-		}(loop, store)
+			// Once committed, return the cached/modified items back to the store. They will
+			// sit in this channel until the next loading cycle has completed.
+			returnToStore <- store
+		}(loop, passStore)
 
 	}
 
 	// Wait until the final commit has completed.
 	commitWait.Wait()
+
+	store.Push((<-returnToStore).Range()...)
+
+	// Once committed, encode to the output location
+	log.Warnf("Manual verification of test results required:")
+	for _, item := range store.Range() {
+		outputEncoder.Encode(item)
+	}
 
 	if err := stream.CloseSend(); err != nil {
 		return fmt.Errorf("close: %v", err)
--- a/flowsim/matsim/fence.go
+++ b/flowsim/matsim/fence.go
@@ -6,7 +6,7 @@ import (
 	"time"
 
 	"github.com/estuary/connectors/flowsim/testcat"
-	"github.com/estuary/flow/go/materialize"
+	"github.com/estuary/protocols/materialize"
 	pm "github.com/estuary/protocols/materialize"
 	log "github.com/sirupsen/logrus"
 )
@@ -62,7 +62,7 @@ func (c *FenceConfig) Execute(args []string) error {
 	}
 
 	// Perform the connector setup (spec, validate, apply) and open the second, overlapping transactions stream.
-	stream2, err := SetupConnectorOpenTransactions(c.ctx, materialization, materialize.AdaptServerToClient(c.driverServer), false)
+	stream2, err := SetupConnectorOpenTransactions(c.ctx, materialization, pm.AdaptServerToClient(c.driverServer), false)
 	if err != nil {
 		return err
 	}
--- a/flowsim/matsim/fence.go
+++ b/flowsim/matsim/fence.go
@@ -29,12 +29,12 @@ func (c *FenceConfig) Execute(args []string) error {
 	if err != nil {
 		return fmt.Errorf("test catalog: %v", err)
 	}
-	catalog, err := testcat.BuildCatalog(c.ctx, testCatalog)
+	materializationSpecs, err := testcat.BuildMaterializationSpecs(c.ctx, testCatalog)
 	if err != nil {
-		return fmt.Errorf("build catalog: %v", err)
+		return fmt.Errorf("build materialization specs: %v", err)
 	}
 
-	materialization := &catalog.Materializations[0]
+	materialization := materializationSpecs[0]
 
 	// Perform the connector setup (spec, validate, apply) and open the first transactions stream.
 	stream1, err := SetupConnectorOpenTransactions(c.ctx, materialization, materialize.AdaptServerToClient(c.driverServer), false)
--- a/flowsim/matsim/fs_adapter.go
+++ b/flowsim/matsim/fs_adapter.go
@@ -56,7 +56,7 @@ func (fsa flowSinkAdapter) Validate(ctx context.Context, req *pm.ValidateRequest
 }
 
 // Apply applies the configuration.
-func (fsa flowSinkAdapter) Apply(ctx context.Context, req *pm.ApplyRequest) (*pm.ApplyResponse, error) {
+func (fsa flowSinkAdapter) ApplyUpsert(ctx context.Context, req *pm.ApplyRequest) (*pm.ApplyResponse, error) {
 	var source = new(image.EndpointSpec)
 	if err := req.Validate(); err != nil {
 		return nil, fmt.Errorf("validating request: %w", err)
@@ -67,7 +67,11 @@ func (fsa flowSinkAdapter) Apply(ctx context.Context, req *pm.ApplyRequest) (*pm
 	req.Materialization = proto.Clone(req.Materialization).(*pf.MaterializationSpec)
 	req.Materialization.EndpointSpecJson = source.Config
 
-	return fsa.DriverServer.Apply(ctx, req)
+	return fsa.DriverServer.ApplyUpsert(ctx, req)
+}
+
+func (fsa flowSinkAdapter) ApplyDelete(ctx context.Context, req *pm.ApplyRequest) (*pm.ApplyResponse, error) {
+	panic("ApplyDelete not adapted")
 }
 
 // Transactions implements the DriverServer interface.
--- a/flowsim/matsim/matsim.go
+++ b/flowsim/matsim/matsim.go
@@ -135,7 +135,6 @@ func (c *Config) ParseConfig() error {
 // NewTestCatalog returns a simple catalog with a single collection using the schema of your TestData
 // and a single materialization setup to use FlowSync with your materialization.
 func (c *Config) NewTestCatalog() (*testcat.TestCatalog, error) {
-
 	var collection, err = testcat.BuildCollection(c.newTestData())
 	if err != nil {
 		return nil, err
@@ -168,7 +167,6 @@ func (c *Config) NewTestCatalog() (*testcat.TestCatalog, error) {
 
 // resolveYamlOrFile takes a string and tries to parse it as yaml or if it can't assumes it's a yaml file.
 func resolveYamlOrFile(in string) (testcat.ConfigMap, error) {
-
 	var out = make(testcat.ConfigMap)
 	if err := yaml.Unmarshal([]byte(in), &out); err == nil {
 		return out, nil
--- a/flowsim/matsim/matsim.go
+++ b/flowsim/matsim/matsim.go
@@ -10,6 +10,7 @@ import (
 	"github.com/brianvoe/gofakeit/v6"
 	"github.com/estuary/connectors/flowsim/testcat"
 	"github.com/estuary/connectors/flowsim/testdata"
+	"github.com/estuary/flow/go/flow/ops"
 	"github.com/estuary/flow/go/materialize/driver/image"
 	pm "github.com/estuary/protocols/materialize"
 	"github.com/jessevdk/go-flags"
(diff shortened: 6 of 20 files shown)


jgraettinger (commented):
Comments below. They're all nits, or opportunities to improve things a bit. No show stoppers.

In general, comments in this could use a little love, but I'm guessing that's because of this PR's draft status? If you haven't already, see [commentary](https://golang.org/doc/effective_go#commentary)

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 4 of 4 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 2 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/65#-:-Mo4fQFL5lbnhEuSj5EM:b-vb8kwi)* status: all files reviewed, 14 unresolved discussions (waiting on @saterus and @snowzach)

___
*[materialize-rockset/Dockerfile, line 17 at r6](https://reviewable.io/reviews/estuary/connectors/65#-Mo4f22H3-CWDkrzvEKj:-Mo4f22H3-CWDkrzvEKk:b-wezc39) ([raw file](https://github.com/estuary/connectors/blob/c1f02baaa26ffda17f332e479d0ac4eb60281d23/materialize-rockset/Dockerfile#L17)):*
> ```Dockerfile
> 
> # Test and build the connector.
> # RUN go test  -tags nozstd -v ./materialize-rockset/...
> ```

Where do we intend to run these tests from, now ? Would it help to thread-through the environment variable ?
___
*[materialize-rockset/driver.go, line 227 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo41Idl2VDLD7yEOntQ:-Mo41Idl2VDLD7yEOntR:b-b19to1) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/driver.go#L227)):*
> ```Go
> 
> 	if err = stream.Send(&pm.TransactionResponse{
> 		// TODO: should I be doing anything with this FlowCheckpoint? I don't *think* that's appropriate for Rockset...
> ```

You've got it right here. The shard recovery log (and not rockset) is the authoritative "commit" of the transaction, so this should be nil.
___
*[materialize-rockset/driver.go, line 238 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo41cNN9a_Qlu9dUajz:-Mo41cNN9a_Qlu9dUak-:b-998jnr) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/driver.go#L238)):*
> ```Go
> }
> 
> func RandString(len int) string {
> ```

nit: is this intended to be exported ? should it be in driver_test.go ?
___
*[materialize-rockset/rest_client.go, line 81 at r2](https://reviewable.io/reviews/estuary/connectors/65#-Mo3yFgB85AGUsIZ3iqy:-Mo3yFgC3QVZXpQu73Zs:b-8vpm3r) ([raw file](https://github.com/estuary/connectors/blob/a8cae4c1d394ce7f78e82235e7e671fa7f1cfb51/materialize-rockset/rest_client.go#L81)):*
> ```Go
> 	req.Header.Add("Accept", headerAccept)
> 	req.Header.Add("Authorization", c.apiKey)
> 	req.Header.Add("Content-Length", fmt.Sprintf("%d", contentLength))
> ```

nit: I believe Content-Length is added automatically, as is User-Agent.
___
*[materialize-rockset/rest_client.go, line 93 at r2](https://reviewable.io/reviews/estuary/connectors/65#-Mo3yVYLAJO6eXVzVJ0e:-Mo3yVYLAJO6eXVzVJ0f:bxbczbk) ([raw file](https://github.com/estuary/connectors/blob/a8cae4c1d394ce7f78e82235e7e671fa7f1cfb51/materialize-rockset/rest_client.go#L93)):*
> ```Go
> 	}
> 
> 	// Do the request
> ```

nit: as a reader, comments in this function are already apparent from the code.
___
*[materialize-rockset/rest_client.go, line 173 at r2](https://reviewable.io/reviews/estuary/connectors/65#-Mo3zRmT2JWtV3cbQ20d:-Mo3zRmT2JWtV3cbQ20e:b-cy8jx8) ([raw file](https://github.com/estuary/connectors/blob/a8cae4c1d394ce7f78e82235e7e671fa7f1cfb51/materialize-rockset/rest_client.go#L173)):*
> ```Go
> 	}
> 	defer resp.Body.Close()
> 	if resp.StatusCode != http.StatusOK {
> ```

Consider having makeRequest read out the full response body ahead of time (returning bytes?), and have it map !StatusOK into its error return value.

This would probably simplify it's usage. If you want to check for cases like 404, you can have it map into ErrNotFound that's then tested against elsewhere.

If you're always POST-ing JSON, you can also pass the post body as an interface and handle the JSON encode within it as well.
___
*[materialize-rockset/rockset.go, line 26 at r3](https://reviewable.io/reviews/estuary/connectors/65#-Mo4-Q4E9YqkkFyrN7jF:-Mo4-Q4F5QSEZjl_mlVx:b-57kggv) ([raw file](https://github.com/estuary/connectors/blob/274e03cf1d356a4eb85a0ce89e931d9acca3f28f/materialize-rockset/rockset.go#L26)):*
> ```Go
> 		return nil, nil
> 	} else if err != nil {
> 		return nil, fmt.Errorf("failed to fetch workspace `%s`: %w", workspace, err)
> ```

nit: for style consistency with our codebase, fmt.Errorf() should provide context about where an error happened but not mark it as "failed". A log output would do that. For example:

```
err = fmt.Errorf("frobbing the quib: %w", fmt.Errorf("mixing the batch: %w", err))
log.WithError(err).Error("failed to fetch workspace")
```
___
*[materialize-rockset/transactor.go, line 27 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo42zys2BYaH5iMrAXR:-Mo42zys2BYaH5iMrAXS:bj9k43f) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L27)):*
> ```Go
> type binding struct {
> 	spec       *pf.MaterializationSpec_Binding
> 	operations map[operation][]json.RawMessage
> ```

nit: comment on what this structure represents?
___
*[materialize-rockset/transactor.go, line 47 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo435i0BoaNL1eKiOgC:-Mo435i0BoaNL1eKiOgD:b-zbt3tc) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L47)):*
> ```Go
> // pm.Transactor
> func (t *transactor) Load(it *pm.LoadIterator, _ <-chan struct{}, loaded func(int, json.RawMessage) error) error {
> 	// Nothing to load
> ```

Panic here, because Load is not expected to be called, correct ?

Just seeking to understand: What _would_ it take to do loads against rockset ? Could we stage a bunch of keys to load, and have it retrieve them efficiently ?
What read-your-writes guarantee does the invocation of the write API provide -- on OK response are we guaranteed to be able to query out the written document ?
___
*[materialize-rockset/transactor.go, line 65 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo4TW1W3yAUJjSUjEYW:-Mo4TW1W3yAUJjSUjEYX:bnt9n1w) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L65)):*
> ```Go
> 
> 	for it.Next() {
> 		select {
> ```

I believe you can drop this |t.ctx|, not tracking it on the |transactor| at all.

Idiomatic practice in Go is to _not_ track a context as a field (though we do it sometimes for $reasons), and to instead pass it as a parameter. In this case, |it| already has the context through the stream it contains, and a context cancellation will bubble up through it.Next().

(Edit to add: actually, I see that you'd still need it on transactor in order to pass it to Commit. We should update the Transactor interface to thread through an appropriate Context. I'll look at that, as I'm in there right now...)
___
*[materialize-rockset/transactor.go, line 127 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo4ZXswDwYRazvE02SJ:-Mo4ZXswDwYRazvE02SK:b-7dn50e) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L127)):*
> ```Go
> }
> 
> func (t *transactor) storeAdditionOperation(b *binding, key *tuple.Tuple, values *tuple.Tuple) error {
> ```

nit: key and value can be by-value rather than *by-reference, right ?
___
*[materialize-rockset/transactor.go, line 146 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo4_P50BP5LHZp30rH4:-Mo4_P50BP5LHZp30rH5:b-t66otr) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L146)):*
> ```Go
> 		if containsString(inference.Types, "object") {
> 			if inference.IsSingleType() {
> 				// TODO: casting the tuple to a string isn't the right way of doing this.
> ```

There's a simpler way to do this (which I had to go code spelunking to recall): The crucial invariant with our tuples encoding is that JSON-encoded representations always pack/unpack as []byte, where string is always a non-encoded regular string. [Here's the specific](https://github.com/estuary/flow/blob/master/crates/tuple/src/value.rs) place where we encode tuple values. [Here's an example](https://github.com/estuary/protocols/blob/main/fdb/tuple/tuple_test.go#L63-L81) of what unpacked values look like from the Go side.

So, what you can instead do here is not worry about the inference type at all. Instead, do a runtime type check of whether value is a `[]byte`. If it is, assume it's a json.RawMessage. Otherwise, just add it to the `document` directly and let json.Marshal deal with it.
___
*[materialize-rockset/transactor.go, line 113 at r8](https://reviewable.io/reviews/estuary/connectors/65#-Mo4gHX_3QfFuu_zQCnp:-Mo4gHX_3QfFuu_zQCnq:bwrxzj5) ([raw file](https://github.com/estuary/connectors/blob/785dd9b2ac2a162bae482418855def5eee5ed577/materialize-rockset/transactor.go#L113)):*
> ```Go
> 		go func(b *binding) {
> 			defer wait.Done()
> 			commitCollection(t, b)
> ```

This error value isn't being used and should be.

I'll suggest using https://pkg.go.dev/golang.org/x/sync/errgroup for this. Gazette's Task abstraction uses it under the hood.
___
*[materialize-rockset/cmd/matsim/matsim.go, line 41 at r8](https://reviewable.io/reviews/estuary/connectors/65#-Mo4fKJpEYa0zTAJ3is5:-Mo4fKJpEYa0zTAJ3is6:b6jxcrt) ([raw file](https://github.com/estuary/connectors/blob/785dd9b2ac2a162bae482418855def5eee5ed577/materialize-rockset/cmd/matsim/matsim.go#L41)):*
> ```Go
> 
> func (e *FakeEvent) UpdateValues() {
> 	// if random.Intn(5) == 0 {
> ```

nit: cleanup ?

saterus (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/65)* status: 0 of 17 files reviewed, 11 unresolved discussions (waiting on @jgraettinger and @snowzach)

___
*[materialize-rockset/driver.go, line 227 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo41Idl2VDLD7yEOntQ:-MooambME2bdxAQtuvzb:b-896fix) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/driver.go#L227)):*

You've got it right here. The shard recovery log (and not rockset) is the authoritative "commit" of the transaction, so this should be nil.
</blockquote></details>

Done.
___
*[materialize-rockset/driver.go, line 238 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo41cNN9a_Qlu9dUajz:-MooanSODZijD_Lpdulo:b-kwbfwp) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/driver.go#L238)):*

nit: is this intended to be exported ? should it be in driver_test.go ?
</blockquote></details>

It's used in both `driver_tests.go` and `matsim.go` to generate simple random values. It's in the driver for historical reasons (at some point the driver was the entire connector), but I'm not sure if there's really a better place for it right now. :shrug:
___
*[materialize-rockset/rest_client.go, line 81 at r2](https://reviewable.io/reviews/estuary/connectors/65#-Mo3yFgB85AGUsIZ3iqy:-MoobFb65auQ_1aqrqzP:bqe72u5) ([raw file](https://github.com/estuary/connectors/blob/a8cae4c1d394ce7f78e82235e7e671fa7f1cfb51/materialize-rockset/rest_client.go#L81)):*

nit: I believe Content-Length is added automatically, as is User-Agent.
</blockquote></details>

After some digging into the http package, you're entirely right. What threw me off was that `http.DumpRequest` doesn't actually print these. I was trying to verify the sizes of the requests I was sending and wasn't seeing it in the logs.
___
*[materialize-rockset/rest_client.go, line 173 at r2](https://reviewable.io/reviews/estuary/connectors/65#-Mo3zRmT2JWtV3cbQ20d:-MoobZ8lDzf_Qt_76vQY:b-nawr8o) ([raw file](https://github.com/estuary/connectors/blob/a8cae4c1d394ce7f78e82235e7e671fa7f1cfb51/materialize-rockset/rest_client.go#L173)):*

Consider having makeRequest read out the full response body ahead of time (returning bytes?), and have it map !StatusOK into its error return value.

This would probably simplify it's usage. If you want to check for cases like 404, you can have it map into ErrNotFound that's then tested against elsewhere.

If you're always POST-ing JSON, you can also pass the post body as an interface and handle the JSON encode within it as well.
</blockquote></details>

I tried to move a lot more of this logic into `makeRequest`. I'm not sure if it's what you were looking for or not, but it did simplify the individual callers.
___
*[materialize-rockset/rockset.go, line 26 at r3](https://reviewable.io/reviews/estuary/connectors/65#-Mo4-Q4E9YqkkFyrN7jF:-MoobqIBCuR7rgpBBSHn:b800sw9) ([raw file](https://github.com/estuary/connectors/blob/274e03cf1d356a4eb85a0ce89e931d9acca3f28f/materialize-rockset/rockset.go#L26)):*

nit: for style consistency with our codebase, fmt.Errorf() should provide context about where an error happened but not mark it as "failed". A log output would do that. For example:

```
err = fmt.Errorf("frobbing the quib: %w", fmt.Errorf("mixing the batch: %w", err))
log.WithError(err).Error("failed to fetch workspace")
```
</blockquote></details>

This was really helpful advice. :thumbsup: 
___
*[materialize-rockset/transactor.go, line 47 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo435i0BoaNL1eKiOgC:-MoobxAT5H8EqNPC4i7z:bk913qf) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L47)):*
> Panic here, because Load is not expected to be called, correct ?

:thumbsup:

>Just seeking to understand: What would it take to do loads against rockset ? 

They provide an endpoint that you can send SQL queries to and they'll return the results as json. They _only_ support named query parameters, not positional parameters. We'll have to update the sql generator to output those queries, but we can do that. 

> Could we stage a bunch of keys to load, and have it retrieve them efficiently ? What read-your-writes guarantee does the invocation of the write API provide -- on OK response are we guaranteed to be able to query out the written document ?

I haven't done any performance testing on this endpoint, but presumably we can load a lot of records this way. Rockset's docs are silent on any guarantees with the Write API. 
___
*[materialize-rockset/transactor.go, line 127 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo4ZXswDwYRazvE02SJ:-MoodZocCbx2g659N5Yq:b-yij1z8) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L127)):*

nit: key and value can be by-value rather than *by-reference, right ?
</blockquote></details>

_Do_ I want do pass it by value? Wouldn't that effectively copy the entire document unnecessarily? We're already going to be copying the bytes when we decode the values into the `DocumentBuffer`, but I didn't think we'd want/need to copy them calling this function.
___
*[materialize-rockset/transactor.go, line 146 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo4_P50BP5LHZp30rH4:-MoodzxhA-_vrWOgJmw2:bbvbjxn) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L146)):*

There's a simpler way to do this (which I had to go code spelunking to recall): The crucial invariant with our tuples encoding is that JSON-encoded representations always pack/unpack as []byte, where string is always a non-encoded regular string. [Here's the specific](https://github.com/estuary/flow/blob/master/crates/tuple/src/value.rs) place where we encode tuple values. [Here's an example](https://github.com/estuary/protocols/blob/main/fdb/tuple/tuple_test.go#L63-L81) of what unpacked values look like from the Go side.

So, what you can instead do here is not worry about the inference type at all. Instead, do a runtime type check of whether value is a `[]byte`. If it is, assume it's a json.RawMessage. Otherwise, just add it to the `document` directly and let json.Marshal deal with it.
</blockquote></details>

This was super helpful. I was fairly certain I wasn't doing this right. :sweat_smile:

saterus:
@jgraettinger I think this is ready for another look. Thanks for all the helpful feedback. :raised_hands:

jgraettinger (changes requested):
> Reviewed 19 of 19 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/65)* status: all files reviewed, 10 unresolved discussions (waiting on @jgraettinger and @saterus)

___
*[materialize-rockset/document_buffer.go, line 14 at r9](https://reviewable.io/reviews/estuary/connectors/65#-Mqz_nB66BSmNwudgx9g:-Mqz_nB66BSmNwudgx9h:b-lmn79) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/document_buffer.go#L14)):*
> ```Go
> 
> // Creates a new buffer. This should be used sparingly, as the buffer is
> // intended to be reused after being emptied with `SplitN`.
> ```

nit: comment is incorrect ; you need to call Clear() right ? What about `SplitN` => `IntoSplits` which also does truncation before return ?
___
*[materialize-rockset/driver.go, line 23 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzaI_zE5oivoJBwQub:-MqzaHij1GD4_eNkOB6H:b-v0oneh) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/driver.go#L23)):*
> ```Go
> 	// Enable verbose logging of the HTTP calls to the Rockset API.
> 	HttpLogging bool `json:"http_logging"`
> 	// Field name which contains CDC change types are located. Values should be one of `["Insert", "Update", "Delete"]`
> ```

Right, this changed out from under you as part of https://github.com/estuary/connectors/issues/81

It's now 'c', 'u', and 'd' for compatibility with Debezium. Here's an example of usage:
https://github.com/estuary/connectors/blob/main/source-postgres/testdata/TestReplicationDeletes.snapshot

In practice this is always located under`/_meta/op`. We might want to make it a `const`.
___
*[materialize-rockset/driver.go, line 87 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzcWXh3Fbh_qfONvr6:-MqzcWXi3dYhKadEBMnb:bq522x3) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/driver.go#L87)):*
> ```Go
> }
> 
> func resourceFromJson(serialized *json.RawMessage) (*resource, error) {
> ```

consider using `pf.UnmarshalStrict` instead of this function
___
*[materialize-rockset/rockset.go, line 87 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzdVTF30gRDyL8Kuau:-MqzdVTF30gRDyL8Kuav:b-p4ti1y) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/rockset.go#L87)):*
> ```Go
> 				return nil, fmt.Errorf("fetching collection `%s/%s` while waiting for it to be created: %w", workspace, collection.Name, err)
> 			} else if col.Status != "READY" {
> 				// Not ready yet, so wait a couple of seconds then we'll try again.
> ```

nit: log to keep the user in the loop about why the invocation is still running ?
___
*[materialize-rockset/transactor.go, line 127 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo4ZXswDwYRazvE02SJ:-Mqzg2orEohYKkiIVCz9:b-mgbfxv) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L127)):*
<details><summary><i>Previously, saterus (Alex Burkhart) wrote…</i></summary><blockquote>

_Do_ I want do pass it by value? Wouldn't that effectively copy the entire document unnecessarily? We're already going to be copying the bytes when we decode the values into the `DocumentBuffer`, but I didn't think we'd want/need to copy them calling this function.
</blockquote></details>

tuple.Tuple is an array type, which means "passing by value" is actually copying the _reference_ to the array but not the array itself. It's more efficient than passing by *tuple.Tuple, because the latter requires the compiler to put in extra "what if this is nil?" checks
___
*[materialize-rockset/transactor.go, line 66 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzdskH4XpujUXDF6rk:-MqzdskH4XpujUXDF6rl:b-9r5ycp) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/transactor.go#L66)):*
> ```Go
> 
> type transactor struct {
> 	// TODO: ctx will be removed when the protocol updates land.
> ```

nit: this has already happened.
___
*[materialize-rockset/transactor.go, line 88 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzeHChAHfMTw_dswUE:-MqzeHCi5W8AE89ljqA5:b-dm3f8v) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/transactor.go#L88)):*
> ```Go
> 	for it.Next() {
> 		select {
> 		case <-t.ctx.Done():
> ```

nit: think you can drop this ; if the context is cancelled, Next will quickly return false.
___
*[materialize-rockset/transactor.go, line 98 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzfHXI1s2LXq7rqmV6:-MqzfHXI1s2LXq7rqmV7:b-ov0dzn) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/transactor.go#L98)):*
> ```Go
> 		var op operation = opNotFound
> 		if t.config.ChangeIndicator == "" {
> 			op = opDocumentUpsert
> ```

as noted above, these values have changed on you. Should we hard-code `/_meta/op` and assume Upsert behavior if it's not set ?

Right now there's some user friction in needing to know ahead of time whether a binding is operating in CDC mode or not, and it seems like that can be resolved by inspecting that location on a per-document basis.
___
*[materialize-rockset/transactor.go, line 163 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzidgA9Mt3Wsd3czEd:-MqzidgA9Mt3Wsd3czEe:bopttyy) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/transactor.go#L163)):*
> ```Go
> 
> func findStringField(b *binding, keys *tuple.Tuple, values *tuple.Tuple, targetFieldName string) string {
> 	// Look for the field in the document's keys.
> ```

I'm uncertain of this approach because of the tight coupling to materialization fields. THere are also tools like https://github.com/buger/jsonparser but I'm unclear that's better. Still thinking on this...

jgraettinger (changes requested):
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/65)* status: all files reviewed, 10 unresolved discussions (waiting on @jgraettinger and @saterus)

___
*[materialize-rockset/transactor.go, line 163 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzidgA9Mt3Wsd3czEd:-MqzsjH7-QVpzuTBcQXl:bsezukn) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/transactor.go#L163)):*

I'm uncertain of this approach because of the tight coupling to materialization fields. THere are also tools like https://github.com/buger/jsonparser but I'm unclear that's better. Still thinking on this... 
</blockquote></details>

Right, part of what's bothering me is I think we're attempting to handle deletions in the wrong place.

There's a "pruning" schema annotation that's been considered but unimplemented for a while now, which I hope we'll actually tackle next quarter. The core idea is that a JSON schema annotation would be able to annotate documents, or sublocations of documents, which are eligble for removal.

It would be plumbed through into the materialization protocol: alongside documents to Store (or perhaps within the StoreIterator) there'd be a boolean indicating that a particular key should be deleted.

If we had that, then CDC operations are just a pattern of annotation we'd use ("prune" is true if /_meta/op is const: "d"), and it's also a more general mechanism.

....

So. Wrapping up: I think we should simply ignore CDC operation types altogether for the moment, and treat everything as a rockset upsert. We can come back at this when we've wired up prune annotations and have incorporated deletions into the materialization protocol.

saterus:
> So. Wrapping up: I think we should simply ignore CDC operation types altogether for the moment, and treat everything as a rockset upsert. We can come back at this when we've wired up prune annotations and have incorporated deletions into the materialization protocol.

While the pruning annotation sounds like a promising general approach, does it fit with our desire to use CDC -> Rockset as a use-case in the short term? If we don't process the deletions, the data that ends up in Rockset will likely be incorrect. Would it be better in the short term to use the Debezium-style operation types to process deletions, knowing that we'll change over to the pruning annotation in the future?

jgraettinger:
I'm not _too_ worried about it because you can still filter deletions at query time (by materializing `op` a filtering value `d`). This is a gray-area call, but IMO better to not implement it and use its absence as motivation for implementing the desired long-term solution, vs implement a solution we're quickly wanting to get rid of / must migrate from.

saterus (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/65)* status: all files reviewed, 10 unresolved discussions (waiting on @jgraettinger and @saterus)

___
*[materialize-rockset/document_buffer.go, line 14 at r9](https://reviewable.io/reviews/estuary/connectors/65#-Mqz_nB66BSmNwudgx9g:-Mr2shqACeD4mQYJrMAz:byp984d) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/document_buffer.go#L14)):*

nit: comment is incorrect ; you need to call Clear() right ? What about `SplitN` => `IntoSplits` which also does truncation before return ?
</blockquote></details>

The comment was definitely outdated. I've added additional comments on the state machine within DocumentBuffer. I can't call `Clear` until after the caller of `SplitN` is finished or we'll remove the documents before they are finished with them. The chunks behind handed out here are all backed by the same buffer to avoid unnecessarily copying the documents, so if we clear it too soon, the documents disappear from the chunks.
___
*[materialize-rockset/driver.go, line 23 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzaI_zE5oivoJBwQub:-Mr2tg_rB3gCvpMtR8my:bqbarti) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/driver.go#L23)):*

Right, this changed out from under you as part of https://github.com/estuary/connectors/issues/81

It's now 'c', 'u', and 'd' for compatibility with Debezium. Here's an example of usage:
https://github.com/estuary/connectors/blob/main/source-postgres/testdata/TestReplicationDeletes.snapshot

In practice this is always located under`/_meta/op`. We might want to make it a `const`.
</blockquote></details>

I've removed the change operation stuff for now based on your feedback below. We can come back to it in the future.
___
*[materialize-rockset/driver.go, line 87 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzcWXh3Fbh_qfONvr6:-Mr2tqyD9R-xeB02KJK_:b4c9lyr) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/driver.go#L87)):*

consider using `pf.UnmarshalStrict` instead of this function
</blockquote></details>

:thumbsup: 
___
*[materialize-rockset/rockset.go, line 87 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzdVTF30gRDyL8Kuau:-Mr2ttRT2ez-j7QnQAum:b1jxr33) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/rockset.go#L87)):*

nit: log to keep the user in the loop about why the invocation is still running ?
</blockquote></details>

Good idea. In production it probably wouldn't be noticeable, but in dev, it's a helpful reminder why your `activate` command is not finished.
___
*[materialize-rockset/transactor.go, line 127 at r4](https://reviewable.io/reviews/estuary/connectors/65#-Mo4ZXswDwYRazvE02SJ:-Mr2uJlAAYYuUKi-aDIr:bjzpb4f) ([raw file](https://github.com/estuary/connectors/blob/d3ef944e2d197bf6a15e636e52a8bd28ae86651f/materialize-rockset/transactor.go#L127)):*

tuple.Tuple is an array type, which means "passing by value" is actually copying the _reference_ to the array but not the array itself. It's more efficient than passing by *tuple.Tuple, because the latter requires the compiler to put in extra "what if this is nil?" checks
</blockquote></details>

I forgot that a Tuple was just a slice. :+1: 
___
*[materialize-rockset/transactor.go, line 66 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzdskH4XpujUXDF6rk:-Mr2u5Rn7OUKv2XOV22b:b-egr8nf) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/transactor.go#L66)):*

nit: this has already happened.
</blockquote></details>

Yeah, I was still using this because of `Store`, but I've removed that now.
___
*[materialize-rockset/transactor.go, line 88 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzeHChAHfMTw_dswUE:-Mr2uAvHCpcqvrUuqbbo:b-ahle5g) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/transactor.go#L88)):*

nit: think you can drop this ; if the context is cancelled, Next will quickly return false.
</blockquote></details>

Good to know.

jgraettinger (commented):
I didn't realize this was waiting on me. LGTM!

> Reviewed 1 of 1 files at r10, 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/65#-:-MrYjJMQClwxoQWxMK9D:b-jkc5t0)* status: all files reviewed, 4 unresolved discussions (waiting on @jgraettinger and @saterus)

___
*[materialize-rockset/transactor.go, line 66 at r9](https://reviewable.io/reviews/estuary/connectors/65#-MqzdskH4XpujUXDF6rk:-MrYj7RwByY-Nlf-HlZV:b-7wax9a) ([raw file](https://github.com/estuary/connectors/blob/09ce6894d50aac86d74b10ac26b36ab4a4c44f18/materialize-rockset/transactor.go#L66)):*
<details><summary><i>Previously, saterus (Alex Burkhart) wrote…</i></summary><blockquote>

Yeah, I was still using this because of `Store`, but I've removed that now.
</blockquote></details>

fyi the StoreIterator has a Context accessor.

PR #67, opened 2021-11-08
Phil/w3c logs

### LINKED ISSUE #64: Support W3C Extended Log file format in parser
We'd like to be able to parse files in the [W3C extended log file format](https://www.w3.org/TR/WD-logfile.html). This format is  similar to the existing TSV format, so ideally this would share some of that existing code. A quick search of crates.io hasn't returned any libraries for parsing this format, so it looks like we'll have to roll our own.

psFried:
- Adds parser support for the W3C Extended Log file format
- Refactored `csv.rs` into a directory with multiple files
- Fixes type inference in the parser so that it can account for dynamic property names

resolves #64


parser/src/format/character_separated/w3c_extended_log.rs
```diff
@@ -0,0 +1,181 @@
...
+        let header =
+            W3cLogParser::parse_header(&prefix).map_err(|e| ParseError::Parse(Box::new(e)))?;
+        // TODO: (optionally) add the extra directives from header as fields to each json object.
+
+        let projections = build_projections(config)?;
+        let headers = resolve_headers(header.fields, projections, &["-"]);
+
+        let reader = csv::ReaderBuilder::new()
+            .delimiter(b'\t')
+            .comment(Some(b'#'))
```
saterus, line 96: This is clever (by telling the csv parser that header rows are comments, we don't have to worry about skipping ahead in the input stream manually), maybe worthy of a comment.

parser/src/format/projection.rs
```diff
@@ -37,17 +44,90 @@ pub enum BuildError {
...
+/// input.
+#[derive(Debug)]
+pub struct Projections {
+    /// Type inferrence information from the root JSON schema.
+    shape: Shape,
+    /// Projections stored with field names exactly as they are in the config/schema.
+    case_sensitive: BTreeMap<String, Rc<Pointer>>,
+    /// Projections stored with field names that have been normalized and converted to lowercase,
+    /// to allow for case-insensitive matching against field names. Technically, these _could_ all
+    /// be included in one big map, but
```
saterus, line 57: I think the end of this comment got lost...

parser/src/format/projection.rs
```diff
@@ -247,11 +312,85 @@ mod test {
...
+                "required": ["locationa", "bee"],
+
             }),
             ..Default::default()
         };
-        let result = build_projections(&config).expect("failed to build projections");
-        insta::assert_debug_snapshot!(result);
+        let projections = build_projections(&config).expect("failed to build projections");
+
+        fn type_info(ty: Option<types::Set>, must_exist: bool, ptr: &str) -> Projection {
```
saterus, line 328: One rename escaped!

saterus (approved):
This looks really good. It's nice how you were able to mostly leverage the stock `csv` parser library. :fireworks:

parser/src/format/character_separated/w3c_extended_log.rs
```diff
@@ -0,0 +1,181 @@
...
+use super::{resolve_headers, CsvOutput};
+use crate::config::ParseConfig;
+use crate::format::projection::build_projections;
+use crate::format::{Output, ParseError, Parser};
+use crate::input::Input;
+use std::io::BufRead;
+
+// TODO: Put a span in the error so we can display the source.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
```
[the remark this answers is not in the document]
psFried: When I added that, I was working with a different approach, where I was using `nom` to parse the headers, which would have made it easy to put span info here. But it ended up being more involved than seemed worth it. I just removed this todo.

Generally, though, it might be nice to someday revisit parser error handling to include spans where possible. I came across the [miette project](https://github.com/zkat/miette) a little while ago, which seemed like a nice way to do that.

parser/src/format/character_separated/w3c_extended_log.rs
psFried: good call, I added a comment.

parser/src/format/projection.rs
psFried: whoops, fixed.

parser/src/format/projection.rs
psFried: fixed

PR #68, opened 2021-11-08
source-postgres: Rewrite to use watermarks instead of LSNs

### LINKED ISSUE #53: source-postgres: Ensure correctness of captured data
We have reason to believe that the transition between table-scanning backfill and replication event processing is subtly incorrect and could miss or duplicate change events in edge cases, especially if there are concurrent "random-access-ish" modifications during the table scan.

The core of the problem is that, in the absence of writes, there's no way to precisely line up the transaction which reads a bunch of data via `SELECT` queries with the sequence of change events we receive via replication. Currently we sample the value of `pg_current_wal_lsn()` at the start of the read transaction and treat that as "the point at which the table was scanned", but this function isn't itself transactional, so not only is there a bit of a race condition there but there's also no guarantee that this LSN corresponds _at all_ with the version of the table we're seeing.

We have anecdotal evidence that Debezium-based CDC solutions can exhibit subtle errors in the captured data, and this is a strong candidate for why that might be (since Debezium uses this same approach).

There are two avenues we might try for a more robust solution:

- PostgreSQL replication must solve this problem somehow, or it wouldn't be possible to add replicas to a live database. We should investigate how exactly this works (probably snapshots?) and whether we could do the same thing.
- Note how this whole problem was qualified with "in the absence of writes" up above. If the transaction which reads data out of the table also emitted _any_ writes, then we could observe where those writes show up in the replication stream and thereby get an actual point of correspondence between the two processes. This is essentially what Netflix's [DBLog](https://netflixtechblog.com/dblog-a-generic-change-data-capture-framework-69351fb9099b) system does, and has the advantage of being reasonably database-agnostic. But it requires that the Flow system have write access to the database, at least for one table.

### LINKED ISSUE #55: source-postgres: Backfill latency improvements
As currently written, the `source-postgres` connector must perform a table backfill to completion before it can resume replication. When the tables are reasonably small this is fine, but if they grow large enough that this backfill takes hours or even days this has two major implications:

1. Adding a new table will essentially pause the capture of all other tables until the backfill finishes.
2. The new table will also not be replicated until the backfill finishes.

The first problem could likely be solved by some refactoring to allow the table-scanning process and replication event processing to be interleaved. There already has to be logic around filtering change events from not-yet-scanned portions of the table, so the only major change is that there needs to be some way to _very briefly_ pause replication processing while each individual chunk is read out. But I suspect we'll need this sooner or later.

The second problem almost feels like a "well what did you expect?" sort of situation, but it's not exactly. The problem is that there may be absurdly gigantic tables in which the majority of the data is old and will never be modified. It would be nice if we could immediately jump to the new data, scan and then start replication on that, and then have a separate batch process which will slowly grind through the older data. This could be provided by some sort of "threshold" feature so we could express things like "only capture rows whose `created_at` column is more recent than `<date>`".

willdonnelly:
This is pretty much the design discussed in #53, although currently backfill queries for multiple tables are issued serially rather than concurrently.

This resolves #53 and also resolves #55 due to how the rewrite interleaves backfills with replication event processing.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/68)


jgraettinger (changes requested):
A bunch of feedback and questions below, but high level I'm super enthusiastic about this PR. Looking good.

> Reviewed 25 of 31 files at r1, 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/68#-:-Mo5pTNF0js70R13etp7:b-m7uavv)* status: all files reviewed, 16 unresolved discussions (waiting on @willdonnelly)

___
*[source-postgres/capture.go, line 105 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5DxiN60I2D4mAvCnR:-Mo5DxiN60I2D4mAvCnS:b-468stu) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L105)):*
> ```Go
> // (if any) are met.
> func RunCapture(ctx context.Context, config *Config, catalog *airbyte.ConfiguredCatalog, state *PersistentState, dest messageOutput) error {
> 	logrus.WithField("uri", config.ConnectionURI).WithField("slot", config.SlotName).Info("starting capture")
> ```

optional nit: logrus.WithFields(logrus.Fields{"url": ..., "slot": ...}).

I'll usually import logrus as log, but that's a preference.
___
*[source-postgres/capture.go, line 166 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5EQwp5VbE-2pBb88t:-Mo5EQwq7EU3c97KgJdA:b8a9ykt) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L166)):*
> ```Go
> 
> 	err = c.streamChanges(ctx)
> 	logrus.WithField("err", err).Debug("change streaming terminated")
> ```

nit: grammar
___
*[source-postgres/capture.go, line 298 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5G6ax0mm5x6U3LcGF:-Mo5G6ax0mm5x6U3LcGG:b-feo2w2) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L298)):*
> ```Go
> 	// state updates on every transaction commit.
> 	logrus.Info("all streams active")
> 	for evt := range c.replStream.Events() {
> ```

Could this be expressed as streaming to a fictional watermark which will never be matched?
___
*[source-postgres/capture.go, line 322 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5Gj-F0DT3thGRgUp6:-Mo5Gj-F0DT3thGRgUp7:b7ik0nw) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L322)):*
> ```Go
> 	var watermarkReached = false
> 	for evt := range c.replStream.Events() {
> 		// Commit events update the current LSN and are otherwise ignored, except for the
> ```

_Should_ these be ignored (as opposed to emitting state updates)?

In my current understanding, the downside is that the connector may produce a lot of documents in between emitted checkpoints, which could impact its latency.

Is it a problem to emit checkpoints which continue to update the read-through LSN? It would cause the present backfill chunk to be restarted from a further-forward LSN, but that seems okay (even desirable) ?

Note it's possible (but not required) to emit the LSN and the scan states as separate partial checkpoints, which the runtime will merge/reduce. The actual checkpoint bytes can be small.
___
*[source-postgres/capture.go, line 379 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5cSdf94J1Why_TXXB:-Mo5cSdf94J1Why_TXXC:bmhiga9) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L379)):*
> ```Go
> 	// Emit any buffered results and update table states accordingly.
> 	for _, streamID := range results.Streams() {
> 		var evts = results.Changes(streamID)
> ```

nit: `events` and `event`.  Non-standard abbreviation impacts readability without saving much.
___
*[source-postgres/capture.go, line 403 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5eIjf-50CRv6isE89:-Mo5eIjf-50CRv6isE8A:bmzw3im) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L403)):*
> ```Go
> 	var results = newResultSet()
> 
> 	// TODO(wgd): Add a sanity-check assertion that the current watermark value
> ```

It could model scans as an outer join of the watermarks table with the core scan query, and have an extra computed column with a "watermark is the correct value" bit. This seems worth doing IMO, but happy to punt for this PR.
___
*[source-postgres/capture.go, line 407 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5fDxu1V2lyujZEgWR:-Mo5fDxu1V2lyujZEgWS:bv78prl) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L407)):*
> ```Go
> 	// than it's worth until we have other evidence of correctness violations though.
> 
> 	// TODO(wgd): We can dispatch these table reads concurrently with a WaitGroup
> ```

FYI for when you get to it: https://pkg.go.dev/golang.org/x/sync/errgroup

Further helps by modeling errors and subtask failure propagation.
___
*[source-postgres/main.go, line 79 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5fgQREP822NvWcxXh:-Mo5fgQREP822NvWcxXi:bvom9jp) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/main.go#L79)):*
> ```Go
> 			"type":        "string",
> 			"title":       "Watermarks Table",
> 			"description": "The name of the table used for 'watermark' writes during backfilling",
> ```

nit: grammar, "during backfills" or "while backfilling" ?  Also not sure 'watermark' should be quoted.
___
*[source-postgres/replication.go, line 85 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5jXpFALQ_MPoWRDPU:-Mo5jXpFALQ_MPoWRDPV:b-da9yqt) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/replication.go#L85)):*
> ```Go
> 		relations: make(map[uint32]*pglogrepl.RelationMessage),
> 		// standbyStatusDeadline is left uninitialized so an update will be sent ASAP
> 		events: make(chan *changeEvent),
> ```

From a performance perspective, I suspect you want a reasonable buffer here so that the message decode vs handling loops can run in parallel.
___
*[source-postgres/replication.go, line 120 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5gQGU-2e6NW1Uz4Fu:-Mo5gQGU-2e6NW1Uz4Fv:bnf0t0n) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/replication.go#L120)):*
> ```Go
> }
> 
> func (s *replicationStream) CurrentLSN() pglogrepl.LSN {
> ```

I'm not following this: why does capture.go call CurrentLSN() to asynchronously fetch the LSN now, vs using an LSN attached to the changeEvent as before ? This seems racy, where previously it wasn't.
___
*[source-postgres/replication.go, line 158 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5jnNoCkp_sRn2W9-6:-Mo5jnNoCkp_sRn2W9-7:b7xxlsn) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/replication.go#L158)):*
> ```Go
> 		// If there's already a change event which needs to be sent to the consumer,
> 		// try to do so until/unless the context expires first.
> 		if s.eventBuf != nil {
> ```

Is this n=1 buffer behavior so that CurrentLSN() doesn't get ahead of what the reader's seeing ?
___
*[source-postgres/replication.go, line 230 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5iartD13vGojlDMLj:-Mo5iartD13vGojlDMLk:b-o6rvni) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/replication.go#L230)):*
> ```Go
> 		}
> 		s.inTransaction = false
> 		atomic.StoreUint64(&s.currentLSN, uint64(msg.TransactionEndLSN))
> ```

Seems like you could define semantics of a re-added `changeEvent.LSN` to be that it's ONLY set if Type == "Commit".
___
*[source-postgres/resultset.go, line 20 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5k8i9-iZNaphCLT3h:-Mo5k8i9-iZNaphCLT3i:b-wzc0u4) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/resultset.go#L20)):*
> ```Go
> type backfillChunk struct {
> 	rows     map[string]*changeEvent
> 	scanKey  []string
> ```

nit: comments on what each of these are please.
___
*[source-postgres/resultset.go, line 48 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5kd0RBJkznSE1GoSh:-Mo5kd0RBJkznSE1GoSi:b-nz8toq) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/resultset.go#L48)):*
> ```Go
> 
> 	// Otherwise add thew new row to the `rows` map and update `scanned`
> 	for _, evt := range evts {
> ```

nit: here too, `events` and `event`
___
*[source-postgres/snapshot.go, line 22 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5ocN3BxC172dT1mIQ:-Mo5ocN4EprZZgPkI1du:bi6yg2) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/snapshot.go#L22)):*
> ```Go
> // `prevKey` if non-nil. This is the entrypoint to the database-specific portion of the
> // backfill process.
> func scanTable(ctx context.Context, conn *pgx.Conn, streamID string, scanKey []string, prevKey []byte) ([]*changeEvent, error) {
> ```

Names are confusing here. I _believe_ `scanKey` is fields which constitute the primary key, while `prevKey` is the tuple-encoded key that was left off at, correct?

`keyColumns` vs `startKey` ?

Also would `scanTableChunk` be a more accurate name?
___
*[source-postgres/snapshot.go, line 23 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5mBL0DvqcqfHh_U5S:-Mo5mBL0DvqcqfHh_U5T:b3zk3fj) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/snapshot.go#L23)):*
> ```Go
> // backfill process.
> func scanTable(ctx context.Context, conn *pgx.Conn, streamID string, scanKey []string, prevKey []byte) ([]*changeEvent, error) {
> 	var snapshot, err = snapshotDatabase(ctx, conn)
> ```

`snapshotDatabase` is still fetching the approximate LSN. That can be removed ?

Do we even _need_ an explicit read transaction anymore ?

I'm sure this is on your future refactoring radar, but `tableSnapshot` feels a bit clunky now, like the whole thing could be lifted into the body of this function.
___
*[source-postgres/snapshot.go, line 43 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5n7d43UChCbsd-0GM:-Mo5n7d43UChCbsd-0GN:bjl108u) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/snapshot.go#L43)):*
> ```Go
> 	var wm = uuid.New().String()
> 
> 	var query = fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s (slot TEXT PRIMARY KEY, watermark TEXT);", table)
> ```

nit: can this use a $1 placeholder ? (thinking about SQL injection).

willdonnelly (commented):
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/68)* status: all files reviewed, 11 unresolved discussions (waiting on @jgraettinger)

___
*[source-postgres/capture.go, line 105 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5DxiN60I2D4mAvCnR:-MoAAnqVDxvMjXXyNtr6:bak0s4) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L105)):*

optional nit: logrus.WithFields(logrus.Fields{"url": ..., "slot": ...}).

I'll usually import logrus as log, but that's a preference.
</blockquote></details>

Yeah, I tend to favor using imports by their default names just because it makes the automatic imports editing work more reliably. Happy to do a quick search-and-replace if you'd prefer `log` though.
___
*[source-postgres/capture.go, line 298 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5G6ax0mm5x6U3LcGF:-MoADferA6X9c2eeJBSc:bcfqwkr) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L298)):*

Could this be expressed as streaming to a fictional watermark which will never be matched?
</blockquote></details>

Yes, after the related change to emit state updates during `streamToWatermark` this works. It will also tie in nicely with my WIP followup change to replace the "watchdog timeout" implementation of polling with a "caught up to watermark" one, but I'll send that as a separate pull request.
___
*[source-postgres/capture.go, line 322 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5Gj-F0DT3thGRgUp6:-MoABoNtBCr9ftRI1RSt:b-q1e00y) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L322)):*

_Should_ these be ignored (as opposed to emitting state updates)?

In my current understanding, the downside is that the connector may produce a lot of documents in between emitted checkpoints, which could impact its latency.

Is it a problem to emit checkpoints which continue to update the read-through LSN? It would cause the present backfill chunk to be restarted from a further-forward LSN, but that seems okay (even desirable) ?

Note it's possible (but not required) to emit the LSN and the scan states as separate partial checkpoints, which the runtime will merge/reduce. The actual checkpoint bytes can be small.
</blockquote></details>

I hadn't really thought about that. My intuition says that you're right: emitting checkpoints while catching up to a watermark just means we might restart from a later point, but then we'll begin by writing a new "dummy" watermark, catching up to that, and repeating the backfill query and from that point we're back into essentially the same state as before with all of the same correctness guarantees.

And yeah, given that we *can* do it we definitely should, especially because of how that ties into the previous question about modelling the tail state as "streaming until a nonexistent watermark".
___
*[source-postgres/capture.go, line 403 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5eIjf-50CRv6isE89:-MoAaWa98MnVs94lbm14:bq7qqjq) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L403)):*

It could model scans as an outer join of the watermarks table with the core scan query, and have an extra computed column with a "watermark is the correct value" bit. This seems worth doing IMO, but happy to punt for this PR.
</blockquote></details>

That sounds like a good way of implementing the check. I'll see about adding that in a followup PR just to avoid adding even more complexity to this one.
___
*[source-postgres/replication.go, line 85 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5jXpFALQ_MPoWRDPU:-MoAFHtU1HdCJPrWJ5ix:b5iggyb) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/replication.go#L85)):*

From a performance perspective, I suspect you want a reasonable buffer here so that the message decode vs handling loops can run in parallel.
</blockquote></details>

Good catch. I left it unbuffered during development so I could make sure the buffering wasn't masking any concurrency bugs, but there should be a good-sized buffer during actual usage. I've added a global variable (for testing) specifying the buffer size.
___
*[source-postgres/replication.go, line 120 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5gQGU-2e6NW1Uz4Fu:-MoAGfdm-KdW9tDqtij3:btpzw0n) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/replication.go#L120)):*

I'm not following this: why does capture.go call CurrentLSN() to asynchronously fetch the LSN now, vs using an LSN attached to the changeEvent as before ? This seems racy, where previously it wasn't.
</blockquote></details>

I can't remember exactly what my reasoning was for this change. Reverted.
___
*[source-postgres/replication.go, line 158 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5jnNoCkp_sRn2W9-6:-MoAHpGW4T3g-KPSXqj-:b9polwo) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/replication.go#L158)):*

Is this n=1 buffer behavior so that CurrentLSN() doesn't get ahead of what the reader's seeing ?
</blockquote></details>

The `eventBuf` field exists because we need to be able to stop blocking on the send to `s.events`, go send a standby status update, and then go back to trying to send that event to the channel. With appropriate buffer sizing this should only happen if the backfill queries take an unexpectedly long time, but it's still an edge-case that needs to be handled.

There's a few other ways we could get the same effect, such as by checking `len(s.events)` before doing `s.receiveMessage()` or by maintaining some sort of "channel not empty" semaphore, this was just the mechanism I was most confident of getting right on the first try.
___
*[source-postgres/replication.go, line 230 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5iartD13vGojlDMLj:-MoAJH4b9aWTu05jnmWp:by91n35) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/replication.go#L230)):*

Seems like you could define semantics of a re-added `changeEvent.LSN` to be that it's ONLY set if Type == "Commit".
</blockquote></details>

Yeah, the capture code only ever uses the LSN when `Type == "Commit"` so there's really no need to include it the rest of the time. Done.
___
*[source-postgres/snapshot.go, line 22 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5ocN3BxC172dT1mIQ:-MoAL15548QPZP4RbfJr:bc2vrju) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/snapshot.go#L22)):*

Names are confusing here. I _believe_ `scanKey` is fields which constitute the primary key, while `prevKey` is the tuple-encoded key that was left off at, correct?

`keyColumns` vs `startKey` ?

Also would `scanTableChunk` be a more accurate name?
</blockquote></details>

I like `keyColumns` as a replacement for `scanKey`. The naming was a bit clunky here because it's trying to be consistent with the use of `scanKey` in the rest of the connector to refer to that value, but I don't see anywhere that `keyColumns` isn't at least equally clear as a name, so I've changed it globally.

I'm concerned that `startKey` could also be confusing, how about `resumeKey`? That to me carries the right implications that it's acting as a resume cursor whose value is the key of some row.
___
*[source-postgres/snapshot.go, line 23 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5mBL0DvqcqfHh_U5S:-MoAKJ0_9xYugPOXaml_:b-d07mym) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/snapshot.go#L23)):*

`snapshotDatabase` is still fetching the approximate LSN. That can be removed ?

Do we even _need_ an explicit read transaction anymore ?

I'm sure this is on your future refactoring radar, but `tableSnapshot` feels a bit clunky now, like the whole thing could be lifted into the body of this function.
</blockquote></details>

This whole file is quite overengineered after the watermark changes, but I'm expecting it to get rewritten in the near future as part of the "make things abstract so we can reuse most of this code for other SQL databases" change, so I didn't spend time refactoring it yet. Happy to do so as part of this PR if you'd prefer.
___
*[source-postgres/snapshot.go, line 43 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5n7d43UChCbsd-0GM:-MoAZBoy0pPsbritExnW:bs3z49j) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/snapshot.go#L43)):*

nit: can this use a $1 placeholder ? (thinking about SQL injection).
</blockquote></details>

Unfortunately no, `CREATE TABLE [IF NOT EXISTS] $1` just results in a syntax error from PostgreSQL.

This table name is coming from the `watermarks_table` property in `config.json`, which seems like it should be reasonably trustworthy. But if we're sufficiently worried about SQL injection in the contents of `config.json` and/or `catalog.json` it wouldn't be too hard to write a function to sanitize table/column names when building queries.

jgraettinger (approved):
:lgtm: 

> Reviewed 31 of 32 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/68#-:-MoAqJ9VAXNPxHv18KjB:bg7igkv)* status: 32 of 33 files reviewed, all discussions resolved (waiting on @jgraettinger)

___
*[source-postgres/capture.go, line 105 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5DxiN60I2D4mAvCnR:-MoAnpFs3P9extjOBgnq:b1ry5b6) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L105)):*
<details><summary><i>Previously, willdonnelly (Will Donnelly) wrote…</i></summary><blockquote>

Yeah, I tend to favor using imports by their default names just because it makes the automatic imports editing work more reliably. Happy to do a quick search-and-replace if you'd prefer `log` though.
</blockquote></details>

Nope, meant this as optional.
___
*[source-postgres/capture.go, line 403 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5eIjf-50CRv6isE89:-MoAoJnA7aW_JmFuS7QO:bn1wz0t) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/capture.go#L403)):*
<details><summary><i>Previously, willdonnelly (Will Donnelly) wrote…</i></summary><blockquote>

That sounds like a good way of implementing the check. I'll see about adding that in a followup PR just to avoid adding even more complexity to this one.
</blockquote></details>

:thumbsup:
___
*[source-postgres/replication.go, line 158 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5jnNoCkp_sRn2W9-6:-MoApDUx1mzbQUXFLF3n:b-svvhpp) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/replication.go#L158)):*
<details><summary><i>Previously, willdonnelly (Will Donnelly) wrote…</i></summary><blockquote>

The `eventBuf` field exists because we need to be able to stop blocking on the send to `s.events`, go send a standby status update, and then go back to trying to send that event to the channel. With appropriate buffer sizing this should only happen if the backfill queries take an unexpectedly long time, but it's still an edge-case that needs to be handled.

There's a few other ways we could get the same effect, such as by checking `len(s.events)` before doing `s.receiveMessage()` or by maintaining some sort of "channel not empty" semaphore, this was just the mechanism I was most confident of getting right on the first try.
</blockquote></details>

Got it, makes sense.
___
*[source-postgres/snapshot.go, line 22 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5ocN3BxC172dT1mIQ:-MoApg3A7Dp7ST_YmvVy:bntzo2f) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/snapshot.go#L22)):*
<details><summary><i>Previously, willdonnelly (Will Donnelly) wrote…</i></summary><blockquote>

I like `keyColumns` as a replacement for `scanKey`. The naming was a bit clunky here because it's trying to be consistent with the use of `scanKey` in the rest of the connector to refer to that value, but I don't see anywhere that `keyColumns` isn't at least equally clear as a name, so I've changed it globally.

I'm concerned that `startKey` could also be confusing, how about `resumeKey`? That to me carries the right implications that it's acting as a resume cursor whose value is the key of some row.
</blockquote></details>

`resumeKey` is excellent
___
*[source-postgres/snapshot.go, line 23 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5mBL0DvqcqfHh_U5S:-MoApjpCDxMjBPxdBo59:b-d7lzgt) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/snapshot.go#L23)):*
<details><summary><i>Previously, willdonnelly (Will Donnelly) wrote…</i></summary><blockquote>

This whole file is quite overengineered after the watermark changes, but I'm expecting it to get rewritten in the near future as part of the "make things abstract so we can reuse most of this code for other SQL databases" change, so I didn't spend time refactoring it yet. Happy to do so as part of this PR if you'd prefer.
</blockquote></details>

Nope, I was just seeking alignment on where it's going. Better to do it as part of a refactoring-only PR that doesn't alter behavior.
___
*[source-postgres/snapshot.go, line 43 at r2](https://reviewable.io/reviews/estuary/connectors/68#-Mo5n7d43UChCbsd-0GM:-MoApwim3u-X4Ds7rCZS:bmc87yf) ([raw file](https://github.com/estuary/connectors/blob/34ecdb5033960e0191ed17bcfb15b0376ab46d51/source-postgres/snapshot.go#L43)):*
<details><summary><i>Previously, willdonnelly (Will Donnelly) wrote…</i></summary><blockquote>

Unfortunately no, `CREATE TABLE [IF NOT EXISTS] $1` just results in a syntax error from PostgreSQL.

This table name is coming from the `watermarks_table` property in `config.json`, which seems like it should be reasonably trustworthy. But if we're sufficiently worried about SQL injection in the contents of `config.json` and/or `catalog.json` it wouldn't be too hard to write a function to sanitize table/column names when building queries.
</blockquote></details>

I'm not really concerned about it, I was just wondering if it was trivial to address here.

PR #69, opened 2021-11-11
connectors: push to ghcr.io with 'dev' tag

willdonnelly:
This is a hackish copy-paste so now there are two build-push-action instances, but it's still the most elegant way I could figure out to express "always upload with the SHA tag (including for PRs), but only use 'dev' tag when `main` has a new build".

If the "push SHA-tagged version for PRs" behavior isn't intentional the two actions could be combined instead.

GitHub Actions are confusing and terrible though, so I have no idea if this will actually work.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -163,3 +163,12 @@ jobs:
           file: ${{ matrix.connector }}/Dockerfile
           push: true
           tags: ghcr.io/estuary/${{ matrix.connector }}:${{ steps.prep.outputs.tag }}
+
+      - name: Push ${{ matrix.connector }} image with 'dev' tag
+        if: ${{ github.event_name == 'push' }}
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ${{ matrix.connector }}/Dockerfile
+          push: true # See 'if' above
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -163,3 +163,12 @@ jobs:
           file: ${{ matrix.connector }}/Dockerfile
           push: true
           tags: ghcr.io/estuary/${{ matrix.connector }}:${{ steps.prep.outputs.tag }}
```
jgraettinger, line 165: does tags take an array ? can we add it here?

willdonnelly: It does take an array, but the semantics of the two push actions are different. The first one will push a hash-tagged image whenever the CI runs, including on not-yet-merged PRs, while the second one needs to only push the `dev`-tagged image when it's been committed.

I considered trying to make it an array which may or may not include the `dev` tag, but figuring out how to do that conditionally in Github Actions YAML was beyond me.

jgraettinger: Ah my bad, I failed to note the `if` conditional.

jgraettinger (approved):
LGTM

PR #71, opened 2021-11-12
source-postgres: add `required` to discovered schemas

willdonnelly:
The JSON schema generated for a table needs to label the primary key columns as required in order for the Flow discovery workflow to work properly.

This is the difference between

```
{
  "properties": {
    "a": { "type": "integer" },
    "b": { "type": "integer" },
    "data": { "type": "string" },
    "id": { "type": "integer" },
    "uuid": { "type": "string" }
  },
  "type": "object"
},
```

and

```
{
  "properties": {
    "a": { "type": "integer" },
    "b": { "type": "integer" },
    "data": { "type": "string" },
    "id": { "type": "integer" },
    "uuid": { "type": "string" }
  },
  "required": [ "id" ],
  "type": "object"
},
```

This fixes github.com/estuary/flow/issues/290


saterus (approved):
Simple! :100: 

Not sure if you want to make this part of this same PR or not, but it looks like it's very close in the code: In our testing, we also realized that it'd be helpful to consider nullability of a column and output that as part of the inferred schema for a field.

willdonnelly:
> Not sure if you want to make this part of this same PR or not, but it looks like it's very close in the code: In our testing, we also realized that it'd be helpful to consider nullability of a column and output that as part of the inferred schema for a field.

I hadn't thought about nullability in the generated schemas. That's absolutely going to be necessary or we'll start seeing validation failures the first time this gets used on a real-world dataset. I'll add a second commit to this PR which adds that.

saterus:
> we'll start seeing validation failures the first time this gets used on a real-world dataset.

Spot on. This is exactly how it manifested in the demo Dave is working on. :dart:

willdonnelly:
Added a second commit which, for instance, changes the SQL column type of `integer` from `{"type":"integer"}` to `{"anyOf":[{"type":"integer"},{"type":"null"}]}` in the discovered schema.

Also added a couple more test cases to verify that `null` can be round-tripped through such columns, and that `integer not null` produces the same `{"type":"integer"}` as before (again, that's an example and should apply to all types).

Issue #57, opened 2021-10-21
Support elastic materialization connector


jixij:
Enable materialization to [elastic](https://www.elastic.co/) , and look into ways to enable spatial-temporal visualizations via [kibana maps](https://www.elastic.co/guide/en/kibana/current/maps.html) or [elastic map](https://www.elastic.co/maps).

PR #76, opened 2021-11-22
elasticsearch materialization connector

jixij:
related issue: #57 

The PR creates the elastic search materialize connector.
The connector focuses on materializing raw flow documents into elastic search service.
The document schema of  elastic search is constructed from flow schema using the existing [schema builder](https://github.com/estuary/connectors/tree/main/materialize-elasticsearch/schemabuilder). 

Some TODOs of the work including 1. manual testing of the driver on different versions and managed services, and see the performances when data load is very heavy. 2. Add integration tests to it, which depends on the testing framework to be switched to work with the new Flow.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/76)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -69,6 +69,7 @@ jobs:
           - materialize-elasticsearch
           - materialize-postgres
           - materialize-snowflake
+          - materialize-s3-parquet
           - materialize-webhook
 
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 jobs:
-  build_parser:
+  build_rust_releases:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
@@ -31,21 +31,29 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             parser/target/
+            materialize-elasticsearch/schemabuilder/target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
-      - name: Test Parser
-        run: cargo test --manifest-path=parser/Cargo.toml --locked
+      - name: Test Rust Apps
+        run: |
+          cargo test --manifest-path=parser/Cargo.toml --locked
+          cargo test --manifest-path=materialize-elasticsearch/schemabuilder/Cargo.toml --locked
 
-      - name: Build Parser
-        run: cargo build --manifest-path=parser/Cargo.toml --locked --release --target=x86_64-unknown-linux-musl
+      - name: Build Rust Apps
+        run: |
+          cargo build --manifest-path=parser/Cargo.toml --locked --release --target=x86_64-unknown-linux-musl
+          cargo build --manifest-path=materialize-elasticsearch/schemabuilder/Cargo.toml --locked --release --target=x86_64-unknown-linux-musl
 
-      - uses: actions/upload-artifact@v1
+      - name: Upload Rust Apps
+        uses: actions/upload-artifact@v2
         with:
-          name: parser_binary
-          path: parser/target/x86_64-unknown-linux-musl/release/parser
+          name: rust_binary
+          path: |
+            parser/target/x86_64-unknown-linux-musl/release/parser
+            materialize-elasticsearch/schemabuilder/target/x86_64-unknown-linux-musl/release/schema-builder
 
   build_connectors:
-    needs: build_parser
+    needs: build_rust_releases
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
@@ -57,10 +65,11 @@ jobs:
           - source-kinesis
           - source-postgres
           - source-s3
+          - materialize-bigquery
+          - materialize-elasticsearch
           - materialize-postgres
           - materialize-snowflake
           - materialize-webhook
-          - materialize-bigquery
 
     steps:
       - uses: actions/checkout@v2
@@ -73,14 +82,16 @@ jobs:
           TAG=$(echo $GITHUB_SHA | head -c7)
           echo ::set-output name=tag::${TAG}
 
-      - uses: actions/download-artifact@v1
+      - uses: actions/download-artifact@v2
         with:
-          name: parser_binary
-          path: parser/target/x86_64-unknown-linux-musl/release/
+          name: rust_binary
 
       # Downloaded artifacts lose their prior permission settings
-      - name: Adjust parser permissions
-        run: chmod +x parser/target/x86_64-unknown-linux-musl/release/parser
+      - name: Adjust rust binary permissions
+        run: |
+          chmod +x \
+          parser/target/x86_64-unknown-linux-musl/release/parser \
+          materialize-elasticsearch/schemabuilder/target/x86_64-unknown-linux-musl/release/schema-builder
 
       - name: Install kafkactl
         if: matrix.connector == 'source-kafka'
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/benbjohnson/clock v1.1.0
 	github.com/bradleyjkemp/cupaloy v2.3.0+incompatible
 	github.com/brianvoe/gofakeit/v6 v6.9.0
+	github.com/elastic/go-elasticsearch/v7 v7.15.2-0.20211104170009-caf202868441
 	github.com/estuary/flow v0.1.1-0.20211014150201-5fb28e9026f9
 	github.com/estuary/protocols v0.0.0-20211129055338-9259b2dca80a
 	github.com/gogo/protobuf v1.3.2
@@ -27,6 +28,7 @@ require (
 	github.com/jessevdk/go-flags v1.5.0
 	github.com/klauspost/compress v1.13.6 // indirect
 	github.com/mattn/go-sqlite3 v2.0.3+incompatible
+	github.com/meirf/gopart v0.0.0-20180520194036-37e9492a85a8
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8 // indirect
 	github.com/prometheus/common v0.31.1 // indirect
 	github.com/sirupsen/logrus v1.8.1
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,6 @@ require (
 	github.com/jessevdk/go-flags v1.5.0
 	github.com/klauspost/compress v1.13.6 // indirect
 	github.com/mattn/go-sqlite3 v2.0.3+incompatible
-	github.com/meirf/gopart v0.0.0-20180520194036-37e9492a85a8
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8 // indirect
 	github.com/prometheus/common v0.31.1 // indirect
 	github.com/sirupsen/logrus v1.8.1
--- a/go.sum
+++ b/go.sum
@@ -216,8 +216,6 @@ github.com/docker/spdystream v0.0.0-20160310174837-449fdfce4d96/go.mod h1:Qh8CwZ
 github.com/dustin/go-humanize v0.0.0-20171111073723-bb3d318650d4/go.mod h1:HtrtbFcZ19U5GC7JDqmcUSB87Iq5E25KnS6fMYU6eOk=
 github.com/dustin/go-humanize v1.0.0/go.mod h1:HtrtbFcZ19U5GC7JDqmcUSB87Iq5E25KnS6fMYU6eOk=
 github.com/dustinkirkland/golang-petname v0.0.0-20191129215211-8e5a1ed0cff0/go.mod h1:V+Qd57rJe8gd4eiGzZyg4h54VLHmYVVw54iMnlAMrF8=
-github.com/elastic/go-elasticsearch/v7 v7.15.1 h1:Wd8RLHb5D8xPBU8vGlnLXyflkso9G+rCmsXjqH8LLQQ=
-github.com/elastic/go-elasticsearch/v7 v7.15.1/go.mod h1:OJ4wdbtDNk5g503kvlHLyErCgQwwzmDtaFC4XyOxXA4=
 github.com/elastic/go-elasticsearch/v7 v7.15.2-0.20211104170009-caf202868441 h1:16ZnvyxVecguRnzozLkpnGhgT6/Y/3aiqRuCcp9f0KE=
 github.com/elastic/go-elasticsearch/v7 v7.15.2-0.20211104170009-caf202868441/go.mod h1:OJ4wdbtDNk5g503kvlHLyErCgQwwzmDtaFC4XyOxXA4=
 github.com/elazarl/goproxy v0.0.0-20170405201442-c4fc26588b6e/go.mod h1:/Zj4wYkgs4iZTTu3o/KG3Itv/qCCa8VVMlb3i9OVuzc=
@@ -555,8 +553,6 @@ github.com/mattn/go-sqlite3 v2.0.3+incompatible h1:gXHsfypPkaMZrKbD5209QV9jbUTJK
 github.com/mattn/go-sqlite3 v2.0.3+incompatible/go.mod h1:FPy6KqzDD04eiIsT53CuJW3U88zkxoIYsOqkbpncsNc=
 github.com/matttproud/golang_protobuf_extensions v1.0.1 h1:4hp9jkHxhMHkqkrB3Ix0jegS5sx/RkqARlsWZ6pIwiU=
 github.com/matttproud/golang_protobuf_extensions v1.0.1/go.mod h1:D8He9yQNgCq6Z5Ld7szi9bcBfOoFv/3dc6xSMkL2PC0=
-github.com/meirf/gopart v0.0.0-20180520194036-37e9492a85a8 h1:7TJiWD1knYDpOAPyFBoKqoyvlsa+UwDw0kv0jVN5Mrk=
-github.com/meirf/gopart v0.0.0-20180520194036-37e9492a85a8/go.mod h1:Uz8uoD6o+eQN19hr6Yro/qKvW+KP6olFq+PK/Nn7gCE=
 github.com/minio/highwayhash v1.0.2 h1:Aak5U0nElisjDCfPSG79Tgzkn2gl66NxOMspRrKnA/g=
 github.com/minio/highwayhash v1.0.2/go.mod h1:BQskDq+xkJ12lmlUUi7U0M5Swg3EWR+dLTk+kldvVxY=
 github.com/modern-go/concurrent v0.0.0-20180228061459-e0a39a4cb421/go.mod h1:6dJC0mAP4ikYIbvyc7fijjWJddQyLn8Ig3JB5CqoB9Q=
--- a/go.sum
+++ b/go.sum
@@ -216,6 +216,10 @@ github.com/docker/spdystream v0.0.0-20160310174837-449fdfce4d96/go.mod h1:Qh8CwZ
 github.com/dustin/go-humanize v0.0.0-20171111073723-bb3d318650d4/go.mod h1:HtrtbFcZ19U5GC7JDqmcUSB87Iq5E25KnS6fMYU6eOk=
 github.com/dustin/go-humanize v1.0.0/go.mod h1:HtrtbFcZ19U5GC7JDqmcUSB87Iq5E25KnS6fMYU6eOk=
 github.com/dustinkirkland/golang-petname v0.0.0-20191129215211-8e5a1ed0cff0/go.mod h1:V+Qd57rJe8gd4eiGzZyg4h54VLHmYVVw54iMnlAMrF8=
+github.com/elastic/go-elasticsearch/v7 v7.15.1 h1:Wd8RLHb5D8xPBU8vGlnLXyflkso9G+rCmsXjqH8LLQQ=
+github.com/elastic/go-elasticsearch/v7 v7.15.1/go.mod h1:OJ4wdbtDNk5g503kvlHLyErCgQwwzmDtaFC4XyOxXA4=
+github.com/elastic/go-elasticsearch/v7 v7.15.2-0.20211104170009-caf202868441 h1:16ZnvyxVecguRnzozLkpnGhgT6/Y/3aiqRuCcp9f0KE=
+github.com/elastic/go-elasticsearch/v7 v7.15.2-0.20211104170009-caf202868441/go.mod h1:OJ4wdbtDNk5g503kvlHLyErCgQwwzmDtaFC4XyOxXA4=
 github.com/elazarl/goproxy v0.0.0-20170405201442-c4fc26588b6e/go.mod h1:/Zj4wYkgs4iZTTu3o/KG3Itv/qCCa8VVMlb3i9OVuzc=
 github.com/envoyproxy/go-control-plane v0.9.0/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
@@ -551,6 +555,8 @@ github.com/mattn/go-sqlite3 v2.0.3+incompatible h1:gXHsfypPkaMZrKbD5209QV9jbUTJK
 github.com/mattn/go-sqlite3 v2.0.3+incompatible/go.mod h1:FPy6KqzDD04eiIsT53CuJW3U88zkxoIYsOqkbpncsNc=
 github.com/matttproud/golang_protobuf_extensions v1.0.1 h1:4hp9jkHxhMHkqkrB3Ix0jegS5sx/RkqARlsWZ6pIwiU=
 github.com/matttproud/golang_protobuf_extensions v1.0.1/go.mod h1:D8He9yQNgCq6Z5Ld7szi9bcBfOoFv/3dc6xSMkL2PC0=
+github.com/meirf/gopart v0.0.0-20180520194036-37e9492a85a8 h1:7TJiWD1knYDpOAPyFBoKqoyvlsa+UwDw0kv0jVN5Mrk=
+github.com/meirf/gopart v0.0.0-20180520194036-37e9492a85a8/go.mod h1:Uz8uoD6o+eQN19hr6Yro/qKvW+KP6olFq+PK/Nn7gCE=
 github.com/minio/highwayhash v1.0.2 h1:Aak5U0nElisjDCfPSG79Tgzkn2gl66NxOMspRrKnA/g=
 github.com/minio/highwayhash v1.0.2/go.mod h1:BQskDq+xkJ12lmlUUi7U0M5Swg3EWR+dLTk+kldvVxY=
 github.com/modern-go/concurrent v0.0.0-20180228061459-e0a39a4cb421/go.mod h1:6dJC0mAP4ikYIbvyc7fijjWJddQyLn8Ig3JB5CqoB9Q=
--- a/materialize-elasticsearch/.snapshots/TestDriverSpec
+++ b/materialize-elasticsearch/.snapshots/TestDriverSpec
@@ -5,18 +5,16 @@
     "definitions": {
       "config": {
         "required": [
-          "endpoint",
-          "Username",
-          "Password"
+          "endpoint"
         ],
         "properties": {
           "endpoint": {
             "type": "string"
           },
-          "Username": {
+          "username": {
             "type": "string"
           },
-          "Password": {
+          "password": {
             "type": "string"
           }
         },
--- a/materialize-elasticsearch/.snapshots/TestDriverSpec
+++ b/materialize-elasticsearch/.snapshots/TestDriverSpec
@@ -129,7 +129,8 @@
             "type": "array"
           },
           "number_of_shards": {
-            "type": "integer"
+            "type": "integer",
+            "default": 1
           },
           "number_of_replicas": {
             "type": "integer"
--- a/materialize-elasticsearch/.snapshots/TestDriverSpec
+++ b/materialize-elasticsearch/.snapshots/TestDriverSpec
@@ -127,6 +127,12 @@
               "$ref": "#/definitions/FieldOverride"
             },
             "type": "array"
+          },
+          "number_of_shards": {
+            "type": "integer"
+          },
+          "number_of_replicas": {
+            "type": "integer"
           }
         },
         "additionalProperties": false,
--- a/materialize-elasticsearch/.snapshots/TestDriverSpec
+++ b/materialize-elasticsearch/.snapshots/TestDriverSpec
@@ -1,0 +1,140 @@
+{
+  "endpoint_spec_schema_json": {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$ref": "#/definitions/config",
+    "definitions": {
+      "config": {
+        "required": [
+          "endpoint",
+          "Username",
+          "Password"
+        ],
+        "properties": {
+          "endpoint": {
+            "type": "string"
+          },
+          "Username": {
+            "type": "string"
+          },
+          "Password": {
+            "type": "string"
+          }
+        },
+        "additionalProperties": false,
+        "type": "object"
+      }
+    }
+  },
+  "resource_spec_schema_json": {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$ref": "#/definitions/resource",
+    "definitions": {
+      "DateSpec": {
+        "required": [
+          "format"
+        ],
+        "properties": {
+          "format": {
+            "type": "string"
+          }
+        },
+        "additionalProperties": false,
+        "type": "object"
+      },
+      "ElasticFieldType": {
+        "required": [
+          "field_type"
+        ],
+        "properties": {
+          "field_type": {
+            "type": "string"
+          },
+          "date_spec": {
+            "$schema": "http://json-schema.org/draft-04/schema#",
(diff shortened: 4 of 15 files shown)


jgraettinger (changes requested):
Thanks! I've left a lot of feedback below, a majority of which are nits.

High level though, this is looking great. I also want to applaud the consistent style and structure of the code -- it feels familiar, idiomatic, and straightforward to read :thumbsup: 

> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/76#-:-MphlJcTEqcEotELSJks:b-um3oen)* status: all files reviewed, 19 unresolved discussions (waiting on @jixij and @saterus)

___
*[.github/workflows/ci.yaml, line 68 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphQjBR-wXM5kXeld9M:-MphQjBR-wXM5kXeld9N:bj5iyk7) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/.github/workflows/ci.yaml#L68)):*
> ```YAML
>           - source-postgres
>           - source-s3
>           - materialize-bigquery
> ```

:+1: thanks for sorting
___
*[materialize-elasticsearch/driver.go, line 44 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphRpW3AglWMCh6BHNs:-MphRpW3AglWMCh6BHNt:btsq69g) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L44)):*
> ```Go
> 	// For single-node clusters, make sure the number_of_replicas is 0, b/c the
> 	// elastic search needs to allocate replicas on different nodes.
> 	// Access the following two fields via getters.
> ```

Heads up: comments on these fields will _eventually_ be user-facing in the UI, and ideally would be written from a user-centric perspective reflective of that.

Technical details: The JSON Schema generator will convert these comments in `description` fields of the generated schema (at least, I think it does), and we already return resource specification schemas in the Spec RPC though we don't do anything with them yet. We have all the plumbing in place, though, to eventually surface these as UI descriptions.
___
*[materialize-elasticsearch/driver.go, line 46 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphSkx7CzU-uL1NAsID:-MphSkx7CzU-uL1NAsIE:b-bnujyn) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L46)):*
> ```Go
> 	// Access the following two fields via getters.
> 	NumOfShards   *int `json:"number_of_shards,omitempty"`
> 	NumOfReplicas *int `json:"number_of_replicas,omitempty"`
> ```

nit: do these need to be pointers, or does the zero-value suit for a "not set" default?
___
*[materialize-elasticsearch/driver.go, line 121 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphTWl_6ucWdQaxC1F9:-MphTWl_6ucWdQaxC1FA:b-qzy2j7) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L121)):*
> ```Go
> 				constraint.Reason = "The root document is needed."
> 			default:
> 				constraint.Type = pm.Constraint_FIELD_OPTIONAL
> ```

My understanding is that:

a) The root document is required (appears as StoreIterator.RawJSON) -- this can be Constraint_LOCATION_REQUIRED (rather than FIELD_REQUIRED) as we don't need _all_ fields having the root projection, just one of them.
b) Keyed locations are also required (they appear as StoreIterator.Key). This also can be LOCATION_REQUIRED.
c) Other locations (appearing as StoreIterator.Values) are ignored by the connector, and should be disallowed.

Does that seem right ?
___
*[materialize-elasticsearch/driver.go, line 165 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphU-x7-ZtV7VK2Pijo:-MphU-x7-ZtV7VK2Pijp:b-pv1oh6) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L165)):*
> ```Go
> 		}
> 
> 		if err = elasticSearch.CreateIndex(res.Index, res.GetNumOfShards(), res.GetNumOfReplicas(), elasticSearchSchema); err != nil {
> ```

What happens if:

 * The index already exists ? (A: we require that it's present schema match this one).
 * The index exists with a different number of shards? (A: we ignore the updated shards, I believe?)
 * The index exists with a different number of replicas ? (A: we ignore here too ?)
 * The index exists with a different schema ? (A: we fail with an error during Apply)

(I came back and updated these after finishing the review).

Would you please note these semantics in a comment ?
___
*[materialize-elasticsearch/driver.go, line 175 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphUMaq6Wj4btFF3Lzt:-MphUMaq6Wj4btFF3Lzu:b-2cxxk7) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L175)):*
> ```Go
> 
> // ApplyDelete is a no-op.
> func (driver) ApplyDelete(ctx context.Context, req *pm.ApplyRequest) (*pm.ApplyResponse, error) {
> ```

Is this truly a no-op, or is it a TODO ? Should it tell Elastic to drop the index ?
___
*[materialize-elasticsearch/driver.go, line 225 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphV81o3ZJF9K_j9WZK:-MphV81o3ZJF9K_j9WZL:b3f9x36) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L225)):*
> ```Go
> 
> 	var log = log.WithField(
> 		"materialization",
> ```

nit: heads up the Flow runtime will attach task & shard info -- including key ranges -- to log messages it gathers from the connector invocation (so you don't need to do it here).
___
*[materialize-elasticsearch/driver.go, line 246 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphVXOkEQERyTPdRnns:-MphVXOkEQERyTPdRnnt:bjmbqko) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L246)):*
> ```Go
> func (t *transactor) Load(it *pm.LoadIterator, _ <-chan struct{}, _ <-chan struct{}, loaded func(int, json.RawMessage) error) error {
> 	var loadingIdsByBinding = map[int][]string{}
> ```

This implementation is :100: correct, but I'll suggest a comment here noting that you're ignoring the `commitCh` argument because you're not actually loading anything until the LoadIterator has completed. Were that not the case you would need to await commitCh because `Commit` is running concurrently.

In the future it may be a meaningful latency optimization to issue loads as soon as a batch fills up. That would require first awaiting `commitCh` however.
___
*[materialize-elasticsearch/driver.go, line 249 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphXqqP0A-zUByWRCG9:-MphXqqP0A-zUByWRCGA:bjeldel) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L249)):*
> ```Go
> 	for it.Next() {
> 		var b = t.bindings[it.Binding]
> 		if b.deltaUpdates {
> ```

nit: you can hoist this panic to the top-level of the function. Not only will it.Next() never be true, but Load() shouldn't be called _at all_ in this case.
___
*[materialize-elasticsearch/driver.go, line 257 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphWwmd3soX3PZ9bVus:-MphWwmd3soX3PZ9bVut:bddxi2i) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L257)):*
> ```Go
> 	for binding, ids := range loadingIdsByBinding {
> 		var b = t.bindings[binding]
> 		for idxRange := range gopart.Partition(len(ids), loadByIdBatchSize) {
> ```

nit: seems like you could replace this dependency with:

```
for start := 0; start < len(ids); start++ {
  var stop = start + loadByIdBatchSize
  if stop > len(ids) {
    stop = len(ids)
  }

  ...
}
```

Go proverb: [A little copying is better than a little dependency](https://go-proverbs.github.io/)
___
*[materialize-elasticsearch/driver.go, line 296 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphZR4J2a_EC5HpV92L:-MphZR4J2a_EC5HpV92M:bu3r1i9) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L296)):*
> ```Go
> 			Body:       bytes.NewReader(it.RawJSON),
> 			OnFailure: func(_ context.Context, _ esutil.BulkIndexerItem, r esutil.BulkIndexerResponseItem, e error) {
> 				log.Error(fmt.Sprintf("failed with response: %+v, error: %v", r, e))
> ```

nit: prefer log.WithFields(logrus.Fields{"error": e, "response": r}).Error("failed to store documents") in order to get structured output of these fields. Phil's recently done a bunch of log-related work that lets us detect and propagate structure like this all the way through!
___
*[materialize-elasticsearch/driver.go, line 304 at r1](https://reviewable.io/reviews/estuary/connectors/76#-Mph_WzGCu7u_v_XzKnD:-Mph_WzGCu7u_v_XzKnE:b4fht1j) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L304)):*
> ```Go
> 	}
> 
> 	return lastErr
> ```

`lastErr` cannot be anything other than`nil` here, right ? because the BulkIndexerItems haven't actually be run yet, just created.

Will `elasticSearch.Commit()` in `transactor.Commit` also error if any of these individual items error ? (in which case, we could drop lastErr?) ?

Also, could these callbacks be race-y ?
___
*[materialize-elasticsearch/driver.go, line 308 at r1](https://reviewable.io/reviews/estuary/connectors/76#-Mphaax9E2X03E06S0io:-Mphaax9E2X03E06S0ip:b96uyx5) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L308)):*
> ```Go
> 
> func (t *transactor) Commit(ctx context.Context) error {
> 	defer func() { t.bulkIndexerItems = t.bulkIndexerItems[:0] }()
> ```

super pedantic nit: I'd suggest having this be just before return, and not in a defer, because the defer behavior isn't actually needed -- if this function errors then Store will never be called again. Given that, the rationale for _not_ deferring it is that the reader could assume there's important semantics there when in fact there isn't.

Another option is to have Store do the truncation immediately before use.
___
*[materialize-elasticsearch/driver.go, line 315 at r1](https://reviewable.io/reviews/estuary/connectors/76#-Mphb6qn26PtXsvHchCh:-Mphb6qoAuo1q2JMCYd8:b8fobum) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L315)):*
> ```Go
> 
> 	for _, b := range t.bindings {
> 		// Using Flush instead of Refresh to make sure the data are persisted.
> ```

nit: mind expanding very slightly on the distinction between Refresh and Flush for those unfamiliar with Elastic ?
___
*[materialize-elasticsearch/elasticsearch.go, line 26 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphfKB6EbldBGg-IkgR:-MphfKB6EbldBGg-IkgS:bp1wxq1) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/elasticsearch.go#L26)):*
> ```Go
> }
> 
> func NewElasticSearch(endpoint string, username string, password string) (*ElasticSearch, error) {
> ```

nit; I don't get the sense that `ElasticSearch` is intended for use outside of this package, right? If that's the case, consider making its methods non-exported.
___
*[materialize-elasticsearch/elasticsearch.go, line 42 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphbP1rA6ztC9HWVRzU:-MphbP1rA6ztC9HWVRzV:b-y8lkqm) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/elasticsearch.go#L42)):*
> ```Go
> 
> // CreateIndex creates a new es index and sets its mappings to be schemaJSON.
> // If an index with the same already exists, and the schema(mappings) of the existing index is inconsistent with
> ```

Ah, great! We should verify schema compatibility as part of Validate though. The reason for this is that Validate is called at build time, and will produce a catalog build error, while ApplyUpsert happens later.
___
*[materialize-elasticsearch/elasticsearch.go, line 104 at r1](https://reviewable.io/reviews/estuary/connectors/76#-Mphdy6B9_BnQyNNFmBn:-Mphdy6B9_BnQyNNFmBo:b-sh3dbn) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/elasticsearch.go#L104)):*
> ```Go
> 	}
> 
> 	var bi, err = esutil.NewBulkIndexer(esutil.BulkIndexerConfig{
> ```

I'm confused by error propagation in these interfaces. Is it our responsibility to track an error passed to OnError, or does that bubble up to Add() or Close() below ? How does it relate to the OnFailure() callback of BulkIndexerItem ?

would you please add a comment discussing errors and their propagation ?
___
*[materialize-elasticsearch/schemabuilder/src/elastic_search_schema_builder.rs, line 23 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphfvXLCs9hFg-aNTmq:-MphfvXLCs9hFg-aNTmr:bowz5cs) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/schemabuilder/src/elastic_search_schema_builder.rs#L23)):*
> ```XML
>     es_type_overrides: &[ESTypeOverride],
> ) -> Result<ESFieldType, Error> {
>     // parse should not fail on the hard-coded url.
> ```

super-nit: use expect("parse should not fail on hard-coded url") to both document this expectation and also make it accessible in stack traces.
___
*[rust-binary/schema-builder, line 0 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphgCXuEeItLuT8tuar:-MphgCXuEeItLuT8tuas:by75shv) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/rust-binary/schema-builder#L0)):*
This should be in .gitignore ?

jixij:
___
*[materialize-elasticsearch/driver.go, line 304 at r1](https://reviewable.io/reviews/estuary/connectors/76#-Mph_WzGCu7u_v_XzKnD:-MpqhIXR6e0SdKlYNj4L:b-ydnay0) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L304)):*

`lastErr` cannot be anything other than`nil` here, right ? because the BulkIndexerItems haven't actually be run yet, just created.

Will `elasticSearch.Commit()` in `transactor.Commit` also error if any of these individual items error ? (in which case, we could drop lastErr?) ?

Also, could these callbacks be race-y ?

</blockquote></details>

great catch. Feel sorry about it. Might be better to construct the BulkIndexerItem in the Commit stage.

jixij:
___
*[materialize-elasticsearch/driver.go, line 44 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphRpW3AglWMCh6BHNs:-Mpqw4m-1c_uPpMMWCt9:bbr6iuq) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L44)):*

Heads up: comments on these fields will _eventually_ be user-facing in the UI, and ideally would be written from a user-centric perspective reflective of that.

Technical details: The JSON Schema generator will convert these comments in `description` fields of the generated schema (at least, I think it does), and we already return resource specification schemas in the Spec RPC though we don't do anything with them yet. We have all the plumbing in place, though, to eventually surface these as UI descriptions.
</blockquote></details>

Thanks for the reminder. I have improved the comments to be per-field and more user-oriented.

jixij:
___
*[materialize-elasticsearch/driver.go, line 46 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphSkx7CzU-uL1NAsID:-Mpqwah2BHKkDQsK9cMD:b7t5hfm) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L46)):*

nit: do these need to be pointers, or does the zero-value suit for a "not set" default?
</blockquote></details>

The initial intention was to default `NumOfShards` to be `1` if not specified, so pointers were introduced in.
To make it simpler and avoid using pointers, I added the validation check to make sure `NumOfShards` must be specified as a positive integer.

jixij:
___
*[materialize-elasticsearch/driver.go, line 257 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphWwmd3soX3PZ9bVus:-MpqzjqG-uwbngH8BNoc:bb6un1j) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L257)):*

nit: seems like you could replace this dependency with:

```
for start := 0; start < len(ids); start++ {
  var stop = start + loadByIdBatchSize
  if stop > len(ids) {
    stop = len(ids)
  }

  ...
}
```

Go proverb: [A little copying is better than a little dependency](https://go-proverbs.github.io/)
</blockquote></details>

Done

jixij:
___
*[materialize-elasticsearch/driver.go, line 308 at r1](https://reviewable.io/reviews/estuary/connectors/76#-Mphaax9E2X03E06S0io:-Mpr-AAs1PHSGFrrfmLW:b-896fix) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L308)):*

super pedantic nit: I'd suggest having this be just before return, and not in a defer, because the defer behavior isn't actually needed -- if this function errors then Store will never be called again. Given that, the rationale for _not_ deferring it is that the reader could assume there's important semantics there when in fact there isn't.

Another option is to have Store do the truncation immediately before use.
</blockquote></details>

Done.

jixij:
___
*[materialize-elasticsearch/driver.go, line 165 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphU-x7-ZtV7VK2Pijo:-Mq-eGJ-DVWmYIJ-CMjN:bdt2pqh) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L165)):*

What happens if:

 * The index already exists ? (A: we require that it's present schema match this one).
 * The index exists with a different number of shards? (A: we ignore the updated shards, I believe?)
 * The index exists with a different number of replicas ? (A: we ignore here too ?)
 * The index exists with a different schema ? (A: we fail with an error during Apply)

(I came back and updated these after finishing the review).

Would you please note these semantics in a comment ?
</blockquote></details>

Yes, I have added some comments to the API of `CreateIndex` to explain the current behavior. The API is shared by both `Validate` and `ApplyUpsert`.

All you statements here are perfect according to the previous implementation. 

And this triggered some more thoughts in this direction. During the revision, I have made some improvements, so that 1) the connector will update the number of replica (or dynamic settings in general) if inconsistency are detected. 2) the connector will stop and issue an error message if inconsistencies are detected in schema, number of shards (or any static settings in general)

jixij:
___
*[materialize-elasticsearch/driver.go, line 175 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphUMaq6Wj4btFF3Lzt:-Mq-gDFc-5St0KREicsW:b-nhmza4) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L175)):*

Is this truly a no-op, or is it a TODO ? Should it tell Elastic to drop the index ?
</blockquote></details>

Agree. I did not pay much attention when adding this API. 

I have made the change to instruct elastic search to drop the index. And I feel this could be a way, although not the perfect, to allow users to resolve inconsistencies detected in index specifications. So I have added the logic here.

jixij:
___
*[materialize-elasticsearch/driver.go, line 225 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphV81o3ZJF9K_j9WZK:-Mq-gmSSEVQKSYbx912I:b-89q77k) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L225)):*

nit: heads up the Flow runtime will attach task & shard info -- including key ranges -- to log messages it gathers from the connector invocation (so you don't need to do it here).
</blockquote></details>

Cool.

jixij:
___
*[materialize-elasticsearch/driver.go, line 246 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphVXOkEQERyTPdRnns:-Mq-gx5Y7iFmQVZK3q95:b8al2lk) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L246)):*

This implementation is :100: correct, but I'll suggest a comment here noting that you're ignoring the `commitCh` argument because you're not actually loading anything until the LoadIterator has completed. Were that not the case you would need to await commitCh because `Commit` is running concurrently.

In the future it may be a meaningful latency optimization to issue loads as soon as a batch fills up. That would require first awaiting `commitCh` however.
</blockquote></details>

Yes, will keep that in mind. Thanks for the reminder.

jixij:
___
*[materialize-elasticsearch/driver.go, line 249 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphXqqP0A-zUByWRCG9:-Mq-h3FM9gq5Xu9NQ4vy:b-yu5q4h) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L249)):*

nit: you can hoist this panic to the top-level of the function. Not only will it.Next() never be true, but Load() shouldn't be called _at all_ in this case.
</blockquote></details>

The delta updates is per-binding. So better to perform the check here. I think that the check might not be needed, since the framework should be able to ensure that, and I removed it.

jixij:
___
*[materialize-elasticsearch/driver.go, line 296 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphZR4J2a_EC5HpV92L:-Mq-hbAk53zDoAr5gbjL:b-y11gt) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L296)):*

nit: prefer log.WithFields(logrus.Fields{"error": e, "response": r}).Error("failed to store documents") in order to get structured output of these fields. Phil's recently done a bunch of log-related work that lets us detect and propagate structure like this all the way through!
</blockquote></details>

Sounds good. I have adjusted this and other logs to be field-based.

jixij:
___
*[materialize-elasticsearch/elasticsearch.go, line 42 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphbP1rA6ztC9HWVRzU:-Mq-hp_03VfrAIIoSPoB:b-65a6hp) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/elasticsearch.go#L42)):*

Ah, great! We should verify schema compatibility as part of Validate though. The reason for this is that Validate is called at build time, and will produce a catalog build error, while ApplyUpsert happens later.
</blockquote></details>

Yes, it is better to detect the issues earlier. I have adjusted the createIndex api with a dryRun mode, so that the validate could perform the check as well.

jixij:
___
*[materialize-elasticsearch/elasticsearch.go, line 104 at r1](https://reviewable.io/reviews/estuary/connectors/76#-Mphdy6B9_BnQyNNFmBn:-Mq-iBhm3A8SVhrWjBdj:b-bttn2e) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/elasticsearch.go#L104)):*

I'm confused by error propagation in these interfaces. Is it our responsibility to track an error passed to OnError, or does that bubble up to Add() or Close() below ? How does it relate to the OnFailure() callback of BulkIndexerItem ?

would you please add a comment discussing errors and their propagation ?
</blockquote></details>

Yes, I have added some comments on the type of errors and how they are delivered to the client.

One thing to note: the Add and Close are responsible to report only a small fraction of errors. The OnError and OnFailure could be called without interrupting the whole workflow.

https://github.com/elastic/go-elasticsearch/blob/main/esutil/bulk_indexer.go#L264-L267
https://github.com/elastic/go-elasticsearch/blob/main/esutil/bulk_indexer.go#L367-L372

jixij:
___
*[rust-binary/schema-builder, line  at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphgCXuEeItLuT8tuar:-Mq-jPPX9-CIekrspNzp:b-maksz9) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/rust-binary/schema-builder#L)):*

This should be in .gitignore ?
</blockquote></details>

This is a temp file that is not useful at all. I should have deleted it earlier. I've just deleted it.

jixij:
___
*[materialize-elasticsearch/schemabuilder/src/elastic_search_schema_builder.rs, line 23 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphfvXLCs9hFg-aNTmq:-Mq-jswj3ElaBBKTsMiT:boyexb3) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/schemabuilder/src/elastic_search_schema_builder.rs#L23)):*

super-nit: use expect("parse should not fail on hard-coded url") to both document this expectation and also make it accessible in stack traces.
</blockquote></details>

Done. Thanks!

jixij:
Thank you Johnny for the detailed reviews and great feedback. Thanks for insights to the flow infrastructure. I have addressed the comments as needed. PTAL!

jixij (commented):
Thanks again Johnny for the reviews and approval. I've made the changes as needed. Not very sure if the current state of the jsonschema descriptions matches the requirements for the UI generator. If needed, I can further tweak them in the followup PR that adds integration tests to the connector.

> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/76#-:-MqFey2PDF4ODCX_UnoJ:betysdx)* status: 13 of 16 files reviewed, 6 unresolved discussions (waiting on @jgraettinger and @saterus)

___
*[materialize-elasticsearch/driver.go, line 44 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphRpW3AglWMCh6BHNs:-MqFWvSXDvex1zETOQA9:b5wr8to) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L44)):*
> https://github.com/estuary/protocols/blob/main/materialize/materialize.proto#L420-L422
Done
___
*[materialize-elasticsearch/driver.go, line 46 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphSkx7CzU-uL1NAsID:-MqFa2zr6zwEadRP-uw9:b5eh91q) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L46)):*
> https://github.com/alecthomas/jsonschema#example)
Cool. I have added the default value. I think we have a snapshot test in-place materialize-elasticsearch/.snapshots/TestDriverSpec, and the diff of the default value is shown.
I've also tweaked other json specs (in the schema builder) to be user-facing for displaying on UI. Not sure exactly how these descriptions will be delivered. If needed, I can move these comments into `description` of the jsonschema tag.
___
*[materialize-elasticsearch/driver.go, line 304 at r1](https://reviewable.io/reviews/estuary/connectors/76#-Mph_WzGCu7u_v_XzKnD:-MqFbP2uAEFpRMDxTwKu:b2oijsb) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L304)):*

No reason to feel sorry about it -- it's why we code review! :thumbsup: 
</blockquote></details>

 👍
___
*[materialize-elasticsearch/driver.go, line 192 at r2](https://reviewable.io/reviews/estuary/connectors/76#-MqF1QTn0JCHIlvJd_nV:-MqFbFWl8pJomuV9SDv_:bqlbe4r) ([raw file](https://github.com/estuary/connectors/blob/cc2900cc9ebf020b1a284e66989ef27d9821f954/materialize-elasticsearch/driver.go#L192)):*

nit: there's a DryRun field which could be set: https://github.com/estuary/protocols/blob/main/materialize/materialize.proto#L420-L422
</blockquote></details>

cool. Added the logic to respect it. 
___
*[materialize-elasticsearch/elasticsearch.go, line 89 at r2](https://reviewable.io/reviews/estuary/connectors/76#-MqF3etzBzuYYjhm8Jdy:-MqFcaIZ5Q9TnAFytinx:b-jm4cjk) ([raw file](https://github.com/estuary/connectors/blob/cc2900cc9ebf020b1a284e66989ef27d9821f954/materialize-elasticsearch/elasticsearch.go#L89)):*

nit: thinking about UX. Should "check index mapping: " be dropped here? Or renamed "incompatibility with existing schema: %w" or something more obvious to the user ?
</blockquote></details>

Yes, I'm dropping the message here. The error message returned from checkIndexMapping is already very informative regarding the schema diffs.
___
*[materialize-elasticsearch/elasticsearch.go, line 95 at r2](https://reviewable.io/reviews/estuary/connectors/76#-MqF3ZrC4x-f9caZPrnz:-MqFc5jZ-t8E1M5ZcJkn:b-rayhq5) ([raw file](https://github.com/estuary/connectors/blob/cc2900cc9ebf020b1a284e66989ef27d9821f954/materialize-elasticsearch/elasticsearch.go#L95)):*
> ```
> check index mapping
> ```
Done
___
*[materialize-elasticsearch/elasticsearch.go, line 277 at r3](https://reviewable.io/reviews/estuary/connectors/76#-MqF5r_4EBpp8x_yc6zg:-MqFc7Z07cJxVSFpr-QY:bb6un1j) ([raw file](https://github.com/estuary/connectors/blob/378107cd0d82a4f18e9c9112a54da030c9740cdb/materialize-elasticsearch/elasticsearch.go#L277)):*

nit: here too. The call to action is to delete the index, but I think it's out of scope here to tell them how to do that (especially given future uncertainty about the deletion workflow)
</blockquote></details>

Done

jgraettinger (commented):
> Reviewed 2 of 3 files at r4, 3 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/76)* status: 15 of 16 files reviewed, 1 unresolved discussion

___
*[materialize-elasticsearch/driver.go, line 46 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphSkx7CzU-uL1NAsID:-MqFnU8j68ERYI4hRi23:b4xhxy5) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L46)):*
<details><summary><i>Previously, jixij (Jixiang Jiang) wrote…</i></summary><blockquote>

> https://github.com/alecthomas/jsonschema#example)
Cool. I have added the default value. I think we have a snapshot test in-place materialize-elasticsearch/.snapshots/TestDriverSpec, and the diff of the default value is shown.
I've also tweaked other json specs (in the schema builder) to be user-facing for displaying on UI. Not sure exactly how these descriptions will be delivered. If needed, I can move these comments into `description` of the jsonschema tag.
</blockquote></details>

ah, great! And... darn. I'd wrongly assumed it was being a little more clever with package comments. Looking through the `reflect` package API, there's simply no way to pull them out, however.

I'm definitely spoiled lately by how thoughtful Rust is about modeling comments and making them intro-spectable (the Rust schema generator we use has awesome support for turning field comments into `title` & `description`).

jixij:
___
*[materialize-elasticsearch/driver.go, line 46 at r1](https://reviewable.io/reviews/estuary/connectors/76#-MphSkx7CzU-uL1NAsID:-MqFriAa5f9nmQndXYRi:b-e1w5cn) ([raw file](https://github.com/estuary/connectors/blob/c34c3f5c713cc1597b6c8fb287c363dbb1e4500c/materialize-elasticsearch/driver.go#L46)):*

ah, great! And... darn. I'd wrongly assumed it was being a little more clever with package comments. Looking through the `reflect` package API, there's simply no way to pull them out, however.

I'm definitely spoiled lately by how thoughtful Rust is about modeling comments and making them intro-spectable (the Rust schema generator we use has awesome support for turning field comments into `title` & `description`).
</blockquote></details>

Aha, got it. Rust sounds very cool.. 
I'll make the fix and make the descriptions available in the schema spec.

PR #78, opened 2021-12-03
Adding in default setup script (source-postgres) 

travjenkins:
### Changes

- Added a new init-db script 
- Changed volumes property to a list
- Updated docker compose to mount file

### Callouts

The init script is currently expecting Postgres version below 14 so not using the `pg_read_all_data` setting


jgraettinger:
@willdonnelly bump ?

willdonnelly:
Sorry, didn't see when review was requested. This looks like a reasonable change, but two thoughts:

1. If we're creating a dedicated `flow_capture` user we should also change the default username in the actual capture connector to `flow_capture` so it'll actually use that.
2. The final `ALTER SYSTEM` statement in the init script is redundant with the `-c wal_level=logical` in the command-line. If it's possible to trigger a database restart at the end of that script it would be better to do that, but if that's not easily doable then the redundancy is probably best (so we're demonstrating two methods of getting that setting).

source-postgres/README.md
```diff
@@ -24,7 +24,7 @@ other supported config options:
 
 ```json
 {
-  "connectionURI": "postgres://flow:flow@localhost:5432/flow"
+  "connectionURI": "postgres://flow_capture:flow_capture@localhost:5432/flow"
```
travjenkins, line 27: Making this match what would be coming down as the defaults.

source-postgres/docker-compose.yaml
```diff
@@ -4,9 +4,13 @@ services:
...
     image: 'postgres:latest'
     command: ["postgres", "-c", "wal_level=logical"]
-    volumes: ["postgres_data:/var/lib/postgresql/data"]
     environment: {"POSTGRES_DB": "flow", "POSTGRES_USER": "flow", "POSTGRES_PASSWORD": "flow"}
     network_mode: "host"
+    volumes:
+      - type: bind
+        source: ./docker-entrypoint-initdb.d/init-user-db.sh
+        target: /docker-entrypoint-initdb.d/init-user-db.sh
+      - postgres_data:/var/lib/postgresql/data
```
travjenkins, line 13: Moving the previous volume down here and adding a new one to bind the new file

source-postgres/docker-entrypoint-initdb.d/init-user-db.sh
```diff
@@ -0,0 +1,14 @@
...
+#  Included as an example of how to set this without using the command line. 
+psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
+    CREATE USER flow_capture WITH PASSWORD 'flow_capture' REPLICATION;
+    GRANT SELECT ON ALL TABLES IN SCHEMA public TO flow_capture;
+    GRANT SELECT ON ALL TABLES IN SCHEMA information_schema, pg_catalog TO flow_capture;
+    CREATE TABLE IF NOT EXISTS public.flow_watermarks (slot TEXT PRIMARY KEY, watermark TEXT);
+    GRANT ALL PRIVILEGES ON TABLE public.flow_watermarks TO flow_capture;
+    CREATE PUBLICATION flow_publication FOR ALL TABLES;
+    ALTER SYSTEM SET wal_level = logical; 
+EOSQL
```
travjenkins, line 14: Taken from the documentation we went to a customer. Making this a default to make folks wanting to get up and running a bit easier.

source-postgres/main.go
```diff
@@ -59,7 +59,7 @@ const configSchema = `{
 			"type":        "string",
 			"title":       "Database Connection URI",
 			"description": "Connection parameters, as a libpq-compatible connection string",
-			"default":     "postgres://flow:flow@localhost:5432/flow"
+			"default":     "postgres://flow_capture:flow_capture@localhost:5432/flow"
```
travjenkins, line 62: Updating the default value to use the `flow_capture` user

willdonnelly (approved):
Looks good to me, I'd go ahead and ignore the CI failures -- they're definitely unrelated.

PR #83, opened 2021-12-08
Support Avro object container files in parser

### LINKED ISSUE #74: Support Avro files in parser
We'd like to support reading avro files in the parser.
These should be associated with the `.avro` extension.
Support for compressed data blocks is likely required, as the use of compression seems widespread.
[This xandr page](https://docs.xandr.com/bundle/log-level-data/page/batch-log-level-data.html) happens to have some pretty good examples that we could potentially use for testing.

psFried:
This adds support for avro files in the parser. This handles all the
officially documented logical types in addition to the primitive types.
We don't currently allow the user to supply a custom read schema in the
config, but that may be a reasonable feature to add in the future.

Some logical types, like timestamps will be converted differently
depending on the types allowed by the JSON schema. The default is to
convert timestamps, dates, and times to the ISO-8601 string
representation. But if the JSON schema disallows strings for a given
location, then the integer representation will be used.

Another potentially reasonable feature that is _not_ implemented here is
a way to set the timezone that's used timestamp values. They are
currently emitted without timezone info.

Resolves #74


parser/src/format/avro.rs
```diff
@@ -0,0 +1,409 @@
...
+    // times don't allow negative values
+    let time = Some(to_secs_and_nanos(duration, units_per_sec))
+        .filter(|&(secs, _)| secs >= 0)
+        .and_then(|(secs, nanos)| NaiveTime::from_num_seconds_from_midnight_opt(secs as u32, nanos))
+        .ok_or_else(|| AvroError::DateTimeOverflow(duration, avro_type, column.to_string()))?;
+
+    Ok(Value::String(time.to_string()))
+}
+
+fn to_secs_and_nanos(duration: i64, units_per_sec: i64) -> (i64, u32) {
```
jixij, line 268: Not exactly sure on the function requirement. Is it a correct function call of `to_secs_and_nanos(1001, 1000)`? And what is the expect output for that?

jixij (approved):
LGTM!! :+1: 
The PR teaches me more on Rust and the parser logic.

parser/src/format/avro.rs
psFried: I'm really glad you asked! I went to add a comment, and spotted a heinous bug where negative timestamps weren't handled correctly. The `chrono` crate expects the sub-second nanos value to always be positive, so if a timestamp is negative, you need to subtract an additional second, and then add back in a the proportional number of sub-second nanos. I fixed that and added a test. Also added a comment explaining what's going on here.

parser/src/format/avro.rs
```diff
@@ -0,0 +1,409 @@
...
+impl Parser for AvroParser {
+    fn parse(&self, config: &ParseConfig, content: Input) -> Result<Output, ParseError> {
+        let iter = AvroIter::from_config_and_input(config, content)?;
+        Ok(Box::new(iter))
+    }
+}
+
+struct AvroIter {
+    reader: Reader<'static, Box<dyn io::BufRead>>,
+    records: BTreeMap<String, Projection>,
```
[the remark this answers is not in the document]
psFried: Good call! Fixed.

parser/src/format/avro.rs
jixij: Looks great! :+1:

PR #85, opened 2021-12-17
Factor out generic logic from `source-postgres` and implement `source-mysql` using that

### LINKED ISSUE #77: Implement a MySQL source connector
I've been working on this for the past week, and while I'm not entirely done with the details the broad strokes are mostly there. Filing this bug so I can update the [Big Board](https://github.com/orgs/estuary/projects/2).

A large chunk of the logic previously written in `source-postgres` is actually the generic "SQL Database Capture" process, so the first step of this is/was to factor out all of that into a generic `sqlcapture` package, and have `source-postgres` implement the various database operations (discover tables, stream change events via replication, write watermarks, scan table chunks, etc) plus a thin `main` function that injects that concrete implementation into the generic machinery that coordinates everything.

A similar bit of refactoring allows the tests to be factored out into another package as well, because most of the existing `source-postgres` tests should be just as applicable to any other SQL database capture.

After all that, implementing the MySQL source "only" requires implementing all those concrete operations one-by-one. This is actually working at the moment, and all the "genericized" tests are passing. However I expect I'll be working on various details for at least the rest of this week.

willdonnelly:
Apologies for how absurdly massive this pull request is [1], the `source-postgres` -> `source-postgres + sqlcapture` refactoring was always going to produce a pretty messy diff, and I think that having the new MySQL code alongside actually helps to clarify what's going on a bit.

Fixes #77 

[1] Also in my defense, the added/removed line counts GitHub shows are mostly test snapshots, plus a bit of code that got moved around in ways the diff algorithm can't follow. There's probably around 1k lines of actual new code here, mostly from the MySQL connector implementation.


source-mysql/main_test.go
```diff
@@ -0,0 +1,200 @@
...
+	dbPass     = flag.String("db_pass", "flow", "Password for the specified database test user")
+	dbName     = flag.String("db_name", "test", "Connect to the named database for tests")
+	dbServerID = flag.Int("db_serverid", 12345, "Unique server ID for replication")
+)
+
+var (
+	TestBackend *mysqlTestBackend
+)
+
+func TestMain(m *testing.M) {
```
psFried, line 30: TIL about `TestMain`. Neat!

source-mysql/replication.go
```diff
@@ -0,0 +1,214 @@
...
+				for _, row := range data.Rows {
+					var before = decodeRow(columnNames, row)
+					rs.events <- sqlcapture.ChangeEvent{
+						Operation: sqlcapture.DeleteOp,
+						Source:    sourceMeta,
+						Before:    before,
+					}
+				}
+			default:
+				return fmt.Errorf("unknown row event type: %q", event.Header.EventType)
```
psFried, line 161: Ya, good idea returning an error here. I was just wondering about this after noticing all the versions above, and I think this is the most reasonable behavior given that a `v3` seems inevitable.

source-mysql/replication.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	return fields
+}
+
+func (rs *mysqlReplicationStream) Events() <-chan sqlcapture.ChangeEvent {
+	return rs.events
+}
+
+func (rs *mysqlReplicationStream) Acknowledge(ctx context.Context, cursor string) error {
+	// TODO(wgd): Figure out how MySQL advances whatever WAL-reserving cursor it has.
```
psFried, line 205: This _seems_ like a question that would need to be resolved before we can really run this, unless I'm just missing some context.

source-mysql/README.md
```diff
@@ -0,0 +1,119 @@
+Flow MySQL Source Connector
```
psFried, line 1: This is an exceptionally well written README, as is the one for source-postgres! 👌

psFried (changes requested):
Overall, this looks great to me. I noticed a few things that are worth changing, and left some comments inline. Should be good to merge once those are addressed. Nice work!

source-mysql/main.go
```diff
@@ -0,0 +1,136 @@
...
+		return &mysqlDatabase{config: &config}, nil
+	})
+}
+
+// Config tells the connector how to connect to the source database and
+// capture changes from it.
+type Config struct {
+	Address  string `json:"address" jsonschema:"default=127.0.0.1:3306,description=Database host:port to connect to."`
+	User     string `json:"user" jsonschema:"default=flow_capture,description=Database user to connect as."`
+	Pass     string `json:"pass" jsonschema:"description=Password for the specified database user."`
```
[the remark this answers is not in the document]
willdonnelly: Done.

source-mysql/main.go
```diff
@@ -0,0 +1,136 @@
...
+	config        *Config
+	conn          *client.Conn
+	defaultSchema string
+}
+
+func (db *mysqlDatabase) Connect(ctx context.Context) error {
+	logrus.WithFields(logrus.Fields{
+		"addr":     db.config.Address,
+		"user":     db.config.User,
+		"pass":     db.config.Pass,
```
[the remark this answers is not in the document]
willdonnelly: Oh yeah, that was a dumb move. Similarly it turned out that `source-postgres` was still logging the password as well, albeit indirectly via the `Config.ToURI()` method. I've checked all references to the password field (and `ToURI()`) and made sure it's not getting logged anywhere else now.

source-mysql/replication.go
willdonnelly: Yeah, this connector won't be ready for production usage until I finish following up on that and ideally add a test (analagous to the PostgreSQL source `TestSlotLSNAdvances`) to validate that behavior. But I'm pretty sure it won't require any fundamental changes to the overall architecture of the code, so I figured it past time to get the review started.

source-mysql/main.go
```diff
@@ -0,0 +1,136 @@
...
+	User     string `json:"user" jsonschema:"default=flow_capture,description=Database user to connect as."`
+	Pass     string `json:"pass" jsonschema:"description=Password for the specified database user."`
+	DBName   string `json:"dbname" jsonschema:"description=Name of the database to connect to."`
+	ServerID int    `json:"server_id" jsonschema:"description=Server ID for replication."`
+
+	WatermarksTable string `json:"watermarks_table,omitempty" jsonschema:"default=flow.watermarks,description=The name of the table used for watermark writes during backfills."`
+}
+
+// Validate checks that the configuration passes some basic sanity checks, and
+// fills in default values when optional parameters are unset.
```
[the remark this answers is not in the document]
willdonnelly: I like the explicit `SetDefaults()` method, done.

source-mysql/replication.go
willdonnelly: Yeah, in general I think it's best to die quickly when we see something unexpected in a replication event stream, since we don't want to make assumptions about the semantics and possibly produce corrupt output. I'm actually torn on having the `v0` and `v1` rows events be accepted, since I haven't personally seen them in my testing.

source-mysql/README.md
willdonnelly: Thanks!

source-mysql/replication.go
```diff
@@ -0,0 +1,214 @@
...
+		Flavor:   "mysql", // TODO(wgd): See what happens if we change this and run against MariaDB?
+		Host:     host,
+		Port:     uint16(port),
+		User:     db.config.User,
+		Password: db.config.Pass,
+	})
+
+	var pos mysql.Position
+	if startCursor != "" {
+		var binlogName, binlogPos, err = splitHostPort(startCursor)
```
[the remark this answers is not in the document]
willdonnelly: Yeah, the backstory is pretty much that I already had the `splitHostPort` function written to handle the address-splitting a few lines above here, and then when I added the resume cursor logic it needed the exact same behavior, but I also couldn't think of a better name for the function.

On reflection I think the right answer is actually to duplicate a few lines of code and define a second `splitCursor` function, because that lets me reword the error messages for each function to be more contextually useful.

source-mysql/replication.go
willdonnelly: On further reflection I've decided to remove the `v0` row change events from the acceptable list for now. Like I mentioned I've only ever seen the `v2` events in my testing, and I've looked at enough other examples of MySQL binlog replication that I'm comfortable accepting the `v1` events as being identical for our purposes, but I don't have a similar amount of confidence that `v0` change events would work correctly.

psFried: 👍 Sounds good

source-mysql/replication.go
psFried: ok, gotcha :+1:

psFried (approved):
I just peeked at the last round of changes, and I think this is looking good. I'm not 100% clear on what else is required in order to get this production ready. Is the only remaining issue just the `Acknowledge` implementation? Either way, I think it's fine to merge this as it is, and tackle the remaining work in a follow on PR.

willdonnelly:
Yeah, the `Acknowledge` implementation is the main piece of work that I know to be required, however I also still need to subject `source-mysql` to the same level of correctness stress-testing as `source-postgres`.

willdonnelly:
Oh, also extending the set of supported datatypes, currently that's just stubbed out enough to make the generic tests pass with support for `int`, `varchar`, `text`, and `double`.

A bit later this afternoon I'll file tracking bugs for MySQL acknowledgements, stress testing, and datatype support, just to make sure they don't get forgotten over the holidays.

willdonnelly:
I'm never sure what the etiquette is when there are multiple reviewers, but I'm inclined to go ahead and merge this today.

@jgraettinger if there's anything you'd like changed I'm happy to make followup changes in another PR.

jgraettinger:
I should have taken myself off. Sounds good, please merge.

willdonnelly:
I've filed issues for the areas of improvement discussed previously:

- https://github.com/estuary/connectors/issues/92
- https://github.com/estuary/connectors/issues/93
- https://github.com/estuary/connectors/issues/94

PR #87, opened 2021-12-20
Fix e2e tests and update parquet output paths

psFried:
**Description:**

This PR rolls up a few different changes. 

- Fixes the e2e tests, which were broken by the removal of `BUILDS_ROOT` from Flow
- Changes the output paths of parquet files to use a padded key range and sequence number, so that a default bucket listing (lexicographic sort) of the files from a single shard will match the order in which they were written.
- Changes the file extension to the more standard `.parquet`

**Workflow steps:**

The workflow itself is unchanged, but the files output by `materialize-s3-parquet` should now be easier to work with when reading them from the S3 bucket.

**Documentation links affected:** none 

**Notes for reviewers:** best to review commits individually


jixij (approved):
:+1:  Looking great.

PR #91, opened 2021-12-23
Rockset bulk ingestion from S3

psFried:
**Description:**

Adds support for creating Rockset collections that initialize data by bulk ingestion from an S3 bucket. Also adds support for handing off from the `materailize-s3-parquet` connector so that it can be used to materialize into S3 to setup for the bulk ingestion.

**Workflow steps:** See the bulk ingestion section in: [materialize-rockset/README.md](https://github.com/estuary/connectors/pull/91/files#diff-1cda328f2ea215b050a6006572dea88bc0852ed6a51c9b6b1c0f97860f90241a)

I created [this link](https://go.estuary.dev/rock-bulk) but have not directed it anywhere yet.

**Notes for reviewers:**

I've manually tested this, but haven't yet added any automated tests. It seems impractical to continue using the current integration testing strategy for bulk loading because it requires a minimum of 5GiB of data in the S3 bucket and at least a medium size Rockset virtual instance. Mocking the Rockset REST API seems like the best approach, but I think it makes sense to kick the tires on this a little more before investing the time into automated tests.


materialize-rockset/transactor.go
```diff
@@ -57,14 +58,48 @@ type transactor struct {
...
+				)
+				waitGroup.Done()
+			}(i)
+		}
+	}
+	waitGroup.Wait()
+	for i, err := range errors {
+		if err != nil {
+			return fmt.Errorf("awaiting bulk ingestion completion for rockset collection '%s': %w",
+				t.bindings[i].rocksetCollection(), err)
```
saterus, line 87: Do we specifically want to wait for all the collections to finish/error before reporting errors? I think you can use https://pkg.go.dev/golang.org/x/sync/errgroup to help here. It will handle propagating the first error from a routine up to the caller and cancelling the others.

materialize-rockset/rockset.go
```diff
@@ -67,31 +77,98 @@ func getCollection(ctx context.Context, client *client, workspace string, collec
...
+		}
+
+		var ready = collection.Status == STATUS_READY
+
+		// We'll check to see if a collection has made no progress and emit a specific warning in that case.  If a
+		// Rockset collection isn't using a large enough instance, then it'll silently refuse to actually use the
+		// integration that was configured. This is a really easy mistake to make, and there doesn't appear to be a way
+		// to actually detect the instance size via the API, so we'll try to log a helpful warning if the collection
+		// goes too long without downloading any of the objects from cloud storage.
+		var bulkIngestStalled = false
```
saterus, line 126: Sounds like you're making the best of what we've got. :grimacing:

materialize-rockset/rockset.go
```diff
@@ -67,31 +77,98 @@ func getCollection(ctx context.Context, client *client, workspace string, collec
...
+		var collectionAge = time.Since(collection.CreatedAt)
+		// If an integration was specified, then we need to _also_ wait for that integration to fully process all of its
+		// files. Technically, this will already be the case if there was enough data in the source bucket to trigger a
+		// "bulk load", because in that case Rockset will not set the status to READY until after all the files have
+		// been uploaded. BUT, if the source bucket didn't contain enough data to trigger the "bulk load", then the
+		// status may be set to READY before all of the source files have been processed. This condition is a guard
+		// against ingesting data out of order in that specific case.
+		if integration != "" {
```
saterus, line 134: If they don't specify an integration, couldn't we return at the top of this function? In that case, there is no bulk ingestion to perform, right?

materialize-rockset/rockset.go
```diff
@@ -67,31 +77,98 @@ func getCollection(ctx context.Context, client *client, workspace string, collec
...
+			logEntry.Warn("Rockset collection integration has not made any progress. Is your Rockset virtual instance large enough to support bulk ingestion?")
+		} else {
+			logEntry.Info("Rockset collection integration has not yet caught up (will retry)")
+		}
+
+		time.Sleep(nextIngestCompletionBackoff(collectionAge))
 	}
 }
+
+func nextIngestCompletionBackoff(timeSinceCreation time.Duration) time.Duration {
```
saterus, line 160: We need to have 5gb to even bother with the bulk ingestion mode, right? Do you have any ballpark of how long that might take to complete? Should we even check in the first 5 minutes?

materialize-rockset/README.md
```diff
@@ -0,0 +1,105 @@
...
+          api_key: <your rockset API key here>
+          max_concurrent_requests: 5
+    bindings:
+      - resource:
+          workspace: <your rockset workspace name>
+          collection: <your rockset collection name>
+        source: example/flow/collection
+```
+
+## Bulk ingestion for large backfills of historical data
```
saterus, line 31: It might be worth adding a simple flowchart or something to help describe the state machine that's happening. It's a workflow that requires user intervention at a few steps and could be easy to mess up.

saterus (approved):
This looks really good. I don't have any major concerns, just clarifying questions.

materialize-rockset/driver.go
```diff
@@ -42,9 +44,48 @@ func (c *config) Validate() error {
...
+}
+
+// Configuration for bulk loading data into the new Rockset collection from a cloud storage bucket.
+type cloudStorageIntegration struct {
+	// The name of the integration in Rockset, which would have been created in the Rockset UI.
+	Integration string `json:"integration"`
+	// The name of the S3 bucket to load data from.
+	Bucket string `json:"bucket"`
+	// The region of the S3 bucket. Optional.
+	Region string `json:"region"`
```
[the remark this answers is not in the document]
psFried: I haven't actually verified this, but I think they may try to discover the region automatically, and this is only required if they don't have permission. I works if I don't specify it with a test bucket is in us-east-2. I'm hesitant to require it, since Rockset doesn't and their users may have the expectation that it's optional.

materialize-rockset/driver.go
```diff
@@ -42,9 +44,48 @@ func (c *config) Validate() error {
...
+	// Prefix of the data within the S3 bucket. All files under this prefix will be loaded. Optional. Must not be set if
+	// 'pattern' is defined.
+	Prefix string `json:"prefix,omitempty"`
+}
+
+func (c *cloudStorageIntegration) Validate() error {
+	var requiredProperties = [][]string{
+		{"integration", c.Integration},
+		{"bucket", c.Bucket},
+		{"prefix", c.Prefix},
```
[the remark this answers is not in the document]
psFried: Whoops, good catch

materialize-rockset/driver.go
```diff
@@ -224,9 +311,13 @@ func (d *rocksetDriver) Transactions(stream pm.Driver_TransactionsServer) error
 		client:   client,
 		bindings: bindings,
 	}
+	// If any bindings have an integration named as part of their resource spec, then await them now, before returning
+	// the opened response. It's important that we await _all_ bindings that are bulk loading before continuing, since
+	// the flow checkpoint that's embedded in the driver checkpoint will be cleared on the first successful transaction.
```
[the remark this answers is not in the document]
psFried: Yeah, that's right. There's just one checkpoint for the whole materialization, so you can't merge two into one without starting over at the beginning.

materialize-rockset/driver.go
```diff
@@ -64,10 +105,17 @@ func (r *resource) Validate() error {
 	if err := validateRocksetName("collection", r.Collection); err != nil {
 		return err
 	}
+	// TODO: move this mutating operation out of Validate
 	if r.MaxBatchSize == 0 {
 		r.MaxBatchSize = 1000
 	}
```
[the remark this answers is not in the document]
psFried: I think an explicit `SetDefaults` function is a worthwhile improvement for now, which is what we've also done elsewhere. There's maybe a decent pattern where we could have an interface that defines a `SetDefaults` function and have a generic unmarshal function use that, but it's a little more than I want to get into right now.

materialize-rockset/driver.go
```diff
@@ -210,6 +278,25 @@ func (d *rocksetDriver) Transactions(stream pm.Driver_TransactionsServer) error
...
 
+	// It's possible that there was a previous materialization with the same name as this one, and that it may have left
+	// a driver checkpoint. This is expected in the case that a user wishes to use a cloud storage integration in
+	// Rockset for bulk loading data for a large backfill as described here: https://go.estuary.dev/rock-bulk
+	// If such a checkpoint is left over, then we'll return it here. The driver checkpoint will then be cleared out
+	// on the next successful commit of a transaction.
+	var flowCheckpoint []byte
+	var s3DriverCheckpoint checkpoint.DriverCheckpoint
+	if len(open.Open.DriverCheckpointJson) > 0 {
```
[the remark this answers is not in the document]
psFried: Good question, I had to look back at the code to verify that this does what I thought it did. It effectively just takes the the new value, which is an empty byte slice. It does this because we _don't _ use merge patches for materialization checkpoints. [Here's the code](https://github.com/estuary/protocols/blob/main/flow/checkpoint_extensions.go) that does the reduction of the old and new checkpoints. For materializations, `Rfc7396MergePatch` is always `false`.

materialize-rockset/rockset.go
psFried: Technically, we await completion at the start of every processing term, so I think it's best to just check early so we don't hold things up in the common case where the collection is already `READY`. Additionally, we create the collection during `Apply` and do this await during `Transactions`, so there could already be some time that's passed.

materialize-rockset/transactor.go
psFried: Oh neat, I didn't know about `errgroup`, but it's exactly what I wanted here. Thanks for the suggestion.

materialize-rockset/rockset.go
psFried: No, because this may still be needed in order to await the `READY` status of a collection that was created without any integration. Previously, this was being done as part of creating the collection during `Apply`, but I ripped it out of there so that this is the only place where we wait for the collection to be ready.

materialize-rockset/README.md
psFried: Yeah, that's a good idea. I think perhaps the user-facing docs might be a better place for it, though, since that would allow using mermaid. @oliviamiannone does that seem reasonable to you? I can help with writing the mermaid if you want.

oliviamiannone: That sounds good to me, I'm not familiar with mermaid so a bit of help would probably save substantial time on my end! Also you might have a better idea of what it should look like. Though I'm happy to contribute as much as I can.

dyaffe (commented):
A few comments after running through this once.  They're all in materialize-rockset/readme.md:

It would help to link in (2) to how to the Flow Template repository so that users can see how to start up.  I could additionally see it making sense  to link to an article on captures.

The materialization examples are missing “connector:” in the yaml and have a title instead.

Tactically, this requires a lot of working knowledge of Flow -- I for instance had to search around to remember how to deploy when working on (3) and also found that in the Template repository.  Afterwards I hit the Storage mapping issue and had to find syntax for that

oliviamiannone:
@dyaffe those are good points, just incorporated relevant changes into the user facing docs (not merged yet, but you can see them here https://github.com/estuary/flow/pull/324) 
As you point out, this is definitely a more intermediate workflow. Framing it as such, as well as listing prerequisites and providing links to more intro-level content, helps a lot.

PR #98, opened 2022-01-06
CSV missing columns with configured headers

### LINKED ISSUE #97: CSV parser should allow missing columns when headers are explicitly configured
Given an example CSV file that doesn't have a header row:

```csv
1
2,3
4,5,6
```

And the following parser configuration:

```yaml
parser:
  format: csv
  csv:
    headers: [a, b, c]
```

The desired behavior is that the CSV file should parse without error, as long as the fields `b` and `c` are nullable. The current behavior is for the parser to return an error after the first row, because it's expecting each row to only have one column (based on the first row that it read).

We should update the parser to successfully parse the file in this scenario. Note that we should also ensure that a CSV file with _more_ columns than there are headers should still fail. For example, the following CSV file should result in an error at line 2, given the same configuration as above with only 3 headers:

```csv
1
2,3,4,5
```

psFried:
Fixes #97

**Description:**

Adds an explicit check for the number of columns in each CSV row instead
of relying on the underlying parser to fail if a row has too many
columns. This allows us to consistently and correctly check the number
of columns in each row, even when headers are provided in the config
instead of being inline at the top of the file.

**Workflow steps:** no change

Adds the short link: https://go.estuary.dev/QRKf3x as a placeholder for documentation on parser configuration. This link currently only appears in the error message, but is intended to point users to an (advanced) "parser configuration" doc.

**Notes for reviewers:** none

### DIFF
--- a/parser/src/format/character_separated/mod.rs
+++ b/parser/src/format/character_separated/mod.rs
@@ -94,7 +94,12 @@ impl Parser for CsvParser {
             .unwrap_or(self.default_delimiter);
 
         let mut builder = csv::ReaderBuilder::new();
-        builder.delimiter(delim);
+        // Configure the underlying parser to allow rows to have more columns than the header row.
+        // This is needed in order to properly parse files using explicitly configured headers
+        // instead of reading the column names from the first row. `CsvOutput` will explicitly check
+        // each row to ensure that it has no more columns than there are headers, which will account
+        // for explicitly configured headers.
+        builder.delimiter(delim).flexible(true);
 
         // The default line ending is CRLF, and we'll stick with that unless the user specifies
         // something different.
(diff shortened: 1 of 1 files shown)


parser/src/format/character_separated/mod.rs
```diff
@@ -406,6 +422,66 @@ mod test {
         Input::Stream(Box::new(Cursor::new(content.into())))
     }
 
+    // Test for: https://github.com/estuary/connectors/issues/97
+    #[test]
+    fn fails_when_there_are_more_columns_than_headers() {
+        let conf = ParseConfig::default();
+
+        // CSV has 2 column headers, but row 3 has 3 columns :boom:
+        let csv = test_input("a,b\n1\n2,3\n4,5,6");
```
saterus, line 431: Maybe split this CSV input into a multi-line string for readability?

saterus (approved):
Looks good. Good test cases. :+1: 

Regarding the docs process, are you going to add that Advanced Parser Configuration page before you merge this? Or will we backfill that later?

parser/src/format/character_separated/mod.rs
psFried: Generally for inline CSVs I've preferred to leave it as a single line because multiline rust strings don't automatically strip indentation. There's a crate that has a procedural macro for it, but I think my preference would be to have any test cases that are significantly larger than this one live as separate files and use `include_bytes!`.

PR #104, opened 2022-01-19
materialize-bigquery: Don't assume the root document exists

willdonnelly:
**Description:**

Subsequent to https://github.com/estuary/protocols/pull/22 (which relaxes the requirement so that a binding for the root document is only `RECOMMENDED` in delta-updates mode), we still need to modify the actual connectors to work with this configuration.

This PR tweaks the BigQuery materialization so that it doesn't assume the existence of a binding for the root document when in
delta-updates mode.

**Workflow steps:**

Once this all works, a catalog entry like:

```yaml
materializations:
  wgd/example/to-bigquery:
    endpoint:
      connector:
        image: ghcr.io/estuary/materialize-bigquery:dev
        config: wgd-materialize-bigquery.config.yaml
    bindings:
      - source: wgd/example/data
        resource: 
          table: "example_table"
          delta_updates: true
        fields:
          recommended: true
          exclude: ["flow_document"]
```

Should be possible. Nothing should change in cases where `flow_document` is not excluded or `delta_updates` is false.

I do not believe the `flow_document` field is discussed in our documentation at present, so no changes here.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/104)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/brianvoe/gofakeit/v6 v6.9.0
 	github.com/elastic/go-elasticsearch/v7 v7.15.2-0.20211104170009-caf202868441
 	github.com/estuary/flow v0.1.1-0.20211207135900-65dc2d2b030a
-	github.com/estuary/protocols v0.0.0-20211129055338-9259b2dca80a
+	github.com/estuary/protocols v0.0.0-20220120193955-e69c14edbdd1
 	github.com/evanphx/json-patch/v5 v5.5.0
 	github.com/go-mysql-org/go-mysql v1.4.0
 	github.com/gogo/protobuf v1.3.2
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/104)* status: all files reviewed, 1 unresolved discussion (waiting on @willdonnelly)

___
*[materialize-bigquery/transactor.go, line 155 at r1](https://reviewable.io/reviews/estuary/connectors/104#-Mtt072I7P5efJ4H9f08:-Mtt072I7P5efJ4H9f09:b-7xr11x) ([raw file](https://github.com/estuary/connectors/blob/ff8a93ca975d826506777b6372bf978f23306bfa/materialize-bigquery/transactor.go#L155)):*
> ```Go
> 
> 		// Convert all the values to database appropriate ones and store them in the GCS file.
> 		var vals = append(it.Key, it.Values...) // TODO(wgd): Was destructive reuse of `it.Key` intentional here?
> ```

I don't believe so. It's not _incorrect_ as written, but it probably does reallocate every iteration since len(Key) ~= cap(Key).

optional nit: hoist `vals` slice out of the loop, and `append(append(vals[:0], it.Key...), it.Values...)` ? Might be less confusing to readers.

jgraettinger (commented):
LGTM

> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/104#-:-Mtt0nB0DNA642DI_NHs:bb6ywmx)* status: all files reviewed, 1 unresolved discussion (waiting on @willdonnelly)

willdonnelly:
> optional nit: hoist `vals` slice out of the loop, and `append(append(vals[:0], it.Key...), it.Values...)` ? Might be less confusing to readers.

Done. I also made two other small tweaks:

- Changed the load query `b.load.sql`'s to have a better dummy value in delta-updates mode: `ASSERT false AS 'Load queries should never be executed in Delta Updates mode.'`. So hopefully we get a cleaner error if some future change led to that being executed by accident.
- Ran `go get -u github.com/estuary/protocols` to pick up https://github.com/estuary/protocols/pull/22, so that once this is merged the CI-built connector will actually work as intended.

jgraettinger (approved):
still LGTM

> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/104#-:-MttNJkt9jeRWDeqMMgy:bh5f435)* status: all files reviewed, 1 unresolved discussion (waiting on @willdonnelly)

PR #106, opened 2022-01-24
materialize-bigquery: Allow setting billing_project_id separately

willdonnelly:
**Description:**

This change adds a new `billing_project_id` configuration parameter (optional, defaults to `project_id` if unset) which will be given to BigQuery (via `bigquery.NewClient`) for billing purposes.

However once we do this, all queries will "conveniently" use that project ID as the default for their own name resolution. To fix that and make the two concepts properly decoupled I had to change a couple places where table names are assembled from "dataset.table" to a fully-qualified "project_id.dataset.table".

I have verified that this works well enough to get through the previously-failing validation step during a catalog build, but haven't tried running the full catalog with this change yet.

**Workflow steps:**

Specify `billing_project_id` in a BigQuery materialization config.

### DIFF
--- a/materialize-bigquery/bigquery.go
+++ b/materialize-bigquery/bigquery.go
@@ -22,13 +22,14 @@ func main() {
 
 // Config represents the endpoint configuration for BigQuery.
 type config struct {
-	ProjectID       string `json:"project_id"`
-	Dataset         string `json:"dataset"`
-	Region          string `json:"region,omitempty"`
-	Bucket          string `json:"bucket"`
-	BucketPath      string `json:"bucket_path"`
-	CredentialsFile string `json:"credentials_file,omitempty"`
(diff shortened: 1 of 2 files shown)


materialize-bigquery/bigquery.go
```diff
@@ -49,7 +50,7 @@ func (c *config) Validate() error {
 
 // DatasetPath returns the sqlDriver.ResourcePath including the dataset.
 func (c *config) DatasetPath(path ...string) sqlDriver.ResourcePath {
-	return append([]string{c.Dataset}, path...)
+	return append([]string{c.ProjectID, c.Dataset}, path...)
```
jgraettinger, line 53: good catch! i suspect we'll be thankful for this in the future...

jgraettinger (approved):

PR #108, opened 2022-01-26
Use protocols from flow repo

psFried:
**Description:**

Updates the flow protocols dependency to use the protocols packages from
the flow repo. The estuary/protocols repo has been re-combined with the
flow repo and the old one will be archived. This removes the dependency
on that repo entirely.

**Workflow steps:**

No change for end users. Developers will now make protocol updates in the main flow repo instead of in estuary/protocols.

**Notes for reviewers:** 

Corresponding flow pr is: estuary/flow#347


jgraettinger (approved):
LGTM

PR #110, opened 2022-01-27
sqlcapture: don't error out on type translation failure

willdonnelly:
**Description:**

When performing discovery, the existence of any table with an unhandled column type can break the entire process. It would be nice if we could simply note the failure and continue, especially since the user may not even care about capturing that table, or about that column of that table at least.

It might be cleanest eventually to just log a warning and not include the troublesome column in the generated schema at all. Sadly, flowctl swallows any warnings we produce, so to get it to the user we have to embed the error into the JSON schema output.

**Workflow steps:**

Currently the easiest way to hit this is to perform discovery when there exists a table with a column of an array type that hasn't been special-cased in.

### DIFF
--- a/sqlcapture/discovery.go
+++ b/sqlcapture/discovery.go
@@ -46,9 +46,19 @@ func DiscoverCatalog(ctx context.Context, db Database) (*airbyte.Catalog, error)
 		for _, column := range table.Columns {
 			var jsonType, err = db.TranslateDBToJSONType(column)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


jixij (approved):
LGTM! 
I was able to run the discover and see the error description as expected.
```
segment_ids:
              description: 'ERROR: could not translate column type "_numeric" to JSON schema: unhandled PostgreSQL type "_numeric"'
```
Thank you!

PR #117, opened 2022-02-03
source-postgres: handle TOAST values in replication events

jgraettinger:
**Description:**

Update source-postgres to omit fields which are unavailable due to TOAST.
A new tests runs through the possibilities for fields under various `REPLIA IDENTITY` settings.

Also add reduction annotations to generated schemas which "do the right thing",
by merging top-level properties of captured event documents.

## Testing

This was tested manually using source-postgres => sqlite materialization.
I ran `flowctl discover` to generate an (unmodified) flow catalog and schemas.
I did a bunch of INSERT and UPDATES into a CDC'd table,
including generating large values which triggered TOAST handling, and playing with different `REPLICA IDENTITY` values.
I observed that events would omit the unchanged TOAST values (as expected), but the materialization correctly rolled up the reduced document values. I also confirmed It honors the `firstWriteWins` semantics of `/_meta/before`

**Workflow steps:**

The workflow doesn't change.  Use `flowctl discover` using the connector image, and then use (e.g. materialize) the collections it produces.

We do need to document the TOAST semantics of postgres captures, for users who fall off of the garden path.
That's not tackled here.

**Notes for reviewers:**

See individual commits.


source-postgres/docker-entrypoint-initdb.d/init-user-db.sh
```diff
@@ -1,14 +1,15 @@
 #!/bin/bash
 set -e
 
-# The alter system command at the end is redundant due command line settings. 
-#  Included as an example of how to set this without using the command line. 
+# The alter system command at the end is redundant due command line settings.
+#  Included as an example of how to set this without using the command line.
 psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" --dbname "$POSTGRES_DB" <<-EOSQL
     CREATE USER flow_capture WITH PASSWORD 'flow_capture' REPLICATION;
+    ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES to flow_capture;
```
willdonnelly, line 8: This is a good addition, and should probably be added to the setup instructions in `README.md` and our docs as well.

willdonnelly (approved):
LGTM

source-postgres/docker-entrypoint-initdb.d/init-user-db.sh
willdonnelly: Ah, I just noticed you mentioned that in the Slack thread too.

Issue #109, opened 2022-01-27
sqlcapture: Make state updates smaller using merge/patch semantics


willdonnelly:
Currently state updates from SQL capture connectors are [serialized in their entirety](https://github.com/estuary/connectors/blob/9d9da6f73255173f2fddb530c84faf49ad536c76/sqlcapture/capture.go#L440) every time we emit an update.

This probably adds up to meaningful amounts of data once you multiply N tables * M checkpoints.

We could be a lot smarter about this by keeping track of a per-stream `dirty` flag, only serializing the `TableState`s of those tables which actually changed, and setting [`Merge=true`](https://github.com/estuary/flow/blob/master/go/protocols/airbyte/catalog.go#L202).

willdonnelly:
(Implicit in the above is that in the steady state all our state updates occur during [`streamToWatermark`](https://github.com/estuary/connectors/blob/9d9da6f73255173f2fddb530c84faf49ad536c76/sqlcapture/capture.go#L249) when we observe a transaction commit via logical replication, and in these cases _literally all we want_ is to bump the cursor value, since there should never be any stream-state changes at this point. Keeping track of per-stream dirty flags is actually overkill for that bit, but still worthwhile overall because it improves the other edge cases a bit.)

PR #126, opened 2022-02-10
sqlcapture: Use merge checkpoints for scalability

willdonnelly:
**Description:**

Previously, each state/checkpoint update output by the source-postgres or source-mysql connectors would restate the state of every table in full. After this change, a per-table "dirty" flag is maintained and used to only emit those tables which have changed since the previous checkpoint.

This fixes https://github.com/estuary/connectors/issues/109

As a special case, this means that during the normal course of tailing replication events after all tables are backfilled, each update will consist of just a new cursor value and the "merge = true" flag.

Previously, the main concern was a situation with dozens or hundreds of tables being captured, where one or more of those tables contains a lot of rapid, small transactions. Since each of those transactions would cause the entire list of tables to be restated, this could in principle cause journal size to blow up several orders of magnitude larger than the actual data. Of course gzipping of journal fragments would reclaim some of that inefficiency, but it's still better to put a cap on it by making these state updates constant in size.

(Note that in general state updates can still vary in size, but that should only happen during backfills, when tables are progressing their "Scanned" upper bound and transitioning into the "Active" state).

**Workflow steps:**

None, this should Just Work in the future when using a newer build of `source-postgres`.

None, to my knowledge.

**Notes for reviewers:**

I feel like the question of "what do we do when the table was deleted and then is later added again" (as exercised in `TestMultipleTables` capture 5) is kind of a tricky edge case in general, and merge updates made it just a bit more complicated. The old behavior was to delete the table from the state entirely when it's removed from the catalog, and thus adding it back in later on would cause the capture to start from the beginning of that table.

For now I'm preserving that behavior in merge-update mode by using an `"Ignored"` state to represent having been deleted. There's no mechanism to ever clean up those `"Ignored"` tombstones though, so I'm not super happy with this either.


willdonnelly:
One additional note: In the test snapshots it looks like this doubles the overall number of state checkpoints, but this is an illusion caused by the "deduplication" logic which squashes down identical (post-cursor-sanitization) state updates into a single event. (That, in turn, is done because it improves test stability quite a bit -- even when the test suite is the only thing hitting the database, sometimes the DB sends us "unexpected" commit events for various reasons.)

So what's actually happening there is that two successive state outputs which were previously identical have now become two different-looking state outputs because only one of them needs to include the modified table backfill progress (the other one is bumping the restart LSN cursor).

jixij (approved):

jixij (approved):
LGTM.  The logic to convert `updating the overall state` into `merging` only the `dirty streams` looks nice!

jgraettinger (approved):
> Reviewed 62 of 62 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/126)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @willdonnelly)

PR #127, opened 2022-02-10
Add titles to generated Kafka spec schema

saterus:
**Description:**

Schemars will render "title" attributes if you use a Markdown heading in the doc comment. This should make it easier to render UI for some of these elements.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

The snapshot of the spec output is probably the main thing to look at. Otherwise it's just tweaking comments.


travjenkins (approved):
lgtm

PR #128, opened 2022-02-11
encode pgtype.numeric ids

jixij:
**Description:**

Per discussions, adding the logic to transform the numeric fields in keys in to a tuple, to make sure 1. the tuple conforms to the FoundationDB tuple format. 2. The fdb-encoded byte slice preserve the order of the original numeric value.

**Workflow steps:**

**Documentation links affected:*

**Notes for reviewers:**

### DIFF
--- a/source-postgres/fdbutils.go
+++ b/source-postgres/fdbutils.go
@@ -11,8 +11,8 @@ import (
 
 const numericKey = "N"
 
-// EncodePgNumericKeyFDB encodes a pgtype.Numeric value to an FDB tuple.
-func EncodePgNumericKeyFDB(key pgtype.Numeric) (tuple.Tuple, error) {
+// encodePgNumericKeyFDB encodes a pgtype.Numeric value to an FDB tuple.
+func encodePgNumericKeyFDB(key pgtype.Numeric) (tuple.Tuple, error) {
 	if encodedKey, err := encodePgNumeric(key); err != nil {
 		return nil, fmt.Errorf("encode pgtype.Numeric: %w", err)
 	} else if encodedBinary, err := key.EncodeBinary(nil, nil); err != nil {
@@ -22,19 +22,14 @@ func EncodePgNumericKeyFDB(key pgtype.Numeric) (tuple.Tuple, error) {
 	}
 }
 
-// DecodeFDBTuple decodes the input TupleElement to its original value.
-// Return nil if the input cannot be decoded.
-func DecodeFDBTuple(t tuple.TupleElement) interface{} {
-	switch t := t.(type) {
-	case tuple.Tuple:
-		if len(t) == 3 {
-			if key, ok := t[0].(string); ok && key == numericKey {
-				if b, ok := t[2].([]byte); ok {
-					var n pgtype.Numeric
-					if err := n.DecodeBinary(nil, b); err == nil {
-						return n
-					}
-				}
+// maybeDecodePgNumericTuple tries to decode the input tuple as a pgNumeric value.
+// returns the decoded value if succeeds, and nil if not.
+func maybeDecodePgNumericTuple(t tuple.Tuple) interface{} {
+	if len(t) == 3 && t[0] == numericKey {
+		if b, ok := t[2].([]byte); ok {
+			var n pgtype.Numeric
+			if err := n.DecodeBinary(nil, b); err == nil {
+				return n
 			}
 		}
 	}
--- a/source-postgres/fdbutils.go
+++ b/source-postgres/fdbutils.go
@@ -1,0 +1,115 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/estuary/flow/go/protocols/fdb/tuple"
+	"github.com/jackc/pgtype"
+)
+
+const numericKey = "N"
+
+// EncodePgNumericKeyFDB encodes a pgtype.Numeric value to an FDB tuple.
+func EncodePgNumericKeyFDB(key pgtype.Numeric) (tuple.Tuple, error) {
+	if encodedKey, err := encodePgNumeric(key); err != nil {
+		return nil, fmt.Errorf("encode pgtype.Numeric: %w", err)
+	} else if encodedBinary, err := key.EncodeBinary(nil, nil); err != nil {
+		return nil, fmt.Errorf("encode binary pgtype Numeric: %w", err)
+	} else {
+		return tuple.Tuple{numericKey, encodedKey, encodedBinary}, nil
+	}
+}
+
+// DecodeFDBTuple decodes the input TupleElement to its original value.
+// Return nil if the input cannot be decoded.
+func DecodeFDBTuple(t tuple.TupleElement) interface{} {
+	switch t := t.(type) {
+	case tuple.Tuple:
+		if len(t) == 3 {
+			if key, ok := t[0].(string); ok && key == numericKey {
+				if b, ok := t[2].([]byte); ok {
+					var n pgtype.Numeric
+					if err := n.DecodeBinary(nil, b); err == nil {
+						return n
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
+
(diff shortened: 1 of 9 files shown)


sqlcapture/tests/tests.go
```diff
@@ -239,3 +247,134 @@ func testCatalogPrimaryKeyOverride(ctx context.Context, t *testing.T, tb TestBac
 	})
 	VerifiedCapture(ctx, t, tb, &catalog, &state, "capture2")
 }
+
+func testEncodePgNumeric(t *testing.T) {
+	var testCases = []struct {
```
willdonnelly, line 252: It's good to have a few human-written test cases to validate the encoding looks like we expect, but I think there should be fewer of these, and then a lot more programmatic tests.

I would recommend trimming down this test table, and writing a second test which:
- Generates a list of K random numbers, where K=1,000 or 10,000 or whatever makes the test run in about 1s
- Encodes each of these numbers down to the final `[]byte` FDB serialization
- Decodes the numbers again and checks the result, just to make sure the whole process works end-to-end
- Iterates over every pair of numbers and checks that the ordering of the `[]byte` serialization matches that of the original numbers.

sqlcapture/tests/tests.go
```diff
@@ -239,3 +247,134 @@ func testCatalogPrimaryKeyOverride(ctx context.Context, t *testing.T, tb TestBac
...
+		},
+	}
+	for _, test := range testCases {
+		var actual, err = sqlcapture.EncodePgNumeric(test.input)
+		require.NoError(t, err)
+		require.Equal(t, test.expected, actual)
+	}
+
+	for p1 := 0; p1 < len(testCases); p1++ {
+		for p2 := 0; p2 < len(testCases); p2++ {
```
willdonnelly, line 362: If you're confident that the test is properly commutative (which it should be) you can just say `p2 <= p1` here to avoid a bunch of redundant work.

sqlcapture/tests/tests.go
```diff
@@ -239,3 +247,134 @@ func testCatalogPrimaryKeyOverride(ctx context.Context, t *testing.T, tb TestBac
 	})
 	VerifiedCapture(ctx, t, tb, &catalog, &state, "capture2")
 }
+
+func testEncodePgNumeric(t *testing.T) {
```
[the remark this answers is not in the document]
jixij: Sounds good. Thanks for the suggestion. The change looks pretty simple, and I've made it.

sqlcapture/tuples.go
```diff
@@ -60,11 +73,95 @@ func unpackTuple(bs []byte) ([]interface{}, error) {
...
 	}
+
 	return xs, nil
 }
 
 func compareTuples(xs, ys []byte) int {
 	return bytes.Compare(xs, ys)
 }
+
+func reverseNegfraction(digits string) []byte {
```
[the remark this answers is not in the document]
jixij: cool. Done

sqlcapture/tests/tests.go
jixij: Done.

sqlcapture/tests/tests.go
jixij: In the new test, the test cases are sorted, so this no longer applies.

jixij:
Thanks Will for the suggestions. I've made the changes. PTAL!

source-postgres/fdbutils.go
```diff
@@ -0,0 +1,115 @@
...
+				if b, ok := t[2].([]byte); ok {
+					var n pgtype.Numeric
+					if err := n.DecodeBinary(nil, b); err == nil {
+						return n
+					}
+				}
+			}
+		}
+	}
+	return nil
```
willdonnelly, line 41: If the input cannot be decoded this should probably be a fatal error. These tuple values should only ever come from our own encoding logic, so a failure here means that either a logic bug or some corruption of the actual encoded key has taken place.

If we silently replace that with a `nil` and the user has specified their own scan key which includes a nullable column, this `nil` will be used as the actual resume column value for backfills and could result in us missing or duplicating certain rows in the capture.

source-postgres/fdbutils.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+// DecodeFDBTuple decodes the input TupleElement to its original value.
+// Return nil if the input cannot be decoded.
+func DecodeFDBTuple(t tuple.TupleElement) interface{} {
+	switch t := t.(type) {
+	case tuple.Tuple:
+		if len(t) == 3 {
+			if key, ok := t[0].(string); ok && key == numericKey {
```
willdonnelly, line 31: Style note: I'm pretty sure you can just write `t[0] == numericKey` to get the same logical behavior here.

I'd recommend combining the two levels of checks into something like `len(t) == 3 && t[0] == numericKey` because that makes the pattern more clear if/when we end up adding more custom encoding cases in the future.

sqlcapture/interface.go
```diff
@@ -106,6 +107,12 @@ type Database interface {
 	DefaultSchema(ctx context.Context) (string, error)
 	// Returns an empty instance of the source-specific metadata (used for JSON schema generation).
 	EmptySourceMetadata() SourceMetadata
+	// EncodeRowKeyForFDB converts a key as necessary to produce a TupleElement,
+	// which is encoded as part of a FoundationDB serialized tuple.
+	// Make sure the conversion is partial-order-preserving.
+	EncodeKeyFDB(key interface{}) (tuple.TupleElement, error)
+	// DecodeKeyFDB decodes the result of `EncodeKeyFDB` to its original form.
+	DecodeKeyFDB(t tuple.TupleElement) interface{}
```
willdonnelly, line 115: This should probably be `(interface{}, error)` because decoding external data always has the possibility of failure.

sqlcapture/capture.go
```diff
@@ -291,15 +291,15 @@ func (c *Capture) streamToWatermark(replStream ReplicationStream, watermark stri
...
+		var rowKey, err = encodeRowKey(tableState.KeyColumns, event.KeyFields(), c.Database)
 		if err != nil {
 			return fmt.Errorf("error encoding row key: %w", err)
 		}
 		if compareTuples(rowKey, tableState.Scanned) <= 0 {
 			if err := c.handleChangeEvent(event); err != nil {
 				return fmt.Errorf("error handling replication event: %w", err)
 			}
-		} else if err := results.Patch(streamID, event); err != nil {
+		} else if err := results.Patch(streamID, event, c.Database); err != nil {
```
willdonnelly, line 302: This is an entirely adequate way of doing things, but since we're adding an argument to the `Patch` method anyway it would be slightly more efficient to pass in `rowKey` which we already computed a few lines prior.

willdonnelly (changes requested):
Looking good. A couple of optional style nits, and one actual complaint (decoding should have an error return, silently returning `nil` could mask a logic bug and result in incorrect captures).

source-postgres/fdbutils.go
jixij: I see. Here needs some explanations,
`DecodeFDBTuple` is called by `DecodeKeyFDB`, which is the external API of the database.
- The logic of `DecodeFDBTuple` is that - if it can decode the tuple, it will return the interpretation results. And if it cannot, it will return nil, indicating that `I cannot interpret it`.

- The logic of `DecodeKeyFDB` is that - if `DecodeFDBTuple` (or any other API in the future) can interpret the tuple, it will return the interpretation results, and if not, it will return the input directly without any interpretations.

So the external  API calls to `DecodeFDBTuple`  never gets a `nil` if the input is not `nil`. And this also explains why `DecodeFDBTuple` does not return any error, b/c if anything results an error, `DecodeFDBTuple` will simply return the input.

(I think the logic might be convoluted, which result in the mis-understandings. I will think about and improve it. There could be a restriction on if the input is a `tuple.Tuple`, it must be interpreted to sth.)

jixij:
Thank you again Will, I've made improvements based on your suggestions and feedbacks. PTAL!

willdonnelly (approved):
LGTM

PR #130, opened 2022-02-14
remove rust code and update program names

psFried:
**Description:**

Use Rust binaries from Flow repo

- Deletes `parser` and switches to using `flow-parser` downloaded from
  Flow releases.
- Deletes `network-proxy-service` and switches to using `flow-network-proxy` downloaded from
  Flow releases.
- Deletes `elasticsearch-schemabuilder` and switches to using `flow-schemalate` downloaded from
  Flow releases.

**Workflow steps:** 

No end-user impact, but developers will need to run `./fetch-flow.sh` prior to running tests or building docker images.

**Notes for reviewers:**

The requirement to run `./fetch-flow.sh` before you can build/test is lamentable, but I couldn't think of a good way to make it easier. LMK if you have a better idea.


fetch-flow.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Fetches the latest binary release of Flow and writes the binaries to the `flow-bin/` directory.
+# The release name is read from the `FLOW_RELEASE` env variable, but it defaults to "dev".
+
+FLOW_RELEASE="${FLOW_RELEASE:-dev}"
```
jixij, line 6: Is the `FLOW_RELEASE` a githash, which allows us to pin down some version of rust bins, maybe for trouble shooting?

psFried: No, it's the name of a Github release. The default points to [this one](https://github.com/estuary/flow/releases/tag/dev). You can still use whatever Flow build you want when you're building locally, though. `fetch-flow.sh` just populates the `flow-bin/` directory in the repository root, but it's not run automatically, except in CI. So you can put whatever binaries you want in that directory if you're trying to debug against a specific flow build.

PR #131, opened 2022-02-14
source-postgres: exclude views from discovery

willdonnelly:
**Description:**

Fixes https://github.com/estuary/flow/issues/365

Neither views nor materialized views can be captured by our connector because neither of them actually show up in the WAL. This makes sense, the only problem is that our discovery logic actually _would_ list views as acceptable streams.

The fix is to tweak the table-discovery query SQL to only return information about tables of type `'BASE TABLE'`. Strictly speaking this does more than just exclude views, but from [the documentation](https://www.postgresql.org/docs/current/infoschema-tables.html) it appears that the other table types we might see are `'FOREIGN'` and `'LOCAL TEMPORARY'`, and I don't think we want discovery to return those either.

**Workflow steps:**

Use `flowctl discover` on a PostgreSQL database, and you shouldn't be given results for things we can't actually capture any more.

None

### DIFF
--- a/source-postgres/capture_test.go
+++ b/source-postgres/capture_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/estuary/connectors/sqlcapture/tests"
 	"github.com/jackc/pglogrepl"
 	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
 )
 
 // TestGeneric runs the generic sqlcapture test suite.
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/131#-:-MvuMH5P26tNwPwCa2FR:bb6ywmx)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @willdonnelly)

Issue #54, opened 2021-10-14
source-postgres: Support columns with array types


willdonnelly:
Fairly self-explanatory, PostgreSQL allows columns to be multidimensional arrays of primitive types, and so the PostgreSQL source connector ought to support that properly.

Currently there's some hacked-in and incomplete support: any array will get serialized as a string using PostgreSQL's own text format, and during discovery there's some hacked-in logic to emit the corresponding schema `{"type":"string"}` for a small subset of possible array types (just integers, floats, and text).

Ideally a proper solution will work generically for the dozens of different PostgreSQL data types, and convert each one to an actual JSON array of the equivalent primitive value. How exactly to handle multidimensionality is still TBD, but it might suffice to translate arrays into something like

```
{
  "type": "object",
  "properties": {
    "dims": <something>,
    "data": {
      "type": "array",
      "items": {"type": <elementType>}
    }
  }
}
```

Or possibly even throw away the dimension information entirely and just return the underlying linear value array.

PR #132, opened 2022-02-14
source-postgres: Add support for array types

willdonnelly:
**Description:**

This fixes https://github.com/estuary/connectors/issues/54

PostgreSQL array types are inherently multidimensional, and according to the documentation:

> [The] current implementation ignores any supplied array size limits, i.e., the behavior is the same as for arrays of unspecified length.
> 
> The current implementation does not enforce the declared number of dimensions either. Arrays of a particular element type are all considered to be of the same type, regardless of size or number of dimensions. So, declaring the array size or number of dimensions in CREATE TABLE is simply documentation; it does not affect run-time behavior.

This means that however we translate array values will need to handle varying sizes and dimensionality on a per-row basis. There are basically three solutions that would satisfy this requirement:

* Ignore dimensionality entirely and just emit a flat array of elements, so `{{1,2},{3,4}}` becomes `[1,2,3,4]`.
* Emit nested arrays, so `{{1,2},{3,4}}` becomes `[[1,2],[3,4]]`.
* Emit an object with a `dimensions` field and an `elements` field, so `{{1,2},{3,4}}` becomes `{"dimensions":[2,2],"elements":[1,2,3,4]}`.

The first option loses data unnecessarily, while the second option would require a confusing JSON schema type with recursion and `anyOf` clauses, so I have chosen the "dimensions and elements arrays" approach here.

**Workflow steps:**

Table discovery via `flowctl discover` will now produce useful schemas for all array types, and values of those types can be captured during backfill and replication.

I have not tested whether this makes array-typed **scan keys** work correctly, and I'd give even odds that it either works or breaks currently.

None to my knowledge

### DIFF
--- a/source-postgres/datatype_test.go
+++ b/source-postgres/datatype_test.go
@@ -2,10 +2,13 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/estuary/connectors/sqlcapture/tests"
 )
+
+const arraySchemaPattern = `{"required":["dimensions","elements"],"type":"object","properties":{"dimensions":{"type":"array","items":{"type":"integer"}},"elements":{"type":"array","items":%s}}}`
 
 // TestDatatypes runs the generic datatype discovery and round-tripping test on various datatypes.
 func TestDatatypes(t *testing.T) {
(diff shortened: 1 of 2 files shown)


psFried (approved):
LGTM
I think the decision to faithfully represent multidimensional arrays makes sense. I imagine that in the future we might also allow converting them into flattened arrays, based on inferred type information from the collection schema. But starting with the non-lossy conversion seems like a good default.

willdonnelly:
> we might also allow converting them into flattened arrays

I'm torn on whether this would be worth implementing. On the one hand it's kind of convenient, but on the other hand it's technically redundant functionality since you can always write a derivation that produces the same output.

Issue #93, opened 2022-01-03
source-mysql: Extend datatype support


willdonnelly:
The initial implementation of `source-mysql` has just enough datatype translation logic (`int`, `varchar`, `text`, and `double`) to make the unit tests pass.

Ideally this should be extended to all column types that MySQL supports, both for discovery (translating MySQL column type names to JSON schema) and for capture (converting the captured values so that their JSON serialization will match the JSON schema).

This may require a bit of additional plumbing in the generic `sqlcapture` package, because currently the discovered type information is not available during the translation of captured values, and the MySQL client library basically only returns `int/float64/[]byte/nil` types, so any nontrivial output translation will probably require that metadata to know when it should be applied.

willdonnelly:
So https://github.com/estuary/connectors/pull/133 takes care of all the "low-hanging fruit" when it comes to MySQL data type support. The biggest issue remaining after that PR is the handling of binary column types like `BINARY(N)` and `BLOB`.

To recap, the issue with those is that we just get a `[]byte` from the MySQL client library we're using (and I believe this is because that's all we actually get at a wire-protocol level), and we would like to emit binary data as a base64-encoded string rather than a human-readable one. This means that we need to get additional context into the value translation logic, so that it can make that sort of decision more intelligently.

I see two obvious options for where we might get this context:

- The MySQL column type, as reported by our existing discovery logic.
- The `ConfiguredCatalog` that was given to the connector at startup. More specifically, the JSON schema associated with each stream could be parsed and then decisions could be made based on factors like "does this field have `"contentEncoding":"base64"` in the currently configured schema?".

The latter option is very Clever, in both good ways and bad ways. I can see all kinds of features that _could_ be interesting. Beyond merely adding/removing `contentEncoding: base64` on any string field, there are other things like how previously with PostgreSQL array handling it was hard to decide how to represent the values, but it would actually be fairly straightforward to support multiple options with the user's schema being how they select between them.

But on the other hand that would be a nightmare to keep consistent between different connectors, so I'm not sure.

willdonnelly:
As of a couple weeks ago (https://github.com/estuary/connectors/commit/1559018e53cd73edbab9fee7d3043788f0b542b3) most column types are now supported, with the following exceptions and caveats:

* The `ENUM` and `SET` types behave inconsistently between backfills and replication. In a backfill they appear as human-readable strings, but replication events contain only an integer index/bitfield.
[...]

PR #133, opened 2022-02-16
source-mysql: Implement column types that currently work

willdonnelly:
**Description:**

This is a decent start to fixing https://github.com/estuary/connectors/issues/93

With this change, many MySQL column types are now supported in both discovery and captures. For those types which aren't supported, there are commented-out entries in the type translation table and commented-out unit test cases which exercise the problem. The remaining problematic types include:

* BIT(n) - This type acts like an integer most of the time inside MySQL, including in replicated change events. However backfill queries get back a []byte containing the raw data.
* BINARY(N), VARBINARY(N), and BLOB types - Strictly speaking these would "work", but as currently implemented the resulting JSON records would contain strings like "\u0001\u0002\u0003" and we'd probably like to base64-encode the data instead.
* ENUM and SET - These types are returned as human-readable strings in backfill queries, and as integer indices/bitfields in replicated change events. Reconciling the two is likely to be tricky, especially if there's any chance at all of schema changes.
* TIMESTAMP - The TIMESTAMP column type is apparently converted between UTC and "the current time zone" at various points, and this is done inconsistently between backfill and replication.
* JSON - Values from the DB look correct, but we need to turn them into json.RawMessage values so that they don't get quoted as strings.

I expect that BIT, BINARY/BLOB, and JSON types can all be fixed with not too much effort by providing richer "expected type" information to the value translation logic.

The ENUM/SET/TIMESTAMP issues are likely to require a bit more thought.

**Workflow steps:**

Try using the MySQL connector and it should work a bit better.


willdonnelly:
I've pushed another commit to this PR which:
1. Adds a bit of caching/plumbing so that the value translation logic in the generic `sqlcapture` machinery, so that `TranslateRecordField` can have access to the `ColumnInfo` struct generated by the table discovery code.
2. Modifies the MySQL implementation of `TranslateRecordField` to translate `BINARY(n)` and `BLOB` columns into `[]byte`  and `JSON` columns into `json.RawMessage`.

source-mysql/replication.go
```diff
@@ -199,7 +199,10 @@ func (rs *mysqlReplicationStream) run(ctx context.Context) error {
 			// that the server's `binlog_format` is not set correctly. Even when it's correctly set
 			// to `ROW`, Query events will still be sent for DDL statements, so we don't want to
 			// return an error here, but we do want these logs to be fairly visible.
-			logrus.WithField("data", data).Info("Query Event")
+			logrus.WithFields(logrus.Fields{
+				"event": data,
```
psFried, line 203: super minor quibble: when the logging format is JSON (always, unless running the connector manually), I'd think that the query would then be duplicated in the logs. Maybe just log `data`?

psFried (approved):
LGTM % a couple of nit picks on logging

source-mysql/discovery.go
```diff
@@ -68,7 +72,24 @@ func (db *mysqlDatabase) TranslateDBToJSONType(column sqlcapture.ColumnInfo) (*j
 	return colSchema.toType(), nil
 }
 
-func (db *mysqlDatabase) TranslateRecordField(val interface{}) (interface{}, error) {
+func (db *mysqlDatabase) TranslateRecordField(column *sqlcapture.ColumnInfo, val interface{}) (interface{}, error) {
+	logrus.WithFields(logrus.Fields{
+		"column": column,
+		"value":  val,
+	}).Info("translating record field")
```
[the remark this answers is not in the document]
willdonnelly: Whoops, yeah, I didn't intend to commit that at all, just forgot to remove it after rebasing a couple days ago. I don't think this is generally desirable logging even at the `Trace` level, it's just too spammy even for that.

Removed now.

source-mysql/replication.go
willdonnelly: I hadn't actually thought about how it looks as JSON, I was just making it look nicer in the human-readable output. However I think this might still be a good change, because the `Query` field of the `QueryEvent` struct is a `[]byte`, and so without this edit the JSON-formatted output looks like:

```json
{
  "data": {
    "SlaveProxyID": 275,
    "ExecutionTime": 0,
    "ErrorCode": 0,
    "StatusVars": "AAAAAAABIACgRQAAAAAGA3N0ZAQhACEA/wAS/wA=",
    "Schema": "dGVzdA==",
    "Query": "QkVHSU4=",
    "GSet": null
  },
  "level": "info",
  "msg": "Query Event",
  "time": "2022-03-04T18:19:39-06:00"
}
```

I think it's helpful to not make the user copy-paste queries into `base64 -d` in order to tell whether a given QueryEvent should be cause for concern.

willdonnelly:
Just had to tweak the value translation logic for `source-postgres` to resolve a minor conflict with the PostgreSQL array type support that just got merged.

source-mysql/replication.go
psFried: > I think it's helpful to not make the user copy-paste queries into base64 -d in order to tell whether a given QueryEvent should be cause for concern.

:+1:

PR #136, opened 2022-02-22
source-mysql: Fail cleanly when `SHOW MASTER STATUS` is empty

willdonnelly:
**Description:**

When connecting to a read replica where `log_bin = OFF` the query `SHOW MASTER STATUS` will return an empty result set (but no error), which currently panics the connector.

This setup won't actually work, so it's correct to fail here, but the failure should be clean, with an error message that suggests the solution.


psFried (approved):
LGTM

PR #138, opened 2022-02-22
source-mysql: Ignore PreviousGTIDsEvent

willdonnelly:
**Description:**

This event appears to be related to binlog segment rotation, and is used to establish an understanding of "what came before" a given file. Since we're not currently using/supporting GTID-based cursors, it is safe to ignore these events.

This change is untested (other than verifying that it doesn't break any existing test cases), because it would take nontrivial work to trigger binlog segment rotation and the fix itself is simple enough that it didn't seem worth the effort to set that up.


psFried (approved):
LGTM

PR #141, opened 2022-02-24
Various logging improvements to source-mysql

psFried:
**Description:**

The go-mysql library uses a separate logging framework, with some pretty
terrible defaults. This adds explicit configuration of that logger so
that it logs to stderr instead of stdout, and so that the log level
matches the level that's used by logrus.

Also adds a few log statements to aid in debugging.

**Workflow steps:** n/a

**Notes for reviewers:** n/a


willdonnelly (approved):
LGTM

PR #143, opened 2022-03-03
sshproxy naming change

jixij:
**Description:**
Switch the source-postgresql connector to work with the new network-connector API.
Two small fixes are made to enable the tests working again, detailed in the PR comments.

**Workflow steps:**

The SSH config in this doc(https://github.com/estuary/flow/blob/master/site/docs/concepts/connectors.md) and related needs to be updated, in which
- rename `sshUser` to `user`,
- rename `remoteHost` to `forwardHost`
- rename `remotePort` to `forwardPort`
- rename `sshPrivateKeyBase64` to `privateKey`, and the content of the field no longer need to be `based64-encoded`. It should be the plain-text of multi-line private key formatted in `yaml`. [This](https://github.com/estuary/flow/blob/master/tests/sshforwarding/materialize-postgres.ssh.config.yaml#L13) is an example of the key and content. (Note: per [yaml doc](https://yaml-multiline.info/), set the block style Indicator to be a pipe, and use indication indicator if needed. In the example it is `|2`)

**Notes for reviewers:**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -101,7 +101,7 @@ jobs:
           context: .
           file: ${{ matrix.connector }}/Dockerfile
           load: true
-          tags: ghcr.io/estuary/${{ matrix.connector }}:test
+          tags: ghcr.io/estuary/${{ matrix.connector }}:test_${{ github.sha }}
           secrets: |
             "rockset_api_key=${{ secrets.ROCKSET_API_KEY }}"
 
@@ -141,7 +141,7 @@ jobs:
           PGPORT: 5432
           PGUSER: flow
 
-        run: CONNECTOR=${{ matrix.connector }} VERSION=test ./tests/run.sh;
+        run: CONNECTOR=${{ matrix.connector }} VERSION=test_${{ github.sha }} ./tests/run.sh;
 
       - name: Source connector ${{ matrix.connector }} integration tests with SSH forwarding
         if: |
@@ -169,7 +169,7 @@ jobs:
           LOCALPORT: 9876
           SSH_FORWARDING_ENABLED: true
 
-        run: CONNECTOR=${{ matrix.connector }} VERSION=test ./tests/run.sh;
+        run: CONNECTOR=${{ matrix.connector }} VERSION=test_${{ github.sha }} ./tests/run.sh;
 
       - name: Materialization connector ${{ matrix.connector }} integration tests
         if: |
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yaml
```diff
@@ -101,7 +101,7 @@ jobs:
           context: .
           file: ${{ matrix.connector }}/Dockerfile
           load: true
-          tags: ghcr.io/estuary/${{ matrix.connector }}:test
+          tags: ghcr.io/estuary/${{ matrix.connector }}:test_${{ github.sha }}
```
jixij, line 104: The current flow runtime pulls the driver image every time it runs.
Appending the gitsha to the tag to make sure the the tag does not exist in the remote repo, so that the built local image does not get overridden by any existing images in the registry.

materialize-postgres/Dockerfile
```diff
@@ -10,8 +10,8 @@ COPY go.* ./
 RUN go mod download
 RUN go install github.com/mattn/go-sqlite3
 
-# Copy in a recent `flowctl` for usage by tests.
-COPY --from=ghcr.io/estuary/flow:dev /usr/local/bin/flowctl /usr/local/bin/flowctl
+# Download recent flow binaries for usage by tests.
+RUN curl -L --proto '=https' --tlsv1.2 -sSf "https://github.com/estuary/flow/releases/download/dev/flow-x86-linux.tar.gz"| tar -zx -C /usr/local/bin/ && cp /usr/local/bin/flowctl-go /usr/local/bin/flowctl
```
jixij, line 15: Download the flow binaries, instead of fetch from image. And seems `flowctl-go` is what we need for now to get the tests pass.

jgraettinger (commented):
A couple comments. Thanks

materialize-postgres/Dockerfile
jixij: Might need some explanations here.
Right now if without any changes, the `/user/local/bin/flowctl` is the rust `flowctl`, which is not ready to be used for testing here. So the `cp` command here is to override the rust `/user/local/bin/flowctl` with go-based  `/user/local/bin/flowctl-go`, which is needed for testing now.  So this is not about copying files around, and I think later we will remove this command after rust `flowctl` is fully developed.

jixij: I tested again with the downloaded rust `flowctl`, and it is working well. So the `cp` command is no longer needed here and everywhere.

jixij:
Thanks Johnny for the review. Changes are  made accordingly. PTAL!

jgraettinger (approved):
LGTM!

Issue #112, opened 2022-01-31
Materialization to Firebolt


dyaffe:
Firebolt is essentially a Snowflake competitor that's up and coming but with good traction.  Because they only accept delta updates, they haven't been able to create integrations with the standard ELT providers (like Fivetran) or with Kafka.  That makes them a big near-term opportunity due to the fact that their customers want to take data from their Kafka clusters and other endpoints. 

We should create a materialization connector for Firebolt that works with delta updates.  [Here](http://docs.firebolt.io/) are their docs.  I can also provide credentials to a test account which they've provisioned to us.

mdibaiee:
Studied Firebolt and our protocol to some extent and had a chat with Johnny and Phil. Here is my current high-level plan, and some notes for tackling this:

1. To add data to Firebolt tables ([fact and dimension tables](https://docs.firebolt.io/working-with-tables.html)), you need to create an External Table that is connected to a S3 bucket. The External Table basically gives you a SQL table interface to the data on S3, which can be saved in various formats (JSON documents or Parquet or ...).
2. You can then execute a query that copies data from this External Table to an actual fact or dimension table. The external table gives you two additional columns to filter on: `source_file_name` and `source_file_timestamp`.
3. Given the recent updates to our protocol, a new pattern has become available: [Recovery Log with Idempotent Apply](https://github.com/estuary/flow/blob/master/go/protocols/materialize/materialize.proto#L170-L204). This pattern matches how Firebolt works: we can store JSON documents on S3, commit them, and on the acknowledgement step copy them from the External Table to the actual table.

We initially preferred to write this connector after having completed the Rust transactor, but that can take a long time to be fully ready, whereas the Go transactor is robust and ready. As such, the decision is to write this connector in Golang using the transactor, however for parts where we need to translate our JSON schema into Firebolt table schemas, we prefer to write Rust crates and call them as binaries from the Go driver. Rust is better suited for such purposes given its Algebraic Data Types, and this will allow us to re-use this code more easily as we transition to using Rust for our connectors.

---

Some known unknowns at the moment:

1. Can we create the External Table as part of the driver or do we need to ask the user to create that table and give us a reference to it?

psFried:
> Can we create the External Table as part of the driver or do we need to ask the user to create that table and give us a reference to it? (Apparently we can)

[...]

PR #146, opened 2022-03-08
materialize-firebolt: first release

mdibaiee:
**Description:**

- New connector for materializing to Firebolt tables: The connector is mostly handling the transaction parts, whereas most of the actual validation of projections and schema, as well as queries for working with Firebolt are generated in a new Schemalate command (https://github.com/estuary/flow/pull/407)
- Firebolt client for sending queries

Solves #112

**Workflow steps:**

- Create a Flow and define a new materialization with this connector's docker image, for example see: https://github.com/mdibaiee/hello-flow/blob/main/flow.yaml#L20

None yet, I suppose we will want to write documentation for this connector as well.

**Notes for reviewers:**

I've updated `go.mod` files here to pull my latest commits from my other PR since I have some new protobuf messages that I wanted to use, see: https://github.com/estuary/flow/pull/407/commits/407ed22c6b107b9bbf4fc3cb7b4e9a9636cb3b7c


materialize-firebolt/transactor.go
```diff
@@ -0,0 +1,287 @@
...
+	if len(t.checkpoint.Files) < 1 {
+		return nil
+	}
+
+	sourceFileNames := map[int][]string{}
+	for _, item := range t.checkpoint.Files {
+		binding := item.Binding
+		fileName := item.Key
+
+		if _, ok := sourceFileNames[binding]; ok {
```
jgraettinger, line 245: How does this case run? It's not possible for there to be more than one file per binding, correct?

materialize-firebolt/transactor.go
```diff
@@ -0,0 +1,287 @@
...
+				Body:   bytes.NewReader(data),
+			},
+		})
+	}
+
+	uploadIterator := s3manager.UploadObjectsIterator{
+		Objects: uploads,
+	}
+
+	err := uploader.UploadWithIterator(aws.BackgroundContext(), &uploadIterator)
```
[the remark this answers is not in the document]
jgraettinger: One question, though: What happens if Firebolt is asked to load a file that doesn't exist? Is that a problem, or a no-op? In `Prepare`, we don't yet know whether we'll see a given binding in the current transaction and must assume we will and add a file for it to the checkpoint.

Even if firebolt were configurable to not care that a loaded file doesn't exist, we _still_ might want to surface and error on that just for sanity's sake -- to ensure the file wasn't scrubbed accidentally or something.

Thinking out loud, the Flow runtime _does_ know which bindings appear in a transaction, and could send that binding list to Prepare, which could populate checkpoint files appropriately. Then `Store` probably _shouldn't_ use the lazy initialization as I suggest above and instead initialize each stream proactively.

If this is sounding reasonable, perhaps a half-way step is to have Store create files even though they might be empty, and we'll fix it later through a coordinated update with the Flow runtime.

materialize-firebolt/transactor.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	// Acknowledge step might end up being run by a separate process
+	// different from the original, and the new transactor process might be initialised with
+	// a new version of the MaterializationSpec. This means it's possible to have an Acknowledge
+	// with MaterializationSpec v1 be run by a transactor initialised with MaterializationSpec v2.
+	// As such, we keep a copy of the queries generated from the MaterializationSpec of the
+	// transaction to which this Acknowledge belongs in the checkpoint to avoid inconsistencies.
+	for binding, fileNames := range sourceFileNames {
+		insertQuery := t.checkpoint.Queries.Bindings[binding].InsertFromTable
```
jgraettinger, line 259: Since we're stuffing queries into each checkpoint anyway, what do you think about pre-filling in these values so that the checkpoint stores completed query literals? Then Acknowledge can side-step needing to know anything about bindings.

Also, sanity checking, what happens if a specific load query is run twice? The second load is a no-op?

materialize-firebolt/transactor.go
```diff
@@ -0,0 +1,287 @@
...
+			return fmt.Errorf("moving files from external to main table: %w", err)
+		}
+	}
+
+	t.checkpoint.Files = []TemporaryFileRecord{}
+
+	return nil
+}
+
+func randomDocumentKey() string {
```
jgraettinger, line 272: nit: use `github.com/google/uuid.NewRandom()` ?

jgraettinger (approved):
Thanks, this is looking good. A bunch of feedback below, mostly nits, some questions, and suggestions on ways to improve structuring.

If you want to land this and address some or all comments as subsequent PRs, that works for me.

jgraettinger:
Looked into the CI break and it's unrelated to this PR and is rather another connector that needs to be updated to reflect an upstream protocols change.

materialize-firebolt/config.go
```diff
@@ -0,0 +1,91 @@
...
+		return fmt.Errorf("missing required password")
+	}
+	if c.S3Bucket == "" {
+		return fmt.Errorf("missing required bucket")
+	}
+	return nil
+}
+
+// GetFieldDocString implements the jsonschema.customSchemaGetFieldDocString interface.
+func (config) GetFieldDocString(fieldName string) string {
```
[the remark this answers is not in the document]
mdibaiee: @jgraettinger Okay, from the code of it, seems like GetFieldDocString does replace description, but not title:

https://github.com/invopop/jsonschema/blob/297bb072302c4423669838e04bb7705277e50224/reflect.go#L531-L536

So I can just remove this method or remove the description tags from the struct. I think I actually prefer this method for description, and just use the tags for title.

materialize-firebolt/transactor.go
mdibaiee: Ah yes, my initial implementation saved multiple files for each binding, I forgot to update this 👍🏽

materialize-firebolt/transactor.go
mdibaiee: Running a query twice loads the data twice (creates duplicate rows), is that a case we need to handle to make sure it's no-op?

materialize-firebolt/transactor.go
mdibaiee: That function has a possibility for error since it uses io operations, this one seems more straightforward without an error

materialize-firebolt/transactor.go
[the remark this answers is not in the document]
mdibaiee: Loading a file into main table is done using a query like this:

```
INSERT INTO files SELECT * FROM files_external WHERE source_file_name IN ('file_name.json');
```

If the file doesn't exist, this query is a no-op because the `SELECT ... WHERE` will have no results.

materialize-firebolt/transactor.go
jgraettinger: Yes, definitely It's expected that we'll regularly see Acknowledge run twice for a given commit: a prior connector completes a transaction, acknowledges it, and then exits. A subsequent connector then re-acknowledges that same transaction as its first action.

If the second acknowledge loads data twice, that's a violation of exactly-once semantics.

Worse, it's possible that an Acknowledge can _partially_ complete (only _some_ of the queries are issued), then perhaps crash while awaiting one of these running INSERT queries. A subsequent invocation then tries the whole Acknowledge again.

The absolute ideal would be to have each individual query be idempotent and atomic -- we can send a single query to firebolt, then crash without getting a response back, and it will either a) complete such that sending the query again is a no-op, or b) fail with no side-effects.

Not sure if or what is achievable w/in Firebolt though. Is there a `CASE WHEN` equivalent ?

materialize-firebolt/transactor.go
jgraettinger: Mostly just reduction of SLOC and removing any possibility of a bug. And even if _we_ know there is no bug, it saves the reader time from wondering / verifying that as well. Being _very_ pedantic, I'd also be thinking about "hey, could these keys _ever_ collide?", and it's true that `rand` is not a cryptographically strong random number generator, while v4 UUID's are (drawing from `/dev/urandom`).

It technically returns an error (as `/dev/urandom` is file-like) but in your shoes I'd simply panic if one actually happens, as that's an infallible operation without something being seriously wrong.

jgraettinger: (This is a suggestion, not a requirement. You asked so I'm just unpacking the internal rationale that led to the suggestion).

materialize-firebolt/transactor.go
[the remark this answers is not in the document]
mdibaiee: I will look into this in a separate PR

materialize-firebolt/transactor.go
mdibaiee: I'll look into this to see how can we make this an idempotent operation 👍🏽

materialize-firebolt/transactor.go
mdibaiee: Makes sense 👍🏽

jixij (approved):

materialize-firebolt/transactor.go
mdibaiee: This should make the insertion idempotent: https://github.com/estuary/flow/pull/407/commits/cd715d1be1b9f99608d3ec8d8f93bd29f5cf9efd

mdibaiee: I just tried it by running this code (basically running each query twice)
```
	for _, query := range t.checkpoint.Queries {
		_, err := t.fb.Query(query)
		if err != nil {
			return fmt.Errorf("moving files from external to main table: %w", err)
		}
		_, err = t.fb.Query(query)
		if err != nil {
			return fmt.Errorf("moving files from external to main table: %w", err)
		}
	}
```

No duplicates are inserted 👍🏽

mdibaiee:
Created issue to fix the materialize-s3-parquet build: https://github.com/estuary/connectors/issues/156

Issue #147, opened 2022-03-09
Azure Postgres Connector Failing due to cron auth


dyaffe:
Our standard steps for setting up a Postgres database don't exactly work.  The first issue happens when one attempts to run the following:

```
postgres=> GRANT SELECT ON ALL TABLES IN SCHEMA information_schema, pg_catalog TO flow_capture;
WARNING: no privileges were granted for "sql_implementation_info"
ERROR: permission denied for table sql_parts
```

It next appears when trying to run a catalog set up to use this same database:

```
flowctl discover --image [ghcr.io/estuary/source-postgres:dev](http://ghcr.io/estuary/source-postgres:dev)
ERRO[0001] running connector failed                      error="exit status 1" logSource="[ghcr.io/estuary/source-postgres:dev](http://ghcr.io/estuary/source-postgres:dev)" operation=discover
FATA[0001] fatal error                                   err="fetching connector bindings: exit status 1 with stderr:\n\n{\"timestamp\":\"2022-03-08T22:47:29.876372355Z\",\"level\":\"INFO\",\"message\":\"received server public key: Ed25519(PublicKey { key: [52, 32, 85, 188, 70, 183, 144, 148, 34, 136, 135, 150, 154, 183, 205, 138, 126, 217, 47, 174, 194, 98, 72, 101, 233, 59, 251, 190, 154, 23, 231, 69] })\"}\ntime=\"2022-03-08T22:47:29Z\" level=info msg=\"initializing connector\" database=postgres host=localhost port=6000 slot=flow_slot user=flow_capture\n{\"timestamp\":\"2022-03-08T22:47:30.40290684Z\",\"level\":\"INFO\",\"message\":\"Unhandled global request: Ok(\\\"[hostkeys-00@openssh.com](mailto:hostkeys-00@openssh.com)\\\")\",\"log.target\":\"thrussh::client::encrypted\",\"log.module_path\":\"thrussh::client::encrypted\",\"log.file\":\"/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/thrussh-0.33.5/src/client/encrypted.rs\",\"log.line\":481}\nunable to list database columns: ERROR: permission denied for schema cron (SQLSTATE 42501)\nError:  unable to list database columns: ERROR: permission denied for schema cron (SQLSTATE 42501)\n"
```

The unable to list all database columns error appears to happen due to our flow_capture user not being able to grant usage on schema cron to `flow_capture`:

```
[...]

PR #148, opened 2022-03-09
source-postgres: Factor out column descriptions from discovery

willdonnelly:
**Description:**

The portion of the column-discovery query that reads column descriptions can fail in some cases, such as when the capture user doesn't have the `USAGE` permission on some schema. In many cases we didn't even want to capture these tables, so it makes little sense for this to be a fatal error for the entire discovery process.

This commit removes the column description from the main column discovery query, and then after the main query completes runs a loop over every discovered column individually requesting the description. Errors are ignored during this step, because the worst that should happen is that our generated JSON schema will not have useful descriptions.

Outside of the specific edge case of schemas which we don't have the `USAGE` permission on this should be a no-op, and the description discovery feature is exercised by `TestDiscoveryComplex` which shows no changes.

This hopefully fixes https://github.com/estuary/connectors/issues/147

**Workflow steps:**

Run `flowctl discover` on a PostgreSQL database where the capture user doesn't have `USAGE` permission on all schemas. Prior to this change that would result in the entire process failing, after this change it should work out fine.

### DIFF
--- a/source-postgres/discovery.go
+++ b/source-postgres/discovery.go
@@ -261,11 +261,7 @@ const queryDiscoverColumns = `
 		c.ordinal_position,
 		c.column_name,
 		c.is_nullable::boolean,
-		c.udt_name,
-		pg_catalog.col_description(
-			format('%s.%s',c.table_schema,c.table_name)::regclass::oid,
-			c.ordinal_position
-		) AS column_description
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM

PR #149, opened 2022-03-09
source-postgres and source-mysql: Validate watermarksTable value

willdonnelly:
**Description:**

The config parameter 'watermarksTable' must be fully-qualified in `'<schema>.<table>'` form in order for the watermark-change detection logic to work properly.

This change improves things slightly by mentioning that fact in the description (and thus in the placeholder config.yaml generated by `flowctl discover`) and validating that (if it's set at all) the watermarksTable property must contain a period. It's not perfect, but it should blunt that rough edge a bit for now.

This would probably have helped with https://github.com/estuary/flow/issues/398

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/alecthomas/jsonschema"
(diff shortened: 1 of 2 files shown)


psFried (approved):
LGTM

jixij (approved):
cool

PR #158, opened 2022-03-28
materialize-firebolt: store JSON as marshalled string, allow nested keys

mdibaiee:
**Description:**

- ~Allows for document keys with `/` to be stored by replacing the `/` with `_` to be compatible with Firebolt~
- In case of JSON objects, marshals them into a string and then stores them as a string to compensate for Firebolt's lack of support for actual JSON objects

**Workflow steps:**

See the other PR https://github.com/estuary/flow/pull/423

N/A

**Notes for reviewers:**

See the other PR https://github.com/estuary/flow/pull/423

### DIFF
--- a/materialize-firebolt/transactor.go
+++ b/materialize-firebolt/transactor.go
@@ -163,13 +163,13 @@ func (t *transactor) projectDocument(spec *pf.MaterializationSpec_Binding, keys 
 
 	// Add the keys to the document.
 	for i, value := range keys {
-		var propName = strings.ReplaceAll(spec.FieldSelection.Keys[i], "/", "_")
+		var propName = spec.FieldSelection.Keys[i]
 		document[propName] = value
 	}
 
 	// Add the non-keys to the document.
 	for i, value := range values {
-		var propName = strings.ReplaceAll(spec.FieldSelection.Values[i], "/", "_")
+		var propName = spec.FieldSelection.Values[i]
 
 		if raw, ok := value.([]byte); ok {
 			var nestedObject = make(map[string]interface{})
--- a/materialize-firebolt/transactor.go
+++ b/materialize-firebolt/transactor.go
@@ -175,16 +175,10 @@ func (t *transactor) projectDocument(spec *pf.MaterializationSpec_Binding, keys 
 			var nestedObject = make(map[string]interface{})
 			err := json.Unmarshal(raw, &nestedObject)
 
-			// If we can parse this raw json as an object, we marshal it
-			// so it can be stored as a stringified JSON object, since
-			// Firebolt does not support JSON objects.
+			// If we can parse this raw json as an object, we store it as a
+			// stringified JSON object, since Firebolt does not support JSON objects.
 			if err == nil {
-				obj, err := json.Marshal(nestedObject)
-				if err != nil {
-					return nil, fmt.Errorf("failed to serialize the nested object: %w", err)
-				}
-
-				document[propName] = string(obj)
+				document[propName] = string(raw)
 			} else {
 				document[propName] = json.RawMessage(raw)
 			}
--- a/materialize-firebolt/transactor.go
+++ b/materialize-firebolt/transactor.go
@@ -163,16 +163,31 @@ func (t *transactor) projectDocument(spec *pf.MaterializationSpec_Binding, keys 
 
 	// Add the keys to the document.
 	for i, value := range keys {
-		var propName = spec.FieldSelection.Keys[i]
+		var propName = strings.ReplaceAll(spec.FieldSelection.Keys[i], "/", "_")
 		document[propName] = value
 	}
 
 	// Add the non-keys to the document.
 	for i, value := range values {
-		var propName = spec.FieldSelection.Values[i]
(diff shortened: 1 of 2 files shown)


materialize-firebolt/transactor.go
```diff
@@ -163,16 +163,31 @@ func (t *transactor) projectDocument(spec *pf.MaterializationSpec_Binding, keys
 
 	// Add the keys to the document.
 	for i, value := range keys {
-		var propName = spec.FieldSelection.Keys[i]
+		var propName = strings.ReplaceAll(spec.FieldSelection.Keys[i], "/", "_")
```
psFried, line 166: There might be a better way of approaching this. Some background:

The `FieldSelection` represents the set of projections that should be materialized. During `flowctl api build`, assuming the user has left the default [`fields`](https://github.com/estuary/flow/blob/master/crates/models/src/materializations.rs#L49) in their binding, we'll call the `Validate` RPC of the connector and resolve this `FieldSelection` by just taking all the fields that the connector tells us are "recommended".

So I think a better way of handling this would be to validate the field names as part of the `Validate` RPC and return `FIELD_FORBIDDEN` for any that contain a `/`. If any fields would need to be renamed, then the expectation would be that the user provides an explicit projection and uses the `fields` configuration in the materialization binding to wire it up. It's more work for a CLI user, but it allows the `projections` of a collection to remain authoritative on the mappings between tabular and JSON representations, and can be guided by good `reason` strings in the constraints returned by `Validate`. It also avoids the possibility of collisions in the case where there's both `foo/bar` and `foo_bar` fields.

materialize-firebolt/transactor.go
```diff
@@ -163,16 +163,31 @@ func (t *transactor) projectDocument(spec *pf.MaterializationSpec_Binding, keys
...
 	}
 
 	// Add the non-keys to the document.
 	for i, value := range values {
-		var propName = spec.FieldSelection.Values[i]
+		var propName = strings.ReplaceAll(spec.FieldSelection.Values[i], "/", "_")
 
 		if raw, ok := value.([]byte); ok {
-			document[propName] = json.RawMessage(raw)
+			var nestedObject = make(map[string]interface{})
```
psFried, line 175: I think the only values that get encoded as `[]byte` in a `Tuple` are JSON objects and arrays. If that's true, then I think you could just use `document[propName] = string(raw)` and not have to parse anything. The tuple-encoding impl is [here](https://github.com/estuary/flow/blob/master/crates/tuple/src/value.rs#L24), and it appears to be the case that only objects and arrays will end up as byte slices.

psFried (changes requested):
I think we should consider a different approach on this one. Happy to talk through any of those suggestions, though.

materialize-firebolt/transactor.go
mdibaiee: @psFried Firebolt _does_ support arrays, so for arrays we want to keep them as they are.

materialize-firebolt/transactor.go
mdibaiee: @psFried I see, okay, I thought since the default recommended fields seem to contain `/`, it would be simple if we just allow the recommendations, only with a change in the field name. This means our recommended fields will not be valid for any nested objects and any nested objects will have to be manually projected, right? Is there a better way where we can still allow the default recommendations?

mdibaiee: Authoritativeness of the projections and avoiding implicit projections makes sense though, I see the point.

mdibaiee: I'm looking to see if there is a way in Firebolt to allow column names with `/` in them (perhaps by some form of quotation or escaping)

mdibaiee: @psFried added validation logic for column names and no longer manipulating them in the connector / schemalate: https://github.com/estuary/flow/pull/423/commits/e24b9685267c021792b8444fb3da39d949ed05d9

mdibaiee: The result is:
```
constraints:<key:\"obj/name\" value:<type:FIELD_FORBIDDEN reason:\"Field names must match the regex ^[a-zA-Z_][a-zA-Z0-9_]*$.\" >
```

Which leads to `obj/name` being omitted from the document and the table.

materialize-firebolt/transactor.go
psFried: ah, ok, that makes sense.

psFried (approved):
Looks good. Thanks for the replies.

materialize-firebolt/transactor.go
mdibaiee: Ah you are right!

materialize-firebolt/transactor.go
jgraettinger: 👍 to resolution here.

>  I get a 500 Internal Error from Firebolt if I try to create a column with / in it, e.g.: CREATE FACT TABLE ("obj/name" TEXT) PRIMARY INDEX "obj/name" fails with 500.

This smells like something they ought to be able to fix -- that in the future "obj/name" would be fine with proper quoting.

PR #164, opened 2022-04-07
Phil/postgres datetime

psFried:
**Description:**

- Migrates the `PostgresSQLGenerator` from the Flow repo into the connectors repo.
- Adds support for materializing `date-time` string values as `TIMESTAMPTZ` columns.

**WARNING:** Once this is merged, the `:dev` tag will start to use this new behavior, which would break any existing materializations that are using `TEXT` for timestamps. To my knowledge, there are no existing materializations that will be affected, so it may not be an issue at all. If there are any affected materializations, then I _think_ that the best way to handle this is to pin their images to a specific tag. As of this writing, this would be `ghcr.io/estuary/materialize-postgres:16bb090`, which is the most recent build. We'd need to switch over to that image _before_ this PR is merged.

**Workflow steps:**

No change to workflows, but now if you have a JSON schema that specifies `{"type": "string", "format": "date-time"}`, those fields will end up as `TIMESTAMPTZ` columns instead of `TEXT`.

None, since we haven't yet documented materialization type mapping.

**Notes for reviewers:**

This is intended to be pretty minimal, and to only address the immediate need. In particular, it doesn't address the issues of how we should be testing these, or allowing for customization.

### DIFF
--- a/materialize-postgres/driver.go
+++ b/materialize-postgres/driver.go
@@ -103,7 +103,7 @@ func newPostgresDriver() pm.DriverServer {
 			if err != nil {
 				return nil, fmt.Errorf("opening Postgres database: %w", err)
 			}
-			return sqlDriver.NewStdEndpoint(parsed, db, sqlDriver.PostgresSQLGenerator(), sqlDriver.DefaultFlowTables("")), nil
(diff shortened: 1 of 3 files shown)


mdibaiee (approved):
LGTM 👍🏽

PR #171, opened 2022-04-13
materialize-snowflake: add reserved words

mdibaiee:
**Description:**

- Specifies the list of reserved words for Snowflake to be quoted. The list was obtained from here: https://docs.snowflake.com/en/sql-reference/reserved-keywords.html

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

N/A


jgraettinger (approved):
LGTM

PR #178, opened 2022-04-19
Refactor Big Query connector with transactor and change authoritative checkpoint

pothibo:
**Description:**

Most of the code changes here is to adapt to the new Transactor interface built for connector, this means replacing the previous sqlDriver interface implementation to follow the same pattern as the one used for [firebolt](https://github.com/estuary/connectors/blob/main/materialize-firebolt/driver.go)

The connector goes through 2 refactoring phase for this changes to
happen. First, it transitions away from the sqlDriver interface that is
more suited for connectors like PostgreSQL and other RDMS, then instead
of setting a fence on the bigQuery dataset, the connector uses a similar
strategy to the one used in Firebolt, which means during the Prepare
phase, the driver check point stores the external files and the queries
that will run for those cloud storage file.

These files gets written to during the Commit phase, and the SQL runs a
merge operation (or append if delta updates) using those external
file. Those external files are basically, in BigQuery's environment, a
different database that can be referenced like any other DB. This
allows, using ExternalDataConfig, to make join SQL query.

**Workflow steps:**
Same as before, it should not include any breaking changes. One thing I'll admit I haven't done yet, is to get a much more complicated collection running with this connector. I've basically only used word_count example and ran it in both previous and current mode and compared the behavior.

Any inputs/ideas are welcomed.

I think the document doesn't need any changes for this connector as the configuration options remains the same and behave the same. The only thing that could possibly change is if the documentation mentions anything about the role of the connector with how authoritative it is.

**Notes for reviewers:**
It's a massive change. I apologize in advance. One thing to note here is that I was heavily influenced by the work done on the  Firebolt connector. I tried to merge the approach there with also some logic that I needed to carry over from the old implementation.

I also removed the tests that was there, mostly because the Binding constructor now requires objects that would need to be mocked and so. However, I didn't look too much into it, if possible, I would like to reintroduce them.


materialize-bigquery/bigquery.go
```diff
@@ -1,214 +1,87 @@
...
+	case "integer":
+		return bigquery.IntegerFieldType
+	case "number":
+		return bigquery.BigNumericFieldType
+	case "timestamp":
+		return bigquery.TimestampFieldType
+	default:
+		return bigquery.BytesFieldType
 	}
 }
```
pothibo, line 87: I'm not sure how useful/stable this method is, I think that is one area where I'm kinda flying blind. Would there be an example that I could use to validate must types?

materialize-bigquery/Dockerfile
```diff
@@ -18,13 +18,15 @@ COPY materialize-bigquery    ./materialize-bigquery
...
 # Test and build the connector.
-RUN go test  -tags nozstd -v ./materialize-bigquery/...
+# RUN go test  -tags nozstd -v ./materialize-bigquery/...
 RUN go build -tags nozstd -v -o ./connector ./materialize-bigquery/...
 
 # Runtime Stage
 ################################################################################
 FROM debian:bullseye-slim
 
+LABEL FLOW_RUNTIME_PROTOCOL="Materialize"
```
pothibo, line 28: I believe this is required moving forward? If not, I can remove.

materialize-bigquery/bigquery.go
```diff
@@ -1,214 +1,87 @@
...
+	fields := binding.FieldSelection.AllFields()
+	schema := make(bigquery.Schema, len(fields))
+
+	for idx, fieldName := range fields {
+		var (
+			err   error
+			field *bigquery.FieldSchema
+		)
+
+		if field, err = fieldSchemaForNameAndProjections(fieldName, binding.Collection.Projections); err == errPrivateProjection {
```
mdibaiee, line 32: Maybe I'm missing something but I couldn't find any place where this error is emitted

materialize-bigquery/binding.go
```diff
@@ -1,225 +1,369 @@
...
+		externalTableAlias: externalTableAlias,
 	}
 
-	return nil
+	name, err := randomString()
+	if err != nil {
+		return nil, err
+	}
 
+	b.Reset(ctx, fmt.Sprintf("%s/%s", cfg.BucketPath, name))
```
mdibaiee, line 109: Just to clarify: are we basically initialising a new writer on an object with a random name here? This is the writer used to write the file corresponding to this binding yeah?

It may be a nitpick but can we somehow express the purpose of this function with a different name? I see that it's called whenever we want a new writer. So perhaps just `binding.NewStorageWriter` or something along those lines

materialize-bigquery/binding.go
```diff
@@ -1,225 +1,369 @@
...
+func (b *Binding) Commit(ctx context.Context) error {
+	return b.writer.commit(ctx)
+}
+
+// Reset is called when a new Writer need to be instantiated.
+// This is done so every pipeline operation in the bigquery materialization
+// can work with it's own writer to avoid data integrity issues.
+// Call Reset when the lifecycle of a bigquery materialization is completed and
+// the data is committed and acknowledged.
+func (b *Binding) Reset(ctx context.Context, name string) error {
```
mdibaiee, line 164: I'm probably missing parts of the picture here, but is it possible to initialise a new writer during the Store phase? Or put another way, my question is, do we really need to create a writer ahead of our Store phase and keep it in the binding? Can it be a short-lived value inside Store?

materialize-bigquery/transactor.go
```diff
@@ -2,287 +2,192 @@ package main
...
+			return err
+		}
 	}
 
 	return nil
 }
 
-func (t *transactor) Acknowledge(context.Context) error {
+func (t *transactor) Acknowledge(ctx context.Context) error {
+	for i, cpBinding := range t.checkpoint.Bindings {
```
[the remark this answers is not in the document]
mdibaiee: I think the idea behind the protocol is that since Commit happens before Acknowledge, during the Commit phase you must make sure that the data is committed properly, and if so,  you do not need to run checks here about whether the data is committed or not. Acknowledge can assume that the files must have been written and comitted successfuly.

mdibaiee (commented):
A few comments below, happy to chat about any of them

materialize-bigquery/bigquery.go
mdibaiee: I think it may be better to panic here if we see a type that we do not recognise and can't translate to BigQuery directly. See the approach here:

https://github.com/estuary/flow/blob/master/crates/schemalate/src/firebolt/firebolt_schema_builder.rs#L160-L179
https://github.com/estuary/flow/blob/master/crates/schemalate/src/firebolt/firebolt_schema_builder.rs#L77-L80

materialize-bigquery/bigquery.go
mdibaiee: I mean `errPrivateProjection`. `fieldSchemaForNameAndProjections` doesn't seem to emit such an error ever

materialize-bigquery/transactor.go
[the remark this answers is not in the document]
pothibo: I think most of the questions you have revolves around the writer{} struct on how it's used. So, I'll try to wrap my answer here to includes all of your other comments. First, my understanding is that the Acknowledge phase can come from 1 of two places:

- After the the stream is opened and `RunTransactor` is called, which would happen when a restart happened between Commit and Acknowledged, and so, the bindings needs to be properly configured to execute a query that includes the cloud storage object and the query.
- After a Commit phase, which is the normal operation flow. In this case, the data is being set up in the Prepare phase, Committed to cloud storage in the Commit phase, and then, everything is packaged and ready to be processed in the Acknowledge phase

So, if that first scenario is indeed possible, it means that the Acknowledged phase is going to attempt to create a query for a file path in Google Storage that we're not 100% sure at this point exists. This could lead to a scenario where the connector tries to execute a query for a file that does not exists and fails in a loop forever since we never go pass that driver checkpoint (loads the connector -> acknowledge failed -> connector shuts down -> loads the connector, etc.)

The reason why I put the `writer{}` inside the Binding object is that, as far as I understand, we either need to store the Google cloud object handle or need to reinitialize it on every phase that requires it. In the current way I got it setup, it would mean that we'd setup those objects in Store/Commit/Acknowledge at the very least.

By having a "mutable" sub-object in `writer` it allowed me to carry some state over the pipeline. Honestly, if it causes issues, I could just rework the logic so that we're just creating a new `writer{}` and use it as a standalone.

materialize-bigquery/binding.go
```diff
@@ -1,225 +1,369 @@
...
-		b.load.extDataConfig.Schema = append(b.load.extDataConfig.Schema, &bigquery.FieldSchema{
-			Name:     identifierSanitizer(col.Name), // Sanitized name without quoting required
-			Repeated: false,
-			Required: col.NotNull,
-			Type:     bigquery.FieldType(colType.SQLType),
-		})
+// Whether the underlying writter has data committed
+// to the storage.
+func (b *Binding) DataCommitted() bool {
+	return b.writer.dataCommitted && b.writer.dataWritten > 0
```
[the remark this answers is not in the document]
pothibo: This is related to the big comment I left above, it's sort of a protection to skip any binding that doesn't have file to it. Maybe the name of the method/fields should be changed? I'll wait for us to discuss the above comment, because if my assumption around the transactor is proven wrong, this might not be necessary.

materialize-bigquery/binding.go
pothibo: I think that makes sense, I'll make the change.

materialize-bigquery/binding.go
pothibo: Not currently, a new writer is only generated at the Prepare phase, or when the transactor is initialized. Similar to the other comment in this PR, this might be refactored if my understanding of "failure mode" is proven wrong.

pothibo:
@jgraettinger @psFried I just had a quick conversation with Mahdi about this PR and one thing we wanted to ask one of you is the semantic value of uploading the content of the stage data in the Commit phase.

This is a place where Firebolt & this connector diverges. In this implementation of the BigQuery connector, I split staging the data with saving it to cloud storage, which means I need to hold variables in memory between phases (Store and Commit). This creates a scenario where I have to store, for each binding, a `writer{}` struct that holds everything between those phases.

Semantically, I like that this connector is only writing to  cloud storage when the Commit phase is reached, but in practice, it creates a trade-off.

The `writer{}` logic needs to be elevated into the binding in this current implementation, and there's a mutable state in each of the binding that might feel weak.

If we were to store the data in cloud storage during the Store phase, that `writer{}` struct could only be a local variable to the Store Phase.

Is there a direction that you 2 would prefer?

psFried (changes requested):
This is clearly a pretty big project, and it's great to see all the progress. Thanks for sticking with this, even though it's really jumping into the deep end. I think there's still a few things to clear up, and some work to do on both performance and correctness testing. Those are just good next steps, though, rather than a criticism of the current state of things.

> If we were to store the data in cloud storage during the Store phase, that writer{} struct could only be a local variable to the Store Phase.
> 
> Is there a direction that you 2 would prefer?

I don't have a strong opinion on that, but I don't think there's anything wrong with maintaining state between store and commit phases. 

Some related considerations: For BQ, each transaction seems to have a relatively high overhead, which causes us to use longer transaction durations within Flow so that we can get more documents in each transaction. This may change with this refactoring, and I think a good next step for this PR would be some high-volume testing so that we can get a sense of what the ideal range of data volume per transaction is. Based on the results of that, I think there's a few things we could consider:

- Using the [Storage Write API](https://cloud.google.com/bigquery/docs/write-api) instead of writing files to cloud storage.
- Chunking stores into pieces of a multipart upload

I'm skeptical of staging documents into a local file before uploading to cloud storage, because Flow should be able to feed in all the Store requests pretty quickly since it literally just has them all in memory.

materialize-bigquery/bigquery.go
pothibo: It is on line 36 because it only `continue` if the err is equal to `errPrivatePRojection` but since the both of you tripped on that line, I'll change it to make it clearer (also you said in another comment that this might not even be needed, so maybe I'll just remove it.

pothibo:
I wanted to get the Storage API Writer included in this PR, but I realized on Friday that the API is in EXPERIMENTAL status and as such, it is probably not wise to use it right this minute.

So, I'll start testing today the current implementation, and gather some information. If the metrics are acceptable, we'll be able to launch this version now, and improve it once the Storage API Writer is released.

materialize-bigquery/schema.go
```diff
@@ -0,0 +1,191 @@
...
+		case "string":
+			if projection.Inference.String_ != nil {
+				s := projection.Inference.String_
+				switch s.Format {
+				case "date-time":
+					fieldType = bigquery.DateTimeFieldType
+				case "date":
+					fieldType = bigquery.DateFieldType
+				case "geography-wkt":
+					fieldType = bigquery.GeographyFieldType
```
pothibo, line 151: I added this field that didn't exist prior. I just saw it in the BigQuery documentation and thought it could be useful to include it. I can definitively remove it if we don't feel like it's worthwhile to keep it.

materialize-bigquery/binding.go
```diff
@@ -1,225 +1,319 @@
...
+			br.Table,
 		)
+
 	} else {
-		// Updates for all columns minus the primary keys plus the document field.
-		var lrUpdates []string
-		for _, colName := range append(spec.FieldSelection.Values, spec.FieldSelection.Document) {
-			var col = tableDef.GetColumn(colName)
-			lrUpdates = append(lrUpdates, fmt.Sprintf("l.%s = r.%s", col.Identifier, col.Identifier))
+		loadQuery = fmt.Sprintf(`
```
psFried, line 267: One thing to look into here is quoting of column and table names. I think the old sql driver had some functionality for quoting identifiers that needed it, depending on which database we're talking to, and we probably want the same behavior here, just with [bigquery-specific rules](https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical). As written, I think this would fail if a column name happened to be a reserved word or start with a number.

materialize-bigquery/binding.go
```diff
@@ -1,225 +1,319 @@
...
+			strings.Join(rColumns, ", "),
 		)
 	}
+	return &sqlQueries{
+		LoadSQL:  loadQuery,
+		WriteSQL: writeQuery,
+	}
+}
+
+// Bigquery only allows underscore, letters, numbers, and sometimes hyphens for identifiers. Convert everything else to underscore.
```
psFried, line 314: `identifierSanitizer` seems to no longer be used, unless I'm missing it? But also, I think it's probably better to quote such naught identifiers rather than transforming them, anyway. Otherwise, it's easy to end up with duplicate column names.

psFried (changes requested):
I think this is looking _much_ better! I left a couple of minor comments inline, but I think the main thing with this PR is just going to be testing it at scale and see how it goes.

materialize-bigquery/binding.go
pothibo: Yes, I meant to use it, and forgot to add it back.

materialize-bigquery/schema.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	if fieldType == "" {
+		return nil, fmt.Errorf("could not map the field to a big query type: %s", fieldName)
+	}
+
+	// This is a test to make sure that the only value in the infered type is either 1
+	// or 2 but it includes a nullable field. If it has more than the expected count, it returns an
+	// error back, ie. Inference.Types = [string, integer] is wrong. [string, null] is OK. [string, integer, null] is wrong
+	expectedSize := 1
```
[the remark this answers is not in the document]
pothibo: I made the change, since it will most likely be an array of 0-3 items so the cost of running through the array is negligible, but doing it this way means we're looping over the slice twice. That's why I didn't use it in the first place.

pothibo:
Closing this PR and reopening a new one to start the conversation for final review.

PR #182, opened 2022-04-21
filesource: use stderr logging instead of Airbyte log messages

willdonnelly:
**Description:**

This changes the `filesource` package to emit its informational / progress-reporting messages to stderr rather than as Airbyte log messages on stdout.

The package previously used a mix of the two, with the most common informational/progress messages being output as Airbyte
messages. Unfortunately the plumbing for Airbyte log messages seems to have been lost in all the connector-interface changes lately, and in general it seems like stderr is preferable for now, so let's just get this working again the easy way.

**Workflow steps:**

Log messages like `sweeping` and `processing file` will start showing up in the logs collection again.

### DIFF
--- a/filesource/filesource.go
+++ b/filesource/filesource.go
@@ -311,8 +311,12 @@ type reader struct {
 }
 
 func (r *reader) sweep(ctx context.Context) error {
-	r.log("sweeping %s starting at %q, from %s through %s",
-		r.prefix, r.state.Path, r.state.MinBound, r.state.MaxBound)
+	log.WithFields(log.Fields{
+		"prefix":   r.prefix,
+		"path":     r.state.Path,
+		"minBound": r.state.MinBound,
+		"maxBound": r.state.MaxBound,
+	}).Info("sweeping")
 
 	var listing, err = r.store.List(ctx, Query{
 		Prefix:    r.prefix,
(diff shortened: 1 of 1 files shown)


mdibaiee:
@willdonnelly Just seeing this, by "the plumbing for Airbyte log messages seems to have been lost in all the connector-interface" do you mean you do not see the Airbyte log messages properly outputted anymore? We have code that is _supposed_ to handle this but there might be a bug.

mdibaiee:
Okay I did some reading about this, I think the problem lies in the fact that the connector proxy runner does not pass any logging parameters to connector_proxy, so connector_proxy just assumes `log.level=warn`. I see in your case, you have some Info logs that are not showing up, so that explains it. I'll see how we can have the `log.level` passed through all the way to connector_proxy.

mdibaiee:
This should fix the root issue of Airbyte log messages not being shown: The log.level of connector proxy is not corresponding to the one passed to the CLI, and I tested it and can confirm the log level is now working as expected. Can you please try with that version?

willdonnelly:
Ah, I see. By "the plumbing seems to have been lost", I just meant that after the refactoring I couldn't find any logic governing the promotion of `AirbyteLogMessage`s to somewhere it would actually show up (which used to be https://github.com/estuary/flow/blob/b8782a40af086fa27db7f5c88cd6acfcbd2c4cfc/go/capture/driver/airbyte/driver.go#L106 before all the refactoring).

Looking at it a bit more it looks like the relevant logic is https://github.com/estuary/flow/blob/master/crates/connector_proxy/src/libs/stream.rs#L64 and then https://github.com/estuary/flow/blob/master/crates/connector_proxy/src/libs/airbyte_catalog.rs#L164, so I think something like the PR you linked should fix the problem.

PR #185, opened 2022-04-25
materialize-postgres: do not convert timestamp values, insert as is

mdibaiee:
**Description:**

- PostgreSQL can handle ISO8601 datetimes itself, so we can instead of parsing the datetime ourselves, just pass it in directly with type: `TIMESTAMPZ` and let postgres handle the conversion.

Solves #184 

**Workflow steps:**

- Use `format: date-time` on a date-time value formatted as ISO8601

N/A

**Notes for reviewers:**

### DIFF
--- a/materialize-postgres/sqlgen.go
+++ b/materialize-postgres/sqlgen.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/estuary/flow/go/protocols/materialize/sql"
 )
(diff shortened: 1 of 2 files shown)


materialize-postgres/sqlgen.go
```diff
@@ -22,20 +21,16 @@ func PostgresSQLGenerator() sql.Generator {
...
 				ByFormat: map[string]sql.TypeMapper{
-					// This format is for RFC3339 timestamps. As of this writing, Flow's schema
-					// validation does not actually validate that String values match their declared
-					// format, so there's no guarantee that the values will parse successfully here.
+					// According to the JSONSchema spec, this format is for RFC3339 timestamps, however
+					// since it seems a lot of usages of this format point to ISO8601 datetimes
+					// we have decided to support ISO8601 for this format as well.
+					// As of this writing, Flow's schema validation does not actually validate that
+					// String values match their declared format, so there's no guarantee that the values will parse successfully here.
 					"date-time": sql.ConstColumnType{
```
psFried, line 29: This could be simplified to `"data-time": sql.RawConstColumnType("TIMESTAMPTZ")`, which always uses the identity function for the ValueConverter.

psFried (approved):
This seems like it should work, but I feel like there might have actually been a problem with the pgx driver using strings instead of a parsed timestamp. Have you tested materializing timestamps and verified that it works ok? As long as it works, I'm good with this.

materialize-postgres/sqlgen.go
mdibaiee: @psFried I tried this on an actual materialization and it seemed to work. I actually realised to get the identity function you need to use `RawConstColumnType`, will do that

PR #188, opened 2022-04-28
source-mysql: Sanity-check binlog_expire_logs_seconds at startup

willdonnelly:
**Description:**

It is critical to the MySQL capture's operation that binlog segments not be purged before we replicate them. In normal operation this is fine, because we start out fully caught-up and consume new events as soon as they occur. But if the capture is offline for a while or otherwise has to rewind back to an older state checkpoint, we need those binlog segments to still be present.

There's no way for us to permanently tell the server "don't purge binlog segments until I'm done with them", all we can do is request that the user set the expiry time longer than any likely downtime.

But it's easy to overlook that section of our documentation, so this new sanity check will cause any `check/discover/read` operation to error out if the expiry time is dangerously short.

However, making that a hard failure could cause trouble in some circumstances, so this commit also adds an 'escape hatch' in the form of the `skip_binlog_retention_check` config parameter.

**Workflow steps:**

Any operation that connects to the MySQL database will now error out if the binlog retention period is set dangerously low.

This shouldn't need any documentation changes, but it does introduce a new shortlink (already set up) to the MySQL connector prerequisites.

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -5,17 +5,22 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/alecthomas/jsonschema"
 	"github.com/estuary/connectors/sqlcapture"
 	"github.com/estuary/flow/go/protocols/airbyte"
 	"github.com/estuary/flow/go/protocols/fdb/tuple"
 	"github.com/go-mysql-org/go-mysql/client"
+	"github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/sirupsen/logrus"
 
 	mysqlLog "github.com/siddontang/go-log/log"
 )
+
+const minimumExpiryTime = 7 * 24 * time.Hour
 
 func main() {
 	fixMysqlLogging()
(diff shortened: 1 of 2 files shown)


mdibaiee (approved):
LGTM

source-mysql/main.go
```diff
@@ -66,9 +71,11 @@ type Config struct {
 	User     string `json:"user" jsonschema:"default=flow_capture,description=Database user to connect as."`
 	Password string `json:"password" jsonschema:"description=Password for the specified database user."`
 	DBName   string `json:"dbname" jsonschema:"description=Name of the database to connect to."`
-	ServerID int    `json:"server_id" jsonschema:"description=Server ID for replication."`
+	ServerID int    `json:"server_id" jsonschema:"description=Server ID for replication (each node in a cluster should have a unique ID)."`
 
 	WatermarksTable string `json:"watermarks_table,omitempty" jsonschema:"default=flow.watermarks,description=The name of the table used for watermark writes during backfills. Must be fully-qualified in '<schema>.<table>' form."`
+
+	SkipBinlogRetentionCheck bool `json:"skip_binlog_retention_check,omitempty" jsonschema:"default=false,description=For advanced users only. Bypasses the 'dangerously short binlog retention' sanity check at startup."`
```
psFried, line 78: I think this was a good idea to add something to the config to allow disabling the check. Just to toss out an idea, this might be better as `ensure_minimum_binlog_retention`, which accepts a duration and always errors if the actual retention is less than the given value. Then we could use a default value of 7 days.

psFried (approved):
LGTM % one optional suggestion

source-mysql/main.go
willdonnelly: I considered making it configurable, but I'm worried that having that value be configurable will lead users to want to tweak it.

Specifically I'm worried about the situation where a not-especially-knowledgeable user is setting up a capture from a preexisting database (especially one that they may not have authority to tweak, or not really feel comfortable configuring). If they see this check failing, there's going to be a strong temptation to just lower the setting to whatever value makes it pass, with shaky reasoning like "well the database has been running fine with that setting until now". Which, sadly, is setting themselves up for more difficult problems down the line.

I could see some utility in setting the threshold _higher_ than 7 days, but the parameter description would get pretty complicated at that point (something like "Specifies the minimum binlog retention period which will be accepted by the connector, with 0 meaning to skip the check entirely, and nonzero values <7 days adjusted up to 7 days for your own good"), and I think users that want to explicitly make it larger are already thinking about the right things and will probably get it right on their own.

psFried: Makes sense :+1:

PR #191, opened 2022-05-06
Don't use top-level $ref in json schemas

psFried:
**Description:**

The library that we use to generate json schemas from Go strucs has some
strange defaults, which causes all the generated schemas to use a
top-level $ref that references a nested definition. Apart from just
being a little weird, this breaks the Flow UI code that generates forms
from json schemas. This commit introduces a shared `go-schema-gen`
package that is used by all Go connectors for generating the schemas,
and sets the required configuration to allow the schemas to work with
our UI.

The second commit removes the network-tunnel-service and the bespoke tunneling implementation in `source-postgres`. Now that the connector proxy is wired up, this is no longer necessary. And the inclusion of a `networkTunnel` property in the schema causes `api discover` to fail due to the conflicting property name.

**Workflow steps:** n/a

**Notes for reviewers:** n/a


mdibaiee (approved):
Nice one, thank you Phil!

materialize-elasticsearch/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN go mod download
 
 # Build the connector projects we depend on.
 COPY materialize-boilerplate ./materialize-boilerplate
+COPY go-schema-gen           ./go-schema-gen
```
willdonnelly, line 14: Probably too trivial a nit to be worth fixing, but maybe put `go-schema-gen` above `materialize-boilerplate` for alphabetical-ordering reasons (and likewise for the other Dockerfiles)?

materialize-elasticsearch/driver.go
```diff
@@ -71,12 +71,12 @@ func (resource) GetFieldDocString(fieldName string) string {
...
 
 func (driver) Spec(ctx context.Context, req *pm.SpecRequest) (*pm.SpecResponse, error) {
-	endpointSchema, err := jsonschema.Reflect(&config{}).MarshalJSON()
+	endpointSchema, err := schemagen.GenerateSchema("Elasticsearch Connection", &config{}).MarshalJSON()
 	if err != nil {
 		return nil, fmt.Errorf("generating endpoint schema: %w", err)
 	}
 
-	resourceSchema, err := jsonschema.Reflect(&resource{}).MarshalJSON()
+	resourceSchema, err := schemagen.GenerateSchema("Elasticsearch resource", &resource{}).MarshalJSON()
```
willdonnelly, line 79: It looks like in general these titles are title-cased, maybe `s/resource/Resource` here?

materialize-firebolt/driver.go
```diff
@@ -18,12 +18,12 @@ import (
...
 
 func (driver) Spec(ctx context.Context, req *pm.SpecRequest) (*pm.SpecResponse, error) {
-	var endpointSchema, err = jsonschema.Reflect(&config{}).MarshalJSON()
+	var endpointSchema, err = schemagen.GenerateSchema("Firebolt Connection", &config{}).MarshalJSON()
 	if err != nil {
 		return nil, fmt.Errorf("generating endpoint schema: %w", err)
 	}
 
-	resourceSchema, err := jsonschema.Reflect(&resource{}).MarshalJSON()
+	resourceSchema, err := schemagen.GenerateSchema("Firebolt resource", &resource{}).MarshalJSON()
```
willdonnelly, line 26: Also here

source-mysql/main.go
```diff
@@ -24,7 +24,7 @@ const minimumExpiryTime = 7 * 24 * time.Hour
 
 func main() {
 	fixMysqlLogging()
-	var schema = jsonschema.Reflect(&Config{})
+	var schema = schemagen.GenerateSchema("MySQL Source Spec", &Config{})
 	schema.Title = "MySQL Source Spec"
```
willdonnelly, line 28: Setting `schema.Title` here is made redundant by the `title` parameter of `GenerateSchema`

willdonnelly (approved):
LGTM, a handful of nits that only _barely_ seemed worth mentioning.

source-postgres/main.go
```diff
@@ -19,8 +18,7 @@ import (
 )
 
 func main() {
-	var schema = jsonschema.Reflect(&Config{})
-	schema.Title = "PostgreSQL Source Spec"
+	var schema = schemagen.GenerateSchema("PostgreSQL Source Spec", &Config{})
```
willdonnelly, line 21: Maybe change this to "PostgreSQL Connection" for symmetry with MySQL?

willdonnelly (approved):
LGTM

materialize-elasticsearch/driver.go
psFried: Thanks for pointing that out. I actually changed this to `Elasticsearch Index`, since `Resource` felt too generic. I think that's more in line with the user experience we want, since if you're materializing to Elasticsearch, you probably already have an idea of what the entities are called in there.

PR #192, opened 2022-05-07
BigQuery: Materialization refactor

pothibo:
**Description:**

**Workflow steps:**

**Notes for reviewers:**


psFried:
>  I understand what the code does, but it doesn't make much sense in terms of why that is build the way it is. It seems like a very limited version of a JSON Serializer and if I wonder if we couldn't use the normal JSON marshaller provided by Go instead? Do we know how that's being used in production?

The purpose of the value converter is to translate between the standard types that are used by Flow, and the types that are expected by a SQL driver. For example, a timestamp would typically be represented as a JSON string in flow, but some drivers would require that it be provided as a `time.Time`. In most cases, the standard types can work just fine, and the identity function is used by default.

materialize-bigquery/binding.go
```diff
@@ -1,225 +1,307 @@
...
-	"cloud.google.com/go/bigquery"
+	"cloud.google.com/go/storage"
+	"github.com/estuary/flow/go/protocols/fdb/tuple"
 	pf "github.com/estuary/flow/go/protocols/flow"
-	sqlDriver "github.com/estuary/flow/go/protocols/materialize/sql"
 )
 
-const tempTableNamePrefix = "flow_temp"
+type Binding struct {
+	// CreateSQL is a template SQL query that is generated
```
mdibaiee, line 15:
```suggestion
	// LoadSQL is a template SQL query that is generated
```

materialize-bigquery/binding.go
```diff
@@ -1,225 +1,307 @@
...
+	// The google cloud storage bucket that is configured
+	// to work with this binding. It is expected to be
+	// configured and working when being assigned here.
+	bucket *storage.BucketHandle
+
+	// ExternalTableAlias is a string generated so that it can be used to reference
+	// the external cloud storage data in a SQL Query. At the core of this, there is data
+	// stored in Google Cloud Storage in a JSON format, and the connector needs to access that data to
+	// merge it into BigQuery's dataset. By defining this table alias here, the Queries stored above can
+	// use this alias when compiling the SQL queries, and job will be configured so that the query runs
```
mdibaiee, line 36: I was a bit confused when reading this: the queries above are already compiled when this struct is being initialised, so this property is not really used to generate those queries, right?

materialize-bigquery/binding.go
```diff
@@ -1,225 +1,307 @@
...
-	}
+func (b *Binding) GenerateDocument(key, values tuple.Tuple, flowDocument json.RawMessage) map[string]interface{} {
+	document := make(map[string]interface{})
+	allValues := append(key, values...)
 
-	// Generate the table definition for this materialization.
-	var tableDef = sqlDriver.TableForMaterialization(targetName, "", generator.IdentifierRenderer, spec)
+	for i, value := range allValues {
+		field := b.Table.Fields[i]
+		// If a value is nil, we need to
```
mdibaiee, line 96: abandoned comment :D

materialize-bigquery/binding.go
```diff
@@ -1,225 +1,307 @@
...
-			strings.Join(pkJoins, " AND "),
-		)
+// DestroyWriter clears the writer and delete the underlying
+// asset in Cloud Storage.
+// When this method returns, no writer is associated to this binding
+// and it's required to assign a new one using the SetWriter() method.
+func (b *Binding) DestroyWriter(ctx context.Context) error {
+	defer func() {
+		b.Writer = nil
+	}()
```
mdibaiee, line 140: If we fail to destroy the writer, is it okay to still set it to null? That means we can't retry destroying it I think

materialize-bigquery/transactor.go
```diff
@@ -2,287 +2,271 @@ package main
...
-		// If we have a data file for this round.
-		if b.load.keysFile != nil {
-
-			// Flush and close the keyfile.
-			if err := b.load.keysFile.Close(); err != nil {
-				return fmt.Errorf("closing external data connection file: %w", err)
+		// Error here doesn't mean it's fatal, it just means the checkpoint didn't have
+		// a checkpoint binding, which is a normal scenario if the last time the driver
+		// stored a checkpoint, the binding didn't exist.
+		if driverBinding, err := t.checkpoint.Binding(i); err == nil {
```
mdibaiee, line 68: This seems a bit odd to me: I think you are somehow mixing the new bindings with the old bindings from the checkpoint, but I don't think they are mixable at all.

You might have more or less number of bindings and their ordering is not going to be the same, so you can't correspond them with each other in any way.

Consider this example:
MaterializationSpec v1 creates checkpoint for bindings:
1. A
2. B
3. C
4. D

MaterializationSpec v2 sends an Open request to the connector, and has these bindings:
1. A
2. G
3. D

They can't be mixed with each other at all. The new Materialization might replace all previous bindings with new ones that are not related.

In Acknowledge, you cannot rely on `t.bindings`, and all your operations must solely depend on the checkpoint, otherwise the potentially new MaterializationSpec that comes with a new transaction can still interfere with the old MaterializationSpec for which we have a checkpoint, but we should protect against that.

materialize-bigquery/transactor.go
```diff
@@ -2,287 +2,271 @@ package main
...
+		}
 	}
 
 	return nil
 }
 
-func (t *transactor) Acknowledge(context.Context) error {
+func (t *transactor) Acknowledge(ctx context.Context) error {
+	for _, cpBinding := range t.checkpoint.Bindings {
+		binding := t.bindings[cpBinding.BindingIndex]
```
mdibaiee, line 238: Per my comment above, we can't do this here. We should only depend on `t.checkpoint` alone for this. `t.binding` is based on the new transaction's bindings, not the previous one. Our only point of reference is checkpoint in order to process the previous transaction.

mdibaiee (commented):
Great job so far! Your comments really help understand what is going on, but there is an important point regarding the Acknowledge phase that we need to address

materialize-bigquery/schema.go
```diff
@@ -0,0 +1,292 @@
...
+				// There is a case where a String_ is present but the values aren't set.
+				// I'm not sure how this can happen since the protobuf struct should
+				// be optional and nil, but it did happen in my testing and I couldn't
+				// figured out why, so this is here as a precaution (P-O)
+				case "":
+					f.fieldType = bigquery.StringFieldType
+					f.sanitizers = append(f.sanitizers, sqlDriver.DefaultQuoteSanitizer)
+
+				default:
+					return nil, fmt.Errorf("inferring string column type as the format is not supported: %s", s.Format)
```
psFried, line 219: I think we should treat the previous case (no format) as the default, and not return an error here. It's expected that there will be format identifiers that we won't be able to recognize.

materialize-bigquery/schema.go
```diff
@@ -0,0 +1,292 @@
...
+
+	// This is a test to make sure that their is only 1 infered type for this projection
+	// ie. Inference.Types = [string, integer] is wrong. [string, null] is OK. [string, integer, null] is wrong
+	if !projection.Inference.IsSingleType() {
+		return nil, fmt.Errorf("don't expect multiple types besides Null and a real type, can't proceed: %s", &projection.Inference.Types)
+	}
+
+	for _, p := range projection.Inference.Types {
+		if p == "null" {
+			includesNull = true
```
psFried, line 276: nit: after the above check for `IsSingleType`, if `p == "null"`, then this field must _always_ be null.

materialize-bigquery/schema.go
```diff
@@ -0,0 +1,292 @@
...
+		Name:        f.name,
+		Type:        f.FieldType(),
+		Required:    projection.Inference.Exists == pf.Inference_MUST && !includesNull, // If Required is False, it means it's nullable
+		Description: projection.Inference.Description,
+	}, nil
+}
+
+var identifierSanitizerRegexp = regexp.MustCompile(`[^\-\._0-9a-zA-Z]`)
+
+func identifierSanitizer(text string) string {
```
psFried, line 290: I think we want to quote identifiers that contain those characters instead of modifying them. It's less surprising for users if their column names show up exactly as they gave them. And if we replace characters with `_` then we could end up with duplicate column names since, for example, `foo/bar` and `foo$bar` would end up as `foo_bar`.

psFried (changes requested):
This has seen a lot of improvement, but still has a ways to go before its ready to merge. I got through much of this, but not all. I'll finish going through tomorrow sometime, but wanted to get you this initial feedback. I think this is getting much closer though, as most of the issues I'm spotting seem relatively straight forward to fix.

materialize-bigquery/config.go
```diff
@@ -0,0 +1,109 @@
...
+func (c *config) StorageClient(ctx context.Context) (*storage.Client, error) {
+	cloudStorageClient, err := storage.NewClient(ctx, c.ClientOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("creating cloud storage client: %w", err)
+	}
+
+	return cloudStorageClient, nil
+}
+
+func (c *config) Validate() error {
```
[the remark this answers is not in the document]
pothibo: Validate() is called implicitly from `pf.UnmarshallStrict()`

materialize-bigquery/schema.go
pothibo: This piece of code was carried over from the past implementation, if I make the change to what you suggest, it will be a breaking change from the previous implementation. We're good with it?

materialize-bigquery/transactor.go
pothibo: I think this new version fixes this.

materialize-bigquery/binding.go
pothibo: The externalTableAlias is used as a configuration map for bigquery so it can locates the table, it's a readonly property that doesn't change once it's created.

Issue #190, opened 2022-05-05
Fix flow_watermarks in CDC Captures 


dyaffe:
Problem:

Currently when a user captures data from a database such as Postgres or MySQL, we automatically discover and capture from the flow_watermarks table.  This obviously isn't something that would be useful for the user so it would be ideal to stop that behavior.

Potential solution:

It would be ideal for us to create flow_watermarks under a different non-public schema for the user.  This would ensure that we wouldn't find it when capturing from the database while also not polluting the database's schemas.

We automatically try to create flow_watermarks, so a solution here would involve updating that creation process as well as our docs which contain SQL queries to create flow_watermarks.

willdonnelly:
Requiring that the watermarks table be in a separate schema seems like it would cause more problems than it solves.

What if we just did the dumbest thing that could work, and modified the `discover` subcommand's implementation to remove the specific stream whose name+namespace matches the watermark table before outputting the discovered catalog?

dyaffe:
That seems like a plan to me

PR #193, opened 2022-05-09
sqlcapture: Exclude watermarks table from discovery results

willdonnelly:
**Description:**

It's basically never useful to capture this table, so we shouldn't be implicitly suggesting that by including it in the `flowctl discover` generated catalog.

This is a simple check performed right before emitting the discovered catalog, so it shouldn't have any impact on other uses of the discovery logic.

Our test coverage doesn't extend to the `sqlcapture.AirbyteMain` subcommands, so this was tested manually by running `flowctl discover` against a local PostgreSQL instance.

This fixes https://github.com/estuary/connectors/issues/190

**Workflow steps:**

Perform discovery (via `flowctl discover` or the new UI) using the `source-mysql` or `source-postgres` connectors, and the watermarks table should no longer be included in the resulting catalog.

### DIFF
--- a/sqlcapture/main.go
+++ b/sqlcapture/main.go
@@ -42,6 +42,18 @@ func AirbyteMain(spec airbyte.Spec, init func(airbyte.ConfigFile) (Database, err
 			if err != nil {
 				return err
 			}
+
+			// Filter the watermarks table out of the discovered catalog before output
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM This looks like a nice UX improvement.

mdibaiee (approved):

PR #202, opened 2022-05-23
MySQL Table Metadata and Related Refactorings

willdonnelly:
**Description:**

This is a pretty large change, but it kind of had to be in order to move from one correct, working state to another. The high-level goal of this PR is essentially "make it possible to run `source-mysql` against databases with `binlog_row_metadata = 'MINIMAL'`", but of course the detail work to make that happen has other implications as well.

This series of commits makes the following high-level changes to the generic `sqlcapture` logic:

- First, all responsibility for translating values into proper JSON-serializable form is moved into the DB-specific logic. That is to say, there's now an expectation that the `map[string]interface{}` values coming from `ScanTableChunk` and the replication events interface are fully translated and we can just output them as needed.
- Second, the `ReplicationStream` is now provided with a set of tables it's supposed to care about. The initial set is given as an argument to `StartReplication` and additional tables can/will be added later on via the `ActivateTable(streamID string)` method.
- Thirdly, a new concept of "persistent, table-specific metadata" is added. This metadata is an opaque `json.RawMessage` which will be serialized in state checkpoints if present. At startup all available metadata is passed to `StartReplication`, and it can be updated during replication using `ChangeEvent`s with a new `MetadataOp` type.
- Finally, every capture now begins with an additional "stream to watermark" operation which must complete before `ActivateTable()` will be used to start processing change events for newly-added streams. A `"Pending"` table mode is used to represent this additional state. This allows us to guarantee that for each table we only ever pay attention to replication events taking place _after_ the connector restart that added its capture binding.

For the most part `source-postgres` doesn't care about these changes, and has only been modified as necessary to keep everything working. This entailed adding some additional arguments/methods to the interfaces, keeping track of the set of "active tables" during replication, and updating test snapshots, but there shouldn't really be any notable changes as far as the user is concerned.

On the other hand `source-mysql` takes full advantage of the new behaviors. The replication logic uses persistent table metadata to keep track of a list of column names and a name->type mapping. This additional information is used to drive change event decoding and value translation, and is thus kind of "the point" of this whole PR.

The metadata is initialized from discovery results the first time a new stream is activated for replication, and is (currently) never updated after that. Instead, it's simply an error which immediately halts the capture if we ever observe an `ALTER TABLE` (or `DROP` or `TRUNCATE` or `RENAME` etc...) on a table which we're actively replicating from. But thanks to the aforementioned changes around the `"Pending"` state, this should only happen if the table is actually altered after we start capturing from it.

In order to process MySQL `QueryEvent`s and determine whether a particular `ALTER TABLE` query impacts a table we're capturing from, `source-mysql` now integrates the Vitess SQL parser package. I've tried to make the implementation fairly future-proof, so in principle we can extend this to support a defined subset of possible table alterations without too much fuss.

**Workflow steps:**

For the most part the connector works the same as before, except that it will now halt if you `ALTER` a table that it's capturing from, but this tradeoff allows it to work on MySQL instances without `binlog_row_metadata = 'FULL'` instead.

I intend to put together a documentation PR after I send this one out, but just for completeness here are the docs changes it will involve:

- https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/#prerequisites should be modified to no longer document the `binlog_row_metadata` requirement, and likewise for the step in the setup SQL.
- https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/#google-cloud-sql (and the following Azure section) should be changed to say that it ought to work but we don't yet have verified setup instructions.
- https://go.estuary.dev/eiKbOh is linked to from a couple of possible capture errors. Currently it just points at the MySQL connector docs, but it should be pointed at a "Troubleshooting" section, or even more specifically an explanation that something has gone wrong with table metadata tracking in a way that is probably unfixable, and so they should tell us about the bug (because these two errors are "should never happen" type ones) and delete+recreate the capture binding for that table.
- https://go.estuary.dev/eVVwet is linked to from some other capture errors, and is used to indicate that an unsupported query event has occurred. This should also go to a troubleshooting section, with several cases describing the possible issues. There are several cases here, including:
  - Queries like `ALTER TABLE` which are forbidden unless/until we start implementing support for them.
  - Queries like `DROP TABLE` or `TRUNCATE TABLE` which we will never support.
  - Queries like `INSERT` or `UPDATE` or `DELETE` which indicate that `binlog_format != 'ROW'`. (Hang on now, that isn't mentioned under "Prerequisites" is it? I'll fix that in the docs PR too)
  - Other queries (I don't know if any exist) which are fine and should be whitelisted so the connector ignores them.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/202)


psFried:
I haven't reviewed all the code here, but I just have to say... your PR description is one of the best written descriptions I've ever read. If there was a Pulitzer for this type of thing, I'd nominate this PR. It really makes the entire thing so much easier to review when it's explained so clearly up front like that. Thanks @willdonnelly :1st_place_medal:

jgraettinger (approved):
LGTM!

I left a bunch of comments and questions below. I spotted no correctness issues, and I think this is good to land as-is, so take this as feedback for future improvements.

> Reviewed 8 of 8 files at r1, 4 of 4 files at r2, 5 of 5 files at r3, 3 of 22 files at r5, 2 of 3 files at r6, 3 of 10 files at r8, 1 of 2 files at r9, 6 of 7 files at r11, 36 of 36 files at r12, 1 of 3 files at r13, 2 of 2 files at r14, 2 of 3 files at r15, 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/202#-:-N3V74sxDRh1drklNyGO:brj0ule)* status: all files reviewed, 10 unresolved discussions (waiting on @psFried and @willdonnelly)

___
*[`source-mysql/backfill.go` line 68 at r1](https://reviewable.io/reviews/estuary/connectors/202#-N2qnsPJ--s_431lssPf:-N2qnsPJ--s_431lssPg:b-jygr3l) ([raw file](https://github.com/estuary/connectors/blob/acbeca5dc3ceefbbd82be50d20b6dc5973a25b3e/source-mysql/backfill.go#L68)):*
> ```Go
> 		}
> 		if err := translateRecordFields(info, fields); err != nil {
> 			return nil, fmt.Errorf("error backfilling table %q: %w", table, err)
> ```

nit: error text "translating record fields to JSON"?

nit, optional: is it possible to pass `val` and `info.Columns` into translateRecordFields (do the indexes align?). Then you could omit some of the map assignment / lookup work being done.
___
*[`source-mysql/replication.go` line 24 at r2](https://reviewable.io/reviews/estuary/connectors/202#-N3Qht-m8SulnSmqNVQi:-N3Qht-m8SulnSmqNVQj:bwhoofj) ([raw file](https://github.com/estuary/connectors/blob/096f1af4a0ba52b4d22c893a345b3644deb434a7/source-mysql/replication.go#L24)):*
> ```Go
> var replicationBufferSize = 1024
> 
> func (db *mysqlDatabase) StartReplication(ctx context.Context, startCursor string, captureTables map[string]bool, discovery map[string]sqlcapture.TableInfo, metadata map[string]json.RawMessage) (sqlcapture.ReplicationStream, error) {
> ```

supernit: `captureTables` could also be `map[string]struct{}` which is slightly more idiomatic for a hash set.

nit: The number of arguments here is starting to be a bit of a smell, and I'm getting confused as to what they are :shrug:

Three of these are essentially maps on streamIDs, so there's an opportunity to move values into a higher-level argument struct?
___
*[`source-mysql/replication.go` line 139 at r3](https://reviewable.io/reviews/estuary/connectors/202#-N3Ub9B97qS6jA5yMgjb:-N3Ub9B97qS6jA5yMgjc:b-j0gawk) ([raw file](https://github.com/estuary/connectors/blob/377655ff338385e540f09645d4623435be8da056/source-mysql/replication.go#L139)):*
> ```Go
> 
> type mysqlTableSchema struct {
> 	Columns     []string          `json:"columns"`
> ```

would it make sense to use a compound struct having both name and type, rather than indirecting through a map?
___
*[`source-mysql/replication.go` line 137 at r16](https://reviewable.io/reviews/estuary/connectors/202#-N3Uo1ljAVqPcwvGdqQC:-N3Uo1lk2--6L7nYzmNM:bwe0tyl) ([raw file](https://github.com/estuary/connectors/blob/03a9e70166ecc501b50e47730021ac20eac499bb/source-mysql/replication.go#L137)):*
> ```Go
> 	gtidTimestamp time.Time // The OriginalCommitTimestamp value of the last GTID Event
> 
> 	// Mutex protecting the table activity and metadata maps, so
> ```

nit: these can be placed into a sub-struct to make it super clear to the reader what is / should be guarded by the mutex vs what isn't.

Example:

```
type mysqlReplicationStream {

  foo struct {
    sync.RWMutex
    activeTables map[string]bool
    ...
  }
}
```

Then use a `rs.foo.RLock()`, `rs.foo.activeTables[bar]` etc.

`foo` could be `mu`, `inner`, `shared`, `guard`, maybe `tables` in this case, etc :shrug: 
___
*[`source-mysql/replication.go` line 175 at r16](https://reviewable.io/reviews/estuary/connectors/202#-N3UpM5QDm4eE-nSxxa9:-N3UpM5QDm4eE-nSxxaA:be5oei3) ([raw file](https://github.com/estuary/connectors/blob/03a9e70166ecc501b50e47730021ac20eac499bb/source-mysql/replication.go#L175)):*
> ```Go
> 		var metadataEvents []sqlcapture.ChangeEvent
> 		rs.tablesMutex.RLock()
> 		if rs.dirtyMetadata != nil {
> ```

nit: this `if` guard can be dropped.
___
*[`source-mysql/replication.go` line 197 at r16](https://reviewable.io/reviews/estuary/connectors/202#-N3V63t42m0WChe6bUE5:-N3V63t42m0WChe6bUE6:b-oc7vum) ([raw file](https://github.com/estuary/connectors/blob/03a9e70166ecc501b50e47730021ac20eac499bb/source-mysql/replication.go#L197)):*
> ```Go
> 
> 		// Process the next binlog event from the database.
> 		var event, err = rs.streamer.GetEvent(ctx)
> ```

Question: could the locking over metadata events be avoided if table activation were inverted a bit, and we were selecting here over _either_ a CDC event _or_ an activation / associated metadata of a table that we're now responsible for tracking?

This is pretty half-baked because my mental picture of the control flow is pretty fuzzy, but I ask because the locking is a bit of a smell and potentially something that can be alleviated by multiplexing on the event types processed by this loop?
___
*[`source-postgres/discovery.go` line 107 at r1](https://reviewable.io/reviews/estuary/connectors/202#-N3QeyVK-dg6kOiUEgfc:-N3QeyVK-dg6kOiUEgfd:b-gmxrnw) ([raw file](https://github.com/estuary/connectors/blob/acbeca5dc3ceefbbd82be50d20b6dc5973a25b3e/source-postgres/discovery.go#L107)):*
> ```Go
> 
> func translateRecordFields(table *sqlcapture.TableInfo, f map[string]interface{}) error {
> 	if f == nil {
> ```

when can `f` be nil ? Would this be, for example, a "before" state of an `insert` event?
___
*[`sqlcapture/capture.go` line 417 at r1](https://reviewable.io/reviews/estuary/connectors/202#-N3QfWXR5h5cdbI6VbQW:-N3QfWXR5h5cdbI6VbQX:b-eu50lz) ([raw file](https://github.com/estuary/connectors/blob/acbeca5dc3ceefbbd82be50d20b6dc5973a25b3e/sqlcapture/capture.go#L417)):*
> ```Go
> 			tableInfo = &discoveryInfo
> 		}
> 		events, err := c.Database.ScanTableChunk(ctx, schema, table, tableInfo, streamState.KeyColumns, resumeKey)
> ```

When is `streamID` expected / allowed to not be in `c.discovery`?
___
*[`sqlcapture/capture.go` line 466 at r1](https://reviewable.io/reviews/estuary/connectors/202#-N3QgkNA0P1BTalw6tbB:-N3QgkNA0P1BTalw6tbC:b-8q0ego) ([raw file](https://github.com/estuary/connectors/blob/acbeca5dc3ceefbbd82be50d20b6dc5973a25b3e/sqlcapture/capture.go#L466)):*
> ```Go
> 			Data:      json.RawMessage(rawData),
> 		},
> 	})
> ```

Would it have been feasible to have `TranslateRecordFields` (note the plural) be part of the DB interface, and keep orchestration of field translation as a concern of `sqlcapture/capture.go`? The major reason for pushing it down was to allow for the use of a DB-internal cache, as I understand, which that interface could still allow?

I'm not specifically advocating for this, just trying to understand the design space.
___
*[`sqlcapture/interface.go` line 132 at r3](https://reviewable.io/reviews/estuary/connectors/202#-N3UcRrwBcpZNCEDX7aY:-N3UcRrwBcpZNCEDX7aZ:b2sujee) ([raw file](https://github.com/estuary/connectors/blob/377655ff338385e540f09645d4623435be8da056/sqlcapture/interface.go#L132)):*
> ```Go
> 	Columns     map[string]ColumnInfo // Information about each column of the table.
> 	PrimaryKey  []string              // An ordered list of the column names which together form the table's primary key.
> 	ColumnNames []string              // The names of all columns, in the table's natural order.
> ```

Related to comment above, a `ColumnInfo` type or similar seems warranted?

willdonnelly:
CI is passing now, other than `materialize-rockset` which seems like it might be some sort of quota issue with the actual destination. Since the previous review was an LGTM % nits and I've fixed all the low-hanging fruit there, I'm going to go ahead and get this merged.

PR #204, opened 2022-05-24
BigQuery: Add support for JSON Form

pothibo:
**Description:**
The UI depends on values provided to the connector to show information
and create the actual form.

Also made a change to the CredentialsJSON field to be a string instead of a byte[] slice because a byte slice would get encoded in JSONSchema differently that is incompatible with how flow deals with the data: https://github.com/invopop/jsonschema/blob/297bb072302c4423669838e04bb7705277e50224/reflect.go#L417-L419

I believe the changes here are backward compatible since string and byte slice can be cast to each type in any direction.

This should fix some issues with the UI showing an error message trying to get test the BQ materialization:

```json
{
    "error": "SchemaBuild",
    "description": "failed to build json schema: at keyword 'media' of schema 'request://schema#/properties/credentials_json': unexpected keyword 'media'"
}
```

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-bigquery/.snapshots/TestSpecification
+++ b/materialize-bigquery/.snapshots/TestSpecification
@@ -52,5 +52,5 @@
     "additionalProperties": false,
     "type": "object"
   },
-  "documentation_url": "https://docs.estuary.dev/#FIXME"
+  "documentation_url": "https://go.estuary.dev/materialize-bigquery"
 }
--- a/materialize-bigquery/.snapshots/TestSpecification
+++ b/materialize-bigquery/.snapshots/TestSpecification
@@ -30,10 +30,7 @@
         "type": "string"
       },
       "credentials_json": {
-        "type": "string",
-        "media": {
-          "binaryEncoding": "base64"
-        }
+        "type": "string"
       }
     },
     "additionalProperties": false,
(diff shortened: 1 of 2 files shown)


materialize-bigquery/.snapshots/TestSpecification
```diff
@@ -30,10 +30,7 @@
         "type": "string"
       },
       "credentials_json": {
-        "type": "string",
-        "media": {
-          "binaryEncoding": "base64"
-        }
+        "type": "string"
```
psFried, line 33: I just realized this was also missing the `secret` annotation. This should also have `"secret": true`, which will cause the value to be encrypted. I just added some [docs on that](https://github.com/estuary/connectors/blob/a161ad0ebb02fbbfe502466f6d8f6c8f85063f05/config_schema_guidelines.md#use-secret-annotation-for-all-sensitive-fields), for reference.

psFried (approved):
LGTM. The `secret` annotation can either be addressed in this PR or a separate one at your preference.

materialize-bigquery/.snapshots/TestSpecification
pothibo: I'll do it in another PR that I'll start tomorrow. Since this is still using the sqlDriver abstraction, I'm not sure how simple it is to do. I need to research it.

[Filed an issue for it](https://github.com/estuary/connectors/issues/206)

PR #207, opened 2022-05-25
MySQL Config Schema

willdonnelly:
**Description:**

This PR tries to clean up the generated configuration JSON-schema for the `source-mysql` connector so that it will look nice in the web UI, and also so that it obeys the new [guidelines](https://github.com/estuary/connectors/blob/main/config_schema_guidelines.md).

Part of that required some changes to the `schemagen` helper package which is used by six connectors:

- Setting `"additionalProperties": true` ([as recommended here](https://github.com/estuary/connectors/blob/main/config_schema_guidelines.md#always-allow-additional-properties-in-endpoint-configurations)) on the root schema.
- Introducing a helper function so that the annotations `jsonschema_extras:"secret=true"` and `jsonschema_extra:"advanced=true"` produce boolean values in the generated schema rather than the strings `"true"` or `"false"`.

Also as a part of this PR, I added some logic to `schemagen.GenerateSchema()` to prevent the resulting schema from containing a `"definitions"` section. I did this for two reasons: firstly, because the `DoNotReference: true` setting means that these will never actually be referenced (instead they're inlined into the top-level schema and no mention is made of the redundant definition), so it's just a bunch of extra noise for no reason. And secondly, because the actual `*jsonschema.Type` representations _aren't shared_ between these definitions and their inlined equivalents, so it would have required additional code (albeit just a few lines) to perform the same boolean-annotation fixup on the definition copies.

**Workflow steps:**

MySQL configs will need to be rewritten, but the actual set of fields hasn't changed so it should be obvious how to do that.

https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL


go-schema-gen/generate.go
```diff
@@ -16,6 +17,32 @@ func GenerateSchema(title string, configObject interface{}) *jsonschema.Schema {
 		DoNotReference: true,
 	}
 	var schema = reflector.ReflectFromType(reflect.TypeOf(configObject))
+	schema.AdditionalProperties = json.RawMessage("true")
```
psFried, line 20: uber nit: This could probably also be `nil` to reduce some noise in the schemas.

psFried (approved):
LGTM

go-schema-gen/generate.go
willdonnelly: Good point! Done and I'll get this merged after the checks finish

PR #208, opened 2022-05-25
BigQuery: Add JSONSchema attribute in annotation for UI Form

### LINKED ISSUE #206: Add Secret annotation to BQ's credentials JSON field
The CredentialsJSON field should be annotated as a secret following the guidelines as defined [here](https://github.com/estuary/connectors/blob/a161ad0ebb02fbbfe502466f6d8f6c8f85063f05/config_schema_guidelines.md#use-secret-annotation-for-all-sensitive-fields).

Since the BQ connector uses the sqlDriver interface abstraction, I'm not sure how that can done directly from the connector, but we need to add that annotation so it can integrate secret safely from the UI.

pothibo:
JSONSchema needs to include some extra attributes to be presented to the
UI so the forms are properly built.

The added logic will have the test failed until
https://github.com/estuary/flow/pull/496 merges and the connector is
rebased to include the new version of the sqlDriver.

The reason for this is because the the DoNotReference needs to be enable
otherwise the test fails and the Schema is formulated in a different
way.

This should be a safe backward compatible fix.

Close #206

### DIFF
--- a/materialize-bigquery/.snapshots/TestSpecification
+++ b/materialize-bigquery/.snapshots/TestSpecification
@@ -48,7 +48,15 @@
       }
     },
     "additionalProperties": false,
-    "type": "object"
+    "type": "object",
+    "definitions": {
+      "credential": {
+        "type": "string",
+        "title": "Credentials",
+        "description": "Google Cloud Service Account JSON credentials in base64 format.",
+        "secret": true
+      }
+    }
   },
   "resource_spec_schema_json": {
     "$schema": "http://json-schema.org/draft-04/schema#",
--- a/materialize-bigquery/.snapshots/TestSpecification
+++ b/materialize-bigquery/.snapshots/TestSpecification
@@ -2,7 +2,6 @@
   "endpoint_spec_schema_json": {
     "$schema": "http://json-schema.org/draft-04/schema#",
     "required": [
-      "billing_project_id",
       "project_id",
       "dataset",
       "region",
--- a/materialize-bigquery/.snapshots/TestSpecification
+++ b/materialize-bigquery/.snapshots/TestSpecification
@@ -2,35 +2,50 @@
   "endpoint_spec_schema_json": {
     "$schema": "http://json-schema.org/draft-04/schema#",
     "required": [
+      "billing_project_id",
       "project_id",
       "dataset",
+      "region",
       "bucket",
-      "bucket_path"
+      "bucket_path",
+      "credentials_json"
     ],
     "properties": {
       "billing_project_id": {
-        "type": "string"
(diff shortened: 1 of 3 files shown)


materialize-bigquery/bigquery.go
```diff
@@ -21,39 +22,30 @@ func main() {
...
-		return "Google Cloud Storage bucket that is going to be used to store specfications & temporary data before merging into BigQuery."
-	case "BucketPath":
-		return "A prefix that will be used to store objects to Google Cloud Storage's bucket."
-	case "CredentialsFile":
-		return "URI that points to a JSON Credentials for Google Clouse Service."
-	case "CredentialsJSON":
-		return "Google Cloud Service Account JSON credentials in base64 format."
-	default:
-		return ""
+func (credential) JSONSchemaType() *jsonschema.Type {
```
psFried, line 27: This seems like a pretty reasonable pattern for customizing schema generation, given that the go struct tags won't work :+1:

materialize-bigquery/bigquery.go
```diff
@@ -21,39 +22,30 @@ func main() {
...
+			"secret": true,
+		},
 	}
+
+	return ty
+}
+
+// Config represents the endpoint configuration for BigQuery.
+type config struct {
+	BillingProjectID string     `json:"billing_project_id" jsonschema:"title=Billing Project ID,description=Billing Project ID connected to the BigQuery dataset. It can be the same value as Project ID."`
```
psFried, line 42: You were right about these struct tags getting a little long. I noticed there's a [feature that uses field comments for the descriptions](https://github.com/invopop/jsonschema#using-go-comments), but I'll defer to your judgement on whether it's worth using. Totally fine to leave it like it is.

materialize-bigquery/.snapshots/TestSpecification
```diff
@@ -2,35 +2,50 @@
   "endpoint_spec_schema_json": {
     "$schema": "http://json-schema.org/draft-04/schema#",
     "required": [
+      "billing_project_id",
```
psFried, line 5: I don't think this should be required. AFAIK it's completely optional, and most users won't need to specify it.

psFried (approved):
LGTM % the `billing_project_id` and of course waiting until the tests will pass

materialize-bigquery/bigquery.go
pothibo: Oh that would be a good idea, but I don't think that would work since the `Reflector{}` is initialized inside the sqlDriver's `Spec()` call which is not accessible from us within a driver.

But, that's something I didn't think about so that is an interesting option for when it can be used!

PR #212, opened 2022-05-26
materialize-firebolt: fix prefix without trailing slash causing issues

mdibaiee:
**Description:**

- If a `prefix` was given without a trailing slash, it would cause issues such as:
```
{"timestamp":"2022-05-12T18:55:50.509439286Z","level":"INFO","message":"invoke connector /connector/connector, [\\"apply-upsert\\"]"}
running external table creation query: response error code 500, Could not access the data source. Please make sure the engine and bucket are on the same AWS region.
```
- This will fix that case as it ensures there is a slash between the prefix and the spec filename. This is consistent with how we handle the prefix in other places of the connector. I also added a note to the JSON schema that the value must now be without a trailing slash.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-firebolt/config.go
+++ b/materialize-firebolt/config.go
@@ -13,7 +13,7 @@ type config struct {
 	AWSSecretKey string `json:"aws_secret_key,omitempty" jsonschema:"title=AWS Secret Key"`
 	AWSRegion    string `json:"aws_region,omitempty" jsonschema:"title=AWS Region"`
 	S3Bucket     string `json:"s3_bucket" jsonschema:"title=S3 Bucket"`
-	S3Prefix     string `json:"s3_prefix,omitempty" jsonschema:"title=S3 Prefix"`
+	S3Prefix     string `json:"s3_prefix,omitempty" jsonschema:"title=S3 Prefix,default=/"`
 }
 
 func (c config) Validate() error {
@@ -50,7 +50,7 @@ func (config) GetFieldDocString(fieldName string) string {
 	case "S3Bucket":
 		return "Name of S3 bucket where the intermediate files for external table will be stored."
 	case "S3Prefix":
-		return "A prefix for files stored in the bucket."
+		return "A prefix for files stored in the bucket, must not have a trailing slash. Example: /my-prefix."
 	case "AWSKeyId":
 		return "AWS Key ID for accessing the S3 bucket."
 	case "AWSSecretKey":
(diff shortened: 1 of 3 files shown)


materialize-firebolt/spec.go
```diff
@@ -23,7 +24,8 @@ func LoadSpec(cfg config, materialization string) (map[string]*pf.Materializatio
 	sess := session.Must(session.NewSession(&awsConfig))
 	downloader := s3manager.NewDownloader(sess)
 	buf := aws.NewWriteAtBuffer([]byte{})
-	existingSpecKey := fmt.Sprintf("%s%s.flow.materialization_spec", cfg.S3Prefix, materialization)
+
+	existingSpecKey := strings.TrimLeft(fmt.Sprintf("%s/%s.flow.materialization_spec", cfg.S3Prefix, materialization), "/")
```
psFried, line 28: Ok, I guess this is what allows the default value to work?

I'm wondering if it might be better to just disallow both leading and trailing slashes? AFAIK it'd be really weird, but technically permissible, for an S3 object key to start with a `/`, so it might be confusing to show it in the UI and then strip off here. Up to you, but just wanted to raise it as something to consider.

psFried (changes requested):
I'm a little uncertain about the `/` being the default value, given that the description says it can't end with a `/`, but I'm not really sure exactly how to proceed, so I'll defer to you on that. Happy to talk through it tomorrow morning if that's helpful.

materialize-firebolt/config.go
```diff
@@ -13,7 +13,7 @@ type config struct {
 	AWSSecretKey string `json:"aws_secret_key,omitempty" jsonschema:"title=AWS Secret Key"`
 	AWSRegion    string `json:"aws_region,omitempty" jsonschema:"title=AWS Region"`
 	S3Bucket     string `json:"s3_bucket" jsonschema:"title=S3 Bucket"`
-	S3Prefix     string `json:"s3_prefix,omitempty" jsonschema:"title=S3 Prefix"`
+	S3Prefix     string `json:"s3_prefix,omitempty" jsonschema:"title=S3 Prefix,default=/"`
```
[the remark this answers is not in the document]
mdibaiee: I had to set the default to something other than an empty prefix, because the UI was complaining about this field being empty, even though it is not marked as required.

materialize-firebolt/spec.go
mdibaiee: Tested this with `/test/blah/` as input, and  the files are put in `bucket/test/blah/...` as expected, so I think we are good 👍🏽

PR #218, opened 2022-05-27
BigQuery: Fix applyUpsert when a table exists

pothibo:
When an upsert happens, I believe that by default it will try to run a query like:

```sql
BEGIN TRANSACTION
CREATE TABLE abc ...;
CREATE TABLE def ....;
COMMIT;
```

Because the `CREATE` statement doesn't include `IF NOT EXISTS`, if one of the table already exists, all the other one won't be created. By forcing the `IF NOT EXISTS`, it allows for all remaining table that does not exist to be created.

I cannot be 100% certain this is the root cause, but I think that's why
a connector can fail during bootup. ApplyUpsert starts by trying to
create all tables for internal uses (_flow_*) and for projections.

When creating those tables, if one of them exists, it will return an
error, and as a result the subsequent table aren't created, which cause
the connector to fail when it starts the transactor and wants to commit
data to the data storage.

### DIFF
--- a/materialize-bigquery/endpoint.go
+++ b/materialize-bigquery/endpoint.go
@@ -167,10 +167,13 @@ func (e *Endpoint) CreateTableStatement(table *sqlDriver.Table) (string, error) 
 	if table.Temporary {
 		builder.WriteString("TEMPORARY ")
 	}
-	builder.WriteString("TABLE ")
-	if table.IfNotExists {
-		builder.WriteString("IF NOT EXISTS ")
-	}
+
+	// Ignoring the table.IfNotExists because we need to
+	// include it here otherwise, the statement would fail if the table exists
+	// and since create gets called multiple time, if one table exists and another doesn't,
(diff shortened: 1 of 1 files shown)


psFried (changes requested):
I don't think I understand how this helps. My understanding is that `table.IfNotExists` would already be `true` because we call [`TableForMaterialization`](https://github.com/estuary/flow/blob/master/go/protocols/materialize/sql/sqlgen.go#L71) to generate them. Am I just missing something? I'd generally prefer to just fix whatever callsite is passing `false` rather than ignoring the value of the parameter.

pothibo:
Meant to add a comment when closing the PR, but it didn't go through. I'm closing this because I don't have an opinion on this. I've seen this happen a few times, and I am not certain this causes the issue, like I mentioned in the description of the PR.

But I don't think we want to ever not use `IF NOT EXISTS` because the code doesn't support alteration to an existing tables.

To me this was an easy fix to test things out in prod and see if it works, but if we want a more thorough investigation, I'll have to change the approach.

PR #236, opened 2022-06-02
source-postgres: Mark 'password' config property as secret

willdonnelly:


travjenkins (approved):
LGTM - compared to the MySQL connector where I know there is a secret field and this is the same change.

PR #243, opened 2022-06-07
ci: push image with version if not already exists

mdibaiee:
**Description:**

- Publish images with a version _IF NOT ALREADY EXISTING_. This is important to avoid accidentally overriding the last version. Version must always be bumped if we want to publish a new version. The `dev` tag will continue to be overriden every time we push.

**Workflow steps:**

Run main branch github action

N/A

**Notes for reviewers:**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,5 +1,9 @@
 name: CI
 
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on:
   push:
     branches: [main]
@@ -166,4 +170,4 @@ jobs:
           context: .
           file: ${{ matrix.connector }}/Dockerfile
           push: true # See 'if' above
-          tags: ghcr.io/estuary/${{ matrix.connector }}:dev
+          tags: ghcr.io/estuary/${{ matrix.connector }}:dev,ghcr.io/estuary/${{ matrix.connector }}:v1


.github/workflows/ci.yaml
```diff
@@ -167,3 +167,18 @@ jobs:
           file: ${{ matrix.connector }}/Dockerfile
           push: true # See 'if' above
           tags: ghcr.io/estuary/${{ matrix.connector }}:dev
+
+      - name: Push ${{ matrix.connector }} image with version tag
+        if: ${{ github.event_name == 'push' }}
+        run: |
+          export IMAGE_VERSION=$(grep VERSION ${{ matrix.connector }}/Dockerfile | cut -d = -f 2)
```
willdonnelly, line 174: Perhaps `grep '^LABEL VERSION'` to avoid accidentally triggering on some `FOOBAR_VERSION` string that might be mentioned somewhere?

.github/workflows/ci.yaml
```diff
@@ -167,3 +167,18 @@ jobs:
...
+          version_exists=$(docker manifest inspect ghcr.io/estuary/${{ matrix.connector}}:$IMAGE_VERSION > /dev/null ; echo $?)
```
willdonnelly, line 175: Does this check have a race condition if there are multiple merges to master when one of them bumps the version number? Or is there something preventing that?

mdibaiee: @willdonnelly that is a good point. I will disable concurrency on main branch following [this](https://github.com/estuary/flow/blob/master/.github/workflows/main.yml#L15-L17)

psFried (changes requested):
I think it's worth at least exploring a different approach for this. Using `grep | cut` to get the version from the Dockerfile seems pretty counterintuitive to me. A more typical practice would be to pass the version in from the outside using `--build-arg`, and I think it would be preferable to do that, as long as the `LABEL` instruction allows referencing args.

My assumption here is that we don't actually _want_ to manage these version numbers by keeping them in the dockerfile. So my thought process is that, given that we don't yet know how we really want to manage connector versions in the long term, it may be better to just hard-code the versions in `ci.yaml`, since that's A, easier to change, and B, more obviously just a placeholder solution. Happy to talk though this, though, if you were thinking of it differently.

mdibaiee:
My idea was to publish images with full semver (`major.minor.patch`), in that case we need a way to get the version and that's pretty much arbitrary I think. People put it in docker images, they sometimes put it in the CI configuration, in a separate config, etc.

For now I updated the PR to just publish the `v1` hardcoded, beside the `dev` tag as suggested, and we can come back to this and perhaps do a bit of research see if we can find a _good_ solution.

PR #244, opened 2022-06-07
Update documentation URLs for source connectors

saterus:
**Description:**

Many of these connectors were missing documentation urls. This field gets extracted by the UI from the connector spec in order to render a contextual link to the connector documentation.

**Workflow steps:**

N/A

* I've added shortlinks for all these connectors

**Notes for reviewers:**

N/A


mdibaiee (approved):
Thank you!

PR #247, opened 2022-06-07
source-mysql: Don't group username/password in config

willdonnelly:
**Description:**

As discussed on Slack the other day [1] a nested object is rendered as a separate tab in the web UI. Travis was of the opinion (paraphrased) that all basic details should be scalar properties of the root configuration so that they all appear in the 'Basic Information' tab.

[1] https://app.slack.com/client/T0120UQAX7X/C012S5U0WP2/thread/C012S5U0WP2-1654034026.249709

**Workflow steps:**

This tweaks the configuration of `source-mysql`, hopefully for the better.

I'll submit a PR for https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/ if/when this is merged.

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -69,13 +69,9 @@ func fixMysqlLogging() {
 // capture changes from it.
 type Config struct {
 	Address  string         `json:"address" jsonschema:"title=Server Address,default=127.0.0.1:3306,description=The host or host:port at which the database can be reached."`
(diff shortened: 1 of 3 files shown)


willdonnelly:
Going to go ahead and merge this, since all our test captures under `estuary/` just got blown away and so nothing should be relying on it right now.

PR #248, opened 2022-06-07
source-postgres: Refactor configuration schema

willdonnelly:
**Description:**

This PR moves several settings (watermarks table, replication slot, publication) into an 'Advanced' subsection, and merges the `host` and `port` properties into a single `address` field for simplicity. This should make creating PostgreSQL captures much more pleasant in the web UI.

Merging this will require a bit of care to not break existing captures, but to my knowledge there's only one of those that's really important to not break.

**Workflow steps:**

The names and locations of various configuration values have changed, but the actual information is the same.

I'll send out a PR for https://docs.estuary.dev/reference/Connectors/capture-connectors/PostgreSQL/ once this is merged.

### DIFF
--- a/source-postgres/README.md
+++ b/source-postgres/README.md
@@ -70,10 +70,9 @@ A minimal `config.json` consists solely of the database connection parameters:
 
 ```json
 {
+  "address": "localhost:5432",
   "database": "flow",
-  "host": "localhost",
   "password": "flow_capture",
-  "port": 5432,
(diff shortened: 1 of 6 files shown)


willdonnelly:
Going to go ahead and merge this, since I just pinned the one production use to a specific hash and all our test captures under `estuary/` just got blown away, so nothing should be relying on this right now.

PR #267, opened 2022-06-15
source-mysql: Ignore INSERT/DELETE events from inactive tables

willdonnelly:
**Description:**

The `binlog_format=ROW` system variable can theoretically [1] be overridden on a per-session basis, and if this happens but only modifies tables that we don't care about capturing, then ideally we should just ignore those query events.

This PR modifies the handling of `INSERT` and `DELETE` statements in the binlog to only be an error if they impact tables that are currently active. It would be nice to also handle `UPDATE`s the same way, but those are more complicated and will require some additional work to make sure the logic is correct, so I wanted to handle the easy cases immediately and see if that helps with the specific failure we're observing.

[1] We have some weak evidence that this may be happening on Amazon RDS MySQL for updates to the `mysql.rds_heartbeat2` table, but even if it isn't this is always a thing that could potentially happen due to user action on the DB.

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -402,8 +402,21 @@ func (rs *mysqlReplicationStream) handleQuery(schema, query string) error {
 				return fmt.Errorf("unsupported operation RENAME TABLE on stream %q (go.estuary.dev/eVVwet)", streamID)
 			}
 		}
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #270, opened 2022-06-17
source-mysql: Sanity-check RDS binlog retention too

willdonnelly:
**Description:**

This adds a query of the `mysql.rds_configuration` table where `name = 'binlog retention hours'`. If the query succeeds then this parameter takes priority over the standard MySQL system variables when doing the binlog retention sanity-check.


psFried (approved):
LGTM

> there's an RDS-specific configuration

:roll_eyes: _of course_ there is

PR #271, opened 2022-06-17
source-mysql: Ignore 'COMMIT' query events in the binlog

willdonnelly:
**Description:**

Most simple operations don't produce these, but it's been observed at least once and they should be ignored rather than halting the capture.


psFried (approved):
LGTM

PR #275, opened 2022-06-20
source-mysql: Fix handling of case-sensitive table names

willdonnelly:
**Description:**

It turns out that, at least in some setups, MySQL uses case-sensitive comparison for table names. In general we try to normalize everything to lowercase within the connector for comparison purposes, but this lowercase-normalization was leaking out into the backfill queries sent to MySQL and causing them to fail.

This has been fixed in the generic `sqlcapture` logic so that now backfill queries will be generated using the schema+name info coming from table discovery, which retains the original casing as returned by the target DB. In addition, the `source-mysql` test helper `CreateTable()` has been tweaked so that generated table names are no longer all lowercase, which means that every test case now exercises this behavior, and also that the table names included in capture output retain the original case.

There is one rough spot remaining, which is that because of our widespread use of lower-case normalization for table names we can't actually support capturing from "case colliding" tables, and in fact the existence of such tables currently breaks discovery and means we can't capture _anything_. But this is only a problem if somebody decides they want to have `mydata.foobar` and `mydata.FooBar` exist and be distinct tables in their database, so it's kind of an edge case and I don't think it's worth the effort it would take to fix at the moment.

**Workflow steps:**

None.

None.


psFried (approved):
LGTM

PR #277, opened 2022-06-21
source-mysql: Use SSL when connecting to MySQL

willdonnelly:
**Description:**

This commit modifies `source-mysql` to use SSL for its connections to the MySQL database. I kind of figured that was the default, but apparently not.

This is done for both the normal/backfill connection and the replication connection. I'm not sure if there needs to be some sort of fallback to not-SSL if SSL connections fail, but my inclination is to just require SSL if possible, and if we see that approach failing in the future we can add fallback logic.

Currently the 'InsecureSkipVerify' TLS configuration is set to true, which means that there's no specific server name which the
SSL certificate is required to match. This means that an active MitM could theoretically intercept the connection (which is why it's called "Insecure" in the option name), but in many cases (such as right now, when I'm running MySQL locally for testing) there may not *be* a server name to verify, so for now I'm skipping that. In the future we could add a `serverName` or `sslServerName` config option, and if set then we could verify that particular name against the certificate.

**Workflow steps:**

Ideally this shouldn't change anything user-visible, except now connections won't fail when the database requires SSL.

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


willdonnelly:
Going to merge this one unreviewed so that Dave can move forward with testing it on whatever cloud provider (Azure?) was having SSL-related issues. Since `SELECT * FROM live_specs WHERE connector_image_name = 'ghcr.io/estuary/source-mysql'` in Supabase indicates that nothing is actually using `source-mysql` in production yet this should be fine.

PR #278, opened 2022-06-27
Add Schema Inference to Kinesis, S3, and GCS Connectors

saterus:
**Description:**

Kinesis, S3, and GCS do not provide any facilities to introspect on the data in its streams/files. This leads to a lackluster user experience, as the discovered bindings all have an empty json schema. This means to get the most value out of Flow, the user needs to author a json schema from scratch. This is less than ideal.

Instead, we'll have these connectors use the `flow-schema-inference` tool to generate simple schemas from the data we read from the streams themselves. Not all of these will be maximally useful, but it's a great start for a user just trying the system. A more advanced user can always edit these schemas with more useful annotations.

**Workflow steps:** (Kinesis example)

1. Login to the AWS CLI.
1. Create a Kinesis stream
1. Seed that stream with some documents
1. Open the UI
1. Select AWS Kinesis as your source
1. Enter your AWS credentials
1. Generate your catalog
1. Admire the what a nice schema you have

Alternatively, you can use the CLI:
```bash
flowctl-go api discover --image=ghcr.io/estuary/source-kinesis:v1 --config=source-kinesis/kinesis-config.json | jq
```

**Picture Worth At Least 1000 words**:

Screenshot 2022-06-27 at 17-23-16 Flow · Create Capture

None. This is an entirely behind the scenes enhancement. Using the Kinesis, S3, and GCS connectors should just get a little better for users.

**Notes for reviewers:**

N/A

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/278)

### DIFF
--- a/filesource/discovery.go
+++ b/filesource/discovery.go
@@ -64,7 +64,7 @@ func discoverSchema(ctx context.Context, conn *connector, streamName string) (sc
 		return DISCOVER_FALLBACK_SCHEMA, nil
 	}
 
-	logEntry.WithField("count", docCount).Info("Got documents for stream")
+	logEntry.WithField("documentCount", docCount).Info("Got documents for stream")
 
 	schema, err := schema_inference.Run(ctx, logEntry, docCh)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


psFried (changes requested):
I think this looks good. I noticed what seem to be a few minor issues, but this should be good to merge once those are addressed. This will be a _huge_ improvement to these connectors once it lands :+1:

PR #282, opened 2022-07-06
source-mysql: Add support for ENUM and SET column types

willdonnelly:
**Description:**

This change required some tweaks to the plumbing around column types. Previously the `DataType` of a column was represented as a string with values like `"int"` or `"text"`, but representing the possible values of an enum column required richer metadata. This was accomplished in a backwards-compatible way by turning the string into an `interface{}` value, and leaving the majority
of column types as a string. Thus the JSON-serialized type names included in state checkpoints are unchanged for other types.

Then a new `mysqlColumnType` struct was defined with an `EnumValues` array and a `Type` property. Currently this is only used for enum and set, but if any other types required richer metadata in the future the same struct can be extended. After deserializing table metadata from state checkpoints during startup, this struct will have gotten turned into a `map[string]interface{}` so a bit of additional logic uses the `mapstructure` package to turn that back into a usable struct.

(It should be noted that the `TestDatatypes` test cases already exercise the state checkpoint metadata round-tripping behavior
because of how they're written, so no new logic was added but that's still being checked).

There's also some new logic to parse a column type name such as `enum('foo', 'a,b,c', 'baz')` into `["foo", "a,b,c", "baz"]`. I'm still not entirely happy with how that part looks, but I believe it should handle all edge cases in practice.

In addition, discovery can now produce useful JSON-schemas for `ENUM` columns, but I don't believe there's a good way to do the same for `SET`, since the default MySQL representation for a set column's value is a comma-separated string like `"foo,bar,baz"` and so I've elected to convert the replicated bitfield version into that form as well.

**Workflow steps:**

Discovery and captures of MySQL tables with `ENUM` and `SET` columns should now work better. Specifically the discovered schema will be more useful and no longer contain a complaint about unrecognized types, and captured values will be useful strings instead of an inconsistent mix of human-readable strings (from backfills) and integer indices/bitfields (from replication).

None, to my knowledge.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/282)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"regexp"
+	"strings"
 
 	"github.com/estuary/connectors/sqlcapture"
 	"github.com/go-mysql-org/go-mysql/client"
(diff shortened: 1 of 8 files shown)


source-mysql/discovery.go
```diff
@@ -158,18 +180,107 @@ func getColumns(ctx context.Context, conn *client.Conn) ([]sqlcapture.ColumnInfo
...
+			return acc.String(), nil
+		} else if bs, ok := val.([]byte); ok {
+			return string(bs), nil
+		}
+		return val, nil
+	}
+	return val, fmt.Errorf("error translating value of complex column type %q", t.Type)
+}
+
+var enumValuesRegexp = regexp.MustCompile(`'((?:''|\\.|[^'])+)'(?:,|$)`)
```
psFried, line 245: I'm just trusting you on this one ;)

source-mysql/discovery.go
```diff
@@ -158,18 +180,107 @@ func getColumns(ctx context.Context, conn *client.Conn) ([]sqlcapture.ColumnInfo
...
+var enumValueReplacements = map[string]string{
```
psFried, line 246: nit: is this list documented somewhere where you can link to in a comment?

psFried (approved):
LGTM % the comment about nullable enum representations

source-mysql/discovery.go
willdonnelly: Yeah, sorry, it's one of those regexes that looks really complicated because it's made entirely of line noise, but I'm decently confident it's correct.

TL;DR: it matches a single-quoted string followed by either a comma or EOL, using non-capturing groups for the alternations on the terminator and the string body characters. The options for string body characters are, in order: "two successive quotes", "anything backslash-escaped", and "any character that's not a single-quote".

source-mysql/discovery.go
willdonnelly: Yeah, the list is just https://dev.mysql.com/doc/refman/8.0/en/string-literals.html#character-escape-sequences plus the `''` mechanism which is also described on that page. I'll add a comment mentioning that, and also one explaining what the regexp is supposed to do.

willdonnelly:
update: CI passed but today ended up being really busy with other stuff, so I'm not going to merge this today after all. This can wait until next week without any problems.

PR #285, opened 2022-07-11
source-postgres: Only error on TRUNCATE if table is active

willdonnelly:
**Description:**

Currently a `TRUNCATE` event in the Postgres WAL will halt the capture unconditionally, but it's only actually an error if it occurs on a table we're capturing. Fixing this avoids random breakage, and makes it possible for the user to remedy the issue themselves by removing a capture binding and then re-adding it, without having to teardown the entire capture.

**Workflow steps:**

None, it should just be a bit more reliable and produce a more useful error message when the user performs an unsupported `TRUNCATE` to a table we're capturing.

Maybe the PostgreSQL docs? Need to check still.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/285)

### DIFF
--- a/tests/materialize/flowctl/build-and-activate.sh
+++ b/tests/materialize/flowctl/build-and-activate.sh
@@ -8,9 +8,11 @@ flowctl api build \
   --directory "${TEST_DIR}"/builds \
   --build-id "${BUILD_ID}" \
   --source "file://${TEST_DIR}/${CATALOG}" \
+  --network host \
   --log.level debug
 
 flowctl api activate \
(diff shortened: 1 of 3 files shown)


mdibaiee (approved):
LGTM

mdibaiee:
@willdonnelly I just cherry-picked @saterus's fix for tests 🤞🏽

willdonnelly:
Thanks for that, I'll go ahead and merge this so we can see whether the capture issue was a real problem or a spurious one.

PR #287, opened 2022-07-15
Retry failed queries in materialize-bigquery

psFried:
**Description:**

Bigquery frequently errors out due to concurrent updates against the
same table. This commit adds retry behavior around query execution so
that the connector does not fail. Unfortunately, there doesn't seem to
be a great way to identify errors due to concurrent updates, and so we
might also retry other types of errors. So we limit the number of retries,
to avoid continuously retrying queries that may never succeed.

**Workflow steps:** n/a

**Notes for reviewers:**

Originally, I'd hoped to also include a change here to have individual shards each use a distinct partition in the `flow_checkpoints_v1` table. While I still think that's beneficial, it seems like it'll take a while to handle migrating live materializations to the new table definitions. And there's good reason to believe that just this retry change by itself will solve or significantly improve the issues we've been seeing.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/287)

### DIFF
--- a/materialize-bigquery/query.go
+++ b/materialize-bigquery/query.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"cloud.google.com/go/bigquery"
+	log "github.com/sirupsen/logrus"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 )
@@ -16,7 +17,6 @@ var (
 
 // query executes a query against BigQuery and returns the completed job.
 func (ep *Endpoint) query(ctx context.Context, queryString string, parameters ...interface{}) (*bigquery.Job, error) {
-
 	return ep.runQuery(ctx, ep.newQuery(queryString, parameters...))
 
 }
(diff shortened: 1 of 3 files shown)


materialize-bigquery/query.go
```diff
@@ -36,29 +36,69 @@ func (ep *Endpoint) newQuery(queryString string, parameters ...interface{}) *big
...
-	return job, nil
+		// I think this is just documenting the assumption that the job must always be Done after
+		// Wait returns. I don't think we should ever hit this condition.
+		if !status.Done() {
+			return job, fmt.Errorf("query not done")
+		}
 
+		return job, nil
+	}
+	panic("unreachable")
```
willdonnelly, line 101: This panic bothers me a little, and the retry loop logic feels like it could be cleaned up slightly by moving the successful-return check above the error handling and letting the `continue` to retry and the `attempt == maxAttempts` failure case naturally fall out of the looping. Something like:

```
for attempt := 0; attempt < maxAttempts; attempt++ {
  <run the query as before>
  if err == nil {
    <sanity-check status.Done here or prior to the err == nil check>
    return job, nil
  }
  if e, ok := err.(*googleapi.Error); ok {
    <terminal error checks>
  }
  logEntry.Warn("job failed (will retry)")
}
logEntry.Error("job failed")
return job, fmt.Errorf("exausted retries: %w", err)
```

materialize-bigquery/query.go
```diff
@@ -36,29 +36,69 @@ func (ep *Endpoint) newQuery(queryString string, parameters ...interface{}) *big
...
+					return nil, err // The error isn't considered retryable
+				}
+			}
+
+			if attempt == maxAttempts {
+				logEntry.Error("job failed")
+				return job, fmt.Errorf("exausted retries: %w", err)
+			} else {
+				logEntry.Warn("job failed (will retry)")
+				// There doesn't seem to be any benefit to waiting before retrying, at least not
```
willdonnelly, line 87: I would strongly recommend adding a little bit of randomized jitter (like maybe a random duration from 50-150ms) to break potential livelocks more quickly and space out retries slightly.

Even if you can't turn up any benefit in small-scale tests, it's simple enough to add now and vastly more likely to help than to hurt us in the future.

willdonnelly (approved):
LGTM % two nits

materialize-bigquery/query.go
psFried: yeah, I probably should have done that before putting up the pr in the first place. Good call out

materialize-bigquery/query.go
psFried: done

PR #291, opened 2022-07-21
materialize-elasticsearch: set default  value of `delta_updates` to `false`

mdibaiee:
**Description:**

- Set default value for `delta_updates` to `false`
- Upgrade from `alecthomas/jsonschema` which is long deprecated to `invopop/jsonschema`. The former did not support boolean `default` values and was causing issues for some connectors.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/291)


materialize-bigquery/bigquery.go
```diff
@@ -76,7 +57,7 @@ type tableConfig struct {
 	base *config
 
 	Table string `json:"table" jsonschema:"title=Table,description=Table in the BigQuery dataset to store materialized result in."`
-	Delta bool   `json:"delta_updates,omitempty" jsonschema:"default=true,title=Delta Update,description=Should updates to this table be done via delta updates. Defaults is false."`
+	Delta bool   `json:"delta_updates,omitempty" jsonschema:"default=false,title=Delta Update,description=Should updates to this table be done via delta updates. Defaults is false."`
```
psFried, line 60: Is this actually changing the default, or was this descriptive of the current behavior?

sqlcapture/discovery.go
```diff
@@ -75,45 +75,43 @@ func DiscoverCatalog(ctx context.Context, db Database) (*airbyte.Catalog, error)
...
-													"reduce": map[string]interface{}{
-														"strategy": "firstWriteWins",
-													},
+			AllOf: []*jsonschema.Schema{
+				{
+					Extras: map[string]interface{}{
+						"properties": map[string]*jsonschema.Schema{
+							"_meta": {
+								Type: "object",
+								Extras: map[string]interface{}{
```
psFried, line 84: It seems a little weird to use `Extras` for these instead of `Properties`. Is there some reason why `Properties` wouldn't work? Probably worth adding a comment, assuming this was intentional.

psFried (approved):
LGTM % the couple of inline comments.

materialize-bigquery/bigquery.go
mdibaiee: @psFried the `default` in JSONSchema was not working, and in Golang the default value for bool is `false`. The description also used to say that it defaults to false.

sqlcapture/discovery.go
mdibaiee: @psFried there is a comment above that explains why

PR #293, opened 2022-07-21
Phil/bq bucket validation

### LINKED ISSUE #292: BigQuery does not handle leading / trailing slashes correctly
`bucket_path: /` causes a silent failure.

Perhaps follow Firebolt in cleaning the path to always be without leading or trailing slashes: https://github.com/estuary/connectors/pull/212/files

psFried:
**Description:**

Fixes #292 

Previously, if a bigquery materialization had a bucket path that started
with a `/`, it would silently fail to materialize any data. This fixes
the object key generation to handle the presence of the /. This allows
configurations with `bucket_path: /` to work, since users seem inclined
to use / instead of leaving it empty.

Additionally, this makes `bucket_path` optional, instead of required. It was always allowed to be empty, so having it be required just seemed a little silly. There's also no way to specify an empty string via the UI, so this means you don't have to set it if you just want to use the root of the bucket.

**Workflow steps:** no change

The table in the  [Bigquery docs](https://docs.estuary.dev/reference/Connectors/materialization-connectors/BigQuery/) will need updated to reflect that `bucket_path` is no longer required.

**Notes for reviewers:** n/a

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/293)

### DIFF
--- a/materialize-bigquery/.snapshots/TestSpecification
+++ b/materialize-bigquery/.snapshots/TestSpecification
@@ -6,7 +6,6 @@
       "dataset",
       "region",
       "bucket",
-      "bucket_path",
       "credentials_json"
     ],
     "properties": {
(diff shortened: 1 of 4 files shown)


mdibaiee (approved):
Thanks Phil!

PR #297, opened 2022-07-26
materialize-postgres: use address instead of host and port

mdibaiee:
**Description:**

- resolves https://github.com/estuary/flow/issues/584

**Workflow steps:**

- Use `address: 127.0.0.1:5437` in materialize-postgres config instead of `host` and `port`

We will need to update https://docs.estuary.dev/reference/Connectors/materialization-connectors/PostgreSQL/ with this change

**Notes for reviewers:**

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/297)


mdibaiee:
@willdonnelly I think the hardest about here is releasing this. This is a breaking change. We have to come up with a process / method of releasing new connector versions.

In this example we might say it's possible to have backward compatible code that reads `host` and `port` and translates it to `address`, but I'm in favour of publishing this as `v2` and going through the migration process to learn how that would work. This is a simple enough change that should be a good example for us to work with.

willdonnelly:
This looks good to me, agreed that we ought to figure out at least a rough version of the connector versioning / release process soon.

One thing worth noting is that all three uses of `materialize-postgres` currently (`estuary/monitoring/postgres`, `estuary/monitoring/hubspot-postgres`, and `estuary/ops/task-stats/to-postgres`) are using `:dev`, which will (presumably) be updated regardless of any versioning/release process improvements. So the first step of releasing this change is probably to move those over to `:v1`.

mdibaiee:
I updated `estuary/monitoring/postgres`, `estuary/monitoring/hubspot-postgres`, and `estuary/ops/task-stats/to-postgres` to use `v1` tag.

willdonnelly (approved):
LGTM

PR #303, opened 2022-07-28
source-postgres: Timeout DB connection after 30 seconds

willdonnelly:
**Description:**

In the absence of this setting the connection will keep trying indefinitely, which in practice means "until the OS TCP connect timeout expires after 4 minutes". This in turn means that even when the target database is completely unreachable and Flow is running the connector in a crash loop, the majority of the time it will look like it's "working" in the UI because the most recent connector run has not yet failed.

I suspect 30 seconds is plenty of time for any non-pathological database setup to connect (and supporting this assertion: the MySQL client-library has a hard-coded 10s deadline already), but this could be made configurable if it ever becomes an issue.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/303)

### DIFF
--- a/source-postgres/helpers_test.go
+++ b/source-postgres/helpers_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/estuary/connectors/sqlcapture"
 	"github.com/jackc/pgconn"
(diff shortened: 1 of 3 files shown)


mdibaiee (approved):
Sounds reasonable to me 👍🏽

PR #306, opened 2022-08-01
SQL capture benchmarks and some throughput improvements

willdonnelly:
**Description:**

This PR adds `go test`-able benchmarks `BenchmarkBackfill` and `BenchmarkReplication` to `source-postgres` and `source-mysql`. These benchmarks take care of initializing the test database with a bunch of rows of synthetic data and then time a capture of that data. They do their best to exclude all of the necessary setup and teardown work from the actual benchmark timing, so that the reported `ns/op` measurements should accurately reflect the marginal cost of capturing a certain amount of data. See "Workflow steps" for more details on how to run those.

In addition, this PR includes several changes motivated by running these benchmarks in a setup with significant (~45ms) DB<->Capture latency:

- Optimizes the `source-postgres` handling of column descriptions during discovery. Previously this was a separate query run in a loop for every column of every discovered table, which added up and made things _very slow_ when run on large databases with nonzero ping. This speeds up connector startup against cloud-hosted DB instances by ~10 seconds, and has a very noticeable impact on test suite runs.
- Increases the backfill chunk size by a factor of 32x. This alone makes backfills substantially faster in setups with nontrivial latency.
- Performs JSON serialization and writing-to-stdout in a separate goroutine for state checkpoints and record output, with a substantial channel buffer acting as a work queue. This allows the primary backfill loop to immediately go start fetching the next chunk of rows, with serialization work taking place while we're waiting on more data from the network.
  - I believe that JSON serialization is still a significant bottleneck at present (primarily because at the end of the benchmark a noticeable amount of time is spent waiting for the worker thread to finish draining the queue), however optimizing this further is probably more effort than it's worth right now.

**Workflow steps:**

The new benchmarks can be invoked like:

```
$ LOG_LEVEL=warn go test -run=None -bench=. -benchtime=1000x -timeout=0 ./source-mysql/ --db_addr=<hostport> --db_password=<secret>
$ LOG_LEVEL=warn go test -run=None -bench=. -benchtime=1000x -timeout=0 ./source-postgres/ --test_connection_uri=postgres://flow_capture:<secret>@<hostport>/postgres
```

For these tests, the unit of work (what `go test` reports as `1 op`) is 1,000 rows of a synthetic dataset, divided evenly across 10 tables. The synthetic data weighs in at an average 192 bytes with significant variability (+/- 132 bytes), so the `12345678 ns/op` values reported by `go test` can be turned into a sustained throughput estimate like `192kB / 12345678 ns = 123 Mbps`.

**Notes for reviewers:**

I did a bit of cleaning-up of individual commits, but the overall sequence of work here is basically as I did it. Let me know if you'd prefer to have the capture optimizations pulled out into a separate PR from the test-refactoring and benchmark-writing part.

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/306)

### DIFF
--- a/source-mysql/throughput_test.go
+++ b/source-mysql/throughput_test.go
@@ -15,7 +15,6 @@ import (
 	"github.com/estuary/flow/go/protocols/airbyte"
 	"github.com/google/uuid"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/sync/errgroup"
 )
 
 // The benchmarks 'BenchmarkBackfill' and 'BenchmarkReplication' initialize
@@ -53,15 +52,10 @@ func benchmarkBackfills(b *testing.B, iterations, numTables, rowsPerTable int) {
 
 	var tb, ctx = TestBackend, context.Background()
 	var tables []string
-	var grp errgroup.Group
 	for i := 0; i < numTables; i++ {
 		var table = tb.CreateTable(ctx, b, fmt.Sprintf("table%d", i), "(id INTEGER PRIMARY KEY, uid TEXT, name TEXT, status INTEGER, modified DATE, foo_id INTEGER, padding TEXT)")
 		tables = append(tables, table)
-		//grp.Go(func() error { return populateTable(ctx, b, tb, table, rowsPerTable) })
 		populateTable(ctx, b, tb, table, rowsPerTable)
-	}
-	if err := grp.Wait(); err != nil {
-		b.Fatalf("error populating tables: %v", err)
 	}
 	var catalog = tests.ConfiguredCatalog(ctx, b, tb, tables...)
 	var dummyOutput = &benchmarkMessageOutput{Inner: json.NewEncoder(io.Discard)}
@@ -100,13 +94,10 @@ func benchmarkReplication(b *testing.B, iterations, numTables, rowsPerTable int)
 		b.Fatalf("capture failed with error: %v", err)
 	}
 
-	var grp errgroup.Group
 	for _, table := range tables {
 		var table = table
-		//grp.Go(func() error { return populateTable(ctx, b, tb, table, rowsPerTable) })
 		populateTable(ctx, b, tb, table, rowsPerTable)
 	}
-	grp.Wait()
 
 	for i := 0; i < iterations; i++ {
 		var state = tests.CopyState(initialState)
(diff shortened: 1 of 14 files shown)


source-mysql/throughput_test.go
```diff
@@ -0,0 +1,202 @@
...
+		"tables":       numTables,
+	}).Info("initializing tables")
+
+	var tb, ctx = TestBackend, context.Background()
+	var tables []string
+	var grp errgroup.Group
+	for i := 0; i < numTables; i++ {
+		var table = tb.CreateTable(ctx, b, fmt.Sprintf("table%d", i), "(id INTEGER PRIMARY KEY, uid TEXT, name TEXT, status INTEGER, modified DATE, foo_id INTEGER, padding TEXT)")
+		tables = append(tables, table)
+		//grp.Go(func() error { return populateTable(ctx, b, tb, table, rowsPerTable) })
```
williamhbaker, line 60: Residual comment?

source-mysql/throughput_test.go
```diff
@@ -0,0 +1,202 @@
...
+		populateTable(ctx, b, tb, table, rowsPerTable)
+	}
+	if err := grp.Wait(); err != nil {
```
williamhbaker, line 63: Depending on if the `grp.Go` above is supposed to be used or not, this may or may not be necessary.

source-mysql/throughput_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+	var initialState = sqlcapture.PersistentState{}
+	if err := sqlcapture.RunCapture(ctx, tb.GetDatabase(), &catalog, &initialState, dummyOutput); err != nil {
+		b.Fatalf("capture failed with error: %v", err)
+	}
+
+	var grp errgroup.Group
+	for _, table := range tables {
+		var table = table
+		//grp.Go(func() error { return populateTable(ctx, b, tb, table, rowsPerTable) })
```
williamhbaker, line 106: Same as in `benchmarkBackfills` about this being commented out.

psFried (approved):
Love it! I had the same question as @williamhbaker about the error groups being seemingly unused, but apart from that this looks good to merge.

source-mysql/throughput_test.go
willdonnelly: Good catch, thanks. I had tried parallelizing the MySQL dataset initialization like I did for Postgres (which doesn't impact the actual benchmark timing but makes the whole thing run faster) but ran into enough issues that I just backed that out and accepted that MySQL benchmark initialization would be slower. I'll go clean up the leftover bits of code from that.

williamhbaker (approved):

PR #308, opened 2022-08-02
source-firestore: first version

mdibaiee:
**Description:**

- A new source for capturing documents from Google Firestore
- A script for adding an updated_at field to all documents in some collections.

**Workflow steps:**

- Discover and deploy a capture with `source-firestore` docker image. 
  - Create a Firestore collectio
  - `flowctl-admin discover --image source-firestore:local --log.level=debug`
  - Set your `googleCredentials` in `acmeCo/source-firestore.config.yaml`
  - `flowctl-admin discover --image source-firestore:local --log.level=debug`
  - Run a temporary data plane: `flowctl-admin temp-data-plane --log.level=debug --network=host`
  - And now deploy `flowctl-admin deploy --log.level=debug --source acmeCo/source-firestore.flow.yaml --network=host`
  - Updating documents and adding new documents will lead to the documents being updated by Flow

- New capture, needs documentation.

**Notes for reviewers:**

- Depends on https://github.com/estuary/flow/pull/616

---
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/308)


source-firestore/main.go
```diff
@@ -0,0 +1,263 @@
...
+
+	var enc = airbyte.NewStdoutEncoder()
+	for {
+		var now = time.Now()
+		for _, stream := range catalog.Streams {
+			var streamName = stream.Stream.Name
+			var streamState = state.Snapshots[streamName]
+
+			var collection = client.Collection(streamName)
+			// FIXME: DocumentID is not ordered, so the user must have a field for us to order the documents on
```
mdibaiee, line 210: This is the main challenge: Firestore's default document identifier is not ordered, from their [docs](https://firebase.google.com/docs/firestore/manage-data/add-data):

> Important: Unlike "push IDs" in the Firebase Realtime Database, Cloud Firestore auto-generated IDs do not provide any automatic ordering. If you want to be able to order your documents by creation date, you should store a timestamp as a field in the documents.

So the user needs to provide us with a column which we can use for chronological ordering of documents.

Given the Airbyte protocol, I'm not sure how can we allow users to configure a cursor for each stream? I see [these two fields](https://github.com/estuary/flow/blob/master/go/protocols/airbyte/catalog.go#L21-L22) in the protocol definition, but I'm not sure how that's mapped from `resource_spec`, or is it at all?

If we can't make this configurable per stream, it needs to be universal across all collections which might be more difficult for users.

source-firestore/main.go
```diff
@@ -0,0 +1,411 @@
...
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+)
+
+type State struct {
+	// Last time a full table scan was run, we use this to know if we need to do a full scan
+	// to reach eventual consistency
+	LastScan string
```
willdonnelly, line 23: Could this be a `time.Time` and use the default JSON serialization/deserialization logic?

source-firestore/main.go
```diff
@@ -0,0 +1,411 @@
...
+	var docsCh = make(chan encDocument)
+
+	var eg = new(errgroup.Group)
+
+	for _, stream := range catalog.Streams {
+		var streamName = stream.Stream.Name
+
+		var collection = client.Collection(streamName)
+		eg.Go(func() error {
+			return fullScan(ctx, lastScan, scanInterval, collection, docsCh, catalog.Tail)
```
willdonnelly, line 251: If I'm understanding the code here correctly there's no synchronization between `listenForChanges` and `fullScan`, so it's possible in some cases [1] that `fullScan` might output an older version of some document _after_ a newer version was already output by `listenForChanges`. Does that sound correct? Do documents have some sort of timestamp/clock value such that this can be reconciled later on?

[1] For a concrete example, this seems like it could happen if the query results in `fullScan` represent a point-in-time snapshot of the dataset, or equivalently if there's enough buffering/latency between Firestore and the connector.

source-firestore/main.go
```diff
@@ -0,0 +1,411 @@
...
+	// to reach eventual consistency
+	LastScan string
+}
+
+type config struct {
+	// Service account JSON key to use as Application Default Credentials
+	CredentialsJSON string `json:"googleCredentials" jsonschema:"title=Credentials,description=Google Cloud Service Account JSON credentials." jsonschema_extras:"secret=true,multiline=true"`
+
+	// How frequently should we scan all collections to ensure consistency
+	ScanInterval string `json:"scan_interval" jsonschema:"title=Scan Interval,description=How frequently should all collections be scanned to ensure consistency. See https://pkg.go.dev/time#ParseDuration for supported values.,default=12h"`
```
willdonnelly, line 31: Would it make sense to allow some sort of "never do a full scan" value here (the obvious options being a zero duration or the literal string `"never"`)?

source-firestore/main.go
willdonnelly: Or is the idea that any particular document is unlikely to be constantly under contention, so at worst that document will be slightly stale until the next `fullScan` run some hours later?

willdonnelly (approved):
LGTM assuming there's a good answer to the "why is it safe to interleave fullScan and listenForChanges output" question

source-firestore/main.go
```diff
@@ -0,0 +1,411 @@
...
+	var lastScan = time.UnixMilli(0)
+	if state.LastScan != "" {
+		var err error
+		lastScan, err = time.Parse(time.RFC3339, state.LastScan)
+		if err != nil {
+			return fmt.Errorf("parsing last scan checkpoint failed: %w", err)
+		}
+	}
+
+	var scanInterval, _ = time.ParseDuration("12h")
```
williamhbaker, line 212: Nit: It would nice if the default `12h` value was declared as a constant, and maybe set in `(c *config) Validate()` if there isn't one supplied by the user. Took me a minute to figure out how the default was set.

source-firestore/main.go
```diff
@@ -0,0 +1,411 @@
...
+	return nil
+}
+
+func encodeDocs(docsCh chan encDocument) error {
+	var enc = airbyte.NewStdoutEncoder()
+
+	for {
+		encDoc := <-docsCh
+
+		if len(encDoc.doc) != 0 {
```
williamhbaker, line 280: Question for my own enlightenment's sake:  What happens if neither of these branches applies? Is such a thing possible? If I had a better grasp on what a doc of length 0 & state of length 0 represented I might suggest adding a log for this scenario, but I'm not sure if that would make sense or not.

source-firestore/main.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	// If we are not tailing, we close the channel after the full scan
+	// to allow the encoding routine to finish so the connector can exit
+	if !tail {
+		close(docsCh)
+		return nil
+	}
+
+	return fullScan(ctx, time.Now(), scanInterval, collection, docsCh, tail)
```
williamhbaker, line 410: Will it matter that the `time.Now()` here is going to be different than the `scanTime` recorded above? I'm guessing probably not, but it seems like it'll make the scan intervals not line up at exactly 12h (or whatever) increments.

source-firestore/main.go
mdibaiee: I thought about that, I just was not sure what the default value is in case of `time.Time` if it's not set? that's why I went with string and parsing it, and then setting the default to unix time 0

source-firestore/main.go
mdibaiee: That is a valid concern, I was not very worried about it since I didn't expect there to be a high rate of edits, but that assumption doesn't always hold.

I'm also not sure whether the query for documents is a point-in-time snapshot or not, if it wasn't the current approach would work, but I couldn't find any information on that from Firestore docs.

It seems tricky to me to ensure that the full scan does not output any documents that have already been outputted by the listener, mostly because it can consume a lot of memory to hold all the documents that are being updated if there are a lot of them, although this memory can be cleaned up once the full scan is over, but this means on larger datasets, we are likely to run into memory issues: the full scan takes longer, there are potentially more updates, and so we have a faster growing list of updated documents which we need to hold onto until the full scan is over. 🤔

source-firestore/main.go
mdibaiee: the literal string sounds good 👍🏽

source-firestore/main.go
mdibaiee: I refactored to use an enum for distinguishing between different requests to encoder

source-firestore/main.go
mdibaiee: Good catch!

source-firestore/main.go
mdibaiee: Apparently it is the zero time, so I guess it works! https://stackoverflow.com/questions/23051973/what-is-the-zero-value-for-time-time-in-go

williamhbaker (approved):

PR #315, opened 2022-08-17
source-http-file uses a maxBound in the future

psFried:
**Description:**

The filesource library uses an upper bound on the `Last-Modified`
timestamp during each sweep. This prevented the source-http-file
connector from ingesting files that are updated very frequently because
the `Last-Modified` timestamp would end up falling outside of the
`maxBound`. Additionally, the failure scenario that the `maxBound` is
meant to address in the first place isn't possible with source-http-file
since it only ever ingests a single file.

This commit updates `filesource.Source` to include the delta that's
applied to the current time in order to determine the `maxBound` for
each sweep. This lets source-s3 and source-gcs retain the existing
behavior, while allowing source-http-file to use a positive value so
that frequently updated files can be captured.

**Workflow steps:**

No changes, just use source-http-file with a frequently updating file, and see that it now works.

N/A - This change just makes the source-http-file connector work as expected in more cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/315)

### DIFF
--- a/source-http-file/main.go
+++ b/source-http-file/main.go
@@ -200,6 +200,13 @@ func main() {
       }`)
 		},
 		DocumentationURL: "https://go.estuary.dev/source-http-file",
+		// Set the delta to far in the future. The scenario that the the `MaxBound` protects agains
+		// is only possible when a listing can return multiple files, which is never the case for
+		// this connector. Setting it in the future allows this connector to fetch files where the
+		// server may set the `Last-Modified` header to the current time when the response is
+		// generated, which could otherwise fall after the `MaxBound`, which is computed prior to
+		// sending any requests.
+		TimeHorizonDelta: time.Hour,
 	}
 
 	src.Main()
--- a/filesource/filesource.go
+++ b/filesource/filesource.go
@@ -46,6 +46,11 @@ type Source struct {
 	Connect func(context.Context, Config) (Store, error)
 	// DocumentationURL links to the Source's extended user documentation.
 	DocumentationURL string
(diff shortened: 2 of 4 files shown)


source-http-file/main.go
```diff
@@ -200,6 +200,13 @@ func main() {
       }`)
 		},
 		DocumentationURL: "https://go.estuary.dev/source-http-file",
+		// Set the delta to far in the future. The scenario that the the `MaxBound` protects agains
```
jgraettinger, line 203: I think this should be zero, and not in-the-future.

If this were an hour in the future:
* Sweep zero starts at now0, and runs with a minBound of 0 and an upperbound of now0+1h.
* The file has a last-modified of now0-2s and is processed.
* The sweep completes, and the new minBound is now0+1h.
* Sweep one starts shortly after at now1, very slightly larger than now0.
* It runs with minBound of now0+1h, and maxBound of now1+1h.
* It sees a file with last-modified of now1-3s _which is outside of the bounds_.

tl;dr after the first sweep, I don't see how the file is processed again.

psFried: We don't update the minBound based on the maxBound, but rather based on the greatest modTime of the files actually processed during the sweep. This is done [here](https://github.com/estuary/connectors/blob/10e168b5bbf956561e0e614f309bd0791c5efff7/filesource/state.go#L111). So I _think_ we're good, but LMK if there's something I'm missing.

jgraettinger: Indeed! Yea I was wrong, I think this is good too.

jgraettinger (approved):
LGTM

PR #317, opened 2022-08-19
source-nativehello: Implement a native gRPC capture

willdonnelly:
**Description:**

This is a complete proof-of-concept capture connector, similar to `source-hello-world` but with a few tweaks that allowed me to show off the resource spec customization better and made it slightly simpler to implement.

This commit also includes a new `source-boilerplate` package which implements functions analagous to `materialize-boilerplate`. I may end up adding a bit more boilerplate because there's currently no equivalent to `RunTransactions()`, instead each capture has to manage receiving `PullRequest.Acknowledge` messages and sending documents/checkpoints for themselves.

Everything appears to work more or less as it should, but this has not really been battle-tested fully and it's possible that there could be some edge case in the connector lifecycle that doesn't work properly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/317)

### DIFF
--- a/source-boilerplate/boilerplate.go
+++ b/source-boilerplate/boilerplate.go
@@ -1,0 +1,230 @@
+package boilerplate
+
+import (
+	"bufio"
+	"context"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"os"
+	"os/signal"
+	"syscall"
+
+	pc "github.com/estuary/flow/go/protocols/capture"
+	protoio "github.com/gogo/protobuf/io"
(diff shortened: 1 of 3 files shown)


mdibaiee (approved):
LGTM! RunTransactions would definitely help for writing simpler connectors by elimating some of the boilerplate, but I rebased my firestore connector on top of this and I'm happy to use this as is

willdonnelly:
This was cherry-picked as part of https://github.com/estuary/connectors/pull/319

PR #319, opened 2022-08-19
source-firestore: support nested collections, use flow protocol

mdibaiee:
**Description:**

- Support nested collectons for source-firestore by discovering them using a wildcard notation:
```
captures:
  acmeCo/source-firestore:
    endpoint:
      connector:
        image: source-firestore:local
        config: source-firestore.config.yaml
    bindings:
      - resource:
          path: orgs/*/runs/*/runResults/*/queryResults
        target: acmeCo/orgs_runs_runResults_queryResults
      - resource:
          path: orgs/*/runs/*/runResults
        target: acmeCo/orgs_runs_runResults
      - resource:
          path: orgs/*/runs
        target: acmeCo/orgs_runs
      - resource:
          path: orgs
        target: acmeCo/orgs
```
- I realised while working on this PR that actually it is not possible for us to not fully scan Firestore whenever we boot up. The process of listening on changes of a query always includes a copy of all documents upfront: https://stackoverflow.com/questions/52955433/can-cloud-firestore-onsnapshot-only-trigger-on-changes-and-not-get-the-initia#
- As a result, I updated the code to remove the need for LastScan state: we scan once when we boot up, and then every `ScanInterval` after that as long as we are awake
- Refactored the code to use native Flow protocol, thanks to @willdonnelly's PR: https://github.com/estuary/connectors/pull/317
- The discovery process is quite complicated since we need to traverse in nested structures, that's the part that took most of the work
- Updated the `build-local.sh` script to work on M1 mac by using `docker buildx`

**Workflow steps:**

- ```flowctl discover --image source-firestore:local```

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/319)

### DIFF
--- a/source-firestore/discovery.go
+++ b/source-firestore/discovery.go
@@ -1,0 +1,201 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	firestore "cloud.google.com/go/firestore"
+	"github.com/estuary/connectors/schema_inference"
+	"github.com/estuary/flow/go/protocols/airbyte"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/api/iterator"
+)
+
+const DISCOVER_DOC_LIMIT = 50
+
+type inferenceRequest struct {
+	document   schema_inference.Document
+	collection string
+}
+
+type inferenceProcess struct {
+	docsCh chan schema_inference.Document
+	count  int
+	open   bool
+}
+
+// The collection name must not have slashes in it, otherwise we end up with parent
+// collections being a prefix of the child collections, which is prohibited by flow
+func collectionRecommendedName(name string) string {
+	return strings.ReplaceAll(name, "/*/", "_").ReplaceAll("/", "_")
+}
+
+// Start a channel where we receive documents for each collection
+// if they match an existing collection (e.g. group/*/subgroup), we send the docs to the
+// existing inference channel, otherwise we create a new inference channel for it.
+// Once we receive DISCOVER_DOC_LIMIT documents for each collection, we close the channel for inference
+// and stop writing to it.
+func discoveryRoutine(
+	ctx context.Context,
+	catalog *airbyte.Catalog,
+	eg *errgroup.Group,
+	ch chan inferenceRequest,
+	inferenceChannels map[string]*inferenceProcess,
+) error {
+	for request := range ch {
+		if process, exists := inferenceChannels[request.collection]; !exists {
+			log.WithField("collection", request.collection).Info("created a new inference channel")
+			var docsCh = make(chan schema_inference.Document)
+
+			var logEntry = log.WithField("collection", request.collection)
+			inferenceChannels[request.collection] = &inferenceProcess{
+				docsCh: docsCh,
+				count:  0,
+				open:   true,
+			}
+
+			eg.Go(func() error {
+				var collection = request.collection
+				var err error
+				var schema schema_inference.Schema
+				log.WithField("collection", collection).Info("schema_inference.Run")
+				schema, err = schema_inference.Run(ctx, logEntry, docsCh)
+				if err != nil {
+					return fmt.Errorf("schema inference: %w", err)
+				}
+
+				log.WithField("collection", collection).Info("finished schema inference")
+				catalog.Streams = append(catalog.Streams, airbyte.Stream{
+					Name:                    collection,
+					RecommendedName:         collectionRecommendedName(collection),
+					JSONSchema:              schema,
+					SupportedSyncModes:      []airbyte.SyncMode{airbyte.SyncModeIncremental},
+					SourceDefinedCursor:     true,
+					SourceDefinedPrimaryKey: [][]string{{firestore.DocumentID}},
+				})
+				return nil
+			})
+
+			docsCh <- request.document
+		} else {
+			if process.open {
+				log.WithField("collection", request.collection).WithField("count", process.count).Info("channel exists, sending document")
+				process.docsCh <- request.document
+				process.count = process.count + 1
+				if process.count > DISCOVER_DOC_LIMIT {
+					process.open = false
+					close(process.docsCh)
+				}
+			} else {
+				log.WithField("collection", request.collection).Info("channel is closed")
+			}
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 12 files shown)


source-firestore/discovery.go
```diff
@@ -0,0 +1,237 @@
...
+
+// Recursively discover collections
+func discoverCollection(
+	ctx context.Context,
+	ch chan inferenceRequest,
+	inferenceChannels map[string]*inferenceProcess,
+	response *pc.DiscoverResponse,
+	collection *firestore.CollectionRef,
+) error {
+	var process, exists = inferenceChannels[collectionToResourcePath(collection.Path)]
```
jgraettinger, line 188: there's a data race over `inferenceChannels`, since it can be concurrently mutated by `discoveryRoutine`.

There's also a race over `process.open` which is read here but concurrently modified elsewhere.

Can you refactor this to accept a context, which is cancelled by discoverRoutine when it's read enough from this collection?

source-firestore/helpers.go
```diff
@@ -0,0 +1,68 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+)
+
+const DEFAULT_SCAN_INTERVAL string = "12h"
```
jgraettinger, line 8: nit: defaultScanInterval is canonical Go. Not SNAKE_CASE.

source-firestore/pull.go
```diff
@@ -0,0 +1,322 @@
...
+
+// struct sent over channel to emitter
+type emitRequest struct {
+	requestType EmitterRequestType
+	arena       []byte
+	binding     uint32
+	state       []byte
+}
+
+func (driver) Pull(stream pc.Driver_PullServer) error {
```
jgraettinger, line 48: no action: a little more boilerplate would go a long way to simplify this further...

source-firestore/pull.go
```diff
@@ -0,0 +1,322 @@
...
+			break
+		} else if err != nil {
+			return err
+		}
+
+		for _, change := range snapshot.Changes {
+			if change.Kind == firestore.DocumentAdded || change.Kind == firestore.DocumentModified {
+
+				// Make sure this document matches the binding path
+				var index = findBindingForCollectionPath(change.Doc.Ref.Parent.Path, c.Bindings)
```
jgraettinger, line 206: is it possible that it wouldn't? I suppose an example would be nested collections not being captured?

Related question: to capture a nested collection, you MUST capture it's parent?

source-firestore/pull.go
```diff
@@ -0,0 +1,322 @@
...
+		}
+	}
+
+	// If we are not tailing, we close the channel after the full scan
+	// to allow the encoding routine to finish so the connector can exit
+	if !c.Tail {
+		return nil
+	}
+
+	return c.FullScan(ctx, scanTime, scanInterval, query)
```
jgraettinger, line 283: nit: this confused me for a moment; comment that we're iterating to the next scan using tail recursion?

source-firestore/pull.go
```diff
@@ -0,0 +1,322 @@
...
+			}
+			if err := c.Stream.Send(&pc.PullResponse{
+				Checkpoint: &pf.DriverCheckpoint{
+					DriverCheckpointJson: []byte("{}"),
+					Rfc7396MergePatch:    true,
+				},
+			}); err != nil {
+				return err
+			}
+		} else if emitDoc.requestType == StateUpdate {
```
jgraettinger, line 309: I'm not seeing where this is produced into the channel?

jgraettinger:
This is a great example of the rock vs hard place of connectors. Rust would offer a lot more "sleep well at night" confidence w.r.t. possible data races and whatnot, but a quick survey of the [ecosystem of firestore clients](https://crates.io/keywords/firestore) is not especially comforting.

source-firestore/pull.go
mdibaiee: You are right, it isn't anymore, I can remove this part 👍🏽

source-firestore/discovery.go
mdibaiee: Refactored, let me know if it makes sense now

source-firestore/pull.go
mdibaiee: Will already acknowledged that we can implement something similar to RunTransactions here as well

source-firestore/pull.go
mdibaiee: @jgraettinger yes, below I check for `index == -1` which is the case where the document doesn't match any of the bindings, in which case we just skip this document.

I haven't tested this, but no, you should not need to capture a parent collection to capture a nested collection thanks to [CollectionGroup](https://firebase.googleblog.com/2019/06/understanding-collection-group-queries.html) queries

source-firestore/discovery.go
jgraettinger: You still have a race over `inferenceChannels`, since discoveryRoutine adds new entries to it while this concurrently access it. Try building / running with --race detection?

source-firestore/pull.go
willdonnelly: I still intend to implement some better boilerplate around the Pull RPC stuff, but I'd prefer to get this PR merged with as few additional changes as possible and then figure out some better boilerplate later.

source-firestore/discovery.go
willdonnelly: I've just pushed a new commit to this PR which reimplements the discovery process in a way that seemed cleaner to me (a single goroutine is spawned to handle each _resource_ which is discovered, and is fed collections to process over a channel from the parent resource's goroutine which acts as a work queue).

There's still a map that gets accessed concurrently from various workers though, so I've added a mutex to guard that.

willdonnelly:
PTAL. I've modified the discovery logic considerably, so it's probably best to review the current state of things as a new piece of code and ignore the diffs from the prior version.

The new logic runs cleanly with `-race` and appears to discover all the resources it ought to (along with useful schemas) when run against the `helpful-kingdom` project.

During testing I encountered an elusive bug where schema inference would sometimes terminate with no output (but manually running the schema inference binary and feeding it the same input bytes worked fine, so it wasn't the input data per se). This does not appear to happen when running without `-race`, but I'm worried that it's still a possibility, so in the event of `schema_inference.Run()` returning an error that error is logged, but then a maximally-permissive placeholder schema is used for that collection, rather than allowing the error to take down the entire discovery process.

jgraettinger (approved):
LGTM (though only briefly skimmed)

Issue #314, opened 2022-08-17
source-postgres: Fix handling of case-sensitive column names


willdonnelly:
I previously went through the code and made sure that the handling of case-sensitive table names ought to work in all kinds of weird edge cases, but the equivalent work for _column names_ has never been done.

One specific situation where we know things can go wrong is when a table is created with a quoted uppercase column name. When this happens, Postgres will retain the original casing and _only a quoted column name_ can ever match that, since it appears that an unquoted but correctly-capitalized column name will get lowercased internally by Postgres. Thus, this error message:

```
unable to execute query \"SELECT * FROM public.trades ORDER BY (Symbol, Timestamp) LIMIT 4096;\":
ERROR: column \"timestamp\" does not exist (SQLSTATE 42703)
```

In this particular case we ended up with a tricky column name for `"Timestamp"` because it's a reserved word and so was automatically quoted by `materialize-postgres`, but I suspect that simply auto-quoting reserved words isn't the right solution because this same situation could pop up with any arbitrary name, in principle.

PR #324, opened 2022-08-24
source-postgres: Quote all column names in backfill queries

willdonnelly:
**Description:**

This change adds a test of several difficult column names which require quoting in PostgreSQL:

  * A primary key column named `table`, which is a reserved word and thus an error if not quoted.
  * A primary key column named `Meta/"wtf"/ID`, which contains two different special characters (and double quotes in particular, which require special handling) as well as capital letters.

The tricky column names are made primary keys, because we need to be able to give them *back* to Postgres when performing backfill queries, but the test also exercises discovery, and captures via backfill and replication, just to make sure the names are handled correctly throughout the capture.

This should fix https://github.com/estuary/connectors/issues/314

**Workflow steps:**

Column names containing special characters and reserved words should already have worked outside of the primary key, but now captures with such a column name in the primary key should work correctly too. And there's a test to make sure it keeps working.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/324)

### DIFF
--- a/source-postgres/backfill.go
+++ b/source-postgres/backfill.go
@@ -64,6 +64,14 @@ func (db *postgresDatabase) ScanTableChunk(ctx context.Context, info sqlcapture.
 	return events, nil
 }
 
+func quoteColumnNames(names []string) []string {
+	var out []string
+	for _, name := range names {
+		out = append(out, name)
+	}
+	return out
+}
+
 // WriteWatermark writes the provided string into the 'watermarks' table.
 func (db *postgresDatabase) WriteWatermark(ctx context.Context, watermark string) error {
 	logrus.WithField("watermark", watermark).Debug("writing watermark")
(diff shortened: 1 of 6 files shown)


source-postgres/backfill.go
```diff
@@ -64,6 +64,14 @@ func (db *postgresDatabase) ScanTableChunk(ctx context.Context, info sqlcapture.
 	return events, nil
 }
 
+func quoteColumnNames(names []string) []string {
+	var out []string
+	for _, name := range names {
+		out = append(out, name)
```
psFried, line 70: This doesn't seem to actually be quoting anything, just copying them to a new slice. But it also seems like it may be unused, so perhaps it just needs deleted.

psFried (approved):
LGTM % the one unused function nit

source-postgres/backfill.go
willdonnelly: Good catch! Yeah, this was dead code and I've deleted it now.

PR #325, opened 2022-08-26
boilerplate: Check for bufio.ErrBufferFull correctly

willdonnelly:
**Description:**

Previously the check was for `io.ErrShortBuffer`, which is not actually the error we get when the incoming message is too large for the buffered reader.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/325)


jgraettinger (commented):
can this be extracted into a function with an accompanying test?

willdonnelly:
Done, PTAL

psFried (approved):
LGTM

PR #329, opened 2022-08-29
source-firestore: The discovered key should be /__path__

willdonnelly:
**Description:**

Previously this was `/__name__`, which is basically just the final path element and thus is frequently something like `27` which is not unique, and this caused the obvious sorts of issues now that the whole nested collections thing is live.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/329)


williamhbaker (approved):
Seems good to me!

PR #330, opened 2022-08-30
materialize-google-sheets: Fix row/column limitations

willdonnelly:
**Description:**

This PR tweaks the startup behavior of the Google Sheets materialization in ways that I think make it more user-friendly, and in the process fixes the issues that led to materializations crashing when run with >26 fields or when they reached >1000 rows.

The row count limit was a weird one, because our use of `InsertRange` requests in the transactor _should have_ resulted in every new output row adding itself to the sheet. But in practice there's some trickery going on, and somehow it's still possible to hit row count limits when appending a new row to the very end of a full sheet. To fix this, we now maintain a "sentinel" row after all our data rows, identified by the value `{}` in the `flow_document` column.

The column limit was much simpler: by default a sheet has 26 columns, and so trying to write more values than this is an error and we needed to resize the sheet to match the number of columns we wanted to write.

Finally I've added a couple of minor UX improvements while I was touching related portions of the code. After these changes, the header row is now "Frozen" so it acts more header-y, and the first column (`flow_document`) is collapsed by default so it doesn't frighten off any users.

**Workflow steps:**

Materializations with more than 25 fields should now work, as should materializations of >1000 rows.

Also users may see column header names automagically getting filled in, since previously they were only written out when a nonexistent sheet was created by the connector.

**Notes for reviewers:**

Now that I've finished writing everything I'm second-guessing the choice to move horizontal-resizing and header-row-writing from the Apply handler to Transactions startup time. I think I still prefer it this way personally, it seems a bit cleaner to have Apply just create a very default-looking sheet where necessary and then do the same initialization at every startup regardless of whether it's a new sheet or a preexisting one, but I'd be happy to move it back if you think that would be better.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/330)

### DIFF
--- a/materialize-google-sheets/main.go
+++ b/materialize-google-sheets/main.go
@@ -307,6 +307,10 @@ func (driver) Transactions(stream pm.Driver_TransactionsServer) error {
 		return fmt.Errorf("recovering sheet states: %w", err)
 	}
 
+	if err := writeSheetSentinels(stream.Context(), svc, cfg.spreadsheetID(), states); err != nil {
+		return fmt.Errorf("writing sheet sentinels: %w", err)
+	}
+
 	bindings, err := buildTransactorBindings(
 		open.Open.Materialization.Bindings,
 		checkpoint.Round,
--- a/materialize-google-sheets/main.go
+++ b/materialize-google-sheets/main.go
@@ -238,6 +238,21 @@ func (driver) apply(ctx context.Context, req *pm.ApplyRequest, isDelete bool) (*
 						Rows:    []*sheets.RowData{{Values: headers}},
 					},
 				},
+				// Hide the first column, which contains Flow internal data
+				&sheets.Request{
+					UpdateDimensionProperties: &sheets.UpdateDimensionPropertiesRequest{
+						Range: &sheets.DimensionRange{
+							SheetId:    sheetID,
+							Dimension:  "COLUMNS",
+							StartIndex: 0,
+							EndIndex:   1,
+						},
+						Properties: &sheets.DimensionProperties{
+							HiddenByUser: true,
+						},
+						Fields: "hiddenByUser",
+					},
+				},
 			)
 		} else if exists && isDelete {
 			description += fmt.Sprintf("Deleted sheet %q.\n", res.Sheet)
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM!

A future question we'll need to answer is: what happens if the field selection changes as the materialization is evolved?

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 2 of 2 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/330#-:-NAp6oG83fdRJJ9sd2JJ:b-qrlsmg)* status: all files reviewed, 1 unresolved discussion (waiting on @psFried and @willdonnelly)

___
*[`materialize-google-sheets/util.go` line 157 at r3](https://reviewable.io/reviews/estuary/connectors/330#-NAp4VZ2ERuGNQX0f9lc:-NAp4VZ2ERuGNQX0f9ld:b-t4c9xg) ([raw file](https://github.com/estuary/connectors/blob/975c2b8da7f0d4b571bd398b29ed22c4fd6a24aa/materialize-google-sheets/util.go#L157)):*
> ```Go
> 				return nil, fmt.Errorf("row %d of range %s is not a string", rowInd, sheetRanges[sheetInd])
> 			} else if *uev.StringValue == sentinelValue {
> 				break // Sentinel value indicating the end of the materialization data
> ```

nit: sanity check that `rowInd` is indeed equal to `len(sheet.Data[0].RowData)-1` ?

willdonnelly:
> nit: sanity check that `rowInd` is indeed equal to `len(sheet.Data[0].RowData)-1` ?

I'm not sure if that sanity check is desirable. If we don't bother doing that check, the user can in theory put their own rollup calculations immediately below the materialization data, which is kind of neat for small use-cases. Like for instance having a couple of cells below some column computing its mean and standard deviation.

Whereas trying to be strict on this point gets us into a weird intermediate state where the user can still do that sort of thing but _only in the one row_ containing the sentinel (since we don't inspect the other cells of that row to make sure they're empty), so it seemed more consistent to just embrace "The sentinel row marks the end of the data we care about, anything below that is Not Our Problem" as the policy there.

PR #331, opened 2022-08-31
source-postgres: Provide a more useful 'slot in use' error

willdonnelly:
**Description:**

If the user creates two captures from the same Postgres database (and doesn't customize the replication slot name in the advanced settings) the second capture will fail because slot `flow_slot` is already in use.

This makes the error message a bit nicer in this situation, from:

    error starting replication: unable to start replication: ERROR:
    replication slot "flow_slot" is active for PID 9967 (SQLSTATE 55006)

To the slightly more user-friendly:

    error starting replication: another capture is already running against
    this database: replication slot "flow_slot" is active for PID 9967

We might want to add some text to https://docs.estuary.dev/reference/Connectors/capture-connectors/PostgreSQL/ discussing this situation and probably clarifying that if you actually *want* more than one distinct capture instance from the same DB they will need to have different replication slot names configured in the advanced settings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/331)

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"regexp"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM - I had this happen to me recently and a better message like this would have helped!

Issue #334, opened 2022-08-31
Update our Snowflake Connector Config


dyaffe:
**Problem**
With AWS Snowflake, at least in `US-WEST-2`, we are currently seeing an issue where we are unable to connect to snowflake with a URL in the format we try with which is:

https://CLIENTID.REGIONID.CLOUDPROVIDER.snowflakecloudcomputing.com

Their connection URL that works is:

https://CLIENTID.snowflake.computing.com

**Proposal**
It occurs to me that the best thing we could do from a users perspective might be to ask for it in the UI in the following format:

https://   USER INPUT HERE  .snowflakecomputing.com 

Then users could update whatever they see in their actual snowflake instance and we could possibly parse it (if required from there).

I could be totally wrong about the plan but this seems reasonable.

willdonnelly:
Purely on aesthetic grounds I would recommend going a step further and just asking for the full domain.

Conveniently the Snowflake client library we use accepts an optional `Host` config property which supersedes their "assemble a domain name from Region and Account and stuff" logic, so I believe that the whole mess of options could be ripped out entirely and replaced with just `Host`.

psFried:
The snowflake client also seems to have special handling for the `us-west-2` region, which makes me wonder whether the configuration specified `US-WEST-2` or `us-west-2`. @dyaffe are you able to confirm which one was in there?

dyaffe:
It looks like we tried it with capitals.  The draft ID was `07:37:2a:db:80:fb:a8:00`

williamhbaker:
I think it will probably work if it is tried with lower-case letters: `us-west-2`. That will cause the DNS logic in the snowflake client [here](https://github.com/snowflakedb/gosnowflake/blob/5da2ab2463b2c7e544722bc58defdb23397287d6/dsn.go#L99-L111) to make the "host" be `account` + `.snowflakecomputing.com`, which I think lines up with what you said works in the original issue (`https://clientid.snowflake.computing.com/` - assuming there's an extra `.` in there and it should actually be `https://clientid.snowflakecomputing.com/`)

Seems like `us-west-2` is special somehow in snowflake's system and the URL for us-west-2 shouldn't contain the region. I can only speculate as to why that would be.

I'll try this out myself and see if I can repro the issue.

williamhbaker:
[...]

PR #335, opened 2022-09-01
materialize-snowflake: add host to config & remove region

williamhbaker:
**Description:**

This fixes the issues seen in #334 when materializing into snowflake on AWS providers. The host for AWS providers is not consistently any combination of account/region/provider for AWS (see [list](https://docs.snowflake.com/en/user-guide/admin-account-identifier.html#non-vps-account-locator-formats-by-cloud-platform-and-region)), so we can't reliably build a URL just from those values.

Fortunately, the snowflake client allows explicitly specifying a "host" value, which bypasses all of the idiosyncrasies of building the host from account/region/provider. This also allows for connecting with the [preferred account identifier format](https://docs.snowflake.com/en/user-guide/admin-account-identifier.html#format-1-preferred-account-name-in-your-organization), which uses the account name and org name, and is easily obtainable from the snowflake UI.

**Workflow steps:**

I was able to materialize into my snowflake account on `us-west-2` using a materialization schema like the following:

```
  examples/<materialization name>:
    endpoint:
        connector:
            config:
              host: <orgname>-<accountname>.snowflakecomputing.com
              account: <accountname>
              user: <my user>
              password: <my user's password>
              database: <database>
              schema: <schema>
              warehouse: <warehouse>
            image: materialize-snowflake:local
    bindings:
      - resource:
          table: <table>
        source: examples/<source collection>
```

https://github.com/estuary/flow/blob/master/site/docs/reference/Connectors/materialization-connectors/Snowflake.md will need updated according to this change to show the new/changed configuration values.

**Notes for reviewers:**

- Cloud provider is removed as a configuration since it is not needed or used at all.
- Region is also removed. Providing a region + host results in invalid URLs from the go-snowflake client. The region parameter is noted as being deprecated in the [snowflake client godocs](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Connection_Parameters).
- A value for account is still required, which I find interesting. But the client will throw an error if "account" is not provided, so we'll keep it as a required configuration value here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/335)

### DIFF
--- a/materialize-snowflake/.snapshots/TestSpecification
+++ b/materialize-snowflake/.snapshots/TestSpecification
@@ -8,6 +8,11 @@
         "title": "Account",
         "description": "The Snowflake account identifier."
       },
+      "host": {
+        "type": "string",
+        "title": "Host URL",
+        "description": "The Snowflake Host used for the connection. Example: orgname-accountname.snowflakecomputing.com (do not include the protocol)."
+      },
       "user": {
         "type": "string",
         "title": "User",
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

PR #336, opened 2022-09-06
new google pubsub materialization

williamhbaker:
Closes https://github.com/estuary/connectors/issues/326

**Description:**

Adds a google pubsub materialization connector.

**Workflow steps:**

This is a new connector. It works with authentication provided with service account credentials, or OAuth 2.0 flow through the UI.

It can be built locally using something like:
```
docker buildx build --platform linux/amd64 --load -t materialize-google-pubsub:local -f materialize-google-pubsub/Dockerfile .
```

If interested, it can be added to the UI of a local development instance of flow, using [this script](https://github.com/estuary/animated-carnival/blob/main/scripts/install-connector-local.sh) from `animated carnival`:
```
./install-connector-local.sh materialize-google-pubsub
```

New docs will need to be created for this connector.

**Notes for reviewers:**

This is pretty light on tests.

For integration tests, there is a [PubSub emulator](https://cloud.google.com/pubsub/docs/emulator) available, and it is possible to easily run this via docker following the instructions in [this blog post](https://medium.com/google-cloud/things-i-wish-i-knew-about-pub-sub-part-3-b8947b49224b), but I couldn't figure out a good way to thread the [environment variables](https://cloud.google.com/pubsub/docs/emulator#env) through for the client running in the connector container - the connector itself would need to be run with the environment variables for testing.

Setting up an actual PubSub topic accessible to our CI also doesn't seem like it would work very well due to the nature of PubSub. Messages are removed when they are ack'd by default - or perhaps not if a [retention policy](https://cloud.google.com/pubsub/docs/replay-overview#configuring_message_retention) is set. Either way seems like it would be really finicky to verify the output if multiple tests were running simultaneously, or if redeliveries occurred since PubSub is "at least once", etc.

The materialization mechanism is really very simple (just sends all the docs it gets to the topic), so I didn't push harder to get a full integration test running. Everything works as expected based on my manual testing.

To enable OAuth 2.0 in production, we'll need to create a new client ID & secret with the PubSub scopes and add those values to the production connectors table.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/336)


go.mod
```diff
@@ -17,14 +17,14 @@ require (
 	github.com/bradleyjkemp/cupaloy v2.3.0+incompatible
 	github.com/brianvoe/gofakeit/v6 v6.16.0
 	github.com/elastic/go-elasticsearch/v7 v7.17.1
-	github.com/estuary/flow v0.1.1-0.20220817221148-d68d91a3b3c1
+	github.com/estuary/flow v0.1.1-0.20220902132733-57dc21d8f214
```
williamhbaker, line 20: todo: update this pin when https://github.com/estuary/flow/pull/659 merges

go-auth/google/config.go
```diff
@@ -0,0 +1,156 @@
+package google_auth
```
williamhbaker, line 1: I debated internally over where to put this. It should be able to be re-used amongst google connectors (sheets, etc.) for authorization with service account & oauth options, so it seemed like having it somewhere it could be accessible to multiple connectors would make sense. In a go monorepo I'd put it in `pkg` but since that's not the case here this is the best I could come up with. Very open to alternatives though.

willdonnelly (approved):
LGTM

go-auth/google/spec.go
```diff
@@ -0,0 +1,24 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	pf "github.com/estuary/flow/go/protocols/flow"
+)
+
+func Spec(scopes ...string) *pf.OAuth2Spec {
+	authUrlTemplate := fmt.Sprintf(
+		"https://accounts.google.com/o/oauth2/auth?access_type=offline&prompt=consent&client_id={{ client_id }}&redirect_uri={{ redirect_uri }}&response_type=code&scope=%s&state={{ state }}",
```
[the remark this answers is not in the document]
williamhbaker: Somewhat surprisingly, the auth URL does not work with those swapped. It might have something to do with the way our stack uses the format strings, since I can't think of a reason why the order of query parameters would matter. I brought the format string out as a constant though, but left those values where they were.

PR #340, opened 2022-09-15
materialize-google-pubsub: enable multiple bindings to multiplex on topics

williamhbaker:
Related to https://github.com/estuary/flow/issues/673, specific to the PubSub materialization

**Description:**

Adds an additional optional field `identifier` to the resource config. This allows for configurations that can materialize multiple collections into the same topic. The `ResourcePath` for each binding must be unique, and providing unique `identifier`s per resource allows for `ResourcePath`s like `mytopic.identifier1`, `mytopic.identifier2` etc.

**Workflow steps:**

Configurations like this will now be possible, where two collections are materialized into the same PubSub topic:

```yaml
materializations:
  bobCo/multi-pubsub:
    endpoint:
      connector:
        image: "materialize-google-pubsub:local"
        config: ../flow/pubsub.config.yaml
    bindings:
      - resource:
          identifier: one
          topic: multiplex-topic
        source: bobCo/greetings1
      - resource:
          identifier: two
          topic: multiplex-topic
        source: bobCo/greetings2
```

The identifier is optional, so binding configurations like this are also possible:

```yaml
    bindings:
      - resource:
          identifier: one
          topic: multiplex-topic
        source: bobCo/greetings1
      - resource:
          identifier: two
          topic: multiplex-topic
        source: bobCo/greetings2
      - resource:
          topic: single-topic
        source: bobCo/greetings1
```

A configuration like this results in a validation error, since the bindings do not all have a unique combination of identifier + topic:
```yaml
    bindings:
      - resource:
          identifier: one
          topic: multiplex-topic
        source: bobCo/greetings1
      - resource:
          identifier: one # Error - identifier + topic are not unique
          topic: multiplex-topic
        source: bobCo/greetings2
      - resource:
          topic: single-topic
        source: bobCo/greetings1
      - resource:
          topic: single-topic # Also error
        source: bobCo/greetings1
```

The identifier is added to the published message as an "attribute", with the key of `identifier`. This will allow consumers to differentiate messages based on this attribute, and could also enable PubSub specific features like [filtering](https://cloud.google.com/pubsub/docs/filtering).

The [PubSub connector docs](https://docs.estuary.dev/reference/Connectors/materialization-connectors/google-pubsub/) will need to have this additional field added to them.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/340)

### DIFF
--- a/materialize-google-pubsub/transactor.go
+++ b/materialize-google-pubsub/transactor.go
@@ -3,6 +3,7 @@ package connector
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	"cloud.google.com/go/pubsub"
 	pf "github.com/estuary/flow/go/protocols/flow"
@@ -11,7 +12,13 @@ import (
 )
 
 type transactor struct {
-	topics []*pubsub.Topic
+	bindings []*topicBinding
+}
+
+type topicBinding struct {
+	identifier string
+	topicName  string
+	topic      *pubsub.Topic
 }
 
 // PubSub is delta-update only.
@@ -27,9 +34,11 @@ func (t *transactor) Store(it *pm.StoreIterator) error {
 	errGroup, ctx := errgroup.WithContext(it.Context())
 
 	for it.Next() {
-		topic := t.topics[it.Binding]
-		res := topic.Publish(ctx, &pubsub.Message{
-			Data:        it.RawJSON,
+		binding := t.bindings[it.Binding]
+		res := binding.topic.Publish(ctx, &pubsub.Message{
+			Data: it.RawJSON,
+			Attributes: map[string]string{
+				IDENTIFIER_ATTRIBUTE_KEY: fmt.Sprintf("%s/%s", binding.topicName, binding.identifier)},
 			OrderingKey: it.Key.String(), // Allows for reading of messages for the same key in order.
 		})
 
@@ -61,8 +70,8 @@ func (t *transactor) Acknowledge(context.Context) error {
 }
 
 func (t *transactor) Destroy() {
-	for _, topic := range t.topics {
+	for _, b := range t.bindings {
 		// Wait for all async messages to finished sending for each topic.
-		topic.Stop()
+		b.topic.Stop()
 	}
 }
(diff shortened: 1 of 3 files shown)


materialize-google-pubsub/transactor.go
```diff
@@ -27,9 +34,11 @@ func (t *transactor) Store(it *pm.StoreIterator) error {
...
 
 	for it.Next() {
-		topic := t.topics[it.Binding]
-		res := topic.Publish(ctx, &pubsub.Message{
-			Data:        it.RawJSON,
+		binding := t.bindings[it.Binding]
+		res := binding.topic.Publish(ctx, &pubsub.Message{
+			Data: it.RawJSON,
+			Attributes: map[string]string{
+				IDENTIFIER_ATTRIBUTE_KEY: fmt.Sprintf("%s/%s", binding.topicName, binding.identifier)},
```
willdonnelly, line 41: Would it be possible to have the default identifier value be `""` and then only include the identifier attribute if it's non-empty? That seems like it might be preferable for the simple use-cases.

materialize-google-pubsub/transactor.go
```diff
@@ -11,7 +12,13 @@ import (
 )
 
 type transactor struct {
-	topics []*pubsub.Topic
+	bindings []*topicBinding
+}
+
+type topicBinding struct {
+	identifier string
```
willdonnelly, line 19: Are the `identifier` and `topicName` fields used separately anywhere? If not it might be more efficient to move the whole `fmt.Sprintf` bit out into setup and just use the same string each time.

willdonnelly (approved):
LGTM although I'm not sold on the UX of having a `"$TOPICNAME/default"` attribute in cases where the user isn't using this multiple-bindings functionality.

materialize-google-pubsub/transactor.go
williamhbaker: Great idea, I updated the code to not include the attribute if the identifier is empty.

materialize-google-pubsub/transactor.go
williamhbaker: I thought about this some more and concluded that the `topicName` is totally redundant to include in the attribute to the topic, since all messages published to a topic will have the same topic name. So I was able to simplify this an remove the string formatting entirely.

Issue #327, opened 2022-08-16
Add OAuth to Google Sheets Materializations


dyaffe:
**Problem**
We currently support OAuth for Google Sheets on the captures side.  We should also add it to materializations.

travjenkins:
Moving over from UI to Connectors as that seems more appropriate.

PR #343, opened 2022-09-16
materialize-google-sheets: add oauth as authentication option

williamhbaker:
Closes https://github.com/estuary/connectors/issues/327

**Description:**

Adds OAuth 2.0 as an authentication option for the google sheets materialization connector.

This is not a backwards compatible change since it changes the shape of the schema around authentication credentials, so before merging this I will update any existing materializations that are not using the `v1` tagged connector image.

We'll also need a project set up with the `https://www.googleapis.com/auth/spreadsheets` scope for this to work.

**Workflow steps:**

Usable via the UI.

The sheets materialization docs will need updated according with this change.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/343)


mdibaiee (approved):
LGTM, thank you!

PR #344, opened 2022-09-19
source-s3: get object with context

williamhbaker:
**Description:**

Uses the passed-in context for `source-s3` `(*s3Store).Read`. This way the retrieving of objects can respect context timeouts.

Most notably, there is a timeout specified for [schema discovery](https://github.com/estuary/connectors/blob/353d119f03a004c6777fae7df67ffdd5b8f33c22/filesource/discovery.go#L111-L119), and without this timeout, it is possible that a single extremely large file will take an excessive amount of time to process.

**Workflow steps:**

**Notes for reviewers:**

Along these same lines, the `List` method doesn't use the provided context. Updating `List` to use a context would be a bit more complicated. I created a separate issue for that work, which isn't causing any immediate problems but might be good to take care of at some point: https://github.com/estuary/connectors/issues/345

Also, while investigating discovery from relatively large files, I found our parsing & _possibly_ inference processes surprisingly slow. If the parsing and inference were much faster, this particular context deadline would be less important, although still useful to have. I have some details to still work out for this, and have created a placeholder for an issue related to this potential optimization: https://github.com/estuary/flow/issues/681

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/344)

### DIFF
--- a/source-s3/main.go
+++ b/source-s3/main.go
@@ -112,7 +112,7 @@ func (s *s3Store) List(_ context.Context, query filesource.Query) (filesource.Li
 	}, nil
 }
 
-func (s *s3Store) Read(_ context.Context, obj filesource.ObjectInfo) (io.ReadCloser, filesource.ObjectInfo, error) {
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

Issue #341, opened 2022-09-16
source-kinesis flaky tests


williamhbaker:
I have seen test flakes in source-kinesis a couple of times. One example: https://github.com/estuary/connectors/actions/runs/3067998999/jobs/4954928763

The relevant message from the logs:

```
							      >	{
							      >	  "$schema": "http://json-schema.org/draft-07/schema#",
							      >	  "type": "object",
							      >	  "required": [
							      >	    "canary",
							      >	    "foo",
							      >	    "id"
Discovered bindings are wrong
							      >	  ],
							      >	  "properties": {
							      >	    "canary": {
							      >	      "type": "string"
							      >	    },
time="2022-09-16T12:58:33Z" level=info msg="caught signal" signal=terminated
							      >	    "foo": {
							      >	      "type": [
							      >	        "boolean",
							      >	        "integer",
							      >	        "string"
							      >	      ]
							      >	    },
							      >	    "id": {
							      >	      "type": "integer"
							      >	    }
							      >	  }
							      >	}
```

Re-running the test results in success.

PR #349, opened 2022-09-21
schema-inference: fix race condition

williamhbaker:
Closes https://github.com/estuary/connectors/issues/341

**Description:**

This fixes a race condition in handling the stdin/out pipes from running schema inference. It has been consistently causing the `source-kinesis` test to fail in an intermittent way. See [recent example](https://github.com/estuary/connectors/actions/runs/3094509636/jobs/5008268197).

Why is only the kinesis test failing? Apparently only `source-firestore` and `source-kinesis` use this code path, and `source-firestore` doesn't have an integration test that verifies bindings like the `source-kinesis` test. Although this change is primarily motivated by the desire to stop the test from flaking, I would imagine this could cause intermittent issues with discovery for both of these connectors in production.

The relevant line from the logs of failing tests is `"failed to decode schema, read |0: file already closed\"`. This arises from the inference code trying to read from a `stdout` that has already been closed by the concurrent call to `cmd.Wait`.

**Workflow steps:**

Run discovery with either `source-kinesis` or `source-firestore`.

N/A

**Notes for reviewers:**

The PR commit history illustrates forcing the failure to occur and then the fix:
- https://github.com/estuary/connectors/pull/349/commits/052bd84b4909b2c4da4621810db3ab071de98c6b adds a simulated delay to the existing code to make the failure inevitable
- https://github.com/estuary/connectors/pull/349/commits/5f2a6bf290f696cbc8bb89186a5721b0190d5835 fixes the race condition. The tests pass even with the artificial delay in place.
- https://github.com/estuary/connectors/pull/349/commits/3463fbed2e1e1dcb149212b591244be682fb6fc5 removes the artificial delay with the fix remaining in place.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/349)

### DIFF
--- a/schema_inference/inference.go
+++ b/schema_inference/inference.go
@@ -57,7 +57,6 @@ func Run(ctx context.Context, logEntry *log.Entry, docsCh <-chan Document) (Sche
 	})
 
 	errGroup.Go(func() error {
-		defer stdout.Close()
 		decoder := json.NewDecoder(stdout)
 
 		// Simulate a delay to reliably reproduce the race condition - the async call to cmd.Wait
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM.

Thanks for this, I had noticed Firestore discovery failing occasionally and I suspect this was the issue.

Issue #345, opened 2022-09-19
source-s3: object listing does not use context


williamhbaker:
In https://github.com/estuary/connectors/pull/344, the `source-s3` connector was updated to use the passed-in context for reading objects.

Object listing also does not use a context, and may be susceptible to extremely large folder listings taking an excessive amount of time, and not being governed by any context timeouts.

This issue is for updating the `source-s3` connector to use the passed-in context for listing.

PR #350, opened 2022-09-21
source-s3: use context for file listing

williamhbaker:
Closes https://github.com/estuary/connectors/issues/345

**Description:**

Threads the context through to the `source-s3` connector `List` operation. https://github.com/estuary/connectors/pull/344 accomplished the same thing for `Read`. This will allow for better handling of context cancellations or timeouts.

**Workflow steps:**

Should be minimal observable impact. This will eliminate the potential for very large and/or slow file listings to lock up schema discovery indefinitely.

N/A

**Notes for reviewers:**

I had initially thought this would take more work to get done, but it turned out to be pretty simple with the `filesource.ListingFunc` adapter already available. There wasn't much possible improvement I could see from refactoring the existing iterator code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/350)

### DIFF
--- a/source-s3/main.go
+++ b/source-s3/main.go
@@ -172,13 +172,13 @@ func (l *s3Listing) Next() (filesource.ObjectInfo, error) {
 			}, nil
 		}
 
-		if err := l.poll(l.ctx); err != nil {
+		if err := l.poll(); err != nil {
 			return filesource.ObjectInfo{}, err
 		}
 	}
 }
 
-func (l *s3Listing) poll(ctx context.Context) error {
+func (l *s3Listing) poll() error {
 	var input = l.input
 
 	if l.output.Prefix == nil {
@@ -190,7 +190,7 @@ func (l *s3Listing) poll(ctx context.Context) error {
 		input.ContinuationToken = l.output.NextContinuationToken
 	}
 
-	if out, err := l.s3.ListObjectsV2WithContext(ctx, &input); err != nil {
+	if out, err := l.s3.ListObjectsV2WithContext(l.ctx, &input); err != nil {
 		return err
 	} else {
 		l.output = *out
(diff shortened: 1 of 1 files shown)


source-s3/main.go
```diff
@@ -106,10 +106,14 @@ func (s *s3Store) List(_ context.Context, query filesource.Query) (filesource.Li
...
 	}
 
-	return &s3Listing{
+	l := &s3Listing{
 		s3:    s.s3,
 		input: input,
-	}, nil
+	}
+
+	return filesource.ListingFunc(func() (filesource.ObjectInfo, error) {
```
willdonnelly, line 114: It feels a little bit weird to construct an `s3Listing` iterator and then wrap it in a closure. Perhaps add a `ctx` field to `s3Listing` instead?

source-s3/main.go
```diff
@@ -188,7 +192,7 @@ func (l *s3Listing) poll() error {
 		input.ContinuationToken = l.output.NextContinuationToken
 	}
 
-	if out, err := l.s3.ListObjectsV2(&input); err != nil {
+	if out, err := l.s3.ListObjectsV2WithContext(context.Background(), &input); err != nil {
```
willdonnelly, line 195: It feels like this was intended to use `ctx` rather than `context.Background()`?

willdonnelly (approved):
LGTM % two nits

source-s3/main.go
williamhbaker: Totally! Thanks.

source-s3/main.go
williamhbaker: Yeah I was trying perhaps a little too hard to not add a context field to a struct, but in this case it is probably more clear, and [not that big of a deal anyway](https://github.com/golang/go/issues/22602). Updated!

willdonnelly: Yeah, I agree with that guidance in general but IMO storing the context in a closure is morally equivalent to storing it in the struct proper.

The way to actually keep the spirit of those guidelines would be to refactor the listing iterator interface so it's `.Next(ctx)`, but I didn't recommend that because it looks like other filesource implementations wouldn't be able to use that (for example, see `source-gcs` which takes the context from `List()` and makes that part of its `*storage.ObjectIterator`)

PR #354, opened 2022-09-23
source-gcs: make ascendingKeys an advanced option

williamhbaker:
**Description:**

The source-gcs `ascendingKeys` configuration option requires careful thought if it is to be used. This configuration could easily have unintended consequences (as we have observed), so this makes it an advanced configuration.

**Workflow steps:**

Compare the `endpointSpecSchema` of the old vs. the new, as reported by `flowctl-admin api spec --image <...>` (note: parser spec omitted for clarity):

Old:
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "title": "GCS Source",
    "type": "object",
    "required": ["bucket"],
    "properties": {
      "ascendingKeys": {
        "title": "Ascending Keys",
        "description": "Improve sync speeds by listing files from the end of the last sync, rather than listing the entire bucket prefix. This requires that you write objects in ascending lexicographic order, such as an RFC-3339 timestamp, so that key ordering matches modification time ordering.",
        "default": false,
        "type": "boolean"
      },
      "bucket": {
        "title": "Bucket",
        "description": "Name of the Google Cloud Storage bucket",
        "type": "string"
      },
      "googleCredentials": {
        "title": "Google Service Account",
        "description": "Service account JSON key to use as Application Default Credentials",
        "type": "string",
        "multiline": true,
        "secret": true
      },
      "matchKeys": {
        "title": "Match Keys",
        "description": "Filter applied to all object keys under the prefix. If provided, only objects whose key (relative to the prefix) matches this regex will be read. For example, you can use \".*\\.json\" to only capture json files.",
        "type": "string",
        "format": "regex"
      },
      "parser": {},
      "prefix": {
        "title": "Prefix",
        "description": "Prefix within the bucket to capture from",
        "type": "string"
      }
    }
  }
```

New:
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "title": "GCS Source",
    "type": "object",
    "required": ["bucket"],
    "properties": {
      "advanced": {
        "description": "Options for advanced users. You should not typically need to modify these.",
        "type": "object",
        "properties": {
          "ascendingKeys": {
            "title": "Ascending Keys",
            "description": "Improve sync speeds by listing files from the end of the last sync, rather than listing the entire bucket prefix. This requires that you write objects in ascending lexicographic order, such as an RFC-3339 timestamp, so that key ordering matches modification time ordering.",
            "default": false,
            "type": "boolean"
          }
        },
        "additionalProperties": false,
        "advanced": true
      },
      "bucket": {
        "title": "Bucket",
        "description": "Name of the Google Cloud Storage bucket",
        "type": "string"
      },
      "googleCredentials": {
        "title": "Google Service Account",
        "description": "Service account JSON key to use as Application Default Credentials",
        "type": "string",
        "multiline": true,
        "secret": true
      },
      "matchKeys": {
        "title": "Match Keys",
        "description": "Filter applied to all object keys under the prefix. If provided, only objects whose key (relative to the prefix) matches this regex will be read. For example, you can use \".*\\.json\" to only capture json files.",
        "type": "string",
        "format": "regex"
      },
      "parser": {},
      "prefix": {
        "title": "Prefix",
        "description": "Prefix within the bucket to capture from",
        "type": "string"
      }
    }
  }
```

I also verified that it shows up & works correctly on a local stack.

GCS capture docs will need updated: https://docs.estuary.dev/reference/Connectors/capture-connectors/gcs/

**Notes for reviewers:**


source-gcs/main.go
```diff
@@ -169,6 +171,20 @@ func main() {
 				"title":       "Prefix",
 				"description": "Prefix within the bucket to capture from"
 			},
+			"advanced": {
+				"properties": {
+				  "ascendingKeys": {
+					"type":        "boolean",
+					"title":       "Ascending Keys",
+					"description": "Improve sync speeds by listing files from the end of the last sync, rather than listing the entire bucket prefix. This requires that you write objects in ascending lexicographic order, such as an RFC-3339 timestamp, so that key ordering matches modification time ordering.",
```
willdonnelly, line 179: This description is really long, and yet it still doesn't communicate everything I'd like it to. Maybe we should add a "For more information see go.estuary.dev/ascending-keys" short link or something?

(For instance I'd like to include a cautionary note about only using this if you're responsible for generating the data and can guarantee that files will always be uploaded strictly in order, and give an explicit example like "If you upload files 'X', 'Z', and 'Y' in that order, it is likely that 'Y' will never be captured).

willdonnelly (approved):
LGTM

As a meta-note though, maybe we should do this to `source-s3` at the same time?

mdibaiee (approved):

Issue #360, opened 2022-09-29
Better Ordering for Connector Input Fields


williamhbaker:
Support for the `order` field was added to the UI in https://github.com/estuary/ui/pull/342.

We should add this field to connector schemas where it makes sense, as the default lexicographic ordering is not great in many cases.

PR #361, opened 2022-09-29
better field ordering for connectors

williamhbaker:
Closes https://github.com/estuary/connectors/issues/360

**Description:**

Adds ordering properties to fields in various connectors. The specific ordering is what seemed reasonable to me, and I'd be open to alternative suggestions. My general thought process for ordering things was to put them in this order:

1. Basic connection information, like a database URL
2. Authentication, in the form of user/password or service account credentials
3. The remaining connector configurations, ordered logically where possible, with "optional" fields generally coming last

The poster child for this algorithm is the configuration for the Postgres connector, which will appear in the order of:
```
address
user
password
database
```

The configs don't always map to that baseline perfectly, so I tried to get as close as possible.

**Workflow steps:**

Build the connector images, update the database records, and view the configs in the UI.

As I have initially proposed it:
- `source-s3` for the removal of the default value for "region" and the suggestion to try us-east-1 if you're not sure
- `source-kinesis` for the removal of the default values for "region", "awsAccessKeyId", and "awsSecretAccessKey"
- `materialize-bigquery` for a small change in the description of "billing_project_id"

**Notes for reviewers:**

It may be interesting to note which connectors I did not update:
- `materialize-google-pubsub`, `materialize-google-sheets`, and `materialize-rockset` only have a single configuration value so there is no need to specify ordering
- `source-firestore` has some work going on with it already so I didn't want to create merge conflicts; this one can be updated later
- `source-kafka` already has ordering

Also see inline comments for specific notes about things I have changed not directly related to ordering.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/361)

### DIFF
--- a/materialize-bigquery/bigquery.go
+++ b/materialize-bigquery/bigquery.go
@@ -23,13 +23,13 @@ func main() {
 
 // Config represents the endpoint configuration for BigQuery.
 type config struct {
-	BillingProjectID string `json:"billing_project_id,omitempty" jsonschema:"title=Billing Project ID,description=Billing Project ID connected to the BigQuery dataset. It can be the same value as Project ID."`
-	ProjectID        string `json:"project_id" jsonschema:"title=Project ID,description=Google Cloud Project ID that owns the BigQuery dataset."`
-	Dataset          string `json:"dataset" jsonschema:"title=Dataset,description=BigQuery dataset that will be used to store the materialization output."`
-	Region           string `json:"region" jsonschema:"title=Region,description=Region where both the Bucket and the BigQuery dataset is located. They both need to be within the same region."`
-	Bucket           string `json:"bucket" jsonschema:"title=Bucket,description=Google Cloud Storage bucket that is going to be used to store specfications & temporary data before merging into BigQuery."`
-	BucketPath       string `json:"bucket_path,omitempty" jsonschema:"title=Bucket Path,description=A prefix that will be used to store objects to Google Cloud Storage's bucket."`
-	CredentialsJSON  string `json:"credentials_json" jsonschema:"title=Credentials,description=Google Cloud Service Account JSON credentials in base64 format." jsonschema_extras:"secret=true,multiline=true"`
+	ProjectID        string `json:"project_id" jsonschema:"title=Project ID,description=Google Cloud Project ID that owns the BigQuery dataset." jsonschema_extras:"order=0"`
+	CredentialsJSON  string `json:"credentials_json" jsonschema:"title=Credentials,description=Google Cloud Service Account JSON credentials in base64 format." jsonschema_extras:"secret=true,multiline=true,order=1"`
+	Region           string `json:"region" jsonschema:"title=Region,description=Region where both the Bucket and the BigQuery dataset is located. They both need to be within the same region." jsonschema_extras:"order=2"`
+	Dataset          string `json:"dataset" jsonschema:"title=Dataset,description=BigQuery dataset that will be used to store the materialization output." jsonschema_extras:"order=3"`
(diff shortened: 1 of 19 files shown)


materialize-bigquery/bigquery.go
```diff
@@ -23,13 +23,13 @@ func main() {
...
+	Region           string `json:"region" jsonschema:"title=Region,description=Region where both the Bucket and the BigQuery dataset is located. They both need to be within the same region." jsonschema_extras:"order=2"`
+	Dataset          string `json:"dataset" jsonschema:"title=Dataset,description=BigQuery dataset that will be used to store the materialization output." jsonschema_extras:"order=3"`
+	Bucket           string `json:"bucket" jsonschema:"title=Bucket,description=Google Cloud Storage bucket that is going to be used to store specfications & temporary data before merging into BigQuery." jsonschema_extras:"order=4"`
+	BucketPath       string `json:"bucket_path,omitempty" jsonschema:"title=Bucket Path,description=A prefix that will be used to store objects to Google Cloud Storage's bucket." jsonschema_extras:"order=5"`
+	BillingProjectID string `json:"billing_project_id,omitempty" jsonschema:"title=Billing Project ID,description=Billing Project ID connected to the BigQuery dataset. Defaults to Project ID if not specified." jsonschema_extras:"order=6"`
```
williamhbaker, line 32: Small update in the wording here for `BillingProjectID`

source-kinesis/connection.go
```diff
@@ -43,30 +43,31 @@ var configJSONSchema = `{
...
-			"type":        "string",
-			"title":       "AWS Endpoint",
-			"description": "The AWS endpoint URI to connect to, useful if you're capturing from a kinesis-compatible API that isn't provided by AWS"
-		},
 		"awsAccessKeyId": {
 			"type":        "string",
 			"title":       "AWS Access Key ID",
 			"description": "Part of the AWS credentials that will be used to connect to Kinesis",
-			"default":     "example-aws-access-key-id",
-			"secret": true
```
williamhbaker, line 62: This doesn't need to be a secret. Also it doesn't seem helpful to have a default value here since `example-aws-access-key-id` is never going to be the correct value.

source-kinesis/connection.go
```diff
@@ -43,30 +43,31 @@ var configJSONSchema = `{
...
+			"order": 0
+
 		},
 		"awsSecretAccessKey": {
 			"type":        "string",
 			"title":       "AWS Secret Access Key",
 			"description": "Part of the AWS credentials that will be used to connect to Kinesis",
-			"default":     "example-aws-secret-access-key",
```
williamhbaker, line 68: Doubly unhelpful to have a default value for a secret field.

source-s3/main.go
```diff
@@ -223,35 +223,40 @@ func main() {
...
+				"secret":      true,
+				"order":       1
 			},
-			"bucket": {
+			"region": {
 				"type":        "string",
-				"title":       "Bucket",
-				"description": "Name of the S3 bucket"
+				"title":       "AWS Region",
+				"description": "The name of the AWS region where the S3 bucket is located.",
```
williamhbaker, line 239: Might not be easy to see from the diff, but this used to have a default value of `us-east-1` and a suggestion to try using `us-east-1` if you are unsure of what region to use. I find it hard to believe that if a user does not know which region they are capturing from, that blindly trying `us-east-1` is going to be a robust strategy. So I re-worded the description and removed the default.

jgraettinger (approved):
LGTM

Skimmed, but I'm agreed with your comments below.
This looks like it was annoying to do. Thank you

Issue #339, opened 2022-09-12
Overhaul Firestore capture memory usage


willdonnelly:
A little over a week ago we ran into some trouble when we tried to capture a real user's dataset with `source-firestore`. The immediate issue was that the reactor pods on which the capture was scheduled started getting OOM-killed. I spent some time last week reproducing the issue locally, and the problem is pretty straightforward. The solution is less so, but we'll get to that.

First the empirical observations. Heavily summarizing a couple days worth of investigation here:

1. While capturing a test dataset, I observe network RX around 12MBps for a couple minutes, during which time memory grows linearly up to ~3.2GiB.
2. Once the download completes there's an immediate spike to ~5.5GiB as the data is processed, and then documents start being emitted.
3. The production dataset which caused OOM issues is around 3x the size of the test data, and this memory behavior appears proportional to the amount of data being synced, so as a very rough estimate the capture was probably trying to use ~16.5GiB, which I think entirely explains the OOMs we observed.

Here's what's happening:

- When you ask Firestore to send you a stream of changes you get these changes as a sequence of consistent "snapshots", and the first snapshot you receive contains the initial state of the dataset.
- Furthermore, the Firestore client library manages this all internally and the first `.Next()` operation doesn't return until the entire initial state is synced.
- This means that your entire dataset (or at least whatever portion of your dataset is matched by a particular query) is going to be held in RAM during this initial sync. And this initial sync happens every time the connector restarts.
- Internally, the Firestore client library accumulates changes into a map keyed by document path, and then after it's all caught up this map is transformed into a list of document changes. This copying approximately doubles worst-case memory usage.

As far as I can tell there is no alternative to this behavior so long as we're going through the Firestore client library. It _might_ be possible to do better by directly using the [gRPC API](https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1#google.firestore.v1.Firestore.Listen) instead, but there are still a lot of uncertainties there.

PR #362, opened 2022-09-29
source-firestore: Big gRPC rewrite

willdonnelly:
**Description:**

This PR consists of:

1. A new implementation of capture logic using the Firestore gRPC `Listen` API directly rather than going through the client library.

    - Doing this was the primary goal of the work, as the client library buffers all documents in memory during the initial sync and we need to...not do that.
    - Since Flow will in turn buffer our output documents until a checkpoint, this implies that we need to emit checkpoints _even before we've reached a consistent point as defined by the server_. In practice we just go ahead and emit an empty checkpoint after every document.
    - Thus the capture can potentially emit duplicate documents now. This was kind of always the case, but we should probably mention that fact in the documentation and call out how collection reduction semantics address this?

2. A whole bunch of test logic, because at this point I'm a bit obsessed with asking "how do I know the output of this capture connector is actually correct in all the possible edge cases?", and ~600 lines of test code and a bunch of snapshot files was the bare minimum I felt comfortable having.

    - Roughly half of the new test code (everything in `capture_test.go`) is really a generic framework for testing `Pull` and `Discover` operations on a gRPC capture. I expect this will get hoisted out into a common package somewhere.

3. Yet Another Rewrite of the discovery process. The new implementation appears to be just as correct as the old one in practice, but it runs *much faster* by using Collection Group queries and processing documents in a tight loop, with sub-collection checking (which requires one RPC per document checked) performed in async workers and using better heuristics to limit the amount of that checking work we do.

4. Also some changes to document metadata. Now the path and update timestamp and deletion flag are all stored under `/_meta`, the suggested schema from discovery has been tweaked to reflect this (and in fact uses reflection from a `documentMetadata` struct), and also we specify a reduction strategy on the top level document of "maximize by `/_meta/mtime`".

This fixes https://github.com/estuary/connectors/issues/339

**Workflow steps:**

Connector usage should be pretty much the same as before, except that the polling interval stuff has been removed because we're relying solely on the "first sync the initial state of the dataset" behavior of the `Listen` RPC.

Probably https://docs.estuary.dev/reference/Connectors/capture-connectors/google-firestore/ but I need to think more about that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/362)


willdonnelly:
I believe this is ready for review now.

source-firestore/pull.go
```diff
@@ -54,261 +39,561 @@ func (driver) Pull(stream pc.Driver_PullServer) error {
...
+}
 
-	var eg = new(errgroup.Group)
+func (s *captureState) BindingIndex(resourcePath string) (uint32, bool) {
+	s.RLock()
+	defer s.RUnlock()
+	if state := s.Resources[resourcePath]; state != nil {
+		return state.bindingIndex, true
+	}
+	return ^uint32(0), false
```
mdibaiee, line 158: Took me a while to understand what this does, could use a comment to clarify it returns the maximum value of uint32

mdibaiee (commented):
Thanks Will! This looks _much better_ than the first version I wrote 😁

source-firestore/pull.go
```diff
@@ -54,261 +39,561 @@ func (driver) Pull(stream pc.Driver_PullServer) error {
...
+		// Listen streams can be a bit over-broad. For instance if there are
+		// collections 'users/*/docs' and 'groups/*/docs' in the database, but
+		// only 'users/*/docs' is captured, we need to ignore any documents
+		// from paths like 'groups/*/docs' which don't map to any binding.
 		return nil
+	} else if docJSON, err := json.Marshal(fields); err != nil {
+		return fmt.Errorf("error serializing document %q: %w", doc.Name, err)
+	} else if err := c.Output.Documents(bindingIndex, docJSON); err != nil {
+		return err
+	} else if err := c.Output.Checkpoint(json.RawMessage(emptyCheckpoint), true); err != nil {
```
williamhbaker, line 517: Does this ever end up "voiding" previous non-empty checkpoints? Might be part of the necessary design, just wanting to make sure I understand correctly.

source-firestore/discovery.go
```diff
@@ -4,48 +4,92 @@ import (
...
-	discoverMaxConcurrentWorkers = 32
+	// discoverSubresourceMinimum and discoverSubresourceProbability work together
+	// to limit the number of documents which we query for subcollections. The logic
+	// here is that if some collection (let's call it "users/*/messages" for example)
+	// has a subcollection ("users/*/messages/*/attachments") we don't need to check
+	// every single message for every user to learn this fact. Furthermore since every
+	// document we check requires a distinct RPC, this could take quite a while to do.
+	//
+	// So instead we can apply some common-sense policy which mimics what a user might
+	// do: check the first N exhaustively and then start skimming through and randomly
```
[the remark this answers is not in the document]
willdonnelly: The "first N" part of the heuristic is less about sequential ordering and more about making sure that when a collection only has a few documents every one of them will be checked for possible nested collections.

Implementing "last N" is trickier because we don't know upfront how many documents are in a collection, so we'd have to do some extra work to maintain a list of the last N documents observed from each collection and then launch some workers to check them for sub-collections only after reaching the end of the collection group. Which would definitely work, but in practice it seems like the IDs in most collections are random (and thus non-sequential) unique IDs anyway, so I'm not sure it's worth the extra effort compared to the "randomly sample P% of documents" approach.

(To be clear just in case you missed it, note that we're still feeding every document we receive into schema inference, the activity that this heuristic governs is solely the "does this document have any nested collections" check since doing that for every single document turned out to be the slow part of discovery.)

source-firestore/pull.go
willdonnelly: Done

source-firestore/main.go
```diff
@@ -23,20 +23,18 @@ type config struct {
 	// Service account JSON key to use as Application Default Credentials
 	CredentialsJSON string `json:"googleCredentials" jsonschema:"title=Credentials,description=Google Cloud Service Account JSON credentials." jsonschema_extras:"secret=true,multiline=true"`
 
-	// How frequently should we scan all collections to ensure consistency
-	ScanInterval string `json:"scan_interval" jsonschema:"title=Scan Interval,description=How frequently should all collections be scanned to ensure consistency. See https://pkg.go.dev/time#ParseDuration for supported values. To turn off scans use the value 'never'.,default=12h"`
+	// Optional name of the database to capture from
+	DatabasePath string `json:"database" jsonschema:"title=Database,description=Optional name of the database to capture from. Leave blank to autodetect. Typically \"projects/$PROJECTID/databases/(default)\"."`
```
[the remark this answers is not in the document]
willdonnelly: Good catch, done.

source-firestore/helpers.go
```diff
@@ -66,3 +65,10 @@ func collectionToResourcePath(collection string) string {
 
 	return strings.Trim(cleanedPath, "/*/")
 }
+
+func documentToResourcePath(documentPath string) string {
```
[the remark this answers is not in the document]
willdonnelly: Just that I missed it when I was adding the type aliases. Done.

source-firestore/pull.go
willdonnelly: No, the `true` argument there says "this is a patch update, to be merged according to https://www.rfc-editor.org/rfc/rfc7396 semantics", and the empty checkpoint is the identity value under merge semantics. So in effect this just says "hey Flow here's a checkpoint so you can finish the transaction whenever it's convenient" without actually modifying the state contents.

source-firestore/helpers.go
```diff
@@ -33,8 +32,8 @@ func findBindingForCollectionPath(collectionPath string, bindings []resource) in
 // which operate on the name of the last collection. Note that this name might not be unique
 // so another check is necessary to make sure we do not mix documents from nested collections
 // that have the same ID, but different parent collections
-func getLastCollectionGroupID(resourcePath string) string {
-	var pieces = strings.Split(resourcePath, "/")
+func getLastCollectionGroupID(path resourcePath) collectionGroupID {
+	var pieces = strings.Split(path, "/")
```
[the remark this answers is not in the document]
willdonnelly: Good catch. I hadn't really looked at `helpers.go` much except as necessary for other refactoring, but yeah I think those should be cleaned up as dead code for now.

source-firestore/pull.go
williamhbaker: Ah thanks! Makes sense.

jgraettinger:
FYI I'm removing myself as a reviewer, I think we're good with the folks who've reviewed already.

mdibaiee (approved):

PR #363, opened 2022-09-29
schema-inference: Fix another race condition

willdonnelly:
**Description:**

More precisely this fixes a race condition which appears to have been introduced by the previous race condition fix. After that fix, depending on scheduling it's possible for the schema-decoding worker to read a complete JSON object and terminate, after which point the errgroup context is canceled [1], and because the command was run with `exec.CommandContext` using that same context, the whole `cmd.Wait()` thing returns `context.Canceled`.

I'm fairly confident that this doesn't actually matter, since at this point we've got our result and this is just shutdown behavior, so I think a decent fix is simply to check for `context.Canceled` from `cmd.Wait()` and not treat that as a failure.

[1] From the docs of `errgroup.WithContext`: "The derived Context is canceled the first time a function passed to Go returns a non-nil error **or the first time Wait returns**, whichever occurs first."


williamhbaker (approved):
Nice!

PR #365, opened 2022-09-30
improve end-to-end test compatibility

williamhbaker:
**Description:**

Some adjustments to the end-to-end tests that allow them to more easily be run on non-linux systems (mac):
- Instead of the test script itself specifically downloading `flow` binaries, the test environment must make them available on `$PATH`. As long as the runner has `flowctl-admin` via `$PATH`, the tests can be run.
- The materialization tests were adjusted to use the `flow` binaries directly instead of running the temp data plane inside of container. This seemed to make things much simpler, and is consistent with how the capture tests run. I'm not 100% sure that the tests couldn't be run on mac with this docker container strategy, but I found it much easier to just run the binaries directly.

**Workflow steps:**

These tests should now be runnable on linux and mac, by invoking the test scripts with the appropriate environment variables.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/365)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -106,8 +106,10 @@ jobs:
           VERSION=$(cat ${{ matrix.connector }}/VERSION | tr -d '\n')
           echo ::set-output name=version::${VERSION}
 
-      - name: Download latest Flow release binaries
-        run: ./fetch-flow.sh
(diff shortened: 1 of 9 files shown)


mdibaiee (approved):
LGTM, thanks!

PR #367, opened 2022-10-03
[wip] network-tunnel library client

mdibaiee:
**Description:**

**Workflow steps:**

**Notes for reviewers:**


tests/materialize/materialize-postgres-rc/setup.sh
```diff
@@ -3,12 +3,13 @@
...
+export PGHOST="${PGHOST:=materialize-postgres-rc-postgres-1.flow-test}"
 export PGPORT="${PGPORT:=5432}"
-export PGDATABASE="${PGDATABASE:=postgres}"
-export PGPASSWORD="${PGPASSWORD:=postgres}"
-export PGUSER="${PGUSER:=postgres}"
+export PGDATABASE="${PGDATABASE:=flow}"
+export PGPASSWORD="${PGPASSWORD:=flow}"
+export PGUSER="${PGUSER:=flow}"
 
+docker compose -f materialize-postgres-rc/docker-compose.yaml up --detach
```
williamhbaker, line 12: Any reason to have this here as opposed to setting it up in github action?

mdibaiee: I just thought it makes sense that each setup.sh file sets up its environment as much as possible (although sometimes earlier CI steps need the service to be up), and in case of running tests locally, it makes things simpler, you just have to do
```
CONNECTOR=x VERSION=y ./tests/run.sh
```

without worrying about what services need to be run, etc.

tests/source-mysql/setup.sh
```diff
@@ -4,10 +4,19 @@ set -e
...
 export RESOURCE="{ \"namespace\": \"test\", \"stream\": \"${TEST_STREAM}\" }"
 
+export MYSQL_DATABASE=test
+export MYSQL_HOST="${MYSQL_HOST:=source-mysql-db-1.flow-test}"
+export MYSQL_PORT="${MYSQL_PORT:=3306}"
+export MYSQL_USER="${MYSQL_USER:=root}"
+export MYSQL_PASSWORD="${MYSQL_PASSWORD:=flow}"
+
+docker compose -f source-mysql/docker-compose.yaml up --detach
+sleep 5
```
williamhbaker, line 14: Would be interesting to try something like [this](https://stackoverflow.com/a/46625302) here.

materialize-postgres-rc/driver.go
```diff
@@ -117,6 +135,31 @@ func newPostgresDriver() pm.DriverServer {
...
+					PrivateKey:  []byte(cfg.NetworkTunnel.SshForwarding.PrivateKey),
+					ForwardHost: host,
+					ForwardPort: port,
+					LocalPort:   "5432",
+				}
+				var tunnel = sshConfig.CreateTunnel()
+
+				// FIXME/question: do we need to shut down the tunnel manually if it is a child process?
+				// at the moment tunnel.Stop is not being called anywhere, but if the connector shuts down, the child process also shuts down.
+				err = tunnel.Start()
```
williamhbaker, line 156: On one hand, it seems like it would be good to call `tunnel.Stop` to clean up the temp key file. But since its running in a container with temporary storage it doesn't seem like it would be strictly necessary so I can't see why it would be required to shut it down manually.

williamhbaker (approved):

tests/source-mysql/setup.sh
mdibaiee: Good idea! Implemented it ✅

PR #376, opened 2022-10-12
source-firestore: Remove 'OrderBy' and retry temporary failures

willdonnelly:
**Description:**

I believe that the 'OrderBy' part of the Listen query is not necessary, and there is (very weak) evidence that it may cause ResourceExhausted errors to occur more frequently. This would make sense, if those errors indicate something like "you're using up too much server-side CPU/RAM" and the ordering requires extra work.

Retrying temporary failures seems like a good idea in general as well, since it allows the capture to make forward progress in some cases where immediately failing the capture wouldn't. I'm torn on whether there should be a maximum retry count or other safeguards, but since failing the entire capture will just result in Flow restarting it after 5 minutes or so, this should be an unambiguous improvement and we can think about improving on this in the future.

**Workflow steps:**

Hopefully captures should be a bit more stable with these changes.

None

**Notes for reviewers:**

Please ignore that the branch is named `wgd/firestore-ratelimit`, it turns out that rate-limiting anything on the connector side is a Bad Idea and just makes things worse. But the other tweaks are definitely helpful.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/376)

### DIFF
--- a/source-firestore/pull.go
+++ b/source-firestore/pull.go
@@ -29,6 +29,10 @@ var firebaseScopes = []string{
 
 const watchTargetID = 1245678
 
+// Non-permanent failures like Unavailable or ResourceExhausted can be retried
+// after a little while.
+const retryInterval = 300 * time.Second
+
 func (driver) Pull(stream pc.Driver_PullServer) error {
 	log.Debug("connector started")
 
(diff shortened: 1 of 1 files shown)


willdonnelly:
Going to go ahead and merge this. I think all of these changes are uncontroversial in the "this wouldn't raise any eyebrows if it had been present in the initial implementation" sense, they pass the automated test suite, and in my local testing this seems to make captures more reliable, so I'd like to try a test capture in production ASAP.

williamhbaker (approved):
LGTM!

Issue #357, opened 2022-09-27
Integrate new `materialize-sql` and `materialize-postgres-rc` packages


williamhbaker:
A new `materialize-sql` package was added in https://github.com/estuary/connectors/pull/347 with various enhancements over the [existing package](https://github.com/estuary/flow/tree/master/go/protocols/materialize/sql). Also in that PR, an updated `materialize-postgres` was created using the new `materialize-sql` package, among other updates. This new postgres materialization exists as `materialize-postgres-rc` and is used internally for stats currently.

We need to incorporate the new `materialize-sql` package into existing sql materialization connectors, and unify the `materialize-postgres-rc` connector into `materialize-postgres`. Having two separate copies of these is untenable for long-term maintenance, and the improvements from the new `materialize-sql` may benefit existing connectors as well.

- [x] integrate `materialize-bigquery`: https://github.com/estuary/connectors/pull/386
- [x] integrate `materialize-snowflake`: https://github.com/estuary/connectors/pull/469
- [x] integrate `materialize-postgres`: https://github.com/estuary/connectors/pull/439
- [x] remove `sqlite` from flow and make it an independent connector (it depends on `materialize-sql`): https://github.com/estuary/flow/issues/442
- [x] remove old `materialize-sql` code from `flow` repo: https://github.com/estuary/flow/pull/873

williamhbaker:
Unassigning myself from this for now. The `materialize-snowflake` integration needs to be completed when time and priorities allow.

PR #386, opened 2022-10-17
materialize-bigquery: use new sql framework

williamhbaker:
**Description:**

Part of https://github.com/estuary/connectors/issues/357

Refactors the `materialize-bigquery` connector to use the `materialize-sql` framework in this repository, instead of the older one in the flow repository. The objective of this PR was to accomplish this while remaining fully backward-compatible, and not introducing any behavior changes for the connector with the code change.

Code updates here include:
- Structural changes in the bigquery connector code to mesh with the new `materalize-sql` framework. This is the majority of the diff.
- Moving all of the query construction into templates, made possible by the new framework. The composition of the queries is easier to reason about in this form.
- Testing:
  - End-to-end tests were created for the bigquery connector, but I have opted not to run them automatically in CI. Interacting with the external bigquery service in CI tests seems like it would be problematic due to latency, availability, etc., and unfortunately there is no local bigquery emulator available. But the tests are useful to run locally for verification of correct behavior.
  - There are some comprehensive fencing tests (`go` tests) which can also be run locally.
  - The `flowsim` package was useful for testing changes locally and a few small updates were made there to bring it into working order. This is also what drove the creation of the `cmd/connector` folder and package name changes in the `bigquery` connector, so that `flowsim` could be use to directly drive the connector and enable more rapid local development.

**Workflow steps:**

Run the `materialize-bigquery` connector as per before. It should work the same way, and is fully backward-compatible.

N/A

**Notes for reviewers:**

See author's in-line comments for specific notes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/386)


materialize-bigquery/sqlgen.go
```diff
@@ -0,0 +1,233 @@
...
+	)
+	VALUES (
+	{{- range $ind, $col := $.Columns }}
+		{{- if $ind }}, {{ end -}}
+		r.{{$col.Identifier}}
+	{{- end -}}
+	);
+{{ end }}
+
+{{ define "installFence" }}
```
williamhbaker, line 156: This fence installation query was copied as-is from the existing code. I think there may be opportunities to simplify the query, but wasn't able to readily identify anything significant. It may be an are for future improvement. The current query does work correctly.

materialize-bigquery/bigquery.go
```diff
@@ -48,166 +44,184 @@ func (c *config) Validate() error {
 	return nil
 }
 
+func (c *config) client(ctx context.Context) (*client, error) {
+	var clientOpts []option.ClientOption
+
+	credentials, err := base64.StdEncoding.DecodeString(string(c.CredentialsJSON))
```
williamhbaker, line 50: Requiring base64-encoded credentials is not consistent with other google connectors we have and generally pretty annoying, but I left it alone for now to maintain backward compatibility. I think it is something that will need to be addressed eventually, and could probably even be done in a backward-compatible way but I have left that (and various other improvement tasks) for another day.

tests/materialize/materialize-bigquery/setup.sh
```diff
@@ -0,0 +1,53 @@
...
+}'
+
+resources_json_template='[
+  {
+    "resource": {
+      "table": "simple"
+    },
+    "source": "${TEST_COLLECTION_SIMPLE}"
+  },
+  {
```
williamhbaker, line 27: A case that is absent: Delta updates.

I found that testing a case like this (say, duplicate keys with delta updates) can be shown to work correctly 100% of the time, but it is cumbersome to verify the results. Timing issues make it so that sometimes the flow runtime sends a higher number of more granular delta updates than others where it sends a smaller number of "condensed" updates. This can vary apparently based on network conditions, or even which bindings are included in the test (their ordering). So I left that case out, but tested it quite a bit manually to be sure it works.

willdonnelly (approved):
LGTM.

Sorry I took so long getting to this, only to send it back without any meaningful review comments, but I promise I really did read it all and this looks correct to me.

tests/materialize/run.sh
```diff
@@ -118,8 +115,16 @@ flowctl-admin temp-data-plane \
     --unix-sockets \
     &
 DATA_PLANE_PID=$!
+
 # Arrange to stop the data plane on exit.
-trap "kill -s SIGTERM ${DATA_PLANE_PID} && wait ${DATA_PLANE_PID} && ${CONNECTOR_TEST_SCRIPTS_DIR}/cleanup.sh" EXIT
+function cleanup() {
+    echo -e "\nexecuting cleanup"
+
+    "${TEST_SCRIPT_DIR}/delete.sh" "" false || true
```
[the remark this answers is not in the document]
williamhbaker: Nope, must have just been some cruft from copying things elsewhere. I cleaned this up now, thanks.

PR #388, opened 2022-10-18
{source,materialize}-boilerplate: Support log.level and log.format

willdonnelly:
**Description:**

This commit adds support for specifying log level and format in the same way as our Airbyte captures and various other portions of the Flow codebase work. This lack is currently breaking `source-firestore` after some of the recent "connect over TCP" and "don't put the connector proxy in the execution path" changes caused it to get run directly with a `--log.level` argument.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/388)

### DIFF
--- a/materialize-boilerplate/boilerplate.go
+++ b/materialize-boilerplate/boilerplate.go
@@ -19,22 +19,18 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
-// TODO: handle the logger properly by setting the log level on logrus
-type Options struct {
-	LogLevel string `long:"log.level" description:"Log level, one of: error, warn, info, debug, trace" default:"info"`
-}
-
 // RunMain is the boilerplate main function of a materialization connector.
 func RunMain(srv pm.DriverServer) {
-	var opts = &Options{""}
-	var parser = flags.NewParser(opts, flags.Default)
+	var logConfig = &logConfig{}
+	var parser = flags.NewParser(logConfig, flags.Default)
 	var ctx, _ = signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
 
 	var cmd = cmdCommon{
-		ctx: ctx,
-		srv: srv,
-		r:   bufio.NewReaderSize(os.Stdin, 1<<21), // 2MB buffer.
-		w:   protoio.NewUint32DelimitedWriter(os.Stdout, binary.LittleEndian),
+		ctx:     ctx,
+		srv:     srv,
+		r:       bufio.NewReaderSize(os.Stdin, 1<<21), // 2MB buffer.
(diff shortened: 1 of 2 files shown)


materialize-boilerplate/boilerplate.go
```diff
@@ -70,6 +51,46 @@ type cmdCommon struct {
...
+		logrus.WithField("err", err).Fatal("unrecognized log level")
+	} else {
+		logrus.SetLevel(lvl)
+	}
+
+	const addr = ":2222"
+	logrus.WithField("addr", addr).Debug("waiting for gRPC connection")
+
+	var conn net.Conn
+	if server, err := net.Listen("tcp", addr); err != nil {
```
dhammika, line 83: Is it missing `defer server.Close()`? Same for client socket.

willdonnelly: It's not really necessary, the listening socket will remain open until the process exits, and it's the OS's responsibility to clean things up at that point.

Even so I'd normally prefer to include the cleanup logic, but in this particular case it'd be a bit of a pain to refactor things so that would work properly (because the command-line parsing library dispatches control to a bunch of distinct `Execute()` methods for each subcommand), so I'm happy just leaving it open.

dhammika: Ya, noticed flow is a bit complicated. Should be fine for short lived single instance runs but I'm not sure if we run these in parallel :)  Perhaps just add a comment to make it explicit.

willdonnelly:
I've rebased this change to the current mainline, now that all of the TCP listening stuff has been reverted.

materialize-boilerplate/boilerplate.go
```diff
@@ -59,6 +55,31 @@ type cmdCommon struct {
...
+	logging *logConfig
+}
+
+// logConfig configures handling of application log events.
+type logConfig struct {
+	Level  string `long:"log.level" env:"LOG_LEVEL" default:"info" choice:"info" choice:"INFO" choice:"debug" choice:"DEBUG" choice:"warn" choice:"WARN" description:"Logging level"`
+	Format string `long:"log.format" env:"LOG_FORMAT" default:"text" choice:"json" choice:"text" choice:"color" description:"Logging output format"`
+}
+
+func (c *cmdCommon) Execute(args []string) error {
```
dhammika, line 68: nit: Common.Execute() just sets logger params. Can we call it ParseArgs() or something more explicit?

dhammika (approved):
:+1:

materialize-boilerplate/boilerplate.go
willdonnelly: Yeah, that naming was a holdover from when I was also moving the TCP socket opening into that function. I've renamed the method and changed the receiver type so the usage looks like `c.cmdCommon.logging.Configure()` now.

PR #396, opened 2022-10-24
Revert TCP connectors back to stdio

willdonnelly:
**Description:**

This commit removes the `FLOW_TCP_PORT=2222` label from all Dockerfiles, removes the bit of code that opened a TCP port from source-boilerplate and materialize-boilerplate, and instead goes back to using os.Stdin and os.Stdout for those.

Unfortunately the logic changes to the boilerplate packages have to be made for all connectors at once, so this change has a fairly large blast radius if something is overlooked, but I don't think that can be helped.

**Workflow steps:**

In theory everything keeps working, and some of the current connector breakage is fixed for the moment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/396)


jgraettinger (approved):
LGTM

PR #397, opened 2022-10-26
Revert standalone network-tunnel from SQL connectors

willdonnelly:
**Description:**

Remove the standalone network-tunnel binary from all SQL database connectors. After the previous reverts to stop the bleeding from the stdio->TCP transition, we're left in a state where the connector proxy is trying to add the networkTunnel configuration when it already exists, so these three connectors produce errors when trying to do `flowctl-admin discover` at the command line.

I'm not sure if having two pieces of code attempting to *consume* the `networkTunnel` config is a problem, I *think* what happens is the connector proxy removes the `networkTunnel` key as part of the adapter logic. If that's correct then the only impact of this is that the `spec` subcommand is currently broken, which isn't a huge deal so long as we don't regenerate the cached endpoint spec schemas in Supabase, but I'm putting together this PR so it'll be ready to go if needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/397)


source-postgres/VERSION
```diff
@@ -1 +1 @@
-v2
+v1
```
mdibaiee, line 1: This means `source-postgres:v2` is still going to be broken?

mdibaiee (approved):

willdonnelly:
Per the discussion in today's standup we're going to try to instead roll things forward into a consistent state by disabling the network tunnel logic in connector_proxy and leaving it in the connectors.

PR #398, opened 2022-10-27
source-mysql: Attempt fallback to non-TLS connections

willdonnelly:
**Description:**

When connecting for backfills and replication, we first try to connect with TLS, and if that fails for any reason we try again without TLS before erroring out.

**Workflow steps:**

This shouldn't change anything except that some MySQL databases which we previously failed to connect to should now be usable.

None, we didn't mention that TLS was required in our current docs anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/398)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -191,17 +191,22 @@ func (db *mysqlDatabase) Connect(ctx context.Context) error {
 	}
 
 	// Normal database connection used for table scanning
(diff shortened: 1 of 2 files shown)


willdonnelly:
I have verified on my local machine that with this change the full test suite passes when run against MySQL with TLS enabled and also with TLS disabled.

jgraettinger (approved):
LGTM

PR #400, opened 2022-10-28
Add Validation Regexp for SSH Tunnel Endpoint

willdonnelly:
**Description:**

For the four connectors which work with SSH tunneling, this PR tweaks the endpoint config JSON schema to both mention the expected shape of the endpoint URI, and add a validation regexp so the UI can check it for them.

I believe that the regexp is correct for this purpose, but it should probably be scrutinized pretty heavily. If I understand correctly the regexp syntax here is just JavaScript's native regexps, so I tested it out in the browser console with variations on `new RegExp('^ssh://[^@]+@[^:]+(:[0-9]+)?$').test('ssh://root@example.com:22')` and it certainly _seems_ to work as desired, though.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/400)


mdibaiee:
The regex is pretty permissive if we want to stop invalid URIs to be allowed (that's up to debate, we may just let the tunnel error out), but these values are matching the regex:
```
ssh://user@- # any invalid character after `@` works
ssh://user@@test
ssh://user@test/ # trailing slash breaks the SSH client, tested it
```

I'd say it's valid to say that it will be hard to be very precise in our validation using a regex as the standards for valid URIs [can be complicated and permissive](https://www.rfc-editor.org/rfc/rfc2396) although SSH may have its own ideas, the best we can do is probably just avoid the user falling into common pitfalls such as the trailing backslash, so perhaps this one which doesn't allow slashes after the initial two, and doesn't allow extra `@`s after the middle one:

```
^ssh:\/\/[^@/]+@[^@:/]+(:[0-9]+)?$
```

willdonnelly:
> The regex is pretty permissive if we want to stop invalid URIs to be allowed

Yeah, my feeling is that this is a non-goal, in fact I don't really care if the regex permits obviously-incorrect line noise like `ssh://://:/@/:/::@@@`. The important thing in my opinion is just to make sure that it rejects *things that might otherwise seem temptingly valid* like a connection URI that forgets the username or lacks the `ssh://` prefix or is just the bare IP address.

After thinking about it a bit more I suspect even the port number stuff is overcomplicating things, if I was writing this again I'd probably just go with `^ssh://.+@.+$` and call it good. That said I also don't see a _problem_ with tightening up the allowed patterns a bit, the one you gave seems like it would work just fine too and probably would save a few users from having a bad setup experience.

mdibaiee:
I don't have a strong preference for the tighter version, I think `^ssh://.+@.+$` is fine if we present the error from ssh client to the user well

mdibaiee (approved):

willdonnelly:
I've switched the pattern to `^ssh://.+@.+$`. I'm pretty sure that when the user starts typing they'll see a little red `must match pattern "^ssh://.+@.+$"` warning below the input box, and in that case it seems like we should aim towards having something that the average user can actually read and understand rather than a huge mess of a regexp that they'll ignore.

Combined with the fixed description text so that it actually mentions the intended pattern I think this should work decently well.

Issue #401, opened 2022-10-28
Accept service account json for bigquery


dyaffe:
**Problem**
We currently require that a user base64's their credentials for only bigquery.  I've observed sessions which show that is confusing.  

**Proposal**
Standardize how credentials are input not to require base64.  We should accept the same form as the google sheets connector which allows the user to upload the json file of credentials that comes directly from Google.

PR #402, opened 2022-10-31
materialize-bigquery: support non-base64 credentials

williamhbaker:
**Description:**

Allows for service account credentials for `materialize-bigquery` to be provided as non-base64 encoded JSON, which is consistent with how all the other google cloud (etc.) connectors work. The spec descriptions are updated to indicate that the credentials don't need to be base64 encoded.

I made this change backwards compatible so that we don't need to bump the version of the connector for this. It will continue to work with base64 encoded JSON, but the spec & docs should direct users to create materializations with unencoded JSON.

Closes https://github.com/estuary/connectors/issues/401

**Workflow steps:**

A bigquery materialization can be created or updated with either base64 encoded JSON, or regular JSON. This also works when adding the JSON file to the config via the UI.

The `materialize-bigquery` docs should be updated to indicate that the credentials should be provided as unencoded JSON.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/402)

### DIFF
--- a/materialize-bigquery/.snapshots/TestSpecification
+++ b/materialize-bigquery/.snapshots/TestSpecification
@@ -11,8 +11,8 @@
       },
       "credentials_json": {
         "type": "string",
-        "title": "Credentials",
-        "description": "Google Cloud Service Account JSON credentials in base64 format.",
+        "title": "Service Account JSON",
+        "description": "The JSON key of the service account to use for authorization.",
         "multiline": true,
         "order": 1,
         "secret": true
--- a/materialize-bigquery/.snapshots/TestSpecification
+++ b/materialize-bigquery/.snapshots/TestSpecification
@@ -12,7 +12,7 @@
       "credentials_json": {
         "type": "string",
         "title": "Service Account JSON",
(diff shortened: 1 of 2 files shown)


materialize-bigquery/.snapshots/TestSpecification
```diff
@@ -11,8 +11,8 @@
       },
       "credentials_json": {
         "type": "string",
-        "title": "Credentials",
-        "description": "Google Cloud Service Account JSON credentials in base64 format.",
+        "title": "Service Account JSON",
+        "description": "The JSON key of the service account to use for authorization.",
```
willdonnelly, line 15: nit: s/key/credentials/ perhaps? The whole JSON credentials blob is more than just the private key.

willdonnelly (approved):
LGTM

materialize-bigquery/.snapshots/TestSpecification
williamhbaker: Seems like it would be better, thanks!

PR #404, opened 2022-11-01
source-mysql: Fix inconsistencies between MySQL and MariaDB

willdonnelly:
**Description:**

This PR tweaks a handful of minor details of `source-mysql` so that it should be able to successfully capture from MariaDB v10.3 and up. Currently there are still a few failures in the automated test suite:

  - The `datetime` datatype round trip test fails because MariaDB truncates fractional seconds instead of rounding.
  - The `json` datatype round trip test fails because MariaDB has `JSON` as an alias for the `LONGTEXT` column type, so discovery and captures will just treat that as a string.
  - The `TestBinlogExpirySanityCheck` test fails on older versions of MariaDB (but not latest) because the `binlog_expire_logs_seconds` system variable doesn't exist and the test code wants to modify that. However the sanity check itself should still work on those DB versions, it will simply fall back to looking at `expire_logs_days` like it should.

**Workflow steps:**

After this is merged, it should be possible to perform discovery and captures against a MariaDB instance using the same `ghcr.io/estuary/source-mysql:v2` connector that we currently offer in the web UI.

Our test suite exercises enough of the core capture behavior that I'm confident that ought to work reliably, however there could be other edge cases around query event parsing and datatype edge cases which we won't encounter until this is run against a real production database.

We should probably tweak https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/ to mention somewhere that MariaDB is also supported.

### DIFF
--- a/source-mysql/datatype_test.go
+++ b/source-mysql/datatype_test.go
@@ -75,7 +75,10 @@ func TestDatatypes(t *testing.T) {
 
 		{ColumnType: "date", ExpectType: `{"type":["string","null"]}`, InputValue: "1991-08-31", ExpectValue: `"1991-08-31"`},
 		{ColumnType: "datetime", ExpectType: `{"type":["string","null"]}`, InputValue: "1991-08-31 12:34:56", ExpectValue: `"1991-08-31 12:34:56"`},
+
+		// This test fails on MariaDB because it truncates fractional seconds rather than rounding (see also: https://jira.mariadb.org/browse/MDEV-16991).
 		{ColumnType: "datetime", ExpectType: `{"type":["string","null"]}`, InputValue: "1991-08-31 12:34:56.987654", ExpectValue: `"1991-08-31 12:34:57"`},
+
 		// TODO(wgd): Timestamps are reported differently in backfills vs replication because backfill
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

If there ends up being a lot of confusion about mysql vs. mariadb, maybe we could just tag & push a mariadb connector image built from the same dockerfile as mysql. Seems like calling it out as compatible with both in the docs should be good unless we end up getting a lot of questions though.

willdonnelly:
> tag & push a mariadb connector image built from the same dockerfile as mysql

Yeah, I started to go down this route for this PR but it proved to be a bit more hairy than I wanted to deal with at the moment. For instance there's no way for one Dockerfile to `INCLUDE` another, using a `FROM` directive would make the mariadb version perpetually one build out of date, and symlinking `source-mariadb/Dockerfile -> source-mysql/Dockerfile` just felt wrong.

And similarly there would need to be a separate VERSION file governing the version tag pushed by CI builds, and you just know that'll fail to get bumped and break something in the future. Fixing all of that would probably require moving more of the "push newly build Docker image to ghcr.io" logic out of `ci.yaml` into a per-connector script or something.

PR #406, opened 2022-11-02
source-firestore: Limit amount of data fetched during discovery

willdonnelly:
**Description:**

This is the *fourth* rewrite of Firestore discovery, and I'm really hoping it will be the last one. The new logic keeps the decently high throughput of the previous attempt (currently it peaks at 16MBps running on my laptop) but applies harder limits to the number of documents fetched.

The new approach is roughly:
 - Separate 'scan workers' and 'inference workers' into separate errgroups. This allows the shutdown logic to be made a bit more centralized at the top level: first wait until all scanners are done, then close all inference channels, then wait until all inference workers are done. And doing that means that we can be a bit more disorganized about spawning workers as described in the next bullet.
 - Spawn a 'scan worker' for each distinct *collection* we observe, rather than each *collection group* as was done previously. This results in a *lot* more parallel threads, so a semaphore is added to bound the number of workers which can execure concurrently (currently limited to 16, in my testing going higher than this didn't provide a meaningful improvement in speed).
 - We already maintained a count of the number of documents processed from a particular resource path, but now the scan workers will check that count after they first acquire their semaphore, so that they can stop immediately if the overall per-resource limit (which is currently 10k documents) has already been reached.
 - Also apply a smaller bound on the number of documents fetched from any specific collection. The existence of this limit allows the iterative 'Get Next' fetch operation to be replaced with a batch "Limit K, Get All" instead. This helps to regain a lot of the throughput lost in the switch from scanning by collection group to scanning individual collections.

Taken all together, discovery now terminates in about 3s when run against my 30GiB synthetic test dataset and about 4-6s when run against a smaller but more complex dataset with ~20 resources.

The new logic may be slightly less reliable in terms of "did we find every possible resource", but the nature of Firestore is that this will always be a probabilistic/ heuristic thing if we don't fetch 100% of documents and check every last one of them for subcollections, so I think that speed [1] is more important.

[1] Well, it's not just about how fast it completes. It's also important that the new logic won't (hypothetically) go and try to fetch a complete dataset of 100M documents and thereby cost the user $60 of Firestore billing every time they hit 'Discover' in the UI.

**Workflow steps:**

None, in theory this should behave similarly to the old implementation, maybe a little bit faster, and the most important part is the improved limits on the total amount of data which will be fetched during Discovery.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/406)

### DIFF
--- a/source-firestore/discovery.go
+++ b/source-firestore/discovery.go
@@ -287,16 +287,13 @@ func (ds *discoveryState) handleDocument(ctx context.Context, doc *firestore.Doc
 }
 
 func (ds *discoveryState) handleCollection(ctx context.Context, coll *firestore.CollectionRef) {
-	ds.shared.Lock()
-	defer ds.shared.Unlock()
-
 	ds.scanners.Go(func() error {
 		var err = ds.discoverCollection(ctx, coll)
 		if err != nil {
 			log.WithFields(log.Fields{
-				"group": coll.ID,
-				"error": err,
-			}).Error("error scanning documents")
+				"collection": trimDatabasePath(coll.Path),
+				"error":      err,
+			}).Error("error scanning collection")
 		}
 		return err
 	})
--- a/source-firestore/discovery.go
+++ b/source-firestore/discovery.go
@@ -19,7 +19,6 @@ import (
 	"github.com/invopop/jsonschema"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
-	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
 )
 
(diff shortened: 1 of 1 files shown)


source-firestore/discovery.go
```diff
@@ -268,25 +277,42 @@ func (ds *discoveryState) processDocument(ctx context.Context, doc *firestore.Do
...
+				ds.handleCollection(ctx, subcoll)
 			}
 			return nil
 		})
 	}
 	return nil
 }
 
+func (ds *discoveryState) handleCollection(ctx context.Context, coll *firestore.CollectionRef) {
+	ds.shared.Lock()
```
williamhbaker, line 289: Is this lock needed?

williamhbaker (approved):
LGTM, one minor question.

source-firestore/discovery.go
willdonnelly: You're right, that's not necessary any more. (It's left over from the previous discovery implementation where this function was checking a shared map to make sure it only spawned one worker thread per collection group)

PR #407, opened 2022-11-03
source-firestore: Stop setting `additionalProperties: false`

willdonnelly:
**Description:**

I thought this had been fixed previously, but apparently we're still generating collection schemas which forbid additional properties of the `/_meta` object, which then immediately breaks because Flow adds `/_meta/uuid`.

**Workflow steps:**

Doing a simple "create capture via discovery then materialize it" setup in the UI should actually work without any manual schema editing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/407)


willdonnelly:
@oliviamiannone the issue you're seeing should be fixed a few minutes after this gets merged.

mdibaiee (approved):

PR #413, opened 2022-11-08
source-gcs: Backport https://github.com/estuary/flow/pull/780

willdonnelly:
**Description:**

This isn't a simple cherry-pick because the 'protocols/airbyte' helper package lives in the Flow repository, and bumping the entire version of that has a bunch of undesired side effects.

So instead, I'm taking the lazy way out and copy-pasting the whole 'protocols/airbyte' package over, then modifying source-gcs to use the new copy. This appears to work as desired with minimal side effects, and while it's not the most maintainable in the long run we will hopefully not be maintaining this backport branch very long.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/413)


willdonnelly:
Oh hang on, I created this PR wrong...

willdonnelly:
Hmm. So my goal here is simply to get a CI build of the version of stuff that I just pushed as `backport/source-gcs-ef602d3`. And to have CI build it, the easiest route is to create a PR, so that's what I did. But does that build the tip commit of the branch, or does that try to cherry-pick it onto `main`?

willdonnelly:
Okay yeah, I skipped a step. https://github.com/estuary/connectors/pull/414 does what I was trying to accomplish here.

PR #414, opened 2022-11-08
source-gcs: Backport https://github.com/estuary/flow/pull/780

willdonnelly:
**Description:**

This isn't a simple cherry-pick because the 'protocols/airbyte' helper package lives in the Flow repository, and bumping the entire version of that has a bunch of undesired side effects.

So instead, I'm taking the lazy way out and copy-pasting the whole 'protocols/airbyte' package over, then modifying source-gcs to use the new copy. This appears to work as desired with minimal side effects, and while it's not the most maintainable in the long run we will hopefully not be maintaining this backport branch very long.

### DIFF
--- a/filesource/filesource.go
+++ b/filesource/filesource.go
@@ -12,8 +12,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/estuary/connectors/protocols/airbyte"
 	"github.com/estuary/flow/go/parser"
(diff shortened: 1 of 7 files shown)


willdonnelly:
Hmm, this doesn't quite work either because our CI setup currently only builds on pushes and PRs targeting the `main` branch...

PR #420, opened 2022-11-10
source-mysql: Add a couple of debug log messages

willdonnelly:
**Description:**

Adds a couple more debug log messages. We're observing one backfill in production which appears to hang for **a very very long time** on each backfill query, and it would be useful to know for certain that the query itself is the part taking all that time (as opposed to the result value processing).

And happily, now that the final couple of "make log levels work" PRs have landed, we can actually turn on debug logging in production via `flowctl-admin shards edit`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/420)

### DIFF
--- a/source-mysql/backfill.go
+++ b/source-mysql/backfill.go
@@ -50,6 +50,11 @@ func (db *mysqlDatabase) ScanTableChunk(ctx context.Context, info sqlcapture.Tab
 
 	// Process the results into `changeEvent` structs and return them
 	var events []sqlcapture.ChangeEvent
+	logrus.WithFields(logrus.Fields{
+		"schema": schema,
+		"table":  table,
+		"rows":   len(results.Values),
+	}).Debug("translating query rows to change events")
 	for _, row := range results.Values {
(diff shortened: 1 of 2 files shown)


source-mysql/backfill.go
```diff
@@ -50,6 +50,11 @@ func (db *mysqlDatabase) ScanTableChunk(ctx context.Context, info sqlcapture.Tab
 
 	// Process the results into `changeEvent` structs and return them
 	var events []sqlcapture.ChangeEvent
+	logrus.WithFields(logrus.Fields{
+		"schema": schema,
+		"table":  table,
+		"rows":   len(results.Values),
+	}).Debug("translating query rows to change events")
```
dhammika, line 57: Doesn't this log sensitive row data or it's not a concern here?

willdonnelly: No, it's just logging the length of the result set, not its contents (which would be a crapton of data we wouldn't want to log purely on pragmatic grounds, in addition to the privacy concerns).

PR #421, opened 2022-11-10
source-{mysql,postgres}: EXPLAIN backfill queries and log result

willdonnelly:
**Description:**

Once per connector restart (per backfilling table), before actually running the backfill query we ask the database to EXPLAIN the query and then log the response.

This should help us diagnose some conditions where backfill queries aren't performing as expected. Any errors which occur during the EXPLAIN process will be logged but not treated as fatal.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/421)


willdonnelly:
Going to go ahead and merge this unreviewed, because I want to see what it says on the capture that's currently having issues. I don't anticipate any side effects from the change (as the only real effect of the new code is some info logging and any failures just get logged and we keep going) and we'll always be free to tweak it later on if needed.

PR #423, opened 2022-11-18
Migrate SQL captures (source-mysql and source-postgres) to Flow gRPC protocol

willdonnelly:
**Description:**

This is a fairly massive refactoring which replaces the Airbyte protocol used by the `source-mysql` and `source-postgres` capture connectors with the native Flow gRPC protocol. This is worthwhile for several reasons:

- Postgres replication slot behavior
  - Currently `source-postgres` doesn't know anything about Flow transaction acknowledgements, so it can't advance the Postgres replication slot's cursor (`restart_lsn`) until the connector restarts (at which point it knows that the state checkpoint it just restarted from must be durably committed).
  - That behavior is very suboptimal, because if the connector stays up for days before being restarted that's multiple days of extra WAL retention on the Postgres DB.
  - The current PR doesn't actually address this problem, but it's a necessary prerequisite -- now that we can receive Acknowledge RPCs from Flow we can tackle this issue in the near future.
- Connector reliability
  - We have seen _so many_ problems caused by intermediate binaries (first `connector_proxy` and later `airbyte-to-flow`).
  - Getting those out of the execution path makes things quite a bit simpler, and now there's one fewer component to look at if anything goes wrong.
- Simpler automated test suites
  - The `capture.DriverServer` interface encapsulates more of the interesting behavior of a connector.
  - We can write generic helper logic (see `source-boilerplate/testing`) which can be applied to *any* gRPC capture.
  - I don't have hard numbers but this allowed removing *a lot* of unnecessary test framework logic from the SQL captures (the big win is that this can be shared with `source-firestore` and any future database-type integrations).

**Workflow steps:**

One of the commits in this PR bumps the `VERSION` files. In theory this change should be more or less backwards-compatible [1], but there's so much changed that I thought the extra caution was worthwhile.

[1] There might be a slight incompatibility around resource specs, since we have to specify our own now instead of using the standard Airbyte resource format, and so I elected not to keep around the fields we don't support and to instead make `primary_key` an explicit property in a more intuitive format. I did leave the names of the schema/table properties as `namespace` and `stream` for the moment (and because `s/namespace/schema` would also be confusing), but I'm still not sure if that's the right call. Also I removed the whole "default namespace if unspecified" thing, I think that was a mistake and in general if people are setting things up in the UI the discovery output will specify the table namespaces explicitly.

We may need to tweak the `source-mysql` and `source-postgres` docs to reflect the changes to the resource spec schema.

**Notes for reviewers:**

The delta and "lines changed" counts from this PR are pretty massive, but a large chunk of that is just churn in the test snapshot representation, so the actual code changes aren't nearly as bad as all that.

I recommend reviewing this commit-by-commit, as I've done my best to group related changes and keep things understandable when read in that light.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/423)


source-boilerplate/boilerplate.go
```diff
@@ -278,3 +281,82 @@ func (a *pullAdapter) SetHeader(metadata.MD) error {
...
+	for _, doc := range docs {
+		var begin = uint32(len(arena))
+		arena = append(arena, []byte(doc)...)
+		slices = append(slices, pf.Slice{Begin: begin, End: uint32(len(arena))})
+	}
+
+	// TODO(wgd): The logic of this function needs to be rewritten to support
+	//   splitting into multiple 'Documents' RPCs in order to keep the size of
+	//   any individual message bounded.
+	if len(arena) > 2*1024*1024 {
```
williamhbaker, line 320: Is this a new limitation? I'm not sure if it was part of airbyte-to-flow as well, or if it is something new that might happen.

sqlcapture/tests/helpers.go
```diff
@@ -1,165 +1,72 @@
...
-		if idx < len(actualLines) {
-			y = actualLines[idx]
-		}
-		if x != y {
-			t.Errorf("snapshot %q mismatch at line %d", snapshotFile, idx)
-			t.Errorf("old: %s", x)
-			t.Errorf("new: %s", y)
-			break
+	var captureCtx, cancelCapture = context.WithCancel(ctx)
+	const shutdownDelay = 200 * time.Millisecond
```
williamhbaker, line 35: This seems like a pretty short delay - sounds like you've tested it and it works reliably though. Might be something we need to adjust if it proves flaky in the future.

source-mysql/main_test.go
```diff
@@ -200,21 +173,61 @@ func (tb *mysqlTestBackend) Query(ctx context.Context, t testing.TB, query strin
...
+func TestAlterTable(t *testing.T) {
+	var tb, ctx = TestBackend, context.Background()
+	var tableA = tb.CreateTable(ctx, t, "aaa", "(id INTEGER PRIMARY KEY, data TEXT)")
+	var tableB = tb.CreateTable(ctx, t, "bbb", "(id INTEGER PRIMARY KEY, data TEXT)")
+	var tableC = tb.CreateTable(ctx, t, "ccc", "(id INTEGER PRIMARY KEY, data TEXT)")
+	tb.Insert(ctx, t, tableA, [][]interface{}{{1, "abc"}, {2, "def"}})
+	tb.Insert(ctx, t, tableB, [][]interface{}{{3, "ghi"}, {4, "jkl"}})
+	tb.Insert(ctx, t, tableC, [][]interface{}{{5, "mno"}, {6, "pqr"}})
+
+	// var catalogA = tests.ConfiguredCatalog(ctx, t, tb, tableA)
```
williamhbaker, line 191: Are these commented lines still needed?

mdibaiee (approved):

source-postgres/.snapshots/TestTruncatedTables-capture2-fails
```diff
@@ -0,0 +1,9 @@
+# ================================
+# Final State Checkpoint
+# ================================
+{"cursor":"0/1111111","streams":{"test.truncatedtables_aaa":{"key_columns":["id"],"mode":"Active"}}}
+# ================================
+# Captures Terminated With Errors
+# ================================
+failed to relay messages: error decoding message: unsupported operation TRUNCATE on table "test.truncatedtables_aaa"
```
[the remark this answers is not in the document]
willdonnelly: Yes, this is the intended outcome. There are two truncations and two captures in the test: the first one to an unrelated table which ought to be ignored, and the second to an actively captured table which is expected to cause an error.

sqlcapture/capture.go
```diff
@@ -603,41 +572,33 @@ func emitMessage(enc MessageOutput, msg interface{}) error {
...
-				Namespace: sourceCommon.Schema,
-				Stream:    sourceCommon.Table,
-				EmittedAt: time.Now().UnixNano() / int64(time.Millisecond),
-				Data:      json.RawMessage(bs),
-			},
-		})
+		// TODO(wgd): Compute the binding index for a particular message more intelligently
+		var sourceCommon = msg.Source.Common()
+		var streamID = JoinStreamID(sourceCommon.Schema, sourceCommon.Table)
+		var binding, ok = c.Bindings[streamID] // TODO(wgd): Unsynchronized access to this map from the worker goroutine should be fine since it's read-only after capture initialization?
```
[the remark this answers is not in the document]
willdonnelly: Yeah, I ran the test suites under the race detector a bunch (that's why I ended up fixing two other race conditions in this PR) and this particular access doesn't appear to be an issue, I just forgot to remove the reminder comment here.

sqlcapture/tests/helpers.go
willdonnelly: Short/long is all relative -- 200ms is enough time for a _fairly long_ network round trip, and at least for local tests (like in our CI builds) where the target database is running as a Docker container on the same machine this appears to be plenty generous.

So empirically this works, and we want it to be as short as possible because this is a constant overhead added to every capture of every test in the automated suite.

That said it might fail if I tried to do another round of the whole "throughput profiling under load" thing. But it shouldn't be too hard to increase the timeout if that becomes an issue.

source-boilerplate/boilerplate.go
willdonnelly: I'm not sure if it's an airbyte-to-flow problem or not, it depends on how the relevant translation is implemented there. It's possible to just implement the output logic in a dumber way where multiple documents never get combined into a single Documents RPC...and now that I've typed that it seems obvious that I should just remove the "coalesce multiple documents" logic for now and just implement this function in the less-theoretically-efficient [1] way.

[1] Neither `source-firestore` nor the `sqlcapture` package ever emit multiple documents via a single call so the efficiency argument is irrelevant in a practical sense at the moment.

williamhbaker (approved):

willdonnelly:
I've rebased the changes after the earlier MySQL PR around timestamp translation, and added a small additional change: the `syncMode` property is now permitted to exist (although still ignored and not mentioned in the resource spec schema) on a resource binding. This should greatly simplify the merge/rollout of this change: all our `:dev` tagged monitoring captures should continue working without any changes, and all other SQL DB captures should be upgradable by just changing `:v2` into `:v3` with no other edits.

If there are no objections and assuming the CI build passes, I'll merge this a bit later today.

PR #426, opened 2022-11-22
source-firestore: Add 'initTimestamp' resource property

willdonnelly:
**Description:**

This is vaguely analagous to how we've used path regexes on filesource captures in the past, as a mechanism for kicking a malfunctioning capture back into the state we need it to be in.

In this particular case, what this allows is taking a collection binding for which replication streaming is failing and removing it from the catalog, and then at a later date we can either re- add it with *the same* timestamp that it used to have, or we can add it with some later timestamp if we need to fiddle around with things manually.

**Workflow steps:**

Nothing should change by default, but now it's possible to manually tweak the initial replication cursor timestamp when adding a capture binding (and **only** when first adding it, the state is authoritative and this new property just impacts the default/initialization behavior, so changes after that point have no effect).

I would recommend we not document this particular knob at the moment, but if we do want to mention it that will need to go in https://go.estuary.dev/source-firestore

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/426)

### DIFF
--- a/source-firestore/.snapshots/TestSpec
+++ b/source-firestore/.snapshots/TestSpec
@@ -39,7 +39,11 @@
           "sync"
         ],
         "title": "Backfill Mode",
-        "description": "Configures the handling of data already in the collection. Refer to go.estuary.dev/source-firestore for details or just stick with 'async'"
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

oliviamiannone:
Following your recommendation to leave undocumented - but leaving "docs pending" label on here to flag it as such for posterity.

PR #430, opened 2022-11-24
Remove `tail`

mdibaiee:
**Description:**

- Remove use of `tail` when running e2e tests, and instead periodically check lines of output against expected lines, and once we reach the same number of lines we compare them one final time to see if the test passes or not
- Remove code that handled non-tailing mode from various connectors

**Workflow steps:**

N/A, supposed to be seamless, unless you were using `--poll`, which now connectors won't handle

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/430)

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -52,13 +52,11 @@ export BROKER_ADDRESS=unix://localhost${TESTDIR}/gazette.sock
 export CONSUMER_ADDRESS=unix://localhost${TESTDIR}/consumer.sock
 
 # Start an empty local data plane within our TESTDIR as a background job.
-# --poll so that connectors are polled rather than continuously tailed.
 # --sigterm to verify we cleanly tear down the test catalog (otherwise it hangs).
 # --tempdir to use our known TESTDIR rather than creating a new temporary directory.
 # --unix-sockets to create UDS socket files in TESTDIR in well-known locations.
 flowctl-go temp-data-plane \
     --log.level info \
-    --poll \
     --sigterm \
     --network "flow-test" \
     --tempdir ${TESTDIR} \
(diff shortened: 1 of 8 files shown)


tests/run.sh
```diff
@@ -105,14 +103,23 @@ flowctl-go api build --directory ${TESTDIR}/builds --build-id test-build-id --so
 
 # Activate the catalog.
 flowctl-go api activate --build-id test-build-id --all --network "flow-test" --log.level info || bail "Activate failed."
-# Wait for a data-flow pass to finish.
-flowctl-go api await --build-id test-build-id --log.level info || bail "Await failed."
-# Read out materialization results.
-sqlite3 -header "${OUTPUT_DB}" "select id, canary from test_results;" > "${ACTUAL}"
+
```
jgraettinger, line 106: Continue to use `api await` here? Seems that the capture connectors in question _do_ actually exit, given the source-test modification, and we can continue to use this property.

Think this makes test UX better since `api await` will fail-fast if a shard has failed, and exits if the dataflow has settled even if the output < expected lines (rather than hanging).

mdibaiee: From my tests, they don't exit. In particular source-postgres and source-mysql will wait until they reach a nonexistant-watermark and will never quit. See this run for example: https://github.com/estuary/connectors/actions/runs/3553412074/jobs/5968911590#step:14:204

mdibaiee: But we do probably need some sort of timeout here in case we don't reach the expected number of lines

jgraettinger: 🤦 of course. Sounds good

williamhbaker (commented):
This LGTM in general, I did have a couple of (pretty basic) questions though -

tests/run.sh
```diff
@@ -105,14 +103,29 @@ flowctl-go api build --directory ${TESTDIR}/builds --build-id test-build-id --so
...
+# Periodically check expected vs actual lines of output, once we reach the same
+# number of lines, we stop the plane and then compare the output
+retry_counter=0
+while true
+do
```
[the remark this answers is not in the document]
mdibaiee: `sqlcapture/tests/helpers.go` is for unit tests, this is for integration tests

sqlcapture/tests/helpers.go
```diff
@@ -23,7 +24,10 @@ import (
 // emitted during the capture, and updates the `state` argument to the final one.
 func VerifiedCapture(ctx context.Context, t *testing.T, tb TestBackend, catalog *airbyte.ConfiguredCatalog, state *sqlcapture.PersistentState, suffix string) []sqlcapture.PersistentState {
 	t.Helper()
-	var result, states = PerformCapture(ctx, t, tb, catalog, state)
+
+	var expectedLines = SnapshotLines(t, suffix)
```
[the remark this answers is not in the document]
mdibaiee: I'm not sure about this to be honest... I can't think of a case at the moment, but can't rule out the possibility either.

tests/run.sh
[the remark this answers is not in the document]
mdibaiee: Actually now that's not needed anymore, Will's latest PR has made unit tests possible without periodic checks

williamhbaker (approved):

PR #431, opened 2022-11-28
materialize-bigquery: use new `materialize-sql` framework with correct edc file "required" schema

williamhbaker:
**Description:**

The first commit is the revert of a revert. This was already reviewed in https://github.com/estuary/connectors/pull/386.

The second commit is a correction on top of that that. The bigquery external data connection file has a sort of schema, and setting the `Required` parameter was not as simple as checking `pf.Inference_MUST`.

There is an additional resolution that has to happen here: `pf.Inference_MUST` indicates that the field is `required` in terms of flow schema, but it can still be `null`. `Required` for the edc schema refers to nullability, so the two aren't directly equivalent. So we need to check to make sure that the column truly cannot be `null` before marking the edc schema as `Required`. Fortunately `col.AsFlatType()` does exactly this check, and we can use its result for the edc schema.

**Workflow steps:**

N/A

**Notes for reviewers:**

The overall diff is large, but the actual change that has not been previously reviewed is in b9a6de068bfab94eb8a9fc73780dee3cb4a1dcc6, and is a 2-line change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/431)


mdibaiee (approved):
LGTM!

williamhbaker:
Merged via 52cc406b97c7c61e6e6c9d0a52d940ea62500db6

PR #433, opened 2022-11-29
source-mysql: Translate TIMESTAMP/DATETIME to RFC3339 strings

willdonnelly:
**Description:**

Previously they were captured as strings in MySQL's preferred timestamp format (`YYYY-MM-DD HH:MM:SS`). This commit changes things so that they are instead (`YYYY-MM-DDTHH:MM:SSZ`).

Note that MySQL TIMESTAMP / DATETIME columns don't actually store any useful time zone information. A TIMESTAMP is defined as being converted to UTC for storage, so we rely on the DB to get that right and just request that it give us the values in UTC as well. For the DATETIME column type we don't even have that assurance, so treating them as UTC is probably wrong but it's still the best we can do.

I'm torn on whether this change requires a version bump. All we previously said was "it's a string" so changing the format is _technically legal_, I think? It would make things a lot easier if we can fix this without a version bump, because then it's simple to go ahead and merge this prior to the big gRPC migration PR that'll be merged in the near future.

**Workflow steps:**

The `TIMESTAMP` and `DATETIME` column types will now be discovered as `type: string, format: date-time` instead of just `type: string` and captured values are emitted as valid RFC3339 timestamps.

None, I don't think we document current behavior at this level of detail.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/433)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/estuary/connectors/sqlcapture"
 	"github.com/go-mysql-org/go-mysql/client"
@@ -139,6 +140,8 @@ func translateRecordFields(columnTypes map[string]interface{}, f map[string]inte
 	return nil
 }
 
+const mysqlTimestampLayout = "2006-01-02 15:04:05"
+
 func translateRecordField(columnType interface{}, val interface{}) (interface{}, error) {
 	if columnType == nil {
 		return nil, fmt.Errorf("unknown column type")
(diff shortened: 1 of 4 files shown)


source-mysql/main.go
```diff
@@ -218,6 +218,11 @@ func (db *mysqlDatabase) Connect(ctx context.Context) error {
 		return fmt.Errorf("unable to connect to database: %w", err)
 	}
 
+	// Set our desired timezone to UTC. This is required for backfills of timestamp columns to behave consistently.
+	if _, err := db.conn.Execute("SET time_zone = '+00:00';"); err != nil {
```
williamhbaker, line 222: I was originally wondering if this changed the time zone of the database itself, but apparently its actually a per session setting, so that's good.

source-mysql/discovery.go
```diff
@@ -165,6 +168,12 @@ func translateRecordField(columnType interface{}, val interface{}) (interface{},
 				return val, nil
 			case "json":
 				return json.RawMessage(val), nil
+			case "datetime", "timestamp":
```
williamhbaker, line 171: Does this still work with existing captures & collections? I guess it would, since its still returning a string, just a more specific kind of string. Assuming that is true, I'm not entirely sure how to feel about versioning of the connector. But ultimately I think there are so few of these deployed yet that it should be fine.

jgraettinger (approved):
LGTM

I think we can do this without a version bump

williamhbaker (approved):

PR #435, opened 2022-11-29
tests: don't use flowctl-go combine

williamhbaker:
**Description:**

The flowctl-go combine command was removed in https://github.com/estuary/flow/pull/818. `materialize-s3-parquet` was the only test actually using this, and it didn't really need to, so this cleans that up and gets the test passing again.

Also added the better exit message to materialization tests that was added to capture tests in https://github.com/estuary/connectors/pull/423

**Workflow steps:**

CI tests will once again work for `materialize-s3-parquet`, or when run locally.

N/A

**Notes for reviewers:**

There test for `elasticsearch` still doesn't work, but that seems to be a separate problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/435)


willdonnelly (approved):
LGTM

PR #440, opened 2022-12-01
source-mysql: Use DB `time_zone` for DATETIME parsing

willdonnelly:
**Description:**

The DATETIME column type is just a raw 'YYYY-MM-DD HH:MM:SS' value with no explicit frame of reference. In general there's an implicit reference of "the user's local time zone", but we don't necessarily know what that is. And we'd really like to emit captured values as RFC3339 timestamps, so we need to interpret the values *somewhere*.

After this PR we assume that if the database's `time_zone` system variable is set to a IANA zone name or a numeric time offset then that is the location/offset in which a DATETIME should be parsed.

If the `time_zone` variable is set to its default value of `'SYSTEM'` we refuse to translate DATETIMEs, because for some reason the MySQL `system_time_zone` variable only contains the current time offset [1] and not a valid zone name which could be used to interpret historical/future dates.

[1] To be concrete: if the system locale is set to 'America/Los_Angeles' then MySQL's `system_time_zone` will report 'PST' or 'PDT' at different times of the year, rather than something like 'PST8PDT' or 'America/Los_Angeles' which could actually be used to interpret historical data. In addition the strings 'PST' and 'PDT' are not valid IANA time zone names, so we just avoid all that complexity by requiring a valid `time_zone` setting.

We should update https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/ to discuss our requirement that the `time_zone` system variable be set to a valid time zone name or numeric offset if the user wants to capture `DATETIME` columns. Then https://go.estuary.dev/80J6rX should be updated to link to that particular bit of the docs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/440)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -235,7 +235,9 @@ func (db *mysqlDatabase) connect(ctx context.Context) error {
 		logrus.Warn("capturing DATETIME values will not be permitted")
 	}
 
-	// Set our desired timezone to UTC. This is required for backfills of TIMESTAMP columns to behave consistently.
+	// Set our desired timezone (specifically for the backfill connection, this has
+	// no effect on the database as a whole) to UTC. This is required for backfills of
+	// TIMESTAMP columns to behave consistently, and has no effect on DATETIME columns.
 	if _, err := db.conn.Execute("SET time_zone = '+00:00';"); err != nil {
 		return fmt.Errorf("error setting session time_zone: %w", err)
 	}
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -230,7 +230,6 @@ func (db *mysqlDatabase) connect(ctx context.Context) error {
 	// Infer the location in which captured DATETIME values will be interpreted.
 	if loc, err := queryTimeZone(conn); err == nil {
 		db.datetimeLocation = loc
-		logrus.WithField("loc", loc.String()).Debug("using datetime location")
 	} else {
 		logrus.WithField("err", err).Warn("unable to determine database timezone")
 		logrus.Warn("capturing DATETIME values will not be permitted")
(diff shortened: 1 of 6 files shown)


source-mysql/main.go
```diff
@@ -245,22 +258,57 @@ func (db *mysqlDatabase) connect(ctx context.Context) error {
...
+		}
+		minutes, err := strconv.Atoi(parts[1])
+		if err != nil {
+			return nil, fmt.Errorf("error parsing %q: %w", tzName, err)
+		}
+		return time.FixedZone("UTC"+tzName, sign*(hours*60*60+minutes*60)), nil
+	}
+
+	// Otherwise let's hope the time zone setting is a valid IANA zone name.
+	return time.LoadLocation(tzName)
```
mdibaiee, line 293: I kinda want to have an easter egg that says "Thank You" if the timezone is set to UTC 😂

mdibaiee (approved):

source-mysql/main.go
```diff
@@ -223,7 +227,16 @@ func (db *mysqlDatabase) connect(ctx context.Context) error {
...
+	if loc, err := queryTimeZone(conn); err == nil {
+		db.datetimeLocation = loc
+		logrus.WithField("loc", loc.String()).Debug("using datetime location")
+	} else {
+		logrus.WithField("err", err).Warn("unable to determine database timezone")
+		logrus.Warn("capturing DATETIME values will not be permitted")
+	}
+
+	// Set our desired timezone to UTC. This is required for backfills of TIMESTAMP columns to behave consistently.
 	if _, err := db.conn.Execute("SET time_zone = '+00:00';"); err != nil {
```
jgraettinger, line 240: Making sure I'm following: we're fetching a persisted `time_zone` on startup, and then overwriting it _for our session only_ as part of making backfills work for TIMESTAMP? This is sound because DATETIME query results are not affected by setting `time_zone`, while TIMESTAMP will consistently be UTC, and no other sessions will observe our write to `time_zone`?

source-mysql/main.go
```diff
@@ -245,22 +258,57 @@ func (db *mysqlDatabase) connect(ctx context.Context) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("error querying 'time_zone' system variable: %w", err)
+	}
+	logrus.WithField("time_zone", tzName).Debug("queried time_zone system variable")
+	if tzName == "SYSTEM" {
+		return nil, errDatabaseTimezoneUnknown
+	}
+
+	// If it looks like a numeric offset then construct a fixed-offset time zone from that.
+	if timeZoneOffsetRegex.MatchString(tzName) {
```
jgraettinger, line 274: Should we just ... not have this behavior? It's extra code to maintain that could have subtle bugs, it's more documentation to write and to communicate to users, and it doesn't seem unreasonable to simply require an IANA time zone name.

If we do keep it, it looks like it needs test coverage (only IANA case is covered currently)

jgraettinger (commented):
Looking great, just a question & comment. Heads up there's a test failure in CI as well

source-mysql/main.go
willdonnelly: That sounds exactly right. The read is of the global setting of `time_zone` which should apply to all sessions by default, and then we modify it for just our session so that the DB will do the right thing and give us `TIMESTAMP` values in UTC.

source-mysql/main.go
willdonnelly: I included this logic because it seemed reasonable to support any legal setting of the `time_zone` variable that wasn't `'SYSTEM'`.

I'm not super attached to keeping it, but it feels worthwhile to me to support as many customer database configurations as we reasonably can, and parsing the numeric offset isn't super complicated (and in fact now that I'm well rested I've realized we can get Go to do the parsing with something like `t, _ := time.Parse("-07:00", tzName); t.Location()` so that avoids hand-written parsing code here).

So I'm inclined to keep this support but replace the hand-written parser with Go's `time.Parse` and then add a new test case which exercises DATETIME support in more time zones than just `America/Chicago`.

source-mysql/main.go
willdonnelly: I've tweaked the comment to hopefully clarify what's going on for future readers.

willdonnelly:
Currently looking into the CI failure. It looks like tests exercising `TIMESTAMP` round-tripping are failing, for what is in hindsight a fairly obvious reason: the tests set the **global** time zone but don't modify the **session** time zone of the connection used for inserting test data, and so that just inherits whatever the server's global time zone was when we launched the current test suite run.

And I never noticed that locally because after running a particular sub-test once the global setting is correct for that specific test if you run it again against the same database. So I think the fix is just to set the session timestamp (for the test-data-inserting connection, which is distinct from the capture backfill connection that we set to UTC so the capture behaves properly. So many distinct variables...)

willdonnelly:
Looks like everything is passing CI now, and with the new `TestDatetimes` test I'm fairly confident we've actually got this correct. Let me know if there are any time zone edge cases that ought to be added there, currently I've got:

- Named time zone which observes DST (with dates in both the summer and the winter)
- Named time zone which does not observe DST
- Explicit offset time zone

jgraettinger (approved):
LGTM!

PR #441, opened 2022-12-01
materialize-*: add x-collection-name to resource spec schemas

mdibaiee:
**Description:**

- Add `x-collection-name` annotation to materialization resource spec schemas, which will be used by the UI to fill in the collection name automatically

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/441)

### DIFF
--- a/go-schema-gen/generate.go
+++ b/go-schema-gen/generate.go
@@ -26,7 +26,7 @@ func GenerateSchema(title string, configObject interface{}) *jsonschema.Schema {
 	schema.Title = title
 	walkSchema(
 		schema,
-		fixSchemaFlagBools(schema, "secret", "advanced", "multiline"),
+		fixSchemaFlagBools(schema, "secret", "advanced", "multiline", "x-collection-name"),
 		fixSchemaOrderingStrings,
 	)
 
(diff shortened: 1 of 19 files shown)


willdonnelly (commented):
LGTM other than a bit of bikeshedding:

What's the rationale for calling it `x-collection-name` rather than simply `collection-name` or maybe `flow-collection-name`? I We already have several other nonstandard (AFAIK) annotations (`secret`, `advanced`, etc) and [X-Prefixes are deprecated](https://www.rfc-editor.org/rfc/rfc6648) in general these days.

mdibaiee:
@willdonnelly we have code that handles `X-` Annotations generically: https://github.com/estuary/flow/blob/f0604a13970933e98522c2f09bcab6193d04e72a/crates/doc/src/annotation.rs#L25-L27

I wasn't aware that this convention is explicitly deprecated. I think of it as a useful distinguisher between JSONSchema's core and custom annotations

willdonnelly:
> we have code that handles x- annotations generically

Oh neat, I wasn't aware of that. That does change things a bit, if the `x-` prefix [1] lets us avoid having to change the Flow schema-parsing code as well.

[1] Fun fact, apparently typing `x-` activates my muscle memory and I'm unable to not type `x-1` by default.

Issue #444, opened 2022-12-05
sqlcapture: Reduce backfill memory usage


willdonnelly:
The recent push to bound memory usage at 1GiB or below puts SQL backfills in a somewhat difficult position. To recap:

- Our correctness strategy relies on reading K rows from the database at a time, covering a particular span of the primary key range, buffering the resulting rows in memory, and then processing replication events until a newly written watermark is reached (which guarantees that we're now fully caught up to a time past the backfill query) while merging any updates observed to row keys falling within the buffered range.
- This requires that we retain K rows worth of data in memory, for some tunable value of K. Actually it's a bit worse than that, as we backfill all tables in parallel and so it's actually K times the number of tables currently being backfilled. Also in the (somewhat rare) edge case of a table with lots of inserts occurring at new random points within the keyspace, any new inserts within that chunk's range will add to the total.
- This can in theory be made to work. Our memory allowance at present is a fairly generous 1GiB, and a typical database row is like 2-4kB of data, so in principle we could set the chunk size to something like 4,096 and still fit many simultaneous tables in memory.
- However it's also tempting to make the buffer-size constants really large for efficiency, because that maximizes throughput from the database especially under conditions of high round-trip latency.

So my feelings at the moment:

- We should go ahead and revert the work I did on throughput-optimal buffer sizing and drop those down to much more modest numbers. This doesn't actually have a lot of effect in low latency setups, and the rest of the time...it's better in my opinion to have something that works reliably, and we have a config setting so we can override the defaults in cases where we both care about a slightly reduced throughput and we're sure it will actually work.
- We should think about explicitly managing the various buffers on disk using RocksDB or BoltDB or something, to offload them out of memory entirely. In the common case we should expect to pretty much just stream them out to disk in sorted order, make few or zero changes while replication catches up, and then read them back in arbitrary order, so it should be possible to do this quite efficiently.

willdonnelly:
Most of this was done. Closing this issue in favor of https://github.com/estuary/connectors/issues/775 which more directly tracks the only remaining work: moving backfill result buffering out of memory and into disk instead.

PR #446, opened 2022-12-05
source-{mysql,postgres}: Lower buffer-size constants

willdonnelly:
**Description:**

This PR lowers various buffer-size tuning constants in `source-mysql` and `source-postgres` so that we can expect captures to actually work properly within the 1GiB connector memory limits. This comes at the cost of lowered throughput in some possible setups, but in the common case the hit shouldn't be too terrible.

Also this PR makes the backfill chunk size an advanced config property of `source-postgres` the same way we already did for MySQL, just for consistency.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/446)


willdonnelly:
I filed a tracking issue and then completely forgot to mention it in the pull request :/

https://github.com/estuary/connectors/issues/444

williamhbaker (approved):
LGTM

Issue #450, opened 2022-12-06
source-firestore: Better behavior when replication fails to catch up


willdonnelly:
Our connector `source-firestore` uses the gRPC `Listen` RPC to stream change events from Firestore. Most of the time this works well.

Unfortunately, we need a resume cursor to pick up at the appropriate point after a restart, Firestore doesn't send us a new resume cursor until after we're fully caught up with the present state of the dataset, and Firestore also requires that we become caught up within 10 minutes or it closes the response stream on us.

This means that if the number of changes to sync is too large (in the realm of >1M changed documents), it might simply be impossible for us to ever become caught up via replication. This is why we introduced the `async` and `none` backfill modes, which both avoid using replication to backfill historical contents of the dataset. However if the connector is offline for a while (and also theoretically if a massive number of new documents are dumped into the dataset all at once) we might reach a point where an ongoing replication stream falls behind and suffers the same issue.

Currently we handle this in the worst possible way, by failing the capture and then just trying again when the connector gets restarted. But each of these retries spends 10 minutes slurping down data as fast as we possibly can, so this ends up producing _a lot_ of read traffic if not noticed and disabled.

We can and should handle this better:

- As an initial stopgap, failure to complete that initial sync should not cause the capture to terminate and restart. Instead we should log a warning, forget about that particular collection for the duration of the current connector run, and **keep going**. This converts the pathological "maximum data rate for 10 minutes, over and over" failure behavior into a more benign (and also more noticeable via monitoring) "no data until the capture restarts, with a single massive spike whenever it restarts".
- Later on we should implement some retry logic, so that the `async` backfill mode can fully recover by jumping the replication stream forward to "now" and triggering a full backfill to pick up anything we might have missed. The "max by `mtime`" reduction behavior of Firestore output collections should take care of any redundancy, but we can also prune the data volume somewhat by omitting any documents from the "recovery backfill" whose mtime predates the last time we were caught up via replication.

willdonnelly:
We've been picking away at Firestore capture stability over time:

- https://github.com/estuary/connectors/pull/825
- https://github.com/estuary/connectors/pull/827
- https://github.com/estuary/connectors/pull/909
- https://github.com/estuary/connectors/pull/912
- https://github.com/estuary/connectors/pull/936
- https://github.com/estuary/connectors/pull/939

And I think we're finally in a tolerably stable state.

[...]

PR #451, opened 2022-12-06
source-firestore: Don't error out if catchup streaming fails

willdonnelly:
**Description:**

See https://github.com/estuary/connectors/issues/450 for a complete summary of the problem, but TL;DR: if replication fails to catch up within 10 minutes Firestore will cut us off, but if we fail the capture then we'll just retry again when the connector gets relaunched in a few minutes. And on monitoring that sure looks like a connector making progress!

So if we instead ignore the issue, that's also not ideal but at least it's not busy-looping and racking up massive read traffic the whole time.

**Notes for reviewers:**

I'm in the process of verifying this locally (it's fairly simple to test, just takes a while since we have to wait 10 minutes for Firestore to cut us off). Once that's done this should be good to go.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/451)

### DIFF
--- a/source-firestore/pull.go
+++ b/source-firestore/pull.go
@@ -480,11 +480,11 @@ func (c *capture) BackfillAsync(ctx context.Context, client *firestore.Client, c
 	return nil
 }
 
-func (c *capture) StreamChanges(ctx context.Context, client *firestore_v1.Client, collectionID string, startTime time.Time) error {
+func (c *capture) StreamChanges(ctx context.Context, client *firestore_v1.Client, collectionID string, readTime time.Time) error {
 	var logEntry = log.WithFields(log.Fields{
 		"collection": collectionID,
 	})
-	logEntry.WithField("startTime", startTime.Format(time.RFC3339)).Info("streaming changes from collection")
+	logEntry.WithField("readTime", readTime.Format(time.RFC3339)).Info("streaming changes from collection")
 
 	var target = &firestore_pb.Target{
 		TargetType: &firestore_pb.Target_Query{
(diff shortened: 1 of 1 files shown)


source-firestore/pull.go
```diff
@@ -577,6 +580,10 @@ func (c *capture) StreamChanges(ctx context.Context, client *firestore_v1.Client
 					return fmt.Errorf("unexpected target ID %d", tc.TargetIds[0])
 				}
 			case firestore_pb.TargetChange_REMOVE:
+				if catchupStreaming && time.Since(catchupStarted) > 5*time.Minute {
```
williamhbaker, line 583: It's not immediately obvious to me why this is 5 minutes, and why it is happening at `TargetChange_REMOVE`. I could surmise that `TargetChange_REMOVE` is what happens when we get cut off for taking too long. Any reason for this being 5 minutes instead of 10 minutes?

willdonnelly: This is pretty much just a heuristic that seemed reasonable to me. What I'm trying to do is detect the specific scenario of "there is a ton of data and we failed to catch up in time", but there are potentially other reasons Firestore might do a `TargetChange.REMOVE` to us, so I put in a heuristic of "it's been a little while" around that.

The precise value is kind of unimportant, but I don't want it to be too close to zero or to the actual threshold, because either of those seemed more likely to result in false positives or negatives. So I went with the midpoint.

williamhbaker: Ah ok - makes sense! Thanks.

williamhbaker (approved):

PR #455, opened 2022-12-07
source-postgres: Ignore the 'cron' schema

willdonnelly:
**Description:**

This PR tweaks `source-postgres` discovery logic to ignore tables in the `cron` schema, which is added by the `pg_cron` extension and generally isn't something the user actually intends to capture.

Also moves the warning when no tables are discovered from the source-mysql discovery code into the generic sqlcapture discovery code so that source-postgres will warn as well.

**Workflow steps:**

It will now be safe to blindly click through the discover-and-publish workflow in the UI when `pg_cron` is installed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/455)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -75,14 +75,6 @@ func (db *mysqlDatabase) DiscoverTables(ctx context.Context) (map[string]sqlcapt
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

PR #456, opened 2022-12-07
source-mysql: Derive default 'node_id' from capture task name

willdonnelly:
**Description:**

This PR changes `source-mysql` to derive its default value for `/advanced/node_id` by hashing the task name, where previously it just hard-coded `u32("Flow")` as the default value.

The issue with a hard-coded default is/was that if a user naively creates two captures from the same MySQL host (which may be required if they want to capture from distinct logical databases) the captures will fight each other for the same ID and crash a lot.

But we don't want to just use pure randomization because some DB admins might want us to use a stable ID even if the user doesn't care to specify one. So instead of being truly random this changes the default behavior to use the upper 32 bits of the SHA-256 hash of the capture task name, which is effectively random but also stable for any particular task.

**Notes for reviewers:**

I vaguely recall somebody (maybe @psFried or @jgraettinger ?) was opposed to pure randomization here, and I think the reasoning was that some DB admins might actually care about our capture having a consistent ID? But now that we're using the Flow gRPC protocol we actually have access to the capture task name, so this ought to work, I think.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/456)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -2,7 +2,9 @@ package main
 
 import (
 	"context"
+	"crypto/sha256"
 	"crypto/tls"
+	"encoding/binary"
 	"encoding/json"
 	"fmt"
 	"net"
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM

Issue #161, opened 2022-04-06
Rockset connector assorted issues


psFried:
Yesterday I had the opportunity to meet with KW from the Rockset team, and he had a bunch of great feedback on our Rockset connector. Here's the list of actionable issues from that:

- There's an [official Go client library](https://pkg.go.dev/github.com/rockset/rockset-go-client) that we could be using, which would resolve a few of these issues. We should probably use that instead of our hand-rolled client.
- The [base URL](https://github.com/estuary/connectors/blob/16bb090ac3fcec2965d22a863219e2cae828dfb2/materialize-rockset/rest_client.go#L21) that we're using is specific to the us-west region. There are multiple other regions, and our connector would not work for customers in those regions. The Go client library handles this automatically.
- We currently handle rate limiting errors by failing and letting the shard supervisor restart the connector. We should instead apply a backoff based on the response status (also handled by the Go library).
- When adding documents via the Write API, it's possible that the response body will contain errors for a subset of added documents, while the HTTP status code is still 200. In order to know that documents were added successfully, we need to be checking the response body in addition to the status code. (The Go client library returns these as `[]DocumentStatus` so we can check them easily)
- We don't currently allow setting many options for created collections. We assumed that users would prefer to use the Rockset UI if they needed to specify a bunch of custom options. But it seems that many Rockset users already create highly customizeed collections via automation, and so we should perhaps expose a more complete set of options for collection creation.
- We currently apply Flow `projections` to the materialized document. This is a little weird for Rockset users because it overlaps with the [Ingest SQL Transformations](https://rockset.com/docs/rollups/) feature in Rockset. We should probably change the constraints returned by `Validate` to only recommend the `flow_document` (any location with an empty pointer = `LOCATION_RECOMMENDED`), and leave all other locations as `FIELD_OPTIONAL` so that a default materialization will simply pass the entire Flow document to Rockset.

PR #457, opened 2022-12-07
materialize-rockset: usability updates & fixes

williamhbaker:
**Description:**

This PR includes several updates & fixes to the `materialize-rockset` connector:
- The was no way to specify a base region URL. Rockset supports [several regions](https://rockset.com/docs/rest-api/), and without this URL specified the connector would default to the us-west region. This would cause problems if the rockset deployment was active in a different region for obvious reasons.
- As a cosmetic enhancement, the "initialize from s3" config was moved to below the advanced configurations, to make it more clear that this too is an advanced configuration. Otherwise in the UI it appeared that it was perhaps something that should be used all the time, as opposed to an advanced, optional configuration.
- An additional cosmetic enhancement: `workspace` and `collection` are required fields for the connector to validate, so this makes that reflected in the JSON spec. The connector would reject configs without those values before. This makes them show up as required in the spec itself and UI.
- The advanced configuration options `event_time_info` & `insert_only` have been deprecated per the Rockset API docs and are no longer used, so these have been removed from our connector config.
- The rockset go client has been updated to the latest version. This version of the client correctly handles base region URLs as they are listed in the rockset docs.
- Errors encountered by worker goroutines spawned to add documents per binding to rockset were not being returned correctly - they would almost always return "Context Cancelled", unless by some miracle (in the case of multiple bindings) the actual error message raced to the one returned by the `ctx.Err()`. This PR fixes that, and that error message working correctly allowed by to identify the next thing...

Presently there is an open issue with the rockset go client because its automatic retrying of `429` errors is not working. This means that a materialization containing a large amount of documents may fail on a rate limiting error, and get into a continuous loop of restarting the same transaction when it restarted every 5 minutes and never being able to complete. When the rockset go client is updated, we can update our connector to use it and fix that. Until then, I did prepare some code to automatically handle these errors on our end, but since it seems like a fix may be imminent on the rockset client side, I'm holding off on including that here for now. Tracking issue for upgrading the client: https://github.com/estuary/connectors/issues/459

**Workflow steps:**

Usage of the `materialize-rockset` connector has changed per the above description.

The [rockset materialization docs](https://docs.estuary.dev/reference/Connectors/materialization-connectors/Rockset/) should be updated:
- Add the new `region_base_url` field
- Remove the configs for `insert_only` and `event_time_info`
- Update `workspace` and `collection` of the resource config to show that they are required

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/457)


materialize-rockset/.snapshots/TestRocksetDriverSpec-TestEndpointSpecSchema
```diff
@@ -2,16 +2,30 @@
   "$schema": "http://json-schema.org/draft/2020-12/schema",
   "$id": "https://github.com/estuary/connectors/materialize-rockset/config",
   "properties": {
+    "region_base_url": {
```
psFried, line 5: Back when we reviewed the first version of the connector with them, we wrote up #161. As mentioned in the issue, their recommendation was to let the client determine the region automatically. I'm guessing you found some problems with that? I'm wondering (hoping) whether this is another case where we should report it and see if they can fix it in their client library. It would be unfortunate for us to have to keep updating these urls every time Rockset expands to a new region or cloud provider.

psFried (commented):
LGTM % discussing whether the base_url thing is a bug in the client.

materialize-rockset/.snapshots/TestRocksetDriverSpec-TestEndpointSpecSchema
williamhbaker: I've confirmed that you do need to set a base URL now. Previously the client would just use the default, which was a US west cluster, but now that there are more clusters it cannot automatically determine which one to use and will error if a base url isn't set. This is the intended behavior of the client & API.

Making this an enum list is a little overkill, but I wanted to make it as smooth as possible to get working in the short term since I could see there being a lot of friction in getting the URL correct without a list to pick from. I think we'll want to make it free-form entry in the future if more URLs added and we have some traction with the connector.

Issue #447, opened 2022-12-06
materialize-sql: allow updating fields to be nullable


mdibaiee:
At the moment we do not allow updates to columns to make them nullable. This seems like an innocent change with no adverse side-effects.

So we can potentially allow our SQL materializations to validate new projections if a column now has `null` as an additional type, and during `ApplyUpsert` we can update the table to mark the column as nullable.

Also see: https://estuaryworkspace.slack.com/archives/C012S5U0WP2/p1670331079792329

PR #458, opened 2022-12-08
materialize-sql: create all non-pks as nullable

mdibaiee:
**Description:**

- All columns except primary keys are made as nullable. this is to allow fields to be made nullable from a collection without causing issues. Once we support automatic schema migration of tables to mark fields as nullable after creation, we can revert back to the more strict behaviour, see: https://github.com/estuary/connectors/issues/447

**Workflow steps:**

- Create a new SQL materialization with non-pk fields that are not marked as nullable, but they will still be created as nullable

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/458)

### DIFF
--- a/materialize-postgres/.snapshots/TestSQLGeneration
+++ b/materialize-postgres/.snapshots/TestSQLGeneration
@@ -1,9 +1,9 @@
 --- Begin "a-schema".target_table createTargetTable ---
 
 CREATE TABLE IF NOT EXISTS "a-schema".target_table (
-		key1 BIGINT NOT NULL,
-		"key!2" BOOLEAN NOT NULL,
-		"nullableKey" TEXT NOT NULL,
+		key1 BIGINT,
+		"key!2" BOOLEAN,
+		"nullableKey" TEXT,
 		"Camel_Case" BIGINT,
 		"a Time" TIMESTAMPTZ,
 		"array" JSON,
@@ -32,8 +32,8 @@ COMMENT ON COLUMN "a-schema".target_table.flow_document IS 'user-provided projec
 --- Begin "Delta Updates" createTargetTable ---
 
 CREATE TABLE IF NOT EXISTS "Delta Updates" (
-		"theKey" TEXT NOT NULL,
-		"nullableKey" TEXT NOT NULL,
+		"theKey" TEXT,
+		"nullableKey" TEXT,
 		"aValue" BIGINT
 );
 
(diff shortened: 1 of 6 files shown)


materialize-sql/table_mapping.go
```diff
@@ -128,6 +128,7 @@ func ResolveTable(shape TableShape, dialect Dialect) (Table, error) {
 	}
 
 	for _, key := range shape.Keys {
+		key.IsPrimaryKey = true
```
mdibaiee, line 131: This was a bit unsettling, the projections for table keys were not being marked with `IsPrimaryKey`, I made it so, but I wonder if there was a reason they were not flagged as `IsPrimaryKey`? @willdonnelly

materialize-postgres/.snapshots/TestSQLGeneration
```diff
@@ -3,13 +3,13 @@
 CREATE TABLE IF NOT EXISTS "a-schema".target_table (
 		key1 BIGINT NOT NULL,
 		"key!2" BOOLEAN NOT NULL,
-		"nullableKey" TEXT,
+		"nullableKey" TEXT NOT NULL,
```
mdibaiee, line 6: We used to create key columns as nullable if they had `null` in their type, was this expected?

materialize-sql/table_mapping.go
willdonnelly: I'm not aware of any reasons, though I'm also not super familiar with how that logic has evolved. Just from skimming the relevant code it looks like the `IsPrimaryKey` field of a `Column` is never actually used for anything -- the primary-keyness is implicit in the column being part of `shape.Keys` here. So it's probably just an oversight.

materialize-postgres/.snapshots/TestSQLGeneration
willdonnelly: Key columns were previously allowed to be null if the input schema permitted nulls, regardless of whether they were singular or compound. I'm pretty sure (though not 100% certain) that was deliberate.

mdibaiee: Hmm, but flow itself does not allow keys to be nullable, right? I can't think of a scenario where a materialization connector will actually be able to materialize a document with nullable keys 🤔

williamhbaker: I believe it is planned for keys to be nullable in flow. The new `materialize-postgres` connector has [some code] (https://github.com/estuary/connectors/blob/e2526149ab5ad42a47bb8a7eefc3e113b722012e/materialize-postgres/sqlgen.go#L187-L191) to allow that even, although it is not supported in the runtime yet. Also see https://github.com/estuary/flow/discussions/643 for other background discussion about nullable keys.

There may be an argument for not applying the `not null` requirement to any column at all at this point, even primary keys. If we implemented nullable key support prior to materialization schema migration, we could end up with broken materializations because a schema key gets changed to nullable. But if we get around to implementing schema migration before nullable keys it wouldn't matter. One caveat is that I'm pretty sure in the case of postgres at least setting the primary key automatically sets `not null` if the primary key is a single column; not sure how it works for composite keys.

mdibaiee:
>  Did you do a test with a materialization created from the old version with not null values, and then transition to a materialization with the new connector? I'm honestly not sure how the projection validation part works when it loads an existing spec from the table, so I'm wondering if these would transition smoothly or not.

I tested this and it works fine (again I think as long as the constraints stay the same, we are fine in validation step).

willdonnelly (approved):
LGTM

PR #463, opened 2022-12-13
source-firestore: Reduce backfill chunk size

willdonnelly:
**Description:**

We're seeing Firestore captures failing in production due to excessive memory usage, and this is the dumbest hammer that we have to potentially reduce memory usage so it's worth trying.

In general our backfills tend to be constrained by journal append rate limits and RTTs to Firestore are fairly low, so this isn't expected to have a large impact on throughput in general.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/463)

### DIFF
--- a/source-firestore/pull.go
+++ b/source-firestore/pull.go
@@ -361,7 +361,7 @@ func (c *capture) Run(ctx context.Context) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #464, opened 2022-12-15
ci: add gh-webhook

mdibaiee:
**Description:**

- add https://github.com/estuary/flow/pull/849 to ci

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/464)


jshearer (approved):
So on _every_ push we re-pull the `:dev` tag as well as the actual tag? Do we frequently want to have `:dev` set to some other build for testing, which would get accidentally overwritten? 

I guess unless we're doing multiple concurrent streams of work on a connector that wouldn't be an issue -- once we're done `:dev`eloping, we'll push to the actual tag and stop caring about `:dev`... so this is probably fine.

There are also likely a few connectors that don't have the `:dev` tag pushed, which will result in the request... silently failing, which is probably fine.

Issue #459, opened 2022-12-08
materialize-rockset needs to handle 429 errors


williamhbaker:
When the current connector hits a rate limit error `429`, it the `Transactions` call and the connector exits. It then gets restarted, trying the transaction from the beginning again, and hits the rate limit again, crashing in a loop and never making progress. 

This seems to stem from the rockset go client itself not retrying `429`'s as intended. Once the rockset go client is updated, we should update our connector to use it.

PR #466, opened 2022-12-19
materialize-rockset: update rockset client to v0.15.4

williamhbaker:
**Description:**

This updates to the latest Rockset client, which includes a fix for retrying rate limit errors. With this updated client, retryable rate limit errors will be automatically handled by the client, rather than returning a terminal error which would crash the connector and cause re-trying of the transaction in a potentially infinite loop.

Closes https://github.com/estuary/connectors/issues/459

**Workflow steps:**

n/a

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/466)


willdonnelly (approved):
LGTM

PR #468, opened 2022-12-19
sqlcapture: Row-key serialization rewrite

willdonnelly:
**Description:**

Our SQL capture connectors keep track of backfill progress by means of a "scan key" (which is usually but not necessarily the primary key of the table) comprising one or more columns of the table being captured. A particular value of the scan key is serialized into an FDB-encoded sequence of bytes, which can then be persisted into a state checkpoint, decoded back into values for a subsequent backfill query, and compared bytewise with other scan key values to establish ordering.

Previously the handling of these keys was somewhat unprincipled, in that this serialization took place _after_ the JSON-output transform was applied to the values coming from the database driver. This mostly worked, because the JSON-output transform turned numbers and strings into the native Go numeric/string types, and most of the fancy domain-specific database types tend to be turned into human-readable JSON strings whose formatting matches something the database will accept back as input.

But strictly speaking it was still wrong to do it this way, and we hit a troublesome edge case in the handling of the MySQL `TIMESTAMP` and `DATETIME` types (which we format as RFC3339 timestamps for output, but which MySQL wants to look like `2022-12-19 16:40:12` at input). This PR refactors things so that the two serialization pathways are entirely decoupled, and then writes a bunch of tests to validate that scan key round-tripping through the encode/decode paths actually works, both in general and for a host of specific column types.

**Workflow steps:**

Theoretically nothing should necessarily have to change. There was very little value transformation happening on the _decoding_ side when working with scan keys, and so any capture which previously worked should continue to work.

None

**Notes for reviewers:**

This is probably best reviewed commit-by-commit:

- The first two commits in the series are just refactoring with no external impact
- The third commit (`sqlcapture: Move row-key encoding into DB-specific logic`) makes the primary functional change
- Then the subsequent five commits add tests (after I realized that really that change should have broken a lot more than it did) and fix a few bugs as they're turned up by the improved test coverage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/468)


source-postgres/backfill.go
```diff
@@ -94,25 +118,36 @@ func (db *postgresDatabase) WatermarksTable() string {
 	return db.config.Advanced.WatermarksTable
 }
 
-func (db *postgresDatabase) buildScanQuery(start bool, keyColumns []string, schemaName, tableName string) string {
-	// Construct strings like `(foo, bar, baz)` and `($1, $2, $3)` for use in the query
-	var pkey, args string
+// The set of MySQL column types for which we need to specify `COLLATE "C"` to get
```
williamhbaker, line 121: In the comment, perhaps "MySQL" should be "Postgres"

williamhbaker (approved):
LGTM!

source-postgres/backfill.go
willdonnelly: I swear, copy-pasting that "do we need some string key comparison" logic introduced no less than 3 distinct flaws. I should have just typed it out by hand the second time :/

Fixed

willdonnelly:
Going to hold off on merging this until tomorrow morning; this PR deliberately doesn't bump the connector version, and I believe that it should be entirely safe [1], but I'd rather things break during the work day if I got something wrong.

[1] There are two reasons I believe this is safe. First, this should have zero impact on any table which has completed its backfill, and new captures would break just as readily after a connector version bump. So only ongoing backfills should be a concern, and we don't have any of those because in general backfills complete _pretty fast_. Secondly, even if a backfill were ongoing and the connector restarted to pick this change up the decode behavior shouldn't be adversely impacted, because the decode path is mostly unaltered and so any value persisted by the old encode logic should get decoded and fed into the next chunk's query the same as before.

PR #473, opened 2022-12-21
sqlcapture: Disable scan key ordering check as a temporary hack

willdonnelly:
**Description:**

This is a new safety check added as part of https://github.com/estuary/connectors/pull/468 but which breaks the rollout of said PR in at least one case.

That case is when the scan key of an ongoing MySQL backfill contains a `DATETIME` column, which previously was serialized as an RFC3339 timestamp (with 'T' instead of ' ' separator and with the hour offset to UTC) and after the aforementioned PR these values are now serialized as the raw text format that MySQL prefers. The problem arises when we compare the old scan key (coming from the initial state checkpoint after the connector restarted) to a new one (the first key of the next backfill chunk after the restart).

In general we double-check the key ordering to help maintain the invariant that bytewise scan-key ordering matches the database's own concept of "greater than" and `ORDER BY`, so in this one case it's safe to just disable the check and keep going.

Once all captures with ongoing backfills are successfully restarted with the new key encoding logic, it will be safe to remove this hack again. Expect that in an hour or so.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/473)

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -484,7 +484,18 @@ func (c *Capture) backfillStreams(ctx context.Context, streams []string) (*resul
 			// key value is greater than the previous cursor value. This together
 			// with the check in resultset.go ensures that the scan key always
 			// increases (or else the capture will fail).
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #475, opened 2022-12-21
source-alpaca: new connector

williamhbaker:
**Description:**

Creates a capture connector for pulling data from the [Alpaca API](https://alpaca.markets/), specifically for stock prices.

The connector collects data in 2 ways: Historical backfilling for a comprehensive collection of trade data starting at a certain point in time, and real-time streaming to get the fastest possible collection of new trade data.

The historical backfill walks through time from the specified start date up to the present, and continues walking through chunks of historical data to keep up with the current time once it has completed the backfill. This will fill in any gaps from the real-time streaming client, as the websocket delivery mechanism is "best effort" with no way to resume the stream from a disconnected client. The streaming client is started once the historical backfill has caught up initially and entered its "wait a little bit and get a little more data" looping phase.

Both of these collection methods running in parallel with generally produce two identical documents for each trade. Trades documents have IDs to uniquely identify them (along with the exchange, date, and symbol - a given symbol on a given exchange will have exactly 1 trade with a specific ID on a given date), so the captured collection should be handled as "last write wins", with the potential for duplicates. Materializing from a collection like this in non-delta mode should be transparent, but when materializing with delta-updates the endpoint will need to allow for these duplicate keys.

I have done a fair amount of testing to verify that the documents produced by the streaming API are identical to the real time API, and so far they are: Any document from the streaming API has an exact copy (sans the flow uuid) in the dataset produced by the historical API covering that same time period.  

**Workflow steps:**

The connector can be configured from the UI, or using a simple config with the CLI like this:
```yaml
  bobCo/alpaca-capture:
    endpoint:
      connector:
        image: "ghcr.io/estuary/source-alpaca:dev"
        config:
          api_key_id: <SECRET>
          api_secret_key: <SECRET>
          feed: iex
          start_date: 2022-11-01T00:00:00Z
          symbols: AAPL,MSFT,AMZN,TSLA,GOOGL,GOOG,NVDA,BRK.B,META,UNH
          advanced:
            is_free_plan: true
    bindings:
      - resource:
          name: trades
        target: bobCo/alpaca-capture/trades
```

Docs will need to be created for this new connector. The `README.md` included in this PR might be a useful starting point.

**Notes for reviewers:**

The connector in the state of this PR works well for handling a relatively small number of symbols. I've been able to capture the top 10 by market cap on the NYSE for example when running locally. The backfill for the entire day of 2022-12-20  from the SIP feed peaked at about ~5000 documents per second, pulling ~5-10 MB/sec of data on my local network. In total the capture wrote 4,243,372 documents / 862 MB of data for all of the trades from those stocks for that entire day. This was with a free plan, and I suspect that I was hitting API rate limits, so a paid plan might be able to go even faster. 

Further enhancements that should be prioritized include:
- Reducing the number of or completely eliminating the presence of duplicated items from the streaming vs. backfill data gathering. Complete elimination may be impossible since I am not confident that there are any ordering guarantees in the events received from the websocket client. It may be possible to implement a size-limited cache of trades seen emitted by the streaming client, and avoid emitting separate documents for the backfiller.
- Allowing for sharding based on a consistent hash of symbols. I can easily see use cases of tracking large numbers of symbols requiring multiple shards to capture in a reasonable way, and we should be able to handling sharding in the connector based on a hash of the symbols.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/475)


source-alpaca/.snapshots/TestSpec
```diff
@@ -0,0 +1,128 @@
...
+        "secret": true
+      },
+      "feed": {
+        "type": "string",
+        "enum": [
+          "iex",
+          "sip"
+        ],
+        "title": "Feed",
+        "description": "The feed to pull market data from. May be overridden within the binding resource configuration.",
```
willdonnelly, line 25: It feels like the whole "specifying feed/symbol in the endpoint config and then also/optionally in each resource" thing is going to be a source of pain in the future, but honestly I don't have a recommendation for how it could be done better.

source-alpaca/main.go
```diff
@@ -0,0 +1,135 @@
...
+	boilerplate "github.com/estuary/connectors/source-boilerplate"
+)
+
+const (
+	defaultMaxBackfillInterval = 15 * time.Minute
+	defaultMinBackfillInterval = 1 * time.Minute
+)
+
+type resource struct {
+	Name    string `json:"name" jsonschema:"title=Name,description=Unique name for this binding. Cannot be changed once set."`
```
willdonnelly, line 17: Something about the need for an explicit name here feels like maybe the domain modeling is off. Some random thoughts:

My naive perspective is that a given feed+symbol pairing is its own distinct stream of data which can be captured into a collection, and combining multiple symbols into a single resource binding is the cause of a lot of the weirdness around endpoint/resource configs here. And I'm reasonably sure one can have multiple capture bindings going into the same collection, so no issue there.

Discovery would be troublesome as currently implemented in the UI. But in a model where each stock symbol is a distinct resource there's actually a pretty nice coherent implementation of a discovery UI where we just return a fixed list of some tens of thousands of possible resources and give the user a search box. I think something like that is on the eventual roadmap for other data sources with a ton of different streams?

source-alpaca/pull.go
```diff
@@ -0,0 +1,156 @@
...
+	} else {
+		// If backfilling is disabled, there's nothing to catch up on. Streaming can start right away.
+		close(caughtUp)
+	}
+
+	if !c.Config.Advanced.DisableRealTime {
+		eg.Go(func() error {
+			// Wait until the backfilling is caught up (or disabled) before starting streaming.
+			// Throughput will most likely be limited by the network or journal append limits. It
+			// may be possible that there is such a large amount of data that the backfilling will
```
willdonnelly, line 143: Is it possible to log a warning/error in the event that backfilling looks like it will probably never catch up? I haven't worked through all the edge cases but something like:

```go
if itemDelay := time.Since(item.Trade.Timestamp); itemDelay > (prevDelay + time.Hour) {
  log.WithField("delay", itemDelay.String()).Warn("backfill delay increasing")
  prevDelay = itemDelay
} else if itemDelay < prevDelay {
  prevDelay = itemDelay
}
```

That is, keep track of some rough estimate of "how far in the past we are" just by looking at each trade document, and log a warning if that number appears to be increasing substantially over time.

willdonnelly (approved):
LGTM modulo a couple of nits around the endpoint/resource config model.

source-alpaca/pull.go
williamhbaker: Seems like a good idea - I'll add logging so that if a backfill cycle takes longer to complete than the actual data period it covers, there will be a warning message emitted. This would be indicative of the backfill being unable to catch up.

source-alpaca/.snapshots/TestSpec
williamhbaker: Yeah I think this is mostly vestigial from the way I had set up the configuration originally, with much more of the config at the resource-level rather than the endpoint-level. Upon reflection, I think it is best to remove these optional overrides entirely, as I don't think they will be used and they complicate things quite a bit.

source-alpaca/main.go
williamhbaker: Functionally I think this is similar to how something like the `source-http-file` capture works, where each binding has a "stream", but the stream is just the last part of the provided URL to uniquely identify it.

The thought about having each symbol being a distinct stream is interesting. In order for data collection to be efficient, we'll need to make requests for grouped symbols. I'm not sure how that would work if each binding was responsible for only a single symbol. I'm thinking in the near future we will end up capturing entire exchanges' worth of symbols at a time, and it would definitely make sense to map an exchange to a stream. It will take some work to handle huge amounts of symbols at a time though, so I'm leaving that as a future enhancement for now.

Issue #422, opened 2022-11-17
New Image for MariaDB


dyaffe:
**Problem**
Our MySQL connector currently also covers MariaDB, but external users don't know that.

**Proposal**
Unfortunately we have a constraint that images have to be unique for connectors, so this isn't as simple as adding a new row to Supabase called "MariaDB".  Instead, we likely need to create a new docker image for MariaDB that is up to date.  From there, I can add the connector to the UI.

PR #479, opened 2022-12-23
build: update CI to allow pushing multiple tags from the same dockerfile

williamhbaker:
**Description:**

This is a CI change that will allow for easily pushing an image with a different tag from a single dockerfile build. This will allow us to have a specific tag for `source-mariadb`, `materialize-timescaledb`, etc. The images themselves are identical (they have the same digests) as their root `source-mysql` and `materialize-postgres`, they just have a different tag.

The variant images will be published on merges to `main` with the same version tag as their root connector.

Closes https://github.com/estuary/connectors/issues/422

**Workflow steps:**

Additional tags for any image can be configured in `ci.yaml` in a similar way. Multiple additional tags can be specified to be built from a single connector.

This PR will end up creating two new connectors: `source-mariadb` and `materialize-timescaledb`. I'd think we should have a docs page for each of these, but the docs should just link to `source-mysql` and `materialize-postgres` (respectively), with clarification on those docs pages that the same configuration applies.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/479)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,27 +71,31 @@ jobs:
       fail-fast: false
       matrix:
         connector:
-          - source-alpaca
-          - source-gcs
-          - source-hello-world
-          - source-http-file
-          - source-test
-          - source-firestore
-          - source-kafka
-          - source-kinesis
-          - source-mysql
-          - source-postgres
-          - source-s3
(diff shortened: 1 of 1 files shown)


mdibaiee (approved):
nice one, thank you!

PR #481, opened 2022-12-27
materialize-postgres: add config to run additional sql on table creation

### LINKED ISSUE #476: materialize-postgres: hypertables are not created when destination is timescale
We have had some use cases where the postgres materialization connector has been used to materialize into timescale.

When the connector creates tables, it does not create them as hypertables. A hypertable is necessary for timescale to be useful, and it is cumbersome for the user to pre-create their table so that the hypertable is in effect for the initial data that is materialized into it.

We should add support for creating tables as hypertables, either into the existing postgres materialization as an advanced option, or somehow forking the connector into a dedicate timescale connector.

williamhbaker:
**Description:**

Adds an optional resource-level configuration for the user to provide additional SQL statements to be executed when creating a table as part of `ApplyUpsert`. This could be used for things like converting a table to a TimescaleDB hypertable before any data is written to it, creating additional indexes, etc.

Closes #476

**Workflow steps:**

As an example for materializing to TimescaleDB where you want the table that the materialization is going to create to be created as a hypertable, you would add this SQL statement to the config:
```sql
SELECT create_hypertable('flow_created_table_name', 'timestamp_column');
```

The postgres materialization docs will need updated with this new configuration. An example for creating a hypertable would be helpful under a TimescaleDB-specific heading.

**Notes for reviewers:**

For the TimescaleDB hypertable simple case, this is a less-direct way of making that possible than something like a checkbox that would cause the connector to use an alternative table creation schema that would automatically include the hypertable statement. We are opting to use the slightly less user-friendly raw SQL input configuration since it is higher leverage: It may be useful for other SQL materializations and cases beyond simple hypertables. For example, a regular postgres materialization might have use for creating additional indexes on created tables. A TimescaleDB materialization might want to create _distributed_ hypertables and/or include some of the numerous options that are possible with hypertable creation - these would not be practical for use to exhaustively maintain as individual connector config. And it makes it possible to include the necessary behaviors in a single postgres connector as opposed to forking into a TimescaleDB connector with legitimately different code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/481)

### DIFF
--- a/materialize-postgres/driver.go
+++ b/materialize-postgres/driver.go
@@ -79,9 +79,10 @@ func (c *config) ToURI() string {
 }
 
 type tableConfig struct {
-	Table  string `json:"table" jsonschema:"title=Table,description=Name of the database table" jsonschema_extras:"x-collection-name=true"`
-	Schema string `json:"schema,omitempty" jsonschema:"title=Alternative Schema,description=Alternative schema for this table (optional)"`
-	Delta  bool   `json:"delta_updates,omitempty" jsonschema:"default=false,title=Delta Update,description=Should updates to this table be done via delta updates. Default is false."`
+	Table         string `json:"table" jsonschema:"title=Table,description=Name of the database table" jsonschema_extras:"x-collection-name=true"`
+	Schema        string `json:"schema,omitempty" jsonschema:"title=Alternative Schema,description=Alternative schema for this table (optional)"`
+	AdditionalSql string `json:"additional_table_create_sql,omitempty" jsonschema:"title=Additional Table Create SQL,description=Additional SQL statement(s) to be run in the same transaction that creates the table." jsonschema_extras:"multiline=true"`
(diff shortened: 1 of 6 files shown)


materialize-postgres/driver.go
```diff
@@ -187,6 +192,8 @@ func (c client) FetchSpecAndVersion(ctx context.Context, specs sql.Table, materi
 }
 
 func (c client) ExecStatements(ctx context.Context, statements []string) error {
+	// Execute statements within a transaction.
+	statements = append(append([]string{"BEGIN;"}, statements...), "END;")
```
mdibaiee, line 196: This will have an effect on all other statements being run as well as the table creation SQL, so I think if we want to limit the blast radius we can move this `BEGIN` and `END` additions to driver.go where we are creating the table creation SQL, what do you think?

mdibaiee (approved):

Issue #320, opened 2022-08-19
Bigquery Not Handling Google Ads Data


dyaffe:
**Problem**
Google ads data by default uses column names that Big Query doesn't handle.  Attempting a materialization leads to the following:

```
time="2022-08-19T22:17:13Z" level=warning msg="job failed (will retry)" attempt=0 error="googleapi: Error 400: Illegal field name: segments.ad_network_type, invalidQuery" jobStatus="&{2 <nil> [] 0xc0004f6dc0}"
time="2022-08-19T22:17:14Z" level=warning msg="job failed (will retry)" attempt=1 error="googleapi: Error 400: Illegal field name: segments.ad_network_type, invalidQuery" jobStatus="&{2 <nil> [] 0xc0006b00b0}"
time="2022-08-19T22:17:14Z" level=warning msg="job failed (will retry)" attempt=2 error="googleapi: Error 400: Illegal field name: segments.ad_network_type, invalidQuery" jobStatus="&{2 <nil> [] 0xc00054a370}"
time="2022-08-19T22:17:15Z" level=warning msg="job failed (will retry)" attempt=3 error="googleapi: Error 400: Illegal field name: segments.ad_network_type, invalidQuery" jobStatus="&{2 <nil> [] 0xc00054a580}"
time="2022-08-19T22:17:15Z" level=warning msg="job failed (will retry)" attempt=4 error="googleapi: Error 400: Illegal field name: segments.ad_network_type, invalidQuery" jobStatus="&{2 <nil> [] 0xc00021c2c0}"
time="2022-08-19T22:17:16Z" level=error msg="job failed, exausted retries" error="<nil>"
```

mdibaiee:
Related: https://github.com/estuary/connectors/issues/295

Issue #295, opened 2022-07-25
BigQuery transforms column name projections, it shouldn't


mdibaiee:
BigQuery materialization is currently transforming projection fields:

https://github.com/estuary/connectors/blob/main/materialize-bigquery/bigquery.go#L179-L184

We had a discussion on this about Firebolt and we decided it's best not to tamper with projection names, and they should be authoritative of what the column is actually called in the destination.

In this case, it's best to quote the column names instead of replacing characters. That's the approach taken in Firebolt.

williamhbaker:
We had some discussion about this and concluded that for now we will continue transforming column name projections.

More info:

BigQuery column names have strict requirements even if they are quoted, see https://cloud.google.com/bigquery/docs/schemas#column_names. I investigated a strategy where the connector would enforce constraints that the projected names must comply with these requirements. This would work, but it would make it very difficult to use with our current capabilities to edit projections in cases where a collection has many fields and the field names contain invalid characters (spaces, colons, etc.). Our replacing of these characters with underscores is consistent with out other systems handle it too - they replace them with underscores.

As an aside, table names are a different story. These can have just about any characters if they are quoted. One exception is that they cannot contain `.`, and neither can column names. We currently don't underscore `.`, which can lead to problems. https://github.com/estuary/connectors/pull/487 fixes that, to replace `.` with underscore.

In the future if we have a system where creating/editing large numbers of projections is easy, we might be able to revisit this and have the connector enforce constraints on field names. It seems like it would be best for the connector to not transform projections (consider the case of `Field!` vs `Field?`, where both are transformed into `Field_` and there is a conflict), but we will have quite a bit of work to get to a point where that is usable.

I will leave this issue open since it might be useful down the road.

PR #487, opened 2023-01-04
materialize-bigquery: improve identifier handling

williamhbaker:
**Description:**

This PR is for a couple of bug fixes to BigQuery, and a bit of refactoring.

It is the result of investigation of https://github.com/estuary/connectors/issues/295. We have decided to continue transforming column names (see my comment on that issue), and while I was experimenting with strategies to not do that I came upon these opportunities for improvement that are still relevant.

The individual commit messages explain what is being changed, but broadly:
- There was a rare case where a materialization might not have any value fields, which would result in an invalid query
- We don't actually need to quote absolutely every identifier, so there's some cleanup there
- We should be replacing `.` with `_`, since it is not valid in column or table names.

Closes https://github.com/estuary/connectors/issues/320

**Workflow steps:**

No changes in the usage of the connector

In the [BigQuery Docs](https://docs.estuary.dev/reference/Connectors/materialization-connectors/BigQuery/) it would be good to highlight naming restrictions, and things that happen with column names:
- Field names cannot contain hyphens. The connector will fail to apply if a field name contains a hyphen. A alternate projection of the field needs to be provided in this case that does not have a hyphen in the name
- Similarly, field names must start with a letter or underscore. Otherwise the connector will fail. Same deal with an alternate projection in this case.
- Any non-alphanumeric or `_` character in a field name will be replaced by `_` in the bigquery column names for fields
- In the case of two fields that become identical after `_` replacement happens (ex: `Field!` and `Field?` both transformed to `Field_`), the connector will fail, needs an alternate projection to work.
- Hyphens are allowed for table names, but any other non-alphanumeric or `_` character in a field name will also be replaced by `_` in the materialized bigquery table names

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/487)


materialize-bigquery/sqlgen.go
```diff
@@ -54,9 +80,14 @@ var bqDialect = func() sql.Dialect {
 	}
 
 	return sql.Dialect{
-		Identifierer:         sql.IdentifierFn(sql.JoinTransform(".", identifierSanitizer(sql.QuoteTransform("`", "\\`")))),
-		UnquotedIdentifierer: sql.UnquotedIdentifierFn(sql.JoinTransform(".", identifierSanitizer(func(s string) string { return s }))),
-		Literaler:            sql.LiteralFn(sql.QuoteTransform("'", "\\'")),
+		Identifierer: sql.IdentifierFn(sql.JoinTransform(".",
+			identifierSanitizer(sql.PassThroughTransform(
+				func(s string) bool {
+					return simpleIdentifierRegexp.MatchString(s) && !sql.SliceContains(strings.ToUpper(s), BQ_RESERVED_WORDS)
```
willdonnelly, line 86: It feels a little bit off to use `ToUpper` here but `ToLower` in the Postgres equivalent.

willdonnelly (approved):
LGTM

materialize-bigquery/sqlgen.go
williamhbaker: Yeah it's kind of weird, and really just because the list on the BigQuery docs has the reserved words in all caps. I think I'll make the common `sql.SliceContains` function do a case-insensitive comparison, since it is used in this same way for all the sql materializations. That should be more consistent.

willdonnelly: That sounds reasonable. This whole identifier-quoting thing should only take place during startup and the resulting queries should be cached in normal operation, so the extra work done by using `EqualFold` should be irrelevant, right?

williamhbaker: It's technically used in the templates for creating and updating a fence which are executed anew for every transaction.

I neglected an important detail though that's making me backpedal on thinking case-insensitivity is a good idea: We also do `SliceContains` during validation to evaluate constraints for fields. I don't actually know off-hand if flow fields are case-sensitive or not but I wouldn't feel great about enforcing that either way.

So - I think I'll actually just add a comment here, re-iterating that the reserved words list for bigquery needs to be  in all caps.

Issue #493, opened 2023-01-06
S3 Connector Allowed Invalid Bucket to be Saved


dyaffe:
**Problem**
We saw an invalid bucket saved in production for task `estuary/actively/verkada_leads/source-s3`.  The connector should catch that and provide an error message.

Here is the current error message:

```
runTransactions: readMessage: connector failed (exit status: 1) with stderr:
{"timestamp":"2023-01-06T17:55:04.215219391Z","level":"INFO","message":"sweeping s3://actively-verkada/leads// starting at "", from 0001-01-01 00:00:00 +0000 UTC through 2023-01-06 17:54:34 +0000 UTC"}
prefix s3://actively-verkada/leads//: during listing: InvalidBucketName: The specified bucket is not valid.
	status code: 400, request id: 5TMYE4RWPMT8S15B, host id: S8Lz940nmuue0jhfj4QopWDauaNW3rqwr2FNBUPyf0qvGVSG7MJre0lAcMh3ovEwl8mnhmS0zIw=
Error: go.estuary.dev/E002: Failed to execute command: airbyte-to-flow: failed with code 1.
```

dyaffe:
I just tried this with `us-east-2` as my region and `stupid-fake-not-working-bucket-234` as my bucket and it let me save.  It appears that the connector just isn't validating my bucket on discover and we should add that.

PR #494, opened 2023-01-06
source-{gcs,s3}: validate bucket is connectable when creating store

williamhbaker:
**Description:**

Adds a check to `source-s3` & `source-gcs` to make sure the bucket is connectable when creating the "store object". This happens during the airbyte protocol `Check` command, so an error connecting to the bucket will be surfaced during validation, rather than at runtime. This is particularly important now that we aren't doing any kind of discovery with the cloud storage connectors, since at no point during the setup does it actually connect to the bucket.

Errors that will be returned during validation will look like this for a non-existent bucket:

- `source-s3`: `connecting to store: validating connection to bucket "not-a-real-bucket-as8djf98": NotFound: Not Found`
- `source-gcs`: `connecting to store: validating connection to bucket "not-a-real-bucket-as8djf98": storage: bucket doesn't exist`

Closes https://github.com/estuary/connectors/issues/493

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/494)

### DIFF
--- a/source-gcs/main.go
+++ b/source-gcs/main.go
@@ -69,6 +69,10 @@ func newGCStore(ctx context.Context, cfg *config) (*gcStore, error) {
 		return nil, fmt.Errorf("creating GCS session: %w", err)
 	}
 
+	if _, err := client.Bucket(cfg.Bucket).Attrs(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM!

Issue #497, opened 2023-01-11
Materialize Pubsub: Create subscription for topic when creating topic


williamhbaker:
Messages delivered to a pubsub topic with no subscription are lost. When we create a pubsub materialization, the common case is for the materialization to create a topic as well and immediately start publishing documents to it. This results in the documents being lost until a subscription is created.

When using the pubsub UI to create a topic, there is an option to create a "default subscription" that ends up being like `your-topic-name-sub`. We should allow the pubsub connector to also create a default subscription in this same way.

This seems to be new behavior since I didn't notice this happening when the connector was originally rolled out, so perhaps this has changed. But we should be able to adapt pretty easily by allowing for creating a default subscription.

PR #498, opened 2023-01-12
materialize-google-pubsub: improve memory & create default subscription

williamhbaker:
**Description:**

Two improvements to pubsub in this PR:

- The first commit provides a config option to create a default subscription when `ApplyUpsert` creates a topic. This is almost always going to be what should happen, otherwise the initially published messages will be dropped since there is no subscription. It is by default `true`. This closes https://github.com/estuary/connectors/issues/497
- The second commit sets the underlying PubSub client to block if its message queue is full. The default behavior was to never block, and this meant that if the store iterator could be read from faster than publish messages could be completed (network bound), unbounded in-memory buffering could occur. The PubSub client handles concurrently sending requests already, with a batch size of 100 and a queue limit of 1000, so this shouldn't have any real impact on connector performance.

**Workflow steps:**

Use the connector like before, possibly configuring the default subscription option.

PubSub connector docs will need updated with the new config option.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/498)

### DIFF
--- a/materialize-google-pubsub/.snapshots/TestSpec
+++ b/materialize-google-pubsub/.snapshots/TestSpec
@@ -99,7 +99,7 @@
       "create_default_subscription": {
         "type": "boolean",
         "title": "Create with Default Subscription",
-        "description": "Create a default subscription when creating the topic. Will be created as \"topic-sub\". Has no effect if the topic already exists.",
+        "description": "Create a default subscription when creating the topic. Will be created as \"\u003ctopic\u003e-sub\". Has no effect if the topic already exists.",
         "default": true
       }
     },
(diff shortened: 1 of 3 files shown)


materialize-google-pubsub/.snapshots/TestSpec
```diff
@@ -95,11 +95,18 @@
         "type": "string",
         "title": "Resource Binding Identifier",
         "description": "Optional identifier for the resource binding. Each binding must have a unique topic \u0026 identifier pair. Included as \"identifier\" attribute in published messages if specified."
+      },
+      "create_default_subscription": {
+        "type": "boolean",
+        "title": "Create with Default Subscription",
+        "description": "Create a default subscription when creating the topic. Will be created as \"topic-sub\". Has no effect if the topic already exists.",
```
mdibaiee, line 102: nit: maybe consider somehow making it clear that `topic` is a variable here, e.g. `<topic>-sub`?

mdibaiee (approved):
LGTM, just a small nit above

materialize-google-pubsub/.snapshots/TestSpec
williamhbaker: That seems like a good idea. I was waffling on the description a bit here myself so the clarification should be helpful.

PR #501, opened 2023-01-13
source-{s3,gcs}: verify bucket existence & permissions with minimal required permissions

williamhbaker:
…equired permissions

**Description:**

This is a follow-up to #494 which added validation checks for cloud storage connectors to check for the existence of a bucket.

I originally had those checks get metadata for the bucket to confirm it exists, but that requires a separate permission from what is otherwise needed. These connectors only need permissions to list and read objects in a bucket and don't necessarily need permission to read metadata about buckets.

The change here is a bit more complicated than before, but produces similar validation results. It only requires list & get objects permissions.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/501)

### DIFF
--- a/source-s3/main.go
+++ b/source-s3/main.go
@@ -5,14 +5,19 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
+	"path"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/estuary/connectors/filesource"
 	"github.com/estuary/flow/go/parser"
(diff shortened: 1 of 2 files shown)


source-s3/main.go
```diff
@@ -87,7 +92,42 @@ func newS3Store(ctx context.Context, cfg *config) (*s3Store, error) {
...
+	})
+	if err != nil {
+		// Note that a listing with zero objects does not return an error here, so we don't have to
+		// account for that case.
+		return fmt.Errorf("unable to list objects in bucket %q: %w", cfg.Bucket, err)
+	}
+
+	// s3:GetObject allows reading object data as well as object metadata. The name of the object is
+	// not important here. We have verified our ability to list objects in this bucket, so we can
+	// interpret a "not found" as a successful outcome.
```
willdonnelly, line 122: It might be worth mentioning in the comment that we don't have to handle the 403 status possibility here because we already verified that we have `s3:ListObjects` up above.

willdonnelly (approved):
LGTM

Issue #505, opened 2023-01-17
materialize-sql: handle numbers formatted as strings


williamhbaker:
It is not uncommon for a captured collection to have a field that is normally a number, but is sometimes encoded as a string. We would like our sql materializations to be able to handle these cases

PR #507, opened 2023-01-18
materialize-sql: allow mapping multiple types if strings are formatted as integers or numbers

williamhbaker:
**Description:**

Adds support for the SQL materialization connectors to handle integers and numbers encoded as strings. This can be enabled by including `format: integer` or `format: number` in the collection schema. The fields will also need to be explicitly included in the materialization since they are optional rather than recommended.

Needs https://github.com/estuary/flow/pull/883 to be merged & deployed before merging this. The broken tests here also require that PR to be merged so that the new formats are allowed.

Closes https://github.com/estuary/connectors/issues/505

**Workflow steps:**

SQL materializations can now handle schema with fields like these:

```yaml
  int_field:
    type:
      - integer
      - string
    format: integer
  num_field:
    type:
      - number
      - string
    format: number
```

They will be materialized into numeric columns, with the string values being converted to the appropriate numeric value (integer or decimal).

You can update the collection schema for a field from this:

```yaml
  int_field:
    type:
      - integer
```

...to this, while the materialization is active:

```yaml
  int_field:
    type:
      - integer
      - string
    format: integer
```

The purpose of allowing this kind of change is to account for scenarios where the data source (Ex: HubSpot) mostly produces numeric values, but sometimes has them encoded as strings. The initial schema inference may only see numeric values, and a string-encoded number might only surface later.

We should update the [schema](https://docs.estuary.dev/concepts/schemas/) docs with these new format annotations. We'll want to point out that for SQL materializations (postgres, snowflake, bigquery, timescale), a `string` type field may be added to a collection field with an existing materialization as long as the field has a single existing type, it is `integer`, or `number`, and a `format: string` / `format: number`  is added with the string. This can be done without re-creating the materialization. And it cannot be done with collection keys, only non-keyed values.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/507)


materialize-sql/type_mapping.go
```diff
@@ -169,6 +231,47 @@ func WithElementConverter(converter ElementConverter) StaticMapperOption {
 	}
 }
 
+// StringCastConverter builds an ElementConverter from the string-handling callback. Any non-string
+// types are returned without modification. This should be used for converting fields that have a
+// string type and one additional type. The callback should convert the string into the desired type
+// per the endpoint's requirements.
+func StringCastConverter(fn func(string) (interface{}, error)) ElementConverter {
```
williamhbaker, line 238: Technically everything seems to work fine with all our current SQL materializations if we don't do this, and just let the database handle an `interface{}` with a concrete type of `string` as its input for a numeric column. Insert queries into postgres (for example) when doing this would sometimes use `1` and sometimes use `'1'` as their parameter value into a `BIGINT` column. This does apparently work but feels really wrong, and may cause bugs I haven't tested for in our current SQL materializations, or perhaps future ones. So I am making it require a little more work to get the correct type for the queries.

williamhbaker:
I made some updates to the original PR; most notably:
- Allowing a `type: string` + `format: integer` (or `format: number`) to be added to a collection with an _existing_ materialization without needing to re-create the materialization
- Disallowing the format annotation to apply to collection keys

materialize-sql/type_mapping.go
```diff
@@ -79,34 +85,90 @@ func BuildProjections(spec *pf.MaterializationSpec_Binding) (keys, values []Proj
...
+		return "", false
+	}
+
+	for _, f := range fields {
+		if f != pf.JsonTypeString {
+			return f, true
+		}
+	}
+
+	// Should never get here - would require both types to be "string".
```
jgraettinger, line 170: nit: panic in this case?

materialize-sql/type_mapping.go
```diff
@@ -79,34 +85,90 @@ func BuildProjections(spec *pf.MaterializationSpec_Binding) (keys, values []Proj
...
+	if projection.IsPrimaryKey || projection.Inference.String_ == nil {
+		return projection.Inference.Types
+	}
+
+	format := projection.Inference.String_.Format
+	if !SliceContains(format, effectiveTypeFormats) {
+		return projection.Inference.Types
+	}
+
+	applicableType, ok := singleNonStringField(projection.Inference.Types)
```
jgraettinger, line 143: the combination of logic here and in `singleNonStringField`, plus the separate separate `effectiveTypeFormats` const, is hard to follow as a reader and is a bit smelly. It feels like too much and unnecessary indirection.

Also, this appears to allow for some unintended cases that probably _shouldn't_ be mapped:
 * `{type: [string, number], format: integer}`
 * `{type: [string, integer], format: number}`

What about having this routine be very explicit, such as:

```
if !projection.IsPrimaryKey && projection.Inference.String_ != nil {
  if reflect.DeepEquals(projection.Inference.Types, []string{"integer", "string") && format == "integer" {
    return []string{"integer"}
  }
  if reflect.DeepEquals(projection.Inference.Types, []string{"number", "string") && format == "number" {
    return []string{"number"}
  }
}
```
?

Note that types are [always in fixed, sorted order](https://github.com/estuary/flow/blob/master/crates/json/src/schema/types.rs#L62-L71) so you can rely on that for slice comparison if you like.

To be really pedantic, we should probably also handle "fractional", which is a type we introduced that is a floating-point and _not_ an integer. It actually can come up in inference, through constructions like: 

```
type: number
not: {type: integer}
```

We do allow it to be specified directly as well, as a JSON schema extension.

materialize-sql/validate.go
```diff
@@ -130,11 +130,8 @@ func checkTypeError(field string, existing *pf.CollectionSpec, proposed *pf.Coll
 		return "The proposed materialization is missing the projection, which is required because it's included in the existing materialization"
 	}
 
-	// Ensure that the possible types of the proposed are a subset of the existing possible types.
-	// The new projection is allowed to contain fewer types than the original, though, since that
-	// will always work with the original database schema.
-	for _, pt := range proposedProjection.Inference.Types {
-		if !SliceContains(pt, existingProjection.Inference.Types) {
+	for _, pt := range effectiveJsonTypes(proposedProjection) {
```
jgraettinger, line 133: nit: dropped comment seems useful?

jgraettinger (commented):
Looking good, a couple of comments below. Thanks!

materialize-sql/validate.go
williamhbaker: Yep I think I killed it on accident...brought it back with some modification.

materialize-sql/type_mapping.go
williamhbaker: Seems reasonable to simplify. I had originally done it this way with the thought of making it really easy to extend to other formats. Being more explicit does simplify things a lot, and it still wouldn't be hard to handle more formats if we ever actually need to, so I think this is a good idea. I made the change like you suggested & everything continues to work well.

I'll work on getting the `fractional` type to work when specified directly.

williamhbaker:
Added support for the `fractional` type here & the corresponding string format. The corresponding flow PR is https://github.com/estuary/flow/pull/889/

williamhbaker:
After some internal discussion, I've removed support for strings formatted as `fractional`, since it's a type that should probably not be used by users and it has some confusing nuances.

I am keeping the bit that adds support for _materializing_ a field with type `fractional` though, since that is trivial & non-user facing, and may come in handy for us internally.

mdibaiee (approved):
LGTM, thank you Will!

PR #509, opened 2023-01-18
source-http-file: skip schema inference

mdibaiee:
**Description:**

- Remove discovery logic from `filesource` entirely

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/509)


williamhbaker:
In https://github.com/estuary/connectors/pull/415 I added the option to skip schema inference to `filesource`. HTTP file was the only one still using it, so now we can get rid of that optional path entirely. I had considered it to be a temporary measure at the time, until we got to this point. We should clean up `filesource/filesource.go` with this as well by making `(src Source) Discover` only have the path of skipping inference and getting rid of the bool option entirely since everything should be doing that now.

PR #510, opened 2023-01-18
source-kinesis: remove local schema inference, use x-infer-schema: true

mdibaiee:
**Description:**

- source-kinesis: remove local schema inference, use `x-infer-schema: true` instead
- Added `sequence_number` and `partition_key` to form a key for each record received from Kinesis, this is based on the docs: "SequenceNumber is the unique identifier of the record within its shard." and "PartitionKey identifies which shard in the stream the data record is assigned to.", which I take to mean a combination of two forms a unique identifier of each record.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/510)

### DIFF
--- a/source-kinesis/capture.go
+++ b/source-kinesis/capture.go
@@ -425,7 +425,21 @@ func (r *shardReader) extractRecords(resp *kinesis.GetRecordsOutput) []json.RawM
 	} else {
 		var result = make([]json.RawMessage, len(resp.Records))
 		for i, rec := range resp.Records {
-			result[i] = json.RawMessage(rec.Data)
+			var j = make(map[string]interface{})
+			if err := json.Unmarshal(rec.Data, &j); err != nil {
+				panic(err)
+			}
+
+			j[metaProperty] = make(map[string]string)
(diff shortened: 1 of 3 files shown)


williamhbaker:
Does have something like stream name + message offset that could be used for the key? I think we'd need a key since I don't believe schema inference figures out keys.

mdibaiee:
Tested this using some dummy data with AWS Kinesis Data Generator and it seemed to be working fine.

williamhbaker (approved):
LGTM - the only thing giving me a slight pause is including the stream name in the key and the data record itself that I had mentioned earlier. Ultimately I think that would be redundant though, since each stream will get its own binding and collection. The combination of sequence number and partition key don't sound like they would be unique _across_ streams, but that's the same as how a primary key in a database table is only unique within that table and is certainly not unique across tables.

mdibaiee:
@williamhbaker I had the same thought regarding stream name, since we are already separating records by stream into separate collections, the keys between streams shouldn't collide, so I don't think we need to include the stream name in the document itself

PR #512, opened 2023-01-18
source-mysql: Replace the all-zeroes `DATETIME` with a valid RFC3339 placeholder

willdonnelly:
**Description:**

The `DATETIME` value `"0000-00-00 00:00:00"` is used by MySQL to indicate invalid data. However we can't process it into a RFC3339 timestamp through the normal parse-then-format machinery because month/day 0 doesn't exist. So we hard-code the replacement `"0000-00-00 00:00:00" -> "0001-01-01T00:00:00Z"` instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/512)


jgraettinger (approved):
LGTM

PR #515, opened 2023-01-20
sqlcapture: Fail cleanly when a capture binding doesn't exist

willdonnelly:
**Description:**

Previously there was a bug where `c.discovery[streamID].PrimaryKey` was used blindly without first verifying that the table actually was found by discovery, leading to a panic. This PR adds a test of the "table listed in bindings doesn't actually exist" scenario and then adds a check so we produce a clean error message instead.

And for good measure I've also fixed a similar bug where `c.discovery[watermarksTable].PrimaryKey` was being dereferenced blindly in the same way.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/515)


williamhbaker (approved):
lgtm

willdonnelly:
So I realized while re-reading this that the new logic kind of implicitly requires that discovery always succeed, and in the past we've tried to allow very restrictive permission setups where for instance a Postgres capture can't even access `pg_catalog` / `information_schema`, and that would work so long as the bindings were already set up.

However we can't make this work on MySQL since we need discovery results to initialize table metadata, and it's quite useful to be able to rely on the discovery results even during normal connector operation (as evidenced by how the code has grown over time to do just that). After bringing this up in standup the consensus was that we should just always require discovery capabilities, so this PR should be fine to merge as-is and we'll need to remove from the Postgres docs the statement:

> The ‘information_schema’ and ‘pg_catalog’ access is required for stream auto-discovery, but not for capturing already configured streams.

And maybe instead document the specific information_schema and pg_catalog tables we require access to (since it's just a few of them that we use).

Issue #506, opened 2023-01-17
Rockset Connector Wording Update


dyaffe:
**Problem**
Our Rockset connector has an option for "initialize from S 3" which is confusing to Rockset.  Their feature is called "Bulk ingest via S3".

**Recommendation**
Can we either just rename it to "Optional bulk ingest via S3" or remove it?

PR #516, opened 2023-01-20
materialize-rockset: remove s3 integration option

williamhbaker:
**Description:**

This removes the s3 integration option from the Rockset materialization. It was not functional in its current state, difficult to use, and likely to result in corrupt Rockset collections since Rockset's S3 integration does not guarantee ordering.

There was also some refactoring & simplification possible since the newer Rockset client we are now using provides a built-in method for waiting for a Rockset collection to be ready to accept writes, so that bit of code was removed from our connector.

I tested this locally with some test data sets and everything still worked like before.

Closes https://github.com/estuary/connectors/issues/506

**Workflow steps:**

Use the connector like before, but now you won't see the S3 integration configuration.

Update Rockset docs to remove the remaining "initializeFromS3" configuration options. Most of the S3 integration stuff has already been removed from Rockset.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/516)

### DIFF
--- a/materialize-rockset/.snapshots/TestRocksetDriverSpec-TestResourceSpecSchema
+++ b/materialize-rockset/.snapshots/TestRocksetDriverSpec-TestResourceSpecSchema
@@ -44,43 +44,6 @@
       "type": "object",
       "title": "Advanced Collection Settings",
       "advanced": true
-    },
-    "initializeFromS3": {
-      "properties": {
-        "integration": {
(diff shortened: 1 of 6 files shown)


psFried (approved):
Love it! 👍🚀 Feels good to see all those deletions

mdibaiee (approved):

PR #517, opened 2023-01-20
source-mysql: decompose range query to make it easier to optimize

jgraettinger:
Instead of using "(key_one, key_two) > (123, 'bar')" constructions, use "key_one > 123 OR (key_one = 123 AND key_two > 'bar')". MySQL and MariaDB's optimizer is able to make better use of this form.

Also:

* Lower a spammy log to trace level.
* Attempt to use ANALYZE rather than EXPLAIN for the backfill query, as its output is more accurate.

**Description:**

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/517)

### DIFF
--- a/source-mysql/backfill.go
+++ b/source-mysql/backfill.go
@@ -54,13 +54,22 @@ func (db *mysqlDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.Di
 	}
 
 	// Build a query to fetch the next `backfillChunkSize` rows from the database.
-	// If this is the first chunk being backfilled, run an `EXPLAIN` on it and log the results
 	var query = db.buildScanQuery(resumeKey == nil, keyColumns, columnTypes, schema, table)
-	db.explainQuery(streamID, query, resumeKey)
+	// Splat the resume key into runs of arguments matching the predicate built by buildScanQuery.
+	// This is super gross, and is a work-around for lack of indexed positional argument support,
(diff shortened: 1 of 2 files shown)


source-mysql/backfill.go
```diff
@@ -153,15 +170,17 @@ func (db *mysqlDatabase) explainQuery(streamID, query string, args []interface{}
...
-	logrus.WithFields(logrus.Fields{
-		"id":    streamID,
-		"query": explainQuery,
-	}).Info("explain backfill query")
-	explainResult, err := db.conn.Execute(explainQuery, args...)
+	// Ask the database to analyze the backfill query.
+	var explainResult, err = db.conn.Execute("ANALYZE "+query, args...)
 	if err != nil {
-		logrus.WithField("query", explainQuery).Warn("unable to execute query")
+		// ANALYZE is not universally supported. Try again with EXPLAIN.
```
willdonnelly, line 176: I believe the bare `ANALYZE` is a MariaDB-ism and the MySQL equivalent is `EXPLAIN ANALYZE`, so it might be worth trying both of those before resorting to a bare `EXPLAIN`.

willdonnelly (approved):
LGTM

source-mysql/backfill.go
jgraettinger: done

PR #519, opened 2023-01-20
Rename all connector binaries in their Dockerfiles

willdonnelly:
**Description:**

All we get from our current OOM-kill monitoring is the name of the process that got killed, so having them all be named "connector" is *rather unhelpful*. This should make OOM-kill events marginally easier to track down.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/519)


willdonnelly:
I'm a little leery of merging this at 5PM on a Friday, so I'll probably hold off on this until Monday even though it would be useful right now to help figure out what's OOMing.

williamhbaker (approved):

PR #522, opened 2023-01-23
materialize-sql: allow updating fields to make them nullable

mdibaiee:
**Description:**

- Allow updating a collection to make a field nullable in `materialize-sql` connectors

**Workflow steps:**

- Create a materialization with non-nullable fields
- Update those fields to make them nullable
- The connector should update your destination database to mark those columns as nullable

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/522)

### DIFF
--- a/materialize-sql/driver.go
+++ b/materialize-sql/driver.go
@@ -116,6 +116,8 @@ func (d *Driver) ApplyUpsert(ctx context.Context, req *pm.ApplyRequest) (*pm.App
 		tableShapes = append(tableShapes, *endpoint.MetaCheckpoints)
 	}
 
+	var statements []string
+
 	for bindingIndex, bindingSpec := range req.Materialization.Bindings {
 		var collection = bindingSpec.Collection.Collection
 		var constraints, loadedBinding, resource, err = resolveResourceToExistingBinding(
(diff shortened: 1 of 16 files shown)


materialize-sql/driver.go
```diff
@@ -141,15 +146,34 @@ func (d *Driver) ApplyUpsert(ctx context.Context, req *pm.ApplyRequest) (*pm.App
 				loadedBinding.FieldSelection.String(),
 			)
 		} else if loadedBinding != nil {
-			continue // Table was already created. Nothing to do.
+			// If the table has already been created, make sure all non-null fields
+			// are marked as such
+			for _, field := range bindingSpec.FieldSelection.Values {
+				var p = bindingSpec.Collection.GetProjection(field)
+
+				if p.Inference.Exists != pf.Inference_MUST {
```
williamhbaker, line 154: If I'm reading this correctly, any time a proposed spec contains fields that are not required the alter table statement will be executed. Would it be possible to compare the proposed spec vs. the loaded spec and only run the table alteration if there's a difference?

williamhbaker:
I was also thinking about going in the other direction: Going from not required -> required. This gets interesting because there may very well be rows in the table that are `null` so changing the column to `not null` wouldn't work. For now I think it's ok & good to not worry about that case. Maybe in the future we could support this case using a default value from the schema, and inserting that value into the column if the row value is `null` before doing the alter table.

materialize-sql/driver.go
mdibaiee: done ✅

williamhbaker (approved):

PR #523, opened 2023-01-24
sqlcapture: validate existance and accessibility of table

mdibaiee:
**Description:**

- At `Validate` and `ApplyUpsert` RPC, make sure we have access to the binding tables
- I'm reusing our Discover logic for finding accessible tables here to be consistent. The code path that I have is probably not the most efficient if user is not choosing all the discovered bindings, but it is _simple_ and means we don't need some new logic.

**Workflow steps:**

- Create a capture to a sql endpoint (e.g. `source-postgres`) and add a binding to a nonexistant (or non-accessible) table. You should see an error like this:
```
    could not find or access table estuary_test_cfngjj
```

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/523)

### DIFF
--- a/sqlcapture/main.go
+++ b/sqlcapture/main.go
@@ -63,6 +63,29 @@ func (d *Driver) Spec(ctx context.Context, req *pc.SpecRequest) (*pc.SpecRespons
 
 // ApplyUpsert applies a new or updated capture to the store.
 func (d *Driver) ApplyUpsert(ctx context.Context, req *pc.ApplyRequest) (*pc.ApplyResponse, error) {
+	var db, err = d.Connect(ctx, string(req.Capture.Capture), req.Capture.EndpointSpecJson)
+	if err != nil {
+		return nil, fmt.Errorf("error connecting to database: %w", err)
+	}
+	defer db.Close(ctx)
+
+	discoveredTables, err := db.DiscoverTables(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, binding := range req.Capture.Bindings {
+		var res Resource
+		if err := pf.UnmarshalStrict(binding.ResourceSpecJson, &res); err != nil {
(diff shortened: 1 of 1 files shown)


sqlcapture/main.go
```diff
@@ -79,16 +79,35 @@ func (d *Driver) Validate(ctx context.Context, req *pc.ValidateRequest) (*pc.Val
...
 	var out []*pc.ValidateResponse_Binding
 	for _, binding := range req.Bindings {
 		var res Resource
 		if err := pf.UnmarshalStrict(binding.ResourceSpecJson, &res); err != nil {
 			return nil, fmt.Errorf("error parsing resource config: %w", err)
 		}
+
+		var streamID = JoinStreamID(res.Namespace, res.Stream)
+
+		if !SliceContains(streamID, discoveredStreams) {
```
willdonnelly, line 141: Unless I've missed some subtlety the whole `discoveredStreams` / `SliceContains` thing could be removed. The `discoveredBindings` map is already keyed by stream ID, so `if _, ok := discoveredBindings[streamID]; ok {` should be equivalent here.

mdibaiee: @willdonnelly I checked the return value of `DiscoverCatalog` and it's `[]*pc.DiscoverResponse_Binding` so it's not a map, but an array

willdonnelly: Ah, my mistake. I was thinking of the underlying `DiscoverTables` operation that each database's interface code implements, which returns the map I was thinking of (and then `DiscoverCatalog` implements the whole "translate that into a slice of `pc.DiscoverResponse_Binding` objects with appropriate JSON schemas" step).

mdibaiee: Thanks, I switched to using that one now 👍🏽

Issue #526, opened 2023-01-30
materialize-google-sheets: connector OOMs if the view is too large


williamhbaker:
The google sheets materialization requires the entire materialized view to fit in memory. If you try to materialize a large collection the connector may run out of memory and hit the memory usage limit, resulting in an OOM kill.

It would probably be better to have some kind of internal kill switch in the connector itself, limiting the materialized view to a reasonable number of rows so that it will not consume all available memory if the view is too large. This would allow for better error messaging to the user.

PR #527, opened 2023-01-31
materialize-google-sheets: limits for maximum number of materialized cells

williamhbaker:
**Description:**

Adds a per-binding limit of 1,000,000 cells that can be part of the materialized view. This will prevent the connector from using an excessive amount of memory when reading from a high cardinality collection. We can fail fast in this case rather than letting the connector run out of memory and be OOM-killed.

Google sheets can actually hold more cells than this (up to 10,000,000 in a workbook). But based on some simple tests with wikipedia recent changes data, 1,000,000 cells of that collection uses about 200mb of data, so this seems like a more reasonable limit.

With this change it is still possible for a materialization to be specified with many bindings that would each stay under the limit, but cumulatively could use more memory. Rather than handling that complicated case, this PR is addressing the more common case of a user creating a materialization with a single binding and not realizing that it will use an unbounded amount of memory.

Closes https://github.com/estuary/connectors/issues/526

**Workflow steps:**

Materializing a collection into google sheets with a large number of keys that would exceed 1,000,000 cells will result in the connector failing and an error message like this:
```
Maximum cells limit of 1000000 exceeded for this materialization. If you are materializing a collection with a large number of unique keys, consider creating a derivation to transform your data into a collection with fewer unique keys.
```

I don't believe we have anything describing memory limits in the [Google sheets materialization docs](https://docs.estuary.dev/reference/Connectors/materialization-connectors/Google-sheets/), so it would be good to add something about that. Along the lines of: A single binding cannot materialize more than 1 million cells; if your collection has an unbounded amount of unique keys you should make a derivation to summarize them into a bounded set of keys.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/527)

### DIFF
--- a/materialize-google-sheets/transactor.go
+++ b/materialize-google-sheets/transactor.go
@@ -13,6 +13,25 @@ import (
 	"google.golang.org/api/sheets/v4"
 )
 
+const (
+	// Maximum number of cells allowed for any single binding to either read from the Store iterator
+	// or hold in-memory in the binding's rows.
+	cellsLimit = 1000000
+)
+
+func checkCellCount(rows int, cellsPerRow int) error {
+	if rows*cellsPerRow > cellsLimit {
+		return fmt.Errorf(
+			"Maximum cells limit of %d exceeded for this materialization. If you are materializing "+
+				"a collection with a large number of unique keys, consider creating a derivation to "+
+				"transform your data into a collection with fewer unique keys.",
+			cellsLimit,
+		)
+	}
+
+	return nil
+}
+
 type transactor struct {
 	bindings []transactorBinding
 	client   *sheets.Service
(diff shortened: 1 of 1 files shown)


jshearer:
> Maximum cells limit of 1000000 exceeded for this materialization. If you are materializing a collection with a large number of unique keys, consider creating a derivation to transform your data into a collection with fewer unique keys.

If we're talking about adding docs around this, I think it would be great to show what a sample derivation would look like, if for no other reason than to clarify what "a collection with fewer unique keys" might mean.

williamhbaker:
> If we're talking about adding docs around this, I think it would be great to show what a sample derivation would look like, if for no other reason than to clarify what "a collection with fewer unique keys" might mean.

Yeah it's tricky to describe the concept succinctly in an error message. We've got this article that might be useful to link from the docs: https://www.estuary.dev/how-to-create-a-real-time-materialized-view-in-postgresql/

mdibaiee (approved):

PR #536, opened 2023-02-10
source-postgres: Fix handling of capitalized table names

willdonnelly:
**Description:**

This fixes a couple of bugs when interacting with tables whose names explicity capitalized, for example the one you get by running `CREATE TABLE "USERS" (id INTEGER PRIMARY KEY, name TEXT);`.

It also adds a new test 'TestCapitalizedTables' which exercises discovery, validation, and capturing from a table named `"USERS"` (with quotes, meaning that it will actually be capitalized on the PostgreSQL side).

Also adds a bonus commit which I _think_ fixes CI build flakiness around the final `DROP_REPLICATION_SLOT` performed during tests. (Added in this PR because its CI builds were consistently failing, and either adding that change fixed it or it's just random and I got lucky)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/536)

### DIFF
--- a/source-boilerplate/testing/testing.go
+++ b/source-boilerplate/testing/testing.go
@@ -271,7 +271,7 @@ func (a *pullAdapter) Send(m *pc.PullResponse) error {
 		for _, documents := range a.transaction {
 			var binding string
 			if idx := documents.Binding; 0 <= idx && int(idx) < len(a.bindings) {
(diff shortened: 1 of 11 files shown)


williamhbaker (approved):
LGTM

PR #539, opened 2023-02-10
source-postgres: Bare minimum of user type support

willdonnelly:
**Description:**

This PR represents the bare minimum we can do to make the use of user-defined types not fail a capture. Concretely, this changes `Type` messages from logical replication to no longer be an error, and adds some tests verifying correct-ish behavior on domain, enum, and tuple types.

There are still a lot of rough edges:

  1. Discovery only "works" to the extent that it doesn't crash. The discovered JSON type for a column of user- defined type is `{description: "<error>"}`, meaning that it accepts anything but complains about it if you look at the schema.
  2. Tuple types aren't serialized especially well as JSON. Rather than becoming a JSON array or object, they are just emitted as a JSON string whose contents is the Postgres text-form representation of the value.
  3. There are two other sorts of user-defined type in Postgres: 'ranges' and 'scalars'. These are not currently tested at all, but in the case of ranges it appears to behave pretty much like a tuple.

None, there's some additional rough spots I'd like to work on before we actually _tell_ people we support user types.

**Notes for reviewers:**

Uploaded as-is just in case we need to get a capture unblocked ASAP. More work will occur next week to try and get user types to a state we're actually happy with.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/539)

### DIFF
--- a/source-postgres/.snapshots/TestUserTypes-Domain-Capture
+++ b/source-postgres/.snapshots/TestUserTypes-Domain-Capture
@@ -1,0 +1,10 @@
+# ================================
+# Collection "test.UserTypes_Domain_bandoleer": 2 Documents
+# ================================
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"usertypes_domain_bandoleer","loc":[11111111,11111111,11111111]}},"id":1,"value":"hello"}
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"usertypes_domain_bandoleer","loc":[11111111,11111111,11111111]}},"id":2,"value":"world"}
+# ================================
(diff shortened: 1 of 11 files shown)


jgraettinger (approved):
LGTM

This is ... not a lot of change LOL. I think this can merge today as a clear improvement for domain and enum types, which are more common.

If it's easy to continue to fail (for now) on tuple types, perhaps that would be ideal, but if not I think we can merge this as-is.

willdonnelly:
I'd say let's go ahead and merge this, then. I intend to dig into tuple types later this afternoon or tomorrow, so if it turns out to be easy enough we can re-disable those (or just make them work properly 🤞) as a fast-follow

PR #540, opened 2023-02-13
source-firestore/backfill: translate document to json-serializable map

mdibaiee:
**Description:**

- resolves https://github.com/estuary/flow/issues/928
- There is no "standard" way of handling NaN values, I think marking them as `null` is the best way to go about it (the alternative being a zero value, but I think that's more problematic)

**Workflow steps:**

**Notes for reviewers:**


mdibaiee:
@willdonnelly I noticed in this PR https://github.com/estuary/connectors/pull/379 we are translating `NaN` into the string `NaN`, that would lead to the type of the colum becoming `[number, string]` which means most materializations will likely not pick it up for materialization.

Do we want to continue with the previous approach and translate to `NaN`, or null?

willdonnelly:
I'm not really sure. I think that `"NaN"` is a more accurate representation of the source data than `null` but the ergonomics around materialization are definitely a bit worse today.

The more fundamental issue, to my mind, is that there is already logic that was supposed to result in us not trying to serialize a NaN value, and yet we saw that happen the other day. Do we know why that was? I'm betting it has to do with replication events versus backfill data paths, but I haven't checked the code to be sure. Regardless of whether we decide to go with `"NaN"` or `null` we ought to fix that so it happens in a single place.

mdibaiee:
> Do we know why that was? I'm betting it has to do with replication events versus backfill data paths, but I haven't checked the code to be sure.

Yes, we use different code paths and libraries when backfilling and when receiving events. When backfilling we use the firestore library which does not give us a typed document, but just a `map[string]interface{}`, hence the new code.

---

I went ahead and changed the code to use `"NaN"` instead of `null`. I will create another issue for us to track down how we can use this in materializations to use the best `NaN` representation for each materialization.

willdonnelly (approved):
LGTM

PR #541, opened 2023-02-13
source-sqlserver: Support captures as non-sysadmin user

willdonnelly:
**Description:**

This involved a multitude of little tweaks, including:

  - Editing the database initialization script to create a new 'flow_capture' user with the necessary permissions.
  - Modifying the test suite setup code to use a different user for test manipulations and captures, so that we can create tables and enable CDC as the sysadmin and then use the restricted 'flow_capture' user for captures.
  - Modifying the test suite to enable CDC for newly created tables, so that the 'flow_capture' user doesn't require that permission.
  - Modifying the connector "ensure watermarks table exists" process to use a simple `SELECT * FROM watermarks` to tell if the table exists, because the complex "check if it exists and if not create it" query requires create permissions even when the table already exists.
  - A few other tweaks that came to my attention while I was debugging permissions issues.

After this, a restricted user (with `SELECT` permissions on the `dbo` and `cdc` schemas, `UPDATE` on `dbo.flow_watermarks`, and nothing else) is able to pass the test suite.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/541)

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -508,7 +508,7 @@ func (rs *mysqlReplicationStream) keyColumns(streamID string) ([]string, bool) {
 	return keyColumns, ok
 }
 
-func (rs *mysqlReplicationStream) ActivateTable(streamID string, keyColumns []string, discovery *sqlcapture.DiscoveryInfo, metadataJSON json.RawMessage) error {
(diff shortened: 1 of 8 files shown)


willdonnelly:
Going to go ahead and merge this so that the connector can actually be useful in production. I don't think any of these changes are particularly controversial anyway.

PR #548, opened 2023-02-16
Capture MongoDB

mdibaiee:
**Description:**

- MongoDB offers no information on last update time of documents. The approach taken here is doing a backfill of all documents the first time the connector is run, and then running a change stream to stream updates since the last time the connector ran.
- For real-time updates I use [change streams](https://www.mongodb.com/docs/manual/changeStreams/) which are only supported if you are running a replicaset

**Workflow steps:**

- Run a mongodb database locally, e.g.:
```
openssl rand -base64 756 > ~/sandbox/keyfile
chmod 400 ~/sandbox/keyfile

docker run --name mongo --network=flow-test --mount type=bind,source=/Users/mahdi/sandbox/keyfile,target=/keyfile -e MONGO_INITDB_ROOT_USERNAME=flow -e MONGO_INITDB_ROOT_PASSWORD=flow --publish 27017:27017 mongo:latest --replSet "rs0" --keyFile /keyfile
```
- Connect to database to initiate replicaset by exec-ing into the docker container:
```
docker exec -it mongo bash
mongosh -u flow -p flow
rs.initiate()
```
- Then connect to database from your local machine using `mongosh` and initialize some data:
```
mongosh -u flow -p flow
db.createUser({ user: 'flow', pwd: 'flow', roles: [{ role: 'read', db: 'test' }] })
db.users.insert({ name: "mahdi" })
db.users.insert({ name: "johnny", "last_name": "graettinger" })
db.users.insert({ name: "tyler", occupation: "soap maker" })
```
- Then run discover and deploy a capture on this connector using `flowctl-go discover` and `flowctl-go deploy`
- Try adding documents after backfill has been done to watch real-time updates

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/548)


willdonnelly:
This looks like a solidly written implementation of what it does, but I have some higher level concerns. If I'm understanding correctly:

- The connector will run a new backfill of all captured collections each time it starts up
- By default we'll only get streaming changes which occur after the change stream was opened in a given connector run.
- Change streaming starts after the backfill, events could potentially be dropped in that gap, and our recourse is just that we'll probably get restarted at some unknown point in the future and redo the backfill

Experience has shown that these JSON document stores can get pretty large, and pulling down the whole dataset over and over is a pain both for us and for the user (also it can cost money -- the MongoDB "serverless" offering costs $0.10 per million reads, and that can add up to real money surprisingly fast).

The lack of timestamps is a bit of a drag, but I think this could still be done better:
1. We should start change streaming before performing a backfill. That way we don't risk missing any change events, and we must already be fine with duplicated documents because we'll be producing a bunch on every restart.
2. It looks like the MongoDB change streaming can take a "Resume Token". We should keep track of a resume token for each change stream and resume from that after a restart if it's present (and also restart things and trigger a new backfill if the resume token is rejected due to age). This then allows us to have another bit of state for "is a backfill required?" and only do that once.
3. We could get even fancier and make the backfill process incremental across restarts as well if there's a good way to resume that from a specific key, in pretty much the same way that our Firestore capture does. But that may not be necessary if the typical backfill completes in less than an hour.

mdibaiee:
> buffer the entire set of replicated changes in order on disk until the backfill completes

With this, I was digging to see whether we have any constraints on disk usage on connectors: at the momnt we don't, and every container has access to the Kubernetes node disk directly. This means that connectors have the ability to fill in all the node's disk space if they wish to. I think this is something we need to limit.

Once we have limits, I am more comfortable with taking this approach. Unfortunately it seems it's not that easy to limit a docker container's storage size, at least not with the storage driver (`overlay2`) and filesystem (`extfs`) that we use today, and it would probably make local development even harder for us. See [`docker run --storage-opt size=1G` docs](https://docs.docker.com/engine/reference/commandline/run/#workdir)

I'm afraid a mongo connector using the disk as buffer can end up filling the node disk (there needs to be a huge amount of backfill and huge size of events for this to happen) and cause eviction of all of our pods in that node. At the moment the best we can do is limit a pod's storage (rather than the connector containers), which would at least limit the blast radius to a single pod.

mdibaiee:
@willdonnelly I believe I found a way for us not to have to buffer change streams, and that's the `startAtOperationTime` option which allows us to ask for change streams to be replayed from a certain time, so the first time we run the connector, we backfill, and then we ask for changes since 5 seconds before the end of backfill (5 seconds margin of error, which is a lot I think). Let me know what you think of this new approach.

mdibaiee:
This new implementation is however limited by the size of the [`oplog`](https://www.mongodb.com/docs/manual/changeStreams/#footnote-start-time), if the connector is stale for a long enough time that the `oplog` has already cleaned up change events since the last run of the connector, we will have a gap of data, which means we will have to somehow ask the connector to run a backfill again.

I don't think we currently have a nice way to do that... one thing we could do is check for this error in the code and if it happens, clear the checkpoint and restart the connector. I will look into this.

source-mongodb/pull.go
```diff
@@ -0,0 +1,267 @@
...
+func (c *capture) ChangeStream(ctx context.Context, client *mongo.Client, binding uint32, res resource, resumeToken bson.Raw) error {
+	var db = client.Database(res.Database)
+
+	var collection = db.Collection(res.Collection)
+
+	log.Debug("listening on changes on collection ", res.Collection)
+	var eventFilter = bson.D{{"$match", bson.D{{"$or",
+	bson.A{
+		// TODO: handle deletion events
+		// bson.D{{"operationType", "delete"}}
```
willdonnelly, line 124: Do we intend to follow up with deletion support in the near future?

source-mongodb/pull.go
```diff
@@ -0,0 +1,267 @@
...
+	if err := eg.Wait(); err != nil && !errors.Is(err, io.EOF) {
+		return err
+	}
+
+	return nil
+}
+
+type capture struct{
+	Output   *boilerplate.PullOutput
+	BackfillFinishedAt *primitive.Timestamp
```
willdonnelly, line 96: This feels like the wrong way to plumb the "when did the backfill complete" information from `BackfillCollection` to `ChangeStream` -- when there are multiple collection bindings I think there's a race condition if backfills complete quickly and/or the change streaming goroutines take longer than expected to get scheduled?

Perhaps this should simply be a return value from `c.BackfillCollection()` which is then provided as an argument to `c.ChangeStream()`?

source-mongodb/pull.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	cursor, err := collection.Watch(ctx, mongo.Pipeline{eventFilter}, opts)
+	if err != nil {
+		// This error means events from the starting point are no longer available,
+		// which means we have hit a gap in between last run of the connector and
+		// this one. We re-set the checkpoint and error out so the next run of the
+		// connector will run a backfill.
+		if e, ok := err.(mongo.ServerError); ok {
+			if e.HasErrorCode(changeStreamFatalErrorCode) && e.HasErrorMessage(resumePointGoneErrorMessage) {
+				if err = c.Output.Checkpoint([]byte("{}"), false); err != nil {
```
willdonnelly, line 142: This will erase checkpoint state for every collection, right?

I guess that in general all change streams of a capture will be resuming from the same point in time in the oplog, but I'm wondering if for instance it's possible for a collection which changes really infrequently to have an older cursor value (since it looks like we only get a new cursor value alongside a document change event), such that it can't be resumed any longer after a restart but we'd probably prefer for other more recent (which I think would tend to be higher-volume and more massive) collections of the capture to not have to redo the backfill?

source-mongodb/pull.go
```diff
@@ -0,0 +1,267 @@
...
+
+		if i % CHECKPOINT_EVERY == 0 {
+			if err = c.Output.Checkpoint([]byte("{}"), true); err != nil {
+				return fmt.Errorf("output checkpoint failed: %w", err)
+			}
+		}
+	}
+
+	// minus five seconds for potential error
+	c.BackfillFinishedAt = &primitive.Timestamp{ T: uint32(time.Now().Unix()) - 5, I: 0 }
```
willdonnelly, line 236: It feels like this timestamp value ought to be a Go `time.Time` here (something like `time.Now().Sub(backfillOverlapDuration)`) which gets converted to a BSON `primitive.Timestamp` in the `SetStartAtOperationTime` part of the logic.

willdonnelly (approved):
LGTM, a few comments but I think only the global variable `BackfillFinishedAt` race condition one _needs_ to be addressed, the others are just suggestions and questions.

psFried:
> This new implementation is however limited by the size of the [`oplog`](https://www.mongodb.com/docs/manual/changeStreams/#footnote-start-time), if the connector is stale for a long enough time that the `oplog` has already cleaned up change events since the last run of the connector, we will have a gap of data, which means we will have to somehow ask the connector to run a backfill again.
> 
> I don't think we currently have a nice way to do that... one thing we could do is check for this error in the code and if it happens, clear the checkpoint and restart the connector. I will look into this.

Just a drive-by note about this, which you may already be aware of. It's pretty cheap and easy to query the oldest oplog record, and use that to determine whether the oplog contains a complete history since the last run. This is pretty common, since AFAIK mongo oplogs are only constrained by a maximum size in bytes (not time), so actual retention depends on usage.

mdibaiee:
@psFried yeah that's available, in this case I just send a request to resume since the last run and if it errors, that basically says the oplog doesn't have the complete history, so we rewind and backfill.

PR #551, opened 2023-02-23
materialize-bigquery: add logging for connector initialization

williamhbaker:
**Description:**

We've seen a couple of instances of an apparent deadlock with bigquery materializations, where the materialization will apparently start up and not write out any data, even though it does not exit with an error.

The cause of this is not immediately clear, so I am adding some logging for the startup the connector. We currently only see "opening bigquery" if everything works perfectly which is not very useful. The logs added here might be leaning a little towards overkill in the other direction, but they will only be logged when the connector restarts and should be useful for making sure the connector is able to initialize successfully.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/551)

### DIFF
--- a/materialize-bigquery/bigquery.go
+++ b/materialize-bigquery/bigquery.go
@@ -58,15 +58,21 @@ func (c *config) client(ctx context.Context) (*client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("creating bigquery client: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


willdonnelly (approved):
LGTM.

This feels a bit spammy and we might want to consider lowering some of these to debug level in the future, but for now I think this is fine since it's just startup logging.

PR #556, opened 2023-02-27
materialize-mongodb: connector for materializing to mongodb collections

mdibaiee:
**Description:**

- materialize documents to MongoDB collections
- No constraints placed on projections, and JSON documents are just marshalled to BSON and inserted, and unmarshalled from BSON to JSON when loading them

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/556)


materialize-mongodb/driver.go
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  return &pm.ApplyResponse{
+    ActionDescription: "",
+  }, nil
+}
+
+// ApplyDelete is no-op as well, since this operation is going to be superseded
+// by an "Apply" operation with no bindings specified
+func (d driver) ApplyDelete(ctx context.Context, req *pm.ApplyRequest) (*pm.ApplyResponse, error) {
```
williamhbaker, line 125: Unless this connector is using the JSON protocol it will still get `ApplyDelete` requests, right? And if it were using the JSON protcol, my understand is that `ApplyUpsert` would need to handle the "Apply with no bindings" to actually delete the collections in the endpoint.

Is the intent here to simply leave the endpoint collections alone on delete? That wouldn't seem totally unreasonable to me, although it would be different than (most?) other materialization connectors we have which do perform some action on the endpoint when a materialization is deleted from flow.

materialize-mongodb/driver.go
```diff
@@ -0,0 +1,201 @@
...
+  // MongoDB supports arbitrary JSON documents so there are no constraints on
+  // any fields and all fields are marked as "recommended"
+  var out []*pm.ValidateResponse_Binding
+  for _, b := range req.Bindings {
+    res, err := resolveResourceConfig(b.ResourceSpecJson)
+    if err != nil {
+      return nil, err
+    }
+
+    constraints := make(map[string]*pm.Constraint)
```
williamhbaker, line 81: If we're only materializing the flow doc, I think we should require the flow document be included in the projections and actually forbid anything else. This is how the materializations for [pubsub](https://github.com/estuary/connectors/blob/2a7aa9c6a5d7432998859cd8706c1d36240af453/materialize-google-pubsub/driver.go#L143-L155) and [not-really-a-materialization webhooks](https://github.com/estuary/connectors/blob/2a7aa9c6a5d7432998859cd8706c1d36240af453/materialize-webhook/driver.go#L95-L107) do it. This would be more explicit than saying that any field is recommended but then ignoring that set of selected fields and always materializing the entire document anyway.

materialize-mongodb/config.go
```diff
@@ -0,0 +1,62 @@
...
+	uri.User = url.UserPassword(c.User, c.Password)
+
+	if c.Database != "" {
+		uri.Path = "/" + c.Database
+	}
+	return uri.String()
+}
+
+type resource struct {
+	Database   string `json:"database" jsonschema=title=Database name"`
```
williamhbaker, line 53: Do we want to allow delta updates at all?

materialize-mongodb/transactor.go
```diff
@@ -0,0 +1,103 @@
...
+
+		if err := loaded(it.Binding, js); err != nil {
+			return fmt.Errorf("sending loaded: %w", err)
+		}
+	}
+
+	return it.Err()
+}
+
+func (t *transactor) Store(it *pm.StoreIterator) (pm.StartCommitFunc, error) {
```
williamhbaker, line 61: It looks to me like we'll need to do something in the transactor to allow for exactly-once semantics with these loads & updates, otherwise zombie shards would end up racing to re-apply/overwrite & undo updates. I don't know mongodb in-depth, but I wonder if it would support enough transactional ability to serve as an [authoritative remote store](https://docs.estuary.dev/reference/Connectors/materialization-protocol/#common-implementation-patterns)?

materialize-mongodb/transactor.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	return nil, nil
+}
+
+func (t *transactor) Destroy() {
+}
+
+
+func sanitizeDocument(doc map[string]interface{}) map[string]interface{} {
```
williamhbaker, line 90: Does MongoDB not work well with storing the value of `v` here directly, where `math.IsNaN(v) == true`? Putting it another way: I am wondering why this conversion in & out of the store is necessary.

materialize-mongodb/transactor.go
```diff
@@ -0,0 +1,103 @@
...
+		binding := t.bindings[it.Binding]
+
+		var doc bson.M
+		if err := json.Unmarshal(it.RawJSON, &doc); err != nil {
+			return nil, fmt.Errorf("bson unmarshalling json doc: %w", err)
+		}
+		// When we specify the _id field below in the `ReplaceOne` call, the value
+		// of _id for the upserted document will be taken from there, so we can
+		// safely remove this field from the document here to avoid a conflict.
+		delete(doc, idField)
```
williamhbaker, line 74: For my own understanding: Is it correct to say that this would only be strictly necessary if the flow document already had a field `_id` that did _not_ originate from this connector storing it - like if the data were captured from mongodb elsewhere, or for some other reason it had a user-provided `_id`?

williamhbaker (commented):
Some high-level questions and comments. Haven't done a deep-dive yet, but would like to see what you think about these:

materialize-mongodb/driver.go
mdibaiee: I think the idea for removing `ApplyDelete` and making it an `ApplyUpsert` with no bindings is exactly that: to avoid deleting things when a materialization is deleted.

My understanding is, if you run a materialization to a destination, and then decide to remove the materialization from Estuary dashboard, you should be able to keep your data.

materialize-mongodb/transactor.go
mdibaiee: Yes, pretty much. When we specify the key in the `ReplaceOne` operation, MongoDB uses that as `_id`. Now if for any reason there is another `_id` specified that was not the tuple of all keys of the collection, then we would have a conflict. So we delete that field to avoid this conflict.

materialize-mongodb/transactor.go
mdibaiee: @williamhbaker I had not written such a materialization before (the ones I had worked on were recovery log + idempotent apply), so I tried to implement the fencing mechanism for MongoDB. Let me know if it makes sense and the implementation looks okay to you.

There is no `KeyBegin` and `KeyEnd` handling since we don't need to run multiple instances of this connector (at least not now). We can revisit that later if necessary.

materialize-mongodb/driver.go
```diff
@@ -0,0 +1,229 @@
...
+			return nil, nil, fmt.Errorf("bumping fence: %w", err)
+		}
+
+		return &transactor{
+			materialization: materialization,
+			client: client,
+			bindings: bindings,
+			fenceCollection: fenceCollection,
+			fence: &fence,
+		}, &pm.TransactionResponse_Opened{}, nil
```
williamhbaker, line 201: For an authoritative remote store materialization (which it looks like MongoDB transactions will support, which is cool) there needs to be a runtime checkpoint returned here. Otherwise the runtime will continue from its last checkpoint in the recovery log. In that case there is a chance that the commit future that was returned in the transactor `Store` phase failed to actually commit. This would result in missing data from the endpoint.

I think this could be accomplished by storing the runtime checkpoint with the `fenceRecord` as well. Then when the fence record is loaded here, the last persisted runtime checkpoint in the endpoint can be returned with the opened response.

materialize-mongodb/transactor.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	var filter = bson.D{{"materialization", bson.D{{"$eq", t.fence.Materialization}}}, {"fence", bson.D{{"$eq", t.fence.Fence}}}}
+	var fence fenceRecord
+	if err = t.fenceCollection.FindOne(tx, filter, options.FindOne()).Decode(&fence); err != nil {
+		return nil, fmt.Errorf("finding fence: %w", err)
+	}
+
+	return func(ctx context.Context, runtimeCheckpoint []byte, runtimeAckCh <-chan struct{}) (*pf.DriverCheckpoint, pf.OpFuture) {
+		return nil, pf.RunAsyncOperation(func() error {
```
williamhbaker, line 123: Following my previous comment about returning the last persisted runtime checkpoint in the opened response, the runtime checkpoint provided here needs to be included in an update to the fence persisted in the endpoint. It should work similarly to the [postgres materialization](https://github.com/estuary/connectors/blob/f8faa1009d2684201d5711d150527d78b136e6e2/materialize-postgres/driver.go#L390-L404). We'd verify that the nonce on the fence had not changed, and then update it with the new checkpoint as part of the entire store phase transaction.

materialize-mongodb/transactor.go
williamhbaker: Made a couple more comments, looks like it is pretty close. Ignoring materialization sharding for now seems reasonable. I guess the fencing mechanism sans key ranges would cause split shards to fail in that case, but at least they wouldn't overwrite each other's data.

materialize-mongodb/transactor.go
mdibaiee: Perfect thanks, this is the kind of stuff I was afraid I'm missing 👍🏽

materialize-mongodb/driver.go
mdibaiee: Addressed this as well 👍🏽

materialize-mongodb/config.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+
+  "net/url"
+)
+
+type config struct {
+	Address  string `json:"address" jsonschema:"title=Address,description=Host and port of the database." jsonschema_extras:"order=0"`
+	User     string `json:"user" jsonschema:"title=User,description=Database user to connect as." jsonschema_extras:"order=1"`
+	Password string `json:"password" jsonschema:"title=Password,description=Password for the specified database user." jsonschema_extras:"secret=true,order=2"`
+	Database string `json:"database" jsonschema:"title=Database,description=Name of the database to capture from." jsonschema_extras:"order=3"`
```
williamhbaker, line 13: The description here `"Name of the database to capture from"` is probably not what we want.

Also: The top-level config has a database, and each resource has a database. From what I can tell, the top-level config database is only used for storing fence records, whereas each individual resource is free to specify its own database.

I believe there is some potential confusion here that needs to be resolved. My initial thought is that resources should not have their own `database` value and should instead just use the top-level one. This would arguably reduce the flexibility of resource configurations, but I think it would simplify the configuration enough to be worth the tradeoff.

williamhbaker (commented):
Some comments from a detailed review.

I built this locally to try kicking the tires with a local mongodb instance and got an error setting up the materialization of `driver error while validating materialization bobCo/mongodb Caused by: invocation failed: connector failed (exit status: 1) with stderr: connecting to database: connection() error occurred during connection handshake: auth error: sasl conversation error: unable to authenticate using mechanism "SCRAM-SHA-1": (AuthenticationFailed) Authentication failed.`

I assume this means that the connector requires some kind of SSL setup on the database?  I am wondering if that is going to be a problem or not. It seems like it would at least be worth pointing out, either in docs or the spec description.

It would also be nice to have an end-to-end test along the lines of https://github.com/estuary/connectors/tree/main/tests/materialize/materialize-postgres as a sanity check if its at all practical.

materialize-mongodb/transactor.go
```diff
@@ -0,0 +1,157 @@
...
+		if err := json.Unmarshal(it.RawJSON, &doc); err != nil {
+			return nil, fmt.Errorf("bson unmarshalling json doc: %w", err)
+		}
+
+		if binding.res.DeltaUpdates == true {
+			_, err := binding.collection.InsertOne(tx, doc, options.InsertOne())
+			if err != nil {
+				return nil, fmt.Errorf("inserting document into collection %s: %w", binding.collection.Name(), err)
+			}
+		} else {
```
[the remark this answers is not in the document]
mdibaiee: Ah! Today I Learned 🤓

mdibaiee:
@williamhbaker regarding the error you got, yes, you need to add a key, I have an example of it here: https://github.com/estuary/connectors/pull/548

I'll look into adding an end-to-end test, but my initial take on these new connectors is to get an MVP out in hands of customers that are asking for them, and see how it works for them, and we may very well end up reworking the connector (as happened with Firestore, multiple times), and once we have a more stable idea of what the connector is going to be like, we make it more solid

williamhbaker (approved):
LGTM!

PR #560, opened 2023-02-28
source-mongodb: fix address without schema throwing error

mdibaiee:
**Description:**

- Apparently if you give a URL with no scheme to `url.Parse`, it actually parses successfully but gives a URL which has `Scheme` equal to the whole address, with an empty `Host` 😁 

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/560)

### DIFF
--- a/source-mongodb/main.go
+++ b/source-mongodb/main.go
@@ -59,7 +59,7 @@ func (c *config) ToURI() string {
 	var address = c.Address
 	var uri, err = url.Parse(address)
 
-	if err != nil || uri.Scheme == "" {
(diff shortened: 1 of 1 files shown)


willdonnelly:
LGTM, although it feels like if `url.Parse` misbehaves on inputs without a scheme we should maybe be checking for the presence of `:` or `://` explicitly rather than trying to infer that from the parse result?

willdonnelly (approved):

PR #562, opened 2023-03-02
materialize-postgres: convert to Flow JSON protocol

williamhbaker:
**Description:**

This PR converts `materialize-postgres` to use the Flow JSON protocol. The change should be entirely transparent with no change to the connector's behavior from a user's point of view, but it took a bit of work to get there.

The commits lay out these general changes & additions:
- Creating the go types for JSON protocol messages related to materializations
- Creating a "boilerplate" layer for starting up a Go-based materialization connector to read JSON protocol messages and handle them asynchronously
- Adapting `materialize-sql` to use the JSON protocol.
- Adapting `materialize-postgres` to use the JSON protocol via the new `materialize-sql`.

The first commit in the series is entirely a copy of `materialize-sql` to `materialize-sql-json`, which later commits change. A significant portion of the diff is from this trivial copying, so it may be best to review commit-by-commit to see the actual changes that were needed to `materialize-sql`. I've opted to do it this way so that materialization connectors can be updated one-by-one in a somewhat more controlled fashion than changing over all of them simultaneously.

**Workflow steps:**

New materializations can be created using the JSON protocol following the same general pattern here. `materialize-postgres` will continue to function as it always has.

N/A

**Notes for reviewers:**

### DIFF
--- a/go/boilerplate/future.go
+++ b/go/boilerplate/future.go
@@ -1,0 +1,45 @@
+package boilerplate
+
+import (
+	"fmt"
+
+	"go.gazette.dev/core/broker/client"
+)
+
+// OpFuture represents an operation which is executing in the background. The operation has
+// completed when Done selects. Err may be invoked to determine whether the operation succeeded or
+// failed. This is copied from gazette's `client` package.
+type OpFuture interface {
+	// Done selects when operation background execution has finished.
+	Done() <-chan struct{}
+	// Err blocks until Done() and returns the final error of the OpFuture.
+	Err() error
+}
+
+// AsyncOperation is a simple, minimal implementation of the OpFuture interface.
+type AsyncOperation = client.AsyncOperation
+
+// NewAsyncOperation returns a new AsyncOperation.
+var NewAsyncOperation = client.NewAsyncOperation
+
+// FinishedOperation is a convenience that returns an already-resolved AsyncOperation.
+var FinishedOperation = client.FinishedOperation
+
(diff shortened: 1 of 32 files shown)


go/boilerplate/future.go
```diff
@@ -0,0 +1,45 @@
+package boilerplate
```
williamhbaker, line 1: The contents of this file were selectively copied from https://github.com/estuary/flow/blob/master/go/protocols/flow/re_exports.go. `OpFuture` is only used in the Flow repo for the sqlite materialization there & the protobuf-based materialization transactor, both of which should be going away. It makes more sense to have this here where it will be used directly by materialization connectors, and eventually it can be removed entirely from the Flow repo when all materialization connectors have converted to the JSON protocol.

materialize-sql-json/.snapshots/TestTableTemplate
```diff
@@ -1,10 +1,10 @@
 
 	CREATE TABLE one."reserved".checkpoints (
-			materialization TEXT,
-			key_begin BIGINT,
-			key_end BIGINT,
-			fence BIGINT,
-			checkpoint TEXT,
+			materialization TEXT NOT NULL,
```
williamhbaker, line 3: This snapshot change is not a result of this PR. It's from running re-running the sqlgen tests, which apparently don't ever get run in CI.

williamhbaker:
Made a bit of a mess of this one, so will re-open as a new PR cleaned up.

PR #563, opened 2023-03-02
materialize-postgres: convert to Flow JSON protocol

williamhbaker:
**Description:**

This PR converts `materialize-postgres` to use the Flow JSON protocol. The change should be entirely transparent with no change to the connector's behavior from a user's point of view, but it took a bit of work to get there.

The commits lay out these general changes & additions:
- Creating the go types for JSON protocol messages related to materializations
- Creating a "boilerplate" layer for starting up a Go-based materialization connector to read JSON protocol messages and handle them asynchronously
- Adapting `materialize-sql` to use the JSON protocol.
- Adapting `materialize-postgres` to use the JSON protocol via the new `materialize-sql`.

The first commit in the series is entirely a copy of `materialize-sql` to `materialize-sql-json`, which later commits change. A significant portion of the diff is from this trivial copying, so it may be best to review commit-by-commit to see the actual changes that were needed to `materialize-sql`. I've opted to do it this way so that materialization connectors can be updated one-by-one in a somewhat more controlled fashion than changing over all of them simultaneously.

I tested this out fairly extensively locally, particularly when transitioning from the existing connector version to the new connector. This is interesting because existing materializations will have the protobuf spec stored as the materialization spec, whereas the new JSON protocol-based connector stores the bindings. There is a compatibility layer built in to `materialize-sql` to handle this. Once we are confident that the new SQL connectors are working as intended we can modify this compatibility layer to _replace_ existing specs instead of converting them. As it is now, the existing specs will only be replaced on new Apply requests.

**Workflow steps:**

New materializations can be created using the JSON protocol following the same general pattern here. `materialize-postgres` will continue to function as it always has.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/563)

### DIFF
--- a/go/boilerplate/future.go
+++ b/go/boilerplate/future.go
@@ -1,0 +1,45 @@
+package boilerplate
+
+import (
+	"fmt"
+
+	"go.gazette.dev/core/broker/client"
+)
+
+// OpFuture represents an operation which is executing in the background. The operation has
+// completed when Done selects. Err may be invoked to determine whether the operation succeeded or
+// failed. This is copied from gazette's `client` package.
+type OpFuture interface {
+	// Done selects when operation background execution has finished.
+	Done() <-chan struct{}
+	// Err blocks until Done() and returns the final error of the OpFuture.
+	Err() error
+}
+
+// AsyncOperation is a simple, minimal implementation of the OpFuture interface.
+type AsyncOperation = client.AsyncOperation
+
+// NewAsyncOperation returns a new AsyncOperation.
+var NewAsyncOperation = client.NewAsyncOperation
+
+// FinishedOperation is a convenience that returns an already-resolved AsyncOperation.
+var FinishedOperation = client.FinishedOperation
+
+// RunAsyncOperation invokes the given function asynchronously and returns an OpFuture which will
+// resolve with its completion or panic.
+func RunAsyncOperation(fn func() error) OpFuture {
+	var op = NewAsyncOperation()
+
+	go func(op *AsyncOperation) {
+		defer func() {
+			if op != nil {
+				op.Resolve(fmt.Errorf("operation had an internal panic"))
+			}
+		}()
+
+		op.Resolve(fn())
+		op = nil
+	}(op)
+
+	return op
+}
--- a/go/boilerplate/materialize.go
+++ b/go/boilerplate/materialize.go
@@ -1,0 +1,327 @@
+package boilerplate
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"math"
+	"os"
+
+	"github.com/estuary/connectors/go/protocol"
+	"github.com/jessevdk/go-flags"
+	log "github.com/sirupsen/logrus"
+	"go.gazette.dev/core/broker/client"
+)
+
+var (
+	inputReader  io.Reader = os.Stdin
+	outputWriter io.Writer = os.Stdout
+	logWriter    io.Writer = os.Stderr
+)
+
+type MaterializeDriver interface {
+	// Spec should produce a synchronous response with a SpecResponse of the driver's specification.
+	Spec(context.Context, protocol.SpecRequest) (*protocol.SpecResponse, error)
+
+	// Validate should produce a synchronous response with a ValidateResponse of the validated
+	// bindings.
+	Validate(context.Context, protocol.ValidateRequest) (*protocol.ValidateResponse, error)
+
+	// Apply should produce a synchronous response with an ApplyResponse of the actions taken by the
+	// driver.
+	Apply(context.Context, protocol.ApplyRequest) (*protocol.ApplyResponse, error)
+
+	// Open should produce a synchronous response with an OpenResponse optionally containing the
+	// Flow runtime checkpoint that processing should resume from. If empty, the most recent
+	// checkpoint of the Flow recovery log is used.
+	//
+	// Or, a driver may send value []byte{0xf8, 0xff, 0xff, 0xff, 0xf, 0x1} as a base64-encoded
+	// string to explicitly begin processing from a zero-valued checkpoint, effectively rebuilding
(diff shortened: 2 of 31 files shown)


go/boilerplate/future.go
```diff
@@ -0,0 +1,45 @@
+package boilerplate
```
williamhbaker, line 1: The contents of this file were selectively copied from https://github.com/estuary/flow/blob/master/go/protocols/flow/re_exports.go. OpFuture is only used in the Flow repo for the sqlite materialization there & the protobuf-based materialization transactor, both of which should be going away. It makes more sense to have this here where it will be used directly by materialization connectors, and eventually it can be removed entirely from the Flow repo when all materialization connectors have converted to the JSON protocol.

materialize-sql-json/.snapshots/TestTableTemplate
```diff
@@ -0,0 +1,18 @@
+
+	CREATE TABLE one."reserved".checkpoints (
+			materialization TEXT NOT NULL,
```
williamhbaker, line 3: This snapshot change (the part with the `NOT NULL` added) is not a result of this PR. It's from running re-running the sqlgen tests, which apparently don't ever get run in CI.

go/boilerplate/materialize.go
```diff
@@ -0,0 +1,327 @@
+package boilerplate
```
williamhbaker, line 1: For reference, see https://github.com/estuary/flow/blob/master/go/protocols/materialize/transactor.go

Many of the comments and concepts in this file were lifted from there.

materialize-sql-json/driver.go
```diff
@@ -22,91 +27,95 @@ type Driver struct {
...
-var _ pm.DriverServer = &Driver{}
+var _ boilerplate.MaterializeDriver = &Driver{}
 
-// Spec implements the DriverServer interface.
-func (d *Driver) Spec(ctx context.Context, req *pm.SpecRequest) (*pm.SpecResponse, error) {
+func (d *Driver) Spec(ctx context.Context, req protocol.SpecRequest) (*protocol.SpecResponse, error) {
 	var endpoint, resource []byte
+	var err error
 
-	if err := req.Validate(); err != nil {
```
mdibaiee: Why are we removing these `Validate()` calls?

mdibaiee (commented):
Thanks Will, looks good in general just two points

materialize-sql-json/driver.go
williamhbaker: The protobuf types provided comprehensive high-level validation methods that were called here and elsewhere. The paradigm going forward is to trust that data we get from the runtime is correct, since it is responsible for ensuring protocol correctness. The runtime can do these extensive validations in a higher leverage way than every connector needing to implement the validations for its JSON protocol types.

jgraettinger: This is in a bit of flux right now, FYI. As of yesterday, there's a pathway where the protobuf protocol and the JSON protocol are one-and-the-same: the JSON connector protocol is quite literally the canonical JSON encoding of the protobuf types.

That would make it practical to have Go connectors use the generated Go package for protobuf types and decode them as JSON, which would also allow the Validate methods to come along.

Apologies for the back and forth, I've been actively working on figuring this all out.

jgraettinger: None of this is a retraction of the position that connectors can essentially trust what they receive from the runtime, by the way, but it would allow connectors to continue to call Validate as an extra sanity check. I'm not taking a position on whether they _should_.

materialize-sql-json/driver.go
```diff
@@ -22,91 +27,95 @@ type Driver struct {
...
 	return resp, nil
 }
 
-// ApplyUpsert implements the DriverServer interface.
-func (d *Driver) ApplyUpsert(ctx context.Context, req *pm.ApplyRequest) (*pm.ApplyResponse, error) {
+func (d *Driver) Apply(ctx context.Context, req protocol.ApplyRequest) (*protocol.ApplyResponse, error) {
+	// TODO(whb): For now we are only handling the case of deleting the entire materialization,
+	// which will be received as an ApplyRequest with no bindings. Future work could enable handling
+	// of deletion for individual bindings that are removed from a spec while the entire spec is not
+	// being deleted.
```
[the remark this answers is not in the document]
williamhbaker: I do remember the discussion about deleting individual bindings being problematic because other shards of the materialization might be running. I would think that all shards of the materialization would be stopped when doing an `Apply` and then re-started, otherwise the same problem would happen with adding bindings. Maybe that's the magic of handling deletes in `Apply` as well. But I'm not 100% sure either, so maybe @jgraettinger can comment here?

We did talk yesterday about deleting an entire materialization as an apply with no bindings. In that case it's connector-dependent on what they will do. Essentially I am advocating here that we continue to delete the tables from the endpoint as we would have when deleting materialization previously. That's helpful when running our own tests of the materialization, and also for users who are iterating on their specs.

materialize-sql-json/driver.go
williamhbaker: That does seem like some interesting flux, so I'll certainly wait and see how that turns out before I merge this, and probably make some changes here it sounds like. If I'm understanding what you are saying correctly this would obviate the need for a bespoke set of Go JSON structs as I have included here, which would certainly be nice.

I still prefer the idea of the trusting the input from the runtime and not validating it as a rule in connectors. Otherwise we're going to be in an ambiguous place where we are always making sure we are doing something that is largely redundant.

There may be selective cases where a validate method would be handy though. One example is pulling the persisted spec from the SQL endpoint. It could be argued that validating that is not totally redundant since it could have been changed in the table somehow.

williamhbaker:
Going to close this out since we won't be merging it in light of more protocol changes that @jgraettinger has planned. I'll keep the branch around since some of it might be useful for modifying the boilerplate with the new grcp protocol, but other than that no further review is going to be useful here.

PR #564, opened 2023-03-03
Comprehensive prerequisite checking for SQL captures

willdonnelly:
**Description:**

This PR factors out various startup checks and initialization behavior that were previously done in an ad-hoc fashion (typically only in the `Pull` RPC) into a single place for each connector, and collectively names these checks and initialization "prerequisites". In addition, checking (and sometimes initialization) logic has been added for various other connector prerequisites that were not previously checked at all. These checks are now performed consistently as part of a `Validate` operation as well as `Pull`, which means that any setup errors can now be reported while the user is still on the "Create a Capture" page in the UI.

There are, broadly speaking, two sorts of connector deployment:

- Sometimes users will be testing the connector out using a test database and will give it very broad credentials, potentially even the superuser account on the database. In this situation we would like to perform as much setup work as we possibly can automatically -- ideally a completely blank-slate database can automatically have CDC enabled, watermarks table created, and any other setup performed without the user needing to think about it.
- But other times users will want to run the connector against their production database, generally with the most restrictive permissions possible. In this situation we want to _verify_ at startup whether we have the required set of permissions and resources, and provide the user a nice summary of all the things they'll need to fix. Ideally they should be presented with these errors while still on the "Create a Capture" page in the UI, as it's pretty user-unfriendly to let them create a capture that instantly goes red and they may not even notice that.

I have attempted wherever possible to make the validation-and-setup routines flexible, such that if a prerequisite is already satisfied **or** an attempt to set it up ourselves succeeds then we're happy, and if both of those options fail then we report a concise and human-readable description of what's wrong.

There are so many potential combinations of database setup and user permissions that it wasn't really feasible to include automated tests of every permutation in our CI builds. However, I have manually tested the three most important scenarios (these being "prerequisite is met", "prerequisite is not met but we can fix that because we're the superuser", and "prerequisite is not met and we can't fix that because we're a restricted user") for every prerequisite check added in this PR.

**Workflow steps:**

At the start of a `Validate` or `Pull` operation the connector will now perform a series of sanity checks and report all errors encountered during these checks as a bulleted list. In general this **should** be a no-op for preexisting captures that currently work, and if that's not the case then that's probably a bug in the implementation of a particular check.

The most notable workflow change is that now in the "Create a Capture" UI the user may be presented with this list of errors if they need to fix their database configuration. This will hopefully be more intuitive to users than the current state of affairs, where they can successfully create a capture even if the database doesn't even have CDC enabled, but then the capture will immediately show up as failed in the UI and they'll have to dig into the details to find out why, and potentially repeat this multiple times as each setup issue in turn causes the capture to fail as the previous one gets fixed.

None of our connector documentation is necessarily _incorrect_ after these changes, but we may want to revisit the recommended database setup instructions as some parts may be unnecessary.

**Notes for reviewers:**

I have attempted to keep each commit reasonably focused on a single chunk of the overall work, so commit-by-commit is probably the way to read this PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/564)


source-postgres/prerequisites.go
```diff
@@ -22,31 +26,107 @@ func (db *postgresDatabase) SetupPrerequisites(ctx context.Context) []error {
...
 func (db *postgresDatabase) prerequisiteReplicationSlot(ctx context.Context) error {
-	// TODO: Verify that the replication slot exists
+	// If the replication slot already exists then we're satisfied
+	var slotName = db.config.Advanced.SlotName
+	var logEntry = logrus.WithField("slot", slotName)
+	var count int
+	if err := db.conn.QueryRow(ctx, fmt.Sprintf(`SELECT COUNT(*) FROM pg_catalog.pg_replication_slots WHERE slot_name = '%s' AND slot_type = 'logical';`, slotName)).Scan(&count); err != nil {
+		return fmt.Errorf("error querying replication slots: %w", err)
+	}
+	if count == 1 {
```
williamhbaker, line 58: I'm assuming its not possible for this and other similar checks to be > 1?

source-mysql/prerequisites.go
```diff
@@ -19,7 +25,14 @@ func (db *mysqlDatabase) SetupPrerequisites(ctx context.Context) []error {
 }
 
 func (db *mysqlDatabase) prerequisiteBinlogFormat(ctx context.Context) error {
-	// TODO: Verify that 'binlog_format' is set to 'ROW'
+	var results, err = db.conn.Execute(`SELECT @@GLOBAL.binlog_format;`)
+	if err != nil || len(results.Values) == 0 {
+		return fmt.Errorf("unable to query 'binlog_format' system variable: %w", err)
```
williamhbaker, line 34: If `len(results.Values) == 0`, it seems like the returned error will not be clear if `err == nil`.

source-mysql/prerequisites.go
```diff
@@ -19,7 +25,14 @@ func (db *mysqlDatabase) SetupPrerequisites(ctx context.Context) []error {
...
+	}
+	var format = string(results.Values[0][0].AsString())
```
williamhbaker, line 36: Is it possible for `results.Values[0]` to be empty?

williamhbaker (approved):
LGTM

source-postgres/prerequisites.go
willdonnelly: That's correct, the checks where I've written `== 1` are querying by unique columns so it should be impossible for the results to be > 1.

And if things are somehow so badly broken that we get _multiple_ results for a particular replication slot name then it's probably better to fail anyway.

source-mysql/prerequisites.go
willdonnelly: Good point. I'm _fairly_ sure that that should never happen on any version of MySQL that we care about supporting but I may as well separate that out so we produce a cleaner error message.

source-mysql/prerequisites.go
willdonnelly: Only in the pedantic sense that we're communicating with a remote server that could in principle be actively malicious and could send us whatever response data it wants.

But practically speaking no, if we get a result row from this query it will contain one value, which is the current setting of the system variable.

PR #565, opened 2023-03-03
Phil/source http ingest

psFried:
**Description:**

Adds the new `source-http-ingest` connector, which allows users to push data into Flow from HTTP requests. This is the first connector to take advantage of the inbound networking features, which are still in the process of being rolled out. The intent is to also use this
connector for end-to-end testing of connector networking.

I created `https://go.estuary.dev/http-ingest`, but it doesn't point to anything yet.
The README is intended to serve as a first draft for the docs page. Please feel free to suggest improvements to it.
This also includes a bit of general documentation for connector developers about inbound networking. This is potentially something we could consider moving into the main docs site, though I don't think there's any urgency to do so.

**Notes for reviewers:**

I went with Rust for this connector primarily because it makes it easy to link to our existing JSON validation code, but also I just like rust :smile: An interesting future possibility is that we could also link this to our parser code for handling different formats and file uploads. I've elected to leave that out for now. I've also elected to leave out all of the bits for dealing with multiple documents in a single request. Think we'll eventually want both of those capabilities, but I think the UX will be better if we package them as two separate connectors. So this connector is now exclusively focused on requests that provide a single JSON object in the body, and a future connector could handle multiple documents in various formats. This seems like a reasonable separation because it allows the default discovery output to be simpler for the common case where you just want a JSON webhook.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/565)


source-http-ingest/README.md
```diff
@@ -0,0 +1,74 @@
...
+|---|---|---|---|---|
+| **** | EndpointConfig |  | object | Required |
+| `/require_auth_token` |  | Optional bearer token to authenticate webhook requests.<br><br>WARNING: If this is empty or unset, then anyone who knows the URL of the connector will be able to write data to your collections. | null, string | `null` |
+
+## Resource configuration reference
+
+| Property | Title | Description | Type | Required/Default |
+|---|---|---|---|---|
+| **** | ResourceConfig |  | object | Required |
+| `/idFromHeader` |  | Set the &#x2F;&#x5F;meta&#x2F;webhookId from the given HTTP header in each request.<br><br>If not set, then a random id will be generated automatically. If set, then each request will be required to have the header, and the header value will be used as the value of &#x60;&#x2F;&#x5F;meta&#x2F;webhookId&#x60;. | null, string |  |
```
williamhbaker, line 71: These tables look a little messed up?

source-http-ingest/src/main.rs
```diff
@@ -0,0 +1,52 @@
+use anyhow::Context;
+use source_http_ingest::run_connector;
+use tokio::io;
+
+fn main() -> anyhow::Result<()> {
+    let runtime = start_runtime()?;
+
+    let stdin = io::BufReader::new(io::stdin());
+    let stdout = io::stdout();
+    let result = runtime.block_on(run_connector(stdin, stdout));
```
williamhbaker, line 10: I'm just generally curious about this: Is there an advantage to using `block_on` vs. `#[tokio::main]` and `.await`?

source-http-ingest/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+
+    // Check to make sure we can successfully create an openapi spec
+    server::openapi_spec(&config, &typed_bindings)
+        .context("cannot create openapi spec from bindings")?;
+    let response = Response::Validated { bindings: output };
+    write_capture_response(response, &mut stdout).await
+}
+
+pub async fn read_capture_request(stdin: &mut io::BufReader<io::Stdin>) -> anyhow::Result<Request> {
+    let mut buf = String::with_capacity(4096);
```
williamhbaker, line 195: 4096 seems plenty long, but I know _materialization_ bindings are pretty detailed. I'm not sure what capture bindings look like exactly off hand, so this probably isn't an issue. But if they have a ton of bindings I'm not sure what would happen here.

source-http-ingest/src/transactor.rs
```diff
@@ -0,0 +1,144 @@
...
+            .context("serializing response")?;
+        self.0.write_all(&resp).await.context("writing response")?;
+        self.0
+            .write_u8(b'\n')
+            .await
+            .context("writing response newline")?;
+        Ok(())
+    }
+
+    pub async fn commit(&mut self, cp: &impl Serialize, merge_patch: bool) -> anyhow::Result<()> {
```
williamhbaker, line 40: Seems like we're always using the same values for `cp` and `merge_patch` so maybe those don't need to be parameters

source-http-ingest/tests/test.rs
```diff
@@ -0,0 +1,247 @@
...
+                    anyhow::bail!(
+                        "got document with unexpected binding {binding}: {}",
+                        doc.get()
+                    );
+                }
+            }
+            Ok(other) => {
+                let json = serde_json::to_string(&other).unwrap();
+                anyhow::bail!(
+                    "unexpecte response on round {i}, expect_commit: {expect_commit}: {json}"
```
williamhbaker, line 101: typo: `unexpecte`

source-http-ingest/src/server.rs
```diff
@@ -0,0 +1,419 @@
...
+        components = components.security_scheme("bearerAuth", sec);
+    }
+
+    let mut paths = openapi::PathsBuilder::new();
+
+    for binding in bindings.iter() {
+        let url_path = ensure_slash_prefix(binding.resource_path[0].as_str());
+
+        let raw_schema = binding.collection.schema.as_ref().unwrap();
+        let openapi_schema = serde_json::from_str::<openapi::Schema>(raw_schema.get())
```
williamhbaker, line 339: Is it possible for there to be a kind of headers inception here? For example, a schema is inferred on a collection generated from ingested webhooks. Then a user goes to their openapi input and see those headers as inputs and provides them as a manual webhook. But also while doing that headers (maybe the same ones?) are added separately by their browser or whatever.

williamhbaker (approved):

source-http-ingest/README.md
psFried: They still render correctly, though I agree it's ugly in text. These are produced by `flow-schemalate`, though, so I'd like to just address it there.

source-http-ingest/src/main.rs
psFried: The main reason for `block_on` is just so we can use `runtime.shutdown_background()`. This was actually cargo-culted over from [this bit in `connector-init`](https://github.com/estuary/flow/blob/f65a9d7f1a89f897b302f79bb89d53c7de682df3/crates/connector-init/src/main.rs#L42), though I'm pretty sure it would apply here as well.

source-http-ingest/src/lib.rs
psFried: The `4096` is actually just the _initial_ capacity. It'll grow automatically if it needs to. I just tend to be in the habit of giving explicit initial sizes to buffers. But in _this_ case, it's probably not providing any benefit since we're likely to read an entire request's worth of bytes in a single go, anyway.

source-http-ingest/src/transactor.rs
psFried: This is one of the few bits of code that I was thinking would be useful for other capture connectors, which is why I thought to keep it generic. I'm kind of inclined to keep it like that, unless you feel strongly otherwise.

williamhbaker: Nope sounds good to me, I can see how it would be useful to possibly extract in the future as well.

source-http-ingest/src/server.rs
psFried: Yeah, and the server will overwrite the user-provided headers with the ones from the request. I just kinda shrugged it off as "too bad there's nothing we can do", but then I just thought of a solution: `readSchema`!

The openapi spec is generated from the `writeSchema` if one is set. So we can just have discover output a `writeSchema` that _doesn't_ include the headers, and a `readSchema` that does include them.

I'm glad you mentioned it, because I think that could be legitimately confusing to users, and having the headers be included in the examples was just noise in the best case. This _will_ still be a little weird with `/_meta/webhookId`, though, since the id has to be required by the `writeSchema`. So users will still see a `webhookId` in the example documents.

psFried: Oh, nevermind then. The DiscoverResponse only has `documentSchemaJson`. There's no way to return separate read vs write schemas at this point. The only other idea that had come to mind was to have a special case in the server that would try remove those fields from the schema when it generates the openapi schema. That seemed a little too gross to do at this point, but now it's kinda seeming like A, our only option, and B, not any worse than providing a confusing UX. I'll give it a shot and see how bad it is.

psFried:
I've been thinking more about the issues in [this thread](https://github.com/estuary/connectors/pull/565#discussion_r1125098968) and I _think_ I may have a better solution. To restate the issue with the original design, it's that the schema we use with the openapi spec contains fields that are automatically added by the connector. This gives the impression that the webhook payload must already include those fields, which is pretty confusing for users.

The solution I ended up with, after exploring several alternatives, is to leave the schemas the way they are, and to just provide an explicit example document that's more minimal. This is what the swagger UI looks like now:

image

The connector allows binding to any Flow collection, and the schemas and keys of those collections can be anything you want. This makes it potentially a little awkward to just show `{"hello": "world!"}` when you're ingesting to a collection that _doesn't_ accept any valid json object. For this reason, that minimal example document is only used if the collection key is `[/_meta/webhookId]`. This is an imperfect heuristic, but seems to provide a smooth UX in all the common cases, while maintaining an acceptable failure mode in the edge case that someone binds a collection with a key of [`/_meta/webhookId]` and a schema that _doen't_ accept the example document (it just returns a validation error).

williamhbaker:
> I've been thinking more about the issues in [this thread](https://github.com/estuary/connectors/pull/565#discussion_r1125098968) and I _think_ I may have a better solution. To restate the issue with the original design, it's that the schema we use with the openapi spec contains fields that are automatically added by the connector. This gives the impression that the webhook payload must already include those fields, which is pretty confusing for users.
> 
> The solution I ended up with, after exploring several alternatives, is to leave the schemas the way they are, and to just provide an explicit example document that's more minimal. This is what the swagger UI looks like now:
> 
> image
> 
> The connector allows binding to any Flow collection, and the schemas and keys of those collections can be anything you want. This makes it potentially a little awkward to just show `{"hello": "world!"}` when you're ingesting to a collection that _doesn't_ accept any valid json object. For this reason, that minimal example document is only used if the collection key is `[/_meta/webhookId]`. This is an imperfect heuristic, but seems to provide a smooth UX in all the common cases, while maintaining an acceptable failure mode in the edge case that someone binds a collection with a key of [`/_meta/webhookId]` and a schema that _doen't_ accept the example document (it just returns a validation error).

This seems pretty reasonable and quite a good improvement for the typical case. Since removing the `_meta` fields from the example schema is not practical (I think that's what you're saying at least) it seems like the change you've got here is a good way to make the connector most usable.

PR #569, opened 2023-03-07
materialize-sqlite: migrate to use materialize-sql

mdibaiee:
**Description:**

- Migrate `materialize-sqlite` from `flow` repository here and use our `materialize-sql` library.
- I had to make `MetaSpecs` optional since given that sqlite is ephemeral, there is no persistence between `ApplyUpsert` and `Transactions` steps, and in general neither checkpoints nor spec validation don't make sense for an ephemeral database
- I also removed fencing since that is also not necessary here
- At the moment the way to inspect the database is to SSH into it and use `sqlite3 /tmp/sqlite.db`. We can add an API on top of this database and use connector connectivity to expose that API so we can use it in the UI / elsewhere.
- One step towards https://github.com/estuary/flow/issues/442

**Workflow steps:**

```
collections:
  acmeCo/greetings:
    schema:
         properties:
           count:
             type: integer
           message:
             type: string
         required:
           - count
           - message
         type: object

    key: [/count]
captures:
  acmeCo/source-test:
    endpoint:
      connector:
        image: ghcr.io/estuary/source-test:v1@sha256:48399a81bed2887c16fb0dd0c6bbd62beb260537e48eabcbd3bf87b7c6958c4b
        config: { "greetings": 5, "skip_state": false }
    bindings:
      - resource:
          stream: greetings
          syncMode: incremental
        target: acmeCo/greetings
materializations:
  acmeCo/sqlite:
    endpoint:
      connector:
        image: materialize-sqlite:local
        config: {}
    bindings:
      - resource:
          table: greetings
        source: acmeCo/greetings
storageMappings:
    "":
        stores:
            - provider: S3
              bucket: acmeCo-source-hubspot
```

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/569)


materialize-sql/driver.go
```diff
@@ -287,16 +296,19 @@ func (d *Driver) newTransactor(ctx context.Context, open pm.TransactionRequest_O
 	var loadedVersion string
 
 	var endpoint, err = d.NewEndpoint(ctx, open.Materialization.EndpointSpecJson)
```
williamhbaker, line 298: I think we'd still want to unconditionally check for `err` here and return if it isn't `nil`, to guard against `endpoint` being `nil` and causing problems later

materialize-sqlite/sqlite.go
```diff
@@ -0,0 +1,357 @@
...
+func (c tableConfig) GetAdditionalSql() string {
+	return ""
+}
+
+func (c tableConfig) Path() sql.TablePath {
+	return []string{c.Table}
+}
+
+func (c tableConfig) DeltaUpdates() bool {
+	return false // SQLite doesn't support delta updates.
```
williamhbaker, line 46: This probably isn't a big deal now, but I don't see any reason why it _couldn't_ support delta updates.

materialize-sqlite/sqlite.go
```diff
@@ -0,0 +1,357 @@
...
+	return "", "", stdsql.ErrNoRows
+}
+
+func (c client) ExecStatements(ctx context.Context, statements []string) error {
+	return c.withDB(func(db *stdsql.DB) error { return sql.StdSQLExecStatements(ctx, db, statements) })
+}
+
+// We don't need a fence since there is only going to be a single sqlite
+// materialization instance writing to the file, and it's ephemeral
+func (c client) InstallFence(ctx context.Context, checkpoints sql.Table, fence sql.Fence) (sql.Fence, error) {
```
williamhbaker, line 106: This is interesting: Since the driver won't supply a checkpoint to the runtime, the runtime will always resume from the recovery log checkpoint. The alternative would be re-starting the materialization from scratch (by returning `pm.ExplicitZeroCheckpoint`). Starting from scratch every time would probably not be great since it could result in huge data transfers every time the connector restarts. Starting from the recovery log checkpoint might be a little surprising to users. I think thats still the best option though since this is meant for internal testing and demos.

williamhbaker (approved):
LGTM.

One high-level comment: I've had issues in the past with concurrent writes resulting in the error `database is locked` with the `go-sqlite3` driver. We might have a tiny chance of that happening here with the load & store phases operating concurrently. I've used client-side locks to handle this before to ensure there is only a single writer at a time. Practically speaking I don't know if we'll ever see that in this connector since its at most 2 threads that will be concurrently writing but it could be something to keep in mind in the future.

williamhbaker:
Another thought: Should we limit how big the database can be somehow? Since it is storing the DB file to disk it might take up quite a lot of space. There are some large collections (wikipedia recent changes) and I am sure users will try to materialize those to sqlite to try things out.

materialize-sqlite/sqlite.go
mdibaiee: Yeah this is a bit ambiguous at the moment, we can play with the current approach and if we have a reason to use a zero checkpoint, we'll go for that instead.

materialize-sqlite/.snapshots/TestSQLiteDriver
```diff
@@ -0,0 +1,11 @@
+test_target:
+theKey, bool, int, string, flow_document
+key1Value, totally different, 0, 33, { "theKey": "key1Value", "string": "notthesame", "bool": false, "int": 33, "number": 2 }
+key2Value, bar, 0, 88, { "theKey": "key2Value", "string": "bar", "bool": false, "int": 88, "number": 56.78 }
+key3Value, baz, 0, 99, { "theKey": "key3Value", "string": "baz", "bool": false, "int": 99, "number": 0 }
+key4Value, <nil>, <nil>, <nil>, { "theKey": "key4Value" }
+
+flow_checkpoints_v1:
```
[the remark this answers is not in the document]
mdibaiee: leftovers!

mdibaiee:
>  Another thought: Should we limit how big the database can be somehow? Since it is storing the DB file to disk it might take up quite a lot of space. There are some large collections (wikipedia recent changes) and I am sure users will try to materialize those to sqlite to try things out.

This is a good point, but I'm not sure how we would do this. Checking the file size of the database before every transaction to see if it's exceeding a specific size and erroring if it does?

If we error, that makes the connector restart and lose its data up to that point. I think that's not great for demoing / experimenting purposes. I would ideally be able to keep my data up to that point, but not add more data. If the user actually wants to clean that data and get new data in, they can disable / enable it in the UI.

If we decide to go ahead with this, that would mean we just halt the connector's progress once we hit the size limit.

PR #574, opened 2023-03-08
source-alpaca: set a minimum start date

williamhbaker:
**Description:**

Data is not available from the historical Alpaca APIs until 01-01-2016. If somebody puts in a start time way before that, the connector will churn through a potentially large amount of empty time ranges while backfilling and take a long time to actually start producing data. This adds a constraint to the configuration that a start date no earlier than 01-01-2016 can be used.

**Workflow steps:**

See https://github.com/estuary/flow/pull/964

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/574)

### DIFF
--- a/source-alpaca/.snapshots/TestSpec
+++ b/source-alpaca/.snapshots/TestSpec
@@ -34,7 +34,7 @@
         "type": "string",
         "format": "date-time",
         "title": "Start Date",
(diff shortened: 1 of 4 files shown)


mdibaiee (approved):
LGTM, thanks!

PR #576, opened 2023-03-09
Add additional tags to connectors

jman0119:
**Description:**

Added additional tags to connectors for Amazon Aurora and Azure SQL Server

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -84,11 +84,15 @@ jobs:
           - path: source-mysql
             additionalTags:
               - source-mariadb
+              - source-amazonaurora
           - path: source-postgres
             additionalTags:
               - source-alloydb
+              - source-amazonaurora
           - path: source-s3
           - path: source-sqlserver
+            additionalTags:
+              - source-azuresqlserver
           - path: materialize-bigquery
           - path: materialize-elasticsearch
           - path: materialize-firebolt
@@ -99,6 +103,7 @@ jobs:
             additionalTags:
               - materialize-timescaledb
               - materialize-alloydb
+              - materialize-amazonaurora
           - path: materialize-rockset
           - path: materialize-s3-parquet
           - path: materialize-snowflake


.github/workflows/ci.yaml
```diff
@@ -84,11 +84,15 @@ jobs:
           - path: source-mysql
             additionalTags:
               - source-mariadb
+              - source-amazonaurora
           - path: source-postgres
             additionalTags:
               - source-alloydb
+              - source-amazonaurora
```
mdibaiee, line 91: These two source connectors will collide with each other. I would suggest we name them like this so we can build two separate images:

```
source-amazon-aurora-mysql
source-amazon-aurora-postgres
```

.github/workflows/ci.yaml
```diff
@@ -99,6 +103,7 @@ jobs:
             additionalTags:
               - materialize-timescaledb
               - materialize-alloydb
+              - materialize-amazonaurora
```
mdibaiee, line 106: for the sake of convention / predictability of names, let's also make this `materialize-amazon-aurora`

.github/workflows/ci.yaml
```diff
@@ -84,11 +84,15 @@ jobs:
...
               - source-mariadb
+              - source-amazon-aurora-mysql
           - path: source-postgres
             additionalTags:
               - source-alloydb
+              - source-amazon-aurora-postgres
           - path: source-s3
           - path: source-sqlserver
+            additionalTags:
+              - source-azuresqlserver
```
mdibaiee, line 95: and `source-azure-sqlserver` here I think would suit our other names best

.github/workflows/ci.yaml
jman0119: should I add -postgres too just to be consistent with how we named in for the source?

mdibaiee (approved):
LGTM, thank you Jenny!

Issue #428, opened 2022-11-23
Request a connector to materialize to redshift


mdibaiee:
### System Name

redshift

### Type

Capture

### Details

From Will Donnely:

I'm not really familiar with Redshift, so I did a bit of reading. Key points:
- It looks like there are basically two ways of getting data into Redshift: [streaming ingestion](https://aws.amazon.com/redshift/redshift-streaming-ingestion/) via Kinesis and [bulk data loading](https://docs.aws.amazon.com/redshift/latest/gsg/rs-gsg-create-sample-db.html) from S3.
- The batch data loading stuff [may already be compatible](https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-format.html#copy-parquet) with what we do in materialize-s3-parquet so it's possible that "Redshift Support" could be as minimal as a connector that does that and then issues COPY FROM queries to trigger bulk loading after it writes out a file.
- It looks like all of the [setup instructions](https://docs.aws.amazon.com/redshift/latest/dg/materialized-view-streaming-ingestion.html) for streaming ingestion are basically "given a Kinesis data stream, here's how you tell Redshift to consume it", so realistically we'd just be writing a materialize-kinesis connector in that case and then it's up to the user to configure Redshift to read from that stream
- It seems...not super complicated to write a materialize-kinesis connector if we're not trying to get fancy with it (delta-updates only, just write records via the [Kinesis API](https://docs.aws.amazon.com/sdk-for-go/api/service/kinesis/#Kinesis.PutRecords) without much fuss). But I'm not sure how that approach would behave when transactions get retried, so it might need to be more complex than that

PR #580, opened 2023-03-11
materialize-redshift: new connector

williamhbaker:
**Description:**

Adds a new connector: `materialize-redshift`

This materialization is based on `materialize-sql` and works most similarly to Bigquery: Data is staged in cloud storage files (`jsonl` in S3 in this case), and then merged into the target table for each binding.

Redshift SQL is based on Postgres and is generally compatible with Postgres SQL. `pgx` in particular has some history of fixing issues & maintaining support for Redshift and worked well for what this connector needs to do.

Even when streaming small incremental updates the stage & load strategy has surprisingly good performance from what I can tell. And it can load large amounts of data (backfill or long configured transaction duration) _very_ efficiently, really only being limited by network bandwidth.

I did some pretty extensive local testing with a wide range of data types, configurations, and data volumes. Aside from some nuances with float fields and timestamps the redshift materialization produces an identical output as a postgres materialization when the endpoint tables are compared.

I don't love that configuring the connector will need the database host & credentials in addition to the S3 service account key credentials for interacting with S3. At this point that is the best we can do though. Streamlining this would be quite involved and could require things like us operating our own staging bucket for individual materializations and supporting IAM role based auth.

Closes https://github.com/estuary/connectors/issues/428

**Workflow steps:**

Use the new connector to materialize collections to redshift

https://github.com/estuary/flow/pull/965

**Notes for reviewers:**

The commit history is an evolution from an initial working version of the materialization to a more refined state, with incremental improvements along the way. See commit messages for additional context that may be helpful.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/580)


mdibaiee (approved):
LGTM, thanks Will! I added two comments to the docs PR

PR #581, opened 2023-03-13
materialize-sql: schema migration to alter table and add new column

mdibaiee:
**Description:**

- Allow new fields to be added to bindings, and in turn add new columns to existing tables for `materialize-sql` connectors. I've tested `materialize-postgres` so far.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/581)

### DIFF
--- a/materialize-sql/.snapshots/TestAlterTableAddColumnTemplate
+++ b/materialize-sql/.snapshots/TestAlterTableAddColumnTemplate
@@ -1,0 +1,4 @@
+
+	ALTER TABLE one."reserved".checkpoints ADD COLUMN
+		checkpoint TEXT NOT NULL;
+  
--- a/materialize-sql/driver.go
+++ b/materialize-sql/driver.go
@@ -143,7 +143,7 @@ func (d *Driver) ApplyUpsert(ctx context.Context, req *pm.ApplyRequest) (*pm.App
 			// fields in the request is identical to the current fields.
 			return nil, fmt.Errorf(
 				"the set of binding %s fields in the request differs from the existing fields,"+
-					"which is disallowed because this driver only supports autoamtic migration for adding new fields. "+
+					"which is disallowed because this driver only supports automatic migration for adding new fields. "+
 					"Request fields: %s , Existing fields: %s",
 				collection,
 				bindingSpec.FieldSelection.String(),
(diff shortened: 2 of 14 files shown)


materialize-sql/.snapshots/TestAlterTableAddColumnTemplate
```diff
@@ -0,0 +1,4 @@
+
+	ALTER TABLE one."reserved".checkpoints ADD COLUMN
+		checkpoint TEXT NOT NULL;
```
mdibaiee, line 3: This example is a bit unintuitive because we don't allow `NOT NULL` columns for this migration but the way the test is written it has everything mapped to be `NOT NULL`. I've left it as-is since it's only testing the templating.

materialize-sql/driver.go
```diff
@@ -138,41 +138,58 @@ func (d *Driver) ApplyUpsert(ctx context.Context, req *pm.ApplyRequest) (*pm.App
...
 			return nil, fmt.Errorf("binding for %s: %w", collection, err)
-		} else if loadedBinding != nil && !bindingSpec.FieldSelection.Equal(&loadedBinding.FieldSelection) {
-			// We don't handle any form of schema migrations, so we require that the list of
+		} else if loadedBinding != nil && !bindingSpec.FieldSelection.Equal(&loadedBinding.FieldSelection) && !isBindingMigratable(loadedBinding, bindingSpec) {
+			// if the binding is not migratable, we require that the list of
 			// fields in the request is identical to the current fields.
 			return nil, fmt.Errorf(
 				"the set of binding %s fields in the request differs from the existing fields,"+
-					"which is disallowed because this driver does not perform schema migrations. "+
+					"which is disallowed because this driver only supports autoamtic migration for adding new fields. "+
```
williamhbaker, line 146: typo: `autoamtic`

materialize-sql/validate.go
```diff
@@ -111,9 +116,7 @@ func ValidateMatchesExisting(existing *pf.MaterializationSpec_Binding, proposed
 	// mention are implicitly forbidden.
```
williamhbaker, line 116: The comment just above here `...and forbid any that aren't already in our map` could perhaps be updated, since that won't be the case anymore.

williamhbaker (approved):
LGTM

PR #590, opened 2023-03-14
materialize-snowflake: sql fixes for delta_updates and no selected values

williamhbaker:
**Description:**

Fixes a couple of edge cases for the generated sql queries in `materialize-snowflake`:
- When a binding is configured with `delta_updates` and `flow_document` is excluded as a field
- When there are no non-key fields selected, AKA there are no value fields being materialized

Both of these are uncommon but can happen, and we've seen examples of them with other connectors.

Closes https://github.com/estuary/connectors/issues/589

**Workflow steps:**

N/A

**Notes for reviewers:**

The first commit is purely a refactoring of the sql generation tests to make it easier to see the diffs from the next couple of changes which are what actually modify the generated queries.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/590)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -6,6 +6,7 @@ import (
 	stdsql "database/sql"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -523,4 +524,10 @@ func (f *scratchFile) put(ctx context.Context, conn *stdsql.Conn, cfg *config) e
 	return nil
 }
 
-func main() { boilerplate.RunMain(newSnowflakeDriver()) }
+func main() {
+	// gosnowflake also uses logrus for logging and the logs it produces may be confusing when
+	// intermixed with our connector logs. We disable the gosnowflake logger here and log as needed
+	// when handling errors from the sql driver.
+	sf.GetLogger().SetOutput(io.Discard)
+	boilerplate.RunMain(newSnowflakeDriver())
+}
--- a/materialize-snowflake/sqlgen.go
+++ b/materialize-snowflake/sqlgen.go
@@ -134,8 +134,12 @@ ALTER TABLE {{ $.Table.Identifier }} ADD COLUMN
 	{{- end }};
 {{ end }}
 
+-- Templated query which joins keys from the load table with the target table, and returns values. It
+-- deliberately skips the trailing semi-colon as these queries are composed with a UNION ALL.
+
 {{ define "loadQuery" }}
-	SELECT {{ $.Table.Binding }}, {{ $.Table.Identifier }}.{{ $.Table.Document.Identifier }}
+{{ if $.Table.Document -}}
+SELECT {{ $.Table.Binding }}, {{ $.Table.Identifier }}.{{ $.Table.Document.Identifier }}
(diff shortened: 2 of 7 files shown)


williamhbaker:
Added an additional commit that disables the `gosnowflake` logger - It is also using `logrus` and the logs it was emitting were problematic for ex: When applying a new materialization and the `flow_materializations_v2` table is not found, it would log an `ERROR` level message, even though we expect this table to not be found sometimes (it will automatically be created) & we handle logging appropriately.

materialize-snowflake/snowflake.go
```diff
@@ -523,4 +524,10 @@ func (f *scratchFile) put(ctx context.Context, conn *stdsql.Conn, cfg *config) e
 	return nil
 }
 
-func main() { boilerplate.RunMain(newSnowflakeDriver()) }
+func main() {
+	// gosnowflake also uses logrus for logging and the logs it produces may be confusing when
+	// intermixed with our connector logs. We disable the gosnowflake logger here and log as needed
+	// when handling errors from the sql driver.
+	sf.GetLogger().SetOutput(io.Discard)
```
mdibaiee, line 531: I'm wondering if we can instead do
```
var logger = log.New()
sf.GetLogger().SetOutput(logger.Writer())
```

Ideally we would have the same instance of `logger` as the rest of the code so it has the same settings, but the idea is instead of discarding these logs, send them to logrus `io.Writer`

materialize-snowflake/sqlgen.go
```diff
@@ -134,8 +134,12 @@ ALTER TABLE {{ $.Table.Identifier }} ADD COLUMN
 	{{- end }};
 {{ end }}
 
+-- Templated query which joins keys from the load table with the target table, and returns values. It
+-- deliberately skips the trailing semi-colon as these queries are composed with a UNION ALL.
+
 {{ define "loadQuery" }}
-	SELECT {{ $.Table.Binding }}, {{ $.Table.Identifier }}.{{ $.Table.Document.Identifier }}
+{{ if $.Table.Document -}}
```
mdibaiee, line 141: Is it only when delta updates are enabled that there is no Document? 🤔

mdibaiee (approved):
LGTM, thanks Will. Two small comments below

materialize-snowflake/sqlgen.go
williamhbaker: It's only possibly to explicitly exclude the `flow_document` if delta updates are enabled, otherwise it is required: https://github.com/estuary/connectors/blob/88fe5895ef5df5c90e7cd8930f2e3bcf19113abf/materialize-sql/validate.go#L71-L76

And by default even with delta updates enabled `flow_document` is recommended, so it will almost always still be there. But we do actually have at least one production task running that is configured like this, for another materialization.

materialize-snowflake/snowflake.go
williamhbaker: Hmm I tried this and it still logs as an `ERROR` level. I think the `Writer()` is set at `info` level, but the logs it writes are still at the level they are logged at - just that it won't log anything below `info`. Aside from that, the internal logs from the `gosnowflake` driver aren't very user friendly and might be confusing for a user so its nice to completely suppress them.

PR #593, opened 2023-03-15
source-postgres: Support UUIDs as primary keys

willdonnelly:
**Description:**

Adds support for Postgres captures with UUID columns in the primary key.

The UUID is represented as a `[16]uint8` when we receive it from the Postgres client library, but can be sliced into a `[]uint8` and the client library will accept that later on when we use it as a resume cursor value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/593)

### DIFF
--- a/source-postgres/.snapshots/TestScanKeyTypes-UUID
+++ b/source-postgres/.snapshots/TestScanKeyTypes-UUID
@@ -1,0 +1,49 @@
+####################################
+### Capture from Start
+####################################
+# ================================
+# Collection "test.ScanKeyTypes_UUID": 1 Documents
+# ================================
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"scankeytypes_uuid","loc":[11111111,11111111,11111111]}},"data":"Data 1","key":"4ab4044a-9aab-415c-96c6-17fa338060fa"}
+# ================================
+# Final State Checkpoint
+# ================================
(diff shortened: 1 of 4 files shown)


jgraettinger:
Confirming my understanding here:

* When capturing a UUID, it's always captured in it's canonical string encoding.
* When the _runtime_ packs a UUID into a tuple, it's packing this canonical string encoding (not its bytes). So, the encoding behavior is different here.
* BUT that's okay, because these FDB keys are strictly an internal implementation detail of source-postgres, and it only needs to be consistent within the connector.

If that's all true than LGTM

jgraettinger:
I might still push for encoding the _string_ representation of the UUID, just for overall cohesion and consistency, and to avoid any potential of a future surprise, but I can't forsee a scenario where it would matter and this isn't high conviction.

willdonnelly:
That's correct, the serialization pathways for capture output and for row key persistence are entirely separate, so this conversion is purely an internal concern. The only properties the row key encoding needs to have are that it preserves key equality/ordering and that we can decode it back into *something* the Postgres client library will accept as a backfill cursor.

For JSON purposes we're going straight from `[16]uint8` to the canonical string encoding, independent of the translation occurring here.

That said this would probably work fine if I format it as a string instead, now that you mention it

jgraettinger (approved):
Still LGTM

PR #599, opened 2023-03-16
allow extending connector images as variants with alternate docs links

williamhbaker:
**Description:**

We re-tag a few connector images as identical variants of their base image. They are exactly the same other than the image tag. Currently these variants produce the same documentation URL in their spec response as the base image, resulting in a somewhat confusing experience when the docs link is followed in the web app, or when the sidebar displays the connector docs.

This PR provides a way for setting an alternative documentation URL for these variants.

It is mostly handled by `connector-variant.Dockerfile` which is a very simple extension of the base dockerfile that allows for a build arg to set the documentation URL. This is made available to the connector code through an environment variable. I've updated the SQL capture & SQL materialization packages to use this environment variable if present - we only currently build variants for these base images.

Configuration of building the variants is still handled in CI, but as an additional job. The variant images depend on the base connector images being available, so building them must come after. A side effect of this & the way github matrix jobs work is that in order to build the variants, all base connector matrix jobs must be successful. So if we have a lot of flaky tests it may provide difficult for the variants to be built. I don't think this is unreasonable since we should endeavor to have a reliable test pipeline, but its something to keep in mind.

Closes https://github.com/estuary/flow/issues/967

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/599)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,45 +71,33 @@ jobs:
       fail-fast: false
       matrix:
         connector:
-          - path: source-alpaca
-          - path: source-gcs
-          - path: source-hello-world
-          - path: source-http-file
-          - path: source-http-ingest
-          - path: source-test
-          - path: source-firestore
-          - path: source-kafka
-          - path: source-kinesis
-          - path: source-mongodb
-          - path: source-mysql
-            additionalTags:
-              - source-mariadb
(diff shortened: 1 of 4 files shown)


willdonnelly (approved):
LGTM

PR #601, opened 2023-03-16
return the error when network tunnel Start() fails

willdonnelly:
**Description:**

Previously this would just be logged and then execution would continue onwards and attempt to connect to localhost, which is confusing and results in sub-optimal error messages for the user.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/601)


williamhbaker (approved):
LGTM!

PR #605, opened 2023-03-17
source-postgres: Disable 'user has REPLICATION' startup check

willdonnelly:
**Description:**

This may be failing in some setups, which would imply that the method we're using to check that permission is insufficiently nuanced.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/605)


willdonnelly:
I am fairly confident that this is the correct fix, but CI builds are having trouble. I'm going to go ahead and merge this unreviewed so once CI works the fix will be live ASAP.

PR #607, opened 2023-03-20
materialize-sql: allow migration to remove a field (mark nullable)

mdibaiee:
**Description:**

- Allow migration to remove a field, this will lead to the column being marked as nullable and left in the database. Does not allow removing of key fields.

**Workflow steps:**

- Remove a non-key field from your collection

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/607)

### DIFF
--- a/materialize-sql/validate.go
+++ b/materialize-sql/validate.go
@@ -94,19 +94,24 @@ func validateNewProjection(resource Resource, projection *pf.Projection) *pm.Con
 
 // ValidateMatchesExisting returns a set of constraints to use when there is a new proposed
 // CollectionSpec for a materialization that is already running, or has been Applied. The returned
-// constraints will explicitly require all fields that are currently materialized, as long as they
-// are not unsatisfiable, and give new constraints for additionally supplied fields.
+// constraints will check type compatibility of projections that exist both in
+// the new and the existing binding, and give new constraints for additionally supplied fields.
 func ValidateMatchesExisting(resource Resource, existing *pf.MaterializationSpec_Binding, proposed *pf.CollectionSpec) map[string]*pm.Constraint {
 	var constraints = make(map[string]*pm.Constraint)
 	for _, field := range existing.FieldSelection.AllFields() {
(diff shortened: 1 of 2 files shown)


materialize-sql/validate.go
```diff
@@ -100,13 +100,16 @@ func ValidateMatchesExisting(resource Resource, existing *pf.MaterializationSpec
 	var constraints = make(map[string]*pm.Constraint)
```
williamhbaker, line 100: Nit: Might update the comment for this function where it says `...will explicitly require all fields that are currently materialized...`

materialize-sql/validate.go
```diff
@@ -124,12 +127,11 @@ func ValidateMatchesExisting(resource Resource, existing *pf.MaterializationSpec
 }
 
 func checkTypeError(field string, existing *pf.CollectionSpec, proposed *pf.CollectionSpec) string {
-	// existingProjection is guaranteed to exist since the MaterializationSpec has already been
-	// validated.
 	var existingProjection = existing.GetProjection(field)
 	var proposedProjection = proposed.GetProjection(field)
-	if proposedProjection == nil {
```
williamhbaker, line 131: If I'm following this correctly, we don't need the `proposedProjection == nil` check here because of the change in this PR & specifically the addition on line 103-105 means we won't even call `checkTypeError` when removing a field.

I'm trying to grok the addition of `existingProjection == nil` below. Since `checkTypeError` is being called out of a loop on `existing.FieldSelection.AllFields()` I would think that `existingProjection` would never be `nil`?

williamhbaker (approved):
LGTM, with a question about `checkTypeError`.

materialize-sql/validate.go
mdibaiee: This actually reminded me to add an additional check here to validate that keys from the existing materialization are present in the new materialization as well so we don't allow the validate step to pass only to error on ApplyUpsert

PR #608, opened 2023-03-20
source-mongodb: use one document to infer type of _id

### LINKED ISSUE #592: source-mongodb: be smarter about type of _id
At the moment we assume `_id` to be a string, but it's possible for our users to want to use a different type for their `_id` fields, such as integer, or even an object.

We can be smarter during discovery and suggest a compatible type for `_id` if we can find any documents in the collection that have `_id` with a different type than string or ObjectId.

mdibaiee:
**Description:**

- Look for one document in each collection during discover to infer type of `_id`, currently supporting `integer` and `number` (I doubt anyone uses this). We previously had a customer who had numerical IDs.
- Resolves #592 

**Workflow steps:**

- Create a collection with integer `_id` and run discover

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/608)

### DIFF
--- a/source-mongodb/pull.go
+++ b/source-mongodb/pull.go
@@ -277,15 +277,7 @@ func sanitizeDocument(doc map[string]interface{}) map[string]interface{} {
 	for key, value := range doc {
 		// Make sure `_id` is always captured as string
 		if key == "_id" {
-			switch value.(type) {
-				case string:
-				default:
-					var j, err = json.Marshal(value)
-					if err != nil {
-						panic(fmt.Sprintf("could not marshal interface{} to json: %s", err))
-					}
-					doc[key] = string(j)
-			}
+			doc[key] = idToString(value)
 		} else {
 			switch v := value.(type) {
 			case float64:
@@ -299,4 +291,19 @@ func sanitizeDocument(doc map[string]interface{}) map[string]interface{} {
 	}
 
 	return doc
+}
+
+func idToString(value interface{}) string {
+	switch v := value.(type) {
+		case string:
+			return v
+		case primitive.ObjectID:
+			return v.Hex()
+	}
+
+	var j, err = json.Marshal(value)
+	if err != nil {
+		panic(fmt.Sprintf("could not marshal interface{} to json: %s", err))
+	}
+	return string(j)
 }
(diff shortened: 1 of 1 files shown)


mdibaiee:
> Is it possible for a mongo collection to have documents with varying types of _id's? If it is, my thought would to have our collection always be type string for _id, and coerce whatever is in the monog collection to a string perhaps in sanitizeDocument.

I don't see anything in the docs suggesting that it is not possible, so I guess it is possible. That's interesting, on another thought, since we use schema inference for mongodb, if they have integer columns, and we coerce them to strings, schema-inference will still end up suggesting `format: integer`, right? so we will materialize the columns as they originally were 🤔

williamhbaker:
> I don't see anything in the docs suggesting that it is not possible, so I guess it is possible. That's interesting, on another thought, since we use schema inference for mongodb, if they have integer columns, and we coerce them to strings, schema-inference will still end up suggesting `format: integer`, right? so we will materialize the columns as they originally were 🤔

What I'm thinking is that the capture connector would manipulate each document it receives from the source and make sure that whatever is in the `_id` field gets written to the collection as its string equivalent, so `1` -> `"1"`, `1.5` -> `"1.5"`, and `"uuidString"` -> `"uuidString"`. Schema inference would always recognize these as being strings since they all would be in the collection and the minimal discovered schema would always indicate the `_id` field as a `string`.

mdibaiee:
@williamhbaker yeah I get that part, but what I mean is, if I really want my `_id` to be an integer when I'm materializing my collection, would schema inference be smart in this case to suggest `type: [string, integer], format: integer` and ensure my materialization is materializing with integer type instead of string?

I think [yes](https://github.com/estuary/flow/pull/949)

mdibaiee:
Well actually now I'm not sure, if we have a minimal schema, does schema inference also modify that schema if it wants to, or does it not touch the minimal schema?

williamhbaker:
Ah I think I understand what you are saying now.

Schema inference leaves the write schema alone and only updates the read schema (hopefully I didn't switch those around), and it does whatever it wants with the read schema.

One thing is that the schema from discovery needs to at least have a field that can be used as a collection key, and I don't think that a field with multiple types would work for that.

PR #612, opened 2023-03-20
materialize-sql: add support for checking prerequisites during validation

williamhbaker:
**Description:**

Adds a "prerequisites" check to sql materializations that is invoked during `Validate` calls.

Currently there are some ad-hoc validations that occur during `NewEndpoint`. These are not comprehensive, and not including them in `Validate` means that they don't run during `Test` invocations. The comprehensive validations are more involved and moving them from `NewEndpoint` to `Validate` is also nice since they won't have to run every time the connector starts up, and are only run when a change to the materialization is being considered.

I generally tried to interpret errors for common user-error cases and return clear & helpful error messages for these cases. This does not always work as well as I'd like depending on what the client library gives us, but overall I think we will have things covered pretty well here.

There is the potential for multiple errors to be independently evaluated for some materializations. In this case we will report as many errors as possible rather than stopping on the first one. This is really only the case for redshift and bigquery which use staging buckets and have the potential for separate client configuration and cloud storage configuration errors.

Also not that there is a small tweak to the snowflake configuration validation: We will now only accept host URLs ending in `.snowflakecomputing.com`. It has never been possible for the materialization to work with anything other than `*.snowflakecomputing.com`, but now the connector will validate that specifically when parsing the configuration. This helps eliminate a lot of possible configuration errors that would have to otherwise be validated in the prereq check for the connector.

**Workflow steps:**

Use connectors like before, but have better error messages when trying to publish them.

I'll take a look at the snowflake docs and see if there is anything that needs updated there.

**Notes for reviewers:**

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -6,10 +6,12 @@ import (
 	stdsql "database/sql"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	boilerplate "github.com/estuary/connectors/materialize-boilerplate"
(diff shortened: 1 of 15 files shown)


materialize-snowflake/snowflake.go
```diff
@@ -88,12 +96,14 @@ func (c *config) Validate() error {
 			return fmt.Errorf("missing '%s'", req[0])
 		}
 	}
+
+	if !validHost(c.Host) {
+		return fmt.Errorf("invalid host %q (must match end in snowflakecomputing.com and not include a protocol)", c.Host)
```
mdibaiee, line 101: Nit: I think it's better if `validHost` itself gives this error, since that function knows what it is checking, I would expect the caller to not have to know what is happening inside the function

mdibaiee (approved):
Thanks Will! LGTM, just a tiny nit

PR #617, opened 2023-03-21
materialize-google-sheets: handle sheets out of order with bindings

williamhbaker:
**Description:**

This removes the strict requirement that sheets be returned from the google sheets API in exactly the same order as bindings are listed. This allows for the user to re-arrange sheets in their spreadsheet and have the materialization continue to work.

Closes https://github.com/estuary/connectors/issues/616

**Workflow steps:**

N/A

**Notes for reviewers:**

### DIFF
--- a/materialize-google-sheets/util.go
+++ b/materialize-google-sheets/util.go
@@ -193,18 +193,33 @@ func loadSheetStates(
 			l1, sheetNames)
 	}
 
+	// Arrange returned sheet data in the same order as configured bindings. A quadratic loop here
+	// should be fine since the number of bindings & sheets will not be very large.
+	var responseSheets []*sheets.Sheet
+	for _, bindingSheetName := range sheetNames {
+		var matched bool
+		for _, sheet := range resp.Sheets {
+			if sheet.Properties.Title == bindingSheetName {
+				responseSheets = append(responseSheets, sheet)
+				matched = true
+				break
+			}
+		}
+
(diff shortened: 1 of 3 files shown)


materialize-google-sheets/util.go
```diff
@@ -193,18 +193,33 @@ func loadSheetStates(
...
+		for _, sheet := range resp.Sheets {
+			if sheet.Properties.Title == bindingSheetName {
+				responseSheets = append(responseSheets, sheet)
+				matched = true
+				break
+			}
+		}
+
+		if !matched {
+			return nil, fmt.Errorf("sheet response missing expected sheet %q", bindingSheetName)
```
willdonnelly, line 210: The mention of "sheet response" in this error message feels like it might be a bit confusing to users. As far as they're concerned the issue is something like "sheet %q does not exist"

willdonnelly (approved):
LGTM

materialize-google-sheets/util.go
williamhbaker: In the case of the user deleting or renaming the sheet (meaning that the sheet response would be missing a binding sheet), there is actually an error triggered a little earlier that is almost certainly even worse, being along the lines of `fetching sheet states: googleapi: Error 400: Unable to parse range: 'greetings2'!A2:A, badRequest`. That would probably be worth looking into improving as a separate effort.

PR #622, opened 2023-03-22
Infer schema for source-http-ingest collections

psFried:
**Description:**

Updates the discovered binding in source-http-ingest to add the `x-infer-schema` annotation to the discovered 'webhook-data' collection. This just seemed like an obvious omission, since the json schema permits almost any valid json object.

**Workflow steps:**

Just setup a new capture using the source-http-ingest connector, and bask in the joy of automatically inferred json schemas.

@oliviamiannone is this something we should document explicitly for each connector? I'm a little uncertain. Happy to update the docs if you think we should.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/622)


mdibaiee (approved):

oliviamiannone:
@psFried I think this can go undocumented for the time being. We can see over time if it's something users get tripped up on -- in that case we can add it.

PR #635, opened 2023-03-31
source-sqlserver: Increase backfill chunk size to 32k

willdonnelly:
**Description:**

When performing backfills we alternate reading chunks from the database tables, performing a watermark write, and then waiting for that write to show up via CDC. This means that backfill throughput is capped at <ChunkSize> / <WatermarkLatency>.

This is true of all SQL captures, but in general the latency from watermark write to CDC event is milliseconds on other databases. SQL Server is special and uses a separate agent process which polls the WAL every 5s by default. This means that we can backfill at most <ChunkSize>/5 rows per second from any one table.

Let's stop the immediate pain by increasing the chunk size 8x, taking us from ~816 docs/sec to ~6.5k docs/sec.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/635)

### DIFF
--- a/source-sqlserver/main.go
+++ b/source-sqlserver/main.go
@@ -91,7 +91,7 @@ func (c *Config) SetDefaults() {
 		c.Advanced.WatermarksTable = "dbo.flow_watermarks"
 	}
 	if c.Advanced.BackfillChunkSize <= 0 {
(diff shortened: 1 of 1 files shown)


willdonnelly:
Merging without review because c'mon, it's a change to a numeric tuning constant and CI passes.

PR #639, opened 2023-04-05
source-sqlserver: Properly quote column names in backfill query

willdonnelly:
**Description:**

Per the documentation it's pretty much always safe to wrap a column name in `[brackets]` when constructing a query, so now we will.

It is possible that there are undocumented edge cases of column names which require further character-level escaping, but if such names exist the documentation doesn't mention the possibility, so this should be adequate until/unless we discover that such a name exists.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/639)


williamhbaker (approved):
LGTM

PR #640, opened 2023-04-05
source-sqlserver: Add support for datetimes as keys

willdonnelly:
**Description:**

This copies the 'TestScanKeyTypes' test to `source-sqlserver` with a test case for `datetimeoffset` columns.

Then the `encodeKeyFDB` function is modified to make this pass by serializing `time.Time` values coming from the database driver using the `sortableRFC3339Nano` format string. No decoding is required, as the string format can be used directly as an input to subsequent backfill queries.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/640)


williamhbaker (approved):
LGTM

PR #641, opened 2023-04-06
source-mongodb: log change stream error details

psFried:
Adds a bit of additional logging when encountering a server error while reading the change stream. This is to help diagnose an issue with a capture running in production.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/641)


williamhbaker (approved):
I looked at this for a few minutes earlier and noted that the log on line 157 was being emitted, implying that the error either wasn't a `mongo.ServerError` or the `changeStreamFatalErrorCode` wasn't the error code it had. I think your logging will help with the second case, but not the first. I'd guess that it probably is a `mongo.ServerError` so hopefully this will work to get us the right error code.

PR #648, opened 2023-04-10
source-postgres: remove parens from ORDER BY

psFried:
**Description:**

This removes the parens from around the column names in the `ORDER BY` of the backfill queries for source-postgres. I ran `EXPLAIN` on some backfill queries and found that, at least in this circumstance, postgres would refuse to use an index when the column names were surrounded with parens, while it would use them as expected if the parens were left off. In this case, we were querying by `(a, b)`, where `a` had a primary key index, and `b` had a secondary btree index. The parens in the `WHERE` clause did not seem to have any effect, though.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/648)


willdonnelly (approved):
LGTM

PR #650, opened 2023-04-10
source-postgres: Lower a couple of buffer-size tuning constants

willdonnelly:
**Description:**

The previous change which made only one table get a chunk backfilled at a time didn't impact memory usage nearly as much as expected, so while that's still probably a good move in general it implies that we're gobbling up memory somewhere else instead.

There are two other notable "buffers" inside of a SQL capture, so let's try reducing the size of both of those buffers and see if that makes any difference.

Both of these are sized primarily for performance -- connector operation remains correct even if the channels were unbuffered, so it should be entirely safe to make them smaller.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/650)

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -187,7 +187,7 @@ const standbyStatusInterval = 10 * time.Second
 // The buffer is much larger for PostgreSQL than for most other databases, because
 // empirically this helps us to cope with spikes of intense load which can otherwise
 // cause the database to cut us off.
-var replicationBufferSize = 64 * 1024 // Assuming change events average ~2kB then 64k * 2kB = 128MB
+var replicationBufferSize = 16 * 1024 // Assuming change events average ~2kB then 16k * 2kB = 32MB
 
 func (s *replicationStream) Events() <-chan sqlcapture.DatabaseEvent {
 	return s.events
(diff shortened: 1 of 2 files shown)


source-postgres/replication.go
```diff
@@ -187,7 +187,7 @@ const standbyStatusInterval = 10 * time.Second
 // The buffer is much larger for PostgreSQL than for most other databases, because
 // empirically this helps us to cope with spikes of intense load which can otherwise
 // cause the database to cut us off.
-var replicationBufferSize = 64 * 1024 // Assuming change events average ~2kB then 64k * 2kB = 128MB
+var replicationBufferSize = 16 * 1024 // Assuming change events average ~2kB then 16k * 2kB = 32MB
```
williamhbaker, line 190: The comment above refers to coping with load spikes which could otherwise cause the database to cut us off - I don't have any kind of feel for how big the buffer needs to be to prevent this from happening, so I can only note that making the buffer smaller feels like it would move closer to that point, whatever it may be.

Is the replication buffer used in backfills? Since we're seeing memory issues with backfills, I'm wondering if the replication buffer would have any expectation of impacting backfills.

williamhbaker (approved):

source-postgres/replication.go
willdonnelly: The load spikes I'm referring to were very hard to hit back when I was doing throughput optimization in the past, and basically require a dedicated database stress-test setup to reproduce.

The replication buffer is always in use, although its fullness will vary depending on how busy the source database is. In theory it fills up while we're performing backfill queries (because the core backfilling-and-merging behavior is necessarily synchronous) and then gets drained once we start waiting for the next watermark change.

PR #651, opened 2023-04-10
source-{mysql,postgres,sqlserver}: Ask nicely to use less RAM

willdonnelly:
**Description:**

New in Go 1.19, the `GOMEMLIMIT` environment variable is added for use-cases such as "the deployment of a web service into containers with a fixed amount of available memory".

I suspect that in at least some cases we OOM during large backfills not because we have >1GiB of *live* data, but rather because the Go GC correctly deduces that we'd get better throughput if we used >1GiB of heap before running a GC cycle. And really, if we don't inform it about our 1GiB memory limit that's a choice it is correct in making.

So, uh, maybe let's try asking it to pretty please keep the Go heap below 900MiB if it can?

**Notes for reviewers:**

In theory we might want to move this into the base connector image at some point in the future, but I figured limiting it to just SQL captures was better for the moment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/651)

### DIFF
--- a/source-mysql/Dockerfile
+++ b/source-mysql/Dockerfile
@@ -48,6 +48,10 @@ COPY --from=builder /builder/connector ./source-mysql
 # Avoid running the connector as root.
 USER nonroot:nonroot
 
+# Ask the Go runtime to keep heap size below 900MiB. Our real memory limit is
+# 1GiB so this is 10% headroom as recommended in https://go.dev/doc/gc-guide
+ENV GOMEMLIMIT=900MiB
+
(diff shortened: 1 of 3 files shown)


willdonnelly:
This probably isn't immediately necessary as I think I've identified a different cause of high memory usage. However this still feels like something we should be setting now that the knob exists, since it's meant for exactly our sort of use-case.

williamhbaker (approved):
Interesting - LGTM

psFried (approved):
LGTM

PR #655, opened 2023-04-11
source-sqlserver: Fix primary key text collation

willdonnelly:
**Description:**

This PR is intended to fix text collation ordering in backfill queries, by explicitly specifying the `Latin1_General_100_BIN2_UTF8` collation order on primary keys whose column type is `{n}{var}char(N)`.

The aforementioned collation basically means "just shut up and sort lexicographically on Unicode code-points" as far as I can tell, and since UTF-8 is designed such that bytewise lexicographic ordering matches Unicode code-point ordering this _should_ mean that backfill query row ordering will exactly match our internal scan key ordering as intended.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/655)

### DIFF
--- a/source-boilerplate/testing/testing.go
+++ b/source-boilerplate/testing/testing.go
@@ -358,6 +358,44 @@ func (v *SortedCaptureValidator) Reset() {
 	v.documents = nil
 }
 
+// OrderedCaptureValidator maintains a list of every document emitted to each output
+// collection in the order they were emitted.
+type OrderedCaptureValidator struct {
+	documents map[string][]json.RawMessage // Map from collection name to list of documents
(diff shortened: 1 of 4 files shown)


willdonnelly:
I'd still like somebody to take a look at this whenever there's time, but I'm going to go ahead and merge this since CI passes, we already do a similar thing in `source-postgres`, and this is impacting a production capture

psFried (approved):
There's some ongoing conversation about this preventing backfill queries from using indexes, but it doesn't seem blocking. LGTM

psFried (approved):
LGTM

Issue #613, opened 2023-03-21
removing materialization bindings should delete the table


mdibaiee:
- [x] materialize-postgres
- [x] materialize-snowflake
- [x] materialize-bigquery
- [x] materialize-sqlite
- [x] materialize-s3-parquet
- [ ] materialize-rockset
- [x] materialize-redshift
- [x] materialize-mongodb
- [x] materialize-google-sheets
- [x] materialize-firebolt
- [x] materialize-google-pubsub (not necessary, see comment below)
- [ ] materialize-elasticsearch

mdibaiee:
materialize-sql: https://github.com/estuary/connectors/pull/614

williamhbaker:
If it makes things easier, for `pubsub` I don't think we should actually delete topics when deleting bindings or the materialization. I think delete for `pubsub` could basically be a noop entirely and we can just leave the topics there. It kind of makes sense to not delete published messages that haven't been consumed yet, which is what deleting the topic would do.

PR #658, opened 2023-04-12
materialize-*: do not delete resources when bindings are removed

mdibaiee:
**Description:**

- Reverts changes made as part of https://github.com/estuary/connectors/issues/613
- I'm keeping the logic for persisting spec of a materialization (this was added to Firebolt and MongoDB for example) so that we can more easily support schema migrations for these connectors in the future

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/658)


tests/materialize/materialize-bigquery/snapshot.json
```diff
@@ -400,7 +400,7 @@
...
+    "actionDescription": "\nCREATE TABLE IF NOT EXISTS `estuary-sandbox`.whb_other_region.flow_materializations_v2 (\n\t\tmaterialization STRING NOT NULL,\n\t\tversion STRING NOT NULL,\n\t\tspec STRING NOT NULL\n)\nCLUSTER BY materialization;\n\nCREATE TABLE IF NOT EXISTS `estuary-sandbox`.whb_other_region.flow_checkpoints_v1 (\n\t\tmaterialization STRING NOT NULL,\n\t\tkey_begin INT64 NOT NULL,\n\t\tkey_end INT64 NOT NULL,\n\t\tfence INT64 NOT NULL,\n\t\tcheckpoint STRING NOT NULL\n)\nCLUSTER BY materialization, key_begin, key_end;\n\nUPDATE `estuary-sandbox`.whb_other_region.flow_materializations_v2 SET version = 'test', spec = '(a-base64-encoded-value)' WHERE materialization = 'tests/materialize-bigquery/materialize';"
```
williamhbaker, line 403: Not something for this PR, but seeing it in the diff reminded me that we should move to using a more stable dataset for bigquery tests than `whb_other_region` - I actually deleted that one recently (not knowing it was used for tests) so I doubt these tests will work anymore. With the new materialization testing system we should be able to run bigquery tests in CI. Previously I held off on adding them to run automatically since there was an inconsistent/long delay in how long it would take the data to be available in bigquery to to validate the materialization results, but now the tests are driven differently and it shouldn't be any issue. So if/when we set up the tests to run automatically it would make sense to establish an "official" dataset for the tests to run in I think.

williamhbaker (approved):

PR #660, opened 2023-04-13
materialize-bigquery: use direct equality for load query

williamhbaker:
**Description:**

Uses direct equality for loads, rather than "is not distinct from".

We've been observing poor load query performance on even modestly sized data sets, making the
connector essentially unusable for these cases. Some quick testing has revealed that using `=` for
comparision in the load query instead of "is not distinct from" makes a tremendous difference in the
performance of the load query join.

This change means that we will not be able to use nullable keys with the connector. For the moment
this is acceptable since it is only very recently that nullable keys were allowed in the runtime and
there are no current bigquery materializations using nullable keys. We will likely be pursuing a
different path for nullable keys involving default values in the future, so long term the direct
equality comparision should continue to work.

**Workflow steps:**

**Notes for reviewers:**

Note that using `=` for load query equality is how Snowflake does it as well: https://github.com/estuary/connectors/blob/34fc559f37853dea2a31abd75c50cf81b8b3e6e2/materialize-snowflake/sqlgen.go#L137-L158 It is possible that Snowflake would have needed to use [`is [not] distinct from`](https://docs.snowflake.com/en/sql-reference/functions/is-distinct-from) to support nullable keys (with unknown performance implications), but it had never been updated for that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/660)

### DIFF
--- a/materialize-bigquery/.snapshots/TestSQLGeneration
+++ b/materialize-bigquery/.snapshots/TestSQLGeneration
@@ -22,11 +22,11 @@ CLUSTER BY key1, key2, boolean, integer;
 SELECT 0, l.flow_document
 	FROM projectID.dataset.target_table AS l
 	JOIN flow_temp_table_0 AS r
-		 ON l.key1 is not distinct from r.key1
-		 AND l.key2 is not distinct from r.key2
-		 AND l.boolean is not distinct from r.boolean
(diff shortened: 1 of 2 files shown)


willdonnelly (approved):
LGTM

PR #661, opened 2023-04-14
materialize-postgres: be more explicit about the default port

williamhbaker:
**Description:**

If no port is specified with the host, the connector will currently use 5432. This is buried in the [inner workings of pgx](https://github.com/jackc/pgconn/blob/1e1135688ea919f66b9425b5bfe06f14ef1e5285/defaults.go#L16) and not made clear to the user. This changes the messaging that the user sees and makes setting the port explicit in our connector code so that it is stable going forward.

**Workflow steps:**

Use a postgres materialization as before with no port specified and observe that it still works. No behavior is actually changed with this PR.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/661)

### DIFF
--- a/materialize-postgres/.snapshots/TestSpecification
+++ b/materialize-postgres/.snapshots/TestSpecification
@@ -6,7 +6,7 @@
       "address": {
         "type": "string",
         "title": "Address",
-        "description": "Host and port of the database.",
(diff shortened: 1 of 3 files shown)


willdonnelly (approved):
LGTM

PR #663, opened 2023-04-14
source-s3: use localstack for integration tests

williamhbaker:
**Description:**

Updates the `source-s3` integration tests to use localstack instead of an actual S3 bucket so they can run reliably.

Also adjusts some test output ordering for `source-s3` and `source-gcs`. Apparently the new sqlite materialization has a slightly different idea of what order these are output in.

**Workflow steps:**

N/A

**Notes for reviewers:**

There was a small adjustment needed to the `source-s3` connector to get non-AWS endpoints to work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/663)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -137,7 +137,7 @@ jobs:
           export_default_credentials: true
 
       - name: Configure AWS credentials from Test account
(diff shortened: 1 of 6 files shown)


mdibaiee (approved):
awesome, thank you!

Issue #348, opened 2022-09-21
source-mysql: Implement limited support for `ALTER TABLE` queries on active tables


willdonnelly:
Currently any use of `ALTER TABLE` on a MySQL table which is actively being captured is an error, and the recommended fix is to remove and then recreate the capture binding. This is because MySQL binlog entries don't include sufficient metadata about column names, so we have to keep track of that information in the state checkpoints after initializing it from discovery results.

We already parse DDL query events which appear in the binlog in order to fail "cleanly" when such a thing happens, so in theory it shouldn't be a ton of work to add support for common cases like "adding a new column to the end of the table".

PR #667, opened 2023-04-18
source-mysql: add limited support for alter table queries on active tables

williamhbaker:
**Description:**

Adds handling for adding columns at the beginning, end, or in the middle of active tables and also dropping columns from active tables.

Closes https://github.com/estuary/connectors/issues/348

**Workflow steps:**

Start a mysql capture and perform `alter table add <name> <type> [first|after <other>]` or `alter table drop column <name>`. The capture will handle these operations and update its state to reflect the addition/removal of the new column and going forward will output captured documents accordingly.

The docs at https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/#unsupported-operations could use updated with this change. Although still technically correct, it would probably be good to state that adding or removing columns is now supported, but other things like changing the type of a column are not.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/667)


willdonnelly (approved):
This looks great to me.

One thing that I noticed while reviewing but which probably _shouldn't_ be addressed in this PR is that the `findStr` function uses strict equality but depending on how much MySQL normalizes the logged query events it may be more correct to use case-insensitive comparison. But then I realized that our backfill query generation for MySQL still doesn't quote column names and our tests don't exercise capitalization behavior anyway, so while that should all be addressed at some point it's probably out of scope for this PR right here.

Issue #537, opened 2023-02-10
source-mysql: Create watermarks table on startup


willdonnelly:
Unlike the Postgres and SQL Server connectors, the MySQL connector doesn't autocreate the watermarks table on startup. The only principled reason for this is that the default watermarks table name is `flow.watermarks` and thus we'll generally need to create a `flow` schema to hold the table, but we've previously discussed doing that and it didn't seem like a problem.

So we should bring MySQL into line with the other SQL captures and automagically create the table for the user if it's not present.

Issue #513, opened 2023-01-18
source-mysql: Flesh out handling of zero-valued `DATETIME` and `TIMESTAMP` columns


willdonnelly:
We previously merged https://github.com/estuary/connectors/pull/512 as a quick fix for the immediate pain of zero-valued `DATETIME`s. This should be improved by implementing the same special-casing for `TIMESTAMP` columns and adding test cases covering both behaviors.

PR #668, opened 2023-04-19
source-mysql: a few fixes/enhancements

williamhbaker:
**Description:**

This PR includes a few fixes/enhancements to `source-mysql`.

#### 1) Attempt to create watermarks database if it does not exist
Closes https://github.com/estuary/connectors/issues/537

With this change we'll try to create the database for the watermarks table if it doesn't exist prior to creating the watermarks table itself if it doesn't exist. This will work for cases where the user configured for the capture has the ability to create databases, which is probably most common for testing setups running with an admin account or similar.

If the database and watermarks table is able to be created the watermarks part of the setup can be completed and the capture run with no other prior database actions.

#### 2) Fail validation when capturing datetime columns if time_zone is not set
Closes https://github.com/estuary/connectors/issues/538

Currently, if the timezone can't be determined from the system `'time_zone'` variable a warning message is logged and captures fail at runtime if a table contains a datetime column.

This adds logic to Validate so that if a timezone can't be determined and any of the tables have datetime columns the capture will fail to publish and the relevant table & columns are returned to the user in an error message.

#### 3) Handle "zero" value timestamps
Closes https://github.com/estuary/connectors/issues/513

[Zero value datetimes were already handled](https://github.com/estuary/connectors/pull/512), so this adds handling for zero value timestamps and a test for each case.

**Workflow steps:**

If you try to create a capture for a table with datetime columns in a database without the `'time_zone'` system variable set (or set in a way that the connector doesn't understand) an error will be returned like this during validation:

```
the capture cannot run due to the following error(s):
     - system variable 'time_zone' must be set to capture datetime columns [my_date, my_date_2] from table "database.table"
```

N/A - Nothing I can see in the docs for `source-mysql` would be relevant to update for these changes.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/668)

### DIFF
--- a/source-mysql/datatype_test.go
+++ b/source-mysql/datatype_test.go
@@ -19,6 +19,10 @@ func TestDatatypes(t *testing.T) {
 	tb.Query(ctx, t, "SET GLOBAL time_zone = 'America/Chicago';")
 	tb.Query(ctx, t, "SET SESSION time_zone = 'America/Chicago';")
 
+	// For testing inputs of "zero" value datetime and timestamp 'NO_ZERO_DATE' must be disabled for
+	// the "zero" values to be inserted into the test database tables.
+	tb.Query(ctx, t, "SET SESSION sql_mode=(SELECT REPLACE(@@sql_mode,'NO_ZERO_DATE',''));")
+
 	tests.TestDatatypes(ctx, t, tb, []tests.DatatypeTestCase{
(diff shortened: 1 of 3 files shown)


willdonnelly (approved):
LGTM

Issue #634, opened 2023-03-30
Update SQL Captures to name collections with schema.table name


dyaffe:
**Problem**
Currently SQL captures name discovered collections with just table names.  This can result in a conflict if two schemas contain the same table name.

**Proposal**
Update discover to output collections with `schema.table` as the postfix of the name

jgraettinger:
When doing this, we should also update the resource config schema to use `schema` instead of `namespace`, which is a hold-over from its Airbyte protocol days.

It will still need to accept `namespace` from deployed capture specs, though.

Issue #631, opened 2023-03-24
CDC connectors should filter well-known flow tables from discovery


dyaffe:
**Problem**
It's relatively common to capture and materialize to the same DB for testing purposes.  When this happens, we capture the three Flow-created tables (watermarks, publications and checkpoints).

It would be nice if we filtered well-known Flow created tables from discovery

willdonnelly:
The materialization tables will need additional logic to be filtered out, but the watermarks table should already be excluded from discovery results. Can you confirm whether the watermarks table was discovered in the test you did earlier?

PR #669, opened 2023-04-19
sqlcapture: discovery improvements

williamhbaker:
**Description:**

There's a couple of changes here which I'm lumping together as "discovery" improvements. There's only a few lines of actual code changing here, and then lots of test output snapshots that result from it.

#### 1) Recommend fully-qualified names for discovered collections
Closes https://github.com/estuary/connectors/issues/634

This change results in recommend discovered collection names to be output like `<schema>.<table>` instead of just `<table>`. This is very helpful in the case of a database containing tables with the same name in different schemas. Currently the catalog recommended by discover for this scenario would cannot be published since there are duplicate collection names. In the fully qualified form there aren't duplicate collection names and the discovery from such a database will work out of the box.

I tested this locally with the 3 sql capture connectors to see what happens when you have previously discovered a set of tables from old connector logic and then re-discover those same tables and/or new tables via the UI. It works well enough: The old tables keep their same collection names, but the new tables are in the form of `<schema>.<table>`. Most importantly the old collections are _not_ removed & replaced with the new collection names. A brand new capture of course will have all the recommended collection names in the fully qualified form.

#### 2) Filter materialization metadata tables from discovery
Closes https://github.com/estuary/connectors/issues/631

We were already filtering the specific `flow_watermarks` table out of discovery for the capture. This extends the filtering to prevent `flow_materializations_v2` and `flow_checkpoints_v1` tables  which are created materializations. They will be filtered out from any schema, since materialization configurations allow for these tables to exist in a user-specified schema.

Currently we only have a materialization connector for postgres, so practically speaking this could only apply for a postgres capture.

It might be worth noting that if a previous capture had discovered these tables, a new discovery will remove them. I don't believe this is an issue with any current postgres captures.

**Workflow steps:**

N/A

**Notes for reviewers:**

In addition to the two changes mentioned above, there's a small tweak to the resource config schema that was requested in https://github.com/estuary/connectors/issues/634

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/669)


willdonnelly (approved):
LGTM

Dumb question: Are periods an allowed character in Flow catalog names or do they get percent-encoded?

williamhbaker:
> LGTM
> 
> Dumb question: Are periods an allowed character in Flow catalog names or do they get percent-encoded?

Periods are allowed and don't get percent-encoded. The only possible weirdness might be materializing from such a collection if the materialization doesn't do well with periods in the destination table names, but all of our current materialization connectors handle it fine and the expectation for materialization connectors is that they are able to somehow deal with all of the characters that could be in a collection name - usually by quoting the identifier, or maybe by returning an appropriate constraint.

Issue #620, opened 2023-03-22
sqlcapture: Correctly handle slashes in the `KeyPtrs` discovery output and collection key decoding


willdonnelly:
Currently our SQL capture connectors just slap a `/` in front of the database column name when filling out the `KeyPtrs` properties of a discovery response. This is blatantly incorrect for column names containing a `/` or `~` character, which should be escaped to `~0` and `~1` respectively.

Likewise the new "use collection key as the primary key for backfills" logic needs to correctly decode a `~0` or `~1` escape code in addition to stripping the leading slash, and should probably also add a check that the only unescaped slash is the first character of the column name.

Issue #353, opened 2022-09-22
Postgres issue with tables with special characters in name


travjenkins:
Running into an issue with the Postgres connector. This is a super aggressive edge but wanted to make sure folks knew about it. 

I have this insane table name:
![image](https://user-images.githubusercontent.com/270078/191829716-74c65436-f704-40ab-9c23-b0c64aee5d92.png)

And when I attempt to get this running as a capture I get the following logs:
```
runTransactions: readMessage: exit status 1 with stderr:

time="2022-09-22T18:52:00Z" level=info msg="initializing connector" address="localhost:6543" database=testing_db slot=flow_slot user=testing
time="2022-09-22T18:52:00Z" level=info msg="discovering tables"
time="2022-09-22T18:52:00Z" level=info msg="starting replication" publication=flow_publication slot=flow_slot startLSN=0/3865E178
time="2022-09-22T18:52:00Z" level=info msg="streaming to watermark" watermark=604fe1e7-038c-4dd4-9abe-bf8e90a999b0
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream="public.\\_%*|-&\\_|*%-\\&_|%*\\-&_|*%\\-_&|\\*%_-|&*%\\_-|&\\%_*-|&\\_%|*-&"
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.anothernewtqablenamehere
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.anothernewtqablenameherehereherehere
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.anothernewtqablenamehereherehereherethisisanotherchange
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.bar_foo
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.flow_checkpoints_v1
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.flow_materializations_v2
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.foo_bar
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.newtablenamehere
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.one
time="2022-09-22T18:52:00Z" level=info msg="activating replication for stream" stream=public.three
[...]

PR #672, opened 2023-04-21
sqlcapture: enhanced handling for unusual identifiers

williamhbaker:
**Description:**

This set of changes improves handling for unusually named tables and primary key columns.

The cross-cutting change to `sqlcapture` for unusual table names is related to recommended Flow collection names on discovery. Flow collection names are allowed only a specific set of characters and have certain requirements for "path-like" sequences like `/../` due to their inclusion in Gazette journal paths. A database like Postgres can have just about any character in a table name (if it is quoted) which can lead to invalid recommended collection names if these table names are used as-is in collection names. The change here applies a simple sanitization where the disallowed characters and any occurrences of a `/` character are replaced with `_`. This closes https://github.com/estuary/connectors/issues/353: `source-postgres` had previously been updated to always quote table names when constructing queries, and the discovery output for strangely named tables will now be acceptable as well.

Another current issue is that primary key columns containing `~` and/or `/` characters are discovered and recommended as collection keys as-is with a `/` prepended to them to turn them into a JSON pointer. But JSON pointers are a hierarchical reference to a location within a JSON document and any occurrences of `~` or `/` in the pointer have special meaning in that sense. The change to `sqlcapture` here handles escaping these characters per RFC6901 so that they can be used as valid collection keys. This closes https://github.com/estuary/connectors/issues/620.

A couple of specific changes were made along with these to `source-mysql`. This connector will now quote key column names and table identifiers when constructing queries similar to `source-postgres`. This allows `source-mysql` to capture from databases with key columns or tables named with [characters that require quoting](https://dev.mysql.com/doc/refman/8.0/en/identifiers.html), which includes any of the kinds of names this PR is addressing. 

**Workflow steps:**

Discover & capture on databases containing identifiers with special characters.

N/A

**Notes for reviewers:**

Much of the impact from this PR involves interacting with the actual flow runtime and its naming requirements, so along with the tests included here I did a fair amount of manual testing with SQL captures, particularly with `source-mysql` which did not previously do much in the way of identifier quoting.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/672)

### DIFF
--- a/source-mysql/.snapshots/TestAlterTable_AddColumnBasic-at_middle
+++ b/source-mysql/.snapshots/TestAlterTable_AddColumnBasic-at_middle
@@ -1,10 +1,10 @@
 # ================================
 # Collection "acmeCo/test/test.altertable_addcolumnbasic_aaa": 2 Documents
 # ================================
-{"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"AlterTable_AddColumnBasic_aaa"}},"data":"mmm","extra_end":"extra_end_11","extra_middle":"extra_middle_1","extra_start":"extra_start_11","id":11}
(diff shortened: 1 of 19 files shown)


willdonnelly (approved):
LGTM

PR #673, opened 2023-04-24
Phil/elasticsearch mappings

psFried:
**Description:**

This changes how Elasticsearch indices are created by the materialize-elasticsearch connector. The connector no longer creates explicit `mappings`, and instead configures the index to
use dynamic [runtime mappings](https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime.html). Eager creation of mappings for every field can unnecessarily bloat the
indices and create performance problems. Using dynamic runtime mappings allows users to still query based on any field in their document, and still allows them to create explicit mappings as the need arises.

This also simplifies the resource configuration, which no longer includes field mappings.

In addition, support was added for authentication using API keys. This was just an opportunistic improvement since the first commit was already a breaking change.

**Workflow steps:**

Setup an elasticsearch materialization, and you'll see that any new indexes will get created using dynamic `runtime` mappings rather than explicit mappings.

[The materialize-elasticsearch docs](https://docs.estuary.dev/reference/Connectors/materialization-connectors/Elasticsearch/) are of course affected, and I'm working on updating them now.

**Notes for reviewers:**

The config schema generation seems kind of unfortunate, but I couldn't figure out how to make the schema generator output a `oneOf` that works with our UI. Is there a better way to do that than just writing the entire schema manually?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/673)


materialize-elasticsearch/driver.go
```diff
@@ -96,27 +170,27 @@ func (driver) Validate(ctx context.Context, req *pm.Request_Validate) (*pm.Respo
...
-			binding.Collection.GetReadSchemaJson(),
-			res.FieldOverides,
-		); err != nil {
-			return nil, fmt.Errorf("building elastic search schema: %w", err)
-		} else if err = elasticSearch.CreateIndex(res.Index, res.NumOfShards, res.NumOfReplicas, schema, true); err != nil {
+		if desc, err := elasticSearch.ApplyIndex(res.Index, res.NumOfShards, res.NumOfReplicas, true); err != nil {
 			// Dry run the index creation to make sure the specifications of the index are consistent with the existing one, if any.
 			return nil, fmt.Errorf("validate elastic search index: %w", err)
+		} else {
+			log.WithFields(log.Fields{
```
williamhbaker, line 190: I'm not 100% sure what to expect from this log message, but if it seemed useful to you to keep here then I'm good with that.

materialize-elasticsearch/elasticsearch.go
```diff
@@ -6,37 +6,39 @@ import (
...
 
-	// TODO(jixiang): test the driver with es version v8, and AWS OpenSearch in addition to Elastic cloud.
-	//                extend the API as needed.
 	elasticsearch "github.com/elastic/go-elasticsearch/v7"
 	"github.com/elastic/go-elasticsearch/v7/esapi"
 	"github.com/elastic/go-elasticsearch/v7/esutil"
 	log "github.com/sirupsen/logrus"
 )
 
+// tOken: Qzl1S3BZY0I4T3RSRm93NE1lWTA6dnlxOFByNmJSdU94dWMyam9qSmx4QQ==
```
williamhbaker, line 19: Left over comment?

materialize-elasticsearch/elasticsearch.go
```diff
@@ -60,92 +62,85 @@ func (es *ElasticSearch) DeleteIndices(indices []string) error {
...
 // if the index does not exist. Otherwise,
 //  1. if the new index has a different mapping (or num_of_shards spec) from the existing index,
 //     the API stops with an error, because the mapping and num_of_shards cannot be changed after creation.
 //  2. if the new index has a different num_of_replica spec from the existing index,
 //     the API resets the setting to match the new.
-func (es *ElasticSearch) CreateIndex(index string, numOfShards int, numOfReplicas int, schemaJSON json.RawMessage, dryRun bool) error {
-	var schema = make(map[string]interface{})
-	var err = json.Unmarshal(schemaJSON, &schema)
+func (es *ElasticSearch) ApplyIndex(index string, numOfShards int, numOfReplicas int, dryRun bool) (string, error) {
+	// var numOfShardsStr = strconv.Itoa(numOfShards)
```
williamhbaker, line 92: Left over comments?

williamhbaker (approved):
LGTM aside from a few left over comments to clean up

williamhbaker:
Re: Config schema generation, I haven't figured out a better way for generating the outputs from JSON annotations.

I've done it a little more of a "meta" level for ex [here](https://github.com/estuary/connectors/blob/93b4b5ceede8a83a9ccd17334471152961df61d1/go-auth/google/config.go#L85-L157) but it's still manual. Not saying its impossible to do otherwise, but if it is I haven't figured it out yet.

materialize-elasticsearch/driver.go
psFried: yeah, I wanted to do something with the action description that's output from the `ApplyIndex` during validation. Many of the error messages in this connector lack details about which binding they relate to, and this just seemed like an easy way to identify that a particular binding was validated without issue.

Issue #544, opened 2023-02-14
source-postgres: Improve captures of user-defined types


willdonnelly:
- Actually capturing user-defined domain and enum types pretty much works today.
- Capturing user-defined tuple and range types needs better value translation, currently it's just stringifying the Postgres text representation of the data.
- Discovery needs to produce vaguely reasonable JSON schemas for domain/enum/tuple/range typed columns.

williamhbaker:
I'm going to close this issue out since https://github.com/estuary/connectors/pull/675 has been merged which improved implemented discovery for enums and ranges, and domain types were already being discovered & captured.

That PR did not address discovery or capture of composite (tuple) types, which would take quite a bit more effort. I opened a separate issue specifically for that (https://github.com/estuary/connectors/issues/677) which includes details about what I found when I was exploring what it might take. For the moment it is not believed these types are very common and thus lower priority to improve handling for.

PR #675, opened 2023-04-25
source-postgres: discover enum & range types as strings

williamhbaker:
**Description:**

This changes discovery for range (user-defined and built-in) and enum (which are always user-defined) columns to output a JSON schema with `string` fields for these columns. They will continue to be output in their text form.

Postgres enum type values are represented as string labels. There is an underlying integer value corresponding to the string label values, but that seems less useful to capture. The enum type in Postgres restricts inputs to just its set of labels, and JSON schema supports a similar concept with the `enum` keyword. However I am opting to _not_ attempt to cast the Postgres enum values into a corresponding JSON schema enum. This is mostly because it would make dealing with additions to the Postgres enum difficult: We would need a mechanism for re-discovering and updating the collection write schema if a new value were added to an allowable Postgres enum. In the future a system may exist where schema violations trigger a re-discovery & publication for the target collection (we've been talking about building this recently, even). It would be possible at that point to apply the same restrictions to the collection JSON schema enum as the Postgres enum if desired, although I'm not sure how useful it would be.

Ranges are a little less straight-forward, but I can't see a better representation for them than Postgres' text form. As an example, a column that accepts ranges of integers can be of type `int4range` and a value like `[1,5)` can be inserted into it. This value is "integers greater than or equal to 1 but less than 5" - it includes information about the upper and lower bounds being inclusive/exclusive. JSON schema does have keywords for `minimum`, `exclusiveMinimum` etc. for numeric types but this does not conceptually map to Postgres ranges where it is the values that describe the range rather than the column type, and Postgres has more range types than just numeric values.

Absent here is any modified handling for user-defined composite or "tuple" types. The text form of these is actually pretty bad and looks something like `"('strVal', 1, false)"`. In principal it would be possible to handle these better. `pgx v5` includes methods for loading information for custom types and registering it with the connection's type registry to allow composite types to be loaded as `map[string]interface{}`, where the string key is the name of the composite type "field" and the value is the value. See https://stackoverflow.com/a/75713614 for an example of this. Conceptually, discovery could find user-defined composite types and register them, and replication could watch for type messages and register new types if they appear. It's not obvious to me that implementing this would be worth the effort at the moment so I'm leaving that as a future TODO if the need arises.

Part of https://github.com/estuary/connectors/issues/544

**Workflow steps:**

Discovery on tables with `enum` or `range` types will produce collection schemas with `string` types. Discovery for composite types will continue to produce a schema with no type and a warning message in the description.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/675)

### DIFF
--- a/source-postgres/.snapshots/TestUserTypes-Range-Capture
+++ b/source-postgres/.snapshots/TestUserTypes-Range-Capture
@@ -1,0 +1,10 @@
+# ================================
+# Collection "acmeCo/test/test.usertypes_range_morsel": 2 Documents
+# ================================
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"usertypes_range_morsel","loc":[11111111,11111111,11111111]}},"id":1,"value":"(1,2]"}
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"usertypes_range_morsel","loc":[11111111,11111111,11111111]}},"id":2,"value":"[3,)"}
+# ================================
+# Final State Checkpoint
+# ================================
+{"cursor":"0/1111111","streams":{"test.usertypes_range_morsel":{"key_columns":["id"],"mode":"Active"}}}
+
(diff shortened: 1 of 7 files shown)


willdonnelly (approved):
LGTM

PR #676, opened 2023-04-25
materialize-bigquery: use JSON type for objects

psFried:
**Description:**\

Updates the materialize-bigquery connector to use the `JSON` column type instead of `STRING` for JSON objects. Most notably, this changes the type that's used for the `flow_document` column. The goal is to make it easier to query these columns.

**Workflow steps:**

No change.

**Notes for reviewers:**

This is just a draft at the moment, so I can test compatibility.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/676)


williamhbaker:
@psFried any reason to keep this open? I've given more thought to how we could accomplish this and it would probably involve building on the changes from https://github.com/estuary/connectors/pull/859. But, likely not a high priority for the foreseeable future.

PR #679, opened 2023-04-26
sqlcapture: add checks for minimum supported database versions

williamhbaker:
**Description:**

Adds a check to `SetupPrerequisites` to verify that the captured database is of the minimum supported version for `source-postgres`, `source-mysql`, and `source-sqlserver`.

This check is implemented in a conservative way, so that only if the database version can conclusively be determined NOT to meet the minimum required we error out of the Validate RPC. Failure to determine the version of the source database will not result in an error and will instead log that the database version couldn't be determined. It's setup this way to account for unusual systems that may work fine to capture from but do not support the standard version queries. I don't know of any systems like this, and my spot-checking of various cloud provided postgres/mysql/sqlserver options didn't turn up anything unusual, but it seems better to give potentially "unknown" versions a chance to run rather than only allow those that comply with the standard version query.

A big chunk of the work for this was testing various versions of the databases in question with our connector. The results are summarized below:

- **Postgres** works with version 10 and up. Our test suite is really flaky on versions 10 and 11 for some reason but everything seems to work fine
- **MySQL** works with version 8. I could not get 5.7 to work at all, see issue: https://github.com/estuary/connectors/issues/682
- **MariaDB** works with version 10.3 which is the lowest version I tried.
- **SQL Server** "works" with SQL Server 2017, however there is an issue with collation for various field types, see issue https://github.com/estuary/connectors/issues/681

Based on the above I set the minimum versions for Postgres to `10.0`, SQL Server to `14.0` (which corresponds to SQL Server 2017), MySQL to `8.0`, and MariaDB to `10.3`. There's a TODO about MySQL `5.7`, and we may be able to relax that if that issue is resolved.

Another thing to mention is that MySQL and MariaDB use the same connector, so we have to rely on the string `MariaDB` being present in the version response to know which is which. The strategy here is to assume that it's MySQL _unless_ the `MariaDB` string is present in the version response (which is true as far as I can tell), and by extension use the "lower" required version of MySQL unless we have that conclusive evidence that the higher required version number for MariaDB should be used. So if a capture is running on a MariaDB instance that does not say it is MariaDB, its version will be compared to the `8.0` required of MySQL (instead of `10.3`), which will err on the side of passing the validation check.

Closes https://github.com/estuary/connectors/issues/399

**Workflow steps:**

Use connectors on very old databases and get an error saying that they don't meet the minimum required version, instead of the connector failing in other less comprehensible ways.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/679)


willdonnelly (approved):
LGTM, although I'm not a huge fan of hard-coding the version numbers in at the point of use and might prefer to define a named constant or variable for those.

PR #683, opened 2023-04-27
source-postgres: Relay Flow acknowledgements to database

willdonnelly:
**Description:**

This has been an outstanding TODO item ever since the first version of `source-postgres`, but over time we've slowly worked our way to a point where it's not too difficult.

This change modifies the generic SQL Capture logic to request checkpoint-acknowledgement messages from Flow, to buffer emitted checkpoint cursors internally until they're acknowledged, and to then relay that acknowledgement to the database-specific logic via the `(ReplicationStream).Acknowledge(cursor)` method which was implemented well over a year ago but never actually used.

In addition, the Postgres test `TestSlotLSNAdvances` has been rewritten to simulate a more realistic scenario, in which there is a constant stream of changes to the database and a single ongoing capture, and we're happy so long as the LSN eventually changes during the normal operation of that capture. The new test logic reliably passes after these changes, but fails just as reliably if the Flow->DB acknowledgement plumbing is broken at any point, so I believe that both the plumbing and the test are working as they should.

**Workflow steps:**

The amount of WAL data retained by the Flow replication slot on Postgres should be greatly reduced, as the slot LSN will consistently advance within tens of seconds rather than after the next connector restart.

None

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/683)


sqlcapture/capture.go
```diff
@@ -628,6 +650,45 @@ func (c *Capture) emitMessage(out *boilerplate.PullOutput, msg interface{}) erro
...
+	defer c.pending.Unlock()
+	if count == 0 {
+		return nil // Zero count isn't permitted in the protocol, but has a sensible meaning of 'ignore it' so let's do that
+	}
+	if count > len(c.pending.cursors) {
+		return fmt.Errorf("invalid acknowledgement count %d, only %d pending checkpoints", count, len(c.pending.cursors))
+	}
+
+	var cursor = c.pending.cursors[count-1]
+	c.pending.cursors = c.pending.cursors[count:]
```
williamhbaker, line 686: Does this allow the memory for everything up until `count` to be freed from the backing array? I'm thinking since they are strings and not pointers...maybe? But I've never been 100% sure about. this myself.

williamhbaker (approved):
LGTM

sqlcapture/capture.go
willdonnelly: TL;DR: the old cursor values will eventually get freed

Longer answer with rambling about garbage collection and runtime internals:

Not immediately (because the slice is essentially just a 3-tuple of `(BackingArrayPtr, Length, Capacity)` and `[n:]` means `(BackingArrayPtr+n, Length-n, Capacity-n)` and AFAIK none of the big production-quality garbage collectors do anything but retain a whole allocation when there's an interior pointer). But later on when we `append()` to the resliced array we will eventually run out of capacity and it'll get reallocated and the retained elements copied, and at that point the entire old array is freed.

So this probably does retain a little bit of extra memory, but a replication cursor looks like `"0/E8E2DF0"` so we're talking about kilobytes at most so I'm not inclined to write a more complicated piece of logic here which I'm more likely to get wrong.

williamhbaker: I see! Thanks for the explanation.

PR #685, opened 2023-04-27
source-postgres: Add DEBUG logs for a production issue

willdonnelly:
**Description:**

In automated tests the new LSN acknowledgement code seems to work, but in production our capture from our own Supabase instance is not advancing the replication slot LSN. We need more visibility here to narrow down the problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/685)


willdonnelly:
YOLO merging because I think a pair of `DEBUG` logging calls is a pretty uncontroversial change

PR #686, opened 2023-04-27
sqlcapture: Treat Acknowledge with empty count as count=1

willdonnelly:
**Description:**

This was the original meaning of the message and there's at least one place within Flow that hasn't been updated to set the count yet, so we should continue to treat it as a count of 1 if unset.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/686)


jgraettinger (approved):
LGTM

Issue #684, opened 2023-04-27
Discovered Text Array in the Postgres Capture Fails with Nullable Text Arrays


dyaffe:
**Problem**
I just created a postgres capture from our `live_specs` table and it happens to have a null nullable text array.  This failed because we discovered it as non-nullable.

cc @williamhbaker

PR #688, opened 2023-04-28
source-postgres: handle nullable columns with array types

williamhbaker:
**Description:**

A captured table column may be an array type and be nullable, in which case the column might contain arrays or `null`. We would previously discover these columns as `{"type":["object"]}` which would disallow any `null` values to be captured from the table. We were also applying the column nullability to the items in the array. But it is not possible for a column array to be declared with non-nullable items in Postgres in a way that we can currently check for, as it requires additional constraints on the table.

This changes discovers for nullable array columns to have a type like `{"type":["object","null"]}` which will allow nullable array columns to be captured when the columns contain `null`.

Closes https://github.com/estuary/connectors/issues/684

**Workflow steps:**

Capture from a database with tables with nullable array columns and see that it now works when they contain `null` values.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/688)

### DIFF
--- a/source-postgres/datatype_test.go
+++ b/source-postgres/datatype_test.go
@@ -10,7 +10,8 @@ import (
 	"github.com/estuary/connectors/sqlcapture/tests"
 )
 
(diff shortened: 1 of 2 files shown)


willdonnelly (approved):
LGTM

Issue #460, opened 2022-12-12
source-postgres: reconsider `FOR ALL TABLES` publication group


jgraettinger:
Currently our connector will try to be helpful and automatically create a publication `FOR ALL TABLES` at startup (expecting this to fail if the named publication already exists). This makes the publication forward-compatible with future tables which may be bound by the capture.

However, this can cause problems for `UPDATE` or `DELETE` of tables having no primary key:
```
cannot delete from table "foobar" because it does not have a replica identity and publishes deletes
```

This is problematic because creating a CDC of a live database has the potential to impact its existing usages.
Ideally the connector would be more selective over the tables included in its publication group.

There is a work-around, which is to set `REPLICA IDENTITY FULL`, as in:
```
# alter table foobar replica identity full;
```

willdonnelly:
Off the top of my head I see two paths for improving this:

1. Remove the `CREATE PUBLICATION` startup behavior entirely, require that the user set it up along with creating the replication user and whatnot, and mention in the docs that they may want to specify just the set of tables they intend to capture. This is similar to how we currently handle table read permissions, for instance.
2. Replace the current `CREATE PUBLICATION <name> FOR ALL TABLES` startup query with `CREATE PUBLICATION <name>` to make sure a default-empty publication exists and then `ALTER PUBLICATION <name> ADD TABLE <foo>, <bar>, <etc>` to add all currently-bound tables. Since it's `ADD TABLE` rather than `SET` the publication will grow to cover all tables that were bound to the capture at any point even if they're later removed, but this seems better than potentially having two captures clobber each other (if a user for some reason configures two captures with distinct replication slots but the same publication name).

jgraettinger:
👍 to option 2) from me, seems like a sound approach.

PR #690, opened 2023-04-28
source-postgres: don't create FOR ALL TABLES publication

williamhbaker:
**Description:**

Up until now `source-postgres` has tried to create a "FOR ALL TABLES" publication if the configured publication did not exist. Creating this publication can cause problems for non-captured tables if those tables do not have a primary key and also do not have a replica identity. In this case the tables can no longer process updates and deletes, since they are included in the FOR ALL TABLES publication and these actions require a table that is part of a publication to have a replica identity. In short, we do not want our capture to have any side effects on non-captured tables.

With this change we'll instead create an "empty" publication at first if one doesn't already exist and ensure all required tables are part of it (including both captured tables and the watermarks table), adding them if they aren't part of it if possible and producing an error if any table is not part of the publication and can't be added to it.

Closes https://github.com/estuary/connectors/issues/460

**Workflow steps:**

Create a capture on a database with non-capture tables. These non-captured tables won't be added to the publication group. Only tables that are part of the capture are added, including any new tables that are later added to the capture.

We should update the `source-postgres` docs. It currently has guidance to create a "FOR ALL TABLES" publication and should instead say to create a publication only for the captured tables if setting up the publication manually.

**Notes for reviewers:**

I don't like adding even more queries that have to run for every bound table every time the connector starts. But we were already doing a `SELECT * LIMIT 0 ...` to verify that we could read from every table so I don't think this change is going to make a huge difference on startup time.

Also note one other potential gotcha that didn't turn out to be that bad: backwards compatibility. Some existing connectors will have had the "FOR ALL TABLES" publication created. This still works just fine though since the connector will see that the publication exists and not try to re-create it, and also the tables will show up in the `SELECT COUNT(*) FROM pg_catalog.pg_publication_tables` just as if they were added explicitly. I did some testing with migrating captures made from the current `:dev` connector to this one and everything worked as expected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/690)

### DIFF
--- a/source-postgres/prerequisites.go
+++ b/source-postgres/prerequisites.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/estuary/connectors/sqlcapture"
 	"github.com/sirupsen/logrus"
@@ -17,6 +18,7 @@ func (db *postgresDatabase) SetupPrerequisites(ctx context.Context) []error {
 		db.prerequisiteReplicationSlot,
 		db.prerequisitePublication,
 		db.prerequisiteWatermarksTable,
+		db.prerequisiteWatermarksInPublication,
 	} {
 		if err := prereq(ctx); err != nil {
 			errs = append(errs, err)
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

PR #694, opened 2023-05-02
source-postgres: add table to publication error message update

williamhbaker:
**Description:**

This changes the error message in the case that a captured table is not part of the configured publication and can't be automatically added to it. For now it's consistent with other prerequisite check errors and carries with it an implication that the user will have to do something manually since the connector couldn't do it automatically.

**Workflow steps:**

N/A

**Notes for reviewers:**


willdonnelly (approved):
LGTM

PR #697, opened 2023-05-03
MongoDB: assortment of issues

mdibaiee:
**Description:**

- Views cannot be used with `ChangeStream`, so we should exclude them from discovery process
- When sanitizing documents, we were not matching against `primitive.M` for nested objects (although `primitive.M` is a type alias for `map[string]interface{}` but it requires matching explicitly), so nested objects were not being sanitized, leading to `NaN` errors
- `ChangeStream` errors were not being surfaced properly, which lead to confusion when a `ChangeStream` would fail

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/697)

### DIFF
--- a/source-mongodb/pull.go
+++ b/source-mongodb/pull.go
@@ -54,7 +54,7 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
 		Output: stream,
 	}
 
-	eg, ctx := errgroup.WithContext(ctx)
+	eg, egCtx := errgroup.WithContext(ctx)
 
 	if err := c.Output.Ready(false); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


source-mongodb/pull.go
```diff
@@ -77,7 +77,13 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
 			}
 		}
 
-		eg.Go(func() error { return c.ChangeStream(ctx, client, idx, res, backfillFinishedAt, resState) })
+		eg.Go(func() error {
```
williamhbaker, line 80: This will get any async stream errors into the logs, but I think its still possible to end up with "context cancelled" being the last thing logged by the connector which I believe would have that error show up in the UI and not be helpful for a user interacting through the UI. This would happen if there are multiple bindings where one completes a backfill and starts the `ChangeStream` and the `ChangeStream` fails right away and cancels the context which the next backfill loop hits and errors out on.

Could we use a separate context for the `errgroup`? Ex. on line 57 use `eg, grpCtx := errgroup.WithContext(ctx)` and then use that `grpCtx` only for `ChangeStream`. That should prevent failed streams from cancelling the backfill context. It would _also_ result in all of the backfills completing (if they can) before an error in the streams is surfaced by `eg.Wait()`. I'm not sure if its important to fail fast in the case of an error'd change stream. I'm assuming its not, but if it is this might not work.

williamhbaker (approved):

source-mongodb/pull.go
mdibaiee: @williamhbaker that's actually an interesting proposal, I think we can have a different context for backfill and changestream jobs and let backfill jobs finish even if a ChangeStream has failed, before we bail out 🤔

mdibaiee: Tested this, works 👍🏽

Issue #692, opened 2023-05-01
User-facing errors for connectors


williamhbaker:
We handle some common errors that occur within connectors in a way that allows for an informative user-facing error message already, and we'd like to continue to expand this coverage for typical error cases encountered by users.

On the connector side, when we encounter one of these errors and handle it we want to omit any "wrapping" of the error as it bubbles up through the connector code and is eventually emitted as a log line on `stdout`. For connectors in Go, a specific error type and code in the boilerplate can make this easier to achieve.

PR #698, opened 2023-05-03
user-facing error messages for connectors

williamhbaker:
**Description:**

The objective of this change is to make it possible for connectors to emit specific user-facing error messages that are caught/handled in the connector code without the extra "wrapping" typical of developer focused errors as they propagate up.

This can now be done by building a `UserError`, providing a message that is meant to be user-facing and optionally the original or "source" error that will be included in the structured log output for diagnostic purposes.

This should not be merged until https://github.com/estuary/flow/pull/1028 is deployed. That change will allow the agent to extract the user-facing message from the structured log.

Along with the creation of `UserError` various connectors have been updated to use it and handle specific common error cases. There are certainly _many_ more connectors and errors that could be handled in this way, but its a start.

As an example, consider this attempted publication for a MongoDB materialization. The configuration is correct except for the database, which doesn't exist.

Prior to any changes to the connector, agent, or UI, we'd get this in the current state:
image
_A dedicated user might be able to read to the end of this error message and guess that something has gone wrong with authentication. Determining the root cause is the database not existing would take some effort._

With the mentioned changes to the agent active but without any changes to the connector this is the message that would be displayed:
image
_This is a little better, and there's less preamble to get through. But there's still quite a bit of error wrapping from the connector that isn't going to mean much to a user and the error message itself is still very low-level since its directly from the go mongo library._

With the changes in this PR implemented in addition to the agent changes, this is the end result:
image
_All of the wrapping is omitted and a more user-facing error is output. Admittedly this still isn't perfect: It would be nice if we could differentiate an incorrect username/password vs. a database not existing, but in the case of the go mongo library that isn't possible. It should be much easier to troubleshoot this error for an end user though. Other connectors can make these kinds of distinctions more clearly, depending on the behavior of the libraries they use._

Closes https://github.com/estuary/connectors/issues/692

**Workflow steps:**

Use connectors and observe better error messages.

N/A

**Notes for reviewers:**

The first commit has a lot of changed files since I did some restructuring of how go connectors copy packages into their docker builds - see the commit message for a full explanation.

The commits following that are implementing `UserError` in various places, including adding explicit error case handling for some widely-used connectors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/698)


go/connector-errors/errors.go
```diff
@@ -0,0 +1,57 @@
...
+// UserError wraps a source error with a user-facing message for the error string. The source error
+// can be provided so that it can be logged separately from the user-facing message for diagnostic
+// purposes.
+type UserError struct {
+	message string
+	source  error
+}
+
+// NewUserError creates a UserError that will output message as the error string.
+func NewUserError(message string, source error) *UserError {
```
willdonnelly, line 19: Typical style for this sort of error-wrapping function in Go would be `NewUserError(err, "this is a message")` so that the short, almost-always-identical prefix can be gotten out of the way first.

willdonnelly (approved):
This looks good to me.

PR #705, opened 2023-05-09
source-{mysql,postgres,sqlserver}: Add support for keyless backfills

willdonnelly:
**Description:**

This is a fairly complicated change, although despite the massive lines-changed count there isn't _that_ much actual code modified -- we just have a lot of lines of test snapshots and adding a cursor property to every backfill document looks like a big deal.

At a high level, this PR:

- Adds a new `Backfill Mode` property to resource configs with three options:
  - `Normal`: The existing behavior that we've always done.
  - `Only Changes`: Skips backfilling the table entirely and just streams new changes. This is a replacement for the old hacky `config.Advanced.SkipBackfills` mechanism which should be deprecated and removed now that we have a less awful way of doing the same thing.
  - `Without Primary Key`: The new keyless backfill mechanism.
- Adds a persistent per-stream backfilled rows counter to the capture state, and embeds that counter value into all backfilled documents in the database-specific LSN/cursor metadata property.
  - On Postgres this looks like `"lsn":[-1,<Counter>,0]`
  - On MySQL this looks like `"cursor":"backfill:<Counter>"`
  - On SQL Server this looks like `"lsn":"","seqval":"<Counter | base64>"`
- Adds new backfill logic so that tables with a `Without Primary Key` backfill mode will be "paginated" using `OFFSET n` queries rather than a greater-than-key comparison. In this mode all replication events will be emitted and "patching" of buffered backfill rows will never occur, although buffering still takes place until the next watermark is observed after each backfill query.
- Discovery is tweaked to suggest, if a table has no primary key or suitable indices, using the `Without Primary Key` backfill mode and a collection key of "whatever the database-specific cursor/LSN property is for this database".

Some rough edges and other details worth noting:

- The values of the `Backfill Mode` property are capitalized and spaced for human consumption, like `"Without Primary Key"`. This feels weird and unnatural for enum values, but it does make things noticeably nicer in the UI and strictly speaking there's nothing wrong with it. As far as I could tell there's no way for us to have the values be some short keyword and then provide a nicer human-readable string to be rendered there.
- In order to suggest the cursor/LSN property as the collection key that property needs to be marked required in discovered schemas. It was simplest to just do this across the board rather than conditionally based on our key recommendation, and AFAIK this should be a safe change to make (as it will only impact future discovered schemas and we will always be setting these fields going forward).
- Likewise the backfill counter value is maintained for and embedded in _all_ backfills, not just keyless ones. This seems reasonable (although it should be noted that counter values could fail to be strictly sequential during a keyed backfill if there are change events which get patched into the buffered result set. This seems fine -- while they're usually sequential it's not a hard requirement and the values are still unique and monotonically increasing).
- Keyless backfill queries use `OFFSET n` for pagination, but on Postgres and SQL Server they use the `ctid` / `%%physloc%%` system columns for ordering and on MySQL they use the combination of all rows in the table. The `ORDER BY ctid` approach was less hassle overall and I believe should be at least as good as any other option there.

**Workflow steps:**

Users should now be able to blindly perform discovery and click "Next" in the UI and end up with a successful capture even from keyless tables. Unless I've overlooked something important, existing uses and new captures of keyed tables should behave exactly the same as before, modulo minor changes around the cursor/LSN property always being present going forward.

I don't recall whether we documented the whole issue of backfill keys and primary keys before. If we do, this should be updated to reflect the new capability in the MySQL/Postgres/SQL Server capture docs.

We should probably also discuss the `Backfill Mode` property in the documentation.

**Notes for reviewers:**

I've tried to mostly put snapshot updates in separate commits from actual code changes, so you might find it useful to review this commit-by-commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/705)


sqlcapture/interface.go
```diff
@@ -73,7 +72,7 @@ type ChangeEvent struct {
 // FlushEvent informs the generic sqlcapture logic about transaction
 // boundaries.
 type FlushEvent struct {
-	Source SourceMetadata
+	Cursor string // TODO(wgd): Maybe json.RawMessage?
```
williamhbaker, line 75: Not sure if this is a WIP todo or something for a future commit. With the way I'm reading this PR it seems like a string worked pretty well.

sqlcapture/capture.go
```diff
@@ -194,11 +201,27 @@ func (c *Capture) Run(ctx context.Context) (err error) {
...
+		if binding == nil {
+			return fmt.Errorf("internal error: no binding information for stream %q", streamID)
+		}
+
+		logrus.WithFields(logrus.Fields{"stream": streamID, "mode": binding.Resource.Mode}).Info("activating replication for stream")
+
 		var state = c.State.Streams[streamID]
-		state.Mode = TableModeBackfill
+		switch binding.Resource.Mode {
+		case BackfillModeNormal, "":
```
williamhbaker, line 213: Unless I'm missing something, I'd tend to prefer having `(r Resource) Validate()` set this value to `"Normal"` if it's an empty string. That would make this case check more direct. But it doesn't look like we are doing an empty string check anywhere other than here, so I don't feel super strongly about this.

williamhbaker (approved):
LGTM

High-level comment on using `ctid` for the postgres keyless ordering key: The postgres docs say that "a row's `ctid` will change if it is updated or moved by `VACUUM FULL`". So I guess running a `VACUUM FULL` in the middle of a backfill would be ill-advised. I'm also wondering about a row being updated changing its `ctid` - could it ever end up that the `ctid` of a row that has not been backfilled is made _lower_ so it ends up getting missed on a later backfill?

source-sqlserver/backfill.go
```diff
@@ -121,6 +152,19 @@ var columnTypeCollatedText = map[string]bool{
 	"nvarchar": true,
 }
 
+func (db *sqlserverDatabase) keylessScanQuery(info *sqlcapture.DiscoveryInfo, schemaName, tableName string) string {
+	var orderColumns []string
```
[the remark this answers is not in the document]
willdonnelly: Good catch, that's left over from when I was using the full set of table rows as the key (before I decided that `%%physloc%%` was the better option).

sqlcapture/capture.go
willdonnelly: Yeah, making that get defaulted somewhere earlier in startup seems reasonable. I think we generally prefer to have `Validate()` be side-effect free (see for instance the `Validate()` vs `SetDefault()` methods on the connector config structs) so I'll probably do it that way though.

willdonnelly:
> LGTM
> 
> High-level comment on using `ctid` for the postgres keyless ordering key: The postgres docs say that "a row's `ctid` will change if it is updated or moved by `VACUUM FULL`". So I guess running a `VACUUM FULL` in the middle of a backfill would be ill-advised. I'm also wondering about a row being updated changing its `ctid` - could it ever end up that the `ctid` of a row that has not been backfilled is made _lower_ so it ends up getting missed on a later backfill?

Yeah, the `ctid` is strictly speaking an unstable row identifier, and AFAIK there's no guarantee that a row won't get moved earlier. However the advantages of using `ctid` or `%%physloc%%` as the key where possible are that:
1. There are fewer edge cases in typical operation -- no weird column types, bizarre collation settings, et cetera can impact the behavior of a keyless backfill.
2. **Actually not an advantage after all**: I _thought_ it would be more efficient, at least for Postgres, since in some previous testing we saw performance advantages from using CTIDs when scanning over large tables. However, I did a quick check just now so I could try and demonstrate that and got unexpected results [1], so I think performance-wise it's a wash for now.

[1] After digging into this some more I've investigated the actual Postgres code that generates `TID Range Scan` plan nodes, and apparently the efficient execution path can only be reached by a query like `SELECT * FROM <table> WHERE ctid <relop> <value>`. Specifically, adding `ORDER BY ctid` to this query forces us back into the "sort the entire dataset" execution path, and so even though it seems like `ORDER BY ctid OFFSET n` ought to be equivalent to the where-clause form it isn't equivalent in terms of performance.

I hadn't previously realized the efficient path was so limited, so as currently implemented this isn't any better than an "order by all columns" approach. This could in principle be fixed by tweaking the Postgres backfills behavior to:
1. Generate the query `SELECT ctid, * FROM ${table} WHERE ctid > ${key} LIMIT ${size}` and use the resume-cursor plumbing instead of the backfill counter offset for keyless backfills too.
2. Extract the `ctid` column value as the row key for backfill results.
3. Probably omit the `ctid` value from the actual backfilled row contents to avoid cluttering things up.

I'm a little leery of doing this because paginated queries without an `ORDER BY` clause worry me quite a bit. Strictly speaking there is no guarantee at all (even assuming no vacuums) that we'll get a stable result ordering or that the N results we receive will be the lowest N values following the where-cursor. In practice I _think_ this is the case but I'm not sure if we should bet on that.

willdonnelly:
There is in fact one terse comment in the relevant Postgres code:

```c
TidRangePath * create_tidrangescan_path(
  PlannerInfo *root, RelOptInfo *rel, List *tidrangequals, Relids required_outer) {
	TidRangePath *pathnode = makeNode(TidRangePath);
        [...]
	pathnode->path.pathkeys = NIL;	/* always unordered */
```

Which _I think_ is explicitly stating that there is no defined sort ordering on the output rows (and thus the planner is correct to require sorting the output if I request `ORDER BY ctid`). The presence of the comment suggests that the programmer who wrote this logic thought that TID range scans would not always produce ordered output, and I'm not about to second-guess them on that.

(However in my own testing I have yet to observe a TID range scan produce unordered results, and we're already being looser on data consistency requirements for keyless tables, so I'm still torn on whether it might be reasonable to rely on CTID ordering in practice...)

willdonnelly:
> Will this work with existing captures which have written data that are then re-discovered (maybe to catch a new table or column)? I'm thinking about reading from those collections if they had not previous had the cursor/LSN property marked as required and had not previously included the value in the output documents.

Hmm, that's a good point, I think things would break in that situation, although I'm not 100% certain of how the catalog merging logic works there.

Which sucks because I would really prefer to just make these properties always-present and always-required for simplicity.

source-sqlserver/backfill.go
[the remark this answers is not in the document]
willdonnelly: Fixed

sqlcapture/interface.go
willdonnelly: Removed

willdonnelly:
> Which sucks because I would really prefer to just make these properties always-present and always-required for simplicity.

As discussed after standup today I've left this logic as it was when I first uploaded this PR -- the cursor/LSN properties will begin being marked as required in all future discovery outputs. Users could potentially hit this rough edge if they re-run discovery on a preexisting MySQL or SQL Server capture and then attempt to materialize an old collection with preexisting data to a new destination, but this should be rare in practice and they can be guided to relax that requirement in their collection schema, which will fix the issue.

Regarding the whole "efficient Postgres backfills by CTID" question I am opting to leave it a little bit unresolved in the interest of getting this merged sooner. We should still be able to change the details of how Postgres backfills work going forward (I could list out several ways of doing that migration without breaking things), and I'd like to get the rest of this work finished before getting bogged down in that again.

willdonnelly:
I believe all outstanding concerns have been addressed, so assuming CI passes after my tweaks I intend to merge this tomorrow morning.

PR #706, opened 2023-05-09
materialize-snowflake: split internal stage PUTs into smaller files

williamhbaker:
**Description:**

We've been uploading data as one large file per binding to Snowflake for staging data, and that has resulted in a few problems:
- Most significantly, the `gosnowflake` driver seems to get bogged down encrypting these files when they are multiple gigabytes in size, which is typical for a non-trivial transaction duration.
- The upload for large files apparently fails entirely if they are too large under some circumstances, resulting in an endless loop of extremely long encryption followed by failed uploads.
- Snowflake can process many small files in parallel, which should offer performance improvement vs. a single large file

This changes the file staging to split the data up into a series of smaller files which are then collectively queried together. The PUTs to Snowflake internal stages can be completed concurrently with populating additional local files, and the smaller local file sizes allow for much faster encryption.

The performance benefits will be evident when running "catch up" materializations on larger datasets. Some rough benchmarking I did with a ~10GB dataset indicates a 2-3x improvement over the current version, with the new version using a 5 minute `maxTxnDuration` and the current version using a 1 second `maxTxnDuration`. This may seem like a strange comparison, but the current version crashes under extended transactions and actually performs best when it using the smallest possible transactions. 10 GB really isn't much in the grand scheme of things, and I expect more of a benefit on even larger datasets.

To test this change, I created a new integration test for `materialize-snowflake` and verified that the snapshot output is identical with the current and new versions. I also verified that the table data in the endpoint for the previously mentioned ~10GB test dataset is the same whether materialized by the old or new version using a variety of different transaction sizes.

**Workflow steps:**

Use snowflake like before, but see it function well with larger transactions.

N/A

**Notes for reviewers:**

An interesting note about the concurrent PUT worker is that it seemed pretty balanced with reading data from the Flow iterators with just a single worker. So there would not be any benefit to running more than worker per binding that I could see, and it keeps things a little simpler.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/706)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bufio"
 	"context"
 	stdsql "database/sql"
 	"encoding/base64"
@@ -10,7 +9,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"regexp"
 	"strings"
 
@@ -18,7 +16,6 @@ import (
 	sql "github.com/estuary/connectors/materialize-sql"
 	pf "github.com/estuary/flow/go/protocols/flow"
 	pm "github.com/estuary/flow/go/protocols/materialize"
-	"github.com/google/uuid"
 	log "github.com/sirupsen/logrus"
 	sf "github.com/snowflakedb/gosnowflake"
 	"go.gazette.dev/core/consumer/protocol"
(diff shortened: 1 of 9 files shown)


willdonnelly (approved):
LGTM

This PR adds enough code that I feel like I should have some sort of more detailed comment about something, but everything looks great to me as-is

dyaffe:
Nice, does this mean we can merge it?  Would love to have some experience understanding how it affects production data flows before our conversation on Thursday.

Issue #703, opened 2023-05-08
source-mysql: Improve heartbeats and timeout behavior


willdonnelly:
So long as a backfill is ongoing or the database is busy everything is fine, but in some cases it looks like MySQL captures might run into trouble if the replication connection is idle for too long. To fix this we should be explicitly requesting periodic heartbeats from the source database and setting a read timeout so that connectivity failures will be more quickly detected.

Specifically, we should modify the replication streaming config to set `HeartbeatPeriod` to something like 60 seconds and `ReadTimeout` to something like 5 minutes. In theory those two changes are all it will take to improve things, but I'd like to do a bit of testing first so I'm not going to just YOLO a PR for that right now.

PR #707, opened 2023-05-10
source-mysql: Request heartbeats and set read timeout

willdonnelly:
**Description:**

This fixes https://github.com/estuary/connectors/issues/703 by requesting heartbeat events from the server every 30s and setting a message read timeout of 5 minutes.

The heartbeats should ensure that the TCP connection remains open under normal conditions even when no other activity is occurring, and the read timeout ensures that we will detect otherwise-silent link failures within minutes rather than hours.

This would be tricky to test automatically in CI, but I have tested manually that this change works as expected, and also verified that without this change a default-configured MySQL instance doesn't send any heartbeat or similar traffic when no changes are occurring, which is entirely consistent with the failures we observed in production.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/707)


williamhbaker (approved):
LGTM

PR #708, opened 2023-05-10
materialize-snowflake: set GOMEMLIMIT=900mb

williamhbaker:
**Description:**

Sets an advisory limit via environment variable in the `materialize-snowflake` dockerfile as a workaround to the memory issues documented in https://github.com/estuary/connectors/issues/709.

This actually seems to work very well in practice and will keep the connector memory right at 900mb when reading load query results for the datasets I have tried it with. This is certainly not an ideal amount of memory to use (we'd like it to use a lot less), and for that I have some initial thoughts in the previously linked issue.

~I'm opting not to make this a global setting via the base image at the moment, since I'd generally prefer to know about high memory use from other connectors via our OOM kill alerts, even if they are something that could be remedied by more active GC, so that we can know to optimize them.~ Actually ended up setting this for all connectors, see comments.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/708)

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -12,4 +12,8 @@ RUN mkdir /home/nonroot/.ssh
 RUN chmod 700 /home/nonroot/.ssh
 RUN chmod 755 /home/nonroot
 
+# Ask the Go runtime to keep heap size below 900MiB. Our real memory limit is
+# 1GiB so this is 10% headroom as recommended in https://go.dev/doc/gc-guide
+ENV GOMEMLIMIT=900MiB
+
 USER root
--- a/source-mysql/Dockerfile
+++ b/source-mysql/Dockerfile
@@ -42,10 +42,6 @@ COPY --from=builder /builder/connector ./source-mysql
 # Avoid running the connector as root.
 USER nonroot:nonroot
 
-# Ask the Go runtime to keep heap size below 900MiB. Our real memory limit is
-# 1GiB so this is 10% headroom as recommended in https://go.dev/doc/gc-guide
-ENV GOMEMLIMIT=900MiB
-
 ENTRYPOINT ["/connector/source-mysql"]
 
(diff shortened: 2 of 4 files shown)


willdonnelly:
LGTM

> I'm opting not to make this a global setting via the base image at the moment, since I'd generally prefer to know about high memory use from other connectors via our OOM kill alerts, even if they are something that could be remedied by more active GC, so that we can know to optimize them.

I'd dispute this. Part of the problem is that the Go GC algorithm has _no reason_ to try and keep memory usage bounded at all if it can improve throughput by holding onto freeable memory for longer. The `GOMEMLIMIT` flag is (explicitly, per the documentation) the official mechanism for how we're supposed to tell the runtime "you're running in a constrained environment, try to keep it down please".

It will still allocate more memory and OOM if the live heap grows too large, the limit is deliberately "soft" to avoid GC thrashing.

williamhbaker:
> I'd dispute this. Part of the problem is that the Go GC algorithm has _no reason_ to try and keep memory usage bounded at all if it can improve throughput by holding onto freeable memory for longer. The `GOMEMLIMIT` flag is (explicitly, per the documentation) the official mechanism for how we're supposed to tell the runtime "you're running in a constrained environment, try to keep it down please".
> 
> It will still allocate more memory and OOM if the live heap grows too large, the limit is deliberately "soft" to avoid GC thrashing.
> 
> So having had this on the SQL captures for a few weeks now and seeing that it helps another connector, I'm thinking it's about time to consider setting it on the base image.

Fair enough, I was a bit on the fence about this. Getting our connector memory monitoring working again would probably be the ideal way to identify memory optimization opportunities rather than letting them blow up because the GC doesn't have the information it needs to run in the environment available to it.

willdonnelly (approved):
LGTM

PR #710, opened 2023-05-10
source-mysql: Ignore /* comments like this */

willdonnelly:
**Description:**

Modifies `source-mysql` to ignore `/* comments like this */` when they're reported as query events via replication.

Also adds a test because this regex is getting long enough to need one even if each individual case is fairly simple.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/710)


willdonnelly:
I'm going to go ahead and YOLO merge this, since a capture in production is erroring on this and aside from the new test it's just adding another case to the "queries we can safely ignore" regex. The specific case is only slightly worrisome because it contains an escaped `*` character, but that's why I added a test at the same time, so I'm pretty sure it'll be fine.

PR #713, opened 2023-05-12
source-sftp: new connector

williamhbaker:
**Description:**

Creates a new connector, `source-sftp`, for capturing data from SFTP servers.

This initial version works with password authentication, which is the immediate need. Certificate-based authentication will probably need to be added at some point as well. Other future work might include connecting over an SSH tunnel for SFTP servers that are not accessible over the public internet.

Most of the heavy lifting is done by the existing `filesource` framework, and the code here facilitates connecting to a remote SFTP server, listing the contents of its filesystem, and streaming data of its files.

In addition to the automated tests included with this PR I did some "real world" testing of the connector, capturing multiple tens of gigabytes of data from a mixture of lots of very small files to one or more very large files, from an OpenSSH server on a connection over the public internet. Everything worked as expected with these tests.

**Workflow steps:**

Use the connector to capture from an SFTP server.

See https://github.com/estuary/flow/pull/1037

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/713)


source-sftp/main.go
```diff
@@ -0,0 +1,397 @@
...
+func configSchema(parserSchema json.RawMessage) json.RawMessage {
+	schema := schemagen.GenerateSchema("SFTP Source", &config{})
+
+	// TODO(whb): This is a little gross but I don't have a better idea for how to embed the parser
+	// schema in the overall schema, and this works.
+	schema.Properties.Set("parser", parserSchema)
+
+	out, err := schema.MarshalJSON()
+	if err != nil {
+		log.Fatal(fmt.Errorf("generating endpoint schema: %w", err))
```
willdonnelly, line 393: Are you intentionally using the `"log"` package here and `logrus` for other logging, or is that accidental? (In either case, this sort of ambiguity is why I prefer consistently referring to the Logrus packge as `log` in new code).

tests/source-sftp/setup.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Set ID_TYPE to string because parsing CSV files will always result in string values.
+export ID_TYPE=string
+
+export RESOURCE="{ \"stream\": \"${TEST_STREAM}\", \"syncMode\": \"incremental\" }"
+
+docker compose -f source-sftp/docker-compose.yaml up --detach
+
+# Wait for the sftp server to be ready.
+go run tests/source-sftp/wait_for_server.go
```
willdonnelly, line 17: Nice use of a helper program 👍

Should the address/username/password flags maybe be specified explicitly? (Unless you're deliberately taking advantage of having the address be `source-sftp-db-1.flow-test:2222` for the capture connector within Docker but `localhost:2222` when running this binary, I guess?)

willdonnelly (approved):
LGTM

tests/source-sftp/setup.sh
williamhbaker: The defaults do happen to work for the instance set up by the automated tests, but I think having the flags set is more clear so I'll add those.

source-sftp/main.go
williamhbaker: Accidental! I'll update to use `logrus` everywhere, aliased as `log`.

source-sftp/main.go
```diff
@@ -0,0 +1,397 @@
...
+}
+
+// visit wraps fs.FileInfo and retains the full path to the file, since fs.FileInfo only provides
+// the file name.
+type visit struct {
+	fs.FileInfo
+	path string
+}
+
+func (s *sftpSource) Read(_ context.Context, obj filesource.ObjectInfo) (io.ReadCloser, filesource.ObjectInfo, error) {
```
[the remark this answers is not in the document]
williamhbaker: Yeah in practice the runtime will nuke any connectors it needs to and there is no strict requirement for connectors to monitor/propagate signals or context cancellations, although it would be nicer if the `sftp` methods could. But it shouldn't be a problem.

Issue #714, opened 2023-05-15
source-mysql: add configuration for setting timezone


williamhbaker:
Previously, if the timezone couldn't be determined from the system `'time_zone'` variable a warning message was logged and captures failed at runtime if a table contained a `datetime` column.

https://github.com/estuary/connectors/pull/668 changed that with some additional logic in `Validate` to make the capture fail to publish if `datetime` column was being captured with no `'time_zone'` system variable set.

Failing to publish with an error from `Validate` is better than successfully publishing and failing later, but there are still some setups (ex: mysql aurora, where a parameter group must be created and attached) where it is difficult to set the `'time_zone'` system variable. Because of this we would like to add a configuration to allow for setting a time zone that will be used for situations when the system variable is not set and cannot be set to allow captures to work with `datetime` columns.

PR #715, opened 2023-05-15
source-mysql: add timezone configuration

williamhbaker:
**Description:**

There are some MySQL-based setups where it is difficult to set the `'time_zone'` system variable. This adds a configuration value to allow for configuring a timezone as part of the capture that will be used for `datetime` columns when the the `'time_zone'` system variable cannot be set on the database by the user.

Closes https://github.com/estuary/connectors/issues/714

**Workflow steps:**

Will need to update the mysql capture docs to reflect this new configuration parameter and describe its usage.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/715)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -104,7 +104,7 @@ func fixMysqlLogging() {
 	}
 
 	mysqlLog.SetDefaultLogger(mysqlLog.NewDefault(handler))
-	// Looking at the source code, it seems that the level names pretty muc" match those used by logrus.
+	// Looking at the source code, it seems that the level names pretty much match those used by logrus.
 	// In the event that anything doesn't match, it'll fall back to info level.
 	// Source: https://github.com/siddontang/go-log/blob/1e957dd83bed/log/logger.go#L116
 	mysqlLog.SetLevelByName(logrus.GetLevel().String())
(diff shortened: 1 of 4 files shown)


source-mysql/main.go
```diff
@@ -244,11 +251,38 @@ func (db *mysqlDatabase) connect(ctx context.Context) error {
 		return fmt.Errorf("unable to connect to database: %w", err)
 	}
 
-	// Infer the location in which captured DATETIME values will be interpreted.
-	if loc, err := queryTimeZone(conn); err == nil {
+	// Infer the location in which captured DATETIME values will be interpreted. The system variable
+	// 'time_zone' is preferentially used if it is set on the database.
```
willdonnelly, line 255: It seems like it might be better to prioritize the config property over the database setting?

I don't have a really strong opinion here, but usually for this sort of "leave blank to autodetect" setting it feels very surprising for explicitly setting the value to _not_ take precedence over the autodetection logic.

willdonnelly (approved):
LGTM

source-mysql/main.go
williamhbaker: Yeah I think you're right, stepping back a bit. I was approaching this kind of reluctantly, leaving it as a "last resort", but it makes more sense for an explicit configuration to take precedence if set rather than for it to not take effect under situations the user might not be aware of.

PR #718, opened 2023-05-17
source-firestore: Deduplicate discovered resources

willdonnelly:
**Description:**

When there are multiple nested collections mapping to the same resource path, such as `foo/1/bar` and `foo/2/bar` mapping to `foo/*/bar`, we only want to emit one discovered binding at the end.

This used to happen as a side-effect of how we launched schema inference workers for each distinct resource path, and I think we accidentally lost that behavior as a side effect of the change which removed schema inference.

This issue was hit by a user in production. I tweaked `TestDiscovery` to make sure it was exercising the issue, fixed it in the obvious manner, and then verified that `TestDiscovery` output changed and only contains one copy of the binding again. That's the full extent of the testing or thought I've put into this change though, so it's possible I've missed something.

(Note that the `TestDiscovery` snapshot shows a pretty large diff because it was never updated after the removing-schema-inference or protocol-refactoring changes -- I'm pretty sure that's all irrelevant and should be ignored, the only important thing for the purposes of this PR is that the set of discovered bindings is the same.)

### DIFF
--- a/source-firestore/.snapshots/TestDiscovery
+++ b/source-firestore/.snapshots/TestDiscovery
@@ -1,226 +1,177 @@
 Binding 0:
 {
     "recommended_name": "flow_source_tests",
-    "resource_spec_json": {
+    "resource_config_json": {
       "path": "flow_source_tests",
       "backfillMode": "async"
     },
     "document_schema_json": {
-      "allOf": [
-        {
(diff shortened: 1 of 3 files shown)


mdibaiee (approved):
LGTM 👍🏽

PR #719, opened 2023-05-18
materialize-sql: recommend string field locations that formatted as numeric

williamhbaker:
**Description:**

Fields that are strings but with a numeric format previously were considered "optional" to include in the materialization. This changes them to be recommended so they are materialized by default without needing specific field selections.

This is helpful for cases like a collection captured from a CSV file where all the values are strings and schema inference is able to infer that they are numeric values encoded as strings and adds the appropriate schema for that. In these cases the expected behavior is that the fields will be materialized without additional configuration.

**Workflow steps:**

**Notes for reviewers:**


mdibaiee (approved):
LGTM!

PR #722, opened 2023-05-19
Jshearer/materialize slack

jshearer:
**Description:**

This fleshes out the Slack materialization connector that @willdonnelly originally wrote for one of our hackathons.

It automatically deals with joining channels when needed, and lets you specify a username and profile picture to communicate as. It also supports rich messages via Slack's `blocks` API

**Note:**
One thing I don't love is the config schema, which results in the following UI:
Screen Shot 2023-05-19 at 13 36 05
By default, `Sender Config` is collapsed. This comes from 
```go
type config struct {
	SenderConfig SlackSenderConfig `json:"sender_config" jsonschema:"title=Slack Config"`
	Credentials  CredentialConfig  `json:"credentials" jsonschema:"title=Authentication" jsonschema_extras:"x-oauth2-provider=slack"`
}
```
I had to extract `SlackSenderConfig` because it was used in a few different places. What I really want is the equivalent to [Serde's flatten attribute](https://serde.rs/attr-flatten.html), but I can't figure out a way to get Go to do the same thing. Any ideas?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/722)

### DIFF
--- a/materialize-slack/Dockerfile
+++ b/materialize-slack/Dockerfile
@@ -2,7 +2,7 @@ ARG BASE_IMAGE=ghcr.io/estuary/base-image:v1
 
 # Build Stage
 ################################################################################
-FROM golang:1.17-bullseye as builder
+FROM --platform=linux/amd64 golang:1.20-bullseye as builder
 
 WORKDIR /builder
 
@@ -12,10 +12,9 @@ COPY go.* ./
 RUN go mod download
 
 # Build the connector projects we depend on.
+COPY go                        ./go
 COPY materialize-boilerplate   ./materialize-boilerplate
 COPY materialize-slack         ./materialize-slack
-COPY go/schema-gen             ./go/schema-gen
-COPY go/connector-errors       ./go/connector-errors
 
 # Test and build the connector.
 RUN go test  -tags nozstd -v ./materialize-slack/...
--- a/materialize-slack/Dockerfile
+++ b/materialize-slack/Dockerfile
@@ -14,7 +14,8 @@ RUN go mod download
 # Build the connector projects we depend on.
 COPY materialize-boilerplate   ./materialize-boilerplate
 COPY materialize-slack         ./materialize-slack
-COPY go-schema-gen             ./go-schema-gen
+COPY go/schema-gen             ./go/schema-gen
+COPY go/connector-errors       ./go/connector-errors
 
 # Test and build the connector.
 RUN go test  -tags nozstd -v ./materialize-slack/...
--- a/materialize-slack/Dockerfile
+++ b/materialize-slack/Dockerfile
@@ -1,0 +1,38 @@
+ARG BASE_IMAGE=ghcr.io/estuary/base-image:v1
+
+# Build Stage
+################################################################################
+FROM golang:1.17-bullseye as builder
+
+WORKDIR /builder
+
+# Download & compile dependencies early. Doing this separately allows for layer
+# caching opportunities when no dependencies are updated.
+COPY go.* ./
+RUN go mod download
+
+# Build the connector projects we depend on.
+COPY materialize-boilerplate   ./materialize-boilerplate
+COPY materialize-slack         ./materialize-slack
+COPY go-schema-gen             ./go-schema-gen
+
+# Test and build the connector.
+RUN go test  -tags nozstd -v ./materialize-slack/...
+RUN go build -tags nozstd -v -o ./connector ./materialize-slack
+
+# Runtime Stage
+################################################################################
+FROM ${BASE_IMAGE}
+
+WORKDIR /connector
+ENV PATH="/connector:$PATH"
+
+# Bring in the compiled connector artifact from the builder.
+COPY --from=builder /builder/connector ./connector
+
+# Avoid running the connector as root.
+USER nonroot:nonroot
+
+LABEL FLOW_RUNTIME_PROTOCOL=materialize
+
+ENTRYPOINT ["/connector/connector"]
--- a/materialize-slack/Dockerfile
+++ b/materialize-slack/Dockerfile
@@ -2,7 +2,7 @@ ARG BASE_IMAGE=ghcr.io/estuary/base-image:v1
 
 # Build Stage
 ################################################################################
-FROM --platform=linux/amd64 golang:1.20-bullseye as builder
+FROM --platform=linux/amd64 golang:1.22-bullseye as builder
 
 WORKDIR /builder
 
--- a/materialize-slack/main.go
+++ b/materialize-slack/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
@@ -122,24 +123,42 @@ func (driver) Validate(ctx context.Context, req *pm.Request_Validate) (*pm.Respo
 
 		var constraints = make(map[string]*pm.Response_Validated_Constraint)
 		for _, projection := range binding.Collection.Projections {
-			constraints[projection.Field] = &pm.Response_Validated_Constraint{
-				Type:   pm.Response_Validated_Constraint_FIELD_OPTIONAL,
-				Reason: "All fields other than 'ts', 'text', and 'blocks' will be ignored",
+			var constraint pm.Response_Validated_Constraint
+
+			switch {
+			case projection.Field == "ts":
+				constraint = pm.Response_Validated_Constraint{
+					Type:   pm.Response_Validated_Constraint_FIELD_REQUIRED,
+					Reason: "The Slack materialization requires a message timestamp",
+				}
+			case projection.Field == "text":
+				constraint = pm.Response_Validated_Constraint{
+					Type:   pm.Response_Validated_Constraint_FIELD_REQUIRED,
+					Reason: "The Slack materialization requires 'text'",
+				}
+			case projection.Field == "blocks":
+				constraint = pm.Response_Validated_Constraint{
+					Type:   pm.Response_Validated_Constraint_FIELD_REQUIRED,
+					Reason: "The Slack materialization would like 'blocks'",
+				}
+			case projection.IsPrimaryKey:
+				constraint = pm.Response_Validated_Constraint{
+					Type:   pm.Response_Validated_Constraint_LOCATION_REQUIRED,
+					Reason: "Keys are all required",
+				}
+			default:
+				constraint = pm.Response_Validated_Constraint{
+					Type:   pm.Response_Validated_Constraint_FIELD_OPTIONAL,
+					Reason: "All fields other than 'ts', 'text', and 'blocks' will be ignored",
+				}
 			}
-		}
-		constraints["ts"] = &pm.Response_Validated_Constraint{
-			Type:   pm.Response_Validated_Constraint_LOCATION_REQUIRED,
-			Reason: "The Slack materialization requires a message timestamp",
+			constraints[projection.Field] = &constraint
 		}
 
-		constraints["text"] = &pm.Response_Validated_Constraint{
-			Type:   pm.Response_Validated_Constraint_LOCATION_RECOMMENDED,
-			Reason: "The Slack materialization requires either text or blocks",
+		if constraints["text"] == nil {
+			return nil, fmt.Errorf("'text' field required")
 		}
-		constraints["blocks"] = &pm.Response_Validated_Constraint{
-			Type:   pm.Response_Validated_Constraint_LOCATION_RECOMMENDED,
-			Reason: "The Slack materialization requires either text or blocks",
-		}
+
 		out = append(out, &pm.Response_Validated_Binding{
 			Constraints:  constraints,
 			DeltaUpdates: true,
@@ -243,13 +262,12 @@ func (d *transactor) Load(it *pm.LoadIterator, loaded func(int, json.RawMessage)
 func (t *transactor) Store(it *pm.StoreIterator) (pm.StartCommitFunc, error) {
 	log.Debug("handling Store operation")
 
-	var vals []tuple.TupleElement
 	for it.Next() {
 		var b = t.bindings[it.Binding]
 		var parsed = buildDocument(b, it.Key, it.Values)
 		var tsStr, tsOk = parsed["ts"].(string)
 		var text, textOk = parsed["text"].(string)
-		var blocks, blocksOk = parsed["blocks"].([]byte)
+		var blocks, blocksOk = parsed["blocks"].(json.RawMessage)
 
 		if !tsOk {
 			return nil, fmt.Errorf("missing timestamp")
@@ -279,14 +297,14 @@ func (t *transactor) Store(it *pm.StoreIterator) (pm.StartCommitFunc, error) {
 			err = json.Unmarshal([]byte(escaped_blocks), &blocksParsed)
 
 			if err != nil {
-				log.Warn(fmt.Sprintf("Here is the field ordering: %#v", vals))
-				return nil, fmt.Errorf("invalid blocks value %q: %w", string(blocks), err)
+				return nil, fmt.Errorf("invalid blocks value %q, %q: %w", reflect.TypeOf(blocks), string(blocks), err)
 			}
-
+		} else {
+			log.Warn(fmt.Sprintf("Blocks apparently NOT okay: %+v", reflect.TypeOf(parsed["blocks"])))
 		}
 
 		// Accept messages from at most 30 minutes in the past
-		if time.Since(ts).Minutes() < 30 {
+		if time.Since(ts).Minutes() < 10 {
 			if err := t.api.PostMessage(b.resource.Channel, text, blocksParsed.BlockSet, b.resource.SenderConfig); err != nil {
 				serializedBlocks, marshal_err := json.Marshal(blocksParsed)
 				if marshal_err == nil {
--- a/materialize-slack/main.go
+++ b/materialize-slack/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"strings"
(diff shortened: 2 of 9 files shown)


materialize-slack/main.go
```diff
@@ -189,62 +177,60 @@ func (driver) NewTransactor(ctx context.Context, open pm.Request_Open) (pm.Trans
...
+	spec       *pf.MaterializationSpec_Binding
+	resource   resource
+	collection string
+}
+
+func buildDocument(b *binding, keys, values tuple.Tuple) map[string]interface{} {
+	var document = make(map[string]interface{})
+
+	// Add the `_id` field to the document. This is required by Rockset.
+	document["_id"] = base64.RawStdEncoding.EncodeToString(keys.Pack())
```
willdonnelly, line 215: Probably don't need this line :)

materialize-slack/Dockerfile
```diff
@@ -0,0 +1,39 @@
+ARG BASE_IMAGE=ghcr.io/estuary/base-image:v1
+
+# Build Stage
+################################################################################
+FROM golang:1.17-bullseye as builder
```
williamhbaker, line 5: This dockerfile will need updated to be like the current ones on the main branch, which use go 1.20, and the directory structure of the imports ("Build the connector projects we depend on.") has changed. See https://github.com/estuary/connectors/blob/60728e31593cc7c9a3d78c961b47eab3e84cca07/materialize-pinecone/Dockerfile#L1-L16.

materialize-slack/main.go
```diff
@@ -0,0 +1,357 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("generating resource schema: %w", err)
+	}
+
+	return &pm.Response_Spec{
+		ConfigSchemaJson:         json.RawMessage(endpointSchema),
+		ResourceConfigSchemaJson: json.RawMessage(resourceSchema),
+		DocumentationUrl:         "https://go.estuary.dev/materialize-slack",
+		Oauth2:                   Spec("channels:read", "groups:read", "im:read", "channels:join", "chat:write", "chat:write.customize"),
+		//	                      Spec("channels:history", "channels:join", "channels:read", "files:read", "groups:read", "links:read", "reactions:read", "remote_files:read", "team:read", "usergroups:read", "users.profile:read", "users:read"),
```
williamhbaker, line 82: Left-over comment that can be removed?

materialize-slack/main.go
```diff
@@ -0,0 +1,357 @@
...
+			return nil, fmt.Errorf("error getting channel: %s, %w", res.Channel, err)
+		}
+
+		if !channelInfo.IsMember {
+			if err := api.JoinChannel(res.Channel); err != nil {
+				return nil, fmt.Errorf("error joining channel: %s, %w", res.Channel, err)
+			}
+		}
+
+		/*
```
williamhbaker, line 116: Maybe a left-over comment here too?

materialize-slack/main.go
```diff
@@ -0,0 +1,357 @@
...
+type driverCheckpoint struct {
+	BindingCollections []string `json:"binding_collections"`
+}
+
+func (c driverCheckpoint) Validate() error {
+	return nil
+}
+
+func (driver) Spec(ctx context.Context, req *pm.Request_Spec) (*pm.Response_Spec, error) {
+	log.Debug("handling Spec request")
```
williamhbaker, line 65: These `log.Debug` statements at the start of every driver/transactor method don't seem like they will be very useful, and also generally redundant with debug messages from the runtime. Can we remove them?

materialize-slack/main.go
```diff
@@ -0,0 +1,357 @@
...
+	resource   resource
+	collection string
+}
+
+func buildDocument(b *binding, keys, values tuple.Tuple) map[string]interface{} {
+	var document = make(map[string]interface{})
+
+	// Add the keys to the document.
+	for i, value := range keys {
+		if i < len(b.spec.FieldSelection.Keys) {
```
williamhbaker, line 234: Since we are requiring that all keys be materialized I don't think this length check is required.

materialize-slack/main.go
```diff
@@ -0,0 +1,357 @@
...
+			var propName = b.spec.FieldSelection.Keys[i]
+			document[propName] = value
+		}
+	}
+
+	// Add the non-keys to the document.
+	for i, value := range values {
+		if i < len(b.spec.FieldSelection.Values) {
```
williamhbaker, line 242: Kind of similar here, I think this length check is not required. We'll get all the values from the field selection. If somehow there are more values from the store iterator than in the field selection we would panic without the length check which seems better than ignoring some of them.

materialize-slack/main.go
```diff
@@ -0,0 +1,357 @@
...
+
+		if !tsOk {
+			return nil, fmt.Errorf("missing timestamp")
+		}
+
+		if !textOk {
+			return nil, fmt.Errorf("missing text")
+		}
+
+		// Parse the timstamp as a time.Time
```
williamhbaker, line 280: Typo in comment

materialize-slack/main.go
```diff
@@ -0,0 +1,357 @@
...
+		}
+
+		return &pf.ConnectorState{
+			UpdatedJson: json.RawMessage(bs),
+			MergePatch:  true,
+		}, nil
+	}, nil
+}
+
+func (transactor) Commit(context.Context) error {
```
williamhbaker, line 339: The transactor interface has changed quite a bit since the time it looks like this branch diverged from main. If you rebase onto the latest from main there will be a few changes to make...transactors no longer have a separate commit or acknowledge method, etc.

materialize-slack/main.go
```diff
@@ -0,0 +1,357 @@
...
+			// Mom can we get rate limiting?
+			// we have rate limiting at home
+			// rate limiting at home:
+			time.Sleep(time.Second * 10)
+		} else {
+			log.Warn(fmt.Sprintf("Ignoring message from the past: %q", ts))
+		}
+	}
+
+	return func(ctx context.Context, runtimeCheckpoint *protocol.Checkpoint, runtimeAckCh <-chan struct{}) (*pf.ConnectorState, pf.OpFuture) {
```
williamhbaker, line 324: I don't think we need to have this be an async commit - see [materialize-google-pubsub](https://github.com/estuary/connectors/blob/60728e31593cc7c9a3d78c961b47eab3e84cca07/materialize-google-pubsub/transactor.go#L28-L63) for an example of a simpler store implementation that I think would work well for this.

williamhbaker (commented):
Left some comments as an initial review. You will need to add this to CI in the github workflow so it actually gets built - there might be some changes needed to actually get it building (I would suggest rebasing on the last from the main branch).

I'll plan on doing another detailed review after that is done.

materialize-slack/Dockerfile
jshearer: I think I had already updated this. The change you're talking about is moving things from e.g `go-materialize-boilerplate` to `go/materialize-boilerplate` right?

materialize-slack/main.go
jshearer: You're saying that the slack materialization connector doesn't need checkpoints, which makes sense now that we don't need to keep track of the last message timestamp and just accept any messages from the last 10 minutes. Removed it!

materialize-slack/main.go
jshearer: I rebased, looks like the only change I could find was removing the `Commit()` and `Acknowledge()` functions

materialize-slack/main.go
```diff
@@ -0,0 +1,302 @@
...
+			// Is this jank? This should really be taken care of by the
+			// serialization logic in slack.Block, but it can't be _that_
+			// bad to do it here... right? see below:
+			// https://api.slack.com/reference/surfaces/formatting#escaping
+			var escaped_blocks = string(blocks)
+			escaped_blocks = strings.ReplaceAll(escaped_blocks, "&", "&amp;")
+			escaped_blocks = strings.ReplaceAll(escaped_blocks, "<", "&lt;")
+			escaped_blocks = strings.ReplaceAll(escaped_blocks, ">", "&gt;")
+
+			err = json.Unmarshal([]byte(escaped_blocks), &blocksParsed)
```
williamhbaker, line 267: Nit: Could simplify as `if err := json.Unmarshal([]byte(escaped_blocks), &blocksParsed); err != nil { ...}`

materialize-slack/main.go
```diff
@@ -0,0 +1,302 @@
...
+
+			if err != nil {
+				return nil, fmt.Errorf("invalid blocks value %q, %q: %w", reflect.TypeOf(blocks), string(blocks), err)
+			}
+		} else {
+			log.Warn(fmt.Sprintf("Blocks apparently NOT okay: %+v", reflect.TypeOf(parsed["blocks"])))
```
williamhbaker, line 273: Is it ok to not have any blocks for a materialized document? If not I think we'd want to error out. If it is ok I don't think we would want to log anything since that could get very spammy if materializing a collection where none of the documents have blocks.

materialize-slack/main.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+
+		// Accept messages from at most 10 minutes in the past
+		if time.Since(ts).Minutes() < 10 {
+			if err := t.api.PostMessage(b.resource.Channel, text, blocksParsed.BlockSet, b.resource.SenderConfig); err != nil {
+				serializedBlocks, marshal_err := json.Marshal(blocksParsed)
```
williamhbaker, line 279: `marshal_err` is pretty ugly - I don't see any reason to not have the error variable called `err` here which would be much more conventional. If nothing else `marshalErr` since variable names in Go a typically camel case.

materialize-slack/main.go
```diff
@@ -0,0 +1,302 @@
...
+				if marshal_err == nil {
+					log.Warn(fmt.Sprintf("Parse Blocks: %+v", string(serializedBlocks)))
```
williamhbaker, line 281: Would we want to fail the connector if this happens? I'm thinking we would.

materialize-slack/main.go
```diff
@@ -0,0 +1,302 @@
...
+				}
+				log.Warn(fmt.Errorf("error sending message: %w", err))
```
williamhbaker, line 283: This may be something where we'd want to retry with rate limiting, but it doesn't look like that is implemented here, although I see the partial comment below and the sleep.

materialize-slack/main.go
```diff
@@ -0,0 +1,302 @@
...
+
+			if err != nil {
+				return nil, fmt.Errorf("invalid blocks value %q, %q: %w", reflect.TypeOf(blocks), string(blocks), err)
+			}
+		} else {
+			log.Warn(fmt.Sprintf("Blocks apparently NOT okay: %+v", reflect.TypeOf(parsed["blocks"])))
+		}
+
+		// Accept messages from at most 10 minutes in the past
+		if time.Since(ts).Minutes() < 10 {
```
williamhbaker, line 277: I'm not really following this. What happens if a message is outside of this time window? Does it just get dropped? I would expect a connector like this to provide at-least-once materialization, but if it is possibly dropping messages it would not do that. It feels like there is some kind of filtering happening here that may not generally be desirable.

williamhbaker (commented):
Thanks for updating the transactor! I left some more comments. You will also need to update `.github/workflows/ci.yaml` in order for the connector to be built automatically.

materialize-slack/main.go
```diff
@@ -0,0 +1,302 @@
...
+		var res resource
+		if err := pf.UnmarshalStrict(binding.ResourceConfigJson, &res); err != nil {
+			return nil, fmt.Errorf("parsing resource config: %w", err)
+		}
+		var channelInfo, err = api.ConversationInfo(res.Channel)
+		if err != nil {
+			return nil, fmt.Errorf("error getting channel: %s, %w", res.Channel, err)
+		}
+
+		if !channelInfo.IsMember {
```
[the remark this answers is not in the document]
jshearer: Awesome, thanks for explaining this. Moved!

materialize-slack/main.go
```diff
@@ -0,0 +1,302 @@
...
+				constraint = pm.Response_Validated_Constraint{
+					Type:   pm.Response_Validated_Constraint_FIELD_OPTIONAL,
+					Reason: "All fields other than 'ts', 'text', and 'blocks' will be ignored",
+				}
+			}
+			constraints[projection.Field] = &constraint
+		}
+
+		if constraints["text"] == nil {
+			return nil, fmt.Errorf("'text' field required")
```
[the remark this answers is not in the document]
jshearer: Ah, good point about `ts`. `blocks` is optional

materialize-slack/main.go
```diff
@@ -0,0 +1,302 @@
...
+
+	api, err := cfg.buildAPI()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var bindings []*binding
+	for _, b := range open.Materialization.Bindings {
+		var res resource
+		err = json.Unmarshal(b.ResourceConfigJson, &res)
```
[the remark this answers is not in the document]
jshearer: Yuup thx

materialize-slack/main.go
jshearer: Yep, leftover from debugging

materialize-slack/main.go
jshearer: Actually removed this whole block, it was left over from debugging. Sorry!

materialize-slack/main.go
jshearer: Yes... I think you're right.

materialize-slack/main.go
jshearer: This should probably be tunable, but the idea is that Slack is different from a regular database in that you can't send a message with a timestamp other than _right now_. So given that, the concept of "backfilling" slack messages doesn't make sense, and we really do only want to deliver messages that happen in roughly real time. 10 minutes was picked as a reasonable window for "real time" slack messages to show up.

materialize-slack/main.go
jshearer: Yeah, retries would be nice. Do we have any common way of implementing that with e.g exponential backoff and all that?

materialize-slack/main.go
jshearer: Removed all of them this time I'm pretty sure 😅

materialize-slack/main.go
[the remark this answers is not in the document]
williamhbaker: In that case, should `blocks` still have the `Response_Validated_Constraint_FIELD_REQUIRED` constraint? If it's optional but useful it seems like it should maybe be recommended instead.

materialize-slack/main.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func (t *transactor) Store(it *pm.StoreIterator) (pm.StartCommitFunc, error) {
+
+	for it.Next() {
+		var b = t.bindings[it.Binding]
+		var parsed = buildDocument(b, it.Key, it.Values)
+		var tsStr, tsOk = parsed["ts"].(string)
+		var text, textOk = parsed["text"].(string)
+		var blocks = parsed["blocks"].(json.RawMessage)
```
williamhbaker, line 250: I _think_ this is ok if `blocks` isn't present in the output from `buildDocument`? `nil` can be coerved into a `json.RawMessage` (which is a []byte)...and later `string([]byte(nil))` I assume is `""`? I'm not 100% sure on this so if you haven't tried it, it might not be a bad idea to test it out.

materialize-slack/main.go
```diff
@@ -0,0 +1,300 @@
...
+		if err := json.Unmarshal([]byte(escaped_blocks), &blocksParsed); err != nil {
+			return nil, fmt.Errorf("invalid blocks value %q, %q: %w", reflect.TypeOf(blocks), string(blocks), err)
+		}
+
+		// Accept messages from at most 10 minutes in the past
+		if time.Since(ts).Minutes() < 10 {
+			if err := t.api.PostMessage(b.resource.Channel, text, blocksParsed.BlockSet, b.resource.SenderConfig); err != nil {
+				return nil, fmt.Errorf("error sending message: %w", err)
+			}
+			time.Sleep(time.Second * 10)
```
williamhbaker, line 286: I'm not sure about this 10 second sleep?  It seems like if there are messages to be posted they should be posted without delay.

materialize-slack/main.go
```diff
@@ -0,0 +1,300 @@
...
+		} else {
+			log.Warn(fmt.Sprintf("Ignoring message from the past: %q", ts))
```
williamhbaker, line 288: Having this be logged at anything other than `debug` is going to be super spammy if this gets pointed at a big collection of historical documents. Also consider using structured logging along the lines of `log.WithField("ts", ts).Debug("ignoring message from the past")`

materialize-slack/slack.go
```diff
@@ -0,0 +1,119 @@
...
+				channels = append(channels, found_channels...)
+				cursor = nextCursor
+			} else {
+				break
+			}
+		}
+
+		api.channelIDs = make(map[string]string)
+		for _, channel := range channels {
+			log.Printf("  + channel %q with id %q", channel.Name, channel.ID)
```
williamhbaker, line 77: This should be logged at some level, and structured: `log.WithFields(log.Fields{...the fields...}).Info("joining channel")`...or something like that

materialize-slack/main.go
williamhbaker: You could probably leverage something like this, from the google sheets materialization: https://github.com/estuary/connectors/blob/0f94d077199f097af12f588dcc491a65e1e922be/materialize-google-sheets/util.go#L305-L327

williamhbaker (approved):
Overall LGTM. Left a few minor comments.

My main uncertainty is with the filtering that the materialization is doing. I understand the reasoning and it sounds like it has been hashed out elsewhere already, so my uncertainty is not blocking. Perhaps the ideal solution would be for the kind of `notBefore` / `notAfter` type of general feature we have discussed elsewhere, and the behavior here will suffice until we have that.

jgraettinger:
:wave: @jshearer We need to land this PR, because it's currently live in our UI but isn't getting updates until it does.

williamhbaker:
👀

materialize-slack/main.go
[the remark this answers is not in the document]
jshearer: Changed

materialize-slack/main.go
jshearer: This is to avoid slamming Slack and hitting the rate-limiter

materialize-slack/main.go
jshearer: narrator: it was not okay. fixed it

jshearer:
Ok, I patched this up and it's now working for the task-failures channel again. Gonna merge

PR #725, opened 2023-05-22
materialize-sql: fix numeric format strings as recommended and handle nullable

williamhbaker:
**Description:**

This PR accomplishes 2 things:
- Originally in 9e6b387 there was a change intended to make string fields formatted as numeric (integer or number) be included in the "recommended" fields. That change inadvertently also cause object and array fields, or indeed ANY field with a single type, to also be recommended. This new version of that change will only recommend fields that map to a `FlatType` of `INTEGER` or `NUMBER` to accomplish the original goal without including other unintended single-type fields.
- Updates the `effectiveJsonTypes` mapping to handle nullable strings formatted as numeric in addition to non-nullable fields of this kind, which closes https://github.com/estuary/connectors/issues/723. These fields will also be recommended by default now.

**Workflow steps:**

Materialize string type fields formatted as numeric, either nullable or required, and see that they are included in the recommended fields by default. Object and array type fields with a single type are not. I tested with a couple of representative SQL materializations (postgres and snowflake) with this `readSchema` to confirm the intended behavior:

```yaml
properties:
  key:
    type: integer
  array:
    type: array
  obj:
    type: object
  stringAsIntegerNotRequired:
    format: integer
    type: string
  stringAsIntegerRequired:
    format: integer
    type: string
  stringAsNumberNotRequired:
    format: number
    type: string
  stringAsNumberRequired:
    format: number
    type: string
  stringIntegerAsIntegerNotRequired:
    format: integer
    type:
      - string
      - integer
  stringIntegerAsIntegerRequired:
    format: integer
    type:
      - string
      - integer
  stringNumberAsNumberNotRequired:
    format: number
    type:
      - string
      - number
  stringNumberAsNumberRequired:
    format: number
    type:
      - string
      - number
required:
  - key
  - stringIntegerAsIntegerRequired
  - stringAsIntegerRequired
  - stringNumberAsNumberRequired
  - stringAsNumberRequired
  - obj
  - array
type: object
```

N/A

**Notes for reviewers:**

The extended checks in `effectiveJsonTypes` are a little verbose but I couldn't think of a more concise way to do it that didn't involve what I thought would be a confusing amount of indirection, so I kept it very explicit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/725)

### DIFF
--- a/materialize-sql/type_mapping.go
+++ b/materialize-sql/type_mapping.go
@@ -112,16 +112,24 @@ func (p *Projection) AsFlatType() (_ FlatType, mustExist bool) {
 }
 
 // effectiveJsonTypes potentially maps the provided JSON types into alternate types for
-// materialization. It currently supports strings formatted as numeric values, either as stand-alone
-// string types or as a string type formatted as numeric + that numeric type. It does not apply to
-// keyed fields.
+// materialization. It currently supports potentially nullable strings formatted as numeric values,
+// either as stand-alone string types or as a string type formatted as numeric + that numeric type.
+// It does not apply to keyed fields.
 func effectiveJsonTypes(projection *pf.Projection) []string {
 	if !projection.IsPrimaryKey && projection.Inference.String_ != nil {
 		switch {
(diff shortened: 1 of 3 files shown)


willdonnelly (approved):
LGTM

PR #726, opened 2023-05-22
filesource: switch to flow protocol

mdibaiee:
**Description:**

- Switch `filesource` and the connectors that use it `source-gcs`, `source-s3`, `source-http-file` and `source-sftp` to flow protocol

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/726)

### DIFF
--- a/filesource/filesource.go
+++ b/filesource/filesource.go
@@ -12,9 +12,12 @@ import (
 	"sync"
 	"time"
 
-	"github.com/estuary/flow/go/parser"
-	"github.com/estuary/flow/go/protocols/airbyte"
 	log "github.com/sirupsen/logrus"
+	pc "github.com/estuary/flow/go/protocols/capture"
+	pf "github.com/estuary/flow/go/protocols/flow"
+	schemagen "github.com/estuary/connectors/go/schema-gen"
+	"github.com/estuary/flow/go/parser"
+	boilerplate "github.com/estuary/connectors/source-boilerplate"
 	"golang.org/x/sync/errgroup"
 )
 
@@ -41,7 +44,7 @@ type Source struct {
 	// given a parser JSON schema it may wish to embed.
 	ConfigSchema func(parserSchema json.RawMessage) json.RawMessage
 	// NewConfig returns a zero-valued Config which may be decoded into.
-	NewConfig func() Config
+	NewConfig func(raw json.RawMessage) (Config, error)
 	// Connect using to the Source's Store using the decoded and validated Config.
 	Connect func(context.Context, Config) (Store, error)
 	// DocumentationURL links to the Source's extended user documentation.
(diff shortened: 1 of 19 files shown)


filesource/filesource.go
```diff
@@ -402,42 +430,16 @@ func (r *reader) makeParseConfig(obj ObjectInfo) *parser.Config {
...
 			return err
 		}
 	}
 
-	// Update and write out state.
-	r.shared.states[r.prefix] = r.state
-	stateWrapper.State.Data = r.shared.states
-
-	if err := r.shared.enc.Encode(stateWrapper); err != nil {
+  if encodedCheckpoint, err := json.Marshal(r.shared.states); err != nil {
```
williamhbaker, line 440: Previously we would aquire the lock for the shared state and update the shared state with the state for this reader before emitting the state. Unless I'm missing something we still need to do that, since I don't see that the shared state is being updated anywhere else.

filesource/filesource.go
```diff
@@ -125,100 +128,115 @@ type connector struct {
...
+		return fmt.Errorf("connecting to store: %w", err)
 	}
 
-	var states = make(States)
-	if args.StateFile != "" {
-		if err = args.StateFile.Parse(&states); err != nil {
-			return fmt.Errorf("parsing state: %w", err)
+	var conn = connector{config: cfg, store: store}
+
+	var states States
```
williamhbaker, line 229: Does this work when `open.StateJson != nil` to unmarshall into? I've hit this before where there's a problem with the `nil` val vs. using `make` to initialize as an empty map.

williamhbaker:
When I try to create a capture through a local stack UI with one of the images built from this, I'm getting an error in the agent post-discovery (actually crashes the agent): `error returned from database: value for domain catalog_name violates check constraint "Must be a valid catalog name"` - I got this with both `source-sftp` and `source-s3`. On the same local stack I can fully publish an image from the `:dev` tag through the UI.

It's not immediately obvious to me looking at the code changes what would cause this. Do you get the same error?

mdibaiee:
@williamhbaker that was an issue I spotted as well, it is because the `RecommendedName` for collection had a trailing slash, I fixed that already 👍🏽

mdibaiee:
@williamhbaker oh interesting... then maybe that's a different problem 👀 
let me try on a local stack

mdibaiee:
@williamhbaker I just tried locally and I was able to publish a capture and can see documents in my collection 🤔

mdibaiee:
@williamhbaker actually I think I was not on latest `master`, let me try there again

williamhbaker:
Nice - that does work for publishing. Anyway, I'm trying to do a really quick sanity check, using the `source-sftp` connector, I'm getting an error with the connector when restarting it after it has processed a file.

I have it configured with a directory named `/inbox` and am getting this:

```json
{
  "_meta": {
    "uuid": "1e0ba91e-fa48-11ed-8400-6b68354aae9b"
  },
  "shard": {
    "kind": "capture",
    "name": "bobCo/sftp2/source-sftp",
    "keyBegin": "00000000",
    "rClockBegin": "00000000"
  },
  "ts": "2023-05-24T15:31:56.701273429Z",
  "level": "warn",
  "message": "parsing state checkpoint: prefix /inbox: expected maxMod > minBound"
}
```

So maybe something is off with the state checkpoint?

mdibaiee:
@williamhbaker good spot! I think this was due to me using `merge_patch: true`, made that `false`.

I also realised we don't really need the `mutex` anymore. The `mutex` was used to avoid a race in writing to stdout, but that is already handled by `stream.Documents` and `stream.Checkpoints`, so we don't need to worry about interleaved stdout writes.

filesource/filesource.go
```diff
@@ -402,42 +422,17 @@ func (r *reader) makeParseConfig(obj ObjectInfo) *parser.Config {
...
 	for _, line := range lines {
-		wrapper.Record.Data = line
-		if err := r.shared.enc.Encode(wrapper); err != nil {
+		if err := r.stream.Documents(r.binding, line); err != nil {
 			return err
 		}
 	}
 
-	// Update and write out state.
 	r.shared.states[r.prefix] = r.state
```
williamhbaker, line 431: I think we still need a lock here for concurrent access of multiple bindings to `r.shared.states`. And along with that, the shared mutex used by all bindings that was previously made by `r.shared.mu = sharedMu` in `Source.Read`.

As an alternative - could an individual binding emit its state checkpoint as a merge patch? As I understand it the state is a [map of "prefixes" to their state](https://github.com/estuary/connectors/blob/cc330e5d80cbdf938c49d9887a623949cbe43668/filesource/state.go#L12). So an individual binding ought to be able to track its own state without any concurrency issues and checkpoint with an appropriate merge patch. I'll caveat by saying I'm not 100% sure if this would work, since I don't fully understand why the merge checkpoint wasn't working before even with the full state object and it works without the merge - but I can see that it does.

williamhbaker (approved):
LGTM % question about needing a lock for `r.shared.states`

filesource/filesource.go
mdibaiee: @williamhbaker Okay I dug a bit deeper into this and used https://github.com/estuary/flow/pull/1060 to troubleshoot this.

Turns out, we emit full state objects, until we are finished with the sweep, then we only emit a state object with `minBound` as its only property. With a merge patch, the other fields stay on, which is not what we want, we want to clear all other properties (that's how the connector expects to receive the checkpoint).

One way to do this, is to have an explicit `null` for all the missing attributes, and that can be achieved by removing `omitempty` json tag from attributes of `State`. This way, we can still emit merge patch to set checkpoint of each prefix independently, but we will also make sure that if our `State` has `nil` values, we actually explicitly send a `null` json value, which in JSON Merge Patch RFC, merging a property with an explicit `null` value just removes that property, exactly what we want.

I tested this, seems to work as expected now, see my last commit 👍🏽 

```
Merge Patch for Checkpoint: {
  "daves-test-bucket-1234562/": {
    "complete": true,
    "maxBound": "2023-05-26T13:25:49Z",
    "maxMod": "2023-05-26T13:23:54Z",
    "minBound": "0001-01-01T00:00:00Z",
    "path": "daves-test-bucket-1234562/Test Spreadsheet - Sheet1.csv"
  }
}
Checkpoint: {
  "daves-test-bucket-1234562/": {
    "complete": true,
    "maxBound": "2023-05-26T13:25:49Z",
    "maxMod": "2023-05-26T13:23:54Z",
    "minBound": "0001-01-01T00:00:00Z",
    "path": "daves-test-bucket-1234562/Test Spreadsheet - Sheet1.csv",
    "records": 5
  }
}
{"ts":"2023-05-26T13:26:20.807749959+00:00","level":"warn","message":"\u001b[36mINFO\u001b[0m[0002] completed sweep of daves-test-bucket-1234562/ from 0001-01-01 00:00:00 +0000 UTC through 2023-05-26 13:25:49 +0000 UTC"}
Merge Patch for Checkpoint: {
  "daves-test-bucket-1234562/": {
    "minBound": "2023-05-26T13:23:54Z"
  }
}
Checkpoint: {
  "daves-test-bucket-1234562/": {
    "complete": true,
    "maxBound": "2023-05-26T13:25:49Z",
    "maxMod": "2023-05-26T13:23:54Z",
    "minBound": "2023-05-26T13:23:54Z",
    "path": "daves-test-bucket-1234562/Test Spreadsheet - Sheet1.csv",
    "records": 5
  }
}
```

PR #727, opened 2023-05-23
replace source-hello-world with source-nativehello

mdibaiee:
**Description:**

- Remove the old `source-hello-world` which was a mess and was using airbyte protocol with `source-nativehello` which is cleaner and uses our own protocol. Bump the `VERSION` file to `v2`

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/727)


source-hello-world/Dockerfile
```diff
@@ -11,19 +9,20 @@ WORKDIR /builder
...
 
 # Build the connector.
-RUN go build -o ./connector -v ./source-hello-world/...
+RUN go build -o ./connector -v ./source-hello-world
 
 
 # Runtime Stage
 ################################################################################
-FROM ${BASE_IMAGE}
+FROM ghcr.io/estuary/base-image:v1
```
williamhbaker, line 25: We do the whole `ARG BASE_IMAGE=...` at the top and then `FROM ${BASE_IMAGE}` thing in all (?) the other connector dockerfiles so it might be best to keep doing that here in case we ever want to change the base image.

source-boilerplate/testing/testing.go
```diff
@@ -284,6 +284,9 @@ func (a *pullAdapter) Send(m *pc.Response) error {
 			var binding string
 			if idx := doc.Binding; int(idx) < len(a.bindings) {
 				binding = string(a.bindings[idx].Collection.Name)
+				if len(binding) == 0 {
```
williamhbaker, line 287: I created the tests for `source-nativehello` as part of the great protocol refactoring from several weeks ago primarily as a means for verifying that the boilerplate changes worked. At this point I don't think the `source-nativehello` tests are very valuable anymore - we have many other connectors that run tests which exercise the boilerplate code paths. So I think I'd rather just remove `source-hello-world/main_test.go` and back this change out here that accounts for a capture spec with zero bindings since it only applies for those tests.

williamhbaker (approved):

PR #728, opened 2023-05-23
source-kinesis: switch to flow protocol

mdibaiee:
**Description:**

- Switch `source-kinesis` to flow protocol and use `source-boilerplate`
- Pull-request is based on #726 

**Workflow steps:**

- Create a capture for Kinesis and generate data to be sent to it

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/728)

### DIFF
--- a/source-boilerplate/boilerplate.go
+++ b/source-boilerplate/boilerplate.go
@@ -384,34 +384,34 @@ func hwHashPartition(partitionId []byte) uint32 {
 	return uint32(highwayhash.Sum64(partitionId, highwayHashKey) >> 32)
 }
 
-func RangeIncludesHwHash(range_ *pf.RangeSpec, partitionID []byte) bool {
+func RangeIncludesHwHash(r *pf.RangeSpec, partitionID []byte) bool {
 	var hashed = hwHashPartition(partitionID)
-	return hashed >= range_.KeyBegin && hashed <= range_.KeyEnd
+	return RangeIncludes(r, hashed)
 }
 
 func RangeIncludes(r *pf.RangeSpec, hash uint32) bool {
 	return hash >= r.KeyBegin && hash <= r.KeyEnd
 }
 
-// RangeOverlap is the result of checking whether one Range overlaps another.
-type RangeOverlap int
+// RangeContain is the result of checking whether one Range contains another.
+type RangeContain int
 
 const (
-	NoRangeOverlap      RangeOverlap = 0
-	PartialRangeOverlap RangeOverlap = 1
-	FullRangeOverlap    RangeOverlap = 2
+	NoRangeContain      RangeContain = 0
+	PartialRangeContain RangeContain = 1
+	FullRangeContain    RangeContain = 2
 )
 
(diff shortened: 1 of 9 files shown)


source-kinesis/capture.go
```diff
@@ -55,8 +58,9 @@ func readStream(ctx context.Context, shardRange airbyte.Range, client *kinesis.K
 type streamReader struct {
 	client             *kinesis.Kinesis
 	ctx                context.Context
+	idx                int
```
willdonnelly, line 61: I _think_ this is the binding index, right? It might be more clear to call this 'bindingIndex' or something which makes the purpose clearer to the reader.

source-boilerplate/boilerplate.go
```diff
@@ -384,7 +384,52 @@ func hwHashPartition(partitionId []byte) uint32 {
...
+	PartialRangeOverlap RangeOverlap = 1
+	FullRangeOverlap    RangeOverlap = 2
+)
+
+func RangesOverlap(rangeOne *pf.RangeSpec, rangeTwo *pf.RangeSpec) RangeOverlap {
+	var includesBegin = RangeIncludes(rangeOne, rangeTwo.KeyBegin)
+	var includesEnd = RangeIncludes(rangeOne, rangeTwo.KeyEnd)
+
+	if includesBegin && includesEnd {
+		return FullRangeOverlap
```
willdonnelly, line 410: Just going off the name of this function I think this is incomplete, there should be a "Full Overlap" case if range one is fully contained within range two, but in that case the begin/end keys of range two will not be within range one so the current checks will report "No Overlap".

If the current semantics are intentional the function should probably be something like `RangeContained`.

source-boilerplate/boilerplate.go
```diff
@@ -384,7 +384,52 @@ func hwHashPartition(partitionId []byte) uint32 {
 	return uint32(highwayhash.Sum64(partitionId, highwayHashKey) >> 32)
 }
 
-func IncludesHwHash(range_ *pf.RangeSpec, partitionID []byte) bool {
+func RangeIncludesHwHash(range_ *pf.RangeSpec, partitionID []byte) bool {
 	var hashed = hwHashPartition(partitionID)
 	return hashed >= range_.KeyBegin && hashed <= range_.KeyEnd
```
willdonnelly, line 389: Since `RangeIncludes` is being added I think this function could be simplified to `return RangeIncludes(range_, hwHashPartition(partitionID))`

willdonnelly (approved):
LGTM

source-boilerplate/boilerplate.go
mdibaiee: I copied this over from `airbyte` library in `estuary/flow` repository: https://github.com/estuary/flow/blob/master/go/protocols/airbyte/partition_range.go#L131-L144

So I would rather not change the code since there is code depending on this logic as it is. I can rename it though 👍🏽

PR #732, opened 2023-05-24
materialize-snowflake: separate MERGE and COPY INTO queries for stores

williamhbaker:
**Description:**

For standard updates materializations Flow uses load queries to determine which keys already exist in the database, and this information is available to the connector during the store phase.

COPY INTO queries are much more efficient than MERGE queries, and we can use the knowledge of which keys already exist when storing new records to issue separate COPY INTO vs. MERGE queries within the same transaction for files staged as new records vs. updates.

The hope is that for cases where a standard updates materialization is storing mostly new rows with relatively few updates in each transaction that this will reduce the time for store phases to complete. On such materializations of large data sets we have observed that the MERGE queries take a long time to complete, and the goal here is to reduce the amount of data processed by the MERGE query and reduce overall time of the store phase.

**Workflow steps:**

N/A

**Notes for reviewers:**

It is difficult to do realistic benchmarking for the scale of data that this change would be beneficial for. I did see modest improvements (~15% time savings) when doing a test materialization with 20 million records into a table where 10 million of those records already existed. I hope that the benefit is greater for larger tables where there is a higher ratio of inserts to updates, which are more like _almost_ delta updates with just a few actual updates sprinkled in.

For testing, I ran the integration tests manually with the new connector image and also verified that the resulting 20 million row tables mentioned above for benchmarking were identical when produced by the original and this new connector version.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/732)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -320,10 +320,10 @@ type binding struct {
 	}
 	// Variables accessed by Prepare, Store, and Commit.
 	store struct {
-		stage     *stagedFile
-		mergeInto string
-		copyInto  string
-		mustMerge bool
+		insertStage *stagedFile
+		mergeStage  *stagedFile
+		mergeInto   string
+		copyInto    string
 	}
 }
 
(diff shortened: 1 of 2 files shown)


willdonnelly (approved):
LGTM

Issue #734, opened 2023-05-25
Materialize-Snowflake not handling Format: Date


dyaffe:
**Problem**
Materialize snowflake seems to only handle `format: date-time` and not `format: date`.  A client has noticed and requested that we fix it.

PR #735, opened 2023-05-25
materialize-snowflake: handle "date" formatted strings

williamhbaker:
**Description:**

Handle strings with format `date` as Snowflake [`DATE` type](https://docs.snowflake.com/en/sql-reference/data-types-datetime#date) columns.

Fortunately this is a very simple change since the text encoding that Snowflake accepts for `DATE` types matches Flow's `date` format, so it is fully backward-compatible. 

I tested backward-compatibility for this by:
1. creating a materialization with `date`s on the current connector (`date` column(s) created as  in Snowflake with type `STRING`)
2. materializing some `date`-formatted string values
3. changing the image of the connector to the new one, built with this change
4. materializing some more `date`-formatted string values and observing that the materialization still worked (values still materialized into `STRING` columns)

A new materialization will create the columns as `DATE` types in Snowflake tables and materialize the values as `DATE`s via Snowflake parsing the encoded strings as `DATE` instead of `STRING`.

Closes https://github.com/estuary/connectors/issues/734

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/735)

### DIFF
--- a/materialize-snowflake/sqlgen.go
+++ b/materialize-snowflake/sqlgen.go
@@ -60,6 +60,7 @@ var snowflakeDialect = func() sql.Dialect {
 		sql.STRING: sql.StringTypeMapper{
 			Fallback: sql.NewStaticMapper("STRING"),
 			WithFormat: map[string]sql.TypeMapper{
+				"date":      sql.NewStaticMapper("DATE"),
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

PR #739, opened 2023-05-29
materialize-sql: validate required fields in collection projections

mdibaiee:
**Description:**

- validate that required fields are marked as `Exists.MUST` in projections during `Validate` and `Apply`

**Workflow steps:**

- Validate a capture with a key that's optional according to document schema of the collection

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/739)

### DIFF
--- a/materialize-sql/endpoint.go
+++ b/materialize-sql/endpoint.go
@@ -184,7 +184,7 @@ func resolveResourceToExistingBinding(
 	} else if loadedBinding != nil {
 		constraints = ValidateMatchesExisting(resource, loadedBinding, collection)
 	} else {
-		constraints = ValidateNewSQLProjections(resource, collection)
+		constraints, err = ValidateNewSQLProjections(resource, collection)
 	}
 
 	return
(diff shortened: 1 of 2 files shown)


williamhbaker (commented):
For new materializations of collections this will ensure that required fields are not nullable in the collection schema, but not for changes to collections that are already being materialized: In `ValidateMatchesExisting`, a field that is already being materialized has its constraint set to `LOCATION_RECOMMENDED` which effectively clobbers the classification for field/location required: https://github.com/estuary/connectors/blob/d7c80eb163dbdfa5e2dce0405e9a77bb5c1c8b20/materialize-sql/validate.go#L130-L133

I believe this change overall is geared toward making SQL materializations not allow nullable keys. It might make it easier to have that be a more explicit requirement which could be enforced higher up in the validation logic. At the least I think `ValidateMatchesExisting` will need to be updated to prevent existing collections from being modified to bypass the "required constraints must not be nullable in the collection schema" check.

williamhbaker (approved):
LGTM

PR #743, opened 2023-06-01
source-sqlserver: Fix primary-key discovery query

willdonnelly:
**Description:**

Previously the query would misbehave in the presence of other non-primary-key constraints on the table (such as UNIQUE or foreign key constraints) because it simply wasn't checking for the 'PRIMARY KEY' constraint type specifically.

This commit fixes that bug and adds a test case which previously would have failed with a `primary key ordering failure` message.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/743)


williamhbaker (approved):
LGTM

Issue #709, opened 2023-05-10
materialize-snowflake high memory use


williamhbaker:
After implementing https://github.com/estuary/connectors/pull/706, we can now run arbitrarily large transactions with snowflake.

Unfortunately the memory use when reading results from Snowflake for Load queries is very high, even resulting in OOM kills of the connector when many keys are loaded. It looks like most of this is coming from the `gosnowflake` driver itself in the form of lots and lots of allocations.

To work around this, https://github.com/estuary/connectors/pull/708 has set the `GOMEMLIMIT` variable to `900mb`. This apparently causes the GC to run more actively and allow for large load transactions while keeping the memory use of the connector right around 900mb.

This is far from ideal, since 900mb of memory is quite a lot and we'd prefer to not use that much.

Options to improve this might include:
- Shelling out to `snowsql`, which can run queries as arguments and output results in JSON. I tested this and found its memory usage stayed very low. 
- Using the [Snowflake SQL REST API](https://docs.snowflake.com/en/developer-guide/sql-api/index) to run the load query, and implementing more efficient downloading and decoding of the results. This would probably be a lot of work and we'd need to handle all the different cloud providers' storage systems.
- Improving the `gosnowflake` driver memory usage upstream.

PR #745, opened 2023-06-02
materialize-snowflake: use streaming json downloader for loads

williamhbaker:
**Description:**

The default behavior of the Snowflake driver is to load & parse query results as arrow batches. This causes relatively high memory usage since they are loaded in large chunks and processed as a batch of columns.

This commit changes the connector to use row-oriented JSON results instead. This is accomplished with the parameter `"GO_QUERY_RESULT_FORMAT": "json"`. Furthermore, the Snowflake driver allows for the use of a "stream downloader" with JSON results which enables streaming load results back to the runtime in a memory-efficient way.

The main caveat when using JSON instead of arrow is (alleged) reduced performance. Although worse performance is plausible, this should not have any meaningful impact on our materialization throughput overall. We run loads in parallel with stores, and the merges that accompany loaded results take a far longer time than parsing the loads.

Closes https://github.com/estuary/connectors/issues/709

**Workflow steps:**

Use `materialize-snowflake` for large transactions and observe that it does not use much memory.

In practice I have seen that the connector memory usage is dramatically reduced, where previously it would be pegged out at ~900MB (via `GOMEMLIMIT`, otherwise it would happily spike into multiple GBs) for the entirety of reading the rows iterator, it now remains at 50-100MB, reflecting the fact that little if any data is being buffered in memory.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/745)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -56,6 +56,7 @@ func (c *config) ToURI() string {
 
 func (c *config) asSnowflakeConfig() sf.Config {
 	var maxStatementCount string = "0"
+	var json string = "json"
 	return sf.Config{
 		Account:   c.Account,
 		Host:      c.Host,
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM!

PR #747, opened 2023-06-05
source-postgres: Re-enable 'user has REPLICATION' check

willdonnelly:
**Description:**

This was previously disabled because it broke captures from Amazon RDS, due to how they use the `rds_replication` role instead of `REPLICATION` (and don't bother to also tweak the `rolreplication` column of `pg_roles` accordingly).

The fix is to add an explicit fallback check for membership in the `rds_replication` role.

As far as I know the new logic should work on all major managed SQL providers, but I haven't actually tested that. We can either spend several hours on some manual testing to make sure this works, which I'm not prepared to do right this moment but will get to sometime soon, or we could just YOLO merge it and then if we see any unexpected task failure notifications revert it again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/747)

### DIFF
--- a/source-postgres/prerequisites.go
+++ b/source-postgres/prerequisites.go
@@ -21,7 +21,7 @@ func (db *postgresDatabase) SetupPrerequisites(ctx context.Context) []error {
 
 	for _, prereq := range []func(ctx context.Context) error{
 		db.prerequisiteLogicalReplication,
-		//db.prerequisiteReplicationUser,
+		db.prerequisiteReplicationUser,
 		db.prerequisiteReplicationSlot,
 		db.prerequisitePublication,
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM. We've got decent task failure notifications now so that seems reasonable to keep an eye on for unexpected failures.

willdonnelly:
Yeah, I'm fairly confident that this change will work on RDS and not adversely impact anything else, so I'm going to go ahead and merge it now and just keep an eye on task failures in case I'm wrong.

The error message `user %q must have the REPLICATION role` is what we'll see in the task failures if this check starts misbehaving and producing false negatives.

PR #748, opened 2023-06-05
materialize-sql: option for ignoring string formatting via field config

williamhbaker:
**Description:**

Adds an escape hatch for materializations sourcing from collections that have string formats where the materialization was created before support for materializing the string formats as specific SQL types was added. This can be used for backwards compatibility. I only know of a single field in a single materialization that needs this in order to run on the latest image of its materialization.

There is a broader concept of our SQL materializations providing backwards compatibility for dialect changes, and this is certainly not that. That will be a hard thing to do and there is not sufficient motivation for figuring out how to do it well right now. This solution is a bit ugly, but is the most minimal change I could come up with that would accomplish the immediate need and shouldn't lock us in to anything in case we want to figure out how to do it "right" in the future.

**Workflow steps:**

Set up a materialization with a field configuration like this for a string field formatted as a numeric and see that it is actually materialized as a string, emulating old behavior:

```yaml
...
        fields:
          recommended: true
          include:
            stringAsNumberNotRequired:
              ignoreStringFormat: true

```

N/A

**Notes for reviewers:**

I tested this manually using the above field configuration and it worked as expected, with the field materialized as a string.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/748)

### DIFF
--- a/materialize-sql/table_mapping.go
+++ b/materialize-sql/table_mapping.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/estuary/flow/go/protocols/fdb/tuple"
 	pf "github.com/estuary/flow/go/protocols/flow"
+	log "github.com/sirupsen/logrus"
 )
 
 // TablePath is a fully qualified table name (for example, with its schema).
@@ -116,6 +117,10 @@ func (t *Table) KeyPtrs() []*Column {
 	return out
 }
 
+type fieldConfigOptions struct {
+	IgnoreStringFormat bool `json:"ignoreStringFormat"`
+}
+
 // ResolveTable maps a TableShape into a Table using the given Dialect.
 func ResolveTable(shape TableShape, dialect Dialect) (Table, error) {
 	var table = Table{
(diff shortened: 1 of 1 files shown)


materialize-sql/table_mapping.go
```diff
@@ -137,6 +142,31 @@ func ResolveTable(shape TableShape, dialect Dialect) (Table, error) {
...
+			if err := json.Unmarshal(col.RawFieldConfig, &options); err != nil {
+				return Table{}, fmt.Errorf("unmarshaling RawFieldConfig for column %s of %s: %w", col.Field, shape.Path, err)
+			} else if options.IgnoreStringFormat {
+				log.WithFields(log.Fields{
+					"collection": table.Source.String(),
+					"field":      col.Field,
+					"format":     col.Projection.Inference.String_.Format,
+				}).Info("ignoring string format for field")
+
+				col.Projection.Inference.String_.Format = ""
```
willdonnelly, line 166: By my reading it's possible for `col.Projection.Inference.String_` to be nil, at least according to the type system. And if that happens when `IgnoreStringFormat` is set this code will panic. That seems like a reasonable response IMO to that corner case, but I wanted to make sure it's intentional.

willdonnelly (approved):
LGTM

materialize-sql/table_mapping.go
williamhbaker: You are right that `col.Projection.Inference.String_` can be `nil` if the types don't include `string`, and that would result in a panic if the field config is set in this way. That kind of failure wouldn't be very user-friendly, but considering the expected usage of this option I'm not too worried about it.

PR #752, opened 2023-06-06
Skip some SQL Capture tests in CI builds to reduce flakiness

willdonnelly:
**Description:**

Skips tests using the `RestartingBackfillCapture` helper function in CI builds, because after several attempts they're still flaky due to some sort of nondeterminism.

Also I threw in a minor error-message tweak that came up earlier today while helping a user get a capture going.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/752)

### DIFF
--- a/source-mysql/datatype_test.go
+++ b/source-mysql/datatype_test.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

PR #753, opened 2023-06-06
source-postgres: Add 'SSL Mode' advanced config option

willdonnelly:
**Description:**

Adds a new 'SSL Mode' advanced config option. This is an enumeration which gets directly plumbed through into a libpq connection URI 'sslmode' parameter, which should have the appropriate effect when set.

When this option is left unset the connector behavior should be unchanged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/753)


source-postgres/main.go
```diff
@@ -127,6 +128,11 @@ func (c *Config) Validate() error {
 			}
 		}
 	}
+	if c.Advanced.SSLMode != "" {
+		if !stringMatches(c.Advanced.SSLMode, []string{"disable", "allow", "prefer", "require", "verify-ca", "verify-full"}) {
```
williamhbaker, line 132: I don't see "allow" in the enum list in the config

williamhbaker (approved):
LGTM

source-postgres/main.go
willdonnelly: Good catch. I was going back and forth on whether to include that one because it's so amazingly worthless as a client option, and I guess I forgot to update one or the other location. I'll go ahead and include it everywhere for completeness.

Issue #749, opened 2023-06-06
mongo capture runtime error: The $changeStream stage is only supported on replica sets


williamhbaker:
We saw this error at runtime with a mongo capture:

`runTransactions: readMessage: change stream on collection <COLLECTION>: (Location40573) The $changeStream stage is only supported on replica sets`

Theoretically this shouldn't happen, as it is something that we should (?) be validating during publishes, since we [ping the primary](https://github.com/estuary/connectors/blob/ea710cdc047e59ffaa9a71c9c5e0a774a14a9e77/source-mongodb/main.go#L103) and this should confirm that replication is enabled.

So it may be possible that something changed on the user's end between publishing and running the capture, or it may be that pinging the primary is possible even if replication is disable & we should check for that specifically.

PR #755, opened 2023-06-09
source-mongodb: validate change stream creation

williamhbaker:
**Description:**

If ReplicaSet is not enabled a MongoDB deployment, the backfill operation will work but then the connector will fail when trying to create a change stream. This adds verification that a change stream can be created during the `Validate` RPC.

The only way I could find to make sure that a change stream can be created is to actually try to create a change stream, so this validation involves creating one and then immediately closing it for each captured mongodb collection. This will add a little bit of overhead to each `Validate` call but it seems pretty insignificant in my testing.

Closes https://github.com/estuary/connectors/issues/749

**Workflow steps:**

If you try to set up a capture on a mongodb instance without ReplicaSet enabled (cannot create a change stream), you'll see an error like this in the UI:

Screenshot 2023-06-09 at 11 56 27 AM

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/755)

### DIFF
--- a/source-mongodb/main.go
+++ b/source-mongodb/main.go
@@ -187,6 +187,23 @@ func (d *driver) Validate(ctx context.Context, req *pc.Request_Validate) (*pc.Re
 			return nil, fmt.Errorf("could not find collection %s in database %s", res.Collection, db.Name())
 		}
 
+		// Ensure a change stream can be initialized on each collection. This verifies that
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

PR #757, opened 2023-06-09
materialize-snowflake: include application identifier

williamhbaker:
**Description:**

Adds the "application" parameter to the `materialize-snowflake` connection details, which is in the form of `$tenant_EstuaryFlow`.

Conceptually this is a simple change since the connector gets the name of the task in all requests it receives, and the `tenant` value is the first part of that task name. Then it's just a matter of inserting that `tenant` value into the connection parameters every time the connection is created.

It was a bit more complicated to wire this all up, since the endpoint-specific logic for the materialization is underneath everything that `materialize-sql` does. The led to quite a bit of mechanical refactoring to plumb the actual `tenant` name down into where it needs to be for setting up the connection. In the end, all sql materializations will end up with this value available to them, although for now `materialize-snowflake` is the only one that has any use for it.

Closes https://github.com/estuary/connectors/issues/756

**Workflow steps:**

No user impact. I did verify that the DSN includes the `account` query parameter with this change.

N/A

**Notes for reviewers:**


mdibaiee (approved):
LGTM

PR #758, opened 2023-06-12
materialize-postgres: pgx batches and temp table improvements

williamhbaker:
**Description:**

The main effort here is to return to using pgx batches instead of the custom pipeline code and re-using temporary tables within the same transaction instead of dropping and re-creating them every time. We no longer need the custom pipeline since we don't require pgBouncer support for larger numbers of connections from many materializations into a single database like we thought would have our reporting system, and we have found that re-creating many temporary tables leads to disk usage bloat.

There is also some other miscellaneous cleanup that I am doing here. See individual commit messages.

**Workflow steps:**

N/A

**Notes for reviewers:**

We should change the ops catalog in the `flow` repo to connect the stats materialization directly to the database rather than pgBouncer before merging this, since presumably connecting via pgBouncer will stop working once this is deployed. See https://github.com/estuary/flow/pull/1082

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/758)


materialize-postgres/driver.go
```diff
@@ -356,90 +348,88 @@ func (t *transactor) addBinding(ctx context.Context, target sql.Table) error {
...
 		if converted, err := b.target.ConvertKey(it.Key); err != nil {
 			return fmt.Errorf("converting Load key: %w", err)
 		} else {
-			batch.queueParams(nil, b.execLoadInsertSQL, converted...)
-		}
-
-		if len(batch.buf) > 1024*1024 {
-			if err := batch.roundTrip(ctx, d.load.conn.PgConn()); err != nil {
-				return err
-			}
```
mdibaiee, line 387: Is it okay that we now will create batches larger than 1Mb?

mdibaiee (approved):
LGTM in general, I have one comment about the 1Mb check we used to have

materialize-postgres/driver.go
williamhbaker: That is a good point, I had for some reason assumed that under the hood pgx managed those between the server to keep from building up a huge buffer, but it definitely doesn't. So I'll need to revise this to keep memory usage from blowing up during large txns.

williamhbaker:
Added periodic flushing of batched queries within the load and store postgres transactions to limit memory use. The batch size limits are somewhat arbitrary since I couldn't find a good way to tabulate how much space the batched queries actually take up, but they seemed to work pretty well in my local testing for materializing ~realistic datasets in large flow transactions.

PR #760, opened 2023-06-13
materialize-google-sheets: expand retries to cover 502/503

jgraettinger:
Expand the request retry logic of `materialize-google-sheets` to cover 502 / 503 errors as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/760)

### DIFF
--- a/materialize-google-sheets/util.go
+++ b/materialize-google-sheets/util.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net/http"
 	"regexp"
 	"time"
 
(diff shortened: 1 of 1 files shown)


materialize-google-sheets/util.go
```diff
@@ -312,9 +313,9 @@ func batchRequestWithRetry(
 		}
 
 		// If we encounter a rate limit error, retry after exponential back-off.
-		if e, ok := err.(*googleapi.Error); ok && e.Code == 429 {
+		if e, ok := err.(*googleapi.Error); ok && retryableErrorCodes[e.Code] {
 			var dur = time.Duration(1<<attempt) * time.Second
-			log.Println("rate limit exceeded. backing off for ", dur)
+			log.Println("received error ", e.Code, http.StatusText(e.Code), ", retrying after ", dur)
```
willdonnelly, line 318: As written this will produce a kind of janky-looking log message like `received error  429 Too Many Requests , retrying after  1s`.

Just for neatness I'd recommend using a proper format string, something like `log.Printf("received error %d (%s), retrying after %s", e.Code, http.StatusText(e.Code), dur)` would read a lot nicer IMO.

willdonnelly (approved):
LGTM

PR #762, opened 2023-06-14
filesource: log errors from parser as last logged line

williamhbaker:
**Description:**

When a `filesource` capture is running the parser and the parser fails, the parser logs an error message and exits with status 1. Currently we log something like `"parser failed: exited with status 1"` from the connector as the last log line when this happens.

This changes the connector to not log anything extra when the parser process exits with an error. The responsibility is on the parser to log the details of the error, and the connector shouldn't clobber that will a less useful generic log. This allows the relevant parser error to be picked up as the shard failure reason and shown in the UI.

See also https://github.com/estuary/flow/pull/1084 for a change in the parser to make the message logged on failure include the details of the error.

**Workflow steps:**

Observe better errors shown in the UI when a `filesource` capture fails due to a problem parsing.

N/A

**Notes for reviewers:**

The first commit is general cleanup and formatting churn in `filesource.go`. The second commit has the actual changes discussed here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/762)

### DIFF
--- a/filesource/filesource.go
+++ b/filesource/filesource.go
@@ -5,18 +5,17 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"regexp"
 	"strings"
 	"time"
 
-	log "github.com/sirupsen/logrus"
+	schemagen "github.com/estuary/connectors/go/schema-gen"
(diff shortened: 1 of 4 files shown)


willdonnelly (approved):
LGTM

PR #764, opened 2023-06-15
materialize-pinecone: return error message from rate limit error message response

williamhbaker:
**Description:**

When a 429 "Too Many Requests" response is received from the OpenAI API, we retry since this is usually a rate limit error that exponential backoff while remedy. It is also possible for this response code to be received if a quota is exceed.

Either way, the API response contains some useful information in its JSON body, and we can extract that and return the API error message in the connector error.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/764)

### DIFF
--- a/materialize-pinecone/client/client.go
+++ b/materialize-pinecone/client/client.go
@@ -52,6 +52,13 @@ func NewOpenAiClient(embeddingModel string, org string, apiKey string) *OpenAiCl
 	}
 }
 
+type openAiEmbeddingsError struct {
+	Error struct {
+		Message string `json:"message"`
+		Type    string `json:"type"`
+	} `json:"error"`
+}
+
 func (c *OpenAiClient) CreateEmbeddings(ctx context.Context, input []string) ([]Embedding, error) {
 	res, err := withRetry(ctx, func() (*http.Response, error) {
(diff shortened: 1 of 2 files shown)


materialize-pinecone/client/client.go
```diff
@@ -80,6 +87,17 @@ func (c *OpenAiClient) CreateEmbeddings(ctx context.Context, input []string) ([]
 	}
 	defer res.Body.Close()
 
+	if res.StatusCode == http.StatusTooManyRequests {
+		var errorBody openAiEmbeddingsError
+		if err := json.NewDecoder(res.Body).Decode(&errorBody); err != nil {
+			log.WithField("error", err).Warn("could not decode error response body")
+		} else if errorBody.Error.Message != "" {
+			return nil, fmt.Errorf("error creating embeddings: %s", errorBody.Error.Message)
```
mdibaiee, line 95: should we include the status code in the log as well? might be useful

williamhbaker: That does seem like it could be helpful - added the status to the error message.

mdibaiee (approved):
Thanks!

PR #767, opened 2023-06-16
source-sqlserver: Improve primary-key discovery query

willdonnelly:
**Description:**

A user ran into a situation where the `CONSTRAINT_NAME` column was the same for distinct rows in `INFORMATION_SCHEMA.TABLE_CONSTRAINTS`, so the join with `INFORMATION_SCHEMA.KEY_COLUMN_USAGE` on that column resulted in multiple rows when we need there to just be one.

I'm pretty sure that the 6-tuple `{CONSTRAINT,TABLE}_{CATALOG,SCHEMA,NAME}` is properly unique, so screw moderation we'll join on all six columns.

I have not added a unit test for this because I'm not sure how the colliding constraint names came to exist. In my attempt to write a test reproducing this I observed that constraint names are normally something like `PK__TestTabl__3213E83FFF1774D4` with a unique ID, so I suspect that either some sort of explicit primary key setup is required or this only occurs on an older version of SQL Server.

However I think that the problem is fairly well understood and I'm confident that this should fix it even without a test reproducing the behavior. So long as this doesn't break the CI build I think we should merge it, and we'll find out on Monday if that fixed the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/767)

### DIFF
--- a/source-sqlserver/discovery.go
+++ b/source-sqlserver/discovery.go
@@ -138,10 +138,19 @@ func getColumns(ctx context.Context, conn *sql.DB) ([]sqlcapture.ColumnInfo, err
 	return columns, nil
 }
 
+// Joining on the 6-tuple {CONSTRAINT,TABLE}_{CATALOG,SCHEMA,NAME} is probably
+// overkill but shouldn't hurt, and helps to make absolutely sure that we're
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #770, opened 2023-06-21
refactor: consolidate "slice contains" logic

williamhbaker:
**Description:**

We have a similar function for determining if a slice contains an element copied in many connectors. This consolidates that into a common function that can be used amongst all connectors.

There are also a few minor cleanups that I noticed while updating these function call sites.

I changed the function signature of our previous standard "slice contains" function to take the slice first and then the element to look for, which is the reverse of how it was. This seems more in line with similar functions, like the one in [golang.org/x/exp](https://pkg.go.dev/golang.org/x/exp/slices#Contains).

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/770)

### DIFF
--- a/go/pkg/slices/slices.go
+++ b/go/pkg/slices/slices.go
@@ -1,0 +1,11 @@
+package slices
+
+func Contains[T comparable](elems []T, v T) bool {
+	for _, elem := range elems {
+		if elem == v {
+			return true
+		}
+	}
+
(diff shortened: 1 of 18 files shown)


mdibaiee (approved):
Thank you!

PR #774, opened 2023-06-22
materialize-sql: add CommitWithDelay function

williamhbaker:
**Description:**

Adds the `CommitWithDelay` for SQL materializations, which can be used to wrap their asynchronous commit operation with an additional delay to spread transactions apart. This has immediate applicability for [Snowflake](https://docs.estuary.dev/reference/Connectors/materialization-connectors/Snowflake/#reducing-active-warehouse-time) and [BigQuery](https://docs.estuary.dev/reference/Connectors/materialization-connectors/BigQuery/#performance-considerations), and may be useful for future situations as well.

The pinned `flow` package is being updated to bring in the changes from https://github.com/estuary/flow/pull/1091, which causes a bit of churn. Other than that, see commit & code comments for more details on `CommitWithDelay` and its usage for `materialize-snowflake` and `materialize-bigquery`.

**Workflow steps:**

Instead of `minTxnDuration`, set `UpdateDelay` on `materialize-snowflake` to spread out transactions to reduce active warehouse time. BigQuery will always use a 2 minute delay, so no additional configuration is needed.

Docs for snowflake and bigquery will need to be updated

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/774)


go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	github.com/benbjohnson/clock v1.3.0
 	github.com/bradleyjkemp/cupaloy v2.3.0+incompatible
 	github.com/elastic/go-elasticsearch/v7 v7.17.1
-	github.com/estuary/flow v0.1.9-0.20230405122715-43db16b1b0cb
+	github.com/estuary/flow v0.2.4-0.20230622152821-5fad88ecf0df
```
williamhbaker, line 22: TODO: Update this if needed after https://github.com/estuary/flow/pull/1091 merges, prior to merging this.

williamhbaker:
The delay is part of the commit so that the load phase for the next transaction can [run asynchronously](https://github.com/estuary/flow/blob/f8d02d6211e8208ccbe81331497ff575175f307f/go/protocols/materialize/transactor.go#L221-L229) while the `OpFuture` that includes the delay is being awaited. If the delay were part of the Store RPC outside of that `OpFuture` it would [block](https://github.com/estuary/flow/blob/f8d02d6211e8208ccbe81331497ff575175f307f/go/protocols/materialize/transactor.go#L260) the next round and hold up processing the next transaction. It's nice to let the processing for the next transaction proceed while the delay is being awaited especially if there is a lot of reduction that can happen for the collection.

mdibaiee (approved):

Issue #773, opened 2023-06-22
source-sqlserver: Filter out included and other non-key columns from secondary index discovery


willdonnelly:
Our current query for enumerating unique secondary indices during discovery isn't doing anything to filter out included columns. We should add a `WHERE ic.key_ordinal != 0` or `WHERE ic.is_included_column = 0` constraint to ignore these because we only care about the index key.

PR #776, opened 2023-06-22
source-sqlserver: Filter included columns from index enumeration

willdonnelly:
**Description:**

When we query the system tables to list potential secondary unique indices for discovery, we need to filter out included non-key columns.

This fixes https://github.com/estuary/connectors/issues/773

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/776)


jgraettinger (approved):
LGTM

PR #777, opened 2023-06-23
source-sqlserver: Fully support `DATETIME` columns

willdonnelly:
**Description:**

Support for the `DATETIME`, `DATETIME2`, and `SMALLDATETIME` columns is added by introducing a new configuration property 'Time Zone' whose value should be an IANA Time Zone name or `-HH:MM` numeric offset, and using that location to interpret the contents of `DATETIME` and friends.

This is basically the same as how we do things in MySQL, except here we're not even trying to autodetect a time zone from the server configuration.

The new time zone property is added at the top level of the configuration so all new users should see it. The default value according to the JSON schema should be `"UTC"` and likewise if the property is empty (as it will be for existing captures) it will default to `"UTC"`.

This fixes https://github.com/estuary/connectors/issues/772

**Workflow steps:**

Existing captures should be unchanged.

New users will no longer be greeted with `error translating column type to JSON schema` warnings when performing discovery, and will be given a hopefully clear input box in the config to change how datetimes are interpreted. If they don't change it from UTC, the captured values will be the same as before (that is, interpreting the YYYY-MM-DD HH:MM:SS.NNN values as UTC) except the schema will at least be useful.

https://docs.estuary.dev/reference/Connectors/capture-connectors/sqlserver/

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/777)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


williamhbaker (approved):
LGTM

PR #778, opened 2023-06-23
source-sqlserver: Support `hierarchyid` column type

willdonnelly:
**Description:**

The `hierarchyid` column type is a weird one. It's an ultra- compact representation of an arbitrary node position within a tree, encoded using an undocumented variable-length binary code of some sort.

When we receive these values from the SQL Server client library they're still in their `[]byte` form, and given that (as far as I can determine) there is no publically available documentation [1] of this binary representation I don't think we should be in the business of decoding them.

So this change amounts to telling our discovery logic that a `hierarchyid` column will be serialized into a base64 string and calling it a day.

This fixes https://github.com/estuary/connectors/issues/771

[1] The best documentation I could find about the actual encoding of this data was https://web.archive.org/web/20210304123429/http://www.adammil.net/blog/v100_how_the_SQL_Server_hierarchyid_data_type_works_kind_of_.html

**Workflow steps:**

One less column type producing scary warnings during discovery.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/778)


williamhbaker (approved):
LGTM

PR #779, opened 2023-06-23
source-postgres: verify that replication slot exists in the configured database

williamhbaker:
**Description:**

Sometimes a replication slot with the name the capture is configured with already exists, but not in the database that the capture is using. This will currently cause the connector to fail after publication with a Postgres error like `replication slot “flow_slot” was not created in this database`. This can happen if a publication is attempted and is able to create the slot but fails for some other reason, and then the user changes their configured database and publishes it again.

With this change we will return a validation error describing this fact when the desired slot name already exists but in another database. The solution is to use a different slot name, or remove the existing slot.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/779)

### DIFF
--- a/source-postgres/prerequisites.go
+++ b/source-postgres/prerequisites.go
@@ -2,10 +2,12 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/estuary/connectors/sqlcapture"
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

PR #784, opened 2023-06-26
source-sqlserver: Quote schema/table names in backfill queries

willdonnelly:
**Description:**

This series of commits consists of a ton of preparatory churn in the test code, followed by a trivial 2-line change to the actual connector logic.

The goal here is just that we ought to be quoting schema and table names when constructing backfill queries, because otherwise a schema named something like `"User"` (which is a reserved word) will result in backfill queries that parse incorrectly and get a confusing error message from the database saying `Invalid usage of the option FIRST in the FETCH statement`.

But testing that required that our test code be able to construct table names with quoted parts like that, which in turn broke the way we set up captures by matching the desired table names to our discovery results, and the simple/easy way to fix that was to cherry-pick some work in progress from the Snowflake CDC branch where this same basic class of problem had also popped up and was fixed via a pretty large-scale refactor of all our test logic to use numeric IDs as a substring of each table name and just match on those IDs.

In this manner, we end up with a >12kLOC pull request for an 8-character functional change to the connector. But the other 12kLOC of changes would have been landed eventually, so I'm happy for that time to be now.

**Workflow steps:**

Captures from SQL Server tables/schemas whose names happen to be reserved words will actually work.

None

**Notes for reviewers:**

See description for a breakdown of why an 8-character functional change adding brackets around table names got blown up into 12kLOC of test churn.

TL;DR: This was a refactor that was already in progress as part of the WIP Snowflake CDC connector, and cherry-picking the whole thing on over and landing it now was the easiest way to enable a minor tweak to the SQL Server test logic.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/784)


williamhbaker (approved):
LGTM

PR #787, opened 2023-07-03
source-mysql: Ignore SAVEPOINT query events as benign

willdonnelly:
**Description:**

I'm unsure why/how `SAVEPOINT` would show up as a query event at all, but it's not inherently a problem. Now a `ROLLBACK` might be a bigger concern, but we'll deal with that if it comes up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/787)


williamhbaker (approved):
LGTM

PR #793, opened 2023-07-07
source-mysql: Disable REPLICATION SLAVE permission check

willdonnelly:
**Description:**

There are a couple of reports that this check may be producing false positives in some setups that should otherwise work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/793)


williamhbaker (approved):
LGTM

dyaffe (commented):
LGTM

PR #802, opened 2023-07-13
Add XID / GTID metadata to Postgres and MySQL captures

willdonnelly:
**Description:**

This PR adds new XID / GTID properties to the source-specific metadata of Postgres / MySQL captures, which will contain the database-assigned transaction ID associated with a particular change. In the case of MySQL GTIDs this property will be omitted if GTID mode isn't enabled on the source database.

Unfortunately deploying this change will probably be a bit of a hassle, as our discovery schemas currently specify `additionalProperties: false` on the `/_meta/source` object, and that means that naively merging this would break all ongoing captures. Our options to resolve this are:

1. Version bump the connector.
2. Modify all production captures and update their schemas to either permit additional properties in that field, or to just add the appropriate transaction-ID property if there's some reason we actually _want_ `additionalProperties: false` semantics there.
3. Some sort of hackery which would only set the `xid` / `gtid` properties if the collection schema contains said properties.

Also we should probably remove `additionalProperties: false` there. I'm not aware of anything important that this protects users from, and it makes otherwise simple changes like this one into a complicated and messy situation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/802)


source-mysql/replication.go
```diff
@@ -148,6 +150,7 @@ type mysqlReplicationStream struct {
 	errCh  chan error         // Error output channel for the replication worker goroutine
 
 	gtidTimestamp time.Time // The OriginalCommitTimestamp value of the last GTID Event
+	gtidString    string    // The GTID value of the last GTID event, formatted as a "<uuid>:<counter>" string.
```
williamhbaker, line 153: The `mysqlSourceInfo` mentions that this is only set if GTIDs are enabled. If GTIDs are not enabled, does this end up in the output document as an empty string, or is it completely omitted?

williamhbaker (approved):
LGTM - based on the discussion earlier today I'm assuming you'll also be changing the schema produced by discovering to remote the `additionalProperties: true` from the metadata object.

source-mysql/replication.go
willdonnelly: Currently it will be completely omitted if GTIDs are disabled.

willdonnelly:
> LGTM - based on the discussion earlier today I'm assuming you'll also be changing the schema produced by discovering to remote the `additionalProperties: true` from the metadata object.

Yeah, this change will be rebased again after I'm done with all the preparatory work.

willdonnelly:
I've updated this PR so that instead of unconditionally setting the transaction ID properties they will now be set _only if the output collection expects them_.

In general this is a terrible idea and we should never be varying capture output based on the schema of the collection it will be going into. But in this particular case it lets us avoid modifying hundreds of collection schemas while still getting this feature for users who want it. And ideally this will be a temporary thing that we can remove in a few months, once we have automatic rediscovery in the event of schema violations and so (hopefully) the majority of those collections will automatically get updated if we flip this on unconditionally.

PR #803, opened 2023-07-13
source-postgres: Ignore 'ParameterStatus' messages

willdonnelly:
**Description:**

These are rare but I believe they might be able to cause the capture to fail just because we don't expect to see that message. They're irrelevant to us, so ignore them.

Also tweaks the 'unexpected message' error to use '%#v' which will hopefully tell us the actual type name so next time I won't have to simply guess what the offending message was.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/803)


williamhbaker (approved):
LGTM

PR #805, opened 2023-07-13
sqlcapture: Permit source metadata to add properties in future

willdonnelly:
**Description:**

See https://github.com/estuary/connectors/pull/802 for the immediate motivation here, we want to add `xid` and `gtid` metadata properties to Postgres and MySQL captures.

Currently the presence of an `additionalProperties: false` clause kind of paints us into a corner where adding new metadata is a really heavyweight operation requiring either a connector version bump or fiddling with all production schemas.

As discussed during standup today, I would like to bite this bullet now by removing the `additionalProperties: false` clause and then modifying collection schemas in production to effectively pretend that clause had never been there.

Since the connector alone controls what properties are going to be present in the metadata object, it changes very rarely, and also in general if we add new metadata we probably want it to be treated as an opt-in nice-to-have thing rather than using the fancy "fail schema validation and rerun discovery and propagate changes" flow we're working on, so just not saying anything and implicitly permitting additional properties seems like the right choice for this specific part of the schema.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/805)

### DIFF
--- a/source-mysql/.snapshots/TestGeneric-KeylessDiscovery
+++ b/source-mysql/.snapshots/TestGeneric-KeylessDiscovery
@@ -88,7 +88,6 @@ Binding 0:
                       "description": "Cursor value representing the current position in the binlog."
                     }
                   },
-                  "additionalProperties": false,
                   "type": "object",
                   "required": [
                     "schema",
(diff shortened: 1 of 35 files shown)


williamhbaker (approved):
LGTM

PR #806, opened 2023-07-13
source-firestore: Tweak discovery limits

willdonnelly:
**Description:**

This commit fiddles with several Firestore discovery details in ways that I hope will make it less likely to OOM on large datasets. This includes:

  - A hard limit on the total number of collection-scanning worker threads that will ever be spawned (currently set to 10k). This should help guard against the theoretical failure mode where a document has a million nested collections under it and we run out of memory simply because of goroutines waiting to run.
  - A new limit on the number of nested collections which will be scanned under any single document. This also helps guard against the 'lots of nested collections' failure mode while making sure we don't rabbit-hole all our discovery work into one small part of the overall tree.
  - Checking the `maxDocumentsPerResource` limit only after acquiring the collection worker thread semaphore.
  - No longer launching a redundant thread inside `handleDocument`, since after removing schema inference all the function really does is loop over subcollections and spawn threads for *those*.
  - Lowering various tuning constants by a factor of 2.

It is my hope that this will make things work better, but really this is all heuristics and hope anyway so I don't have any hard proof. I have verified that the very basic `TestDiscovery` automated test still passes, but that's about all I have done to verify this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/806)

### DIFF
--- a/source-firestore/discovery.go
+++ b/source-firestore/discovery.go
@@ -32,20 +32,28 @@ const (
 	// So instead we can apply some common-sense policy which mimics what a user might
 	// do: check the first N exhaustively and then start skimming through and randomly
 	// picking a document to check every so often.
-	discoverSubresourceMinimum     = 100
+	discoverSubresourceMinimum     = 50
 	discoverSubresourceProbability = 0.01
 
 	// discoverMaxDocumentsPerCollection limits the maximum number of documents which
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #809, opened 2023-07-14
materialize-redshift: handle async upload errors while encoding rows

williamhbaker:
**Description:**

There is a possible deadlock if the uploader goroutine exits with an error in the midst of calls to `encodeRow`. In this case the pipe will never be read to the call to write to the pipe via `(*stagedFile).encoder.Encode` will block indefinitely.

The fix is to close the write end of the pipe on error from the uploader. There is also some handling for routing the error from the uploader goroutine to the call to `encodeRow` so that it can be surfaced as the connector's exit error.

Additionally I have fixed/updated the end-to-end tests, which were used to test this change.

**Workflow steps:**

Run the materialization with the chance of deadlock. This could happen, for example, if the S3 bucket were removed in the middle of a transaction, or made unavailable for some other reason.

N/A

**Notes for reviewers:**

### DIFF
--- a/materialize-redshift/staged_file.go
+++ b/materialize-redshift/staged_file.go
@@ -31,25 +31,27 @@ import (
 // uploading the file, the file location is returned as well as a callback that can be used to
 // delete the created file.
 type stagedFile struct {
-	cols         []*sql.Column
-	client       *s3.Client
-	uploader     *manager.Uploader
-	bucket       string
-	bucketPath   string
-	group        *errgroup.Group
-	uploadOutput *manager.UploadOutput
-	pipeWriter   *io.PipeWriter
-	encoder      *json.Encoder
-	started      bool
+	cols          []*sql.Column
+	client        *s3.Client
+	uploader      *manager.Uploader
+	uploaderErrCh chan (error)
+	bucket        string
+	bucketPath    string
+	group         *errgroup.Group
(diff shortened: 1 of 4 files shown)


materialize-redshift/staged_file.go
```diff
@@ -74,6 +76,16 @@ func (f *stagedFile) start(ctx context.Context) {
...
+			// This send cannot block since we are the only sender on the buffered channel. The
+			// error value sent here may be picked up in a call to encodeRow. Or if the error
+			// results after a flush begins, it will retrieved via group.Wait() from the error
+			// returned below.
+			f.uploaderErrCh <- err
+
+			// Close the writer channel now that the uploader has failed and no further writes to
+			// the pipe will be read. This will cause new & outstanding writes to the pipe to return
+			// a "closed pipe" error.
+			f.pipeWriter.Close()
```
jgraettinger, line 88: this can race with a concurrent use of `pipeWriter`. I think you want to be closing `r` here, which also propagates ErrClosedPipe

https://pkg.go.dev/io#PipeReader.Close

jgraettinger: Also, I think you could use PipeReader.CloseWithError to deliver the causal error to the write loop, rather than having `uploadErrCh` ?

jgraettinger (approved):
LGTM

PR #810, opened 2023-07-14
materialize-redshift: default to VARCHAR(MAX) when no MaxLength annotation is available

williamhbaker:
**Description:**

It is not uncommon for collections to have long strings, and it is currently not common at all for collection schemas to have `maxLength` annotations.

Furthermore, it is not currently practical to expect users to set `maxLength` annotations on schemas discovered from SQL captures etc. as these annotations will be overwritten. This makes materializing long string fields from SQL captures tenuous at best.

The change here is an inverted treatment of redshift string columns: We will create them with the maximum length by default, but respect `maxLength` annotations if they are there for possibly tightening the column limits.

It is possible to alter Redshift columns after the fact to have a shorter length for `VARCHAR`, so excessively long string columns can easily be adjusted after the fact.

**Workflow steps:**

The `materialize-redshift` documentation will need to be updated with this change.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/810)

### DIFF
--- a/materialize-sql/type_mapping.go
+++ b/materialize-sql/type_mapping.go
@@ -328,13 +328,12 @@ func (m ProjectionTypeMapper) MapType(p *Projection) (MappedType, error) {
 	}
 }
 
-// MaxLengthMapper checks if the projection is a STRING type Projection having a MaxLength.
-// If it is, it invokes WithLength with the MaxLength to map the Projection and returns
-// the result. Otherwise, it invokes and returns Fallback.
+// MaxLengthMapper checks if the projection is a STRING type Projection having a MaxLength. If it
+// is, it invokes WithLength to map the Projection and returns the result. Otherwise, it invokes and
+// returns Fallback.
 type MaxLengthMapper struct {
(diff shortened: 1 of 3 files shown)


materialize-sql/type_mapping.go
```diff
@@ -328,13 +328,12 @@ func (m ProjectionTypeMapper) MapType(p *Projection) (MappedType, error) {
 	}
 }
 
-// MaxLengthMapper checks if the projection is a STRING type Projection having a MaxLength.
-// If it is, it invokes WithLength with the MaxLength to map the Projection and returns
-// the result. Otherwise, it invokes and returns Fallback.
+// MaxLengthMapper checks if the projection is a STRING type Projection having a MaxLength. If it
+// is, it invokes WithLength to map the Projection and returns the result. Otherwise, it invokes and
+// returns Fallback.
```
williamhbaker, line 333: I had structured this weirdly before, so this is some cleanup.

PR #811, opened 2023-07-15
source-mysql: Handle zero timestamps with fractional seconds

willdonnelly:
**Description:**

These values occur if you have a column with a type such as `timestamp(6)` and insert the all-zeroes time into them.

We can handle this by just doing a prefix match for the zero `YYYY-MM-DD HH:mm:SS` part, since MySQL won't allow you to insert a value like `0000-00-00 00:00:00.123456` anyway -- the zero month/day are only allowed if it's the zero timestamp, so no information is lost by just doing a simple prefix match and replacing with a constant `0001-01-01T00:00:00Z` instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/811)


williamhbaker (approved):
LGTM

PR #813, opened 2023-07-17
sqlcapture: normalize time values to be RFC3339

mdibaiee:
**Description:**

- Normalize `format: time` values to be RFC3339 times (specifically the timezone needs to be normalized). This pull-request only works when we have https://github.com/estuary/flow/pull/1114 also merged since we currently do not support timezone offsets in `format: time` in our jsonschema validation logic

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/813)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,7 +54,7 @@ require (
 	google.golang.org/genproto v0.0.0-20230131230820-1c016267d619
 	google.golang.org/grpc v1.52.3
 	google.golang.org/protobuf v1.28.1
-	vitess.io/vitess v0.16.2
+	vitess.io/vitess v0.15.3
 )
 
 require (
@@ -130,6 +130,7 @@ require (
 	github.com/klauspost/cpuid/v2 v2.0.9 // indirect
 	github.com/kr/fs v0.1.0 // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
+	github.com/mattn/go-ieproxy v0.0.9 // indirect
 	github.com/mattn/go-isatty v0.0.17 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.4 // indirect
 	github.com/minio/asm2plan9s v0.0.0-20200509001527-cdd76441f9d8 // indirect
--- a/source-postgres/datatype_test.go
+++ b/source-postgres/datatype_test.go
@@ -65,10 +65,14 @@ func TestDatatypes(t *testing.T) {
 
 		// TODO(wgd): The 'timestamp with time zone' type produces inconsistent results between
 		// table scanning and replication events. They're both valid timestamps, but they differ.
-		// {ColumnType: `timestamp with time zone`, ExpectType: `{"type":"string","format":"date-time"}`, InputValue: `'January 8, 1999'`, ExpectValue: `"1999-01-07T16:00:00-08:00"`},
+		// {ColumnType: `timestamp with time zone`, ExpectType: `{"type":["string","null"],"format":"date-time"}`, InputValue: `'January 8, 1999'`, ExpectValue: `"1999-01-07T16:00:00-08:00"`},
 		{ColumnType: `time`, ExpectType: `{"type":["integer","null"]}`, InputValue: `'04:05:06 PST'`, ExpectValue: `14706000000`},
 		{ColumnType: `time without time zone`, ExpectType: `{"type":["integer","null"]}`, InputValue: `'04:05:06 PST'`, ExpectValue: `14706000000`},
-		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06 PST'`, ExpectValue: `"04:05:06-08"`},
+		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06 PST'`, ExpectValue: `"04:05:06-08:00"`},
+		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06 UTC'`, ExpectValue: `"04:05:06Z"`},
+		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06.123 UTC'`, ExpectValue: `"04:05:06.123Z"`},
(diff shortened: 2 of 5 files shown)


jgraettinger:
Can you walk me through how we arrived here, at a schema-based fix-up, instead of having individual connectors be responsible for using valid time formats?

I'm pretty uncomfortable with normalization behaviors which are driven by the target collection schema -- every time we've dabbled with this we've come to regret the cyclic dependency -- and I think that having connector output depend on the collection schema should be a last resort. Especially for our owned connectors where we deeply control their means of output.

mdibaiee:
@williamhbaker this is so much simpler now, I wish I knew this part existed before I got on this!

@jgraettinger I pretty much took the lead of what we do in airbyte-to-flow and tried to do something similar here, although at some point it did occur me it might be a better idea to use our knowledge of the database directly to do this, but I was not sure if that is already available or would need more wiring, but turns out we already have that available. The code is much much simpler now.

go.mod
```diff
@@ -19,9 +19,10 @@ require (
 	github.com/benbjohnson/clock v1.3.0
 	github.com/bradleyjkemp/cupaloy v2.3.0+incompatible
 	github.com/elastic/go-elasticsearch/v7 v7.17.1
-	github.com/estuary/flow v0.2.4-0.20230623123341-b5c714e822e3
+	github.com/estuary/flow v0.3.4-0.20230714153247-33844329ca87
 	github.com/evanphx/json-patch/v5 v5.6.0
 	github.com/go-mysql-org/go-mysql v1.5.0
+	github.com/go-openapi/jsonpointer v0.19.6
```
williamhbaker, line 25: I don't think we still need all of these additions to `go.mod`, so it would be best to revert them unless they are needed.

source-postgres/datatype_test.go
```diff
@@ -68,7 +68,8 @@ func TestDatatypes(t *testing.T) {
 		// {ColumnType: `timestamp with time zone`, ExpectType: `{"type":"string","format":"date-time"}`, InputValue: `'January 8, 1999'`, ExpectValue: `"1999-01-07T16:00:00-08:00"`},
 		{ColumnType: `time`, ExpectType: `{"type":["integer","null"]}`, InputValue: `'04:05:06 PST'`, ExpectValue: `14706000000`},
 		{ColumnType: `time without time zone`, ExpectType: `{"type":["integer","null"]}`, InputValue: `'04:05:06 PST'`, ExpectValue: `14706000000`},
-		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06 PST'`, ExpectValue: `"04:05:06-08"`},
+		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06 PST'`, ExpectValue: `"04:05:06-08:00"`},
```
williamhbaker, line 71: My memory is fuzzy on the specific case we needed to handle - something about fractional seconds? Might be good to add a test case for that if it's not already covered.

williamhbaker (approved):
LGTM with a couple more comments

source-postgres/datatype_test.go
```diff
@@ -68,7 +68,8 @@ func TestDatatypes(t *testing.T) {
 		// {ColumnType: `timestamp with time zone`, ExpectType: `{"type":"string","format":"date-time"}`, InputValue: `'January 8, 1999'`, ExpectValue: `"1999-01-07T16:00:00-08:00"`},
 		{ColumnType: `time`, ExpectType: `{"type":["integer","null"]}`, InputValue: `'04:05:06 PST'`, ExpectValue: `14706000000`},
```
jgraettinger, line 69: @willdonnelly these test cases are odd -- this is outputting a unix timestamp? Why that, as opposed to `04:05:06--8` (and also, we don't know the date, right) ?

mdibaiee: @jgraettinger I think this is because the `ExpectType` is set to `integer`

jgraettinger: yep, I'm asking _why_ are we mapping these into an integer? Seems they should be RFC3339, no?

source-postgres/replication.go
```diff
@@ -420,10 +422,14 @@ func (s *replicationStream) decodeChangeEvent(
 		return nil, fmt.Errorf("error encoding row key for %q: %w", streamID, err)
 	}
 
-	if err := translateRecordFields(nil, bf); err != nil {
+	discovery, ok := s.tables.discovery[streamID]
+	if !ok {
+		return nil, fmt.Errorf("unknown discovery info for stream %q", streamID)
+	}
+	if err := translateRecordFields(discovery, bf); err != nil {
```
mdibaiee, line 429: I found a bug where during replication we were not providing discovery information to `translateRecordFields` (this was not noticed because we did not rely on discovery info in this function up to this point)

source-postgres/datatype_test.go
mdibaiee: @jgraettinger ah sorry. So in this case, the column type is `time`, which is time without timezone.

pgx handles this value as a unix timestamp: https://pkg.go.dev/github.com/jackc/pgtype#Time

Since it doesn't have timezone, to format it as RFC3339 would require us to assign it an arbitrary timezone (UTC being the natural choice), but that is not really accurate, since the time does not really have a timestamp.

jgraettinger: > pgx handles this value as a unix timestamp: https://pkg.go.dev/github.com/jackc/pgtype#Time

Ah -- and to pedantically clarify it's not a unix timestamp but is instead the number of microseconds since midnight. Very different things, but I do follow that that implies a representation like "04:05:06.123456", which also would need to omit a timezone offset (because the timezone is unknown), and is thus not a valid `format: "time"`, and in that sense just spitting out the number of microseconds instead is reasonable.

For other DB's we asses the system's timezone or allow it to be configured and use that to resolve the ambiguity, but this was definitely written before we established that practice. Thanks

willdonnelly: Yeah, this logic predates all of the work on supporting configurable time zones for this sort of timezone-free datatype.

If this were being implemented for the first time today I'd probably favor converting the microseconds count into an RFC3339 time with appropriate time zone conversion, but that's probably beyond the scope of this PR to do now since it would require making sure there aren't any existing uses that would be broken.

PR #819, opened 2023-07-17
Revert "source-mysql: Handle RDS binlog configuration query errors"

willdonnelly:
Reverts estuary/connectors#818

Turns out that we can get indistinguishable failures on RDS and non-RDS setups, so the whole "get binlog expiry time" function will need to be rewritten to start with an explicit "if we're running against RDS, check this, otherwise check that" branch.

Reverting this in the meantime so it's not breaking other non-RDS captures.

### DIFF
--- a/source-mysql/prerequisites.go
+++ b/source-mysql/prerequisites.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


willdonnelly:
YOLO merging as this is just a revert and the previous PR broke at least one otherwise-successful task in production.

PR #821, opened 2023-07-17
source-mysql: Always include `""` as possible ENUM value

willdonnelly:
**Description:**

Apparently MySQL will never actually guarantee that your ENUM column doesn't contain illegal values. Note that this is distinct from nullability, an `ENUM(...) NOT NULL` column can still contain error values which will be represented by the empty string when you retrieve them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/821)


williamhbaker (approved):
LGTM

PR #824, opened 2023-07-18
source-firestore: Recursively sanitize arrays via backfill

willdonnelly:
**Description:**

We have value-sanitization logic (which currently just replaces NaN values with the string "NaN") which is applied to backfilled documents, but previously it wouldn't recurse into lists which made it of limited usefulness.

This rewrite replaces `sanitizeDocument` with `sanitizeValue` (because it makes the logic cleaner once we're handling lists- of-values and maps-of-values both) but other than recursing into lists there should be no functional change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/824)


williamhbaker (approved):
LGTM

PR #825, opened 2023-07-18
source-firestore: Limit number of concurrent backfill queries

willdonnelly:
**Description:**

Adds a semaphore which should ensure that only 2 (at current setting) `BackfillAsync` worker threads can be holding onto chunks of backfilled documents at any given moment. Hopefully this will make memory usage more controllable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/825)

### DIFF
--- a/source-firestore/pull.go
+++ b/source-firestore/pull.go
@@ -40,6 +40,11 @@ const retryInterval = 300 * time.Second
 // Log progress messages after every N documents on a particular stream
 const progressLogInterval = 10000
 
+const (
+	backfillChunkSize   = 1024
+	concurrentBackfills = 2
+)
+
 func (driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) error {
 	log.Debug("connector started")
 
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM
one optional nit pick is that the [x/sync/semaphor](https://pkg.go.dev/golang.org/x/sync/semaphore) package would give you named functions like `.Acquire(1)` instead of `semaphor <- struct{}{}`, which I personally find easier to read. TBH I kinda understand why you'd opt not to add a dependency, though, so 🤷

willdonnelly:
Yeah, we should probably clean up the Firestore connector to use `x/sync/semaphore` in the near future, but there's already a similar use of channel-as-semaphore in the discovery logic and I didn't want to expand the scope of this PR too much, since it's mostly just a quick fix in the hopes of fixing a production capture so it stops getting OOM killed.

willdonnelly:
I've uploaded https://github.com/estuary/connectors/pull/826 to address the use of `x/sync/semaphore`

PR #826, opened 2023-07-18
source-firestore: Use x/sync/semaphore instead of void channels

willdonnelly:
**Description:**

Should be no functional difference, but the code's a little bit cleaner and it handles context cancellation better in theory.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/826)


willdonnelly:
Obsolete. This was done as part of another PR

PR #828, opened 2023-07-19
materialize-redshift: explicitly lock table at start of commit

williamhbaker:
**Description:**

All indications are that obtaining an exclusive lock on the potentially concurrently accessed checkpoints table should cause other transactions to "wait" until they can get the lock. In practice I have seen this not be the case when the lock statement was part of a statement batch, so hopefully a more direct lock statement execution will do the trick.

I was able to manually test this somewhat by simply opening two sessions in a Redshift cluster and getting a lock on a table with the first session. Then when I tried to get a lock with the second session, it blocked (no error) until the first session's transaction committed.

**Workflow steps:**

Redshift materializations with more than one shard operating in the same metadata schema should work without error.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/828)

### DIFF
--- a/materialize-redshift/driver.go
+++ b/materialize-redshift/driver.go
@@ -618,8 +618,6 @@ func (d *transactor) commit(ctx context.Context, fenceUpdate string, hasUpdates 
 	}
 	defer txn.Rollback(ctx)
 
-	var batch pgx.Batch
-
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #832, opened 2023-07-19
materialize-redshift: automatically alter table columns for long strings

williamhbaker:
**Description:**

The previous strategy of creating string columns with default length of `4096` and using `maxLength` schema annotations for longer strings was untenable for certain use cases involving huge numbers of tables materializing many very long strings.

This commit makes the connector smart enough to "enlarge" Redshift `VARCHAR` columns for these long strings automatically. A string longer than the default length of `256` bytes will trigger an `ALTER TABLE ALTER COLUMN` to set the column to be the maximum length, AKA `VARCHAR(MAX)`, which is `65535`. Strings longer than this will result in an error since there is not way for Redshift to store them.

This binary setting of either `256` or `VARCHAR(MAX)` is necessary due to potential races between split shards of the materialization. Finer grains would possibly result in one shard undoing the work of another by "shrinking" the column.

Since the connector inspects the table at startup for their actual maximum lengths, there is an option for the user to adjust columns to be smaller than the maximum length but larger than the default length if they wish.

In addition to running the end-to-end tests, I manually tested this to make sure it correctly responds to long strings showing up in collections, particularly with table and column names containing mixed capitalizing and special characters.

**Workflow steps:**

Materialize long strings without needing to alter the tables manually or add schema annotations.

The `materialize-redshift` docs will need to be updated according to this change.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/832)


willdonnelly (approved):
LGTM

PR #835, opened 2023-07-20
materialize-redshift: minimize concurrency for upload workers

williamhbaker:
**Description:**

This configures the S3 uploader to use a single goroutine for uploading a file in attempts to minimize memory usage of the connector.

We already have a fairly high degree of concurrency for uploads just from the fact that many materializations have more than 1 binding - sometimes many more. And each binding actually gets a separate upload thread for both loads and stores.

I am not totally confident that for the kinds of uploads we are doing (with an `io.Reader` and not an `io.ReadSeeker`) that multiple goroutines are used for uploads, but there is some sparse documentation suggesting that memory usage is directly related to the concurrency factor so this seems like it is worth a try.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/835)

### DIFF
--- a/materialize-redshift/driver.go
+++ b/materialize-redshift/driver.go
@@ -547,8 +547,9 @@ func (d *transactor) Load(it *pm.LoadIterator, loaded func(int, json.RawMessage)
 			return fmt.Errorf("converting Load key: %w", err)
 		}
 
-		b.loadFile.encodeRow(converted)
-
+		if err := b.loadFile.encodeRow(converted); err != nil {
+			return fmt.Errorf("encoding row for load: %w", err)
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

> And each binding actually gets a separate upload thread for both loads and stores.

Would it be feasible to only run one upload at a time? The runtime guarantees that you'll see _all_ of the `Request.Store` for binding N before you see _any_ for binding N+1, so connectors could use the transition from N => N+1 to await an ongoing upload and start a next one.

PR #836, opened 2023-07-20
materialize-redshift: truncate strings longer than VARCHAR limits

williamhbaker:
**Description:**

Setting these configuration options enables automatic truncation of strings too long to fit in a `VARCHAR(MAX)` column or `VARCHAR` field of a `SUPER` column. This, it would seem, is preferable to the connector being unable to materialize collections containing these strings at all.

See AWS documentation for [TRUNCATECOLUMNS](https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-conversion.html#copy-truncatecolumns) and [parsing options for SUPER](https://docs.aws.amazon.com/redshift/latest/dg/super-configurations.html#parsing-options-super).

**Workflow steps:**

Materialize collections with very long strings directly, though in a truncated state, per Redshift's limits.

The `materialize-redshift` documentation will need updates for this, which can be combined with the updates needed from https://github.com/estuary/connectors/pull/832.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/836)

### DIFF
--- a/materialize-redshift/.snapshots/TestSQLGeneration
+++ b/materialize-redshift/.snapshots/TestSQLGeneration
@@ -111,7 +111,8 @@ CREDENTIALS 'aws_access_key_id=accessKeyID;aws_secret_access_key=secretKey'
 REGION 'us-somewhere-1'
 JSON 'auto ignorecase'
 DATEFORMAT 'auto'
-TIMEFORMAT 'auto';
+TIMEFORMAT 'auto'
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

This a great resolution considering, and it's 🤯 to me that redshift has reached a place where they've institutionalized convenient discarding of their user's data.

PR #839, opened 2023-07-21
materialize-redshift: escape leading null escape sequences

williamhbaker:
**Description:**

Redshift will terminate values going into `VARCHAR` columns where the null escape sequence occurs. This is especially a problem when it is the first thing in the string and is followed by other characters, as this results in the load error "Invalid null byte - field longer than 1 byte". Adding an additional escape to the beginning of the string allows the value to be loaded with the initial `"\u0000"` as a regular string.

Load options like `ESCAPE` and `NULL AS` seem somewhat promising for better configuration of this behavior, but these options are not supported when copying from JSON files, see [aws docs](https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-format.html#copy-json-data-file).

If the null escape sequence occurs elsewhere in a string value going into a `VARCHAR` column, the value will be cutoff. This is probably not ideal but just checking if a string starts with the sequence is enough to allow data to be loaded without needing to search through the entirety of every string field. Fortunately, values in the `flow_document` `SUPER` column load without issue and are true to the source data from the collection. For example, collection document like `{"value":"hello\u0000 world"}` will be loaded into the `value` column as `"hello"`, but will still be `"hello\u0000 world"` in the `flow_document` `SUPER` column.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/839)

### DIFF
--- a/materialize-redshift/sqlgen.go
+++ b/materialize-redshift/sqlgen.go
@@ -26,7 +26,22 @@ var rsDialect = func() sql.Dialect {
 		sql.ARRAY:   sql.NewStaticMapper("SUPER", sql.WithElementConverter(sql.JsonBytesConverter)),
 		sql.BINARY:  sql.NewStaticMapper("VARBYTE"),
 		sql.STRING: sql.StringTypeMapper{
-			Fallback: sql.NewStaticMapper("TEXT"), // Note: Actually a VARCHAR(256)
+			Fallback: sql.NewStaticMapper("TEXT", sql.WithElementConverter( // Note: Actually a VARCHAR(256)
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM, I guess.

I'm really not a fan of how this mixes string-escaping levels and creates ambiguity, but I trust that you've thought about the problem more than I and so this is probably the least bad option.

dyaffe:
Any chance we can get this in soon?  Would be good to finish the backfill this weekend

PR #840, opened 2023-07-21
materialize-sql: allow for adding required fields

williamhbaker:
**Description:**

The primary objective of this PR is to allow for new required fields to be added to materialized collections. The desired behavior in this case is to add the column to the materialized table as nullable. Previously we did not allow required fields to be added to a collection that was being materialized by a SQL materialization.

As part of testing this, I found several other scenarios we need special handling for: Adding or removing the `flow_document` field via field selection, re-adding a previously removed field from a collection (the column will already exist), and generally just various differences in how the connectors handle things like dropping null constraints for columns if the column doesn't have a null constraint.

This drove moving these kinds of operations out of templated queries and into behaviors of the connectors to allow for them to handle the requested changes however they need to, usually with some kind of error handling. For example, some databases don't support adding a column to a table if it doesn't already exist. In these cases the easiest thing to do is to try adding the column, and if the returned error is complaining about the column already existing, just carry on.

**Workflow steps:**

I tested this by doing the following with all the SQL materializations (except `materialize-sqlite`):
- Add a required field to a materialized collection -> It should be added to the table as a nullable column
- Add a non-required field to a collection -> It should be added as a nullable column as well
- Remove a required field that was part of the existing materialization -> NOT NULL constraint should be dropped from the table, if applicable
- Remove a non-required field that was part of the existing materialization -> This should work, with nothing done to the table
- Re-add a previously removed field -> This should not fail, even though the column already exists in the table
- Remove & re-add `flow_document` from a delta updates materialization via field selection
- Remove & re-add `flow_published_at` via field selection

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/840)


materialize-sql/.snapshots/TestTableTemplate
```diff
@@ -1,10 +1,10 @@
 
 	CREATE TABLE one."reserved".checkpoints (
-			materialization TEXT,
-			key_begin BIGINT,
-			key_end BIGINT,
-			fence BIGINT,
-			checkpoint TEXT,
+			materialization TEXT NOT NULL,
```
williamhbaker, line 3: These snapshot updates are not a result of anything in this PR, they just haven't been updated since we went back to creating columns as `NOT NULL` if they are required in collection schemas. I've added running the `materialize-sql` package tests to SQL materialization connectors so we don't get out of sync in the future.

materialize-sql/driver.go
```diff
@@ -360,39 +352,6 @@ func (d *Driver) NewTransactor(ctx context.Context, open pm.Request_Open) (pm.Tr
 	return transactor, &pm.Response_Opened{RuntimeCheckpoint: cp}, nil
 }
 
-func isBindingMigratable(existingBinding *pf.MaterializationSpec_Binding, newBinding *pf.MaterializationSpec_Binding) bool {
```
williamhbaker, line 363: This function was all about making sure you can't remove keys from a materialized collection, and that any new fields are not required.

We specifically don't want to restrict new fields to only non-required ones anymore, and collection keys are always required to be materialized via constraints - see [here](https://github.com/estuary/connectors/blob/2ed68817d2f043bb34781cab47211aa3d2d46fd4/materialize-sql/validate.go#L95-L97) for new materializations and [here](https://github.com/estuary/connectors/blob/2ed68817d2f043bb34781cab47211aa3d2d46fd4/materialize-sql/validate.go#L134-L139) for modifications to existing materializations.

materialize-bigquery/sqlgen.go
```diff
@@ -122,23 +122,6 @@ CLUSTER BY {{ range $ind, $key := $.Keys }}
 	{{- end}};
 {{ end }}
 
--- Alter column and mark it as nullable
-
-{{ define "alterColumnNullable" }}
-ALTER TABLE {{ $.Table.Identifier }} ALTER COLUMN {{ $.Identifier }} DROP NOT NULL;
```
williamhbaker, line 128: Previously this would error if the column did not have a `NOT NULL` constraint, which made it effectively impossible to remove a non-required field from a materialized collection.

willdonnelly (approved):
LGTM

psFried (changes requested):
I haven't done a thorough review, but one thing I noticed is that the new `AddColumnToTable` and `DropNotNullForColumn` functions seem like they'll still modify the tables even when `req.DryRun == true`. I think the desired behavior would be for those to still return an action description in that case, though. So it probably means passing a `dryRun` boolean parameter to those?

williamhbaker:
> I haven't done a thorough review, but one thing I noticed is that the new `AddColumnToTable` and `DropNotNullForColumn` functions seem like they'll still modify the tables even when `req.DryRun == true`. I think the desired behavior would be for those to still return an action description in that case, though. So it probably means passing a `dryRun` boolean parameter to those?

Yuuuup good call. I pushed a commit to address this. Thanks!

psFried (approved):
👍👍 LGTM

PR #842, opened 2023-07-24
source-mysql: Ignore all 'FLUSH' query events in replication

willdonnelly:
**Description:**

All `FLUSH FOO` queries should be entirely irrelevant to our capture connector.

Also fixes the logging of these benign queries so that they'll be visible at DEBUG level if we need to see them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/842)


williamhbaker (approved):
LGTM

PR #845, opened 2023-07-25
source-mongodb: consider change events with fullDocument=null

mdibaiee:
**Description:**

- See the comment in code

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/845)

### DIFF
--- a/source-mongodb/pull.go
+++ b/source-mongodb/pull.go
@@ -181,7 +181,14 @@ func (c *capture) ChangeStream(ctx context.Context, client *mongo.Client, bindin
 			if err = c.Output.Documents(binding, js); err != nil {
 				return fmt.Errorf("output documents failed: %w", err)
 			}
-		} else {
+		} else if ev.FullDocument != nil {
+			// FullDocument can be "null" if another operation has deleted the
+			// document. This happens because update change events do not hold a copy of the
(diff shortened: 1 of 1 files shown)


source-mongodb/pull.go
```diff
@@ -181,7 +181,14 @@ func (c *capture) ChangeStream(ctx context.Context, client *mongo.Client, bindin
 			if err = c.Output.Documents(binding, js); err != nil {
 				return fmt.Errorf("output documents failed: %w", err)
 			}
-		} else {
+		} else if ev.FullDocument != nil {
+			// FullDocument can be "null" if another operation has deleted the
```
williamhbaker, line 185: I'm not sure if there is anything we could do differently or better, but is it possible that the change event stream could contain an insert and then a delete for a document? In that case I assume we'd get a delete event for a document that there was never an insert event for, so it would look like something was deleted that never existed.

williamhbaker (approved):
LGTM

source-mongodb/pull.go
mdibaiee: @williamhbaker insert events always have a full document that is part of the event itself, this comment really only applies to updates. I will clarify this in the comment.

Issue #822, opened 2023-07-18
materialize-bigquery needs a better error message when inputting invalid JSON credentials


williamhbaker:
We have some handling for error messaging if a user's JSON credentials cannot be parsed: https://github.com/estuary/connectors/blob/a5c46ecac86cb8828ebee01b42a054db3fe1224e/materialize-bigquery/bigquery.go#L194-L200

However, the `endpoint` has already been created at the point of this check: https://github.com/estuary/connectors/blob/a5c46ecac86cb8828ebee01b42a054db3fe1224e/materialize-sql/driver.go#L77C29-L80

So we should really be handling the possibility of invalid JSON credentials in the BigQuery `NewEndpoint` function.

PR #847, opened 2023-07-25
materialize-bigquery: verify that credentials are valid JSON

williamhbaker:
**Description:**

Verify that the configured credentials are valid JSON, to provide a more instructive error message than `bigquery.NewClient` provides if given credentials that are not formatted as JSON.

Closes https://github.com/estuary/connectors/issues/822

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/847)

### DIFF
--- a/materialize-bigquery/bigquery.go
+++ b/materialize-bigquery/bigquery.go
@@ -35,6 +35,9 @@ func (c *config) Validate() error {
 	if c.ProjectID == "" {
 		return fmt.Errorf("expected project_id")
 	}
+	if c.CredentialsJSON == "" {
+		return fmt.Errorf("expected credentials_json")
+	}
 	if c.Dataset == "" {
 		return fmt.Errorf("expected dataset")
 	}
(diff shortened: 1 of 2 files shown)


mdibaiee (approved):

materialize-bigquery/bigquery.go
```diff
@@ -187,16 +197,9 @@ func newBigQueryDriver() *sql.Driver {
 
 func prereqs(ctx context.Context, ep *sql.Endpoint) *sql.PrereqErr {
 	cfg := ep.Config.(*config)
+	client := ep.Client.(*client)
```
mdibaiee, line 200: How come we don't need to call `cfg.client` anymore? `cfg.client` seems to be doing a lot of stuff 🤔

williamhbaker: It's already part of the endpoint, initialized here: https://github.com/estuary/connectors/blob/5afc5133a39a9ac989d63f36f5265825a4c88650/materialize-sql/driver.go#L77C11-L79

The genesis of these invalid-credentials-json errors was kind of confusing before, since we were doing a check here for exactly that, but it wasn't being hit. And that was because the client already gets created as part of building the endpoint that gets passed to here.

That said, I think this structure could be improved, and at some point in the future I'd like to look at making the `prereqs` checks part of the `Client` interface.

PR #854, opened 2023-07-26
materialize-bigquery: use FLOAT64 for format: number

mdibaiee:
**Description:**

- FLOAT64 can represent a larger range of values than BIGNUMERIC, and although BIGNUMERIC is better suited to loss-less numerics, in case of `format: number`, they will be processed using a double representation, which means we don't gain anything from using BIGNUMERIC here.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/854)


williamhbaker:
LGTM. This should be backwards-compatible with existing tables that were created with `BIGNUMERIC` columns since the JSON encoding of the values is the same.

Issue #830, opened 2023-07-19
Ignore TRUNCATE commands for MySQL


dyaffe:
**Problem**
We currently will stop processing a MySQL DBs CDC log upon seeing a `TRUNCATE` call.  This prevents moving forward with an entire capture.

We have some proof that truncate commands are used to trim DBs but users would still like to see the records from their DB.  This is both from live users and from what Fivetran does.

**Proposal**
Ignore `TRUNCATE` calls

PR #855, opened 2023-07-26
source-{mysql,postgres}: Ignore TRUNCATE events

willdonnelly:
**Description:**

Table truncation is a bit of a weird middle ground, where its formal semantic meaning is something like "delete all existing rows" but in practice it's often used as a way of expunging old data simply to keep tables small.

In Postgres and MySQL, table truncation is reported as its own event type, distinct from row deletions. And we don't have any good way to convert a truncate event into a sequence of deletions because that would require keeping track of every single row in a table as part of our persistent capture state, which obviously doesn't scale.

So for now the best thing we can do within our data model is to ignore them. The good news is that this is often what the user would like to have happen, but we may need to revisit whether there's any way to model the truncation operation explicitly so that other behaviors could be implemented downstream.

One hacky option would be to add some sort of 'truncation epoch' counter/timestamp in the document metadata, so that after a table has been truncated future capture documents will at least be distinguishable from documents before that most recent truncation. But this requires further discussion and so for now we're doing the easy thing and ignoring them.

This fixes https://github.com/estuary/connectors/issues/830

**Workflow steps:**

Captures will no longer stall if an active table is truncated. Instead they will keep on going and ignore the truncation entirely, other than emitting a warning.

We probably need to update the `source-mysql` and `source-postgres` docs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/855)

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -488,7 +488,7 @@ func (rs *mysqlReplicationStream) handleQuery(schema, query string) error {
 		}
 	case *sqlparser.TruncateTable:
 		if streamID := resolveTableName(schema, stmt.Table); rs.tableActive(streamID) {
-			return fmt.Errorf("unsupported operation (go.estuary.dev/eVVwet): %s", query)
+			logrus.WithField("table", streamID).Warn("ignoring TRUNCATE on active table")
 		}
 	case *sqlparser.RenameTable:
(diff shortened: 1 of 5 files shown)


jgraettinger (approved):
LGTM

PR #857, opened 2023-07-26
source-http-ingest: log json parse errors

psFried:
**Description:**

This is kind of a quick and dirty fix to make sure we log request errors due to invalid JSON. These were not previously logged because parsing failed in the axum json extractor, so axum sends the response automatically. The normal request logging only logs the status, though, and we want to log the whole body, at least for now. So we'll probably end up doing some more refactoring of this soon, but for now this ought to get us what we need.

One other potentially unrelated fix here is upping the request body limit from the default of 2MiB to 64MiB, which matches the maximum document size allowed by Flow.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/857)

### DIFF
--- a/source-http-ingest/src/server.rs
+++ b/source-http-ingest/src/server.rs
@@ -2,10 +2,12 @@ use crate::{transactor::Transactor, Binding, EndpointConfig};
 
 use anyhow::Context;
 use axum::{
-    extract::{Json, Path, State},
+    body::Bytes,
+    extract::{DefaultBodyLimit, Json, Path, State},
     routing, Router,
 };
 use http::status::StatusCode;
+use serde_json::Value;
 use utoipa::openapi::{self, schema, security, OpenApi, OpenApiBuilder};
 use utoipa_swagger_ui::SwaggerUi;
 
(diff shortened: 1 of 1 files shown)


source-http-ingest/src/server.rs
```diff
@@ -233,35 +241,67 @@ impl Handler {
...
 
+        // Try to deserialize the body as json _before_ acquiring the handlers lock,
+        // to prevent slow connection from impacting others.
+        let json: Value = match serde_json::from_slice(&body) {
+            Ok(j) => j,
+            Err(err) => {
+                let content_type = request_headers.get("content-type");
+                let content_encoding = request_headers.get("content-encoding");
+
+                let req_body = String::from_utf8_lossy(&body);
```
jgraettinger, line 274: perhaps write a bounded prefix? otherwise this could blow out the log volume pretty easily

source-http-ingest/src/server.rs
```diff
@@ -233,35 +241,67 @@ impl Handler {
...
+                &collection_path,
+            ));
+        }
+
         let start = std::time::Instant::now();
         tracing::debug!(path = %collection_path, "looking up handler for path");
         let mut handlers_guard = self.handlers_by_path.lock().await;
 
         let Some(collection) = handlers_guard.get_mut(collection_path.as_str()) else {
+            tracing::info!(uri_path = %collection_path, "unknown uri path");
```
jgraettinger, line 297: warn?

jgraettinger (approved):
LGTM % comments

source-http-ingest/src/server.rs
psFried: This was intentional, so that you can turn the log level to `warn` if it gets too spammy due to bots or something. I'm assuming you don't have a very strong opinion here and were just pointing it out because it seemed accidental

source-http-ingest/src/server.rs
psFried: Good idea. Done.

PR #858, opened 2023-07-26
source-http-ingest: handle request decompression

psFried:
**Description:**

The previous commit seems to have broken request decompression.  This commit fixes it. TBH, I don't actually see why decompression ever would have worked in the first place, which makes me suspicious that there was some sort of issue in my earlier test setup. In other words, this might just be implementing request decompression for the first time, rather than fixing a regression.

In either case, this explicitly adds a request decompression middleware.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/858)

### DIFF
--- a/source-http-ingest/Cargo.lock
+++ b/source-http-ingest/Cargo.lock
@@ -38,10 +38,41 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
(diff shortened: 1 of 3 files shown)


psFried:
@jgraettinger I'm gonna go ahead and merge this pre-review, since I've tested locally and I think it may resolve the issue. 🤞

PR #859, opened 2023-07-27
materialize-bigquery: use schema from table for edc file

williamhbaker:
**Description:**

Instead of defining the external file schema from our current dialect definitions, introspect the exist table columns to build a schema that is ensured to be compatible with the table. This way we can change the dialect and as long as the JSON encoding of the values is the same, the loads from the external file will still work.

This allows for changing the materialized column for JSON type `number` to be a `FLOAT64`, and for it to be backward-compatible with existing `BIGNUMERIC` columns.

This was tested by setting up a materialization on the old connector with a `number` type in the schema, materializing a bit of data, switching to the version of the connector from this PR, and then materializing some more data. Also, for the column introspection to EDC schema part, I tested with a variety of "challenging" field and table names, with various mixes of capitalization and special characters.

**Workflow steps:**

Materialize JSON fields of type `number` and get `FLOAT64` columns, which are capable of storing larger values (at reduced precision).

N/A

**Notes for reviewers:**

This is a step toward changing object and array column types to `JSON` in bigquery, but doesn't quite get there. The JSON encoding for these values actually must be different than it is currently: Right now we encode them as a string, which would get loaded into a Bigquery `JSON` column as a string, rather than its JSON interpretation. And we can't change the encoding to be JSON bytes because it would break existing materializations.

It should be possible to use the table schema introspection used here to figure out if an existing table has column types of `STRING` or `JSON` for corresponding flow field types of `object` or `array`, and use an alternate encoding depending on the existing table. That wasn't looking like it would be an easy thing to do so I'm holding off for now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/859)

### DIFF
--- a/materialize-bigquery/bigquery.go
+++ b/materialize-bigquery/bigquery.go
@@ -78,13 +78,11 @@ func (c *config) client(ctx context.Context) (*client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("creating bigquery client: %w", err)
 	}
-	log.WithField("projectID", billingProjectID).Info("bigquery client successfully created")
 
 	cloudStorageClient, err := storage.NewClient(ctx, clientOpts...)
 	if err != nil {
 		return nil, fmt.Errorf("creating cloud storage client: %w", err)
 	}
-	log.Info("cloud storage client successfully created")
 
 	return &client{
 		bigqueryClient:     bigqueryClient,
--- a/materialize-bigquery/.snapshots/TestSQLGeneration
+++ b/materialize-bigquery/.snapshots/TestSQLGeneration
@@ -9,7 +9,7 @@ CREATE TABLE IF NOT EXISTS projectID.dataset.target_table (
 		integer INT64 NOT NULL,
 		string STRING NOT NULL,
 		`defAULT` STRING,
-		number BIGNUMERIC,
+		number FLOAT64,
 		person_place_ STRING,
 		source_name STRING,
 		`with-dash` STRING,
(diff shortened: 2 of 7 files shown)


materialize-bigquery/bigquery.go
```diff
@@ -78,13 +78,11 @@ func (c *config) client(ctx context.Context) (*client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("creating bigquery client: %w", err)
 	}
-	log.WithField("projectID", billingProjectID).Info("bigquery client successfully created")
```
williamhbaker, line 81: Cleaning up some extraneous logging in this file that was added in https://github.com/estuary/connectors/pull/551 which I always intended to remove but never got around to until now.

materialize-bigquery/bigquery.go
```diff
@@ -391,13 +388,5 @@ func (c client) InstallFence(ctx context.Context, _ sql.Table, fence sql.Fence)
...
 	fence.Checkpoint = checkpoint
 
-	log.WithFields(log.Fields{
-		"fence":            fence.Fence,
-		"keyBegin":         fence.KeyBegin,
-		"keyEnd":           fence.KeyEnd,
-		"materialization":  fence.Materialization.String(),
-		"checkpointsTable": fence.TablePath,
-	}).Info("fence installed successfully")
-
```
mdibaiee, line 401: I find these kinds of logs useful, perhaps keep them around as `Debug` logs?

mdibaiee (approved):

materialize-bigquery/bigquery.go
williamhbaker: I'll keep these ones around installing & reading the fence around since otherwise it is kind of weird when the last thing logged is "installing fence", which can sometimes need to wait for a long time if a previous commit had a lock on the checkpoints table.

PR #860, opened 2023-07-27
source-mysql: Translate empty `json` columns as `null`

willdonnelly:
**Description:**

The empty string is not valid JSON and so if we somehow got an empty string from a `json` column and made it a `json.RawMessage` it would fail later on when we tried to serialize it.

We have reason to believe that this can happen, and if it does in fact happen it seems reasonable to translate the empty string into JSON `null`.

(See https://bugs.mysql.com/bug.php?id=98496 for one way that this sort of situation could arise)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/860)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -200,6 +200,9 @@ func (db *mysqlDatabase) translateRecordField(columnType interface{}, val interf
 			case "blob", "tinyblob", "mediumblob", "longblob":
 				return val, nil
 			case "json":
+				if len(val) == 0 {
+					return nil, nil
+				}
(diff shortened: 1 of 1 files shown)


willdonnelly:
I tested this locally with the reproduction steps from https://bugs.mysql.com/bug.php?id=98496 (declaring a `json not null` column and doing an `INSERT IGNORE` of SQL `NULL` into it) and have verified that replication failed without this fix and works as expected after the fix.

williamhbaker (approved):
LGTM

PR #862, opened 2023-07-27
source-http-ingest: accept array bodies

psFried:
**Description:**

Updates the source-http-ingest connector to accept arrays of JSON objects.  If the body of a request is a JSON array, then each item in the array must be an object that conforms to the write schema of the collection.

In the case that `idFromHeader` is set in the resource config, arrays are dealt with by suffixing the header value with the index of each array element.

This commit does _not_ update the openapi spec. Doing so may be desirable, but it's totally necessary right now, and I'd prefer to do it in a subsequent commit.

**Workflow steps:**

start sending http requests with array bodies ;)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/862)

### DIFF
--- a/source-http-ingest/src/server.rs
+++ b/source-http-ingest/src/server.rs
@@ -112,71 +112,103 @@ fn required_string_header<'a>(
         })
 }
 
-fn err_response(
-    status: StatusCode,
-    err: anyhow::Error,
-    request_path: &str,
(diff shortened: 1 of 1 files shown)


psFried:
YOLOing this one, since it's low impact and fixes (🤞) an issue with a production capture

PR #867, opened 2023-07-31
source-mysql: Normalize invalid enum values to ""

willdonnelly:
**Description:**

Tweaks value translation logic to normalize invalid enum values to `""`. This happened automatically via backfills, but replication will give us the enum values as an integer and we have to translate `0` as `""` ourselves.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/867)


willdonnelly:
This PR also changes the MySQL test support code to use `INSERT IGNORE` for all test dataset initialization. I believe that this should be a backwards compatible fix (if not we'll find out from this CI run) and it's a change I've had to make locally several times in the past month to reproduce particular edge cases around data handling.

jgraettinger (approved):
LGTM

PR #868, opened 2023-07-31
sqlcapture: No error if primary key changes after backfill

willdonnelly:
**Description:**

If the selected primary key for a binding changes (most likely because the binding has been changed to point at a new collection whose key is different) then we shouldn't really care so long as we're not in the middle of an ongoing backfill.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/868)

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -339,6 +339,26 @@ func (c *Capture) updateState(ctx context.Context) error {
 			return fmt.Errorf("stream %q: primary key must be specified", streamID)
 		}
 
+		// See if the stream is already initialized. If it's not, then create it.
+		var streamState, ok = c.State.Streams[streamID]
+		if !ok || streamState.Mode == TableModeIgnore {
+			c.State.Streams[streamID] = &TableState{Mode: TableModePending, KeyColumns: primaryKey, dirty: true}
+			continue
+		}
+
+		// The following safety checks only matter if we're actively backfilling the table or
+		// intend to start backfilling it shortly. But checking for that would risk silently
+		// failing if a new keyed backfill mode were added in the future, so instead we check
+		// for the set of known-to-be-fine situations here.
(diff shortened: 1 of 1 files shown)


sqlcapture/capture.go
```diff
@@ -339,6 +339,26 @@ func (c *Capture) updateState(ctx context.Context) error {
...
+			c.State.Streams[streamID] = &TableState{Mode: TableModePending, KeyColumns: primaryKey, dirty: true}
+			continue
+		}
+
+		// The following safety checks only matter if we're actively backfilling the table or
+		// intend to start backfilling it shortly. But checking for that would risk silently
+		// failing if a new keyed backfill mode were added in the future, so instead we check
+		// for the set of known-to-be-fine situations here.
+		var notBackfilling = (streamState.Mode == TableModeIgnore)
+		notBackfilling = notBackfilling || (streamState.Mode == TableModeActive)
```
jgraettinger, line 354: nit:
```
var notBackfilling = (streamState.Mode == TableModeIgnore)
    || (streamState.Mode == TableModeActive)
    || (streamState.Mode == TableModeKeylessBackfill)
    ...
```

jgraettinger (approved):
LGTM % nit

sqlcapture/capture.go
willdonnelly: Sadly Go doesn't let you line the operators up nicely like that and the best I can do is:

```
var notBackfilling = (streamState.Mode == TableModeIgnore) ||
	(streamState.Mode == TableModeActive) ||
	(streamState.Mode == TableModeKeylessBackfill) ||
	(streamState.Mode == TableModePending && binding.Resource.Mode == BackfillModeOnlyChanges) ||
	(streamState.Mode == TableModePending && binding.Resource.Mode == BackfillModeWithoutKey)
```

PR #870, opened 2023-08-01
source-mysql: Ignore many benign table alterations

willdonnelly:
**Description:**

I have examined the parser AST structs and have added explicit failure cases for the types of table alteration which I believe might possibly be an error, then changed the default to ignore any of the other 14 alteration types that we don't need to care about because they're totally irrelevant to our capture.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/870)


jgraettinger (approved):
LGTM

PR #871, opened 2023-08-01
source-mysql: Allow ALTER TABLE to drop key constraints

willdonnelly:
**Description:**

We don't care about non-primary key constraints at all, and I think it's safe to allow dropping the primary key constraint too, since we do support capturing from tables without a primary key in the first place.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/871)


jgraettinger (approved):
LGTM

PR #873, opened 2023-08-02
source-batchsql: Proof of concept

willdonnelly:
**Description:**

This is a small and fairly dumb connector which provides SQL database capture by running user-provided queries periodically against the database and translating any result rows into JSON.

This PoC supports Postgres only, but the concept applies to any SQL database and probably a lot of the logic could be factored out if desired. This connector is:

  + About an order of magnitude simpler than `source-postgres` by lines of code.
  + Able to capture from views or even from the results of a complex ad hoc query.
  + Completely unable to generate useful schemas, although it *does* contain some logic to discover tables and views and emit bindings which blindly `SELECT * FROM <foo>` those entities.
  + Unable to capture incremental changes when such changes can't be easily identified using a `WHERE` clause.
  + Less able to handle massive datasets, although depending on the data it may still be a reasonable approach.
  + Currently stateless, although persisting cursor values for each binding would be an easily added feature. I'm torn on whether it's actually desirable though, restarting from the "no cursor" case after each connector restart has a certain simplicity that I kind of like.

But it's still interesting and there are situations where this could be the tool of choice, especially if/when persistent cursor state is added.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/873)


source-batchsql/datatypes.go
```diff
@@ -0,0 +1,98 @@
...
+	case [16]uint8: // UUIDs
+		var s = new(strings.Builder)
+		for i := range x {
+			if i == 4 || i == 6 || i == 8 || i == 10 {
+				s.WriteString("-")
+			}
+			fmt.Fprintf(s, "%02x", x[i])
+		}
+		return s.String(), nil
+	case pgtype.Bytea:
```
willdonnelly, line 34: I blindly copy pasted the value translation logic from `source-postgres` here, however I'm not entirely sure if we're receiving `pgtype` values since I wrote the main query logic using the Go `database/sql` interface and that might change how values from the database are returned to us.

I haven't spent any time testing the extent of datatype support since this was just a quick proof-of-concept.

source-batchsql/main.go
```diff
@@ -0,0 +1,480 @@
...
+	Output    *boilerplate.PullOutput
+}
+
+func (c *capture) Run(ctx context.Context) error {
+	log.WithFields(log.Fields{
+		"address":  c.Config.Address,
+		"user":     c.Config.User,
+		"database": c.Config.Database,
+	}).Info("connecting to database")
+	var db, err = sql.Open("pgx", c.Config.ToURI())
```
willdonnelly, line 333: The `sql.Open` call and associated config-to-DSN mapping is essentially the only part of this connector which is database-specific (well, also the datatype mapping stuff potentially but I haven't investigated that yet), so in principle almost all of this logic could be factored out into a helper package and the per-database connector code would be 100-200 lines saying essentially "here's what a configuration looks like, here's a function that takes a config and returns a generic `*sql.DB`, and here's how to translate output values".

source-batchsql/main.go
```diff
@@ -0,0 +1,480 @@
...
+			Key:                defaultKey,
+		})
+	}
+
+	return &pc.Response_Discovered{Bindings: bindings}, nil
+}
+
+var catalogNameSanitizerRe = regexp.MustCompile(`(?i)[^a-z0-9\-_.]`)
+
+func recommendedCatalogName(schema, table string) string {
```
willdonnelly, line 248: This function is copied from `sqlcapture` and could be exported from there in a real implementation of this concept.

source-batchsql/main.go
```diff
@@ -0,0 +1,480 @@
...
+	if r.PollInterval != "" {
+		if _, err := time.ParseDuration(r.PollInterval); err != nil {
+			return fmt.Errorf("invalid poll interval %q: %w", r.PollInterval, err)
+		}
+	}
+	return nil
+}
+
+// documentMetadata contains the source metadata located at /_meta
+type documentMetadata struct {
```
willdonnelly, line 117: For now I've chosen to have the document metadata be as simple as possible, with a "polled at" timestamp and a "result row index" counter which I see as being kind of analagous to how a file source has file and offset metadata.

source-batchsql/main.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+	if len(params) > 0 {
+		uri.RawQuery = params.Encode()
+	}
+	return uri.String()
+}
+
+// Resource represents the capture configuration of a single resource binding.
+type Resource struct {
+	Name         string   `json:"name" jsonschema:"title=Name,description=The unique name of this resource."`
```
willdonnelly, line 88: This is a hacky way of getting some sort of "unique name" for a given capture binding. I don't think I ended up using it, but this is what persistent state would be associated with if I had implemented that. There's obviously some overlap with what we discussed yesterday about bindings having a stable ID.

dyaffe:
Would it be possible to add the schema discovery type options we have in our current Postgres Capture to something like this?  I could imagine it being amazing to have the option to either "capture via SQL Queries" where a user would enter one or many SQL queries that we would capture into collections or "Find tables to capture" using this connector. 

Maybe that's a future enhancement.  I feel like this by itself would be good enough for the current client.  Would love to give it to them to try it soon if possible!

source-batchsql/main.go
```diff
@@ -0,0 +1,480 @@
...
+		}
+		cursorValues = nextCursor
+		time.Sleep(pollInterval)
+	}
+	return ctx.Err()
+}
+
+func (c *capture) poll(ctx context.Context, db *sql.DB, bindingIndex int, tmpl *template.Template, cursorNames []string, cursorValues []any) ([]any, error) {
+	var templateArg = map[string]any{
+		"subsequent": len(cursorValues) != 0,
```
jgraettinger, line 396: nit: what about `IsFirstQuery` ?

source-batchsql/main.go
```diff
@@ -0,0 +1,480 @@
...
+	pf "github.com/estuary/flow/go/protocols/flow"
+	"github.com/invopop/jsonschema"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
+
+	_ "github.com/jackc/pgx/v4/stdlib"
+)
+
+// Config tells the connector how to connect to and interact with the source database.
+type Config struct {
```
jgraettinger, line 27: sanity checking - you believe `Config` is extensible to other DBs ? How might that work?

jgraettinger (approved):
LGTM

This is really neat & flexible and seems quite sound, for a PoC. Approving so we can start trying it out for customer use cases.

I quite like your call of not bothering with state persistence but having the connector be long-lived. That gives us _mostly_ incremental capture, while avoiding locking down at this early time what its state checkpoints are going to look like.

Given that, the only "gotcha" I can think of is that the moment we have real use-cases running on it, the `Config` will be hard to change. Same with any template arguments introduce.

willdonnelly:
> Would it be possible to add the schema discovery type options we have in our current Postgres Capture to something like this?

Anything is possible with sufficient code, the question here is about how we want to balance development effort. My current stance is that I'd rather avoid trying to do fancy schema generation and just rely on our existing schema inference support, for a couple of reasons:

- Firstly, the JSON schema we generate is essentially a guarantee "this is what the column values should look like after being processed through the database, the client library, and then our connector's value plumbing and JSON serialization", and ensuring that's correct for all value types in every edge case represents a fairly hefty chunk of work.
- And unfortunately we can't necessarily just factor this out of our main CDC capture implementations and reuse it, for boring technical reasons. Basically this PoC is written using the generic Go `database/sql` interface as much as possible, in order to minimize the amount of database-specific code a particular connector would require. But that means that column values get processed a little bit differently from the normal CDC connector, and so basically require a whole different (but _hopefully_ much smaller) translation function before we JSON serialize them.
- Finally there's the pragmatic argument: this connector should generally not be a user's first choice, they should use our much more polished CDC connectors wherever possible. This discovery output isn't meant to be an easy "click Next and it will work" thing, because that would result in running a dumb `SELECT * FROM foobar` every 5 minutes which _really_ doesn't scale well [1]. The only reason I threw in discovery at all was so that the user could have a nice list of "here's the tables you might want to capture" in the UI, along with an obvious place to go and edit the select query to be less dumb.

[1] The obvious counterargument here is that it's possible (at least in Postgres) to use a `SELECT xmin, * FROM foo WHERE xmin > $1 ORDER BY xmin;` query, which would be a bit smarter. In fact here's an example of how that works in a test catalog:

```
bindings:
  - resource:
      name: "foobar"
      template: >
        SELECT xmin, * FROM public.foobar
        {{if .subsequent}}
            WHERE xmin::text::bigint > $1 ORDER BY xmin::text::bigint
        {{end}}
      cursor: ["xmin"]
      poll: 5m
    target: "acmeCo/something/foobar"
```

The problem is that this only works for tables (no views), and so trying to get any more clever than `SELECT * FROM foobar` means we have to start paying attention to the types of things and implementing conditional logic et cetera. So for now I did the dumbest thing that could possibly work there.

source-batchsql/main.go
willdonnelly: Sounds good, I'll change that.

jgraettinger:
Two follow-up questions then:

* Should this be called source-postgres-query ? Since you intend multiple connectors for distinct DBs (which I agree with).
* Should we have "discovery" not actually do _any_ enumeration of the existing DB catalog? Instead, it could return a single suggested binding with an inferred schema annotation and a placeholder name and query, which is basically a "hello world" you would start to write your query into. @dyaffe I'm _not_ suggesting that this is where we want to ultimately end up, but I _am_ worried about precedent we're setting with attempting to scan the DB catalog and return garden-path queries. A placeholder "your SQL goes here" query with comments and template usage guide, etc, is seeming more in line with what we can actually support today. (I'm not sure if the UI will show it in a reasonable text box though 🤔 )

jgraettinger:
Also FYI @travjenkins, I said just yesterday "I can't think of a reason why users might need to manually add capture bindings". Well.... this connector (which allows users to capture arbitrary SQL queries) is one good counter example where they might.

willdonnelly:
> Should this be called source-postgres-query ? Since you intend multiple connectors for distinct DBs (which I agree with).

Yes, I think so. I'm working on an update to this PR right now which does that refactoring and renaming. Although I still kind of prefer `source-postgres-batch` or something, at least to me "batch" evokes a pretty intuitive understanding of what it's doing while "query" just feels kind of confusing in a name. But I can see how other people's intuitions could differ there.

dyaffe:
I like `source-postgres-batch` -- seems more generic as this evolves

jgraettinger:
> When you say "not view" are you suggesting that we should exclude views from discovery entirely

Yep I'm suggesting this ☝️.

The intuitions are that a) we think the common PLG use-case is going to be physical table replication without WAL access, and b) views can be quite large in their result set, and there's no obvious "always works" way to grab only the _incremental_ changes of a view without more user involvement (for example, selecting a reasonable cursor field, etc).

So discovery would then focus on physical tables where we can deliver a reasonable and performant UX without the user needing to modify discovered bindings at all.

willdonnelly:
I've uploaded an all-new iteration of this PR which renames `source-batchsql` to `source-postgres-batch`, factors out the generic logic into a separate file with a parameterized driver struct `BatchSQLDriver` and only about 150 lines of Postgres specific code, and tweaks the discovery resources as we've just discussed, to exclude views and use `xmin` cursors for table queries.

One thing I wanted to call out though:

> a reasonable and performant UX

Note that `WHERE xmin > $1` queries don't scale super great, AFAIK they're not satisfied using any sort of index and require a scan over the entire table. However since this is a batch-mode connector it just issues one query and processes the entire result set no matter how large it is or how long that takes, and then waits 5 minutes between queries, so in practice it'll probably scale _okay_ for tables below tens/hundreds of GiB.

jgraettinger:
> One thing I wanted to call out though

Yep, noted! Thanks. Agreed this would work only for small to medium tables -- that's the target right now.

By "efficient" I was referring to the actual data coming out of the capture task, since we price on that. It would be poor and aggravating UX if we garden-path'd people into capturing even smallish tables or views as a full refresh done over and over again, and then charged them for every one of them.

willdonnelly:
Makes sense. I've uploaded one final modification to this PR, in which I removed the `translateRecordField` function that I previously copy-pasted from `source-postgres` and turned the translation into a no-op for now. I've verified that we don't really seem to need any translation for most data types, even oddball ones like `cidr` which I know would be really ugly if `source-postgres` didn't do anything about it seem to be getting translated into a reasonable JSON string via the different plumbing used by this connector.

(Maybe we should have used that for `source-postgres` too. Oh well, hindsight is 20/20)

willdonnelly:
After a tiny bit more fiddling I believe the connector is in a more or less mergeable state. It's still not super polished, but it seems to basically work. Let me know if anybody wants more time for a review, but otherwise I'll go ahead and merge it in a bit since I've still got that LGTM from earlier.

One thing worth calling out: currently there's a...not quite a _bug_ around table alterations, but a suboptimal case which causes the connector to exit and restart. Arguably this might be desirable, since it means we'll recapture all column values! But the actual reason for it is that the Go Postgres client library is doing some prepared-query caching and the table alteration makes the old query no longer valid. I believe that _in addition to the error_ the client library also invalidates the old cache entry, so if we checked for this error and retried it would work correctly and keep going without a restart.

jgraettinger:
Still LGTM to land!

Just FYI, with near absolute certainty the next ask will be to improve discovery to have rough equivalence w/ the CDC connector, where it outputs the correct collection key and table-derived schema.

PR #875, opened 2023-08-02
materialize-sql: handle materializing fields with multiple types

williamhbaker:
**Description:**

Materialize fields with multiple types as their value converted to JSON. This provides a reasonable representation of the data for the table column. Previously we prohibited these values from being materialized at all.

**Workflow steps:**

Include any field you want in field selection and have it be a column in your materialized table, except for columns with a single type of `null`. I considered allowing `null` types to be materialized as well but it would not have been much of a simplification, and would have required some kind of column type be assigned for this value, which would be strange. It's probably a mistake to include a field having only the type of `null` in the selected fields, since more than likely its type will later change to something else which would be an incompatible migration.

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/875)


mdibaiee (approved):
This is great! Thanks Will

materialize-sql/driver.go
```diff
@@ -191,7 +191,7 @@ func (d *Driver) Apply(ctx context.Context, req *pm.Request_Apply) (*pm.Response
 					if !previousNullable && !newCol.MustExist {
 						alterAction, err := endpoint.Client.DropNotNullForColumn(ctx, req.DryRun, newTable.Identifier, newCol.Identifier)
 						if err != nil {
-							return nil, fmt.Errorf("dropping NOT NULL constraint for on longer required field for column '%s' in table '%s': %w", newTable.Identifier, newCol.Identifier, err)
+							return nil, fmt.Errorf("dropping NOT NULL constraint for no longer required field for column '%s' in table '%s': %w", newTable.Identifier, newCol.Identifier, err)
```
mdibaiee, line 194: nit: this message is pretty hard to read, I had to read it multiple times until I understood what is happening 😁 

Maybe something like this:
```suggestion
							return nil, fmt.Errorf("dropping NOT NULL constraint for column '%s' in table '%s' because it is no longer a required property: %w", newTable.Identifier, newCol.Identifier, err)
```

PR #876, opened 2023-08-03
materialize-snowflake: batch table creation and run batches concurrently

williamhbaker:
**Description:**

Creating hundreds of tables in a single apply operation takes quite a while when run using a single statement. This change creates batches of table creation statements and runs them with concurrent workers.

Snowflake does not support DDL statements in transactions, so there's no reason to run them as part of a transaction. We just need to make sure the materialization spec update happens only after all of the table alterations are complete.

With this strategy, we can create several hundred tables within the 5 minute timeout period allowed for an Apply RPC.

**Workflow steps:**

Create a materialization with up to ~300 bound collections. For testing purposes I used 400 tables. The Apply action takes 1-2 minutes with this many tables. I was also able to materialize documents from all of these collections to each of the 400 created tables.

N/A

**Notes for reviewers:**

N/A

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/876)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -20,6 +20,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	sf "github.com/snowflakedb/gosnowflake"
 	"go.gazette.dev/core/consumer/protocol"
+	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

Issue #2, opened 2021-06-03
DynamoDB Source


psFried:
We'd like to have a source connector for dynamodb. We've done a bit of research and thinking on it already, which I'll write up here.

DynamoDB has a feature called [streams](https://docs.amazonaws.cn/en_us/amazondynamodb/latest/developerguide/Streams.html) that allows reading CDC events. There is a separate feature that can automatically forward these CDC events to a Kinesis stream. Our connector should use dynamodb streams, _not_ the kinesis feature. The kinesis feature has a seemingly higher cost, and CDC events may be duplicated in the kinesis stream. Additionally, the kinesis feature does not scale automatically to adjust for the throughput on the dynamodb table. Dynamodb streams has a lower cost and scales automatically as the dynamodb table itself scales. The dynamodb streams API is very similar to the Kinesis API, though, so the two implementations can likely share a lot of code.

When processing CDC data, it's important to process events in the correct order. If you have an insert followed by a delete on the same key, you will end up with incorrect results if you processed the delete before the insert. This is complicated by design of Shards in dynamodb streams. Shards each have a parent shard, and each shard may have 0 or more child shards. You must read a parent shard entirely before you start reading a child shard if you want to process documents in the correct order. This is problematic because of how we distribute reads of the dynamodb (or kinesis) shards among the set of Flow consumer shards. Consumer shard A may decide to process dynamo shard X before Consumer shard B has finished reading the parent of X. Solving this seems like it would require some shared state.

Another possible approach for dealing with ordering is to use the `ApproximateCreationDateTime` of each dynamodb record with a `maximize` reduction strategy on the root document so that the most recent CDC document for each key is the one that will take precedence.

williamhbaker:
I've spent some time working on this and getting a working implementation.

For a single Flow consumer shard it's reasonably straightforward, but becomes more complicated when considering multiple Flow consumer shards for reasons alluded to in the original issue description, in addition to considerations for backfilling existing data from a table.

I'll describe the general mode of operation first, followed by a description of the challenges and a _possible_ way to work with scaled-out Flow consumer shards.

## General Mode of Operation (Single Flow Shard)

[...]

PR #877, opened 2023-08-03
source-dynamodb: new connector

williamhbaker:
**Description:**

New capture connector for DynamoDB.

I've attempted to describe the really important parts in code comments that handle the more interesting aspects of the connector. For additional background, see the issue description for https://github.com/estuary/connectors/issues/2, and especially the reply [here](https://github.com/estuary/connectors/issues/2#issuecomment-1622197597), which describes the connector's workings in more detail.

Very generally, the connector captures data from one or more DynamoDB tables by:
- Discovering any tables it has access to that have the appropriate DynamoDB streams enabled
- Generating a minimal schema for these tables. The schema will include the partition key and sort key, if there is a sort key. Other elements of the schema will need to be determined by schema inference.
- Backfills the table by alternating between doing backfill scan queries for 1 hour, and completely catching up reads of stream shards in order to keep the shard data from expiring.
- Once the backfill for a table is done, continuously reads data from its streams indefinitely.

This implementation does NOT support splitting Flow task capture shards. Enabling this would take some more work to enable coordination between the capture shards.

Closes https://github.com/estuary/connectors/issues/2

**Workflow steps:**

Use the new connector to capture from DynamoDB.

New documentation will need to be written, which I will do concurrently with merging this PR.

**Notes for reviewers:**

The most difficult part of this connector to test is around the reading of elaborate DynamoDB stream shard configurations, such as one with a long lineage of shards and split child shards. That sort of thing is not possible to generate in the emulator and not really practical to induce/test in a live database. So this is an area I would consider higher risk, since I haven't been able to thoroughly test it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/877)


source-dynamodb/.snapshots/TestDiscovery-additional_table_without_stream_enabled
```diff
@@ -0,0 +1,96 @@
...
+            "_meta"
+          ],
+          "properties": {
+            "_meta": {
+              "type": "object",
+              "required": [
+                "op"
+              ],
+              "properties": {
+                "approximateCreationDateTime": {
```
jgraettinger, line 34: `createdAt` ? Or `sourceCreatedAt` ? This name is a mouthful, its description can note that it's approximate, and `format` already conveys that it's a date-time.

source-dynamodb/.snapshots/TestDiscovery-additional_table_without_stream_enabled
```diff
@@ -0,0 +1,96 @@
...
+    "recommended_name": "discoverTable",
+    "resource_config_json": {
+      "table": "discoverTable"
+    },
+    "document_schema_json": {
+      "$defs": {
+        "discoverTable": {
+          "type": "object",
+          "required": [
+            "partitionKey"
```
jgraettinger, line 12: should we use `/_meta/partitionKey` or even `/_meta/key` ?

The partition key isn't _literally_ part of the document, as I understand, so it would be a bit more consistent in the sense that "connector-added properties go under /_meta". It also roughly matches the behavior of other schema-less connectors that key off of /_meta/...

It believe this will work w/ end-to-end materializations, despite `/_meta` filtering from recommended fields, because it's the collection key.

source-dynamodb/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+FROM ghcr.io/estuary/base-image:v1
+
+WORKDIR /connector
+ENV PATH="/connector:$PATH"
+
+# Bring in the compiled connector artifact from the builder.
+COPY --from=builder /builder/connector ./source-dynamodb
+
+LABEL FLOW_RUNTIME_PROTOCOL=capture
+LABEL CONNECTOR_PROTOCOL=flow-capture
```
jgraettinger, line 36: nit: you only need `LABEL FLOW_RUNTIME_PROTOCOL=capture` now. `CONNECTOR_PROTOCOL` isn't used.

source-dynamodb/backfill.go
```diff
@@ -0,0 +1,322 @@
...
+	// Arrange for stopCh to be closed after the provided duration to signal the backfill workers to
+	// stop.
+	stopCh := make(chan struct{})
+	stopTimer := time.AfterFunc(dur, func() { close(stopCh) })
+	defer stopTimer.Stop()
+
+	// Run the backfill workers. They will continuously pull segements from t.activeSegements and
+	// run backfill scans, returning only if they run out of segments or stopCh is closed.
+	eg, groupCtx := errgroup.WithContext(ctx)
+	for idx := 0; idx < backfillConcurrency; idx++ {
```
jgraettinger, line 63: we'll spawn three backfill workers per-binding, no matter how many bindings there are?

If I'm back-filling from 10 tables, that's 30 concurrent backfill requests, which I would expect to fight with each other for CPU. It seems pretty easy to get into situations where this would perform poorly, as compared to fixed parallelism of backfills across _all_ tables, no?

Or is the implicit assumption that there just aren't many bindings, given how discovery works (haven't gotten there yet)?

source-dynamodb/backfill.go
```diff
@@ -0,0 +1,322 @@
...
+			return nil, fmt.Errorf("scanning table %q segment %d: %w", t.tableName, segment, err)
+		}
+
+		return scanned, nil
+	}
+
+	return nil, fmt.Errorf("failed to scan segment %d of table %q after %d attempts", segment, t.tableName, maxAttempts)
+}
+
+// keyAttributeWrapper makes it possible to serialize a DynamoDB attribute value as JSON for
```
jgraettinger, line 240: another option would be to use FoundationDB tuple, also used for encoding of packed keys b/w the runtime & connectors.

CDC connectors [use it](https://github.com/estuary/connectors/blob/main/sqlcapture/tuples.go) to encode row keys that are persisted into state checkpoints.

Not a deal-breaker, but if it's easy to change i'd recommended FDB tuples for overall consistency across connectors.

source-dynamodb/capture.go
```diff
@@ -0,0 +1,473 @@
...
+			"segments":         t.totalBackfillSegments,
+			"rcus":             t.effectiveRcus,
+			"backfillDuration": backfillDuration.String(),
+		}).Info("backfilling table")
+		if err := c.backfill(ctx, t, backfillDuration); err != nil {
+			return fmt.Errorf("backfilling table '%s': %w", t.tableName, err)
+		}
+	}
+
+	// TODO(whb): If we were ever going to allow these Flow consumer shards to be split, we would
```
jgraettinger, line 262: This is all fine / no changes required, but something to ponder:

Does dynamodb offer a monotonic version or time attached to records from both streaming and backfill? If so, we could achieve eventual consistency by using the `max` reduction annotation while also concurrently doing backfills and streaming.

jgraettinger (approved):
LGTM with some comments

source-dynamodb/.snapshots/TestDiscovery-additional_table_without_stream_enabled
williamhbaker: `createdAt` sounds good, the long-winded version was a direct copy from what DynamoDB's API calls it.

source-dynamodb/.snapshots/TestDiscovery-additional_table_without_stream_enabled
williamhbaker: The partition key is part of the document - it's like any of the other columns in the DynamoDB record, except it has a schema applied to it, so it is always required to be the same type and will always be there with the record. I called it `partitionKey` for this test ([ref](https://github.com/estuary/connectors/pull/877/files#diff-1e8f7e60698434a0b474287b9be50f4d4862cd6bc4591fb54586fcfd739b12fcR27)), but it could be called anything, and would almost always be a relevant property to include in the document.

source-dynamodb/backfill.go
williamhbaker: The main thing I grappled with for this is if there was a table with a low provisioned capacity, the workers for that table might spend a lot of time just sitting around waiting for the rate limiter to refill. So with a fixed worker pool the entire pool could end up filled with workers doing nothing but waiting intermittently.

But you're right If there are a lot of tables, this wouldn't be optimal. I'll see if I can come up with another way to structure this. One thing that comes to mind is using an additional semaphore that a worker must acquire after awaiting the rate limiter for the table, but prior to actually doing the work.

source-dynamodb/capture.go
williamhbaker: Unfortunately it doesn't have any kind of version or timestamp for backfill records, and all you get from stream records is an "approximate" timestamp for the record. The closest thing we could get would be the use the approximate stream record timestamp and a wall clock time for when the backfill scan was executed, which I think would only work _most_ of the time.

source-dynamodb/.snapshots/TestDiscovery-additional_table_without_stream_enabled
jgraettinger: Ah, in that case I rescind my comment, as it seems best not to re-define an existing concept. Happy to go either way though.

source-dynamodb/.snapshots/TestDiscovery-additional_table_without_stream_enabled
jgraettinger: :+1: i retract my comment then

source-dynamodb/backfill.go
jgraettinger: ah! okay, that's important context. I'm happy to merge as-is and see where the actual problems shake out.

source-dynamodb/capture.go
jgraettinger: Blarg. Okay, sounds good then considering.

source-dynamodb/backfill.go
williamhbaker: Updated to use FDB tuples for checkpoint resume keys! It wasn't too difficult and makes things a little more compact.

PR #878, opened 2023-08-03
source-postgres-batch: Fix schemas to look nicer in the UI

willdonnelly:
**Description:**

Tweaks some `jsonschema` annotations so things look nicer in the UI (mostly by making the query template properly multiline). Also adds the first test of this connector, a snapshot test of the spec response.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/878)


willdonnelly:
YOLO merging this as I believe it's low risk

PR #879, opened 2023-08-03
source-postgres-batch: Order query results by xmin

willdonnelly:
**Description:**

During the rewriting I guess I accidentally removed the query clause `ORDER BY xmin::text::bigint`. But we still need that, since the `WHERE xmin > $1` part only works if we can assume that the last value of the previous iteration was also the largest `xmin` value. Which I think is generally true by default, but we shouldn't assume that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/879)

### DIFF
--- a/source-postgres-batch/main.go
+++ b/source-postgres-batch/main.go
@@ -99,9 +99,9 @@ func connectPostgres(ctx context.Context, configJSON json.RawMessage) (*sql.DB, 
 }
 
 const tableQueryTemplateTemplate = `{{if .IsFirstQuery -}}
-  SELECT xmin AS txid, * FROM %[1]s;
+  SELECT xmin AS txid, * FROM %[1]s ORDER BY xmin::text::bigint;
 {{- else -}}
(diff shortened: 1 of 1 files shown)


willdonnelly:
However to play devil's advocate here: if the query results are _not_ ordered by `xmin` then the last row's `xmin` value will necessarily be smaller than it would otherwise be.

And a smaller-than-necessay `xmin` cursor means that we would read some number some duplicate rows the next time we issue a polling query. And duplicate reads are something we'd like to avoid as much as possible, but they're fine in a data-correctness sense.

So lacking an `ORDER BY` clause means that _if the database returned things in a screwy order_ we might potentially see some excess amount of data, but doesn't compromise the correctness of the captured data.

And in practice, the result rows appear to be in order of increasing `xmin` anyway.

So depending on whether the presence of an explicit `ORDER BY` changes the database performance of this query, this could be an acceptable tradeoff to make. We just don't know either way.

williamhbaker (approved):
LGTM

PR #880, opened 2023-08-04
source-google-drive: initial implementation

jgraettinger:
A fairly standard `filesource` connector backed by Google Drive.

Add `Extra` field to `filesource.ObjectInfo` to enable pass-through of
driver custom data.

Add `Oauth2` field to `filesource.Source` to enable inclusion of Oauth2
config specification.

Add additional `filesource` validation checks
  * Ensure the store can be connected to
  * Ensure a path regex compiles.

**Workflow steps:**

Tested manually using `flowctl` and a local stack thus far.

Use the connector by copy / paste-ing a URL to a Google Drive folder you want to capture, such as `https://drive.google.com/drive/folders/1cxcuroqMp8628jp9I52Sl4Hxk2EqAiqP`.

The connector requires the `https://www.googleapis.com/auth/drive.readonly` scope.
I've tested using a service account only, but it re-uses the standard Google credential mechanisms.

I created a shortlink https://go.estuary.dev/source-google-drive which has not been written yet.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/880)


jgraettinger:
Thanks @williamhbaker, that's a great catch and we need to do something else here.

You're quite right that we don't get an explicit "you're not authorized" response. Instead the listing "succeeds" and just returns ... nothing.

I'm thinking to introduce a validation check that will fail if the root bucket literally returns _nothing_. It would pass if there's a folder (with no files in it), or a file that's not matched by pattern properties, or anything else, but would fail if and only if it returns nothing at all.

PR #884, opened 2023-08-04
source-postgres-batch: Select `xmin` as `txid`

willdonnelly:
**Description:**

This makes naive postgres-to-postgres sync workflows Just Work without complaining about `xmin` being a forbidden column name.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/884)


williamhbaker (approved):
LGTM

PR #887, opened 2023-08-07
source-mysql: Support `ALTER TABLE foo CHANGE ...` sometimes

willdonnelly:
**Description:**

This commit adds support for `ALTER TABLE foo CHANGE COLUMN` and `ALTER TABLE foo MODIFY COLUMN` queries in cases which don't actually rename or reposition the columns. It's possible in theory to add support for renaming/repositioning (since the effects are basically the equivalent of a DROP / ADD sequence) but I have opted to omit that part in order to get a quick fix together for the simple case.

I also factored out the datatype parsing in discovery into a separate function, because I thought it was the same logic as we use in replication event processing. It turns out the two functions are subtly different but I think the code looks nicer factored-out so I'm hesitant to revert it even though it's a completely unrelated change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/887)

### DIFF
--- a/source-mysql/.snapshots/TestAlterTable_Unsupported-capture2-fails
+++ b/source-mysql/.snapshots/TestAlterTable_Unsupported-capture2-fails
@@ -5,5 +5,5 @@
 # ================================
 # Captures Terminated With Errors
 # ================================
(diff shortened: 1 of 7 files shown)


williamhbaker (approved):
LGTM

PR #890, opened 2023-08-09
materialize-redshift: handling for long strings as load keys

williamhbaker:
**Description:**

It's possible for collection keys that are strings to be longer than the 255 bytes permitted in a Redshift TEXT column. In this case the load table needs to be created with wider VARCHAR columns to hold the full string value.

We'll keep track of the longest string key values seen for a load transaction, and if it ends up being longer than 255 bytes the load table will be created with all VARCHAR columns wide enough for the longest value seen. Also we will explicitly disable truncation of column values for the load tables, since this could lead to data inconsistency for very long string keys with identical leading 65,535 bytes. That brings along the limitation that string fields used as keys can't be any longer than 65,535 bytes, which doesn't seem totally unreasonable.

**Workflow steps:**

Materialize collections with long string fields that are used as keys.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/890)

### DIFF
--- a/materialize-redshift/.snapshots/TestSQLGeneration
+++ b/materialize-redshift/.snapshots/TestSQLGeneration
@@ -36,19 +36,6 @@ COMMENT ON COLUMN "Delta Updates".aValue IS 'A super-awesome value.
 auto-generated projection of JSON at: /aValue with inferred types: [integer]';
 --- End "Delta Updates" createTargetTable ---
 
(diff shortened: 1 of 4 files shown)


mdibaiee (approved):
LGTM, I will make sure the integration test I wrote for this case is also applied to Redshift in my branch 👍🏽

PR #891, opened 2023-08-09
source-mysql: Stringify malformed JSON column values

willdonnelly:
**Description:**

It's kind of gross to have to do this, but if the user's source database is so badly broken that their JSON columns contain non-empty but invalid JSON then we should still try and do _something_ with those values other than failing the capture.

Since the fault lies in the source data and our generated for a JSON column is `{}` we are both morally and practically able to do whatever we like with these values, so I have chosen to wrap them in an object as `{"invalidJSON": "<stringifiedValue>"}` so that at least such values are clearly identifiable.

This is completely untested because I still don't actually know how to get invalid non-empty values into JSON columns. I figure as long as CI still passes we merge this and see if it fixes the production capture that's currently hitting this issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/891)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -201,7 +201,16 @@ func (db *mysqlDatabase) translateRecordField(columnType interface{}, val interf
 				return val, nil
 			case "json":
 				if len(val) == 0 {
+					// The empty string is technically invalid JSON but null should be
+					// a reasonable translation.
 					return nil, nil
+				}
+				if !json.Valid(val) {
+					// If the contents of a JSON column are malformed and non-empty we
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

jgraettinger:
Also LGTM and IMO it can land as-is.

* An optimization, though: Use a wrapper type that also implements https://pkg.go.dev/encoding/json#Marshaler
* Have it delegate to its contained json.RawMessage
* If the delegate MarshalJson fails, _then_ use the `invalidJSON` handling.

This just avoids paying an extra cost to check it's valid JSON when it almost certainly is.

PR #892, opened 2023-08-09
source-mysql: Downgrade EXPLAIN error logging to WARN

willdonnelly:
**Description:**

The failure of our `EXPLAIN ...` queries is entirely benign. That whole piece of logic is just to help us with debugging backfill query behavior, and if the explain queries all fail we will just shrug and keep going.

So in my opinion `WARN` severity is more appropriate than `ERROR` here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/892)


williamhbaker (approved):
LGTM

PR #894, opened 2023-08-10
sqlcapture: Log document contents when JSON serialization fails

willdonnelly:
**Description:**

The hope is that the Go format `%#v` will produce a reasonable rendering of the value that's failing JSON marshalling so we can figure out how exactly things are going wrong in the first place.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/894)


jgraettinger (approved):
LGTM

PR #898, opened 2023-08-11
sqlcapture: Remove in-memory buffering of backfill rows

willdonnelly:
**Description:**

This PR removes in-memory buffering of backfill rows from all SQL Capture connectors. It accomplishes this by instead emitting backfill rows as soon as they are received from the database, but then ensuring that no checkpoints are emitted during subsequent replication processing until the next watermark is reached.

This behavior guarantees that Flow will reduce together any sequence of concurrent replication events with the backfilled row state and so no correctness violations will ever be observed in the resulting sequence of documents, however it does make a couple of edge cases slightly less clean:

- If a row is updated concurrently with the backfill query which first reaches that part of the table, such that the update replication event is part of the same Flow transaction as the initial backfill read, Flow will reduce the documents together to produce one with a `/_meta/op` of update, with no prior insert of that row.
- Likewise, deletion of a row concurrent with its backfill could result in a document first being reported as a deletion, with no prior insert of that row.

This does not compromise the correctness of the data, so we recently decided that guaranteeing these properties was not sufficient reason to keep buffering backfill results in connector memory (considering the constraints that imposed on us where we had to keep backfill chunk sizes low to avoid OOMing but then that dramatically lowered throughput in some cases).

This PR ended up including a grab-bag of other small changes as well:

- The "emitter goroutine" abstraction has been removed in favor of a very direct callback-driven stream processing of backfill events. The ability to overlap JSON serialization with other processing was not worth the increased memory usage and difficulty of debugging that it introduced.
- The MySQL backfill function has been rewritten to use a streaming query mechanism, since otherwise MySQL wouldn't benefit from these other changes.
- Some MySQL datatype handling was very slightly tweaked because the streaming backfill subtly changed how a couple of data types got serialized. See commit description for details on that one.

I have tested locally that all three SQL Capture connectors pass their full test suites after this change, including 100-iteration runs of the (now slightly modified to match the new semantics) correctness invariants stress test which verifies that we never omit rows or report incorrect state transitions.

**Workflow steps:**

Nothing should change externally.

I don't _think_ our documentation explicitly guarantees anywhere that the first observation of a row will be an insert, but if it does that should be removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/898)


williamhbaker (approved):
LGTM

With the removal of the emitter goroutine, it sounds like all of the computational work is happening on the main thread. Do you think we're leaving any performance on the table since we give connectors 2 cores to use?

willdonnelly:
> With the removal of the emitter goroutine, it sounds like all of the computational work is happening on the main thread. Do you think we're leaving any performance on the table since we give connectors 2 cores to use?

I think reaching the point where we're consistently CPU-bound in production captures would be a really nice problem to have, honestly.

And I think we have plenty of other avenues for optimization in the future which don't require the same sort of "throwing messages into a buffer to be handled asynchronously by a separate thread" structure that led to memory and debugging issues. Just off the top of my head:

- Backfilling multiple tables concurrently within a single backfill iteration, now that they won't be competing for buffer memory.
- Using a more efficient JSON serialization library which can take advantage of the structural similarity between every document from a particular table.
- Distributing tiny batches of work (tens or hundreds of documents) to worker threads for serialization and then blocking on the entire result.

PR #900, opened 2023-08-14
source-{mysql,postgres,sqlserver}: Backfill chunk size 50,000

willdonnelly:
**Description:**

Subsequent to https://github.com/estuary/connectors/pull/898 which removed the large in-memory buffer of backfill results (and thus made connector memory usage independent of backfill chunk size) and some Slack discussion about why we still don't want to set this to 1M or something absurd, we decided to bump this up to "something greater than 32,768 but less than 65,536"

Despite the fact that we've been using powers of two for this in the past there's no actual technical requirement or performance related reason that it be a power of two, and a sum of powers such as `48*1024 = 49,152` isn't as instantly recognizable anyway, so I figure we might as well make it look nice to less technical users and have instead increased the default to 50,000 here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/900)

### DIFF
--- a/source-mysql/.snapshots/TestGeneric-SpecResponse
+++ b/source-mysql/.snapshots/TestGeneric-SpecResponse
@@ -64,7 +64,7 @@
             "type": "integer",
             "title": "Backfill Chunk Size",
             "description": "The number of rows which should be fetched from the database in a single backfill query.",
-            "default": 32768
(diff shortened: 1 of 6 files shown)


jgraettinger (approved):
LGTM

Issue #837, opened 2023-07-20
materialize-redshift: optimize memory use


williamhbaker:
There is some memory overhead associated with using the S3 uploader to stream data into S3 when materializing to Redshift. For an individual binding it's not much, but when there gets to be a large number of bindings it can be significant.

One attempt to tweak this was made in https://github.com/estuary/connectors/pull/835, where the concurrency was lowered. This doesn't seem to have helped much.

There is probably something similar happening for loads and stores that is causing this relatively high memory usage proportional to the number of bindings:
- Load requests are received in an effectively random order with respect to bindings. So each staged file uploader accumulates a small amount of data to upload gradually over the course of the load phase, only flushing its respective buffer when it has gotten enough data (5 MB is the minimum buffer size). If there are many 10's of bindings, each one could have up to 5 MB of buffered data in the worst case scenario.
- Store requests are received in a specific order: For their respective binding, they are always consecutive and increasing. A single binding's uploader will constantly receive data and upload it as the buffer fills, but at the end there will be some residual in the buffer that is not flushed until _all_ stores are received.

An option for optimizing memory usage associated with Stores is to flush the uploader when the stores for the binding have moved on to the next one - we know that we will see no further stores for the "previous" binding in this case.

Loads may be more difficult because of their random order. It could be worth looking into using the lower-level S3 client rather than the S3 upload manager for this, or considering strategies like spilling to disk to handle extreme cases if that's needed.

PR #901, opened 2023-08-16
materialize-mysql: limit batch size by memory

mdibaiee:
**Description:**

- limit the size of a batch by its size in memory rather than number of rows
- according to https://github.com/estuary/connectors/issues/837, store RPCs iterate over the data in binding order, so once a binding has been passed, we can drain its batch to avoid keeping it in memory after that. I'm implementing that here to minimise memory usage of store RPC. Unfortunately for load we can't do this as that is not ordered by binding. One question I have is... is this assumption of ordering in Store RPC something we can rely on in the long-term?

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/901)


materialize-mysql/driver.go
```diff
@@ -637,12 +665,14 @@ func (d *transactor) Load(it *pm.LoadIterator, loaded func(int, json.RawMessage)
 			}
 
 			batches[it.Binding] = append(batches[it.Binding], converted...)
+			batchSize += size.Of(converted)
```
williamhbaker, line 668: I'm not terribly excited about the new dependency here and the runtime inspection of a variable's size. Since we are building these converted values into strings to write to a CSV file in the `drainBatch` functions, I think we could re-arrange things so a batch is a `[]string`, and the sizing could be based on the byte length of those strings.

More broadly, have you considered streaming directly to the reader? Squinting at the code here a little bit I think you could set up an `io.Pipe()` that would send rows directly through to the reader in `mysql.RegisterReaderHandler`. The CSV writer would write to the write end of the pipe, and the "`ReaderHandler`" would then read from the read end.

williamhbaker:
Not directly related to these changes: Does the CSV loading work if the target table's columns are in a different order than the values received from the runtime? IIRC it's possible to re-arrange columns in MySQL and this could be a problem if we are requiring that the table columns are in the same order as the CSV columns.

williamhbaker:
Another observation/question: Would it be possible to add a pre-requisite check to verify that local data loading is enabled? I've seen a case where the materialization has failed after publishing with the error `Error 3948: Loading local data is disabled; this must be enabled on both the client and server sides`.

materialize-mysql/driver.go
mdibaiee: That's actually interesting, I think now that we don't need the arguments separately anymore, we could do something like that, to use a `bytes.Buffer` to hold the CSV encoding of each batch and then measuring the size of that is also much easier 🤔

I'll look into this, thanks Will

materialize-mysql/driver.go
```diff
@@ -258,30 +273,28 @@ func (c client) AddColumnToTable(ctx context.Context, dryRun bool, tableIdentifi
...
-
-			if !dryRun {
-				if err := c.withDB(func(db *stdsql.DB) error { return sql.StdSQLExecStatements(ctx, db, []string{query}) }); err != nil {
-					return err
+		query = fmt.Sprintf(
+			"ALTER TABLE %s ADD COLUMN %s %s;",
+			tableIdentifier,
+			columnIdentifier,
+			columnDDL,
+		)
```
mdibaiee, line 281: The previous approach was failing when the identifier was quoted, because when using placeholders and parameters, identifiers should not be quoted. So instead I'm just trying out the query and if it fails with the specific error of "column exists", we just ignore that error.

materialize-mysql/driver.go
```diff
@@ -527,69 +540,105 @@ func (t *transactor) addBinding(ctx context.Context, target sql.Table, varchars
...
+	record, err := rowToCSVRecord(converted)
+	if err != nil {
+		return fmt.Errorf("error encoding row to CSV: %w", err)
 	}
 
-	writer.Flush()
-	if err := writer.Error(); err != nil {
-		return fmt.Errorf("flush writing csv: %w", err)
+	batch.w.Write(record)
+	batch.w.Flush()
```
williamhbaker, line 586: Do we need to flush after every write? Or is it sufficient to flush in drain batch?

williamhbaker (approved):
LGTM - I think it's tantalizingly close to being able to stream CSV encoded records directly to the MySQL reader without buffering anything, but it's fine for that to be a task for another day if it is working well as you have it. The batch state machine is pretty much all there with `setupBatch`, `Write` and `drainBatch`, it would just need an `io.Pipe` worked in.

materialize-mysql/driver.go
```diff
@@ -527,69 +540,105 @@ func (t *transactor) addBinding(ctx context.Context, target sql.Table, varchars
...
+	var writer = csv.NewWriter(&buff)
+
 	var readerName = fmt.Sprintf("batch_data_%s", readerSuffix)
 
 	mysql.RegisterReaderHandler(readerName,
 		func() io.Reader {
 			return &buff
 		},
 	)
-	defer mysql.DeregisterReaderHandler(readerName)
```
williamhbaker, line 583: I noticed we aren't calling this anywhere anymore - should we be? Maybe in `drainBatch`?

materialize-mysql/driver.go
```diff
@@ -527,69 +540,105 @@ func (t *transactor) addBinding(ctx context.Context, target sql.Table, varchars
...
+}
+
+type batchMeta struct {
+	// use w to write csv records for a batch
+	w *csv.Writer
+	// use buff to check how much data has been written so far
+	buff *bytes.Buffer
+}
+
+func (batch batchMeta) Write(converted []any) error {
```
williamhbaker, line 579: Does `Write` need to be exported here?

mdibaiee:
@williamhbaker i don’t think we need à `pipe` necessarily. If we want to pipe directly to MySQL I think it may be sufficient to run the query in a go routine and wait for its exit once we have reached our desired batch size, right?

all we are doing at the moment is waiting for the buffer to reach our desired size before we execute the query.

one challenge is that the `csv.Writer` class is not an `io.Writer` and its `Write` method does not tell you how many bytes it wrote (we may measure the bytes we wrote to it, but the CSV encoding itself adds some bytes for quoting, etc.). The use of the buffer is so that I can measure how much data is in memory. Also the same reason for flushing after every write, instead of keeping the writes in `csv.Writer`‘s temporary buffer, we keep it in our own buffer, allowing us to check its size.

If we stream directly to MySQL, we will also have this challenge of knowing how much data there is in memory. We could have a middle-man that reads from one side, counts, and writes the other end I suppose.

williamhbaker:
Do bytes read from the reader in `RegisterReaderHandler` get sent to MySQL as they are read, or are they buffered on the client side? I've assumed that they are sent over as they are read (perhaps with some small buffering for efficiency), and that  a synchronous pipe could be used where the write end is used as the `io.Writer` in `csv.NewWriter` and the read end is used in `RegisterReaderHandler`. There wouldn't be a need to count bytes then since very little would ever be retained in memory.

mdibaiee:
@williamhbaker I think in either case, this data is going to be buffered _somewhere_, whether that's the server or the client, it is a resource intensive process if we don't batch up the requests (we could have up to 5GB of data in one transaction, or perhaps more), that's while the performance gain growth is kind of logarithmic, we don't get much performance boost after a certain size of query, so I think I would prefer keeping a batching approach to avoid excessive memory usage on client or server.

Issue #902, opened 2023-08-16
materialize-redshift: more informative error message if ping timeout check fails


williamhbaker:
This is often the result of not having "publicly accessible" enabled for the cluster, see https://docs.aws.amazon.com/redshift/latest/mgmt/serverless-connecting.html#serverless-publicly-accessible, so our error message should reflect that, and also the fact that Estuary's IP address needs to be allowed through any firewall settings.

Issue #841, opened 2023-07-24
materialize-redshift: report errors from COPY via `stl_load_errors` table


williamhbaker:
When we have a failed COPY operation, right now the connector will output the error it receives from Redshift, which says that the job failed and to check the `stl_load_errors` table for more information.

The `stl_load_errors` does contain useful information, but that's not very helpful unless you have access to the table.

It _should_ be possible for the connector to use these error messages and actually check the `stl_load_errors` and report the relevant information. One of the columns in `stl_load_errors` is the path of the file that was attempted to be uploaded, and this is a unique UUID for every individual COPY operation. So, we should be able to query `stl_load_errors` using the file path and report back the error reason in that way.

Issue #717, opened 2023-05-16
materialize-redshift: use merge query for stores


williamhbaker:
When the redshift materialization was originally created there was no ability to use a "merge" query for storing data, like BigQuery uses.

There is now apparently a way to do a merge: https://docs.aws.amazon.com/redshift/latest/dg/r_MERGE.html

This would probably be a better way to do stores. We currently create a temporary table, load data from S3 into that temporary table, delete items that exist in the temporary table from the main table, and then copy everything from the temporary table into the main table. Using the MERGE query would save us the "delete from the main table" step which may be a performance improvement.

PR #908, opened 2023-08-18
materialize-redshift: usability improvements

williamhbaker:
**Description:**

Several improvements to `materialize-redshift`:
- Use the newly available [merge query](https://docs.aws.amazon.com/redshift/latest/dg/r_MERGE.html) for merging new data into target tables (Closes https://github.com/estuary/connectors/issues/717)
- Report errors from the `sys_load_error_detail` table if a `COPY INTO` query fails (Closes https://github.com/estuary/connectors/issues/841)
- Provide a more detail error message if there is a timeout in `Validate` when trying to ping the database (Closes https://github.com/estuary/connectors/issues/902)

**Workflow steps:**

No user-facing changes in functionality

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/908)


materialize-redshift/driver.go
```diff
@@ -877,64 +877,136 @@ func (d *transactor) commit(ctx context.Context, fenceUpdate string, hasUpdates
...
+		return fmt.Errorf("COPY INTO error for table '%s' but could not query sys_load_error_detail: %w", table, err)
+	}
+
+	log.WithFields(log.Fields{
+		"errMsg":    loadErrInfo.errMsg,
+		"errCode":   loadErrInfo.errCode,
+		"colName":   loadErrInfo.colName,
+		"colType":   loadErrInfo.colType,
+		"colLength": loadErrInfo.colLength,
+	}).Warn("loadErrInfo")
```
mdibaiee, line 956: This seems redundant given that we have a catch-all case below, or is it used for some other purpose?

mdibaiee (approved):
LGTM, just a small comment

materialize-redshift/driver.go
williamhbaker: Hopefully it is redundant all the time, I have it here just in case there's something unexpected that happens with the explicit error codes handled in the switch statement. For example, if a `1216` code actually ends up being something other than an "Input line is not valid", that way we'll have all of the information in the logs at least.

PR #909, opened 2023-08-18
source-firestore: Improve change streaming reliability (hopefully)

willdonnelly:
**Description:**

This PR makes several changes which I hope should all contribute to making change streaming more reliable.

To recap: The fundamental challenge that `source-firestore` grapples with frequently is that when we first open a `Listen` RPC for some stream, **we must catch up to the present time within 10 minutes at most** or else Firestore will cut us off and we won't even have an updated resume cursor for our trouble.

This PR does several things which should hopefully make it more likely that we'll get and remain caught up:

- A `WaitGroup` is added which keeps track of the number of replication workers which are _not_ currently caught up. Backfill workers consult this wait group before requesting any data and block unless/until it reaches zero. This should avoid situations where backfill work competes with streaming catchup for limited throughput.
- The change-streaming retry interval is decreased from 5 minutes to 1 minute. This means that in the event of a retryable interruption of streaming, there will be 5x less data accumulated before we resume.
- The change-streaming retry behavior is tweaked to more closely resemble the connector-restart behavior, by incrementing the relevant WaitGroup again and requesting a resume from the most recent consistent point. I'm actually not entirely sure how the old behavior worked at all, since whenever it retried it would send _the same exact Listen RPC_ again. Possibly this only ever worked because the connector itself got restarted occasionally as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/909)

### DIFF
--- a/source-firestore/discovery.go
+++ b/source-firestore/discovery.go
@@ -18,6 +18,7 @@ import (
 	"github.com/invopop/jsonschema"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
 	"google.golang.org/api/option"
 )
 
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #910, opened 2023-08-21
materialize-elasticsearch: revamp connector

williamhbaker:
**Description:**

A fairly comprehensive re-working of the `materialize-elasticsearch` connector. There's not much left of the original, so the full diff might be tough to make sense of. The individual commits reflect an evolution of working through these changes and may be somewhat more useful to review on a commit-by-commit basis.

The high level changes:
- Create new indices with index mappings based on the field selection for the connector. This required bringing in everything needed to produce constraints for proposed fields, map a collection's JSON types to Elasticsearch types, and handle incompatible schema changes with `Unsatisfiable` constraints.
- Improve memory efficiency by streaming & batching instead of buffering an entire transaction's documents into memory. Along with this, switch to using the [mGet API](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-get.html) for load requests, which has the notable advantage over the search API of not relying on the Elasticsearch index to be refreshed in order to read writes from the previous transaction.
- Previously the number of shards & the number of shard replicas were required to be set for each index. The number of shards for an index is significant because the number of shards cannot be changed after an index is created - meaning that Elasticsearch just doesn't allow for this to happen. The number of shard replicas can be changed after an index is created. To make the connector easier to use in a common case, providing a number of shards will now be optional, and if omitted the number of shards will be created using whatever is set as default on the cluster (his happens automatically in ES if you don't provide a specific value), which is as far as I can tell a pretty standard practice. The number of replicas is being moved to an advanced, optional endpoint-level configuration, and will similarly use the cluster default if not specified.
- Normalize Flow collection names into compatible Elasticsearch index names. Most notably this provides handling for collection names with capital letters, as capital letters aren't allowed in Elasticsearch index names.
- Add network tunnel capabilities.

**Workflow steps:**

Use the revised connector to materialize collections to Elasticsearch.

The connector docs will need updated based on these changes.

**Notes for reviewers:**

An area of uncertainty that remains for me, which I am opting not to tackle now: Should the connector be creating indices directly, or should it be creating [index aliases](https://www.elastic.co/guide/en/elasticsearch/reference/current/aliases.html)? There is an indirect mechanism for changing the number of shards in an index after creation using the [split index API](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-split-index.html). This creates a new index with a new name. I could imagine users wanting to split an index created by Flow as they accumulate more and more data into it. This would currently be cumbersome to do in a way that would keep the "original" index name, and it could be made easier using aliases.

I'm leaving the current "create an index directly" strategy unchanged since it should at least work for common cases where a materialized collection is appropriate for whatever the default number of shards is, or the user knows how many shards they want up front, but using aliases instead might be something to consider in the future depending on how the connector gets used.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/910)


materialize-elasticsearch/validate.go
```diff
@@ -0,0 +1,264 @@
...
+		constraints[field] = constraint
+	}
+
+	for _, proj := range boundCollection.Projections {
+		if !deltaUpdates && proj.IsRootDocumentProjection() {
+			// Standard updates materializations cannot change the projection field of the root
+			// document after the initial publication.
+			if proj.Field != existing.FieldSelection.Document {
+				constraints[proj.Field] = &pm.Response_Validated_Constraint{
+					Type: pm.Response_Validated_Constraint_UNSATISFIABLE,
```
psFried, line 169: I think this should maybe be `FIELD_FORBIDDEN`. It's permissible for a collection to have multiple projections of the root document. And I think we'd want to say that whichever one is on the existing field selection is required, and all others are just forbidden.

I believe (but please correct me if you think otherwise) if you return an unsatisfiable constraint, then the validation crate will fail the build, even if that field isn't explicitly included.

psFried (approved):
This looks like a _huge_ improvement. I left a few comments we should resolve, but good to merge after that.

materialize-elasticsearch/driver.go
```diff
@@ -28,124 +40,281 @@ type credentials struct {
...
 }
 
 // GetFieldDocString implements the jsonschema.customSchemaGetFieldDocString interface.
 func (resource) GetFieldDocString(fieldName string) string {
 	switch fieldName {
 	case "Index":
-		return "Name of the ElasticSearch index to store the materialization results."
+		return "Name of the Elasticsearch index to store the materialization results."
 	case "DeltaUpdates":
 		return "Should updates to this table be done via delta updates. Default is false."
```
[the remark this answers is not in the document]
williamhbaker: Not disagreeing that the description is pretty bad, but I am reluctant to spend much time working on the wording for 2 reasons:
- It's the exact same wording as all other SQL materializations use. I could see us making a more sweeping change to the wording of everything at once, but it would be sort of weird to have this one individually be different
- Currently the description doesn't show up at all for boolean fields (there's just a check box), so this is basically non-user-facing text anyway

materialize-elasticsearch/type_mapping.go
```diff
@@ -0,0 +1,124 @@
...
+			return property{Type: elasticTypeIp}
+		default:
+			if p.IsPrimaryKey {
+				return property{Type: elasticTypeKeyword}
+			} else {
+				return property{Type: elasticTypeText}
+			}
+		}
+	case pf.JsonTypeObject:
+		return property{Type: elasticTypeFlattened}
```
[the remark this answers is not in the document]
williamhbaker: Seems fair to `index: false` the root document projection. I'm not sure if it would be desirable to have some amount of indexing to make the field more useful, but it could always be indexed after the fact if that is desirable.

materialize-elasticsearch/client.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+type client struct {
+	es *elasticsearch.Client
+}
+
+func (c *client) createMetaIndex(ctx context.Context, replicas *int) error {
+	props := map[string]property{
+		"version":   {Type: elasticTypeKeyword},
+		"specBytes": {Type: elasticTypeBinary},
```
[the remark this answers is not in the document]
williamhbaker: Binary types don't get indexed. There's no reason to index the version either, so I'll prevent that from being indexed as well.

materialize-elasticsearch/validate.go
williamhbaker: Interesting point about there possibly being more than one root document projection 🤔 .

I know it can get weird if a constraint is returned for a projection that doesn't exist in the collection, as would be the case if we returned a "field required" constraint for the root document projection if that projection no longer existed.

With multiple root document projections, I have observed that the first one alphabetically gets treated as the "document" when the materialization is published. The others will only be materialized as normal fields if they are explicitly required, which I guess makes since with the constraint being "location required", and one of the root projections can satisfy that. This can also get weird if the first alphabetical root projection is explicitly excluded on an initial publication, but on a later publication that exclusion is removed - then the "document" with respect to the standard updates materialization switches to the first one in alphabetical order.

I'll have to test this out and think about it some more, but I think I am tending to agree that the only strategy that makes sense is to require a single root field, but forbid the others. This would need to happen both with the initial publication (take the first one alphabetically, forbid the rest), and on later ones (require the previously seen document field, forbid other root fields).

psFried: > This would need to happen both with the initial publication...

I don't think this would be necessary. With the initial publication, you just return `LOCATION_RECOMMENDED` for both, and the runtime will pick one. Then on later publications, we'd require the existing root projection, and forbid the others. So the user could decide whichever one they want on the initial publication, but from then on they cannot change it.

williamhbaker: Yep, after thinking about it for a bit more, I agree.

PR #912, opened 2023-08-21
source-firestore: Log number of docs read before catchup failure

willdonnelly:
**Description:**

I believe something screwy may be occurring in at least some of these cases and we're failing to catch up because Firestore is not actually sending us anything. But it's hard to be sure just from reading the debug-level logs because there's still plenty of change events from other collections, so logging the actual counter value on catchup failure should make this more blatantly obvious if it happens.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/912)


willdonnelly:
Merging a trivial debug-logging change without review

PR #913, opened 2023-08-22
sqlcapture: Tweak replication diagnostics trigger logic

willdonnelly:
**Description:**

Previously the diagnostics were run only when it's determined that the replication stream is idle, but there's another possible failure mode where we observe a low rate of changes (and thus the idle timeout never triggers) but still never see the watermark.

A flat 5-minute timeout (cancelled once we reach the watermark) more closely models the situations where diagnostics are likely to be needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/913)


willdonnelly:
Going to merge this unreviewed. It's a small debug-logging change, CI is green for `source-mysql` and `source-postgres`, and I would like to see the output on a production task ASAP.

PR #916, opened 2023-08-22
sqlcapture: Ignore views and other non-BASE TABLE entities

willdonnelly:
**Description:**

Postgres discovery logic has long ignored views and other entities whose type is not `BASE TABLE`. Unfortunately it looks like MySQL and SQL Server discovery omitted that bit and would happily produce bindings for views, which will never really work properly since they don't have replication events.

This commit changes things so that MySQL and SQL Server discovery should correctly filter these entities, but in such a way that they do not produce an error at capture time for any preexisting captures. Instead views will be silently auto-ignored at capture time.

Hopefully we can unwind some of that logic later on once nobody has any views in their catalogs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/916)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -16,19 +16,28 @@ import (
 )
 
 func (db *mysqlDatabase) DiscoverTables(ctx context.Context) (map[string]*sqlcapture.DiscoveryInfo, error) {
+	var tableMap = make(map[string]*sqlcapture.DiscoveryInfo)
+	var tables, err = getTables(ctx, db.conn)
+	if err != nil {
+		return nil, fmt.Errorf("error discovering tables: %w", err)
+	}
+	for _, table := range tables {
+		var streamID = sqlcapture.JoinStreamID(table.Schema, table.Name)
+		tableMap[streamID] = table
+	}
+
 	// Enumerate every column of every table, and then aggregate into a
 	// map from StreamID to TableInfo structs.
-	var columns, err = getColumns(ctx, db.conn)
+	columns, err := getColumns(ctx, db.conn)
 	if err != nil {
 		return nil, fmt.Errorf("error discovering columns: %w", err)
 	}
-	var tableMap = make(map[string]*sqlcapture.DiscoveryInfo)
 	for _, column := range columns {
 		// Create or look up the appropriate TableInfo struct for a given schema+name
 		var streamID = sqlcapture.JoinStreamID(column.TableSchema, column.TableName)
 		var info, ok = tableMap[streamID]
 		if !ok {
-			info = &sqlcapture.DiscoveryInfo{Schema: column.TableSchema, Name: column.TableName}
+			continue // Ignore information about excluded tables
 		}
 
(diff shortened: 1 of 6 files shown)


willdonnelly:
Tests appear to pass on MySQL, but let's wait for the full CI run to complete because this could potentially break Postgres or SQL Server if there's a mistake.

source-mysql/discovery.go
```diff
@@ -16,19 +16,28 @@ import (
 )
 
 func (db *mysqlDatabase) DiscoverTables(ctx context.Context) (map[string]*sqlcapture.DiscoveryInfo, error) {
+	var tableMap = make(map[string]*sqlcapture.DiscoveryInfo)
```
jgraettinger, line 19: you intend to eventually get rid of this separate query over tables?

instead of introducing a separate query for tables, could you instead update the columns query to return `table_type` (as you have already done below) and then condition on the column's table type?

source-mysql/discovery.go
```diff
@@ -287,11 +294,35 @@ func (db *mysqlDatabase) translateRecordField(columnType interface{}, val interf
...
+}
+
+const queryDiscoverColumns = `
+  SELECT col.table_schema, col.table_name, col.ordinal_position, col.column_name, col.is_nullable, col.data_type, col.column_type, tbl.table_type
+  FROM information_schema.columns col
+  JOIN information_schema.tables tbl
+    ON (col.table_schema = tbl.table_schema AND col.table_name = tbl.table_name)
   ORDER BY table_schema, table_name, ordinal_position;`
 
 func getColumns(ctx context.Context, conn *client.Conn) ([]sqlcapture.ColumnInfo, error) {
```
jgraettinger, line 328: should ColumnInfo below be extended with BaseTable ?

jgraettinger (approved):
LGTM but wondering if this can be simplified back into a single discovery query

source-mysql/discovery.go
willdonnelly: I started that way initially, but the column-discovery helper function currently returns a list of _column_ information objects and adding an "is this a column of a base table" flag to that felt worse. And since we're doing filtering on both the type _and_ the names of tables now, it felt cleaner to just divide up the responsibilities correctly -- now the column-discovery logic just lists every column of every table, and the table-discovery logic decides what tables we're interested in.

source-mysql/discovery.go
willdonnelly: Ah, I forgot to revert my changes to this query after I decided that just factoring out table-discovery as its own distinct thing was cleaner. I'll revert these edits here.

source-mysql/discovery.go
willdonnelly: To be clear, I think that it makes sense to keep table discovery and column listing distinct inside the database-specific discovery code here going forward, the bits I hope to revert are the existence of the "BaseTable" flag in the resulting discovery info and the conditional logic around auto-ignoring views bound to existing captures.

Issue #751, opened 2023-06-06
PostgreSQL Partition Polish


willdonnelly:
It is currently possible for `source-postgres` to capture from partitioned tables but the setup process is non-obvious and basically requires us to guide a user through it. There are a few tweaks we should make which would make it much easier for users to get this working by themselves:

1. Firstly, the `CREATE PUBLICATION` queries found in our documentation and in the connector's startup validation code should be modified to set `publish_via_partition_root = true`.
2. Second, discovery should be modified to deliberately exclude tables which are a subpartition of some other table. We should only suggest capturing from the root table.
3. Those changes will make the garden path Just Work but error reporting if the user does something wrong (such as forgetting to set that publication parameter) will be suboptimal (it will just say "table is not in publication" and not elaborate why). So as a stretch goal we might want to add some logic such that _if a table fails that check_ we proceed to check whether it's a partitioned table **and** the `publish_via_partition_root` flag is unset, and in that particular case we emit a more specific error message.

PR #917, opened 2023-08-23
source-postgres: Exclude child partitions from discovery

willdonnelly:
**Description:**

This partially addresses https://github.com/estuary/connectors/issues/751 (specifically point #2) but the other bits of polish (documentation updates and better error messages) are still needed at some point.

This commit factors out table-enumeration from column-listing for simplicity (because the query was starting to get hard to follow) and to match what we did for MySQL and SQL Server just yesterday.

The new table-enumeration query is completely different from the old one but I believe it returns the same set of tables as before, minus child partitions. I ran a couple of the discovery tests to spot-check the results and added a couple new ones, so I think it is correct.

As with the other connectors, the column-listing query no longer tries to also filter for table eligibility. Its only job is to list every column of every table in the database.

**Workflow steps:**

Child partitions of a partitioned table should no longer be returned by discovery, which means that blindly refreshing the bindings of a capture should be safe in more circumstances.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/917)

### DIFF
--- a/source-postgres/discovery.go
+++ b/source-postgres/discovery.go
@@ -24,8 +24,12 @@ const (
 
 // DiscoverTables queries the database for information about tables available for capture.
 func (db *postgresDatabase) DiscoverTables(ctx context.Context) (map[string]*sqlcapture.DiscoveryInfo, error) {
-	// Get lists of all columns and primary keys in the database
-	var columns, err = getColumns(ctx, db.conn)
+	// Get lists of all tables, columns and primary keys in the database
+	var tables, err = getTables(ctx, db.conn)
+	if err != nil {
+		return nil, fmt.Errorf("unable to list database tables: %w", err)
+	}
+	columns, err := getColumns(ctx, db.conn)
 	if err != nil {
 		return nil, fmt.Errorf("unable to list database columns: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


source-postgres/discovery.go
```diff
@@ -49,15 +53,15 @@ func (db *postgresDatabase) DiscoverTables(ctx context.Context) (map[string]*sql
...
 	for _, column := range columns {
 		var streamID = sqlcapture.JoinStreamID(column.TableSchema, column.TableName)
 		var info, ok = tableMap[streamID]
 		if !ok {
-			info = &sqlcapture.DiscoveryInfo{
-				Schema:    column.TableSchema,
-				Name:      column.TableName,
-				BaseTable: true, // PostgreSQL discovery queries only ever list 'BASE TABLE' entities
-			}
+			continue
```
williamhbaker, line 64: I think this should only happen if its a column in a system table, if I'm reading things right. If not, it might be worth logging something here since (I assume) it should never happen otherwise.

williamhbaker (approved):
LGTM

source-postgres/discovery.go
willdonnelly: A system table or a child partition, now. This is basically the same logic we already used for primary-key aggregation: if the table isn't already in the list then an earlier discovery step has decided we don't care about it and should ignore this information.

Logging whenever this happens would be unusably spammy even for debug level IMO.

PR #920, opened 2023-08-23
source-mysql: Support 'DROP DATABASE' queries for unbound schemas

willdonnelly:
**Description:**

Adds support for `DROP DATABASE` queries when they're dropping a schema which none of the active bindings are part of. Note that we'll still error out and halt the capture if a schema is dropped while we're actively capturing tables from it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/920)

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -468,6 +468,12 @@ func (rs *mysqlReplicationStream) handleQuery(schema, query string) error {
 	switch stmt := stmt.(type) {
 	case *sqlparser.CreateDatabase, *sqlparser.CreateTable, *sqlparser.Savepoint, *sqlparser.Flush:
 		logrus.WithField("query", query).Debug("ignoring benign query")
+	case *sqlparser.DropDatabase:
+		// Remember that In MySQL land "database" is a synonym for the usual SQL concept "schema"
+		if rs.schemaActive(stmt.GetDatabaseName()) {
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

source-mysql/replication.go
```diff
@@ -648,6 +654,27 @@ func (rs *mysqlReplicationStream) tableActive(streamID string) bool {
...
+		}
+	}
+	return false
+}
+
+// splitStreamID is the inverse of JoinStreamID and splits a stream name back into
+// separate schema and table name components. Its use is generally kind of a hack
+// and suggests that the surrounding plumbing is probably subtly wrong, because the
+// normalization from schema+table to stream IDs is supposed to go one way only.
+func splitStreamID(streamID string) (string, string) {
```
jgraettinger, line 673: 🤔 Go is pretty happy with map keys like:

```
struct StreamID {
   Schema string
   Table string
}
```

Perhaps there's an opportunity to defer this encoding until it's actually needed?

A `StreamID` type could have a method for producing it's `canonical.encoding`.

willdonnelly: Yeah, that's how we ought to do it at some point. Ideally as part of that refactoring the whole "case normalization" can be dropped in its entirety. It made sense originally, but now that almost everything else is properly case-aware the main thing that normalization does is to make it impossible to capture from tables with "the same name" capitalized differently.

PR #924, opened 2023-08-24
source-sqlserver: Support text primary keys without forcing `COLLATE Latin1_General_100_BIN2_UTF8` on backfills

willdonnelly:
**Description:**

Textual primary keys present an interesting challenge for SQL capture connectors. In our SQL capture's standard mode of operation we need to be able to represent "the current position of the backfill" as a single key, which replication events will sort either before (meaning that they are within the already-backfilled portion of the table) or after (meaning that we haven't reached this row yet).

If text keys were always sorted lexicographically by Unicode code-point this would be trivial, but sadly they're not. Databases love to apply weird locale-specific rules about how punctuation and non-Latin alphabetic characters will behave, sometimes they'll even chunk things up into words or otherwise sort in a non-character-based fashion.

This is an issue in all of our SQL captures. Currently `source-postgres` specifies `<columnName> COLLATE "C"` for textual key columns and `source-mysql` specifies `BINARY <columName>` and `source-sqlserver` specifies `<columnName> COLLATE Latin1_General_100_BIN2_UTF8`. However the current state of affairs is especially bad for `source-sqlserver` because the `Latin1_General_100_BIN2_UTF8` collation isn't supported in database versions more than a few years old.

In general there are two ways we could address this:

1. Give up on comparing keys to determine if a particular row has been backfilled already. This implies that we would have to emit all replication events (for tables whose key ordering can't be compared) and thus the output would no longer provide the nice property of "every row will first be observed as an `INSERT` event from the backfill", however the data will all still be correct.
2. Implement logic (plus data tables) which understands specific column collations and can encode the column values into a more trivially-comparable form.

We initially attempted option 2, writing code which can automatically generate collation-ordering data tables by interrogating the database to see how it sorts various keys in a particular collation. Unfortunately it turns out that in some cases (mainly `NVARCHAR` columns) this turns into an impossible rabbit-hole. But we already have working logic which can encode the default SQL Server collation of `SQL_Latin1_General_CP1_CI_AS` so we may as well use it.

To handle all other cases we've ended up going with option 1 by:
1. Implementing a new `"UnfilteredBackfill"` mode of operation which is basically the same as a normal backfill except that instead of doing a key comparison it unconditionally emits any replication events for that table.
2. Modifying discovery logic to flag text columns whose collation ordering cannot currently be reproduced within the connector
3. Using the `"UnfilteredBackfill"` mode when a table's key ordering is considered 'unpredictable'.

I'm not really sure whether the "collated key encoding" logic truly pulls its weight. On the one hand it's already written and it allows us to continue offering the "first row observation comes from backfill" property when the default key collation is in use, but on the other hand it's a lot of additional logic just to provide a consistency property which is kind of niche and which we can't even provide for all tables. I'd almost prefer to just stop doing it in general for all captures from SQL databases, since we're going to run into similar issues the moment we try and stop requiring a UTF-8 binary sort in the other databases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/924)


jgraettinger:
One other thought: there's an opportunity here to re-define a "Normal" backfill to actually mean an unfiltered one, and to name our current meaning of "Normal" backfill something else.

Rationale is that it allows for more dramatic updates to how backfill functions (for example, changing the collation we push down into the DB) without breaking existing capture specs that have resources with "Normal" backfill mode.

I don't think we could apply a change while a backfill is _currently_ running, but it would at least give us flexibility to change implementation where a _future_ new-"Normal" backfill might occur.

willdonnelly:
I would actually be open to going even further and abolishing "Normal" backfills in favor of the unfiltered variant all the time.

The "cleanliness" invariant that we achieve by filtering out replication events from not-yet-backfilled rows is a fairly niche requirement which is difficult to even describe (especially now that Flow is handling the reduction of replication events concurrent with the backfill of a row and so we can't guarantee that the first observation of a row won't be a deletion or update), and even a best-effort approach like you just outlined is still a bunch of additional complexity to provide a property of dubious value that AFAIK we don't even document.

Like, what's the actual description of that which we might put in our documentation? Something like:

> If your table's primary key consists of integers _or_ strings in a UTF-8 sortable collation, then it's guaranteed that you will observe a particular row state at most once. If the table contains a textual primary key which is not collated in UTF-8 bytewise order, a row state may be observed twice: once as a replication event and once when the backfill reaches it.

I'm not convinced it's worth all the effort at this point.

willdonnelly:
Now that you've mentioned it I definitely prefer defining the "Normal" backfill mode to be the unfiltered one and being more clear that the one that does additional filtering is "Precise", and I don't think this should be merged without at least making that name change.

But I think I'm even more in favor of removing these newly-named "precise backfills" as a feature entirely. Are you aware of any use-cases where it's helpful to guarantee that a given row-state will be captured at most once (and/or that after the first observation of a row a subsequent change can't be reported as an insert, I suppose) but which wouldn't have already been broken by the recent buffering changes which made it possible for the first observation to be an update or deletion?

jgraettinger:
> Are you aware of any use-cases where it's helpful to guarantee that a given row-state will be captured at most once .. ?

Yes. It seems silly to be so nit-picky about precise backfills given our Captures => Materialization product, but the place where they really matter is transformations.

When you pair a precise backfill with REPLICA IDENTITY FULL, a neat thing that happens is that every captured row is a logical two-tuple of `(retracted-state, next-state)`. Furthermore, you're **guaranteed** that every `retracted-state` you see has already been seen as the last `next-state`.

Given those guarantees, you can then implement differential updates of a continuous calculation by logically subtracting the retracted state, and then logically adding the next state.

Obviously this is an advanced concept and we'd need to build more product / packaging to make it useful to more users, but it's such a powerful and unique capability I'd be pretty loathe to give it up. Especially if all that remains to keep it is a key comparison in the connector (it must track the cursor anyway). There's a reason Netflix designed DBLog in this way.

I don't believe we've undone any of this capability in the switch to leveraging the runtime combiner by the way. So long as the connector emits backfill results _and then_ WAL events in each backfill chunk, the combiner will apply `firstWriteWins` annotations to correctly preserve the `before` / `after` state in any backfill. The only oddity it introduced is that your after state could be a deletion where there is no "before" state, but that's easily detected from looking at the single document.

willdonnelly:
> Yes [...] transformations.

That makes sense, I can see how that capability could be worth the effort in some situations.

In that case I think we should definitely rename the current "normal" backfill mode to "Precise" and the new unfiltered one to "Standard" or something. This PR includes the changes we previously discussed to leave the resource config's mode setting empty by default, so the default will actually still be "use your best judgement and pick precise where possible" but users won't need to worry about the distinction in almost all cases.

I think given that we're going to keep precise backfills at all, we should also keep the logic I've implemented here for predicting how `SQL_Latin1_General_CP1_CI_AS` columns are collated, since that's likely to be _overwhelmingly_ the most common collation in real databases and that collation alone shouldn't be a major ongoing support burden. We should plan to also add support for the `Latin1_General_100_BIN2_UTF8` collation soon I think, but then probably draw a line there and not add any others for SQL Server.

jgraettinger:
That sounds reasonable to me 👍 thanks.

sqlcapture/capture.go
```diff
@@ -541,29 +570,29 @@ func (c *Capture) handleReplicationEvent(event DatabaseEvent) error {
...
-
-	// While a table is being backfilled, events occurring *before* the current scan point
-	// will be emitted, while events *after* that point will be patched (or ignored) into
-	// the buffered resultSet.
-	if compareTuples(change.RowKey, tableState.Scanned) <= 0 {
-		if err := c.emitChange(change); err != nil {
-			return fmt.Errorf("error handling replication event for %q: %w", streamID, err)
+	if tableState.Mode == TableModePreciseBackfill {
+		// While a table is being backfilled, events occurring *before* the current scan point
+		// will be emitted, while events *after* that point will be patched (or ignored) into
```
jgraettinger, line 581: nit: comment is out of date

sqlcapture/capture.go
```diff
@@ -594,11 +623,21 @@ func (c *Capture) backfillStream(ctx context.Context, streamID string) error {
 	var lastRowKey = streamState.Scanned
 	var eventCount int
 	var err = c.Database.ScanTableChunk(ctx, discoveryInfo, streamState, func(event *ChangeEvent) error {
-		if compareTuples(lastRowKey, event.RowKey) > 0 {
-			// Sanity check that the DB must always return rows whose serialized
-			// key value is greater than the previous cursor value. This together
-			// with the check in resultset.go ensures that the scan key always
-			// increases (or else the capture will fail).
+		if streamState.Mode == TableModePreciseBackfill && compareTuples(lastRowKey, event.RowKey) > 0 {
+			// Sanity check that when performing a "standard" backfill the DB's ordering of
```
jgraettinger, line 627: nit: "standard" is now "precise".

sqlcapture/main.go
```diff
@@ -17,7 +18,7 @@ import (
 
 // Resource represents the capture configuration of a single table.
 type Resource struct {
-	Mode BackfillMode `json:"mode" jsonschema:"title=Backfill Mode,description=How the preexisting contents of the table should be backfilled. This should generally not be changed.,enum=Normal,enum=Only Changes,enum=Without Primary Key,default=Normal"`
+	Mode BackfillMode `json:"mode,omitempty" jsonschema:"title=Backfill Mode,description=How the preexisting contents of the table should be backfilled. This should generally not be changed.,default=,enum=,enum=Normal,enum=Precise,enum=Only Changes,enum=Without Primary Key"`
```
jgraettinger, line 21: This is a _little_ odd, but I see how you're using Go's default zero behavior here to avoid caring whether there's an explicit `""` or it's just not set.

Optional nit: I believe you could instead have this be `*BackfillMode` and then remove the `""` variant.
It's probably not worth changing, though, given how it impacts the switch statements that use Mode.

jgraettinger (approved):
LGTM!

A healthy chunk of the collation stuff went over my head, but what I grok'd all made good sense.

sqlcapture/main.go
willdonnelly: Yeah, I went back and forth a couple of times on how exactly to represent this property's optionality and default value. I think the current approach is the least ugly overall, but I'd be happy to change it if you'd prefer.

sqlcapture/capture.go
willdonnelly: Fixed

sqlcapture/capture.go
willdonnelly: Fixed

sqlcapture/main.go
jgraettinger: Nope, good as-is, especially if you've already considered it.

PR #927, opened 2023-08-25
source-sqlserver: Stop CI tests on first failure

willdonnelly:
**Description:**

We run a lot of tests in CI and they should all pass most of the time, but if a single one fails then it can take a while before we actually learn this and there's really no reason to be waiting that long. So if one fails, stop the test run immediately so we get quicker feedback and/or an opportunity to retry if it looks like flakiness.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/927)


williamhbaker (approved):
LGTM

Issue #687, opened 2023-04-28
Request a connector to capture from DynamoDB


ecoCart123:
### System Name

DynamoDB

### Type

Capture

### Details

I want to capture changes in DynamoDB and replicate them to Redshift

dyaffe:
Thanks for the request!  Just a quick question -- have you thought about Amazon's DMS service and is there a reason that wouldn't work for your use case?  We would love to understand that for our prioritization around building the connector.  

Just FYI, usually captures and materializations from NoSQL DB's are relatively quick for us to build (1-2 weeks).

ecoCart123:
Hi there! I had not yet, but thank you for that suggestion. I have been
using fivetran as well as investigating the usage of glue for this process,
but I wanted to ask about dynamoDB with estuary. But the noSQL
differentiation makes a lot of sense in terms of connectors

dyaffe:
Great -- let us know if you're still interested after checking out DMS.  There are some advantages if we were to do it.  A few examples:

1. Lower latency
2. Simpler schematization of your data
3. The ability to push it wherever you want (even outside of AWS)

ecoCart123:
After looking at the cost comparison I would actually love to switch to
estuary if there is a DynamoDB connector! AWS DMS looks useful but I
require support for streaming and low latency. I have been trying to use
glue streaming but having issues with permissions and the high costs of the
glue notebooks using spark, and I think estuary might be a much better
solution than 5tran.

Could I confirm that I'd like estuary to request to add the DynamoDB as a
capture? Once that is added then I think I would be able to switch our
companies data replication service from 5tran to estuary :)

dyaffe:
We don't currently support it but we can prioritize it -- I'll reach out as soon as it's available in the app if that works for you?

ecoCart123:
that sounds great thank you!

dyaffe:
Hey -- quick question for you as we are getting close to releasing these connectors:

Would you want to be able to configure a new value for the partition key and sort key when materializing to DynamoDB or would you want just a partition key?

PR #928, opened 2023-08-28
materialize-dynamodb: new connector

williamhbaker:
**Description:**

New materialization connector for DynamoDB.

The first commit is lifting the validation code that was written for `materialize-elasticsearch` in https://github.com/estuary/connectors/pull/910 into a common library that can be used by other materialization connectors, including this new DynamoDB materialization. This also includes a number of new test cases and fixes for edge cases that those test cases cover. Eventually I would like to use this validation library for SQL materializations as well.

The remaining commits are the actual code for the new materialization. A few interesting things about this materialization:
1) We're only going to support materializing collections with at most 2 collection keys. This maps to DynamoDB tables in an obvious way, since DynamoDB tables can have either a single partition key as their primary key, or a compound key with a partition key + sort key. It's not obvious how to handle collections with more than 2 keys, so the current recommendation would be to use a derivation to transform the collection into one with at most 2 keys. This could evolve if an obvious way to handle collections with more than 2 keys emerges, but that will take some user input.
2) DynamoDB row sizes are limited to 400KB, which is fairly small. The connector will only require the collection key and document fields, with all other fields optional. This will provide the maximum compatibility for collections with large documents, and allow for selecting other top-level fields to be materialized as separate "columns" if desired.
3) From some rough performance testing on a local stack, a single materialization shard can max out the 1000WCU/3000RCU limits of a DynamoDB partition without breaking a sweat. When new tables are created they only have a single partition, so a single materialization shard can easily push that to its limits. Theoretically as tables grow they get more partitions, so at some point there could be a benefit to sharding the materialization, but it probably won't be very useful for many cases. Because of this I've kept the load/store concurrency low, only pipelining requests to DynamoDB with reading the load/store iterators, which makes it easier to manage backoffs.

Closes https://github.com/estuary/connectors/issues/687

**Workflow steps:**

Use the new connector to materialize to DynamoDB.

New docs will need to be written.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/928)


mdibaiee (approved):
LGTM 👍🏽

PR #931, opened 2023-08-28
source-mysql: Ignore `(CREATE|ALTER|DROP) VIEW` statements

willdonnelly:
**Description:**

Since we don't capture from views in the first place we should be fine to just unconditionally ignore these.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/931)


jgraettinger (approved):
LGTM

PR #933, opened 2023-08-29
source-postgres: Normalize unrepresentable timestamps to zero

willdonnelly:
**Description:**

Postgres permits timestamps with years from 4713 BC to 294276 AD, but we serialize timestamps to RFC3339 strings, which can only go from the year 0 AD to 9999 AD.

In general timestamps with years >9999 aren't actually real data, they're typos like `20231`. So instead of trying to get clever, we'll normalize any out-of-range values to `0000-01-01T00:00:00Z`.

An exception is currently made for the special `infinity` timestamp value, which is still translated as `9999-12-31T23:59:59Z`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/933)


williamhbaker (approved):
LGTM

PR #934, opened 2023-08-29
source-mysql: Ignore `DROP TRIGGER` query events via regex

willdonnelly:
**Description:**

These don't parse successfully (at least in some cases) but we also don't care about them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/934)


williamhbaker (approved):
LGTM

PR #936, opened 2023-08-29
source-firestore: Restart failed streams, skipping ahead to the current time

willdonnelly:
**Description:**

Modifies `source-firestore` so that when a change stream fails to catch up in time it will:

1. Immediately skip ahead to the current time and continue streaming changes from there.
2. Set a new 'Inconsistent' flag on the resource state, indicating that the output is no longer guaranteed to be perfectly consistent with the source dataset.
3. Fire off a sneaky timer in the background which will force a connector restart after a good long while, which gives the connector an opportunity to re-establish consistency at startup (and avoids the complexity of cancelling and restarting backfills at arbitrary times).

In addition, the connector will now trigger a new backfill of any inconsistent bindings the next time the connector restarts, while also skipping the change stream ahead again to maximize the chance that it remains caught up. Backfills are now rate-limited so that at least `backfillRestartDelay` (currently 6 hours) must elapse since the start of the previous backfill before the next one begins.

**Workflow steps:**

Maybe high-volume Firestore collections will actually work reliably. We'll see.

**Notes for reviewers:**

I'm not super happy with this code, but it's what I've got right now. I think there's a decent chance it will solve the problem for the moment, and I really don't want to do the level of rewrite that seems to be necessary to actually do this better.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/936)

### DIFF
--- a/source-firestore/.snapshots/TestAddedBindingSameGroup-one
+++ b/source-firestore/.snapshots/TestAddedBindingSameGroup-one
@@ -6,5 +6,5 @@
 # ================================
 # Final State Checkpoint
 # ================================
-{"Resources":{"flow_source_tests/*/users/*/docs":{"ReadTime":"<TIMESTAMP>"}}}
+{"Resources":{"flow_source_tests/*/users/*/docs":{"Backfill":{"Completed":true,"Cursor":"","MTime":"<TIMESTAMP>","StartAfter":"<TIMESTAMP>"},"ReadTime":"<TIMESTAMP>"}}}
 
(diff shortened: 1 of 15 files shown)


willdonnelly:
The part of this PR that I'm the least happy with is the three lines that fire off a long timer and then force a connector restart. My original goal was to implement logic within the connector which could shut down any ongoing backfill of a particular collection and trigger a new one internally, and I still think that would be a good improvement to make eventually. But there wasn't time to implement it today, so I figured I should clean up what I've got and upload it for now.

williamhbaker (approved):
LGTM

PR #938, opened 2023-08-30
connectors: verify that addresses do not contain leading or trailing spaces

williamhbaker:
**Description:**

Adds a check for various connectors for endpoint address/URLs that might contain leading or trailing spaces, since this is something that has come up more than once and can be kind of hard to debug.

Creating a separate package for such a simple check feels like a bit of overkill, but it lets us have a single point to change the wording of the error message if we want to in the future.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/938)

### DIFF
--- a/go/util/util.go
+++ b/go/util/util.go
@@ -1,0 +1,16 @@
+package util
+
+import (
+	"fmt"
+	"strings"
+)
+
+// CheckEndpointSpaces checks for extraneous leading or trailing whitespace and returns an error if
+// there are any. It should be used as part of validation that an endpoint URL/address is
+// well-formed.
+func CheckEndpointSpaces(fieldName string, ep string) error {
+	if strings.TrimSpace(ep) != ep {
+		return fmt.Errorf("'%s' must not contain leading or trailing spaces", fieldName)
+	}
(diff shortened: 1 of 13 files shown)


mdibaiee:
@williamhbaker thanks! this feels like it could be useful for many other fields as well, not just endpoints. I can see it being useful for s3 bucket url, s3 host, in some cases the username, API keys, AWS credentials, aws region, etc.

I'm wondering if there is a way for us to make this easier to check against multiple properties of a connector's config. Perhaps we can use go struct field tags to mark these properties, and use `reflect` like we do in `go-schema-gen` to automatically run this check against all properties of a configuration that are marked this way?

williamhbaker:
@mdibaiee that does seem like a better way to do it. I've been thinking about this a bit and trying to come up with a way to have the check without adding it to every single connector (which is a problem with my initial implementation, too) and having to maintain consistency in that for every new connector going forward, and so far haven't come up with anything. So I'll think about it some more and hopefully come up with something better that allows it to be used more generally.

PR #939, opened 2023-08-30
source-firestore: Fix a couple dumb oversights

willdonnelly:
**Description:**

Fixes a couple of things I overlooked in https://github.com/estuary/connectors/pull/936. Specifically, the handling of a streaming catchup failure used to return from the function, so after my changes that branch continued on to the more general "report an error" handling when it was supposed to continue the loop instead. Also there was a null pointer dereference in the backfill initialization logic which could only occur on a preexisting capture from before that PR.

**Notes for reviewers:**

I'm still out today, just happened to be around for a few minutes and figured I'd throw this together since I have the context to see exactly what's wrong.


williamhbaker (approved):
LGTM

PR #943, opened 2023-09-01
materialize-mysql: convert date-time and time to server timezone

mdibaiee:
**Description:**

- Re-uses a lot of logic from `source-mysql` to fetch `time_zone` from database if `timezone` configuration is not set
- If no `timezone` can be found, ~we allow the materialization to proceed, however if we find a `date-time` or `time` field in the materialization, we will not materialize it, and we will error out explaining why. I did this to be symmetrical with `source-mysql`~ since we have `flow_published_at` now for materializations by default, so unless the user explicitly disables this column, they will have at least one `date-time` field being materialized. I think it's better if we just error out in materialization side when there is no timezone.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/943)


williamhbaker (approved):
LGTM. Made a small logging addition in the event of an error querying the database timezone.

Issue #919, opened 2023-08-23
materialize-sql: return unsatisfiable constraints for incompatible type changes to collection keys


williamhbaker:
It's possible to change the type of a collection key. SQL materializations will currently always return a "field required" constraint for existing keys, rather than checking to see if the types have changed in a compatible way.

Issue #903, opened 2023-08-16
materialize-sql: don't allow the root document projection to be changed for a standard updates materialization


williamhbaker:
This is a very similar issue to https://github.com/estuary/connectors/issues/897, and resolving the two should probably be done together.

The problem here is that the root document (by default `flow_document`) can be re-projected as a different field without our SQL materializations' validations catching it. This will result in a failure at runtime since the new column for the new root document projection will have existing rows with nothing in them, resulting in a `null` value being sent back as a load response for those keys.

Issue #897, opened 2023-08-11
materialize-sql: don't let `flow_document` be excluded from standard updates materializations


williamhbaker:
There is currently no validation check to ensure that `flow_document` is an included projection for standard updates materializations when _updating_ a materialization binding. There is a check for it when creating a new binding.

This means it is possible to publish a standard updates materialization with `flow_document` included, and then in a later publication remove `flow_document` without changing to a delta updates materialization.

A standard updates materialization with `flow_document` excluded will basically work like a delta updates materialization since we have things like [this](https://github.com/estuary/connectors/blob/7a47c94b3dee842eb939cad0e7deeef66e845f2c/materialize-postgres/sqlgen.go#L127-L129) in the queries, but that's probably not what users would expect unless they specifically opt-in to delta updates.

Issue #886, opened 2023-08-07
materialize-sql: use bigger numeric types for strings formatted as numeric values


williamhbaker:
We occasionally see very large integer or decimal numbers with schemas like `{type: string, format: integer}`, for example.

These are currently treated the same way as a `{type: integer}`, and cause problems when attempting to materialize into a destination column that cannot fit the size of the value, which is often limited by a 64 bit integer.

Where possible, we should ensure the largest possible column is used for strings formatted as numeric values, since they are more likely to contain values that do not fit into a typically sized column.

PR #946, opened 2023-09-05
materialize-sql: use different columns for numeric format strings 

williamhbaker:
**Description:**

Allows `materialize-sql` connectors to use different columns for numeric format strings via the `FlatType` `STRING` and the `WithFormat` property of `StringTypeMapper`. This is probably the way we should have always handled these numeric-strings, but it was not particularly expedient to do so until now.

Along with that, this change consolidates the constraint-producing validation code that has been developed for `materialize-elasticsearch` and `materialize-dynamodb` by bringing it into `materialize-sql` connectors. A number of constraint edge cases are fixed as a result of this. Constraints for SQL materializations will now be based on the DDL of the materialized column (ignoring nullability), rather than the strict set of JSON types the Flow collection schema includes.

Closes https://github.com/estuary/connectors/issues/919
Closes https://github.com/estuary/connectors/issues/903
Closes https://github.com/estuary/connectors/issues/897
Closes https://github.com/estuary/connectors/issues/886

**Workflow steps:**

**Notes for reviewers:**

There was not a good way that I could find to represent large, exact, decimals. AFAICT every system requires setting some specific value for _scale_, which is the number of digits after the decimal point, and this is unknown for decimal numbers in the general case. Large integers can often be represented in these numeric types using a scale of 0 (no decimals), such as a `NUMERIC(38,0)` column type allowing numbers with 38 total digits, with 8 after the decimal point.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/946)


materialize-sql/type_mapping.go
```diff
@@ -387,6 +380,30 @@ type StringTypeMapper struct {
 var _ TypeMapper = StringTypeMapper{}
 
 func (m StringTypeMapper) MapType(p *Projection) (MappedType, error) {
+	// TODO(whb): This bit of hackery is to provide backwards compatibility for materializations
```
williamhbaker, line 383: This section was moved out of `table_mapping.go`, since I think it makes more sense here. It is still necessary to have this because of one postgres materialization we have running.

willdonnelly (approved):
LGTM

PR #950, opened 2023-09-06
source-mysql: Ignore 'OtherRead' statements

willdonnelly:
**Description:**

Apparently `ANALYZE TABLE` queries show up in the binlog and then get parsed into a `*sqlparser.OtherRead`. These should be entirely ignorable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/950)


willdonnelly:
Merging without review because this is a small whitelist addition and I'd like to get a production capture unstuck that much quicker.

PR #951, opened 2023-09-07
materialize-pinecone: allow empty namespaces and enforce a single binding with empty namespace for starter plans

williamhbaker:
**Description:**

Pinecone free tier starter pods can't use namespaces, we need to not use a namespace when the target index pod is a "starter". This means that an index on a starter pod can also only have a single binding.

**Workflow steps:**

Do a materialization on a Pinecone free tier account with a starter plan pod. This will succeed if you have a single binding with an empty (empty string) for the namespace.

Pinecone materialization docs should be updated with this change.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/951)

### DIFF
--- a/materialize-pinecone/driver.go
+++ b/materialize-pinecone/driver.go
@@ -15,8 +15,10 @@ import (
 )
 
 const (
-	textEmbeddingAda002             = "text-embedding-ada-002"
-	textEmbeddingAda002VectorLength = 1536
+	textEmbeddingAda002                = "text-embedding-ada-002"
+	textEmbeddingAda002VectorLength    = 1536
+	starterPodType                     = "starter"
+	emptyNamespaceResourcePathSentinel = "FLOW_EMPTY_NAMESPACE"
 )
 
 type config struct {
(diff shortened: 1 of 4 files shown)


materialize-pinecone/driver.go
```diff
@@ -184,6 +177,34 @@ func (d driver) Validate(ctx context.Context, req *pm.Request_Validate) (*pm.Res
 		entry.Warn("Metadata field 'flow_document' will be indexed. This may not result in optimal memory utilization for the index.")
 	}
 
+	if indexDescribe.Database.PodType == starterPodType {
+		// "Starter" pod types on the free tier cannot use namespaces. Namespaces are required for
+		// differentiating multiple bindings in the same index, so starter pods cannot have more
+		// than 1 binding, and the binding can't have a namespace set.
+		if len(req.Bindings) > 1 {
+			return nil, fmt.Errorf(
+				"cannot materialize more than 1 binding to index '%s' which has pod type '%s'. consider materializing fewer bindings, or using an index with a different pod type",
```
jgraettinger, line 186: users who are seeing this are probably just as new to pinecone as they are to us. what about:

"Your Pinecone index is of type {PodType}, which cannot use Pinecone's namespace feature. This materialization is thus unable to have more than one bound collection. Consider removing bindings, or upgrade your Pinecone plan."

materialize-pinecone/driver.go
```diff
@@ -184,6 +177,34 @@ func (d driver) Validate(ctx context.Context, req *pm.Request_Validate) (*pm.Res
...
+		}
+
+		if len(req.Bindings) == 1 {
+			res, err := resolveResourceConfig(req.Bindings[0].ResourceConfigJson)
+			if err != nil {
+				return nil, fmt.Errorf("resolving resource config for single starter plan index binding: %w", err)
+			}
+			if res.Namespace != "" {
+				return nil, fmt.Errorf(
+					"namespace must be left blank when materializing to index '%s' with pod type '%s', since these pod types do not support namespaces: namespace was '%s'",
```
jgraettinger, line 199: "Your Pinecone index is of type {PodType}, which cannot use Pinecone's namespace feature. You can still proceed by removing the Pinecone Namespace from your bound collection's Resource Configuration, or by upgrading your Pinecone plan."

jgraettinger (approved):
LGTM % rewording user-facing errors. thanks!

PR #952, opened 2023-09-08
materialize-mongodb: don't use transactions

williamhbaker:
**Description:**

In practice we have seen problems with using MongoDB transactions for storing documents related to a
low default transaction timeout on MongoDB servers, and high memory usage on the server for large
transactions, which is probably why there is a fairly low default transaction timeout.

This converts the connector to an "at least once" style of materialization, forgoing the use of
transactions. With these changes, the materialization will work with unbounded Flow transaction
sizes, and be able to achieve "at least once" semantics in the general case, and "effectively once"
when idempotent reductions are used.

This PR also updates the connector to use bulk operations for loads and stores, which is a significant
performance improvement. Loads are done using the `find` operation with a reasonably sized list of 
IDs to retrieve. Stores are done using the `bulkWrite` operation.

**Workflow steps:**

`materialize-mongodb` docs should be updated.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/952)

### DIFF
--- a/materialize-mongodb/transactor.go
+++ b/materialize-mongodb/transactor.go
@@ -6,9 +6,7 @@ import (
 	"fmt"
 	"math"
 
-	pf "github.com/estuary/flow/go/protocols/flow"
 	pm "github.com/estuary/flow/go/protocols/materialize"
-	"go.gazette.dev/core/consumer/protocol"
 
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
@@ -16,10 +14,6 @@ import (
 )
 
 type transactor struct {
-	materialization string
-	fenceCollection *mongo.Collection
-	fence           *fenceRecord
-
 	client   *mongo.Client
 	bindings []*binding
 }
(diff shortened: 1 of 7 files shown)


materialize-mongodb/transactor.go
```diff
@@ -6,143 +6,279 @@ import (
...
+	ctx context.Context,
+	loaded func(i int, doc json.RawMessage) error,
+	batches <-chan loadBatch,
+) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case batch := <-batches:
+			if batch.keys == nil {
```
willdonnelly, line 182: Trivial nit: If you're just checking whether the channel is closed I think `case batch, ok := <-batches: if !ok ...` is more idiomatic. But this definitely works so feel free to ignore.

willdonnelly (approved):
LGTM

Some of this went over my head, but I don't see any obvious issues.

One high-level question though: is this expected to work seamlessly for preexisting MongoDB materializations?

materialize-mongodb/transactor.go
williamhbaker: Oh nice, that is better. It felt a little weird checking for the zero-value.

williamhbaker:
> One high-level question though: is this expected to work seamlessly for preexisting MongoDB materializations?

It will cause them to start over from the beginning and re-backfill. Considering how the materialization worked previously it is hard to imagine this being an issue, but I will check any existing materializations that are active and see if it warrants a version bump or something else.

dyaffe:
I think it's safe.  We don't really have any meaningful ones live yet which is likely why we've never seen this issue.

jgraettinger (approved):
LGTM

PR #955, opened 2023-09-13
source-mongodb: incremental backfill

mdibaiee:
**Description:**

- A new checkpoint structure which allows distinguishing between Backfill and Streaming states, and tracks last backfilled item, timestamp of backfill start, and resume token for streaming
- Converts old checkpoints to the new model
- Use `RawValue` for storing `BackfillLastId` so that we can be sure that we are not losing the original value of the `_id` during decoding and encoding processes of mongodb driver and/or checkpoint

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/955)

### DIFF
--- a/source-mongodb/main.go
+++ b/source-mongodb/main.go
@@ -100,8 +100,8 @@ func (d *driver) Connect(ctx context.Context, cfg config) (*mongo.Client, error)
 		return nil, err
 	}
 
-	if diff, err := OplogTimeDifference(ctx, client); err != nil {
-		return nil, fmt.Errorf("could not read oplog time difference: %w", err)
+	if diff, err := oplogTimeDifference(ctx, client); err != nil {
+		log.WithField("err", err).Warn("could not read oplog time difference")
 	} else {
 		log.WithField("timediff", diff).Debug("read oplog time difference")
 
--- a/source-mongodb/main.go
+++ b/source-mongodb/main.go
@@ -24,6 +24,8 @@ import (
 )
 
 const (
+	// Minimum oplog time difference: see the comment on OplogTimeDifference in
+	// oplog.go
 	minOplogTimediff = 24 * 60 * 60 // 24 hours, in seconds
 )
 
--- a/source-mongodb/main.go
+++ b/source-mongodb/main.go
@@ -14,12 +14,17 @@ import (
 	boilerplate "github.com/estuary/connectors/source-boilerplate"
 	pc "github.com/estuary/flow/go/protocols/capture"
 	pf "github.com/estuary/flow/go/protocols/flow"
+	log "github.com/sirupsen/logrus"
 
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 	"go.mongodb.org/mongo-driver/mongo/readpref"
 	mongoDriver "go.mongodb.org/mongo-driver/x/mongo/driver"
+)
+
+const (
+	minOplogTimediff = 24 * 60 * 60 // 24 hours, in seconds
 )
 
 type resource struct {
@@ -91,6 +96,16 @@ func (d *driver) Connect(ctx context.Context, cfg config) (*mongo.Client, error)
 	client, err := mongo.Connect(ctx, options.Client().ApplyURI(cfg.ToURI()))
 	if err != nil {
 		return nil, err
+	}
+
+	if diff, err := OplogTimeDifference(ctx, client); err != nil {
+		return nil, fmt.Errorf("could not read oplog time difference: %w", err)
+	} else {
+		log.WithField("timediff", diff).Debug("read oplog time difference")
+
+		if diff < minOplogTimediff {
+			log.Warn(fmt.Sprintf("the current time difference between oldest and newest records in your oplog is %d seconds. This is smaller than the minimum of 24 hours. Please resize your oplog to be able to safely capture data from your database: https://go.estuary.dev/NurkrE", diff))
+		}
 	}
 
 	// Any error other than an authentication error will result in the call to Ping hanging until it
--- a/source-mongodb/main.go
+++ b/source-mongodb/main.go
@@ -101,10 +101,8 @@ func (d *driver) Connect(ctx context.Context, cfg config) (*mongo.Client, error)
 	}
 
 	if diff, err := oplogTimeDifference(ctx, client); err != nil {
-		log.WithField("err", err).Warn("could not read oplog time difference")
+		return nil, fmt.Errorf("could not read oplog, access to oplog is necessary: %w", err)
 	} else {
-		log.WithField("timediff", diff).Debug("read oplog time difference")
-
 		if diff < minOplogTimediff {
 			log.Warn(fmt.Sprintf("the current time difference between oldest and newest records in your oplog is %d seconds. This is smaller than the minimum of 24 hours. Please resize your oplog to be able to safely capture data from your database: https://go.estuary.dev/NurkrE", diff))
 		}
--- a/source-mongodb/oplog.go
+++ b/source-mongodb/oplog.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+type oplogRecord struct {
+	Ts primitive.Timestamp `bson:"ts"`
+}
+
+const tsProperty = "ts"
+
+func OplogTimeDifference(ctx context.Context, client *mongo.Client) (uint32, error) {
+	var db = client.Database("local")
+	var oplog = db.Collection("oplog.rs")
+
+	// get latest record
+	var opts = options.FindOne().SetSort(bson.D{{tsProperty, SortDescending}})
+	var latest oplogRecord
+	if err := oplog.FindOne(ctx, bson.D{}, opts).Decode(&latest); err != nil {
+		return 0, fmt.Errorf("querying latest oplog record: %w", err)
+	}
+
+	// get oldest record
+	opts = options.FindOne().SetSort(bson.D{{tsProperty, SortAscending}})
+	var oldest oplogRecord
+	if err := oplog.FindOne(ctx, bson.D{}, opts).Decode(&oldest); err != nil {
+		return 0, fmt.Errorf("querying oldest oplog record: %w", err)
+	}
+
+	return latest.Ts.T - oldest.Ts.T, nil
+}
--- a/source-mongodb/oplog.go
+++ b/source-mongodb/oplog.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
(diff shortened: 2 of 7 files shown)


source-mongodb/pull.go
```diff
@@ -140,30 +179,32 @@ type changeEvent struct {
...
+
 	var opts = options.ChangeStream().SetFullDocument(options.UpdateLookup)
-	if resumeToken != nil {
-		opts = opts.SetResumeAfter(resumeToken)
-	} else if backfillFinishedAt != nil {
-		opts = opts.SetStartAtOperationTime(backfillFinishedAt)
+	if state.StreamResumeToken != nil {
+		opts = opts.SetResumeAfter(state.StreamResumeToken)
+	} else if !state.BackfillStartedAt.IsZero() {
+		var startAt = &primitive.Timestamp{T: uint32(state.BackfillStartedAt.Unix()) - 1, I: 0}
```
williamhbaker, line 204: Do we need to go back in time any more than 1 second? I'm not sure how MongoDB streams work but am wondering if there is some "approximation" to when the oplog events are recorded that could be off by more than 1 second from our recorded wall clock time of when the backfill began, or even just drift between clocks of the server vs. connector.

source-mongodb/pull.go
```diff
@@ -255,54 +295,123 @@ func (c *capture) ChangeStream(ctx context.Context, client *mongo.Client, bindin
...
+			checkpointJson, err := json.Marshal(checkpoint)
+			if err != nil {
+				return fmt.Errorf("encoding checkpoint to json failed: %w", err)
+			}
+
+			if err = c.Output.Checkpoint(checkpointJson, true); err != nil {
+				return fmt.Errorf("output checkpoint failed: %w", err)
+			}
+
+			if diff, err := OplogTimeDifference(ctx, client); err != nil {
```
williamhbaker, line 381: Conceptually here I think we want to know if the backfill is falling behind how fast it needs to be going in order to not lose the oplog event it needs when starting streaming. Is that something we can conclusively check for in `ChangeStream()`? It kind of looks like a `startAt` can be set, but if that `startAt` is earlier than anything in the oplog there won't be an error. So the scenario is that in the time the backfill took to complete, the change events we needed around `startAt` have fallen off.

It looks a little bit like we need to grab an initial resume token from the change stream prior to starting the backfill from some point earlier than the backfill began, and then resume from that token when transitioning into the stream phase. That way if the token is gone we will error and know that there is missing data. Then, we could periodically check to make sure that resume token still exists while the backfill is on-going to be able to fail fast.

source-mongodb/pull.go
```diff
@@ -255,54 +295,123 @@ func (c *capture) ChangeStream(ctx context.Context, client *mongo.Client, bindin
...
+				if diff < minOplogTimediff {
+					log.Warn(fmt.Sprintf("the current time difference between oldest and newest records in your oplog is %d seconds. This is smaller than the minimum of 24 hours. Please resize your oplog to be able to safely capture data from your database: https://go.estuary.dev/NurkrE", diff))
+				}
 			}
 		}
 	}
 
-	if err = c.Output.Checkpoint([]byte("{}"), true); err != nil {
-		return nil, fmt.Errorf("sending checkpoint failed: %w", err)
+	state.Status = StatusStreaming
```
williamhbaker, line 395: Do we need to check `cursor.Err` after `cursor.Next` returns `false`? Maybe not if an error would always be surfaced in `cursor.Decode`.

source-mongodb/pull.go
```diff
@@ -140,30 +179,32 @@ type changeEvent struct {
...
+	}}}}}
+
 	var opts = options.ChangeStream().SetFullDocument(options.UpdateLookup)
-	if resumeToken != nil {
-		opts = opts.SetResumeAfter(resumeToken)
-	} else if backfillFinishedAt != nil {
-		opts = opts.SetStartAtOperationTime(backfillFinishedAt)
+	if state.StreamResumeToken != nil {
+		opts = opts.SetResumeAfter(state.StreamResumeToken)
+	} else if !state.BackfillStartedAt.IsZero() {
```
williamhbaker, line 203: Will `state.BackfillStartedAt.IsZero` ever be `true`?

source-mongodb/pull.go
```diff
@@ -96,16 +114,14 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
...
-			if backfillFinishedAt, err = c.BackfillCollection(ctx, client, idx, res); err != nil {
+		if resState.Status == StatusBackfill {
+			if err = c.BackfillCollection(ctx, client, idx, res, &resState); err != nil {
 				return fmt.Errorf("backfill: %w", err)
 			}
 		}
 
 		eg.Go(func() error {
-			return c.ChangeStream(egCtx, client, idx, res, backfillFinishedAt, resState)
+			return c.ChangeStream(egCtx, client, idx, res, &resState)
```
williamhbaker, line 124: I think there is a race between emitting documents & checkpoints for concurrent bindings here: If there are 2 `ChangeStream` workers running concurrently, their calls to `c.Output.Documents` and `c.Output.Checkpoint` can be interleaved with each other, and also to the running `c.BackfillCollection` call. `c.Output.Documents` and `c.Output.Checkpoint` do take a lock, but it is only for their specific invocation and it is released when they return.

An example inconsistency this could lead to is:
1. Backfill collection 'A' emits some documents
2. ChangeStream collection 'B' emits some documents
3. Backfill collection 'A' emits a checkpoint; this commits the documents from (1) and (2) to flow
4. The connector crashes before ChangeStream collection 'B' emits its state checkpoint
5. When the connector starts back up, ChangeStream collection 'B' resumes from an older state checkpoint and emits duplicate documents to ones that were previously emitted and committed to flow

It could be possible to prevent this by having a kind of "emit documents and checkpoints" function that emits both under a common lock.

williamhbaker (commented):
Reviewed with some comments.

A more general thought: It would be nice to have a test exercising all of the backfill & change stream resumption logic. I have found tests like [these](https://github.com/estuary/connectors/blob/b23219f576fc2cf1a02dcb1342a024aa5ab5291b/source-dynamodb/main_test.go#L21-L118) producing a snapshot like [this](https://github.com/estuary/connectors/blob/main/source-dynamodb/.snapshots/TestCapture) to be helpful in demonstrating the correctness for a similar system.

source-mongodb/main.go
```diff
@@ -93,6 +100,16 @@ func (d *driver) Connect(ctx context.Context, cfg config) (*mongo.Client, error)
 		return nil, err
 	}
 
+	if diff, err := OplogTimeDifference(ctx, client); err != nil {
```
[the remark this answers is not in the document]
mdibaiee: > I'm assuming that wouldn't work here since we're not exactly checking for the actual oplog retention period, and rather the inferred retention period from the oldest and newest records - and that could be skewed if there are very infrequent updates. Curious if you have any other thoughts on this though.

Yep this is the issue, we can't check an actual retention period for Mongo versions prior to 4.4, and after 4.4 it is an optional parameter that can be set: https://www.mongodb.com/docs/manual/core/replica-set-oplog/#minimum-oplog-retention-period

I don't really know if we need to support Mongo < 4.4, I've assumed the answer is yes. If the answer is no, this is simpler because we can just ask the user to set [storage.oplogMinRetentionHours](https://www.mongodb.com/docs/manual/reference/configuration-options/#mongodb-setting-storage.oplogMinRetentionHours) and we can check its value to make sure it is sufficient. Perhaps we can have both implementations if we want to support < 4.4 versions.

source-mongodb/oplog.go
```diff
@@ -0,0 +1,44 @@
...
+// it is used as a measure of how long do oplog records stay in the oplog.
+// Note that this value can fluctuate depending on the density of oplog
+// records during a period, so continuously monitoring this value is necessary
+// and a single value is not representative
+func OplogTimeDifference(ctx context.Context, client *mongo.Client) (uint32, error) {
+	var db = client.Database("local")
+	var oplog = db.Collection("oplog.rs")
+
+	// get latest record
+	var opts = options.FindOne().SetSort(bson.D{{tsProperty, SortDescending}})
```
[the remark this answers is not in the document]
mdibaiee: A single record would just return `0` as the time difference, and no records would error.

I think having no records is _theoretically_ possible. How could we handle it? We could log a warn that we could not read the oplog time difference if we can't read it, instead of erroring.

source-mongodb/pull.go
mdibaiee: The 1 second here is only because `primitive.Timestamp` here only supports seconds precision, and I wanted to have the safety of one second as we have loss of precision from the original timestamp (I think lost precisions are rounded, so if we have >50 milliseconds, we end up rounding up a full second).

source-mongodb/pull.go
mdibaiee: We can check to see if the time elapsed since start of backfill is approaching `OplogTimeDifference`, in which case we are getting close to losing the oplog records to start a `ChangeStream` since beginning of backfill

source-mongodb/pull.go
mdibaiee: @williamhbaker it shouldn't be

source-mongodb/pull.go
mdibaiee: We do it a few lines below

williamhbaker: Ah I missed that. But we emit the checkpoint first? That seems like it could (maybe) result in a case where we emit a checkpoint showing that the backfill is done if `cursor.Next()` returns false with an error, and the backfill isn't really done yet.

source-mongodb/oplog.go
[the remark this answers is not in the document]
williamhbaker: Yeah I think we'd have to log a warning instead of erroring in that case. I could imagine a static database that had been populated longer than `{oplog_retention_period}` ago with no changes since then, and we would want to be able to capture from such a database.

source-mongodb/pull.go
williamhbaker: I see - still wondering if we should have a longer safety factor though: For a case where MongoDB's notion of what time it is differs from what time our connector thinks it is a significant amount, to make sure that we don't miss any changes that happen right around the exact time that the backfill begins. It would be pretty unlikely, and would basically have to be a change in the first key in sorted order happening at almost the exact same time as the backfill starts, but maybe a ~5 minute buffer would be worth it to make that basically impossible.

source-mongodb/pull.go
mdibaiee: @williamhbaker good point!

source-mongodb/oplog.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// get oldest record
+	var opts = options.FindOne().SetSort(bson.D{{tsProperty, SortAscending}})
+	var oldest oplogRecord
+	if err := oplog.FindOne(ctx, bson.D{}, opts).Decode(&oldest); err != nil {
+		return fmt.Errorf("querying oldest oplog record: %w", err)
+	}
+	var oldestTs = time.Unix(int64(oldest.Ts.T), 0)
+
+	if oldestTs.After(ts) {
```
williamhbaker, line 63: Will this work if:
1. A backfill is started on a collection that has nothing in the oplog
2. During the backfill, there's a change on the table, resulting in an oplog entry with a timestamp after the backfill started
3. When this check runs, it will see  the oplog timestamp coming after the backfill started timestamp

I don't have much knowledge on how MongoDB oplogs work so maybe this isn't an issue at all, if hypothetically there are periodic "heartbeat" types of changes always recorded on some frequency or something like that.

williamhbaker (approved):
LGTM, assuming the "querying the oplog outside of its retention window" thing is resolved. I left a question about that latest commit.

source-mongodb/oplog.go
williamhbaker: One final question (I think 😃 ): Will this cause an error if the oplog is completely empty? That could mean that nothing has ever been written to the oplog, which would be fine, and we wouldn't want to fail on that. Or it could mean that the record(s) we really need to make sure we captured are now gone, which we would want to fail on.

mdibaiee: > Will this work if:
 >   A backfill is started on a collection that has nothing in the oplog
    During the backfill, there's a change on the table, resulting in an oplog entry with a timestamp after the backfill started
    When this check runs, it will see the oplog timestamp coming after the backfill started timestamp

We will error in this case as well. I thought about this a bit. I think one way to resolve this is to not set `backfill_started_at` unless we have actually captured anything. If we have captured any documents during backfill, there _must_ be records in the oplog, and so we must be able to find the timestamp, however if we can't find anything to capture, it is possible that there is nothing in the oplog.

mdibaiee: Okay, actually on further investigation, it seems it is not possible to have an empty oplog: during the initialization process, MongoDB itself writes configuration for the replicaset, which generates some oplog records, so any MongoDB instance that has replicasets enabled (a pre-requisite for our capture), will have some oplog records available.

mdibaiee: > Also - is access to the oplog collection granted under the same circumstances as access to a change stream?

No, I think with this new change we require access to the oplog collection, so I'm adding a check for that 👍🏽

mdibaiee:
@williamhbaker also added tests similar to that of dynamodb, thanks!

Issue #864, opened 2023-07-28
source-kinesis: use localstack for integration tests


williamhbaker:
We should use the localstack AWS emulator for kinesis integration tests to speed up their execution and reduce costs

Reference https://github.com/estuary/connectors/pull/663 for doing this with the `source-s3` integration tests

PR #956, opened 2023-09-13
source-kinesis: use localstack for integration tests 

williamhbaker:
**Description:**

- Uses `localstack` docker containers for Kinesis integration tests instead of actually calling to the real AWS API.

- Moves the "endpoint" configuration to an advanced option, which is consistent to other AWS captures
and materializations. This is not a field that a user is ever likely to use. There aren't any
existing kinesis captures that use this field, so this is not a breaking change.

- Also cleans up and fixes the set of capture tests for Kinesis, which had gotten into quite a state
of disarray.

Closes https://github.com/estuary/connectors/issues/864

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/956)

### DIFF
--- a/source-kinesis/connection.go
+++ b/source-kinesis/connection.go
@@ -12,12 +12,16 @@ import (
 )
 
 // Config represents the fully merged endpoint configuration for Kinesis.
-// It matches the `KinesisConfig` struct in `crates/sources/src/specs.rs`
 type Config struct {
-	Endpoint           string `json:"endpoint,omitempty" jsonschema:"title=AWS Endpoint,description=The AWS endpoint URI to connect to (useful if you're capturing from a kinesis-compatible API that isn't provided by AWS)"`
 	Region             string `json:"region" jsonschema:"title=AWS Region,description=The name of the AWS region where the Kinesis stream is located"`
 	AWSAccessKeyID     string `json:"awsAccessKeyId" jsonschema:"title=AWS Access Key ID,description=Part of the AWS credentials that will be used to connect to Kinesis"`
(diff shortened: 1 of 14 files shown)


source-kinesis/connection.go
```diff
@@ -12,12 +12,16 @@ import (
...
-	Endpoint           string `json:"endpoint,omitempty" jsonschema:"title=AWS Endpoint,description=The AWS endpoint URI to connect to (useful if you're capturing from a kinesis-compatible API that isn't provided by AWS)"`
 	Region             string `json:"region" jsonschema:"title=AWS Region,description=The name of the AWS region where the Kinesis stream is located"`
 	AWSAccessKeyID     string `json:"awsAccessKeyId" jsonschema:"title=AWS Access Key ID,description=Part of the AWS credentials that will be used to connect to Kinesis"`
 	AWSSecretAccessKey string `json:"awsSecretAccessKey" jsonschema:"title=AWS Secret Access Key,description=Part of the AWS credentials that will be used to connect to Kinesis"`
+
+	Advanced advancedConfig `json:"advanced,omitempty"`
+}
+
+type advancedConfig struct {
+	Endpoint string `json:"endpoint,omitempty" jsonschema:"title=AWS Endpoint,description=The AWS endpoint URI to connect to (useful if you're capturing from a kinesis-compatible API that isn't provided by AWS)"`
```
willdonnelly, line 24: Moving this property into the advanced config will break any existing configs. Are we prepared for whatever breakage that might cause?

willdonnelly (approved):
LGTM

source-kinesis/connection.go
williamhbaker: Yep there aren't any real usages of this connector yet, any the only two tasks that exist (one is our own testing task, and another is a disabled & probably abandoned one) don't have this field set so it shouldn't actually break them.

PR #957, opened 2023-09-13
source-mongodb: backfill when binding has no state

mdibaiee:
**Description:**

- Instead of a global condition for starting backfill (this was never intentional AFAIK), check whether that specific binding has state or not
- Key the checkpoint with collection name and resource config, so that by changing the collection name it is possible to trigger a re-backfill

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/957)

### DIFF
--- a/source-mongodb/pull.go
+++ b/source-mongodb/pull.go
@@ -67,34 +67,19 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
 		if err := pf.UnmarshalStrict(binding.ResourceConfigJson, &res); err != nil {
 			return fmt.Errorf("parsing resource config: %w", err)
 		}
-		// the deprecated way of storing checkpoints
 		var resState, _ = prevState.Resources[resourceId(res)]
-
-		// the new way of storing checkpoints
-		var bindingState, _ = prevState.Resources[bindingId(binding, res)]
-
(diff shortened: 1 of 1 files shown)


williamhbaker:
We had some additional discussion about how to handle this:
- Long-term we do want changing the target Flow collection of a binding to trigger a re-backfill of the collection, but we don’t want to add that special behavior to mongodb right now since it would be different than other captures & want to hold off for a more holistic solution, whatever that might end up looking like.
- For the moment, we will add logic to the capture to clear a binding’s state when it is removed and re-added, which is how other captures work. That way a re-backfill can be triggered by removing & re-adding the binding to the spec.

I pushed a couple of commits to this branch to accomplish the checkpoint update upon removing & re-adding a binding, also keeping the fix check for individual resource state vs. global state for doing a backfill of a collection. This worked in local testing:
1. On the current `dev` image, create a capture with 1 MongoDB collection and let that backfill and stream a change
2. Still on the current `dev` image, add a binding to that capture and stream a change. Only the streamed change document is captured
3. Switch to the image built from this branch, and remove the newly added binding. The log line `binding removed from catalog` is output as expected upon removing the binding
4. Add the binding back in, and the collection is re-backfilled

willdonnelly (approved):
LGTM

PR #958, opened 2023-09-13
source-mysql: Ignore ALTER COLUMN table alterations

willdonnelly:
**Description:**

A while back I made a list of all the table alterations which seemed like they could possibly be an issue, and made those all explicit errors so the default could be to ignore unhandled alterations (since there are a ton of benign ones).

This list was a little bit over-broad, and as far as I can see there is no way for an `ALTER COLUMN` clause to ever be a problem (since they just modify column defaults and a couple of other properties we don't care about) so these should be ignored as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/958)

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -546,7 +546,7 @@ func (rs *mysqlReplicationStream) handleAlterTable(stmt *sqlparser.AlterTable, q
 		switch alter := alterOpt.(type) {
 		// These should be all of the table alterations which might possibly impact our capture
 		// in ways we don't currently support, so the default behavior can be to log and ignore.
-		case *sqlparser.AlterColumn, *sqlparser.RenameColumn:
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

jgraettinger:
Do we ever care about a column's type being altered?

williamhbaker:
> Do we ever care about a column's type being altered?

That comes through in a `sqlparser.ChangeColumn` event I believe, which is handled in a later switch branch.

willdonnelly:
Yeah, as far as I can tell `ALTER COLUMN` is just about fiddling with default values (and some sort of "invisible" flag) which isn't something we need to keep track of. Column name/type changes are done with `CHANGE COLUMN` or `MODIFY COLUMN` clauses.

PR #960, opened 2023-09-14
materialize-sql: don't use numeric types for numeric formatted strings that are primary keys

williamhbaker:
**Description:**

Fixes an oversight from #946: We historically have not converted strings with integer or number format to their equivalent numeric type if they are collection keys, and need to continue to not do that. This is mostly because "number" types are usually converted to floats, which cannot be used as table primary keys.

It could possibly be argued that strings formatted as integers should be able to be materialized as integer primary keys. There's an option for that for (new) connectors going forward, but for now we need to keep the behavior the same to maintain compatibility. It may also be possible to change this later with some kind of table introspection, where the connector is able to convert types based on the actual table columns, but that would take significantly more thought.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/960)


willdonnelly (approved):
LGTM

PR #962, opened 2023-09-15
source-postgres: Advanced config option for discovery schemas

willdonnelly:
**Description:**

This commit adds a new advanced config option containing a list of strings. If unset the connector will continue to operate as always, but if this list is non-empty then discovery will only return tables in schemas which are part of that list.

I have confirmed that this doesn't break any of our existing test suite and also verified manually that the filtering logic works as expected.

I'm not entirely happy with the field title/description text, I just wrote the first thing that came to mind so we could get this out. I'm also not entirely sure if this array of strings will look good in the UI form or if a single text input for a comma-separated list of schema names might be better.

**Workflow steps:**

Nothing should change by default, but users can now explicitly select the schemas that they want to discover tables within.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/962)

### DIFF
--- a/source-postgres/.snapshots/TestGeneric-SpecResponse
+++ b/source-postgres/.snapshots/TestGeneric-SpecResponse
@@ -68,6 +68,14 @@
             ],
             "title": "SSL Mode",
             "description": "Overrides SSL connection behavior by setting the 'sslmode' parameter."
+          },
+          "discover_schemas": {
+            "items": {
+              "type": "string"
+            },
+            "type": "array",
+            "title": "Discovery Schema Selection",
+            "description": "If this is specified only tables in the selected schema(s) will be automatically discovered. Leave blank to discover tables from all schemas."
           }
         },
         "additionalProperties": false,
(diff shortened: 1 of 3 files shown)


travjenkins (approved):
Did a quick review and seems fine to me. I won't be able to catch any specific `go` issues though.

williamhbaker (approved):
The rendering of array config options in the UI currently isn't great. This is how it would look as an empty array:

image

The biggest problem I think is that it doesn't show the description of the configuration.

An array does seem like the right conceptual choice though, aside from that. As it is, the UI form would probably take some external explaining for a user to understand, which is _probably_ alright for now.

willdonnelly:
@williamhbaker Yeah, that's pretty much how I expected it would look, and it's not great. But I think you're right that `[]string` is semantically what we want (and it's also a bit simpler in the code and avoids edge cases like "what if the user had a schema name containing a comma"), so maybe we just accept that for now and hope that array properties in the UI will be tweaked to look a bit cleaner at some point in the future?

willdonnelly:
That's good to know, thanks Travis.

I'm going to go ahead and merge this as-is then.

PR #963, opened 2023-09-15
source-postgres: Only apply schema filtering to catalog discovery

willdonnelly:
**Description:**

We use the same table-discovery logic inside of a capture in order to get various table metadata, and it's possible for a table to be a configured binding even if it's excluded from discovery. Likewise the watermarks table may be accessible and in use even if it's in a schema which has been filtered out, and that shouldn't be an error.

I've verified that this doesn't break the discovery filtering logic or normal captures, however I haven't actually gone to the effort of reproducing a capture which fails because of the old way it worked. I'm fairly confident this will fix it nonetheless.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/963)


williamhbaker (approved):
LGTM

PR #965, opened 2023-09-19
materialize-elasticsearch: don't retry errors during Validate

williamhbaker:
**Description:**

We have a pretty long retry duration setup for allowing extended retries of retryable errors for data-related operations, but this does not work well when performing connectivity checks.

A prime example is if the server is using a self-signed certificate, which we currently do not support. This will return a 5xx error that is currently retried, and will result in a timeout cancellation and mask the real error.

We could theoretically try to discriminate errors even further to decide which ones to retry, but for now it seems generally safer to fail fast when doing validation connectivity checks and allow the user to be the retry mechanism.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/965)

### DIFF
--- a/materialize-elasticsearch/driver.go
+++ b/materialize-elasticsearch/driver.go
@@ -211,7 +211,7 @@ func (c config) Validate() error {
 
 // toClient initializes a client for connecting to Elasticsearch, starting the network tunnel if
 // configured.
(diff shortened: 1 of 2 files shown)


willdonnelly (approved):
LGTM

PR #966, opened 2023-09-20
source-mongodb: _meta/op

mdibaiee:
**Description:**

- Add `_meta/op` to source-mongodb documents
- The tests helped find another issue, when emitting deleted documents we were not stringifying the `_id` the same way we do for other events, so there was a discrepancy there, fixed it.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/966)

### DIFF
--- a/source-mongodb/discovery.go
+++ b/source-mongodb/discovery.go
@@ -24,10 +24,12 @@ const idProperty = "_id"
 const (
 	metaProperty    = "_meta"
 	deletedProperty = "deleted"
+	opProperty      = "op"
 )
 
 type documentMetadata struct {
-	Deleted bool `json:"deleted,omitempty" jsonschema:"title=Delete Flag,description=True if the document has been deleted, unset otherwise."`
+	Deleted bool `json:"deleted,omitempty" jsonschema:"title=Delete Flag (deprecated, use op instead),description=True if the document has been deleted, unset otherwise."`
+	Op bool `json:"op,omitempty" jsonschema:"title=Change Operation,description=Change operation type: 'c' Create/Insert, 'u' Update, 'd' Delete."`
 }
 
 func generateMinimalSchema() json.RawMessage {
--- a/source-mongodb/discovery.go
+++ b/source-mongodb/discovery.go
@@ -23,12 +23,10 @@ const idProperty = "_id"
 
 const (
 	metaProperty    = "_meta"
-	deletedProperty = "deleted"
 	opProperty      = "op"
 )
 
 type documentMetadata struct {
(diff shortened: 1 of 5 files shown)


source-mongodb/discovery.go
```diff
@@ -24,10 +24,12 @@ const idProperty = "_id"
 const (
 	metaProperty    = "_meta"
 	deletedProperty = "deleted"
+	opProperty      = "op"
 )
 
 type documentMetadata struct {
-	Deleted bool `json:"deleted,omitempty" jsonschema:"title=Delete Flag,description=True if the document has been deleted, unset otherwise."`
+	Deleted bool `json:"deleted,omitempty" jsonschema:"title=Delete Flag (deprecated, use op instead),description=True if the document has been deleted, unset otherwise."`
+	Op bool `json:"op,omitempty" jsonschema:"title=Change Operation,description=Change operation type: 'c' Create/Insert, 'u' Update, 'd' Delete."`
```
williamhbaker, line 32: It would be good to include this `_meta/op` property in the minimal discovered schema, since it is controlled by the connector.

source-mongodb/discovery.go
```diff
@@ -24,10 +24,12 @@ const idProperty = "_id"
 const (
 	metaProperty    = "_meta"
 	deletedProperty = "deleted"
+	opProperty      = "op"
 )
 
 type documentMetadata struct {
-	Deleted bool `json:"deleted,omitempty" jsonschema:"title=Delete Flag,description=True if the document has been deleted, unset otherwise."`
+	Deleted bool `json:"deleted,omitempty" jsonschema:"title=Delete Flag (deprecated, use op instead),description=True if the document has been deleted, unset otherwise."`
```
williamhbaker, line 31: What would it take to completely get rid of this `deleted` property? I think the longer we wait to remove it, the harder it will be since more captures will have been created with it. But also going forward it's going to be confusing to have both `op: d` and `delete: true` in the documents - I would wonder what the difference is, which I could trust, etc. I'm assuming we would need to evaluate current production uses of the connector to see if a version bump and/or migration would be needed.

williamhbaker (approved):
LGTM

source-mongodb/discovery.go
mdibaiee: @williamhbaker this is the line that does that, the minimal schema of mongodb is generated from this struct's reflection

williamhbaker: Ah I didn't even notice that! Thanks.

PR #967, opened 2023-09-20
materialize-sql: compress outbound data for warehouse materializations

williamhbaker:
**Description:**

Adds compression of outbound data for `materialize-snowflake`, `materialize-redshift`, and `materialize-bigquery`. The largest part of the effort was re-working the file staging to split files into multiple parts if they weren't already doing that, to enable efficient parallel processing of the files being loaded. Previously redshift and bigquery could process a single huge file in parallel since they were uncompressed, but compression requires splitting the files up for that to keep happening.

GZIP compression is being used since it supported across the board for our current warehouse materializations. A fast compression level (`flate.BestSpeed`) is used since using a higher compression level doesn't add much benefit for space utilization, but does come at a much higher CPU cost: Testing on some synthetic data, I found that the `flate.BestCompression` level only reduced file sizes by ~5-10% and took at least 2x as long. There is a compression level faster even than `flate.BestSpeed` called `flate.HuffmanOnly`, which did require noticably less CPU, but produced file sizes that were about twice as large. So `flate.BestSpeed` seems like a pretty good compression level to use, although it is easy to change that in the future if we want to.

In terms of real-world performance, I tested the before & after throughput materializing several GB of data and didn't see any noticeable performance degradations on my local stack. In particular, the processing of the data in the warehouses was not adversely impacted, suggesting that the impact of handling compressed data is not a major performance hit, and the parallel processing of many compressed files is probably working about as well as parallel processing a single uncompressed file. This may not translate directly into production use cases, but it does not seem like there are any performance showstoppers from introducing this compression.

**Workflow steps:**

N/A

**Notes for reviewers:**

May be helpful to review commit-by-commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/967)


mdibaiee (approved):
LGTM! small nits

PR #969, opened 2023-09-20
source-s3: more useful error message on invalid credentials or missing bucket

williamhbaker:
**Description:**

Invalid credentials is a common error case, and currently this returns the "raw" AWS S3 client error that talks about a request signature mismatch. This won't make any sense to a user who tried to publish the capture, so we can handle that error (and some others) and produce a more understandable error message.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/969)


willdonnelly (approved):
LGTM

PR #971, opened 2023-09-20
source-postgres: Fix column discovery when lacking permissions

willdonnelly:
**Description:**

This commit replaces the old column-discovery query with a drop-in replacement which uses various `pg_catalog` tables instead of `information_schema.columns` and thereby avoids the `has_column_privilege()` check which is part of that view.

Previously it was possible for the capture user to have sufficient permissions to discover a table and its primary key, but the column-discovery query would return no information. This would result in a generated schema like:

```
{
  "schema": {
    "$defs": {
      "foobar": {
        "type": "object",
        "required": [ "id" ],
        "$anchor": "foobar",
        "properties": {}
      }
    },
    "allOf": [
      { ...the usual _meta portion of the schema... },
      { "$ref": "#foobar" }
    ]
  },
  "key": [ "/id" ]
}
```

And because that schema doesn't constrain the primary-key types in any way this would result in a very ugly `location /id accepts "array", "boolean", "null", "number", "object", "string" in schema file://flow.json#[...], but locations used as keys may only be null-able integers, strings, or booleans` error (for every table impacted).

This PR fixes that by using the underlying `pg_catalog` information instead, which avoids any "does the user have access to this column" checks and means that we can generate correct schemas even when we don't have access to the table itself. Which means that at the end of the day the user will get a much friendler `user "flow_capture" cannot read from table "public.foobar"` error at publication time.

**Workflow steps:**

In theory nothing should change for any preexisting captures or any future captures which have adequate permissions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/971)

### DIFF
--- a/source-postgres/.snapshots/TestDiscoveryWithoutPermissions
+++ b/source-postgres/.snapshots/TestDiscoveryWithoutPermissions
@@ -1,0 +1,106 @@
+Binding 0:
+{
+    "recommended_name": "discoverywithoutpermissions_117535",
+    "resource_config_json": {
+      "namespace": "public",
+      "stream": "discoverywithoutpermissions_117535"
+    },
+    "document_schema_json": {
+      "$defs": {
+        "PublicDiscoverywithoutpermissions_117535": {
+          "type": "object",
+          "required": [
+            "id"
+          ],
+          "$anchor": "PublicDiscoverywithoutpermissions_117535",
+          "properties": {}
+        }
+      },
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM. The question of omitting tables from discovery that the user does not have access vs. producing them but erroring is interesting. I do agree that producing them and erroring is probably best.

PR #972, opened 2023-09-21
materialize-bigquery: better error message on prereq check for dataset or project not found

williamhbaker:
**Description:**

Return a clear error message if the configured dataset or project ID can't be found.

Under certain circumstances BigQuery will return a useless error message like "Request could not be served" when the project ID or dataset is malformed in a certain way, but the 404 response code always means that one of the two could not be found.

There is a bit of code for `if googleErr.Message != ""` which is sort of peculiar and I think a leftover from prior refactoring and debugging which is being removed. We'll just pass through the error as-is if its not a 404, since other cases are pretty well described as far as I can tell.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/972)

### DIFF
--- a/materialize-bigquery/bigquery.go
+++ b/materialize-bigquery/bigquery.go
@@ -207,30 +207,25 @@ func prereqs(ctx context.Context, ep *sql.Endpoint) *sql.PrereqErr {
 
 	var googleErr *googleapi.Error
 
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

PR #974, opened 2023-09-21
source-postgres: Use `[lo, hi)` TID-range for keyless backfills

willdonnelly:
**Description:**

Previously our keyless backfill queries worked by going

    SELECT * FROM table ORDER BY ctid LIMIT n OFFSET m;

This worked in a theoretical sense, but on large tables it forced the database to sort the entire dataset on every query because the query planner doesn't recognize that `ORDER BY ctid` can be done more efficiently than that.

After this change our keyless backfill queries will look like:

    SELECT ctid, * FROM table WHERE ctid >= $1 AND ctid < $2;

This should be considerably more efficient because the Postgres query planner has a special case for this (see `tidpath.c`).

The current logic is fairly naive. It keeps track of the last CTID of the previous backfilled row as the backfill cursor, and queries the range (PrevBlock, PrevOffset+1) up to (PrevBlock+QuerySize+1, 0) where `QuerySize = ChunkSize / 100`. Note that this is a conservative estimate and means that keyless backfills will normally read fewer than `ChunkSize` rows per query.

There is one potential issue that I can see, though it seems unlikely to arise in practice. The issue is that currently we determine that a backfill has reached the end of a table when the backfill query yields an empty result set. But in principle this could happen here if there are sufficiently large gaps in the CTIDs of live rows. With the default chunk size of 50k rows this can only happen if 500 consecutive block indices are unused.

I suspect it may be possible to construct a pathological table which would exhibit this issue, but that it is unlikely to happen in real datasets.

**Workflow steps:**

Nothing changes as far as the user is concerned, except maybe their keyless backfills run faster.

### DIFF
--- a/source-postgres/backfill.go
+++ b/source-postgres/backfill.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/estuary/connectors/sqlcapture"
+	"github.com/jackc/pgtype"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


jgraettinger:
Closing because we used a ctid cursor with LIMIT instead.

PR #975, opened 2023-09-21
materialize-mysql: fix escaping issue with DSN

williamhbaker:
**Description:**

The MySQL driver didn't work with passwords containing special characters because of the URL escaping coming from `uri.String()`. Notably, this was causing Oracle HeatWave materializations to fail with an authentication error because these passwords require special characters. Using the `mysql.Config` to construct the DSN fixes that.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/975)

### DIFF
--- a/materialize-mysql/.snapshots/TestConfigURI-IncorrectSSL
+++ b/materialize-mysql/.snapshots/TestConfigURI-IncorrectSSL
@@ -1,2 +1,2 @@
-will:secret1234@tcp(example.com:3306)/somedb?tls=
+will:secret1234@tcp(example.com:3306)/somedb
 invalid 'sslmode' configuration: unknown setting "whoops-this-isnt-right"
--- a/materialize-mysql/driver.go
+++ b/materialize-mysql/driver.go
@@ -13,7 +13,6 @@ import (
 	"fmt"
 	"io"
 	"net"
-	"net/url"
 	"slices"
 	"strings"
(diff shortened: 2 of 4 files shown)


materialize-mysql/.snapshots/TestConfigURI-IncorrectSSL
```diff
@@ -1,2 +1,2 @@
-will:secret1234@tcp(example.com:3306)/somedb?tls=
+will:secret1234@tcp(example.com:3306)/somedb
```
psFried, line 1: Is the change here significant? I'm assuming that `tls=` is equivalent to not having that parameter at all, but just wanted to double check.

psFried (approved):
LGTM % one question

materialize-mysql/.snapshots/TestConfigURI-IncorrectSSL
williamhbaker: This output can only be reached if an invalid SSL mode is selected, which will cause a validation error, so we'd never actually try to connect with this DSN. On a purely academic level, I'd suspect that omitting the parameter entirely is more along the lines of what MySQL wants, since that is the output from the mysql library `FormatDSN()` method when the TLS config is set to an empty string.

PR #976, opened 2023-09-21
source-postgres: Use `WHERE ctid > $1` in keyless backfills

willdonnelly:
**Description:**

Previously our keyless backfill queries worked by going

    SELECT * FROM table ORDER BY ctid LIMIT n OFFSET m;

This worked in a theoretical sense, but on large tables it forced the database to sort the entire dataset on every query because the query planner doesn't recognize that `ORDER BY ctid` can be done more efficiently than that.

After this change our keyless backfill queries will look like:

    SELECT ctid, * FROM table WHERE ctid > $1 LIMIT n;

This should be considerably more efficient because the Postgres query planner has a special case for this (see tidpath.c):

    > EXPLAIN SELECT ctid, * FROM \"test\".\"foobar\" WHERE ctid > $1 LIMIT 100;
    QUERY PLAN=Limit  (cost=0.00..1.97 rows=100 width=42)
    QUERY PLAN=  ->  Tid Range Scan on foobar  (cost=0.00..32.51 rows=1651 width=42)
    QUERY PLAN=        TID Cond: (ctid > '(0,0)'::tid)

We are implicitly assuming that the results of a TID range scan will be in ascending CTID order even though there is no explicit `ORDER BY ctid` clause in the query (because adding that causes the query planner to add an explicit sort of the whole dataset).

To make sure this doesn't cause any problems, there is a new sanity check in the keyless backfill result processing which will fail the capture with an error if result rows (within a single chunk) are not in ascending CTID order.

**Workflow steps:**

No visible impact other than keyless backfills going much faster.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/976)

### DIFF
--- a/source-postgres/backfill.go
+++ b/source-postgres/backfill.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/estuary/connectors/sqlcapture"
+	"github.com/jackc/pgtype"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #978, opened 2023-09-22
source-mongodb: order on $natural instead of ts

mdibaiee:
**Description:**

there are no indexes defined on the oplog.rs collection, and the recommendation is to use the $natural order (the order by which records are written to disk) in order to read from this collection. This works for oplog.rs since there are only insertions, and never updates, so the $natural order is the order in which records have been inserted. There are performance implications for using any sort other than $natural since all documents must be scanned for those sorts.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/978)

### DIFF
--- a/source-mongodb/oplog.go
+++ b/source-mongodb/oplog.go
@@ -15,7 +15,7 @@ type oplogRecord struct {
 	Ts primitive.Timestamp `bson:"ts"`
 }
 
-const tsProperty = "ts"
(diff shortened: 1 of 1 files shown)


mdibaiee:
I'm going to merge to unblock users, please do a review after 👍🏽

PR #979, opened 2023-09-22
source-mongodb: emit converted state, handle mixed state 

mdibaiee:
**Description:**

- Previously if we had to convert the state of a connector with old checkpoint format, we would only emit the updates to the checkpoint as backfills were progressing, so if midway through backfilling of collections, we had a restart, we would end up with a state that is partly old format, and partly new format
- This pull-request handles two issues: it allows us to parse mixed formats between old and new formats, and it also emits a checkpoint update as soon as the conversion is done to avoid this issue of mixed up formats
- Additionally, set a `batchSize` of `1024` for backfill query
- Ordering on _id seems like the most likely cause of the performance regression.
We can use natural ordering, and still do incremental backfills. Since we start reading oplog changes since the beginning of the backfill, it means that even if we miss some updates to documents during the backfill, once we start listening on change streams, we will apply those updates after the backfill. Natural order is close to insertion order but is not guaranteed to be insertion order. New insertions since the beginning of the backfill will be covered by the change stream so we should be good for covering those events after the initial backfill is done
- There is a bug in _meta/op where we do not emit it during backfill, we should emit _meta/op: "c" during backfill. The schema of the _meta/op is also wrong at the moment, needs to be of type string

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/979)


source-mongodb/pull.go
```diff
@@ -320,7 +341,7 @@ func (c *capture) BackfillCollection(ctx context.Context, client *mongo.Client,
 		state.BackfillStartedAt = time.Now()
 	}
 
-	var opts = options.Find().SetSort(bson.D{{idProperty, SortAscending}})
+	var opts = options.Find().SetSort(bson.D{{idProperty, SortAscending}}).SetBatchSize(BACKFILL_BATCH_SIZE)
```
williamhbaker, line 344: We'll still return the results for the entire collection here right and not just `BACKFILL_BATCH_SIZE` docs? From a skim of the MongoDB docs I'm ~95% sure if that's how it works but wanted to ask.

More importantly: Do you have a good idea of what this does to connector memory? I'm wondering if it gets all `BACKFILL_BATCH_SIZE` docs at once and has to store them all in memory, or if the client is smarter than that. It could be a problem if the documents are very large to get that many at once.

williamhbaker (approved):
LGTM % a couple of questions

source-mongodb/pull.go
```diff
@@ -137,7 +157,7 @@ type capture struct {
 }
 
 type deprecatedCaptureState struct {
-	Resources map[string]bson.Raw `json:"resources"`
+	Resources map[string]json.RawMessage `json:"resources"`
```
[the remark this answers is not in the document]
mdibaiee: @williamhbaker `json.RawMessage` and `bson.Raw` are both `[]byte` under the hood. In case of the initial `Unmarshal`, `json.RawMessage` basically does not touch the underlying bytes and you get the original raw byte array that was found under the key.

Afterwards we just cast that `json.RawMessage` to `bson.Raw`, but the underlying data stays the same byte array

source-mongodb/pull.go
mdibaiee: > We'll still return the results for the entire collection here right and not just BACKFILL_BATCH_SIZE docs? From a skim of the MongoDB docs I'm ~95% sure if that's how it works but wanted to ask.

Yes I've verified that we do still capture all of the documents, the batch size is explained [here](https://www.mongodb.com/docs/manual/reference/command/find/) and [here](https://www.mongodb.com/docs/manual/reference/method/cursor.batchSize/). The first one explains what happens with `find` command itself, and the second one explains what happens when you set that option on the `cursor` that you get from the command. The driver [does both](https://github.com/mongodb/mongo-go-driver/blob/v1.11.0/mongo/collection.go#L1235-L1237).

The behaviour you mentioned happens if you set `singleBatch` to true, which we don't, and by default that is `false`.

source-mongodb/discovery.go
```diff
@@ -27,7 +27,7 @@ const (
 )
 
 type documentMetadata struct {
-	Op bool `json:"op,omitempty" jsonschema:"title=Change Operation,description=Change operation type: 'c' Create/Insert, 'u' Update, 'd' Delete."`
+	Op string `json:"op,omitempty" jsonschema:"title=Change Operation,description=Change operation type: 'c' Create/Insert, 'u' Update, 'd' Delete."`
```
williamhbaker, line 30: I don't think the description is rendering correctly in output collections; it gets cut off after the `,` so it just ends up being `"Change operation type: 'c' Create/Insert"`. Relatedly, it would be good to have a test that snapshots the discovered schema so we can more clearly it see it up front and prevent regressions.

source-mongodb/.snapshots/TestDiscover
```diff
@@ -0,0 +1,83 @@
...
+      "properties": {
+        "_id": {
+          "type": "string"
+        },
+        "_meta": {
+          "$schema": "http://json-schema.org/draft/2020-12/schema",
+          "$id": "https://github.com/estuary/connectors/source-mongodb/document-metadata",
+          "properties": {
+            "op": {
+              "type": "string",
```
williamhbaker, line 22: Ah seeing this snapshot actually made me think of something else: For other captures with `_meta/op`, it's not actually a `string`, its an `enum` of the possible values, ex: https://github.com/estuary/connectors/blob/654a9467aa5d1790ce66e6e055296b1c28dd8c63/source-postgres/.snapshots/TestDiscoveryCapitalization#L49-L54

mdibaiee: @williamhbaker added enum to it 👍🏽

willdonnelly (approved):
LGTM

source-mongodb/pull.go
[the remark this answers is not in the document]
psFried: > Afterwards we just cast that json.RawMessage to bson.Raw, but the underlying data stays the same byte array

@mdibaiee I'm a little late to the party here, but this does not sound safe at all. `json.RawMessage` is expected to be JSON-encoded data, whereas `bson.Raw` is expected to be BSON-encoded. The two are not compatible, so I don't see how this would work without parsing the JSON and converting it to BSON. I'd think a type cast would just result in an invalid `bson.Raw` value.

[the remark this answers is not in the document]
willdonnelly: Oh, that's a good catch. It seems likely that `bson.Raw` being an alias for `[]byte` is serialized into JSON as a base64-encoded string, and the `json.RawMessage` would still be that string (including double-quotes around it), so there still needs to be a step that strips those quotes and decodes base64 into bytes before casting to BSON.

PR #980, opened 2023-09-25
materialize-snowflake: allow per-resource schema configuration

williamhbaker:
**Description:**

Snowflake sessions & transactions can span across different schemas. This change allows for resource-level schema configuration, similar to other SQL materializations like materialize-postgres.

Particular attention is given to making the change compatible with existing materializations: If no schema is set for a resource OR the resource schema is the same as the endpoint configuration schema (as will be the case for all existing tasks), the resource path is still just the name of the target table.

Manual testing was done to confirm that tables with alternate schemas than the endpoint configuration get updated & materialized to correctly and can co-exist with tables in other alternate schemas and/or the endpoint configuration schema. Also the "migration" scenario was tested, where a pre-existing materialization was re-published with this new image, with confirmation that it did not re-materialize anything (the old checkpoint still held), both initially and if the schema was explicitly configured to be the same as the endpoint schema, including permutations of capitalizations that Snowflake disregards.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/980)

### DIFF
--- a/materialize-snowflake/.snapshots/TestSQLGeneration
+++ b/materialize-snowflake/.snapshots/TestSQLGeneration
@@ -1,5 +1,5 @@
---- Begin target_table createTargetTable ---
-  CREATE TABLE IF NOT EXISTS target_table (
+--- Begin "a-schema".target_table createTargetTable ---
+  CREATE TABLE IF NOT EXISTS "a-schema".target_table (
     key1 INTEGER NOT NULL,
     key2 BOOLEAN NOT NULL,
     boolean BOOLEAN,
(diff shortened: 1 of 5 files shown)


jgraettinger (approved):
LGTM

> Manual testing was done to confirm that tables with alternate schemas than the endpoint configuration get updated & materialized to correctly and can co-exist with tables in other alternate schemas and/or the endpoint configuration schema. Also the "migration" scenario was tested, where a pre-existing materialization was re-published with this new image, with confirmation that it did not re-materialize anything (the old checkpoint still held), both initially and if the schema was explicitly configured to be the same as the endpoint schema, including permutations of capitalizations that Snowflake disregards.

:heart:

PR #981, opened 2023-09-26
source-postgres: Set `publish_via_partition_root` option

willdonnelly:
**Description:**

This fixes https://github.com/estuary/connectors/issues/751 (in combination with a Flow docs PR which I'll be merging around the same time)

This commit sets the `publish_via_partition_root` flag everywhere that `source-postgres` executes or documents a `CREATE PUBLICATION` query:

 - When, during prerequisite validation, the connector autocreates a publication it will now also attempt to set the flag. It does this in a followup `ALTER PUBLICATION` query because versions of Postgres <13 don't have this option and we don't want the create query to fail in those cases.
 - The example queries in the README now show it being set.
 - The `init-user-db.sh` script which initializes the CI database for tests will now set it as documented in the README.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/981)

### DIFF
--- a/source-postgres/README.md
+++ b/source-postgres/README.md
@@ -83,6 +83,7 @@ CREATE USER flow_capture WITH PASSWORD 'secret' REPLICATION;
 GRANT pg_read_all_data TO flow_capture;
 
 CREATE PUBLICATION flow_publication FOR ALL TABLES;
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

PR #985, opened 2023-09-27
source-mongodb: run a single change stream

mdibaiee:
**Description:**

**Workflow steps:**

**Notes for reviewers:**


source-mongodb/pull.go
```diff
@@ -3,24 +3,26 @@ package main
...
+	"golang.org/x/sync/errgroup"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
+// Limit the number of backfills running concurrently to avoid
+// too much memory pressure
+const ConcurrentBackfillLimit = 20
```
williamhbaker, line 24: If a capture has more than 20 bindings, I believe this will mean that bindings 20+ will need to wait until at least some of those initial 20 have finished.

Here's an idea: I see that there is a method `cursor.RemainingBatchLength()`, which IIUC describes how many documents have been loaded into memory for the iterator, and once that gets to `0` the next call will require a network request. With this information we may be able to efficiently process an entire batch's documents for a binding, and then potentially switch to a different binding and process a batch for it. A single threaded version of this would loop, and in each loop pick a binding that still needs a backfill done at random and process 1 batch for it. A multi-threaded version could start a separate go routine for each binding, but use a semaphore to limit the active batches to something smaller like 5 at a time, while still progressing the backfill for each binding somewhat evenly, at random.

This may work poorly though if there is a lot of overhead in requesting a new cursor vs. continuing a previous cursor. But if that's the case, I would wonder if doing 20 at a time is too many.

source-mongodb/pull.go
```diff
@@ -316,15 +316,19 @@ func (c *capture) ChangeStream(ctx context.Context, client *mongo.Client, bindin
...
 
 	return nil
 }
 
-const BackfillBatchSize = 1024
+// We use a very large batch size here, however the go mongo driver will
+// automatically use a smaller batch size that prevents memory usage to blow out
+// of proportion. The large number here basically is a way for us to ask the
+// driver to choose a reasonable batch size for us.
+const BackfillBatchSize = 262144
```
williamhbaker, line 331: We are calculating when to checkpoint off of having emitted this many documents, and a single checkpoint should not have this many documents. We're using 50k for SQL captures as their default backfill chunk size so that might be a good choice here too, see https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1692036679355679

source-mongodb/pull.go
```diff
@@ -183,83 +148,118 @@ type resourceState struct {
...
+		"$and", bson.A{
+			bson.D{{"$or",
+				bson.A{
+					bson.D{{"operationType", "delete"}},
+					bson.D{{"operationType", "insert"}},
+					bson.D{{"operationType", "update"}},
+					bson.D{{"operationType", "replace"}},
+				},
+			}},
+		  bson.D{{"$or", collectionFilters}},
```
williamhbaker, line 221: It would be really nice to get _all_ events from all collections of all types. That way we have the most up to date resume cursor possible - I'm thinking about if there was a source where we were capturing only a single collection of many, and that collection had a low rate of changes compared to others and the resume token we got from that single collection ends up getting pushed out of the oplog. This would then require filtering out oplog events we don't care about in the connector. I believe SQL captures do something like this. Would that be practical here?

source-mongodb/pull.go
```diff
@@ -416,7 +431,7 @@ func (c *capture) BackfillCollection(ctx context.Context, client *mongo.Client,
 			if diff, err := oplogTimeDifference(ctx, client); err != nil {
 				return fmt.Errorf("could not read oplog, access to oplog is necessary: %w", err)
 			} else {
-				// TODO: This is just spamming logs, but users won't see these logs. Should we
+				// TODO: Users won't see these logs until there is an error. Should we
```
williamhbaker, line 434: TBH I think we should elevate this check to just being a check that is done in `Validate`, and only errors if its a new enough version of MongoDB to query the retention period directly (I think that's a thing?) and that retention period is lower than some threshold. Then there could be a "skip retention sanity check" option like `source-mysql`. It could be up for debate how this would work on versions too old for that retention period to be queried directly - either always fail unless the "skip" option is checked, or always pass with a warning.

Mainly my thoughts here are motivated by seeing strange things in some of my own testing: Logs like "your retention period is 35 seconds" when I have loaded some test data, etc. It just doesn't seem practical to check for a sufficient oplog period in this way, altough I certainly understand there is precedent for it.

source-mongodb/pull.go
```diff
@@ -174,7 +133,13 @@ func (s *captureState) Validate() error {
...
 const (
 	StatusBackfill = 0
-	StatusStreaming = 1
+	// Value of "1" was previously used to designate `StatusStreaming`, however
+	// an update in how we process streams meant that we needed to backfill
+	// bindings again to end up with a clean state, so `StatusStreaming` was
+	// updated to "2", and bindings that still had `Status = 1` in their state
+	// were backfilled. All of this is to say: do not re-use the value 1 in this
+	// enum.
+	StatusStreaming = 2
```
williamhbaker, line 142: Having 2 possible statuses here for expressing "is the backfill done" seems a little smelly to me, especially since I don't think the `StatusBackfill` is being used anywhere. How about a boolean property like `BackfillComplete`? All states going forward would have this initialized as `false` until it is set to `true` by the updated connector, so this would also accomplish the desire to re-backfill all collections.

source-mongodb/pull.go
```diff
@@ -183,83 +148,118 @@ type resourceState struct {
...
-		// connector will run a backfill.
-		//
-		// There is potential for optimisation here by only resetting the checkpoint
-		// of the collection which can't resume instead of resetting the whole
-		// connector's checkpoint and backfilling everything
+		// this one.
 		if e, ok := err.(mongo.ServerError); ok {
 			if e.HasErrorMessage(resumePointGoneErrorMessage) || e.HasErrorMessage(resumeTokenNotFoundErrorMessage) {
-				return fmt.Errorf("change stream on collection %s cannot resume capture, this is usually due to a small oplog storage available. Please resize your oplog to be able to safely capture data from your database: https://go.estuary.dev/NurkrE. After resizing your uplog, you can remove the binding for this collection add it back to trigger a backfill: %w", res.Collection, e)
+				return fmt.Errorf("change stream cannot resume capture, this is usually due to insufficient oplog storage. Please consider resizing your oplog to be able to safely capture data from your database: https://go.estuary.dev/NurkrE. After resizing your oplog, you can remove all bindings for this capture and add them back to trigger a backfill: %w", e)
```
williamhbaker, line 246: Are we resetting the resume token anywhere if all bindings are removed?

source-mongodb/pull.go
mdibaiee: I see your point, I personally wante to avoid touching data that we didn't need to touch, hence the filtering, but I can see that it has implications for our connector being able to work in certain scenarios, that should work 👍🏽

source-mongodb/pull.go
```diff
@@ -316,15 +316,19 @@ func (c *capture) ChangeStream(ctx context.Context, client *mongo.Client, bindin
...
+		return fmt.Errorf("cursor error for change stream: %w", err)
 	}
 
 	defer cursor.Close(ctx)
 
 	return nil
 }
 
-const BackfillBatchSize = 1024
+// We use a very large batch size here, however the go mongo driver will
```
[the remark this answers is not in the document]
mdibaiee: I think so, in my tests, collections with smaller documents tend to have larger (effective) batch sizes, and collections with larger documents tend to have smaller batch sizes.

Specifically, I ran tests on three collections with these results:
listingsAndReviews 752 effective batch size, 17.9kb average document size
tweets 10042 effective batch size, 1.9kb  average document size
companies 3653 effective batch size, 4kb average document size

With these numbers it seems the effective byte size of a batch is around 15-20MB.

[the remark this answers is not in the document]
mdibaiee: Okay, I actually have a precise answer now: MongoDB cannot return results larger than the maximum BSON document size, which is 16mb, so the batch size will always be capped at 16mb. [docs](https://www.mongodb.com/docs/v7.0/reference/limits/#mongodb-limit-BSON-Document-Size)

source-mongodb/pull.go
mdibaiee: I updated this, so now we first check the `oplogMinRetentionHours` configuration, if it is set, we require that it be more than 24 hours. If it is not set, we use the _approximation_ method to see if the oplog _seems_ small, we do let the user know though that this is an approximation.

source-mongodb/pull.go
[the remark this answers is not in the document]
williamhbaker: Nice, that seems pretty reasonable.

mdibaiee:
@williamhbaker a change stream requires more than access to the oplog itself, because the events for the change stream require both access to read the oplog and access to read the documents that have been changed.

With the change to do a change stream on the whole instance, we require access to read all databases in the instance. Filtering for the specific collections we need does not help with this unfortunately.
See [this page of docs](https://www.mongodb.com/docs/manual/changeStreams/#access-control)

> To open a change stream on an entire deployment, applications must have privileges that grant [changeStream](https://www.mongodb.com/docs/manual/reference/privilege-actions/#mongodb-authaction-changeStream) and [find](https://www.mongodb.com/docs/manual/reference/privilege-actions/#mongodb-authaction-find) actions on all non-system collections for all databases in the deployment.

source-mongodb/pull.go
```diff
@@ -220,6 +220,12 @@ func (c *capture) ChangeStream(ctx context.Context, client *mongo.Client, resour
 
 		var startAtWithSafety = streamStartAt.Add(oplogSafetyBuffer)
 		if err := oplogHasTimestamp(ctx, client, startAtWithSafety); err != nil {
+			cp, err := json.Marshal(captureState{})
```
williamhbaker, line 224: I believe this is where we would end up if the backfill took so long that the required position in the oplog is no longer available. In this case I think we'd want to not reset the checkpoint, and just error out, since re-doing the backfill will likely end up just repeating and failing again.

mdibaiee: @williamhbaker I think the issue is we can't be sure that's exactly what is happening here, other cases could be happening if we are unable to find the timestamp in the oplog, for example: the backfill was done, we started a change stream on a single database (not the instance), but there were no events for a while (an idle database). This means we don't have a resume token yet.

The connector is then restarted and we try to start the change stream from the timestamp again, but this time we can't find the timestamp in the oplog anymore.

source-mongodb/pull.go
mdibaiee: @williamhbaker this actually is a very neat way of handling this. I implemented it and tested it.

I'm currently using 20 as the limit for bindings running concurrently. Given that a batch can't be larger than 16mb, it means at maximum this process will use around 320mb memory (not counting for other memory usage of the driver).

williamhbaker: Nice! 20 does _seem_ like a lot, although I certainly haven't tested it. Generally I haven't seen much benefit to more than 3-5 concurrent requests like these. 320mb of raw data + overhead is kind of a lot too so it might make sense to dial that back. Hard to testing meaningfully though without production use cases though.

source-mongodb/pull.go
williamhbaker: Hmm I see. This seems fine for now since we haven't seen anything close to not being able to backfill in ~24 hrs, but might be good to keep an eye on going forward.

source-mongodb/pull.go
```diff
@@ -132,141 +109,213 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
...
-	} else if !state.BackfillStartedAt.IsZero() {
+	} else if !streamStartAt.IsZero() {
 		var oplogSafetyBuffer, _ = time.ParseDuration("-5m")
-		if err := oplogHasTimestamp(ctx, client, state.BackfillStartedAt.Add(oplogSafetyBuffer)); err != nil {
+
+		var startAtWithSafety = streamStartAt.Add(oplogSafetyBuffer)
+		if err := oplogHasTimestamp(ctx, client, startAtWithSafety); err != nil {
+			cp, err := json.Marshal(captureState{})
+			if err != nil {
+				return fmt.Errorf("marshalling checkpoint for removed bindings: %w", err)
```
williamhbaker, line 228: We should update these error messages

source-mongodb/pull.go
```diff
@@ -132,141 +109,213 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
...
+				return fmt.Errorf("change stream cannot resume capture, this is usually due to insufficient oplog storage. Please consider resizing your oplog to be able to safely capture data from your database: https://go.estuary.dev/NurkrE. The connector will now attempt to backfill all bindings again: %w", e)
+			} else if e.HasErrorMessage(unauthorizedMessage) {
+				log.WithField("err", e).Warn("could not start instance-level change stream, trying database-level change stream.")
+				// Check that there is only one database, if so, we can start a change
+				// stream on this database. If there is more than one database
+				// configured, we error out as we do not support multiple change streams
+				// on different databases. An instance-level change stream must be used
+				// for that scenario
+				var dbName = resources[0].Database
+				for _, res := range resources[1:] {
```
williamhbaker, line 265: Does this work if there is only a single resource?

source-mongodb/pull.go
```diff
@@ -132,141 +109,213 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
...
+					if res.Database != dbName {
+						return fmt.Errorf("We are unable to establish an instance-level change stream on your database due to an error. Please consider granting permission to read all databases to the user: %w", err)
+					}
+				}
+
+				var db = client.Database(dbName)
+				// TODO: what happens if a user switches from readAnyDatabase to
```
williamhbaker, line 272: This does seem like an important consideration. I don't know how resume tokens work for a "global" change stream vs a per-database one. Would switching permissions from read@<db> to readAnyDatabase break the capture? Presumably _increasing_ the connector's permissions shouldn't break things, but if the resume tokens aren't compatible then it would, which wouldn't be good. If that does break things, I think ideally we would store the type of permission being used in the state checkpoint, only emitting it with the very first checkpoint, and then always continue to try to use that mechanism.

source-mongodb/pull.go
```diff
@@ -412,25 +482,25 @@ func (c *capture) BackfillCollection(ctx context.Context, client *mongo.Client,
...
-
 	if err := cursor.Err(); err != nil {
 		return fmt.Errorf("cursor error for backfill %s: %w", res.Collection, err)
 	}
 
-	state.Status = StatusStreaming
+	state.Backfill.Done = true
 	var checkpoint = captureState{
 		Resources: map[string]resourceState{
 			resourceId(res): *state,
```
williamhbaker, line 506: Nit: This function could take a `resourceState` directly (not a reference), to avoid needing to do this dereference.

source-mongodb/pull.go
```diff
@@ -412,25 +482,25 @@ func (c *capture) BackfillCollection(ctx context.Context, client *mongo.Client,
 			if err = c.Output.Checkpoint(checkpointJson, true); err != nil {
```
williamhbaker, line 482: With backfills running concurrently, these calls to `Documents` and `Checkpoint` will be interleaved and one binding could potentially commit another one's documents etc. It's wouldn't be the end of the world since the result would be a possible duplication of documents, but still probably good to avoid. I think what we'll need to do is group up all of the returned documents from a single mongo network-batch send send them atomically with the corresponding checkpoint update.

source-mongodb/pull.go
```diff
@@ -132,141 +109,213 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
...
+		"state":             state,
+		"streamResumeToken": state.StreamResumeToken,
+		"streamStartAt":     streamStartAt,
+	}).Info("listening on changes")
+
+	var eventFilter = bson.D{{"$match", bson.D{{
+		"$and", bson.A{
+			bson.D{{"$or",
+				bson.A{
+					bson.D{{"operationType", "delete"}},
```
williamhbaker, line 209: At the risk of being pedantic...are there _other_ events we should be listening for? Should we filter out any events? It seems like it would be best to take everything the database will give us and do all of the filtering in the connector, extracting a resume token even from non CRUD events.

williamhbaker (approved):
LGTM % some final comments

source-mongodb/pull.go
mdibaiee: Yes it works: https://go.dev/play/p/fsPSvjMP8YT

source-mongodb/pull.go
mdibaiee: Had a chat with Johnny on this as well, yes I think using the state to keep track of the mode of change stream is important: we will still break, but we will break understanding why and can help the user. The idea is to keep track of the mode, and if there is a change in the mode, we automatically do a backfill and switch to the new mode.

mdibaiee:
Scenarios to test:
- [ ] Resuming a connector that has been running on the current v1 version with this version of the connector
- [ ] Stopping a backfilling connector of this version midway through and continuing, by the end we should have all of the data
- [ ] Instance-wide change stream
- [ ] Database-level change stream
- [ ] Switching from instance-wide change stream to database-level change stream should trigger a backfill and set the scope
- [ ] If the resume token is lost in oplog compactions, we must reset the checkpoint and restart the connector to start a backfill again
- [ ] If a backfill has finished, but we have had no resume token, if the timestamp of the beginning of backfill - 5 minutes is not in the oplog anymore, we must reset the checkpoint and restart the connector to start a backfill again

source-mongodb/pull.go
mdibaiee: Actually I did a test of this scenario and turns out the resume token is not invalid across these two change streams. In this case I think it is okay to continue using the resume token of the other change stream, since that means we already have covered all the events relevant to the actively captured databases / collections. If a new database / collection is added, we will inevitably backfill it, so it should also be okay in that case.

PR #987, opened 2023-09-28
source-postgres-batch: Only allocate `cursorValues` if there's data

willdonnelly:
**Description:**

Allocating outside the loop means that the first polling iteration on an empty table will transform a cursor `nil` into `[]any{nil}` (of appropriate length) which is counted as "not the first query" the next time we try polling.

We'd prefer to instead keep running the "start" query until the first time it produces data, so that allocation is moved inside the loop where it will only be executed the first time it's actually needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/987)

### DIFF
--- a/source-postgres-batch/driver.go
+++ b/source-postgres-batch/driver.go
@@ -528,9 +528,6 @@ func (c *capture) poll(ctx context.Context, bindingIndex int, tmpl *template.Tem
 		columnPointers[i] = &columnValues[i]
 	}
 
-	if len(cursorValues) != len(cursorNames) {
-		cursorValues = make([]any, len(cursorNames))
-	}
 	var count int
 	for rows.Next() {
 		if err := rows.Scan(columnPointers...); err != nil {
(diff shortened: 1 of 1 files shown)


source-postgres-batch/driver.go
```diff
@@ -559,6 +556,9 @@ func (c *capture) poll(ctx context.Context, bindingIndex int, tmpl *template.Tem
 			return nil, fmt.Errorf("error emitting checkpoint: %w", err)
 		}
 
+		if len(cursorValues) != len(cursorNames) {
```
jgraettinger, line 559: nit: perhaps a comment that this addresses the "empty result set" case?

jgraettinger (approved):
LGTM

source-postgres-batch/driver.go
willdonnelly: Done

PR #989, opened 2023-09-29
source-dynamodb: parallelize discovery

williamhbaker:
**Description:**

The DescribeStreams API has an account-wide limit 10 requests per second, which makes it
difficult to efficiently discover large numbers of tables without hitting that limit, or timing out
the discover.

Because of this, don't run DescribeStream for every discovered table's stream - it's just not
practical.

Also run `discoverTable` operations in parallel, since the DescribeTable API has a limit of 2,500
requests per second. And increase the maximum number of retries the clients will use to allow for a
degree of rate limiting from system backpressure during normal operation.

These changes allow a source with 100's of tables to be successfully discovered, at the expense of
only finding issues with disabled streams when the connector runs rather than during the publication
steps. If streaming is enabled for a table and configured correctly, it's probably not going to be
explicitly disabled, so this seems like a fair trade-off.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/989)

### DIFF
--- a/source-dynamodb/discovery.go
+++ b/source-dynamodb/discovery.go
@@ -4,17 +4,17 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sync"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
-	"github.com/aws/aws-sdk-go-v2/service/dynamodbstreams"
-	streamTypes "github.com/aws/aws-sdk-go-v2/service/dynamodbstreams/types"
 	cerrors "github.com/estuary/connectors/go/connector-errors"
 	pc "github.com/estuary/flow/go/protocols/capture"
 	pf "github.com/estuary/flow/go/protocols/flow"
 	"github.com/invopop/jsonschema"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 )
 
 type discoveredTable struct {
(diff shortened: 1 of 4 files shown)


source-dynamodb/discovery.go
```diff
@@ -264,29 +284,11 @@ func discoverTable(ctx context.Context, c *client, table string) (discoveredTabl
...
-		log.WithFields(log.Fields{
-			"table":     tableDescribe.Table.TableName,
-			"streamArn": tableDescribe.Table.LatestStreamArn,
-			"status":    streamDescribe.StreamDescription.StreamStatus,
-		}).Warn("streaming enabled for table but stream was not active; table will not be captured")
-		return discoveredTable{}, false, nil
-	}
-
-	if streamDescribe.StreamDescription.StreamViewType != streamTypes.StreamViewTypeNewAndOldImages {
+	if tableDescribe.Table.StreamSpecification.StreamViewType != types.StreamViewTypeNewAndOldImages {
```
mdibaiee, line 287: Is it also possible that a table's `StreamSpecification` sets `StreamViewType` to something other than `StreamViewTypeNewAndOldImages`, but the last stream of the table is made with `StreamViewTypeNewAndOldImages`?

i.e.

```
streamDescribe.StreamViewType == StreamViewTypeNewAndOldImages
tableDescribe.StreamSpec.StreamViewType != StreamViewTypeNewAndOldImages
```

mdibaiee (approved):
LGTM % question above

source-dynamodb/discovery.go
williamhbaker: I don't believe so - per the [API docs](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TableDescription.html), the `StreamSpecification` for the `TableDescription` is for the _current_ stream on the table, and you can only have a single stream for a table. And by all accounts even the `Table.StreamSpecification.StreamEnabled` should be a direct replacement of `streamDescribe.StreamDescription.StreamStatus != streamTypes.StreamStatusEnabled`.

PR #990, opened 2023-09-29
source-mysql-batch: Tweaks to work nicer in the UI

willdonnelly:
**Description:**

Tweaks a the resource config schema metadata so it should look nicer in the UI, and hopefully with the cursor no longer marked as required the generated bindings from discovery won't fail the form validation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/990)


willdonnelly:
Merging unreviewed because I don't have a lot of time and it's a couple trivial changes to a newly added connector.

PR #996, opened 2023-10-06
source-mysql-batch: Remove default 'DBName' setting

willdonnelly:
**Description:**

The default setting for this property should be leaving it unset, not `"mysql"`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/996)


willdonnelly:
Merging without review because this is a trivial edit.

PR #997, opened 2023-10-06
source-mysql-batch: Make (Config).ToDSN() match client library

willdonnelly:
**Description:**

When originally writing this code I accidentally looked at the latest version of the client library DSN parsing logic, however this has been changed since the version (v1.5.0) we're currently using.

The old version we're using has the dumbest possible "parsing" logic and just splits the string on some separators without any encoding/decoding of special characters, and so we need to use an equally dumb string-splatting behavior when constructing the connection string.

If/when we upgrade the client library version this logic should be changed back to using `url.URL{}.String()` for symmetry with the newer parsing logic.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/997)

### DIFF
--- a/source-mysql-batch/main.go
+++ b/source-mysql-batch/main.go
@@ -5,7 +5,6 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
-	"net/url"
 	"strings"
 
 	schemagen "github.com/estuary/connectors/go/schema-gen"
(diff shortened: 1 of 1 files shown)


willdonnelly:
I'm going to merge this without review because there are no production captures using this connector yet other than a couple of test ones that Dave and I created and it will be obvious if this fixes things. Plus it's 5pm on a Friday so I don't want to impose on anybody just to get a rubber-stamp here.

PR #1000, opened 2023-10-10
source-mongodb: revert changes related to document decoding

mdibaiee:
**Description:**

- reverting the decoding changes from #985 

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1000)

### DIFF
--- a/source-mongodb/pull.go
+++ b/source-mongodb/pull.go
@@ -440,13 +440,9 @@ func (c *capture) BackfillCollection(ctx context.Context, sp *semaphore.Weighted
 		}
 		state.Backfill.LastId = rawDoc[idProperty]
 
-		var doc = make(bson.M)
-		for key, val := range rawDoc {
-			var v interface{}
-			if err = val.Unmarshal(&v); err != nil {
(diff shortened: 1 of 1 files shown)


williamhbaker:
Mind adding a description of how this was tested? From what I can gather, the previous encoding is suspected of producing incorrect results...nested arrays within an additional array structure I think? It is not immediately obvious to me that this change addresses that, but a test case would be pretty convincing.

jgraettinger (approved):
LGTM - this is removing an optimization where we attempt to avoid re-decoding the top level document?

mdibaiee:
@jgraettinger @williamhbaker  yes, reverting back an optimization to avoid re-decoding top level document: the issue is that `RawValue`'s structure of decoding an object / array are different from what an original `cursor.Decode` yields. It works the same for scalar values, but not for objects and arrays apparently. The reasoning from what I gather is that it wants to ensure that it can encode those value back to exactly what they were in BSON format, so in an attempt to not be _lossy_, the decoding is more verbose.

jgraettinger:
As a follow-up, then, please add integration test coverage that exercises these data types.

PR #1001, opened 2023-10-10
source-mysql-batch: Streaming queries and default polling interval

willdonnelly:
**Description:**

When writing the initial implementation of `source-mysql-batch` I mistakenly assumed that the `go-mysql-org/go-mysql` implementation of the `database/sql` driver interface probably used streaming queries.

(This was an easy mistake to make since the `database/sql` result rows abstraction implements an iterator protocol which is clearly intended to allow streaming processing, and the MySQL client library does have methods for streaming query execution. I didn't think about it too hard and forgot about the extra steps we had to take to make streaming result processing work for the MySQL CDC connector.)

This PR replaces the use of `*sql.DB` with the lower-level `*client.Conn` representation and then explicitly uses a streaming query execution method when polling a binding.

It also adds a mutex so that only one binding can be actively executing its query at a time, because doing it that way was simpler than adding connection pooling, and because it's useful operationally to only have one thing at a time going on in the logs, and because this keeps memory usage as low as possible for now.

While I was working on all that, I also went ahead and added a connector-level advanced config option for setting the default polling interval. It can still be overridden on a per-binding basis and defaults to 24 hours if not set anywhere, but now the user can adjust a single number and have that impact all bindings which haven't specified their own polling interval setting.

**Workflow steps:**

Queries with massive resultsets should now work instead of getting the connector OOM-killed.

**Notes for reviewers:**

This is probably not the most elegantly-factored code I could have written, but we needed this fixed ASAP and _functionally_ it should be correct. Any internal plumbing cleanups can be done later as needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1001)

### DIFF
--- a/source-mysql-batch/driver.go
+++ b/source-mysql-batch/driver.go
@@ -29,7 +29,7 @@ type BatchSQLDriver struct {
 	DocumentationURL string
 	ConfigSchema     json.RawMessage
 
-	Connect               func(ctx context.Context, configJSON json.RawMessage) (*client.Conn, error)
+	Connect               func(ctx context.Context, cfg *Config) (*client.Conn, error)
 	GenerateResource      func(resourceName, schemaName, tableName, tableType string) (*Resource, error)
 	ExcludedSystemSchemas []string
 }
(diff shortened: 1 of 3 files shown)


source-mysql-batch/driver.go
```diff
@@ -454,13 +457,18 @@ func (c *capture) worker(ctx context.Context, bindingIndex int, res *Resource) e
 		return fmt.Errorf("error parsing template: %w", err)
 	}
 
-	var pollInterval = 24 * time.Hour
+	// Polling interval can be configured per binding. If unset, falls back to the
+	// connector global polling interval. If that's also unset, falls back to 24h.
+	var pollStr = "24h"
```
jgraettinger, line 462: nit: this is not required / is duplicated given SetDefaults()

jgraettinger (approved):
LGTM % nit

source-mysql-batch/driver.go
willdonnelly: Ah, good catch. I'll clean that up in a followup so I can click the merge button that much sooner here.

PR #1003, opened 2023-10-11
source-postgres: don't allow empty schema in discovery schema selection list

williamhbaker:
**Description:**

If an empty schema is provided in the discovery schema selection list, there will be no tables discovered since all of them will be filtered out. This may inadvertently happen if a row is added to the array UI component and then not removed. It is not always obvious to the user in this case that they are actually providing a 0-length string as their schema filter.

Ideally we might have controls that would somehow combine a variable length array of items and the fact that if any of them are there, then they must not be 0 length. I couldn't find an easy way to do this with our use of schema-gen: Newer versions of `github.com/invopop/jsonschema` do allow some control over the format of `items` within an array, but not for setting `minLength` that I could see, and upgrading to one of these newer versions (we are using 0.5 which is quite old) may bring in various other breaking changes that don't seem worth right tackling now.

So the connector will return a validation error if any of the provided schemas are an empty string, with a description on what to do instead.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1003)

### DIFF
--- a/source-postgres/.snapshots/TestGeneric-SpecResponse
+++ b/source-postgres/.snapshots/TestGeneric-SpecResponse
@@ -75,7 +75,7 @@
             },
             "type": "array",
             "title": "Discovery Schema Selection",
-            "description": "If this is specified only tables in the selected schema(s) will be automatically discovered. Leave blank to discover tables from all schemas."
+            "description": "If this is specified only tables in the selected schema(s) will be automatically discovered. Omit all entries to discover tables from all schemas."
           }
         },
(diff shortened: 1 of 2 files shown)


willdonnelly (approved):
LGTM.

It feels really dumb that there's a straightforward modification we'd like to make to our config schema which the tools we're using prevent us from making, but until we're ready to invest the time and effort to specify a minimum length on the strings I guess this works.

PR #1004, opened 2023-10-11
source-mysql-batch: Remove redundant 24h default poll interval

willdonnelly:
**Description:**

Just a minor cleanup pointed out in https://github.com/estuary/connectors/pull/1001, since the `SetDefaults()` logic fills in a default polling interval of `"24h"` we can assume the config property is non-empty later on.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1004)


williamhbaker (approved):
LGTM

PR #1005, opened 2023-10-11
source-mysql-batch: Optimize document serialization

willdonnelly:
**Description:**

This adds a new helper package `./go/encrow` for serializing tabular data such as we get from SQL queries, and modifies `source-mysql-batch` to use this new package.

The helper package is more efficient than using the standard library `encoding/json` to serialize a `map[string]any` for two reasons:

1. The top level of document serialization is specialized to only work on objects with a specific arity and list of property names. This is represented with a new `encrow.Shape` abstraction.
2. Serializing the individual values of each document is done with the github.com/segmentio/encoding/json package, which is a micro-optimized JSON encoder/decoder package intended as a faster drop-in replacement for the stdlib one.

The specialization alone resulted in about a 3x speedup even while still using the stdlib encoder for each field value. Adding the segmentio JSON encoder provided another 2x speedup on top of that.

Benchmark results for demonstration (measuring single core performance on my laptop):

    $ GOMAXPROCS=1 go test -v ./go/encrow/ -bench=. -benchtime=30s -count=3
    cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
    BenchmarkStandardSerialization
    BenchmarkStandardSerialization      4982           7267985 ns/op
    BenchmarkStandardSerialization      4867           7644533 ns/op
    BenchmarkStandardSerialization      4099           8395409 ns/op
    BenchmarkShapeSerialization
    BenchmarkShapeSerialization        26900           1334689 ns/op
    BenchmarkShapeSerialization        26907           1337353 ns/op
    BenchmarkShapeSerialization        27273           1327145 ns/op

I have written a normal test and a fuzz test which attempt to verify that for every possible object mapping keys to values, the output of the new encoder exactly matches the standard library encoder. This turned up one bug (now fixed) in how an object with zero properties was serialized, and other than that I have seen no failures after a decently large amount of fuzzing.

So as far as I can tell this ought to be a pretty straightforward improvement to throughput. Currently just for `source-mysql-batch` but this same package can be used for all our SQL captures in the future.

**Workflow steps:**

Nothing changes, `source-mysql-batch` captures run up to 6x faster.

**Notes for reviewers:**

I'm not 100% settled on this design, but I was just playing around with approaches to optimizing our JSON serialization of result rows and this seems to work pretty well, as demonstrated in the benchmark results included up above.

It would be even simpler (literally a one-line code change) to just drop in the segmentio JSON encoder instead of the standard library one but skip all the "shape optimization" stuff, however the custom code I wrote for specializing the object shape really does provide a noticeable improvement (about 2x) even on top of that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1005)

### DIFF
--- a/go/encrow/encrow.go
+++ b/go/encrow/encrow.go
@@ -1,0 +1,78 @@
+// Package encrow provides a JSON encoder optimized for SQL table rows.
+package encrow
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/segmentio/encoding/json"
+)
+
+// A Shape represents the structure of a particular document.
+type Shape struct {
+	arity    int
+	prefixes []string
+	swizzle  []int
+}
+
+// NewShape constructs a new Shape corresponding to the provided field names.
+func NewShape(fields []string) *Shape {
+	// Construct a list of swizzle indices which order the fields by name.
+	var swizzle = make([]int, len(fields))
+	for i := 0; i < len(fields); i++ {
+		swizzle[i] = i
+	}
+	sort.Slice(swizzle, func(i, j int) bool {
+		return fields[swizzle[i]] < fields[swizzle[j]]
+	})
+
+	// Sort names
+	var sortedNames = make([]string, len(fields))
+	for i, j := range swizzle {
+		sortedNames[i] = fields[j]
+	}
+
+	return &Shape{
+		arity:    len(sortedNames),
+		prefixes: generatePrefixes(sortedNames),
+		swizzle:  swizzle,
+	}
+}
+
+func generatePrefixes(fields []string) []string {
+	var prefixes []string
+	for idx, fieldName := range fields {
+		var quotedFieldName, err = json.Marshal(fieldName)
+		if err != nil {
+			panic(fmt.Errorf("error escaping field name %q: %w", fieldName, err))
+		}
+		if idx == 0 {
+			prefixes = append(prefixes, fmt.Sprintf("{%s:", quotedFieldName))
+		} else {
+			prefixes = append(prefixes, fmt.Sprintf(",%s:", quotedFieldName))
+		}
+	}
+	return prefixes
+}
+
+// Encode serializes a list of values into the specified shape.
+func (s *Shape) Encode(values []any) ([]byte, error) {
+	var err error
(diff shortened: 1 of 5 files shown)


go/encrow/encrow.go
```diff
@@ -0,0 +1,78 @@
...
+			prefixes = append(prefixes, fmt.Sprintf("{%s:", quotedFieldName))
+		} else {
+			prefixes = append(prefixes, fmt.Sprintf(",%s:", quotedFieldName))
+		}
+	}
+	return prefixes
+}
+
+// Encode serializes a list of values into the specified shape.
+func (s *Shape) Encode(values []any) ([]byte, error) {
```
jgraettinger, line 59: suggestion: pass `bs` into this function and initialize as `bs = bs[:0]`.

This allows for buffer re-use across rows, or the caller can pre-allocate a new buffer with a desired capacity, or pass in `nil` to match the current behavior.

Downside is that it does enable the caller to make use-after-free errors where they can't today with the PR'd API. The garden path case is fine, because `Output.Documents` fully serializes its input before returning.

jgraettinger (approved):
LGTM - this looks great and timely, and appears more broadly applicable then just SQL captures.

go/encrow/encrow.go
willdonnelly: Done. In my benchmarks it doesn't appear to provide a noticeable speedup, but it certainly shouldn't hurt.

willdonnelly: Correction (after Johnny pointed out that I accidentally shadowed the buffer variable in the benchmark code), this actually does improve throughput a bit, something like a 40% speedup:

```
Before (No Buffer Reuse):
BenchmarkShapeSerialization        26218           1369746 ns/op
BenchmarkShapeSerialization        26358           1373856 ns/op
BenchmarkShapeSerialization        26253           1325392 ns/op

After (With Buffer Reuse):
BenchmarkShapeSerialization        37581            959920 ns/op
BenchmarkShapeSerialization        37747            956241 ns/op
BenchmarkShapeSerialization        37480            927488 ns/op
```

PR #1009, opened 2023-10-12
source-mysql-batch: Persist cursor values in state checkpoint

willdonnelly:
**Description:**

This PR implements support for persistent cursors in the `source-mysql-batch` connector. These cursors are stored in the state checkpoint as an object like `{"Streams":{"test_basic_capture_826935":{"CursorNames":["id"],"CursorValues":[123]}}}`.

Changes to the cursor-names list will result in a stream getting restarted from the empty cursor, the same as removing-and-recreating should.

Since these state checkpoint updates have to be serialized to JSON and we've seen repeatedly that JSON serialization can be the limiting factor on capture throughput, I've minimized this extra work in the laziest possible way by making the capture only emit a checkpoint update after every thousand rows (plus one when the query is finished, if needed), so the checkpoints should only account for about 0.1% more serialization work than the result rows themselves require.

Introducing persistent state means that the connector's restart behavior is no longer easily tested via ad hoc uses of `flowctl raw capture`, so I've also written one automated test which sanity-checks that the cursor restart behavior is working as it should in the simplest case of a single-column cursor which is an integer ID. As a side effect this also verifies that everything else required to capture from a simple table is basically working.

**Workflow steps:**

Cursor values will now persist across connector restarts rather than solely within a single connector invocation. No action is required by the user to make this happen if they have specified a cursor already.

**Notes for reviewers:**

I went back and forth on this but ultimately decided not to try and align with the nascent binding behavior spec at all just yet, on the theory that it's probably easier to migrate to the correct behavior in the future if the old behavior is clearly distinct.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1009)


williamhbaker (approved):
LGTM

PR #1010, opened 2023-10-12
materialize-sql: Use github.com/segmentio/encoding/json for encoding

willdonnelly:
**Description:**

The `stagedFile` abstraction used for BigQuery/Snowflake/Redshift materializations uses a `*sql.CountingEncoder` which in turn uses a `json.Encoder` to serialize row contents.

The standard library `encoding/json` package is fairly slow, and the `github.com/segmentio/encoding/json` package is a drop-in replacement which performs 2-3x better in most benchmarks, so swapping out the package used in `materialize-sql/encoder.go` should theoretically give us a noticeable throughput improvement for basically no actual effort.

**Notes for reviewers:**

This change is completely untested, but should be quite safe provided that it builds and passes CI in the first place. I don't have any hard numbers on how much of a bottleneck JSON serialization represents for these materializations, but it's probably not trivial and we have some long-running ones where it should be pretty blatantly obvious if this helps.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1010)

### DIFF
--- a/materialize-sql/encoder.go
+++ b/materialize-sql/encoder.go
@@ -3,9 +3,10 @@ package sql
 import (
 	"compress/flate"
 	"compress/gzip"
-	"encoding/json"
 	"fmt"
 	"io"
+
+	"github.com/segmentio/encoding/json"
 )
 
 const (


jgraettinger (approved):
LGTM we tested this w/ bigquery using the PR image and 👍

PR #1015, opened 2023-10-13
materialize-bigquery: set user-agent header

williamhbaker:
**Description:**

Sets the user-agent option for the `materialize-bigquery`client, and plumbs through the information from the build in order to populate that user-agent.

The user-agent will look like this, and is (as best I can tell) compatible with the requirements for the Google Partner program: `"Estuary/materialize-bigquery-v1-b4c0562 (GPN:Estuary Technologies, Inc.;)"`

The version tag and commit SHA are provided as docker build args. For simplicity they are provided to all connectors via our CI configuration, although `materialize-bigquery` is the only one that will use them for now.

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1015)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -172,7 +172,10 @@ jobs:
           context: .
           file: ${{ matrix.connector }}/Dockerfile
           load: true
-          build-args: BASE_IMAGE=ghcr.io/estuary/base-image:${{ steps.prep.outputs.tag }}
+          build-args: |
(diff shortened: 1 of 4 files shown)


mdibaiee (approved):

jgraettinger:
Is this still needed?

williamhbaker:
Probably not, I was keeping it around in case we wanted to use this instead of the simpler version in https://github.com/estuary/connectors/pull/1118. I'll close this PR but keep the branch around just in case.

PR #1016, opened 2023-10-13
Adding Google Cloud SQL variants for mysql, sqlserver and postgres

dyaffe:
**Description:**

Adding Google Cloud SQL variants for mysql, sqlserver and postgres

This is required for their new partner program.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1016)

### DIFF
--- a/materialize-sqlserver/VARIANTS
+++ b/materialize-sqlserver/VARIANTS
@@ -1,1 +1,2 @@
 materialize-google-cloud-sql-mysql
+materialize-azure-sqlserver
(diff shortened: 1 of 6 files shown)


materialize-sqlserver/VARIANTS
```diff
@@ -0,0 +1,2 @@
+materialize-google-cloud-sql-mysql
```
willdonnelly, line 1: This looks like a copy-paste error?

willdonnelly (changes requested):
LGTM except I think there's a mistake in the `materialize-sqlserver` variants list which needs to be fixed.

materialize-sqlserver/VARIANTS
dyaffe: Fixed that

willdonnelly (approved):
LGTM

PR #1023, opened 2023-10-19
source-postgres-batch: Feature parity with MySQL Batch

willdonnelly:
**Description:**

This PR implements several changes to `source-postgres-batch` which should bring it back to feature parity with `source-mysql-batch` after the recent tweaks to that connector. This includes:

- A connector-level advanced config option for the default polling interval to use when the resource binding declines to specify a value. If unset the default for `source-postgres-batch` will be 5 minutes, as it was before this option was added.
- Using the `encrow` helper package to optimize serialization throughput on large query resultsets.
- Persisting cursor values in the state checkpoints so that restarting the connector doesn't force large incremental table captures to restart from the beginning.
- An automated test which verifies that the state-checkpoint restart behavior works correctly in the common case, and as a side effect verifies that the whole capture process is basically working in at least that one scenario.

**Workflow steps:**

No user action is required for any of these changes, any preexisting captures should keep on working without any changes. The only new knob is the polling interval setting in the advanced configuration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1023)


willdonnelly:
One notable difference between `source-mysql-batch` and `source-postgres-batch` which this PR doesn't address is that the MySQL capture will only capture from one table at a time, while this one is happy to be running queries against multiple large tables simultaneously.

We should definitely resolve the difference so that the connectors behave similarly, but I'm not sure if the correct resolution is to add a mutex/semaphore to the Postgres connector or to go back and fix the MySQL connector to use a connection pool (plus maybe also a semaphore to bound the maximum amount of concurrency), so I've chosen to do nothing about that just yet.

jgraettinger (approved):
LGTM

PR #1025, opened 2023-10-20
materialize-sqlserver: use bulk insert for load phase

mdibaiee:
**Description:**

- Batch together documents from the same binding in one bulk insertion operation as long as they consecutive in Load phase, this should help in performance of the connector during load phase.
- Tested by running a materialization on the connector: can see the performance boost on a materialization I had set up previously, but yet to be seen on the customer's data
- Integration tests are run on this connector in CI pipeline
- Realised we don't have tests for compound keys, so I added some tests to verify this doesn't break compound key loading

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1025)


materialize-sqlserver/driver.go
```diff
@@ -459,21 +459,64 @@ func (d *transactor) Load(it *pm.LoadIterator, loaded func(int, json.RawMessage)
...
+
+			var err error
+			batch, err = txn.PrepareContext(ctx, mssqldb.CopyIn(b.tempLoadTableName, mssqldb.BulkOptions{}, colNames...))
+			if err != nil {
+				return fmt.Errorf("load: preparing bulk insert statement on %q: %w", b.tempLoadTableName, err)
+			}
+		}
+
+		if it.Binding != lastBinding {
+			if _, err := batch.ExecContext(ctx); err != nil {
```
jgraettinger, line 490: nit: if you conditioned this on `batch != nil` then you could eliminate the repetition above.

Or, something like:


```go
	for it.Next() {
		var b = d.bindings[it.Binding]

		if lastBinding != it.Binding {
		        // Submit previous `batch`.
			if batch != nil {
				if _, err := batch.ExecContext(ctx); err != nil {
					return fmt.Errorf("load: batch insert on %q: %w", b.target.Identifier, err)
				}
			}
			// Prepare next `batch` for the updated binding.
			var colNames = []string{}
			for _, col := range b.target.Keys {
				// Column names passed here must not be quoted, so we use Field instead of Identifier
				colNames = append(colNames, col.Field)
			}

			var err error
			batch, err = txn.PrepareContext(ctx, mssqldb.CopyIn(b.tempLoadTableName, mssqldb.BulkOptions{}, colNames...))
			if err != nil {
				return fmt.Errorf("load: preparing bulk insert statement on %q: %w", b.tempLoadTableName, err)
			}
		}

		if converted, err := b.target.ConvertKey(it.Key); err != nil {
			return fmt.Errorf("converting Load key: %w", err)
		} else if _, err := batch.ExecContext(ctx, converted...); err != nil {
			return fmt.Errorf("load: writing data to batch on %q: %w", b.tempLoadTableName, err)
		}
	}

	// You already have this logic to handle the final batch at EOF:
	if batch != nil {
		if _, err := batch.ExecContext(ctx); err != nil {
			return fmt.Errorf("load batch insert on %q: %w", d.bindings[lastBinding].target.Identifier, err)
		}
	}
```

jgraettinger (approved):
LGTM

A comment below to improve factoring is all

jgraettinger:
> Tested by running a materialization on the connector: can see the performance boost on a materialization I had set up previously, but yet to be seen on the customer's data
Integration tests are run on this connector in CI pipeline
Realised we don't have tests for compound keys, so I added some tests to verify this doesn't break compound key loading

:heart:

jgraettinger:
No changes required, just noting it: handling for the Store iterator below could use the same factoring -- it's not actually required that you track different batches per-binding, since you'll see all stores of a given binding before any stores of the next binding.

PR #1030, opened 2023-10-24
Fleshing out python connectors

jshearer:
**Description:**

* `pull_upstream.sh` imports upstream connector from optional subtree, recording commit/location for license reasons
* `Dockerfile.example` is an example Python connector dockerfile with all the fixings
* `source-airtable` is an example connector that I imported and tested running, but didn't actually modify yet
* `README.md` describes the python connector development process as I understand it

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1030)


source-airtable/test.flow.yaml
```diff
@@ -0,0 +1,43 @@
+---
+captures:
+  acmeCo/source-airtable:
+    shards:
+      logLevel: debug
+    endpoint:
+      connector:
+        image: ghcr.io/estuary/source-airtable:local
+        config:
+          pokemon_name: pikachu
```
jgraettinger, line 10: lol well this isn't gonna work

source-airtable/Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM python:3.11-slim as builder
```
jgraettinger, line 1: I think we need to find a way to have this be a template execution, collapsed with `Dockerfile.example` from this commit, with an easy garden path so that devs can largely ignore Docker unless they need custom packages, etc.

Copy / paste of Dockerfiles will get unwieldy pretty much immediately (it's already unwieldy for our technical connectors in this repo), and while it's not in scope, they _also_ could benefit from establishing a better precedent here.

We should probably pick and try to standardize on a tool like https://github.com/hairyhenderson/gomplate for executing a custom template given a JSON input over the CLI (this is literally just the first Google search match and looked reasonable, but I have no strong opinion on the particular template renderer we use).

source-airtable/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+ENV PYTHONUNBUFFERED=1 \
```
jgraettinger, line 3: nit: please comment on what these are and why we need them.

source-airtable/__main__.py
```diff
@@ -0,0 +1,7 @@
+from python import shim_airbyte_cdk
+from source_airtable import SourceAirtable
+
+shim_airbyte_cdk.CaptureShim(
+    delegate=SourceAirtable(),
+    oauth2=None
```
jgraettinger, line 6: nit: there's a value for this spec already in the DB that we can pluck out and place here

python/README.md
```diff
@@ -0,0 +1,187 @@
...
+This module contains the building blocks to build new Flow connectors in Python, as well as adapting external Python connectors written in standard protocols to Flow. A Python connector is a subclass of the `Connector` classes in the `capture`, `derive`, and `materialize` (once it's written) submodules. Convenience classes `shim_airbyte_cdk.CaptureShim` and `shim_singer_sdk.CaptureShim` are provided to adapt existing Airbyte or Singer/Meltano connectors.
+
+A Python connector is composed of a few critical pieces:
+* `pyproject.toml` defines the Python module and its dependencies. For example:
+  ```toml
+  [tool.poetry.dependencies]
+  # If we're pulling in an open-source connector that doesn't need any changes, we can define it here. Subdirectory can be omitted if the connector lives at the root of the repo.
+  source_bigdata = { git = "https://github.com/big/data.git", subdirectory = "connector" }
+  # If we're working with an open-source connector that we have pulled in-tree, we can define it like this, where the connector was pulled into the subdirectory "upstream-connector"
+  source_bigdata = { path = "upstream-connector" }
```
jgraettinger, line 11: I think I'm 👎 on having the additional indirection of `upstream-connector`.

The extra level of nesting is awkward, and if we're making changes to it's implementation and behavior then it's no longer "upstream", in the sense that it's a unmodified & vendor'd chunk of code. And we pretty much _always_ intend to make _some_ modification to the upstream connector (for example, by simplifying its usually-incorrect schema and enabling schema inference).

Can we just have it be the current directory, as `source-pokemon` does ?

python/README.md
```diff
@@ -0,0 +1,187 @@
...
+  ```
+* In addition to being useful documentation for other people running your connector, `test.flow.yaml` or some other valid Flow spec is needed to invoke your connector locally. You'll point `flowctl` at this file to test your connector when developing locally.
```
jgraettinger, line 13: The vision is that various `test.flow.yaml` files will become fully-automated acceptance tests for each connector, which is required for every connector and driven by our CI.

We would have actual working credentials (sops protected) which grab an actual set of data from the live API using the connector, that's then verified against an expectation.

More plumbing to do here, of course.

python/README.md
```diff
@@ -0,0 +1,187 @@
...
+  ```python
+  from python import shim_airbyte_cdk
+  from source_bigdata import SourceBigData
+
+  shim_airbyte_cdk.CaptureShim(
+      delegate=SourceBigData(),
+      oauth2=None
+  ).main()
+  ``` 
+* A `Dockerfile` to build your connector into a container image that will be run in production. An example dockerfile can be found in this directory. Make sure to `source-changeme` with the name of your connector. If you are writing a materialization connector, make sure to also change the `LABEL FLOW_RUNTIME_PROTOCOL` to `materialization`.
```
jgraettinger, line 24: as noted, I think we need a path to templating these, so that the garden path is very lightweight and can take advantage of changes to the base template.

python/README.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Pulling an open-source connector in tree
+While using existing builds of open-source connectors is all well and good, there will come a time in every connector's life that we need to make changes to it. Before deciding to pull a connector in-tree, first make sure that the version of the connector you're going to pull is licensed to allow this usage. Then, the connector can be imported like so
+
+```bash
+$ git remote add -f --no-tags bigdata https://github.com/big/data.git 
+$ ./python/pull_upstream.sh bigdata/master path/in/source/repo path/in/this/repo
```
jgraettinger, line 31: nit: maybe make these examples more explicit, like:

./python/pull_upstream.sh bigdata/master upstream/path/to/source-foobar ./source-foobar

Also, is there a way to pass in a specific revision?

python/README.md
```diff
@@ -0,0 +1,187 @@
...
+    "required": [
+      "stream",
+      "syncMode"
+    ]
+  },
+  "documentationUrl": "https://docs.bigdata.com/integrations/sources/bigdata"
+}
+```
+
+### Debugging
```
jgraettinger, line 126: :heart: this is awesome

python/pull_upstream.sh
```diff
@@ -0,0 +1,77 @@
...
+        git merge -s ours --no-commit --allow-unrelated-histories "$SOURCE_COMMIT"
+    fi &&
+
+    git diff --color=never "$OLD_TREEISH" "$SOURCE_COMMIT:$SOURCE_PREFIX" \
+        | git apply -3 --directory="$DEST_PREFIX" || git mergetool
+
+    if (( $? == 1 )); then
+        echo "Uh-oh! Try cleaning up with |git reset --merge|."
+    else
+        git commit -em "import: Pull in changes from external repo
```
jgraettinger, line 54: nit: make the first line more specific ? Ex:

`import: $DEST_PREFIX through $SOURCE_COMMIT` or something ?

otherwise they'll all look the same in `git log`.

python/pull_upstream.sh
```diff
@@ -0,0 +1,77 @@
...
+Remote Repo URL: $SOURCE_REMOTE_URL
+Source Commit name: $SOURCE_COMMIT
+Source Commit ID: $SOURCE_SHA1
+Source Repo Prefix: $SOURCE_PREFIX/
+Import Path: $DEST_PREFIX/
+
+# Feel free to edit the title and body above, but make sure to keep the
+# ${FUNCNAME[0]}: line below intact, so ${FUNCNAME[0]} can find it
+# again when grepping git log.
+${FUNCNAME[0]}: $SOURCE_SHA1 $SOURCE_PREFIX $DEST_PREFIX"
```
jgraettinger, line 64: neat!

python/README.md
jgraettinger, line 31: Oh, one more thought here (and I'm making this up as I go, but...) -- I'm thinking that the user should be required to provide the license of the imported code and bake it into the commit message?

In that way, we're tracking the individual licenses of each import commit.

Perhaps we take in both short form (MIT, APACHE-2, etc), where we verify in `pull_upstream.sh` that it's a known license type, and also in long form as a perma-link to a URL the accompanying upstream license.

This does two things: 
* It facilitates making license review an explicit part of the code review process.
* It lets us write tools that scan the commit history and generate all applicable licenses for imported code and URL proofs of the license.

source-airtable/Dockerfile
jgraettinger: As discussed over Slack, another option is to use relative symbolic links -- tracked by git -- paired with `--build-arg`'s that pass in (minimally) the connector name.

`--build-args` can also read from the environment so there may be an opportunity to selectively use .env files to override ARGs of a common Dockerfile ?

jshearer: It turns out that you can just specify which Dockerfile to use with `docker build -f <path/to/Dockerfile>`, so I updated the Dockerfile template to use build args, and `build-local.sh` to provide them.

python/README.md
jshearer: How strongly do you feel about this? I ask because Python really wants its modules to be in subfolders. I tried changing to pulling the upstream connector directly into the `source-airtable` folder and that didn’t work because it’s a separate module that’s not configured like Poetry expects (i.e it has `setup.py` and requirements.txt instead of `pyproject.toml`). 

We _could_ tell people that when they import a connector, they also have to rejigger its python packaging to be consistent with all the rest, but given that it “just works” as a separate module imported by Poetry through `pyproject.toml`, I’m inclined to leave it be. Thoughts?

python/README.md
jshearer: Yup agreed. See updated `python/Dockerfile` and `build-local.sh`

python/README.md
jgraettinger: That's fair. I hadn't considered the impacts of setup.py et al.

In that case, though, a more conventional layout would be `source-airtable/source_airtable`, right? Given that's the actual name of the module, and is also how it's organized in the upstream repo.

I was mostly reacting to the "upstream-connector" naming.

python/README.md
jshearer: Done

source-airtable/__main__.py
jshearer: Done

python/pull_upstream.sh
```diff
@@ -18,14 +18,30 @@ git-merge-subpath() {
...
         SOURCE_COMMIT_FULL="${SOURCE_COMMIT_FULL%/*}"
-        echo "$SOURCE_COMMIT_FULL"
         # Get the URL
         SOURCE_REMOTE_URL=$(git remote get-url "$SOURCE_COMMIT_FULL")
     fi
 
     local SOURCE_SHA1
     SOURCE_SHA1=$(git rev-parse --verify "$SOURCE_COMMIT^{commit}") || return 1
 
+    LICENSE_FOUND=$(\
```
jgraettinger, line 28: :heart: this is excellent

python/activate.sh
```diff
@@ -3,16 +3,10 @@
...
-rm ${ROOTDIR}/pyproject.toml || true
-rm ${ROOTDIR}/poetry.lock || true
-
-# Link activated connector.
-ln ${ROOTDIR}/${CONNECTOR}/pyproject.toml ${ROOTDIR}/pyproject.toml
-ln ${ROOTDIR}/${CONNECTOR}/poetry.lock ${ROOTDIR}/poetry.lock
-
-cd ${ROOTDIR} && poetry install
\ No newline at end of file
+poetry -C ${ROOTDIR}/${CONNECTOR} install
```
jgraettinger, line 11: way better.

supernit: spell out `--directory` param instead of `-C` ?

jgraettinger (approved):
Giving an LGTM now because I think this is good to land.

One note I had is that `common` is ... pretty common. We will very likely eventually post a `flow-sdk` to pypi and perhaps should set up for that naming structure soon-ish, while deferring actually extracting out the package. But I'm cool to land as-is and consider that later.

source-airtable/test.flow.yaml
jshearer: Replaced with a legit stub

python/README.md
jshearer: There is now!

jshearer:
Renamed `common` to `flow-sdk` 👍

PR #1031, opened 2023-10-24
source-postgres: Disable parallel workers for keyless backfills

willdonnelly:
**Description:**

Sets `max_parallel_workers_per_gather = 0` before running any backfill queries and resets it to the default afterwards. This avoids an edge case in which the query planner decides to use parallel sequential scans to satisfy the `WHERE ctid > $1` query, which breaks our ordering assumptions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1031)


jgraettinger (approved):
LGTM

PR #1032, opened 2023-10-25
source-mysql-batch: Persist `LastPolled` timestamp

willdonnelly:
**Description:**

Adds a `LastPolled` property to the per-stream state checkpoints and uses that to implement polling after consistent intervals even across task restarts.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1032)

### DIFF
--- a/source-mysql-batch/driver.go
+++ b/source-mysql-batch/driver.go
@@ -475,6 +475,7 @@ type captureState struct {
 type streamState struct {
 	CursorNames  []string
 	CursorValues []any
+	LastPolled   time.Time
 }
 
 func (s *captureState) Validate() error {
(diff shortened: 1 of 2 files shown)


source-mysql-batch/driver.go
```diff
@@ -595,11 +574,39 @@ func (c *capture) poll(ctx context.Context, bindingIndex int, tmpl *template.Tem
...
+	}
+	pollInterval, err := time.ParseDuration(pollStr)
+	if err != nil {
+		return fmt.Errorf("invalid poll interval %q: %w", res.PollInterval, err)
+	}
+
+	// Sleep until it's been more than <pollInterval> since the last iteration,
+	// then update the "Last Polled" timestamp.
+	if !state.LastPolled.IsZero() && time.Since(state.LastPolled) < pollInterval {
+		select {
```
jgraettinger, line 591: nit: add a log of when we will next awake to poll the table?

jgraettinger (approved):
LGTM

source-mysql-batch/driver.go
willdonnelly: Done

PR #1033, opened 2023-10-25
source-mysql: Ignore query events beginning with `--` comments

willdonnelly:
**Description:**

This is absolutely the wrong way to handle comments in general, but it's at least consistent with how we're mishandling other comment styles at the moment and will unblock a current production capture that's hitting this issue.

In the longer term we should either:

1. Implement proper comment-stripping logic prior to the whole "should we ignore this query" regexp.
2. Convert query parse failures into a warn-and-ignore rather than the current fatal error, and possibly remove the whole "ignore some benign queries via regexp" thing entirely.

The rationale for (2) here is that being strict about parse failures has never, to my knowledge, actually caught a real issue -- the handful of DDL queries which we actually want to produce task failures are also the ones which reliably parse correctly, so this extra strictness is just a source of occasional issues in production.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1033)


jgraettinger (approved):
LGTM

PR #1036, opened 2023-10-26
source-mysql-batch: Less collision prone numbered placeholders

willdonnelly:
**Description:**

The previous implementation of numbered placeholders used the syntax `:N` which proved to be a dumb design decision since that appears as a substring of every timestamp.

(Note that this isn't *quite* as dumb as it sounds because the garden path for most users has them using the default query template which has no such issue, the only time this comes up is if you customize the query template and hard-code a specific timestamp as part of one of the queries. Unfortunately that's not a super uncommon thing to want to do, as specifying an initial cursor value for the first query is the obvious way to avoid backfilling a bunch of historical data)

The new placeholder syntax added in this PR looks like `@flow_cursor_value[N]` which ought to be substantially less collision prone. It might still be a good idea to make this substitution logic actually tokenize the input query (at least enough to avoid doing replacements within strings) but this should be solid enough in practice, I suspect.

After this PR both placeholder syntaxes are supported, so that there can be a convenient time window to change over existing uses to the new syntax. Once nothing is using the old syntax it can be removed by deleting the relevant regexp clause and chunk of expansion code.

While I was editing this logic, I also went ahead and added proper error returns and a check for out-of-range parameter indices, so if things fail in the future we should get a proper error message rather than a Go panic.

**Workflow steps:**

No user action is required. Future captures will just work with `@flow_cursor_value[N]` syntax. Preexisting captures will need to be updated to use that syntax in the future, and I intend to do that after this is live.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1036)

### DIFF
--- a/source-mysql-batch/.snapshots/TestQueryTemplate-SubsequentOneCursor
+++ b/source-mysql-batch/.snapshots/TestQueryTemplate-SubsequentOneCursor
@@ -1,1 +1,1 @@
-SELECT * FROM `test`.`foobar` WHERE (`ka` > :0) ORDER BY `ka`;
+SELECT * FROM `test`.`foobar` WHERE (`ka` > @flow_cursor_value[0]) ORDER BY `ka`;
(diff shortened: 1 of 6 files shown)


jgraettinger (approved):
LGTM

PR #1040, opened 2023-10-30
Variants of DB connecters for Amazon RDS

jonwihl:
Amazon RDS Source Connectors:

- MySQL
- PostgreSQL
- SQL Server
- MariaDB

Amazon RDS Materialization Connectors:

- MySQL
- PostgreSQL
- SQL Server
- MariaDB

Docs for all sources/materializations have been created

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1040)

### DIFF
--- a/materialize-mysql/VARIANTS
+++ b/materialize-mysql/VARIANTS
@@ -2,3 +2,4 @@ materialize-amazon-aurora-mysql
 materialize-mariadb
 materialize-mysql-heatwave
 materialize-google-cloud-sql-mysql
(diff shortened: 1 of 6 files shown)


dyaffe (changes requested):
I don't think these are case sensitive, but it may be good to switch source Postgres and SQL Server to be all lower case.  Otherwise LGTM.

dyaffe (approved):
LGTM

PR #1042, opened 2023-11-01
materialize-motherduck: new connector

williamhbaker:
**Description:**

New connector to materialize data to [Motherduck](https://motherduck.com/).

A couple of interesting things to point out:
1) The connector only works in "delta updates" mode. This is because currently doing large-scale updates on tables in Motherduck results in out of memory errors from Motherduck itself. So practically speaking there is not yet a reliable way to do a standard updates materialization. The connector will require that it be configured with delta updates, but still have the checkbox present (enabled by default) to allow us to transition into a possible standard updates version in the future.
2) Motherduck doesn't support transactions yet, but also does not error if transaction statements are used. The connector's `Store` code is written to use a transaction, even though it doesn't work transactionally, with the hopes that Motherduck will eventually honor transactions.

Performance and resource utilization isn't _great_, but I am hoping this improves with continued development of DuckDB/Motherduck and the Go DuckDB driver. In particular, there is a high amount of memory overhead from the driver opening a connection to Motherduck, and somewhat weirdly execution of statements that read data only from cloud storage sometimes seem to push work back onto the client application (connector), for reasons and purposes I don't fully understand. The implementation here does appear to work well enough and is both simple and obvious, so I hope it benefits transparently from ecosystem improvements. There may be shorter-term reasons to investigate other performance related workarounds (and I have some ideas) but I'm holding off on that for now, until a specific need arises.

**Workflow steps:**

Docs will be created in parallel with merging this PR.

**Notes for reviewers:**

The first commit is a pretty large diff and is mostly a mechanical refactoring of test-related code, so the subsequent commits are probably the most interesting.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1042)


materialize-motherduck/staged_file.go
```diff
@@ -0,0 +1,213 @@
+package main
```
williamhbaker, line 1: This is pretty much a copy/paste from redshift, with a few changes - namely redshift uses a manifest file, whereas Duckdb uses a list of file names.

materialize-motherduck/driver.go
```diff
@@ -0,0 +1,474 @@
...
+		if err := tplUpdateFence.Execute(&fenceUpdate, d.fence); err != nil {
+			return nil, pf.FinishedOperation(fmt.Errorf("evaluating fence template: %w", err))
+		}
+
+		return nil, pf.RunAsyncOperation(func() error {
+			// NB: Motherduck doesn't actually support transactions yet, but this code is written
+			// like it does. Eventually Motherduck will probably support transactions, and the
+			// connector pretending like it already does doesn't hurt anything and may make the
+			// eventual transition easier.
+			txn, err := d.storeConn.BeginTx(ctx, nil)
```
mdibaiee, line 428: Given that motherduck does not support transaction, it means this implementation cannot guarantee exactly-once semantics, right?

Since if `storeQuery` is executed, but `fenceUpdate` isn't, then the transaction might get repeated, and we end up with duplicate data being inserted.

I think given the lack of transactionality, we may want to look into the other pattern for materializations, "Recovery Log is Authoritative with Idempotent Apply":

https://github.com/estuary/flow/blob/f8d02d6211e8208ccbe81331497ff575175f307f/go/protocols/materialize/transactor.go#L58-L77

Although this will require the ability to run the copy from staged file to target table idempotently, which I'm not sure if DuckDB supports. If it doesn't, then the current implementation is the best we can do I guess

mdibaiee (approved):
LGTM % one comment

materialize-motherduck/driver.go
williamhbaker: Yep it will be "at least once".

There's not a direct way to do an idempotent copy/load of files unfortunately. There _may_ be a way to approximate an idempotent load by including the filename of the loaded file(s) as a materialized column (DuckDB makes this available via the `read_json` etc. functions) and running the copy queries in a way that excludes filenames that have already been loaded.

Balancing the effort and potential trickiness (what if the source collection already has a field of `filename`, etc.) of the currently possible `filename`-based idempotent loads vs. the fact that _DuckDB_ does support multi-statement transactions (it's just a MotherDuck limitation) and [MotherDuck lists transactions as a current DuckDB feature that is not yet supported & increased DuckDB feature coverage is continuously being worked on](https://motherduck.com/docs/architecture-and-capabilities/#considerations-and-limitations), I am inclined to wait & hope for transaction support in MotherDuck to resolve this.

PR #1044, opened 2023-11-01
source-mysql: Pin MySQL docker image to 8.0.35

willdonnelly:
**Description:**

The latest MySQL docker image release is consistently causing CI builds to fail on the same prerequisite check because of an incorrect binlog retention setting.

This should be investigated and fixed, but for now this unblocks `source-mysql` pushes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1044)


willdonnelly:
This appears to fix the CI build issues we're seeing and should probably be merged to make pushes reliable. We can figure out why `mysql:latest` stopped working later.

williamhbaker (approved):
LGTM

I think we may need a `go mod tidy` to fix some broken builds (bigquery, ~sheets,~ and snowflake), not related to this change though.

willdonnelly:
I went ahead and ran `go mod tidy` and added that to this PR out of convenience. It looks like it's probably fixed the Snowflake build issue but I'll let CI be the ultimate judge of that.

PR #1050, opened 2023-11-02
python: Update airbyte and singer `CaptureShim`s to support schema inference.

jshearer:
**Description:**

Since connector specs are usually wrong, we want a quick and easy way to configure a connector to just use our schema inference. This means emitting a bare schema containing just the key fields (non-null, since keys can't be null), and `x-infer-schema: true`. 

_With `usesSchemaInference=True`_:
```yaml
collections:
  acmeCo/custom_fields:
    schema:
      type: object
      properties:
        gid:
          type: string
      x-infer-schema: true
    key:
      - /gid
```
_With `usesSchemaInference=False`_:
```yaml
collections:
  acmeCo/custom_fields:
    schema:
      type:
        - "null"
        - object
      properties:
        gid:
          type:
            - "null"
            - string
        resource_type:
          type:
            - "null"
            - string
        name:
          type:
            - "null"
            - string
        created_at:
          type:
            - "null"
            - string
          format: date-time
        project:
          type:
            - "null"
            - object
          properties:
            gid:
              type:
                - "null"
                - string
            resource_type:
              type:
                - "null"
                - string
            name:
              type:
                - "null"
                - string
    key:
      - /gid
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1050)


jgraettinger (approved):
LGTM

A thing I'm legit unsure of is whether we need to programatically strip stuff out of the connector's JSON schema, or whether we want to do that as our recourse when a connector's schema proves to be wrong.

singer_sdk is actually quite proactive about verifying that emitted records match its schema. I believe we're far less likely to encounter problems with these connector schemas.

For airbyte, a bit motivation for this work is getting these connectors version controlled and having tooling where it's really easy for us to drive a connector locally and have a development feedback loop where we can respond to its encountered failures quickly and easily.

Given all that, I'm on the fence about whether we want a "nuke from orbit" approach, or to pass through the connector's schema with x-infer-schema: true (which effectively means the read-schema is the intersection of the emitted schema and read schema).

I _think_ I lean towards saying we should try passing through the connector schema as-is, augmented with x-infer-schema, and see how good or bad that is in practice. What do you think?

jshearer:
Based on our talk post-standup, we decided to keep connector schemas and fix them one-by-one. So the only functionality change here is adding `x-infer-schema`.

jgraettinger:
:+1: There's another compelling reasons for keeping connector schemas that occurred to me: it allows for connector schemas to use reduction and deletion annotations.

PR #1051, opened 2023-11-02
sqlcapture: Use 'Heartbeats' when streaming indefinitely

willdonnelly:
**Description:**

Previously after ensuring that we're caught up and all tables are fully backfilled the connector would enter a "stream forever" state in which it just processes changes indefinitely as they come in from the source database. This was implemented by simply running the "stream to watermark" function with a nonexistent watermark value, which would never be observed and so streaming would never end.

There were two reasons this behavior might be suboptimal:

  1. We're not actively confirming that the connection remains alive. In the past we've seen the occasional case (usually involving an SSH bastion intermediary) where the DB is fully idle, the underlying TCP stream gets silently disconnected, and it takes hours before TCP-level keepalives cause this to be detected by the connector.

  2. In Postgres specifically, we can only advance the replication cursor in response to transaction commits on the logical DB we're capturing from, but the WAL is shared between all logical databases on a given server. Thus if there's activity on some other database, WAL retention will grow and grow until a change is made on the captured-from database.

We could in principle have addressed (2) by spawning a worker thread as part of the PostgreSQL connection-opening logic, which would emit periodic watermark writes and thereby guarantee a consistent low rate of changes on the source DB. However because there exist possible failure modes like (1) I decided that a more general change to the SQL CDC structure was warranted.

After this PR, the "stream forever" operation now works like:

  1. Decide on a watermark UUID.
  2. Start streaming replication events until that UUID is reached.
  3. Write the watermark UUID to the database after one minute.

This ensures not only that there's a constant low level of changes moving through the system, but also that we're consistently observing our own watermark writes and remaining caught up to real time.

As a part of this change I also reworked the "progress logging" code, since we now expect change streaming to consistently stop every minute or so and this made the old "periodically log progress" behavior a poor fit. The resulting sequence of log messages looks something like:

    time="2023-11-02T15:25:46-05:00" level=info msg="no tables currently require backfilling"
    time="2023-11-02T15:25:46-05:00" level=info msg="streaming replication events indefinitely"
    time="2023-11-02T15:25:46-05:00" level=info msg="streaming to watermark" watermark=77c64b9d-dedc-4e43-b601-0f1764eb7f2f
    time="2023-11-02T15:26:46-05:00" level=info msg="processed replication events" events=41093
    time="2023-11-02T15:26:46-05:00" level=info msg="streaming to watermark" watermark=3f4ddb65-d948-4c6d-808d-e774c190d85e
    time="2023-11-02T15:27:46-05:00" level=info msg="processed replication events" events=42155
    time="2023-11-02T15:27:46-05:00" level=info msg="streaming to watermark" watermark=631b1118-88f4-4a8e-b327-2ed286c4fe5b
    time="2023-11-02T15:28:46-05:00" level=info msg="processed replication events" events=42393

I have verified (by running TestGeneric/StressCorrectness with the tuning constants turned up much higher) that the new logic should work just as well as the old in all three SQL CDC connectors.

**Workflow steps:**

Nothing should change as far as users are concerned, except that idle Postgres databases will no longer have excessive WAL retention when another database on the server is active, and we might be a bit more robust about detecting other edge case failures or connection disruption.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1051)

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -14,6 +14,7 @@ import (
 	boilerplate "github.com/estuary/connectors/source-boilerplate"
 	"github.com/google/uuid"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 )
 
 // PersistentState represents the part of a connector's state which can be serialized
@@ -117,9 +118,9 @@ type Capture struct {
 }
 
 const (
-	nonexistentWatermark   = "nonexistent-watermark" // The watermark which will be used for the final "tailing" stream call.
-	streamIdleWarning      = 60 * time.Second        // After `streamIdleWarning` has elapsed since the last replication event, we log a warning.
-	streamProgressInterval = 60 * time.Second        // After `streamProgressInterval` the replication streaming code may log a progress report.
+	heartbeatWatermarkInterval = 60 * time.Second // When streaming indefinitely, write a heartbeat watermark this often
+	streamIdleWarning          = 60 * time.Second // After `streamIdleWarning` has elapsed since the last replication event, we log a warning.
+	streamProgressInterval     = 60 * time.Second // After `streamProgressInterval` the replication streaming code may log a progress report.
 )
 
 // Run is the top level entry point of the capture process.
@@ -283,7 +284,7 @@ func (c *Capture) Run(ctx context.Context) (err error) {
 	// gets emitted. This ensures that streams reliably transition into the Active
 	// state on the first connector run even if there are no replication events
 	// occurring.
-	logrus.Debug("no tables currently require backfilling")
+	logrus.Info("no tables currently require backfilling")
 	if err := c.emitState(); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


sqlcapture/capture.go
```diff
@@ -454,50 +417,70 @@ func (c *Capture) streamToWatermark(ctx context.Context, replStream ReplicationS
...
+				return workerCtx.Err()
+			case <-time.After(heartbeatWatermarkInterval):
+				if err := c.Database.WriteWatermark(workerCtx, watermark); err != nil {
+					return fmt.Errorf("error writing next watermark: %w", err)
+				}
+			}
+			return nil
+		})
+
+		// Perform replication streaming until the watermark is reached (or the
```
jgraettinger, line 437: you thought about running this with `group.Go` ?

In this context, because your using `workerCtx` I agree this implementation seems correct. It just threw my pattern-matching instinct which was looking for multiple calls to `group.Go` with a single blocking Wait that's looking for the first returned error.

Though I suppose a difference is that an error to write the watermark could be buried by a secondary context cancellation error of `streamCatchup` ?

jgraettinger (approved):
LGTM with one comment below

sqlcapture/capture.go
willdonnelly: That's a good point, I hadn't thought about the potential for `streamCatchup` to return a useless context-cancellation error and hide the real failure. I'll change this so that both operations run via `group.Go`.

PR #1052, opened 2023-11-03
source-postgres: Make watermarks permissions error reporting work

willdonnelly:
**Description:**

Previously it was possible in some cases to set up the watermarks table such that the capture user didn't have permission to update the table *and yet those failures weren't reported* which is bad.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1052)


willdonnelly:
Merging without review, because CI passes and it's basically just `s/Query/Exec/` in the watermark write function, which would be more correct (since no results are expected) even if it weren't found to fix the error reporting issue.

jgraettinger (commented):
LGTM

PR #1054, opened 2023-11-03
source-postgres: Truncate oversize string/byte/JSON columns

willdonnelly:
**Description:**

Adds some code to truncate overly large (defined here as >1MiB, an arbitrarily chosen value which should be large enough for most reasonable purposes and small enough that even several such fields is still within the Flow document size limits) string/byte/JSON columns coming from Postgres.

This logic isn't bulletproof: I'm sure there are plenty of other column types which could in theory cause trouble, including array columns. But this handles the four most common column types which could contain megabytes of data, and in practice it will probably be sufficient for now.

In the longer term, this truncation function should be made a part of the common `sqlcapture` logic and should be done as part of the final JSON serialization of a change event, so that it will apply to every possible value type in all three databases.

But that's a bit more complicated and I wanted to get the minimum useful version of this feature out sooner.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1054)

### DIFF
--- a/source-postgres/discovery.go
+++ b/source-postgres/discovery.go
@@ -21,7 +21,7 @@ const (
 	infinityTimestamp         = "9999-12-31T23:59:59Z"
 	negativeInfinityTimestamp = "0000-01-01T00:00:00Z"
 	RFC3339TimeFormat         = "15:04:05.999999999Z07:00"
-	truncateColumnThreshold   = 1 * 1024 * 1024 // Arbitrarily selected value
+	truncateColumnThreshold   = 8 * 1024 * 1024 // Arbitrarily selected value
 )
 
 // DiscoverTables queries the database for information about tables available for capture.
--- a/source-postgres/discovery.go
+++ b/source-postgres/discovery.go
@@ -21,6 +21,7 @@ const (
 	infinityTimestamp         = "9999-12-31T23:59:59Z"
 	negativeInfinityTimestamp = "0000-01-01T00:00:00Z"
 	RFC3339TimeFormat         = "15:04:05.999999999Z07:00"
+	truncateColumnThreshold   = 1 * 1024 * 1024 // Arbitrarily selected value
 )
 
 // DiscoverTables queries the database for information about tables available for capture.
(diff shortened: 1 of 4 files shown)


source-postgres/discovery.go
```diff
@@ -250,8 +257,45 @@ func translateRecordField(column *sqlcapture.ColumnInfo, val interface{}) (inter
 			fmt.Fprintf(s, "%02x", x[i])
 		}
 		return s.String(), nil
+	case string:
+		if len(x) > truncateColumnThreshold {
```
williamhbaker, line 261: Would it be practical to add some indication that these fields have been truncated if they do get truncated? It will be very obvious for some types from the output being `{"flow_truncated":true,"original_size":%d}` but for ex. a string that gets truncated won't have that kind of indication from what I can tell.

willdonnelly: Yeah, we can implement pretty much whatever transformation we want here, I just went with what was described to me in the initial feature request. Do you have anything specific in mind?

williamhbaker (approved):
Code LGTM.

A higher-level question about rolling this out: We have definitely been capturing documents with text fields larger than 1 MB but less than the system limits and even often materializing these fields too, if they don't cause an error in the endpoint (which they usually do, but not _always_ AFAIK). So maybe we'd want to do some kind of communications prior to deploying this, and giving users a heads-up that it is going to start happening.

dyaffe:
From FT's documentation, they do this: If a value in a STRING column exceeds the maximum allowed length, it is truncated to the maximum length allowed.

Because of that, I think I'd be okay with just starting to do the truncation.  The only question I have is whether 1 MB is the right size.  It could make sense for us to accept larger documents into Flow than that and have the materialization connector decide what to do with them.  Even 8 MB docs could make sense for instance if the redshift connector trimmed those to 1 MB and the snowflake one did so to 4 MB or something.

willdonnelly:
Yeah, the issue with larger limits here is that this is a _per-column_ limit which we hope will be enough to keep the _total_ document size within Flow limits in most cases. So long as there's just one massive column we can easily push it higher, but this means that fewer massive columns are required to push a document over the Flow limit.

But IIRC the Flow limit is 64MiB, so even an 8MiB limit per column is probably still plenty.

In the long run it would be better to integrate these limits into the overall JSON document serialization and do some sort of dynamic "truncate all oversize columns equally to the maximum threshold which keeps the overall document within the 64MiB limit" logic, but that's a lot more complex .

dyaffe:
In general we've seen 1-2 columns that are large but I've never so far seen more than 2.  An 8 MB per-column system limit feels pretty defensible and I'd imagine we could tune this later if we had to?

willdonnelly:
Yeah, it's just a constant defining the maximum size in bytes and should be trivial to adjust in the future if needed. I'll bump it up to 8MiB and then get this merged as soon as I'm back at my computer.

willdonnelly:
I've increased the truncation threshold to 8MiB and am merging this now.

Note that binary blob columns are serialized as base64 strings but the size limit is applied to the actual byte data, meaning that a binary blob can be around 10-11MiB when serialized in a Flow document. This is unlikely to matter in practice so I've made no effort to address that.

PR #1055, opened 2023-11-06
sqlcapture: Reset replication cursor when no bindings

willdonnelly:
**Description:**

Resets the replication cursor when there are no bindings.

This should allow the user to get things working again after WAL deletion events by disabling and then re-enabling all bindings, rather than having to use the Flow CLI to download their task spec, rename it, and publish that.

**Workflow steps:**

If I understand correctly that disabled bindings simply aren't sent to the connector in the bindings list at all (which I'm fairly sure is the case, since otherwise that wouldn't work at all?), then disabling all bindings in the UI should allow a capture task to recover after WAL deletion renders the old restart checkpoint no longer usable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1055)

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -403,6 +403,17 @@ func (c *Capture) updateState(ctx context.Context) error {
 		}
 	}
 
+	// If no bindings are present, forget the old replication cursor. This is safe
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #1057, opened 2023-11-06
Update Singer `CaptureShim` to work with v`0.33`

jshearer:
This updates the Singer shim to work with changes in `singer-sdk` v`0.33`. Specifically, we have to pull `sync_all()` in house because the way it references `write_message()` is not compatible with our "clever" solution to monkey-patching that method to capture its output. My write-up on this from a slack thread:

-   When we import a module in python, it imports the `__init__.py`s above that module first. So importing `singer_sdk._singerlib` will import `singer_sdk/__init__.py`, and `singer_sdk/_singerlib/__init__.py`. See the screenshot for the import tree of `singer_sdk._singerlib.messages`
    Screen Shot 2023-11-03 at 14 01 45

-   In order to monkeypatch `write_message` we need to get a reference to its containing module `singer_sdk._singerlib.messages` *before* that is used by anything else, specifically `singer_sdk.tap_base` which is where the problematic import lives. But the problem is, `singer_sdk/__init__.py` also imports `singer_sdk.tap_base` which imports `singer_sdk._singerlib.messages`, so by trying to import `singer_sdk._singerlib.messages` to monkey-patch `write_message`, we also end up triggering the import of `tap_base` which imports `write_message` directly, thereby preventing us from overwriting it in time

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1057)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,7 +18,7 @@
                     "remoteRoot": "."
                 }
             ],
-            "justMyCode": true
+            "justMyCode": false
         }
     ]
 }
--- a/python/flow_sdk/shim_singer_sdk.py
+++ b/python/flow_sdk/shim_singer_sdk.py
@@ -1,22 +1,20 @@
 import os
+import sys
 import singer_sdk._singerlib as singer
 import singer_sdk.metrics
+from singer_sdk.streams import Stream
 import typing as t
 
 from .capture import Connector, request, response, Response
-from . import flow, logger 
+from . import flow, init_logger
 
 class Config(t.TypedDict):
     pass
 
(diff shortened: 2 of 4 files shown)


jgraettinger (approved):
LGTM

Do you think there's a change that's worth trying to upstream, to facilitate this?

jshearer:
If there is, it's more in-depth than the change here. It would involve setting up some sort of configurable indirection for message writing and then making sure that everything currently calling `write_message` uses that instead. Plus, we would still want to drive `sync()` ourselves in order to get parallel streams, whenever we decide to do that.

PR #1058, opened 2023-11-06
Add `source-criteo`

jshearer:
**Description:**

* Pull in [`tap-criteo`](https://github.com/edgarrmondragon/tap-criteo), and fix what I believe is a typo in schemas
* Configure `tap-criteo` with sops-encrypted creds, and document the process
* Fix Singer shim validation: for some reason, `singer-sdk`'s `Tap.run_connection_test()` tests _all_ streams that the connector reports, rather than just those that are selected. The consequence is that `validate` is far too greedy, trying to read data even from disabled streams. This was showing up as auth errors since the Criteo creds we have only grant access to a single stream.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1058)


jgraettinger (commented):
This is great, my only concern is the "Unknown" license type of the import commit.

The repo does declare a license (apache 2) here: https://github.com/edgarrmondragon/tap-criteo/blob/683463aeb9aea32593432d703f22cc15c2da1750/pyproject.toml#L14

jshearer:
Ah okay, I didn't think to look in `pyproject.toml` -- I was just looking for a `LICENSE.md`. I will update the commit message 👍

jgraettinger (approved):
LGTM but I don't understand the DOCS_URL environment variable: why wouldn't that be part of the shim signature called by `main.py`? It could be an optional kwarg for example.

jshearer:
It's because of CI, we build everything with [`connector-variant.Dockerfile`](https://github.com/estuary/connectors/blob/main/connector-variant.Dockerfile) which allows for different connector variants having different docs URLs. 

> The connector must support optionally reading the DOCS_URL environment variable
# for its spec response.

PR #1059, opened 2023-11-07
source-postgres: Replication diagnostics and watermark logging

willdonnelly:
**Description:**

Adds a few queries to be run as part of the "automated replication diagnostics" triggered when watermark streaming takes unexpectedly long to complete.

Also adds some additional logging calls in the deferred watermark- writing worker spawned by `streamForever`, so that:

  1. There is no way for the worker to exit without logging why it did so.
  2. At debug level, we can see when the watermark write actually succeeded, since it's theoretically possible that this could be some time after we dispatched the write query.

**Workflow steps:**

No user action is required, this is all internal debug logging stuff.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1059)

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -654,6 +654,41 @@ func (s *replicationStream) Close(ctx context.Context) error {
 }
 
 func (db *postgresDatabase) ReplicationDiagnostics(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #1061, opened 2023-11-07
souce-mongodb: limit datetimes to min/max range

mdibaiee:
**Description:**

- source-mongodb's bson types do not support encoding datetimes outside of the range of 0 to 9999 to JSON. This pull-request puts a constraint over datetime values so that they cannot exceed this range, and if so they are normalized to the min/max value.
- Tested using the newly added test cases for date-time

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1061)


mdibaiee:
Yes, otherwise `json.Marshal` fails on values outside of the safe range

PR #1062, opened 2023-11-07
materialize-google-pubsub: hex-encode ordering key

williamhbaker:
**Description:**

The packed key can be an arbitrary sequence of bytes, including sequences of non-UTF8 characters when converted to a string. If used as an ordering key, this causes a PubSub error.

Instead, hex-encode the packed key to use as an ordering key, which is always valid UTF8.

**Workflow steps:**

**Notes for reviewers:**

This change to the ordering key could hypothetically result in out of order reading of messages with the same packed key from a materialized topic where the key was originally published previously as a raw string, but then an update gets published again with the hex-encoded string. Since there is only 1 active pubsub materialization and its currently broken because of this issue I don't think it is likely to be an actual problem in practice.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1062)

### DIFF
--- a/materialize-google-pubsub/transactor.go
+++ b/materialize-google-pubsub/transactor.go
@@ -1,11 +1,13 @@
 package connector
 
 import (
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 
 	"cloud.google.com/go/pubsub"
 	pm "github.com/estuary/flow/go/protocols/materialize"
+	"github.com/minio/highwayhash"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

materialize-google-pubsub/transactor.go
```diff
@@ -34,7 +50,7 @@ func (t *transactor) Store(it *pm.StoreIterator) (pm.StartCommitFunc, error) {
 
 		msg := &pubsub.Message{
 			Data:        it.RawJSON,
-			OrderingKey: string(it.PackedKey), // Allows for reading of messages for the same key in order.
+			OrderingKey: fmt.Sprintf("%d", PackedKeyHash_HH64(it.PackedKey)),
```
jgraettinger, line 53: nity-nit: it's merged and not a big deal, but `%08x` would make these uniform size and lexicographically comparable: https://go.dev/play/p/D40eHGzzjWU

PR #1063, opened 2023-11-08
Return `resource_path` in discovered response

psFried:
**Description:**

Updates all capture connectors to start returning the `resource_path` as part of discover responses.

**Notes for reviewers:**

My main goal in this was obviously to avoid making any changes to existing `resource_path`s, since changing existing resource paths seems like it could break something during discovery merge. Mostly, this was easy to do, since I was able to just copy the behavior from `Validate` and make sure `Discover` returns the same thing. The one case where this was not possible was `source-http-ingest`, which needed to use a different resource path. See the commit message for more details on that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1063)


williamhbaker (approved):
LGTM. Props for getting all the test snapshots updated and the tests passing!

PR #1064, opened 2023-11-08
materialize-bigquery: compatibility for fields with hyphens

williamhbaker:
**Description:**

BigQuery now supports "Flexible column names", which means column names can include hyphens, among other things.

We were never normalizing hyphens as underscores in column names, so historically collections with field names containing hyphens would cause an apply failure. With BigQuery's support for "Flexible column names", the table will now actually be created, but the external tables we use to load data still don't support these special column names.

This commit replaces the column names in the generated JSON files and SQL queries with generic placeholders, so that column names with dashes aren't a problem for external tables. With this we can create & materialize to tables with hyphens in the column names.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1064)


jgraettinger (approved):
LGTM!

PR #1065, opened 2023-11-09
Add `source-shopify`

jshearer:
**Description:**

Add a `source-shopify` connector, including OAuth config. I had to disable some of the streams until we apply for/get approved for [special](https://shopify.dev/docs/api/usage/access-scopes#orders-permissions) [OAuth scopes](https://shopify.dev/docs/api/usage/access-scopes#protected-customer-data-permissions), but everything else appears to work well!

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1065)


jgraettinger (approved):
LGTM

PR #1068, opened 2023-11-10
source-mysql: Fix a rare shutdown deadlock

willdonnelly:
**Description:**

It is rare but possible (observed once ever in production) for the MySQL capture to deadlock while shutting down, because the event channel buffer is full and the replication worker is attempting to send another event, while the top-level capture loop has already stopped processing events and is waiting for the final error result from shutting down the replication worker.

The fix for this is to allow context cancellation to unblock these sends, which means that the replication worker will finish shutting down, which means it will send the final error result, which allows the top level shutdown to finish.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1068)


williamhbaker (approved):
LGTM

PR #1074, opened 2023-11-15
source-mysql: Validate that binary logging is enabled

willdonnelly:
**Description:**

Fixes the `SHOW MASTER STATUS` prerequisite check to error out when the result is empty, and adds a new check which explicitly queries whether the `log_bin` system variable is set to `ON`.

As a result, when running against a database without binary logs enabled the error will now look like:

    the capture cannot run due to the following error(s):
     - binary logging is not enabled: system variable 'log_bin' = "OFF"
     - unable to query latest binlog position (is binary logging enabled?)

This is slightly redundant in this case, but there are some cases where we could theoretically see an empty `SHOW MASTER STATUS` result even if binary logging is enabled (for instance, if the destination for binlog files is full or incorrectly permissioned so they can't actually be opened on disk), so it seemed better to have both checks.

So the explicit `log_bin` check allows us to give the user a more actionable error in that particular situation, while the "empty result" check means that other edge cases will still prevent the user from publishing a failing capture.

**Workflow steps:**

Users will no longer be able to create a MySQL / MariaDB capture against a database which has binary logging disabled.

We should mention the requirement that `log_bin = ON` in https://docs.estuary.dev/reference/Connectors/capture-connectors/mariadb/ right next to the `binlog_format = ROW` requirement.

We might also want to mention it in the MySQL docs page, but it's less critical there because the default is enabled for MySQL.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1074)

### DIFF
--- a/source-mysql/prerequisites.go
+++ b/source-mysql/prerequisites.go
@@ -23,6 +23,7 @@ func (db *mysqlDatabase) SetupPrerequisites(ctx context.Context) []error {
 	}
 
 	for _, prereq := range []func(ctx context.Context) error{
+		db.prerequisiteBinlogEnabled,
 		db.prerequisiteBinlogFormat,
 		db.prerequisiteBinlogExpiry,
 		db.prerequisiteWatermarksTable,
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #1075, opened 2023-11-15
source-http-ingest: return resource_path in discovered for real

psFried:
**Description:**

Apparently _someone_ committed a TODO instead of actually returning the resource path as part of discover. It's fixed now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1075)


willdonnelly (approved):
LGTM

Issue #968, opened 2023-09-20
materialize-bigquery: reading loaded documents is slow


williamhbaker:
Reading loaded documents for `materialize-bigquery` is done using a `RowIterator`: https://github.com/estuary/connectors/blob/5270541d48bb9cda534a50358b80c8106f5d65c3/materialize-bigquery/transactor.go#L217-L240

I don't have any reason to believe there is anything overtly wrong about reading query results like this - it is the most obvious way to do it, and is a recommended way to read query results.

It is relatively slow though. The reason for this is that BigQuery limits the amount of data that can be returned in a single "page" to 10MB (see [BigQuery docs](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query?hl=en) for `maxResults`), and API calls to re-fill the page buffer can take several seconds to complete. With many millions of documents, the amount of time it takes can add up: On a local stack with 10MM rows of synthetic data in a single materialized table, I have observed that each buffer refill request for a load query takes about 6 seconds and returns ~15,000 document, and I get a rate of reading loaded documents of ~2500 docs/second. This means it takes over an hour to process a full 10MM response of loaded documents, even though the load query itself took only a minute or so.

At a higher level I don't think there is anything wrong with our connector code or the BigQuery Go client. I have been able to produce similar long pauses between iterator buffer refills with the Python client, for example. It seems to just be a consequence of using the BigQuery REST API.

There are alternate strategies we may be able to use for reading loaded documents. The two that I have briefly investigated would likely require additional authentication scopes, and so would not be backward compatible:
- Using the [storage read API](https://cloud.google.com/bigquery/docs/reference/storage). This would involve putting the load query results in a separate table - a temporary table if possible - and using the storage read client to read the table. This certainly requires an additional permission to create the read job. I have not actually tried to see if this is any faster than the REST API, but the documentation for the storage read system suggest that it can split a table into multiple streams for reading, so it has some potential. The BigQuery Go client has a [method to enable using the storage read API](https://pkg.go.dev/cloud.google.com/go/bigquery#Client.EnableStorageReadClient), but when I tested this it had no effect - possibly because I didn't have the required auth scope.
[...]

PR #1076, opened 2023-11-16
source-bigquery-batch: Initial implementation

willdonnelly:
**Description:**

This PR implements a complete batch BigQuery capture connector including some basic test coverage. Notable points:

 - This capture uses the same sort of "user must specify cursor columns" query template behavior as the Batch MySQL connector.
 - The discovery code can only discover tables from a single "dataset" (the BigQuery word for SQL schemas). This is due to BigQuery limitations and wasn't easily avoidable. Note that it's entirely possible to configure bindings from other datasets (assuming appropriate access permissions) and that should work fine, we just can't autodiscover across datasets.
 - There's no BigQuery emulator we can run locally / in CI builds, so I've been running tests on the `pivotal-base-360421` project from my machine.
 - I have verified that datatype handling is generally sane for integer / string / boolean / timestamp / bytes / json column types, but have not exhaustively tested beyond that.

This connector is basically a copy-paste of `source-mysql-batch` which has been extensively tweaked to actually work when running against BigQuery. This brings us up to three parallel copies of largely the same logic, so we're probably at about the point where we'll need to start factoring out the commonalities, but I've refrained from doing so just yet.

I have created https://go.estuary.dev/source-bigquery-batch pointing at https://docs.estuary.dev/reference/Connectors/capture-connectors/bigquery-batch/. The latter still needs to be created, I'll see about doing that tomorrow.


source-bigquery-batch/driver.go
```diff
@@ -0,0 +1,698 @@
...
+	var keys []*discoveredPrimaryKey
+	for _, key := range keysByTable {
+		keys = append(keys, key)
+	}
+	return keys, nil
+}
+
+var databaseTypeToJSON = map[string]*jsonschema.Schema{
+	"INT64":   {Type: "integer"},
+	"FLOAT64": {Type: "number"},
```
williamhbaker, line 360: If there's a BigQuery primary key that is a `FLOAT64`, what happens to Flow collection since IIRC Flow collections can't have `number` as a primary key type?

source-bigquery-batch/driver.go
```diff
@@ -0,0 +1,698 @@
...
+}
+
+var catalogNameSanitizerRe = regexp.MustCompile(`(?i)[^a-z0-9\-_.]`)
+
+func recommendedCatalogName(table string) string {
```
williamhbaker, line 365: I know that we're still doing this everywhere, but I don't think connectors actually need to still be doing it since [the runtime does it now](https://github.com/estuary/flow/pull/1083), too. Certainly not something we need to clean up now, but seeing this tickled my memory of that change.

source-bigquery-batch/driver.go
```diff
@@ -0,0 +1,698 @@
...
+	var params []bigquery.QueryParameter
+	for idx, val := range cursorValues {
+		params = append(params, bigquery.QueryParameter{
+			Name:  fmt.Sprintf("p%d", idx),
+			Value: val,
+		})
+	}
+	q.Parameters = params
+
+	rows, err := q.Read(ctx)
```
williamhbaker, line 597: Not requesting anything different here right now, but I'm betting this is going to be relatively slow, for reasons similar to those [noted for reading loads](https://github.com/estuary/connectors/issues/968) of `materialize-bigquery`. Fortunately, I have seen that the slowness when running in our production environment isn't nearly as bad as when I've tested on a local stack, but it's still not great. Enabling the [storage read API](https://pkg.go.dev/cloud.google.com/go/bigquery#Client.EnableStorageReadClient) makes reading results _much_ faster when I've tested it, but is pretty brutal on memory use for the connector. It could be something we need to look into more down the line.

source-bigquery-batch/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# caching opportunities when no dependencies are updated.
+COPY go.* ./
+RUN go mod download
+
+COPY go                 ./go
+COPY source-boilerplate ./source-boilerplate
+
+RUN go install -v ./go/...
+RUN go install -v ./source-boilerplate/...
+RUN go install -v github.com/jackc/pgx/v4
```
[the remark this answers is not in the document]
willdonnelly: That's a dumb copy-paste oversight (twice now!). In the `source-postgres-batch` Dockerfile it serves a purpose (because getting that installed ahead of time means that connector rebuilds don't have to rebuild that branch of the dependency tree), and it's just been carried over by mistake. I'll remove that from here and the MySQL Dockerfile.

source-bigquery-batch/docker-compose.yaml
```diff
@@ -0,0 +1,17 @@
+version: "3.7"
```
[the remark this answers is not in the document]
willdonnelly: Oh yeah, I should delete this file. I spent a bunch of time trying to make things work with the emulator in this config file -- the statement that "there's no emulator" should be read as "there's no emulator that actually works well enough to run tests against", and I know that from having wasted a decent chunk of time trying it.

source-bigquery-batch/driver.go
```diff
@@ -0,0 +1,698 @@
...
+
+const queryDiscoverTables = `SELECT table_schema, table_name, table_type FROM %[1]s.INFORMATION_SCHEMA.TABLES;`
+
+type discoveredTable struct {
+	Schema string
+	Name   string
+	Type   string // Usually 'BASE TABLE' or 'VIEW'
+}
+
+func discoverTables(ctx context.Context, db *bigquery.Client, dataset string) ([]*discoveredTable, error) {
```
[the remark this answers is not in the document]
willdonnelly: Oh, that would work. I was aware that it could be done by executing more SQL queries to list them all, but that seemed overly complicated and so I just did the simplest thing that could work for now. I'm not sure how people use multiple datasets in BigQuery, but it seemed like common use-cases probably wouldn't span multiple datasets anyway, so I'm also inclined to just leave it like this for now.

source-bigquery-batch/driver.go
willdonnelly: That's a trick question, BigQuery doesn't allow tables to have floats as primary keys (which is rather weird given that the primary key property is purely advisory and even uniqueness isn't actually enforced AFAIK).

Which, now that you mention it, makes it a bit silly for me to have added it here in the first place, since this type translation is only applied to primary keys in batch capture discovery. But on the other hand it is the correct mapping of that type, so ¯\_(ツ)_/¯

source-bigquery-batch/driver.go
willdonnelly: > Enabling the [storage read API](https://pkg.go.dev/cloud.google.com/go/bigquery#Client.EnableStorageReadClient) makes reading results much faster when I've tested it, but is pretty brutal on memory use for the connector

My experience has generally been that trading off increased memory usage for throughput is a bad move -- nearly all of the work getting SQL CDC memory usage under control boiled down to "remove some buffering that I previously added as a throughput optimization", so I'd say we shouldn't do that unless/until we have evidence that it's not fast enough as-is for real-world uses.

source-bigquery-batch/driver.go
willdonnelly: Ah, I wasn't aware of that. I'm going to leave this as-is for symmetry with the other connectors for now, but we should definitely start removing that from connectors sometime soon.

PR #1079, opened 2023-11-20
Update Singer shim to emit state checkpoints

jshearer:
**Description:**

We don't have to do any multiplexing of stream-specific state for Singer as far as I can tell because the Meltano SDK manages mapping stream-specific into keys in the `"bookmarks"` state-key under the hood. All we have to do is persist the state values, and provide the latest state to the `Tap()` constructor when restarting.

Also update specs for `source-criteo` and `source-shopify` to make them work better in the UI

**Note:** Airbyte state is more complex and I'm still researching how to handle it, so this just covers Singer state for the moment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1079)

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -28,6 +28,9 @@ A Python connector is composed of a few critical pieces:
   ).main()
   ```
 
+### Requirements
+In order to build python connectors, you will need to install [Poetry](https://python-poetry.org/docs/#installation).
+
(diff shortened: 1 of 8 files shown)


jgraettinger (approved):
LGTM

PR #1081, opened 2023-11-21
source-mongodb: do not set reduce: merge strategy for mongo collections

mdibaiee:
**Description:**

- we initially did this in order to be consistent in behaviour with our sql captures which emit the full document and `_meta/op: "d"`, however the reduction strategy is problematic when it comes to documents that have one of their columns removed in an update, however because of the reduction strategy they retain that column in the materialization. so we have decided to revert this change to just have mongodb emit a document with id and `_meta/op: "d"` that ends up clearing all other columns in the materialization
- Tested by running tests, and manually ran a test where I deleted a field from a document (using `$unset`) and in the materialization the column was set to null.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1081)

### DIFF
--- a/source-mongodb/.snapshots/TestDiscover
+++ b/source-mongodb/.snapshots/TestDiscover
@@ -29,14 +29,8 @@ Binding 0:
               "description": "Change operation type: 'c' Create/Insert 'u' Update 'd' Delete."
             }
           },
-          "type": "object",
-          "reduce": {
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #1082, opened 2023-11-21
source-mongodb: discover multiple databases in an instance

mdibaiee:
**Description:**

- Our capture logic allows for capturing data from multiple databases since we rely on the resource config's `Database` field to find the right database to connect to. We also use an instance-wide change stream, and we already have a check in place which ensures that if multiple databases are specified in resource configs, we error if they try to run a capture without access to an instance-level change stream.
- This change changes the discovery logic to allow the user to discover across an instance, optionally specifying a comma-separated list of database names to capture

**Workflow steps:**

**Notes for reviewers:**


source-mongodb/main.go
```diff
@@ -97,9 +96,7 @@ func (c *config) ToURI() string {
 
 	uri.User = url.UserPassword(c.User, c.Password)
 
-	if c.Database != "" {
-		uri.Path = "/" + c.Database
-	}
+	uri.Path = "/"
```
williamhbaker, line 99: nit: Remove this line entirely?

williamhbaker (approved):
LGTM

source-mongodb/main.go
mdibaiee: @williamhbaker I actually initially did, but MongoDB client complains about this, as it seems to add its own `?query=` parameters, and if there is no slash, it considers the URI to be invalid without a slash:

```
connecting to database: error parsing uri: must have a / before the query ?
```

PR #1083, opened 2023-11-21
Phil/resource path pointers

psFried:
**Description:**

Updates the batch sql and webhook capture connectors to return `resource_path_pointers` as part of their spec response. This is an incremental step toward updating all of the connectors, which is meant to address the biggest current pain points (sql batch connectors). The `source-http-ingest` connector is only updated now because I had already manually updated it in production.

The ultimate goal of this is to allow taking advantage of the new discover merge behavior, which relies on the presence of `resource_path_pointers`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1083)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/databricks/databricks-sdk-go v0.24.0
 	github.com/databricks/databricks-sql-go v1.5.1
 	github.com/elastic/go-elasticsearch/v8 v8.9.0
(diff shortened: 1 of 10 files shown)


williamhbaker (approved):
LGTM aside from the test failures. Maybe needs a `go mod tidy`? Also I am wondering if you need to update the spec test snapshots.

psFried:
source-mysql tests are failing due to an unrelated compilation error. I'm gonna merge this as is, and try to fix source-mysql in a separate PR.

PR #1086, opened 2023-11-22
materializations: efficient applies for SQL materializations

williamhbaker:
**Description:**

This set of changes is for making SQL materialization `Apply` operations much more efficient.

The current issue is that if there are a large number of actions that need to be taken in an `Apply`, it might time out if the connector is not able to efficiently carry out all the actions in the ~5 minute time limit. There's a couple of related, contributing factors to this: 1) SQL materializations are very inefficient at applying changes, largely because they have to try each one individually because they don't know what the actual state of the extant table is, and 2) Progress from a previous apply attempts cannot be resumed from, since the set of changes is computed only from the proposed spec vs. a previously persisted spec, and a spec does not get persisted unless the entire set of actions is completed.

The bulk of the work in these changes is adding the capability for all SQL materializations to introspect the state of existing tables and columns via whatever `INFORMATION_SCHEMA` view the endpoint provides, and for filtering out from the computed actions those that aren't necessary. This allows for much more efficient queries to accomplish the remaining actions, and for incremental progress to be made across `Apply` invocations if the initial attempts fail for some reason.

In addition to the expanded test coverage around `Apply`'s, I did _a lot_ of manual testing for these changes, as they touch every single SQL materialization. Mostly the automated tests cover the relatively happy paths, whereas manual testing was done to confirm behavior with really heinous schema, table, and column names & quantities.

See individual commit messages for more details, and it may be useful to review commit-by-commit.

**Workflow steps:**

You can successfully apply changes with lots and lots of columns, and/or with many new tables. I have verified that all SQL materializations are able to create 400 new tables in a single `Apply` now, though they may still fail for other reasons with that many tables.

`materialize-databricks` doesn't have an efficient way of dropping NOT NULL's for large amounts of columns that I could find, but other than that all SQL materializations are also able to add 500+ columns and/or alter 500+ columns without timing out, usually within just a couple of seconds.

For the broad removal of the "Additional Table Create SQL" configuration, there probably should be some doc updates, although I'm not sure how much we have documented that off hand. It'll only be in `materialize-postgres` (and its variants) now though.

**Notes for reviewers:**

I'm pretty far from satisfied with how the runtime parameters are passed to the `sql.Dialect` constructors and the proliferation of untyped magic strings that results. I had to extend this to additional materializations as part of this work, and I'd imagine it will need to continue extending for future work (dynamic type conversion based on column types?). I couldn't figure out a better way to do it right now though.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1086)


materialize-snowflake/.snapshots/TestSQLGeneration
```diff
@@ -1,25 +1,25 @@
 --- Begin "a-schema".target_table createTargetTable ---
-  CREATE TABLE IF NOT EXISTS "a-schema".target_table (
```
williamhbaker, line 2: I cleaned up some extra spacing that was in these query templates, which where a lot of the diff on the file is coming from.

materialize-bigquery/sqlgen.go
```diff
@@ -123,6 +129,32 @@ CLUSTER BY {{ range $ind, $key := $.Keys }}
...
+ALTER TABLE {{$.Identifier}}
+{{- range $ind, $col := $.AddColumns }}
+	{{- if $ind }},{{ end }}
+	ADD COLUMN {{$col.Identifier}} {{$col.NullableDDL}}
+{{- end }};
+{{- end -}}
+{{- if $.DropNotNulls -}}
+{{- if $.AddColumns }}
+
+{{ end -}}
```
mdibaiee, line 149: Is this newline break intentional?

mdibaiee (approved):
LGTM, thanks Will!

PR #1087, opened 2023-11-27
Phil/all resource path pointers

psFried:
**Description:**

Updates all remaining capture connectors in this repo to return `resource_path_pointers` as part of their `spec` response. This allows them to take advantage of the new discover merge behavior, which merges bindings based on the extracted resource paths.

I removed the `ResourcePath` from the Discovered responses as part of this, since those were only recently introduced. I didn't remove the resourcePath from the Validated responses, since our validation logic still requires them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1087)

### DIFF
--- a/source-firestore/.snapshots/TestDeletions-two
+++ b/source-firestore/.snapshots/TestDeletions-two
@@ -1,9 +1,4 @@
 # ================================
-# Collection "flow_source_tests/*/docs": 2 Documents
-# ================================
-{"_meta":{"path":"projects/project-id-123456/databases/(default)/documents/flow_source_tests/TestDeletions/docs/3","mtime":"<TIMESTAMP>","delete":true}}
-{"_meta":{"path":"projects/project-id-123456/databases/(default)/documents/flow_source_tests/TestDeletions/docs/4","mtime":"<TIMESTAMP>","delete":true}}
-# ================================
 # Final State Checkpoint
 # ================================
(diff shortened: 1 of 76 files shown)


source-firestore/.snapshots/TestDeletions-two
```diff
@@ -1,9 +1,4 @@
 # ================================
-# Collection "flow_source_tests/*/docs": 2 Documents
-# ================================
-{"_meta":{"path":"projects/project-id-123456/databases/(default)/documents/flow_source_tests/TestDeletions/docs/3","mtime":"<TIMESTAMP>","delete":true}}
```
williamhbaker, line 4: Do you know why this snapshot changed?

williamhbaker (approved):
LGTM - I think there are more connectors that need updated though? `source-alpaca` and `source-dynamodb` - maybe those are the only 2 missing.

source-firestore/.snapshots/TestDeletions-two
psFried: Not fully, no. Best I can tell, the tests were just broken. Which makes sense, given that they aren't run as part of CI, so it'd be easy for them to get outdated. But the test in this case clearly deletes those two documents from the source prior to verifying the snapshot. And the behavior of them being missing from the snapshot seems consistent with the `TestDeletions-one` snapshot, which also doesn't include the deleted documents.

williamhbaker (approved):
LGTM!

PR #1088, opened 2023-11-28
materialize-snowflake: run load and store queries concurrently

mdibaiee:
**Description:**

- Run load and store operations concurrently for up to 5 bindings at a time
- Also upload files concurrently by having multiple `putWorker`s
- Tested by running the integration tests as well as manually tested on more than 5 bindings at a time to check the limiting logic

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1088)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -377,8 +377,8 @@ func (c client) withDB(fn func(*stdsql.DB) error) error {
 }
 
 type transactor struct {
-	cfg    *config
-	tenant string
+	cfg *config
+	db  *stdsql.DB
 	// Variables exclusively used by Load.
 	load struct {
 		conn *stdsql.Conn
@@ -405,10 +405,15 @@ func newTransactor(
 
 	dialect := snowflakeDialect(cfg.Schema)
 
+	db, err := stdsql.Open("snowflake", cfg.ToURI(ep.Tenant))
+	if err != nil {
+		return nil, fmt.Errorf("newTransactor stdsql.Open: %w", err)
+	}
+
 	var d = &transactor{
 		cfg:       cfg,
 		templates: renderTemplates(dialect),
-		tenant:    ep.Tenant,
+		db:        db,
 	}
 
 	if d.updateDelay, err = sql.ParseDelay(cfg.Advanced.UpdateDelay); err != nil {
@@ -495,7 +500,7 @@ func (d *transactor) Load(it *pm.LoadIterator, loaded func(int, json.RawMessage)
 	for it.Next() {
 		var b = d.bindings[it.Binding]
 
-		if err := b.load.stage.start(ctx, d.load.conn); err != nil {
+		if err := b.load.stage.start(ctx, d.db); err != nil {
 			return err
 		} else if converted, err := b.target.ConvertKey(it.Key); err != nil {
 			return fmt.Errorf("converting Load key: %w", err)
@@ -531,10 +536,6 @@ func (d *transactor) Load(it *pm.LoadIterator, loaded func(int, json.RawMessage)
 
 	// In order for the concurrent requests below to be actually run concurrently we need
 	// a separate connection for each
-	db, err := stdsql.Open("snowflake", d.cfg.ToURI(d.tenant))
-	if err != nil {
-		return fmt.Errorf("load stdsql.Open: %w", err)
-	}
 
 	for iLoop, queryLoop := range subqueries {
 		var query = queryLoop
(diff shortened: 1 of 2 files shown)


materialize-snowflake/snowflake.go
```diff
@@ -547,38 +550,65 @@ func (d *transactor) Load(it *pm.LoadIterator, loaded func(int, json.RawMessage)
...
+	for iLoop, queryLoop := range subqueries {
+		var query = queryLoop
+		var i = iLoop
+		group.Go(func() error {
+			var b = d.bindings[i]
+
+			log.WithField("table", b.target.Identifier).Info("load: starting querying documents")
+			// Issue a join of the target table and (now staged) load keys,
+			// and send results to the |loaded| callback.
+			rows, err := d.load.conn.QueryContext(sf.WithStreamDownloader(groupCtx), query)
```
williamhbaker, line 567: I don't think using the same connection for all goroutines is going to result in a throughput benefit. A single connection must execute queries sequentially, so sharing this same connection will still end up with the load queries still being run one after another. I do think a `*sql.Conn` is thread safe (although I'm not 100% sure), but it [has locking](https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/database/sql/sql.go;l=504-524;drc=894d24d617bb72d6e1bed7b143f9f7a0ac16b844) that will result in concurrent calls waiting for each other to complete.

Using a separate connection for each load goroutine seems like a viable option though. It would probably be fine to just do `db.QueryContext` and let the `db` manage connection pooling, rather than keeping a reference to one specific connection for the life of the transactor.

materialize-snowflake/snowflake.go
```diff
@@ -628,30 +660,42 @@ func (d *transactor) commit(ctx context.Context) error {
...
+		var b = bLoop
+		group.Go(func() error {
+			if !b.store.stage.started {
+				// No table update required
+			} else if err := b.store.stage.flush(); err != nil {
+				return err
+			} else if !b.store.mustMerge {
+				log.WithField("table", b.target.Identifier).Info("store: starting direct copying data into table")
+				// We can issue a faster COPY INTO the target table.
+				if _, err = txn.ExecContext(groupCtx, b.store.copyInto); err != nil {
```
williamhbaker, line 677: I have a similar concern here as the one above in `Load`: `txn` uses the same connection, and any concurrent operations with that connection will have to be run sequentially even if they are called from separate goroutines (assuming the `txn` is threadsafe - otherwise something even stranger might happen).

This is harder though, since we do need to run all of these operations in the same transaction to maintain exactly-once delivery.

TBH I'm not seeing how concurrent store operations are going to help us. One wild idea would be to switch to an idempotent apply style of storing, if that is possible with Snowflake, since that could probably be done concurrently. Or, we could concurrently load staged files into staging tables, and then in a single transaction merge the staging table data into the target tables. Both of these ideas are fraught with complexity and uncertainty of what if any benefit they would provide, so I'm certainly not advocating for them right now.

mdibaiee: That's a good point. I thought it is possible to run multiple statements in a single transaction concurrently so long as we committed after having run all of the statements. In that case I'll revert this part. I think the uploading of the files is most likely to have the most significant impact on performance improvement though.

materialize-snowflake/snowflake.go
```diff
@@ -518,29 +523,63 @@ func (d *transactor) Load(it *pm.LoadIterator, loaded func(int, json.RawMessage)
...
-	// and send results to the |loaded| callback.
-	rows, err := d.load.conn.QueryContext(sf.WithStreamDownloader(ctx), loadAllSQL)
+	group, groupCtx := errgroup.WithContext(ctx)
+	group.SetLimit(MaxConcurrentLoads)
+	// Used to ensure we have no data-interleaving when calling |loaded|
+	var mutex sync.Mutex
+
+	// In order for the concurrent requests below to be actually run concurrently we need
+	// a separate connection for each
+	db, err := stdsql.Open("snowflake", d.cfg.ToURI(d.tenant))
```
williamhbaker, line 534: I'd suggest creating a `*db` in `newTransactor`, and having the `*db` reference be a property of `transactor`, basically taking the place of `tenant` as a property. If that's not possible for some reason, we'll at least want to do a deferred close of the db here.

materialize-snowflake/staged_file.go
```diff
@@ -133,10 +135,12 @@ func (f *stagedFile) start(ctx context.Context, conn *stdsql.Conn) error {
...
 	f.putFiles = make(chan string)
 
-	// Start the putWorker for this transaction.
-	f.group.Go(func() error {
-		return f.putWorker(f.groupCtx, conn, f.putFiles)
-	})
+	for i := 0; i < MaxConcurrentUploads; i++ {
+		// Start the putWorker for this transaction.
+		f.group.Go(func() error {
+			return f.putWorker(f.groupCtx, conn, f.putFiles)
```
williamhbaker, line 141: I hadn't thought of this before...all these workers are using the same `conn`, which will somewhat diminish their concurrency benefit. They won't actually be able to upload in parallel. They _will_ be able to pick up jobs off the channel though, which will effectively multiply the number of files that can be buffered on disk, so we'd probably never notice a difference. But, it would probably make the most sense for them to get a handle to a `*db` and use that instead of a specific connection.

williamhbaker (approved):
LGTM % a couple final comments

PR #1089, opened 2023-11-28
source-bigquery-batch: Minor adjustments

willdonnelly:
**Description:**

Adds a small check so that resource specs with empty cursor column names will be rejected at startup. Also throws in a couple of logging tweaks inspired by trying to read a capture task's logs yesterday, and moves the polled-at timestamp update so that it occurs _after_ acquiring the "only update one binding at a time" mutex which is obviously more correct.

These changes are made to `source-bigquery-batch` and `source-mysql-batch` for consistency. The analagous changes for batch Postgres have been added to the "bring source-postgres-batch back to feature parity" PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1089)


jgraettinger (approved):
LGTM

PR #1090, opened 2023-11-28
materialize-bigquery: use parameterized query for persisting the materialization spec

williamhbaker:
**Description:**

This is a temporary fix for allowing materialize-bigquery to use a parameterized query instead of a potentially large SQL string when inserting or updating its materialization spec.

BigQuery does not allow for SQL strings to be larger than 1 MiB in size, which is not outside the realm of possibility for complex Flow materialization specs. The size of the parameters in a parameterized query is not included in this size limit, so running the query as a parameterized query works much better for these large specs.

A more permanent fix will be rolled into #1086 shortly, but this will be able to unblock any current publications which are hitting the size limit.

I tested this with a huge materialization spec containing about 50,000 fields (100 bindings with 500 fields each). The current `dev` image fails to publish it due to the query size limit, but the image built from this branch is able to publish successfully.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1090)


jgraettinger (approved):
LGTM

PR #1099, opened 2023-11-30
materializations: support re-creating tables on increments to `backfill` counter

williamhbaker:
**Description:**

Adds support for applicable materializations to re-create (replace, or drop & re-create) tables (or their equivalent) on increments to the `backfill` counter for the binding.

Applicable materializations are those that persist a Flow materialization spec and produce constraints for proposed binding changes based on that. This is all of the SQL materializations + `materialize-dynamodb` & `materialize-elasticsearch`.

Generally we'll try to do the replacement as an atomic action if the endpoint supports it, either through a `CREATE OR REPLACE TABLE ...` statement or with a transaction to drop & create the table anew. Some endpoints don't support either of these though so for those its a best effort drop & then re-create.

Also of some note is that SQL materializations which utilize the "authoritative endpoint" pattern & fencing semantics will do a best-effort increment of the fence for the materialization prior to replacing tables. This may help prevent prior invocations of the connector from racing to commit data to the replaced table. For now I haven't implemented this transactionally with doing the actual table replacements since many systems don't support the transactional DDL required, but its something we could add to those that do later. Future work is also planned to automatically disable & re-enable materialization shards when `backfill` increases, which will further reduce the chances of these data races.

Depends on the runtime changes in https://github.com/estuary/flow/pull/1292, which add the `backfill` counter.

Reference issue: https://github.com/estuary/flow/issues/1276

**Workflow steps:**

We will need to create/update docs for the new `backfill` counter and how it can be used.

**Notes for reviewers:**


materialize-bigquery/.snapshots/TestApply
```diff
@@ -373,6 +373,33 @@
...
+    "Type": "STRING"
+  },
+  {
+    "Name": "optionalVal1",
+    "Nullable": "YES",
+    "Type": "INT64"
+  },
+  {
+    "Name": "requiredVal1",
+    "Nullable": "NO",
```
williamhbaker, line 397: In these snapshots, you can tell the table was replaced because it just has the original fields from the "base" spec, and that this `requiredVal1` is not nullable. The progression of the tests is to create the table from the "base" spec, add some columns and drop the nullability, and then re-create the table from the original "base" spec. So we can see that the table has conclusively been replaced.

materialize-dynamodb/.snapshots/TestValidate
```diff
@@ -539,4 +539,45 @@
...
+{
+	"flow_document": {
+		"type": 2,
+		"reason": "The root document is required for a standard updates materialization"
+	},
+	"flow_published_at": {
+		"type": 4,
+		"reason": "This field is able to be materialized"
+	},
+	"key": {
```
williamhbaker, line 552: For these snapshots, the main difference is from the `unsatisfiable type change` test is that the `key` field has an unsatisfiable type change, but when incrementing the backfill counter the constraints are as if it is a new binding, rather than a change from the original.

materialize-dynamodb/driver.go
```diff
@@ -360,16 +371,11 @@ func resolveResourceConfig(specJson json.RawMessage) (resource, error) {
 
 func createTable(
 	ctx context.Context,
-	dryRun bool,
```
williamhbaker, line 363: This wasn't needed, since `doAction` already accounts for the dry run.

materialize-mysql/sqlgen.go
```diff
@@ -136,6 +136,15 @@ CREATE TABLE IF NOT EXISTS {{$.Identifier}} (
 ) CHARACTER SET=utf8mb4 COLLATE=utf8mb4_bin {{- if $.Comment }} COMMENT={{Literal $.Comment}} {{- end }};
 {{ end }}
 
+-- Templated replacement of a materialized table. MySQL doesn't support
+-- transactional DDL and also doesn't have a CREATE OR REPLACE TABLE
+-- type of operation so this is the best we can do.
+
+{{ define "replaceTargetTable" }}
+DROP TABLE IF EXISTS {{$.Identifier}};
+{{ template "createTargetTable" . }}
```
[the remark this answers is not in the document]
williamhbaker: Hmmm good catch I hadn't noticed that about multi-statement queries with the MySQL driver. That seems like a pretty bad limitation, and along with the weirdness about query parameters I've opted to just not enable multi-statement queries and instead run two separate queries for table replacement, where the first one drops the existing table and the second one creates it.

mdibaiee (approved):

PR #1101, opened 2023-12-04
sqlcapture: use stateKey for tracking binding state 

williamhbaker:
**Description:**

Adds support for state keys to SQL CDC captures.

The general idea is that a given stream (which is the combined schema & table name of a captured table) has its state accessible in the driver checkpoint via the runtime-provided `stateKey`.

After some consideration, for the moment I have settled on representing this association between `streamID` and `stateKey` at the `sqlcapture` "framework" level, rather than threading it deeper into individual captures driven by the framework. This seems to work out decently well since a replication stream change event already has the information needed to construct a `streamID`, but getting the `stateKey` out of this requires some higher-level knowledge, much like how getting the correct binding index to emit a document also requires some additional knowledge.

In addition to the automated tests, I manually tested this by:
1. Create a capture of multiple tables using the old connector image
2. Switch to the new image, and see that the state was migrated correctly...no re-backfilling or anything like that
3. Use the backfill counter to trigger a re-backfill. Removing & re-adding bindings still works for that too. Also incrementing the backfill counter for _all_ bindings will reset the cursor.

**Workflow steps:**

After this change is merged, SQL CDC captures will be able to use backfill counters to trigger a re-backfill of a captured table.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1101)


sqlcapture/capture.go
```diff
@@ -30,30 +30,35 @@ func (ps *PersistentState) Validate() error {
...
+	var bindings []*Binding
+	for _, binding := range c.Bindings {
+		if slices.Contains(modes, BackfillMode(c.State.Streams[binding.StateKey].Mode)) {
+			bindings = append(bindings, binding)
 		}
 	}
-	sort.Strings(streams)
-	return streams
+	slices.SortFunc(bindings, func(a, b *Binding) int {
+		if a.StreamID < b.StreamID {
```
willdonnelly, line 43: The body of this comparison function could just be `return strings.Compare(a.StreamID, b.StreamID)`, I think?

sqlcapture/capture.go
```diff
@@ -395,11 +402,21 @@ func (c *Capture) updateState(ctx context.Context) error {
 
 	// Likewise streams may be removed from the catalog, and we need to forget
 	// the corresponding state information.
-	for streamID, state := range c.State.Streams {
-		var _, streamExistsInCatalog = c.Bindings[streamID]
-		if !streamExistsInCatalog && state.Mode != TableModeIgnore {
-			logrus.WithField("stream", streamID).Info("stream removed from catalog")
-			c.State.Streams[streamID] = &TableState{Mode: TableModeIgnore, dirty: true}
+	// TODO(whb): We can eventually remove this when backfill counters are fully incorporated into
```
willdonnelly, line 405: Just checking my understanding here: The plan is that "backfill counter" will be incremented if the user disables+re-enables a single binding, even if they don't use the new "go re-backfill this" flow?

(Because in addition to permitting users to re-backfill a collection by toggling the binding, it's also important that this logic here currently _prevents_ the possibility of disabling a stream, missing some amount of replication events, and then re-enabling it _without_ triggering that new backfill.)

sqlcapture/capture.go
```diff
@@ -608,7 +634,8 @@ func (c *Capture) backfillStreams(ctx context.Context) error {
 }
 
 func (c *Capture) backfillStream(ctx context.Context, streamID string) error {
-	var streamState = c.State.Streams[streamID]
+	var stateKey = c.Bindings[streamID].StateKey
```
willdonnelly, line 637: It might be cleaner to modify `backfillStreams()` to provide the randomly selected binding as the argument here rather than just the name, in order to avoid indexing into the bindings map here.

willdonnelly (commented):
This looks directionally correct to me. I've added a couple of comments on specific bits, but overall this seems like the right approach.

How to transition over existing captures is a tricky question though, because the way the binding keys are implemented here it looks like the old stream IDs and the new state keys would be sharing the same "key space" during that switchover. If not for that, it seems like it ought to be reasonably straightforward to add some additional logic to `updateState()` which would check for bindings which have no state associated with `stateKey` but do have old-style state associated with the stream ID, and copy over the state appropriately.

sqlcapture/capture.go
williamhbaker: That's a good point, we'll probably need to keep this state-clearing logic around since its not really valid to disable + resume a binding without re-backfilling it; I always seem to forget about that. I don't know that we'd always want to increment the backfill counter on re-enabling a binding, and although that might be possible, I don't think we'd do it right away 🤔.

sqlcapture/discovery_test.go
```diff
@@ -16,52 +16,52 @@ func TestRecommendedCatalogName(t *testing.T) {
 		{
 			schema: "something",
 			table:  "typical",
-			want:   "something.typical",
```
williamhbaker, line 19: The changes in this file aren't from this PR, but I found that they needed updated when running all of the `sqlcapture` tests.

sqlcapture/capture.go
williamhbaker: It seems useful to still have the `streamID` available for logging and getting the `discoveryInfo`, so passing the binding would require accessing that property of the binding a bunch of times or declaring a new variable, which doesn't seem much better. Let me know if I'm missing anything though.

sqlcapture/capture.go
```diff
@@ -402,21 +448,29 @@ func (c *Capture) updateState(ctx context.Context) error {
...
-	// anew, as they should after such an event).
-	if len(c.Bindings) == 0 {
+	// If all bindings have been removed or had their backfill counter incremented, forget the old
+	// replication cursor. This is safe because logically if no streams are active or a re-backfill
+	// has been requested for them then we can't miss any events of interest when the replication
+	// stream jumps ahead, and doing this allows the user an easy recovery path after WAL deletion,
+	// they just need to either increment the backfill counter for all bindings, or disable all
+	// bindings and then re-enable them (which will cause them all to get backfilled anew, as they
+	// should after such an event).
+	if allStreamsAreNew {
```
willdonnelly, line 473: Hmm, so changing the check to work this way means that if the backfill counters for every binding are incremented all at once we'll also reset the cursor? Nice!

willdonnelly (approved):
LGTM

Everything seems in order, let's ship it.

PR #1103, opened 2023-12-04
materialize-postgres: remove null bytes from string values

williamhbaker:
**Description:**

`materialize-postgres` does not allow string values to have the null byte sequence "\u0000" anywhere in them. This changes `materialize-postgres` to strip these values out of materialized string fields.

Completely removing the "\u0000" (replacing with "") seemed most consistent with other systems that do handle strings with null bytes, like Snowflake and BigQuery, since they completely strip out the null bytes rather than replacing them with some sentinel like a space etc.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-postgres/sqlgen.go
+++ b/materialize-postgres/sqlgen.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	sql "github.com/estuary/connectors/materialize-sql"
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #1106, opened 2023-12-05
source-mysql: Ignore `DROP FUNCTION` query events

willdonnelly:
**Description:**

Another one that can't be parsed by the Vitess query parser.

I'm just about at the point of recommending we turn query parse failures into a benign log-and-ignore, since this has never once been a real issue -- the queries we actually need to handle or forbid are parsed correctly, and these failures are just a bunch of edge cases we don't care about. But that's more work than just adding another prefix string into the ignores regexp, so it can wait until after the production issue is unblocked.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1106)


willdonnelly:
This is a pretty trivial change and CI passes, so I'm going to go ahead and merge without review to unblock a production task that's hitting this.

PR #1110, opened 2023-12-05
Phil/path pointer updates

psFried:
**Description:**

Rolls up a few miscellaneous updates. First and foremost is that it updates the `resource_path_pointers` to match the order that's used by airbyte connectors. This ensures that the resource path that's extracted using `resource_path_pointers` will match the ones returned in the Validated response.

As of right now, the resource path that's extracted using `resource_path_pointers` is only used for a single discover operation, so changing the `resource_path_pointers` is no big deal. But soon, updates to an existing `resource_path_pointers` value will result in an error, so this is just making everything clean and consistent while we still have the chance.

I also tried to make the source-kafka tests slightly more robust.

Finally, I updated the python version requirements for source-criteo and source-shopify to match the version that now comes with Github hosted runners.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1110)


psFried:
> Ah, so the problem is that the order of the resource path pointers was wrong?

Yep, that's right.

PR #1118, opened 2023-12-07
materialize-bigquery: set user agent in bigquery client

psFried:
Sets the User-Agent header in the bigquery client, to enable GCP to track partner attribution.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1118)


williamhbaker:
Is this all we need to set? I had done this a while back and never merged it, which is more complicated but also what I thought we needed: https://github.com/estuary/connectors/pull/1015

psFried:
Let's catch up on VC tomorrow about this

williamhbaker (approved):
LGTM. We can go with this unless we need something fancier.

PR #1120, opened 2023-12-08
Bump `python/pyproject.toml` to include `py.typed`. 

jshearer:
This is so that users of this module can use  Mypy typing. Also bump the version to match the version on pypi: https://pypi.org/project/flow-sdk/

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1120)


jgraettinger (approved):
LGTM

PR #1121, opened 2023-12-11
materialize-databricks: check status of warehouse in prereqs

mdibaiee:
**Description:**

- The best practice for Databricks integrations is to check the status of the warehouse and give a human-readable error if the status is not active. I've updated the code to give such error in PreReqs, and I've reduced the timeout so that we don't have jobs waiting a long time for warehouses. Warehouses take around 5 minutes to start, and we don't want to hang the UI for 5 minutes waiting for the warehouse to come online. We still will ping the warehouse, which will initiate its booting up, but we will let the user know that the warehouse is not started yet, so they have to wait.
- Added `SchemName` parameter name to connection string, this ensures that temporary tables created are in that schema. I tested this by changing the schema in config and running the new connector, the temporary tables are created in the configured schema.
- Updated the integration tests `checks.sh` file to clean uuids from the snapshot to make the test more reproducible.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1121)

### DIFF
--- a/materialize-databricks/config.go
+++ b/materialize-databricks/config.go
@@ -2,34 +2,32 @@ package main
 
 import (
 	"fmt"
-	"strings"
-	"net/url"
-	"github.com/invopop/jsonschema"
 	"github.com/iancoleman/orderedmap"
+	"github.com/invopop/jsonschema"
+	"net/url"
+	"strings"
 )
 
 // config represents the endpoint configuration for sql server.
 type config struct {
-	Address     string           `json:"address" jsonschema:"title=Address,description=Host and port of the SQL warehouse (in the form of host[:port]). Port 443 is used as the default if no specific port is provided." jsonschema_extras:"order=0"`
-	HTTPPath    string           `json:"http_path" jsonschema:"title=HTTP path,description=HTTP path of your SQL warehouse"`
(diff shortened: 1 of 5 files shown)


materialize-databricks/config.go
```diff
@@ -111,15 +109,15 @@ func (c *config) ToURI() string {
 
 	var params = make(url.Values)
 	params.Add("catalog", c.CatalogName)
+	params.Add("schema", c.SchemaName)
```
mdibaiee, line 112: This is the real change, the rest of changes in this file are `go fmt`

tests/materialize/materialize-databricks/checks.sh
```diff
@@ -0,0 +1,2 @@
+# cleanup the snapshot by replacing uuids with a placeholder so that the test is reproducible
+sed -i '' 's/.\{36\}\.json/<uuid>/g' ${SNAPSHOT}
```
williamhbaker, line 2: Would it make sense to include this in the chain here? https://github.com/estuary/connectors/blob/f282c882c66de33204c54ede7950641ea19b3df2/tests/materialize/materialize-databricks/fetch.sh#L7-L9

williamhbaker (approved):
LGTM

PR #1123, opened 2023-12-11
materialize-databricks: UpdateDelay

mdibaiee:
**Description:**

- Use `sql.CommitWithDelay` to delay commits as a cost-saving measure.
- Tested by manually running this, and commits are spaced by 30 minutes. It takes about 5 minutes for Pro and Classic type of SQL warehouses to boot up (Serverless SQL Warehouses are supposed to be faster to boot up, but we don't have access to those to test them), but that does not lead to a timeout in our code, as in my tests the connector seems to work fine using this configuration.
- Added logic so that if auto-stop is configured to be more than 15 minutes, we disable Update Delay, since it is not as useful in that scenario.
- Disabled databricks sql and databricks SDK's internal logging facilities so that we can directly control what errors get logged
- Documentation: https://github.com/estuary/flow/pull/1308/files#diff-79327be85e79183cf88e533070c7a9f8a04dbd230e1d328cd7478d981eae8c2f

This is a sample graph of a Databricks materialization task running with this setting:
image

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1123)

### DIFF
--- a/materialize-databricks/driver.go
+++ b/materialize-databricks/driver.go
@@ -280,6 +280,11 @@ func (c client) PreReqs(ctx context.Context, ep *sql.Endpoint) *sql.PrereqErr {
 			errs.Err(fmt.Errorf("The selected SQL Warehouse is stopping, please start the SQL warehouse and try again."))
 		}
 
+		if res.AutoStopMins >= 15 {
+			log.Warn(fmt.Sprintf("Auto-stop is configured to be %d minutes for this warehouse, disabling update delay. To save costs you can reduce the auto-stop idle configuration and tune the update delay config of this connector. See docs for more information: https://go.estuary.dev/materialize-databricks", res.AutoStopMins))
+
+			cfg.Advanced.UpdateDelay = "0s"
+		}
 	}
 
 	// Use a reasonable timeout for this connection test. It is not uncommon for a misconfigured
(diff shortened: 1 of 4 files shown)


materialize-databricks/driver.go
```diff
@@ -277,6 +279,12 @@ func (c client) PreReqs(ctx context.Context, ep *sql.Endpoint) *sql.PrereqErr {
 		case databricksSql.StateStopping:
 			errs.Err(fmt.Errorf("The selected SQL Warehouse is stopping, please start the SQL warehouse and try again."))
 		}
+
+		if res.AutoStopMins >= 15 {
```
williamhbaker, line 283: Uber-nit: We probably don't need to log this if the materialization is already explicitly configured to have a 0s update delay

materialize-databricks/driver.go
```diff
@@ -277,6 +279,12 @@ func (c client) PreReqs(ctx context.Context, ep *sql.Endpoint) *sql.PrereqErr {
...
+			log.Warn(fmt.Sprintf("Auto-stop is configured to be %d minutes for this warehouse, disabling update delay. To save costs you can reduce the auto-stop idle configuration and tune the update delay config of this connector. See docs for more information: https://go.estuary.dev/materialize-databricks", res.AutoStopMins))
+
+			cfg.Advanced.UpdateDelay = "0s"
```
williamhbaker, line 286: Since we're setting this in `NewTransactor` I don't think we need to do it here too

williamhbaker (approved):
LGTM

PR #1124, opened 2023-12-11
Add detailed docs for how to write OAuth configs

jshearer:
I realized while writing this that there is a _lot_ of possible detail to include, and I may have gone too deep in certain areas and not deep enough in others. But I think this is at least a useful start. Thoughts?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1124)

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -387,6 +387,9 @@ By convention we have picked the field `credentials` to receive injected OAuth v
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jonwihl:
LGTM. Only addition I would make is to indicate that you made need to add the keyword 'credentials' to other parts of the code. Like in situations where the 'authentication' keyword is being replaced by 'credentials'.

jonwihl (approved):

PR #1128, opened 2023-12-13
source-mongodb: enable capturing from shared database instances

psFried:
**Description:**

The current approach to mongodb backfills relies on time-based checks to ensure that we haven't missed any replication events that happen during a backfill. These checks caused problems for captures from shared database servers, like the entry-level instances on MongoDB Atlas.  Such shared servers don't provide unlimited access to the oplog, since the oplog itself is shared by many users. This raises the possibility that querying the oplog can potentially return no events, in the case that it gets filled with events that we aren't authorized to read.

This commit changes the behavior around oplog checks to account for that possibility, with the ultimate goal of being able to robustly capture from the sorts of cheap/free instances that users are likely to test with.

It's no longer necessarily considered a terminal error if we observe no events in the oplog.  We check the oplog before starting the backfill, and persist a boolean that says whether we've observed any events in the oplog. After the backfill is complete, we once again check if there's any events in the oplog. If there were previously events in the oplog, and after the backfill there are none, then we consider that to indicate a possible loss of consistency and start over.

We still check the timestamp of the oldest oplog entry, as before, and ensure that it's older than the start time of the backfill.  The only difference is that we now allow skipping this check in the specific case that the oplog contained zero entries before we started the backfill.

This all required some fairly significant refactoring, and I tried to improve both test coverage and readability along the way.

**Workflow steps:**

- Create a free-tier instance on MongoDB Atlas, but don't add any data
- Setup a capture
- It works!

TODO: still checking to see if there's stuff in the docs that should be updated

**Notes for reviewers:**

Discussed in [this Slack thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1702402565306109)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1128)

### DIFF
--- a/source-mongodb/pull.go
+++ b/source-mongodb/pull.go
@@ -39,6 +39,15 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
 		}
 	}
 
+	var resources = make([]resource, len(open.Capture.Bindings))
+	for idx, binding := range open.Capture.Bindings {
+		var res resource
+		if err := pf.UnmarshalStrict(binding.ResourceConfigJson, &res); err != nil {
+			return fmt.Errorf("parsing resource config: %w", err)
+		}
+		resources[idx] = res
+	}
+
 	var ctx = stream.Context()
 
 	client, err := d.Connect(ctx, cfg)
@@ -53,6 +62,13 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
 		}
 	}()
 
+	// Ensure that we're able to query the oplog, and get the timestamp of the oldest entry
+	// so that we can check for the possibility that we've missed events during downtime.
+	oldestOplogTime, err := checkOplog(ctx, client)
+	if err != nil {
+		return err
+	}
+
 	var c = capture{
 		cfg:    cfg,
 		Output: stream,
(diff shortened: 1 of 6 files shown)


source-mongodb/pull.go
```diff
@@ -134,11 +190,39 @@ type capture struct {
 type captureState struct {
 	Resources map[string]resourceState `json:"resources,omitempty"`
 
+	ChangeStreamEmptyBeforeBackfill *bool `json:"change_stream_empty_before_backfill,omitempty"`
```
mdibaiee, line 193: I would rename this to `OplogEmptyBeforeBackfill` just because change stream is an API on top of oplog, and we check the oplog for being empty, not the change stream. To be very technical, we may have an empty change stream when tailing if there are no new events, but there might still be records in the oplog.

source-mongodb/pull.go
```diff
@@ -372,6 +446,53 @@ func (c *capture) ChangeStream(ctx context.Context, client *mongo.Client, resour
...
+	if backfillStartTime.IsZero() {
+		// We're starting a new backfill, so emit a state update indicating
+		// whether we've observed events in the oplog.
+		var isEmpty = oldestOplogEntry.IsZero()
+		state.ChangeStreamEmptyBeforeBackfill = &isEmpty
+		return true, ""
+	} else if state.wasChangeStreamEmptyBeforeBackfill() && !oldestOplogEntry.IsZero() {
+		// We're resuming an in-progress backfill and there is now an event
+		// that's been observed in the change stream. Update the state accordingly
+		var superHighTechFalse = false
```
mdibaiee, line 468: 😂

mdibaiee (commented):
Thanks Phil, a few comments

PR #1130, opened 2023-12-15
sqlcapture: DEBUG-log the results of startup table discovery

willdonnelly:
**Description:**

Sometimes it would be pretty convenient if we could turn on debug logs in production and see the full list of all tables discovered and their columns and the types of those columns.

Since by default we're only running at INFO level in production this should be fine, although it will make test output noisier.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1130)


williamhbaker (approved):
LGTM

PR #1134, opened 2023-12-18
Update Singer Python connector shim 

jshearer:
Expose the proper config flag to switch between incremental and polling mode.

Also add simple validation to make sure that the user doesn't pick a disallowed mode based on what the connector advertises

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1134)


python/flow_sdk/shim_singer_sdk.py
```diff
@@ -240,12 +251,7 @@ def sync_streams(self, delegate: singer_sdk.Tap):
     "type": "object",
     "properties": {
         "stream": {"type": "string"},
-        "syncMode": {"type": "string", "enum": ["incremental", "full_refresh"]},
-        "namespace": {"type": "string"},
-        "cursorField": {
-            "type": "array",
-            "items": {"type": "string"},
-        },
+        "replication_mode": {"type": "string", "enum": ["INCREMENTAL", "FULL_TABLE"]},
```
jgraettinger, line 254: i see `replication_method` above, but `replication_mode` here. Also, should `replication_key` be schematized ?

jshearer: Wow, that's a typo. Both done.

jgraettinger (approved):
LGTM

PR #1135, opened 2023-12-18
materializations: validate and apply based on existing tables

williamhbaker:
**Description:**

This changes many materializations to use the actual structure of materialized resources in how they do `Validate` and `Apply.` The materializations that care about this sort of thing are those that create "typed" columns and produce constraints off of those columns.

There are a couple of immediate motivations:
- It allows these materializations to work better with pre-created tables, where the pre-created table might be incompatible in a way that we can fix: Either missing columns, or has non-null columns that we don't have selected fields for and/or have nullable fields selected for. It is actually surprising how often this is a problem, usually because a new materialization is created to materialize to a previously materialized-to table and the source collection schema has evolved.
- If there is an unsatisfiable field with a desired source collection schema change, it will now be possible for the user to update their table and then have the publication succeed. For example, a column that was previously only integers now has random strings - in this case, the user could run a query on their table to change the column type to a string and do whatever casting is necessary, and a subsequent Validate of the connector would succeed based on that changed table state.

A lot of the work here is elevating the `ExistingColumns` concept implemented for SQL materializations in https://github.com/estuary/connectors/pull/1086 into the `materialize-boilerplate`, now called `InfoSchema`. There is also quite a bit of re-plumbing and abstracting the common actions done during Apply into a common `Applier` interface and `ApplyChanges` routine that will efficiently carry out the needed actions for implementors of `Applier.`

As a practical callout, there will be additional overhead in the code for materializations that produce constraints: They now need to know how to interpret their materialized resources' "columns" (or equivalent concept) and map that back to a way to produce constraints vs. a proposed Flow collection projection. This is pretty easy for a system like ElasticSearch where the text for an index mapping you create matches up exactly with the text of a described index mapping, but this is not generally the case and SQL systems must describe this reverse mapping explicitly, along with the "forward" mapping of Flow types to column types that already exists. I've thought about how to make expressing these relationships less redundant (ref: the typical `TypeMapper` construction when building a `Dialect` for a SQL materialization) but couldn't come up with anything that wouldn't require some pretty massive changes, and I didn't want to add that to this already quite large PR. So I'm leaving that as a possible future improvement for now.

**Workflow steps:**

Create materializations to existing tables and they will be modified as needed if possible to be compatible with applied changes. Validations may fail if existing columns are incompatible with the proposed spec.

**Notes for reviewers:**

The first 2 commits are entirely refactoring with no functional changes, and represent of decent chunk of the changed code for this PR. I tried to structure the remaining commits in logical chunks of work, with the last several having some "finishing touches" type of stuff that revises earlier work. So it may be easiest to review commit-by-commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1135)


materialize-sql/std_sql.go
```diff
@@ -61,27 +62,6 @@ func StdSQLExecStatements(ctx context.Context, db *sql.DB, statements []string)
 	return conn.Close() // Release to pool.
 }
 
-// StdIncrementFence increments the value for the fence for all checkpoints of the provided
-// materialization.
-func StdIncrementFence(ctx context.Context, db *sql.DB, ep *Endpoint, materialization string) error {
```
williamhbaker, line 66: Minor callout here: We won't need this anymore with runtime-driven applies since all materialization shards will be stopped when the apply happens, so there's no reason to fence off other instances.

materialize-sql/std_sql.go
```diff
@@ -292,38 +272,6 @@ func (col anyColumn) String() string {
 	return string(col)
 }
 
-// StdListTables is a convenience function for getting a list of tables for Client implementations
-// which use Go's standard `sql.DB` type and systems with an information_schema schema.
-func StdListTables(ctx context.Context, db *sql.DB, catalog string, schema string) ([]string, error) {
```
williamhbaker, line 297: I removed the tests that used this function; they were the ones that added more than one table at a time. These tests made the overall tests much more complicated and its trivial to test this scenario manually so it didn't seem worthwhile to keep.

materialize-boilerplate/info_schema.go
```diff
@@ -0,0 +1,211 @@
...
+	// Sort the output to make it a little more clear which paths are ambiguous if there are a lot
+	// of them.
+	slices.SortFunc(out, func(a, b []string) int {
+		return strings.Compare(joinPath(i.locatePath(a)), joinPath(i.locatePath(b)))
+	})
+
+	return out
+}
+
+func joinPath(in []string) string {
```
mdibaiee, line 202: Why not just use https://pkg.go.dev/strings#Join?

mdibaiee (commented):
Thanks a lot Will, this is large, I have a few questions / suggestions. I haven't reviewed the implementation for every connector in detail yet, my questions are mostly about `materialize-boilerplate`, but I'm looking through connectors now

materialize-boilerplate/apply_test.go
```diff
@@ -0,0 +1,301 @@
...
+			require.NoError(t, err)
+			require.Equal(t, tt.want, app.getResults())
+			actions := got.ActionDescription
+
+			// Dry run, not concurrent.
+			req.DryRun = true
+			got, err = ApplyChanges(ctx, req, app, is, false)
+			require.NoError(t, err)
+			require.Equal(t, testResults{}, app.getResults())
+			require.Equal(t, actions, got.ActionDescription)
```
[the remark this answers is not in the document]
williamhbaker: `app.getResults()` clears the mock app's "memory" of changes, so each of these little blocks is a fresh run. It is saying that if you do a dry run vs. not a dry run starting from scratch, you should get the same actions output, but the dry run won't actually do anything via the `testResults{}`. These are pretty low level unit tests I put in to make sure that dry run works as it should, and that concurrent vs. not current effectively end up doing the same things.

materialize-boilerplate/info_schema.go
williamhbaker: I may have had some grander plan for this at some earlier iteration, but yeah that's all we'd need now, and I'll swap that in!

materialize-boilerplate/apply.go
```diff
@@ -0,0 +1,232 @@
...
+type BindingUpdate struct {
+	// NewProjections is the projections of a proposed materialization spec that are included in the
+	// field selection but there are no fields in the materialized resource for.
+	NewProjections []pf.Projection
+
+	// NewlyNullableFields is the fields in the endpoint that should now be made nullable, that
+	// weren't before. They may need to be nullable because a materialized field is now nullable and
+	// wasn't before, or because they exist as non-nullable in the destination and aren't included
+	// in the materialization's field selection.
+	NewlyNullableFields []EndpointField
```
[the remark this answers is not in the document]
williamhbaker: I did it this way so we can remove nullability constraints for fields that we don't have projections for, for example if a column exists in a table that is non-nullable but we aren't materializing a field for it when the `Apply` comes through. In that case we don't have a projection available. I've seen this happen a couple of times when users have re-materialized to tables that were previously created by materializations that had an optional field selected that was always present in the schema, but on the re-materialization they didn't include that optional field.

It certainly does make cases where we have to re-state the column definition more tricky though - although another benefit is that a user could have a pre-created column with certain characteristics (like a different encoding than we'd use for a new projection) and there'd be no risk of us blowing that away if a `NOT NULL` is dropped.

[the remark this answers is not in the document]
williamhbaker: Adding to that...I thought about having both the `EndpointField` _and_ a `pf.Projection` if it was available, but then decided that since it has to work with just the `EndpointField` anyway we might as well go with that all the time.

materialize-boilerplate/apply.go
```diff
@@ -0,0 +1,229 @@
...
+	// carried out with a "CREATE OR REPLACE ..." type of action.
+	ReplaceResource(ctx context.Context, spec *pf.MaterializationSpec, bindingIndex int) (string, ActionApplyFn, error)
+
+	// UpdateResource updates an existing resource. The `BindingUpdate` contains specific
+	// information about what is changing for the resource. `NewProjections` are assured to not
+	// already exist in the destination, and `NewlyNullableFields` are assured to non-nullable in
+	// the destination. It's called for every binding, although it may not have any `BindingUpdate`
+	// parameters. This is to allow materializations to perform additional specific actions on
+	// binding changes that are not covered by the general cases of the `BindingUpdate` parameters.
+	UpdateResource(ctx context.Context, spec *pf.MaterializationSpec, bindingIndex int, applyParams BindingUpdate) (string, ActionApplyFn, error)
```
jgraettinger, line 70: nit: applyParams => bindingUpdate ?

Should it be pluralized, along with the struct itself (there can be _many_ updates for the binding)?

materialize-boilerplate/apply.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		if !is.HasResource(binding.ResourcePath) {
+			// Resource does not yet exist, and must be created.
+			desc, action, err := applier.CreateResource(ctx, req.Materialization, bindingIdx)
+			if err != nil {
+				return nil, fmt.Errorf("getting CreateResource action: %w", err)
+			}
+			addAction(desc, action)
+		} else if existingBinding != nil && existingBinding.Backfill != binding.Backfill {
```
jgraettinger, line 124: What about a DropResource action of the Applier interface, and remove ReplaceResource ?

That would make it an easier change, if we decide to tweak the circumstances under which we drop tables.

materialize-boilerplate/apply.go
```diff
@@ -0,0 +1,229 @@
...
+				return nil, fmt.Errorf("getting ReplaceResource action: %w", err)
+			}
+			addAction(desc, action)
+		} else {
+			// Resource does exist and may need updated for changes in the binding specification.
+			params := BindingUpdate{
+				NewlyDeltaUpdates: existingBinding != nil && !existingBinding.DeltaUpdates && binding.DeltaUpdates,
+			}
+
+			for _, proposedField := range binding.FieldSelection.AllFields() {
```
jgraettinger, line 138: nit: you could shorten up the line length by proposedField => field and proposedProjection => projection without loss of clarify IMO.

materialize-sql/endpoint.go
```diff
@@ -92,6 +108,9 @@ type Endpoint struct {
 	NewTransactor func(ctx context.Context, _ *Endpoint, _ Fence, bindings []Table, open pm.Request_Open) (pm.Transactor, error)
 	// Tenant owning this task, as determined from the task name.
 	Tenant string
+	// ConcurrentApply of Apply actions, for system that may benefit from a scatter/gather strategy
+	// for changing many tables in a single publication.
```
jgraettinger, line 112: nit: tables are no longer changed in a publication.

materialize-boilerplate/validate.go
```diff
@@ -207,7 +207,22 @@ func (v Validator) validateMatchesExistingBinding(
 
 		// Build constraints for new projections of the binding.
 		if _, ok := constraints[p.Field]; !ok {
-			constraints[p.Field] = v.c.NewConstraints(&p, deltaUpdates)
+			c := v.c.NewConstraints(&p, deltaUpdates)
+
+			// Continue to recommended any optional fields that were included in a prior spec's
```
jgraettinger, line 212: You say:

> But, it will help to preserve continuity for "legacy" specs if
the definition of recommended fields changes for a materialization and there is a change in the
materialized resource. 

I'm a little tired, but: is this check worth the code? It seems like it could be trivially omitted and :shrug: but it may have come about due to an acute problem you saw.

jgraettinger (approved):
Phew, there's a lot here. Left many comments below, but no deal breakers.

This is a pretty important change, and the _degree_ of change makes it important to get in soon, so feel free to push back or address QoL stuff as future PR's.

materialize-boilerplate/apply.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+// ActionApplyFn is a callback that will be executed to carry out some action to achieve a change to
+// a materialized resource.
+type ActionApplyFn func(context.Context) error
+
+// BindingUpdate is a distilled representation of the typical kinds of changes a destination system
+// will care about in response to a new binding or change to an existing binding.
+type BindingUpdate struct {
+	// NewProjections is the projections of a proposed materialization spec that are included in the
```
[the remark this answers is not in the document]
williamhbaker: :+1: on the grammar fix

It doesn't take an EndpointField because there is not a materialized "column" in the destination since its a new projection, hence there is no existing field information to create an EndpointField from. Indeed, the Projection must be used to create a new column in the destination system. I'll extend the comment a bit to describe that.

materialize-boilerplate/info_schema.go
```diff
@@ -0,0 +1,156 @@
...
+type LocatePathFn func([]string) []string
+
+// TranslateFieldFn takes a Flow field name and outputs a translated string that can be used to find
+// the field in the InfoSchema, with translations applied in a similar way as LocatePathFn.
+type TranslateFieldFn func(string) string
+
+// InfoSchema contains the information about materialized collections and fields that exist within
+// the endpoint.
+type InfoSchema struct {
+	// resources is a mapping of string keys representing locations within the destination system to
```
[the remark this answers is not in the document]
williamhbaker: The keys are coming from the materialized fields that exist in the endpoint. They will end up being the same as the output of the LocatePathFn. But, they are populated from for ex. querying `INFORMATION_SCHEMA` and pushing the results in the `InfoSchema` struct. I'm updating the comment to be this, which will hopefully help:

```go
// resources is a mapping of string keys representing locations of materialized fields within
// the destination system to EndpointFields with their metadata. It is populated by calls to
// (*InfoSchema).PushField, which adds materialized fields to InfoSchema as reported by the
// endpoint. As a convention, if the location has multiple components (like a schema & table
// name), these components are joined with dots.
```

materialize-boilerplate/apply.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// NewlyDeltaUpdates is if the materialized was standard updates per the previously applied
+	// materialization spec, and is now not. Some systems may need to do things like drop primary
+	// key restraints in response to this change.
+	NewlyDeltaUpdates bool
+}
+
+// Applier represents the capabilities needed for an endpoint to apply changes to materialized
+// resources based on binding changes. Many of these functions should return an ActionApplyFn, which
+// will be executed only if the `Apply` is not a dry-run.
```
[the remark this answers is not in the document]
williamhbaker: Yeah, we'll need the callback stuff for concurrent executions, and being able to build up the action description in a deterministic way, which seems useful. It does complicate the implementations a little bit. Depending on when this gets merged with respect to updating the flow pin to when dry-run doesn't exist, I'll remove the dry-run stuff here, which will be kind of nice.

materialize-boilerplate/apply.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// PutSpec upserts a spec into the metadata table.
+	PutSpec(ctx context.Context, spec *pf.MaterializationSpec, version string, exists bool) (string, ActionApplyFn, error)
+
+	// CreateResource creates a new resource in the endpoint. It is called only if the resource does
+	// not already exist.
+	CreateResource(ctx context.Context, spec *pf.MaterializationSpec, bindingIndex int) (string, ActionApplyFn, error)
+
+	// ReplaceResource replaces a resource in the endpoint so that it can start materializing from
+	// the beginning. It may be called even if the resource doesn't already exist. This is usually
```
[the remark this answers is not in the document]
williamhbaker: Yeah, I guess it wouldn't actually be called if the materialized resource doesn't already exist. The only way ReplaceResource gets called is if the materialized resource already exists and the backfill counter for the new spec is larger than the one in the prior spec (and it exists in the prior spec). I think this comment was a holdover from a prior time. I'll update it.

materialize-boilerplate/apply.go
williamhbaker: I'll change `applyParams` to `bindingUpdate` since `applyParams` was a holdover from the prior SQL-only version of this. `bindingUpdate` as singular is still making more sense to me though.

materialize-boilerplate/apply.go
williamhbaker: I think you're alluding to make a table replacement always be a drop followed by a create, which is an interesting idea since that's what some systems have to do anyway.

We've already got the implementations for all materialization done and working with a distinct "replace" action, which will sometimes do an actual "replace" if the destination supports it, sometimes do a drop & create within a transaction, and failing all of that does a non-transactional drop & replace. So I'm not too inclined to rework all of those right now, and maybe this could be a future improvement if we need to tweak dropping tables.

materialize-sql/endpoint.go
williamhbaker: :+1: Kind of makes me wonder if supporting this concurrent pattern is even worth it anymore. But I guess its still nice for them to go fast if possible.

materialize-boilerplate/validate.go
williamhbaker: The described scenario in the comment is a bit of a stretch. A stronger motivation is for systems like dynamodb where we allow field selection, but the destination system has only a limited concept of a "column" (only the keys for dynamodb). In that case I think its useful to recommend fields that were previously selected, even if their "as new" constraints are optional, based on the field selection from the prior spec.

materialize-boilerplate/info_schema.go
```diff
@@ -0,0 +1,156 @@
...
+func (i *InfoSchema) HasResource(resourcePath []string) bool {
+	rk := joinPath(i.locatePath(resourcePath))
+	_, ok := i.resources[rk]
+	return ok
+}
+
+// ExtractProjection finds the projection in a Flow collection spec that must correspond to a field
+// reported by the endpoint. Fields in the collection spec must have the same one-way translation
+// applied to them that a field does when materializing it in order to do this lookup.
+func (i *InfoSchema) ExtractProjection(efn string, collection pf.CollectionSpec) (pf.Projection, bool, error) {
```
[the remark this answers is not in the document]
williamhbaker: `efn` is a little vague. As you've noted the challenge is a Flow field name vs. a translated endpoint field name. The functions where `efn` is used are only used internally to the boilerplate package, so they don't really need to be part of its "public" API. I'll change the name of this to `endpointFieldName` to make it more obvious, and make these functions non-exported for now.

materialize-boilerplate/apply.go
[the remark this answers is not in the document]
jgraettinger: Right, I suppose I was thinking "why can't the caller of Applier do more of the heavy-lifting for dialect mapping", but only later realized "because dialects are a SQL-ism only".

Then I wondered whether there was any helpful middle-ground between "here's a Projection, figure it out" vs the nice unpacking work we do for SQL, but nothing obvious came to me.

materialize-boilerplate/apply.go
jgraettinger: :+1: to doing this in the future, but I continue to think separate Drop and Create callbacks are more appropriate. The current Replace feels like an over-generalization of the CREATE OR REPLACE sql pattern... which we just saw today (with bigquery clustering keys) doesn't always work either.

materialize-boilerplate/validate.go
jgraettinger: Oh! Okay, that's a _great_ reason for doing this, but it didn't come through for me at all. Would you please update the comment to give that example?

PR #1137, opened 2023-12-19
source-snowflake: Initial implementation

willdonnelly:
**Description:**

This PR adds a Snowflake CDC connector.

I'll come back and flesh out the PR description in the morning and make sure it's got docs and CI setup and whatnot, but I believe the core connector implementation is now complete so I might as well get this uploaded and start the code review ball rolling.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1137)


source-snowflake/capture.go
```diff
@@ -0,0 +1,650 @@
...
+
+	SequenceNumber int `json:"seq" jsonschema:"description=The sequence number of the staging table from which this document was read"`
+	TableOffset    int `json:"off" jsonschema:"description=The offset within that staging table at which this document occurred"`
+}
+
+func (snowflakeDriver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) error {
+	log.Debug("connector started")
+
+	var checkpoint = new(captureState)
+	if len(open.StateJson) > 0 {
```
williamhbaker, line 43: Might need to check for a state value of `{}` here or otherwise account for that possibility to avoid ending up with a `nil` `checkpoint.Streams` if there is no previously persisted state.

source-snowflake/capture.go
```diff
@@ -0,0 +1,650 @@
...
+	fmt.Fprintf(queryBuf, "SELECT * FROM %s", stagingName.QuotedName())
+	if len(quotedOrderColumns) > 0 {
+		fmt.Fprintf(queryBuf, " ORDER BY %s", strings.Join(quotedOrderColumns, ", "))
+	}
+	fmt.Fprintf(queryBuf, " LIMIT NULL OFFSET %d", streamState.TableOffset)
+	var readStagingTableQuery = queryBuf.String()
+
+	// Read contents of staging table and output to Flow.
+	logEntry.WithField("query", readStagingTableQuery).Info("querying staging table")
+	rows, err := c.DB.QueryContext(ctx, readStagingTableQuery)
```
williamhbaker, line 418: Definitely test out the streaming JSON downloader option, which we implemented for `materialize-snowflake` load queries, to prevent extremely high memory usage & OOM kills: https://github.com/estuary/connectors/pull/745. FYI I pulled this branch down and threw together a docker file to get a capture working, and am seeing memory usage very near 1 GB for the connector when capturing a modestly sized table.

Also, the driver has some weirdness around capturing numbers, see https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Supported_Data_Types and the `WithHigherPrecision` option. It might not do this if using the streaming JSON downloader, but as it is I noticed the capture was capturing integers as strings.

source-snowflake/discovery.go
```diff
@@ -0,0 +1,383 @@
...
+	// as primary keys. But there's no simple way to select just the columns we need
+	// from these nonstandard not-a-table queries, and that makes them unstable if
+	// new columns were to be added in the future.
+	//
+	// So we wrap the database connection with SQLX and use its struct-tag reflection
+	// to select the columns of interest *by name*, and set Unsafe() so that it will
+	// ignore unmatched columns.
+	var xdb = sqlx.NewDb(db, "snowflake").Unsafe()
+	var tables []*snowflakeDiscoveryTable
+	if err := xdb.Select(&tables, "SHOW TABLES;"); err != nil {
```
williamhbaker, line 94: This seems to work, but can you not get what is needed from `information_schema.tables`?

source-snowflake/discovery.go
```diff
@@ -0,0 +1,383 @@
...
+	Table      string `db:"table_name"`
+	ColumnName string `db:"column_name"`
+	KeySeq     int    `db:"key_sequence"`
+}
+
+func performSnowflakeDiscovery(ctx context.Context, cfg *config, db *sql.DB) (*snowflakeDiscoveryResults, error) {
+	log.Debug("performing discovery")
+
+	// We have to use `SHOW FOO` queries to get some of the information we need, such
+	// as primary keys. But there's no simple way to select just the columns we need
```
williamhbaker, line 85: There's [RESULT_SCAN](https://docs.snowflake.com/en/sql-reference/functions/result_scan), although maybe that is what you mean by not being a simple way, since it is a bit clunky.

source-snowflake/discovery.go
```diff
@@ -0,0 +1,383 @@
...
+			continue
+		}
+		streams[tableID].PrimaryKey = append(streams[tableID].PrimaryKey, discoveredPK.ColumnName)
+		if len(streams[tableID].PrimaryKey) != discoveredPK.KeySeq {
+			log.WithFields(log.Fields{
+				"stream": tableID.String(),
+				"pkey":   streams[tableID].PrimaryKey,
+				"seq":    discoveredPK.KeySeq,
+			}).Warn("primary key sequence mismatch")
+			return nil, fmt.Errorf("internal error: primary key sequence mismatch")
```
williamhbaker, line 172: I'm not entirely sure why, but I hit this error when trying to discover a checkpoints table that our snowflake materialization makes, `FLOW_CHECKPOINTS_V1`. It's primary keys don't seem to show up in order when I use a `SHOW PRIMARY KEYS`: 
image

source-snowflake/discovery.go
```diff
@@ -0,0 +1,383 @@
...
+		out.Extras["type"] = []string{s.jsonType, "null"} // Use variadic form.
+	} else {
+		out.Type = s.jsonType
+	}
+	return out
+}
+
+var snowflakeTypeToJSON = map[string]columnSchema{
+	// "NUMBER":  {jsonType: "number"}, // The 'NUMBER' column type is handled in code
+	"TEXT":          {jsonType: "string"},
```
williamhbaker, line 374: I think we'll want to handle `VARIANT` types also. Maybe the "any possible type" schema is best for them still, but I think they are fairly common and not something we'd want to show an error for - which if I'm understanding things correctly, we currently would.

source-snowflake/capture.go
```diff
@@ -0,0 +1,650 @@
...
+
+type snowflakeSourceMetadata struct {
+	sqlcapture.SourceCommon
+
+	SequenceNumber int `json:"seq" jsonschema:"description=The sequence number of the staging table from which this document was read"`
+	TableOffset    int `json:"off" jsonschema:"description=The offset within that staging table at which this document occurred"`
+}
+
+func (snowflakeDriver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) error {
+	log.Debug("connector started")
```
williamhbaker, line 40: Needs a `stream.Ready(false)` somewhere

williamhbaker (commented):
I took a quick look and made some comments on this draft - you're probably already aware of these things, but here it is anyway just in case.

source-snowflake/discovery.go
```diff
@@ -0,0 +1,383 @@
...
+
+	var db, err = connectSnowflake(ctx, cfg)
+	if err != nil {
+		return nil, fmt.Errorf("error connecting to snowflake: %w", err)
+	}
+	defer db.Close()
+
+	discoveryResults, err := performSnowflakeDiscovery(ctx, cfg, db)
+	if err != nil {
+		log.WithField("err", err).Fatal("error discovering snowflake collections")
```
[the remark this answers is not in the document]
willdonnelly: I have no idea why these errors are being logged like this and not just returned.

(I'd hazard a guess that I may have started writing the initial version of this discovery logic in some prototyping context where this was just `func main()` or something and simply never fixed this when turning it into an actual capture connector)

Fixed.

source-snowflake/capture.go
willdonnelly: Good catch, I've changed the logic to be `if checkpoint.Streams == nil { checkpoint.Streams = make(...) }` which should ensure that it's always initialized correctly.

source-snowflake/discovery.go
willdonnelly: Oh dang, I don't think I actually tested multiple-column primary keys very much. This is an obvious result of using `SHOW PRIMARY KEYS` as the enumeration query, there's no actual `ORDER BY key_sequence` occurring at any point.

I'll have to go re-investigate whether there's any way to get the necessary info from `information_schema` or if we have to just cope with `SHOW PRIMARY KEYS` returning unordered results and sort them ourselves...

willdonnelly: Yep, I've just double-checked. The `information_schema.table_constraints` view has existed for at least five years and for that entire time people keep asking and being told that there is still no way to actually get the _column names_ corresponding to a table's primary-key constraint other than by running `SHOW PRIMARY KEYS`.

I'll have to modify the logic to handle sorting the results of that query within the connector.

source-snowflake/discovery.go
willdonnelly: Yeah, that could probably be made to work but it seemed simpler to just use the `sqlx` package to automatically map result column names to struct fields, and I'd rather not change that now unless there's a compelling reason.

(The `sqlx` package was already an indirect dependency brought in by _something_ in the connectors repository, so trying to avoid another package dependency doesn't seem like enough reason IMO)

source-snowflake/discovery.go
willdonnelly: In this case yeah, I think all the information is present in both `SHOW TABLES` and `information_schema.tables`.

The real sticking point is `SHOW PRIMARY KEYS` which cannot be replaced with an `information_schema` query at all. IIRC, because of that I originally _only_ used the `SHOW FOO` style queries here for consistency, but then it turned out that `SHOW COLUMNS` didn't give us all of the information we need while `information_schema.columns` did, so now we're forced to have an inconsistent mix of data sources.

So this could in principle be swapped out for `SELECT * FROM information_schema.tables` with a little bit of name-changing in the corresponding struct tags, but I'm not convinced either option is strictly better so I'd rather just leave it as-is.

source-snowflake/capture.go
willdonnelly: Done.

source-snowflake/capture.go
willdonnelly: I've modified the capture to use streaming JSON here, and added proper value translation plumbing to parse stringified numeric and boolean columns back into the appropriate JSON types.

source-snowflake/discovery.go
willdonnelly: Variant types (including object/array) are rather tricky to handle. The values we get back from the gosnowflake client library are stringified and appear to be JSON, so it's tempting to translate them all to `json.RawMessage` values so they get embedded directly into the captured documents. However I can't find any actual documentation which would justify doing this. I'll have to investigate this one some more tomorrow morning.

willdonnelly: The connector now has support for the `VARIANT` / `OBJECT` / `ARRAY` Snowflake column types. Discovery will produce schemas of `{}` / `{type: object}` / `{type: array}` respectively and captures produce the obvious JSON representations.

source-snowflake/discovery.go
willdonnelly: The necessary sorting has been added explicitly in the code (see https://github.com/estuary/connectors/pull/1137/files#diff-3a385a923b61bcbfd54148783038ddf3a99bc5ed2235a25246767b42425b9587R117)

willdonnelly:
PTAL. The main changes since the previous round of PR feedback are:

- The per-binding state in checkpoints is now keyed using the runtime-provided `StateKey` property.
- Using `StateKey` meant that the keys can no longer be reliably decomposed into the source table schema/name, which was previously something we used as part of the cleanup logic when bindings were deleted. To remedy this, the unique ID associated with the staging resources of a particular source table is now stored in the per-binding state as well.
- And since we're explicitly keeping track of that unique ID now, there is no need to implicitly compute it based on the source table name. In fact it's better to have it incorporate a random nonce, because this doesn't have the same issues with potential cleanup/initialization conflicts when incrementing the backfill counter. So we do that now.
- Also instead of simply deleting the state entries for removed bindings, we now set a boolean to mark them as disabled. This eliminates some theoretical edge cases where staging resources might fail to ever be cleaned up if the first and only attempt to delete them failed, at the cost of slightly larger state checkpoints. If that ever becomes an issue, we can reintroduce some logic to prune disabled entries after verifying that all associated Snowflake entities have been deleted.

source-snowflake/discovery.go
```diff
@@ -0,0 +1,400 @@
...
+								},
+								"reduce": map[string]interface{}{
+									"strategy": "merge",
+								},
+							},
+							Required: []string{"op", "source"},
+						},
+					},
+					"reduce": map[string]interface{}{
+						"strategy": "merge",
```
williamhbaker, line 311: We've had issues with top-level merge reductions in some captures before, see https://github.com/estuary/connectors/pull/1144

That's mainly an issue with captures that do not have a firm schema, but I am wondering what happens with this capture if tables are altered to have columns removed, and values are not present at all for queries on the later staging tables resulting from those?

williamhbaker (approved):
LGTM - one question about the `reduce: merge` discovery output

source-snowflake/main.go
```diff
@@ -0,0 +1,146 @@
...
+
+func main() {
+	boilerplate.RunMain(new(snowflakeDriver))
+}
+
+type snowflakeDriver struct{}
+
+// config represents the endpoint configuration for Snowflake.
+type config struct {
+	Host      string         `json:"host" jsonschema:"title=Host URL,description=The Snowflake Host used for the connection. Must include the account identifier and end in .snowflakecomputing.com. Example: orgname-accountname.snowflakecomputing.com (do not include the protocol)." jsonschema_extras:"order=0"`
```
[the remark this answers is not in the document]
willdonnelly: Done

source-snowflake/discovery.go
willdonnelly: Hmm, yeah, in that case the merged documents would end up with additional properties which are no longer in the source table. As I understand it, the reason we currently have `merge` reduction semantics on the top level of SQL CDC documents is because we need the metadata properties to be recursively merged so that `/_meta/before` will have `firstWriteWins` semantics.

It could be argued that we should revisit this at some point, but for now I think it's probably sufficient to note that this behavior for the new Snowflake capture is consistent with our other SQL CDC connectors (and in fact I think I basically just copy-pasted it from those).

williamhbaker: If I'm understanding correctly then, other CDC captures would do the same thing if they had a column completely removed via an alter table statement, with reduced materialized rows going forward carrying on now absent values via the merge reduction. And that's a bit of a different problem than for example dynamodb where a field may be completely absent to represent the lack of a value, where it has no concept of columns. I'd agree that it makes sense to proceed then as it is.

PR #1143, opened 2023-12-19
source-kafka: fix config field names

mdibaiee:
**Description:**

- The JSONSchema and the configuration are not linked by code, and in the JSONSchema the credential keys are `aws_access_key_id` and `aws_secret_access_key`, however in the code the struct has `access_key_id` and `secret_access_key`. I had not hit this issue in my tests since I used a local, manually prepared configuration file, but I noticed the discrepancy when trying this out on the UI
- Also removed `PLAIN` as the default `mechanism`, since when you first open the jsonform, it defaults to `PLAIN`, and if you switch to AWS IAM, the default value stays in somehow and becomes part of the config, which is confusing because the user will see it in their config, but in reality it has no effect when using IAM.
- Exclude `__amazon_msk_canary` from discovered bindings since it seems to cause an endless loop of this error:
```
SASL authentication error: SaslAuthenticateRequest failed
```
Although I am not entirely sure if this is specific to this stream or might happen on other streams. I have one stream of data that seems to be doing okay with (I tried increasing the load on this stream and it seems to be fine). What's interesting is in case of `__amazon_msk_canary`, we are able to capture a lot of events, and then the connector fails. I'm assuming there are authentication limitations for accessing this internal topic.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1143)

### DIFF
--- a/source-kafka/src/kafka.rs
+++ b/source-kafka/src/kafka.rs
@@ -312,7 +312,7 @@ fn parse_message<'m>(msg: &'m BorrowedMessage<'m>) -> Result<serde_json::Value, 
     })
 }
 
-static KAFKA_INTERNAL_TOPICS: [&str; 1] = ["__consumer_offsets"];
+static KAFKA_INTERNAL_TOPICS: [&str; 2] = ["__consumer_offsets", "__amazon_msk_canary"];
 
 fn reject_internal_topics(topic: &&MetadataTopic) -> bool {
     !KAFKA_INTERNAL_TOPICS.contains(&topic.name())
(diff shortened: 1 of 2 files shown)


source-kafka/src/kafka.rs
```diff
@@ -312,7 +312,7 @@ fn parse_message<'m>(msg: &'m BorrowedMessage<'m>) -> Result<serde_json::Value,
     })
 }
 
-static KAFKA_INTERNAL_TOPICS: [&str; 1] = ["__consumer_offsets"];
+static KAFKA_INTERNAL_TOPICS: [&str; 2] = ["__consumer_offsets", "__amazon_msk_canary"];
```
psFried, line 315: just for future reference, you can use a slice (`&[&str]`), which doesn't require an explicit length

psFried (approved):
LGTM

Issue #1146, opened 2023-12-20
technical connectors: respect per-binding state key


williamhbaker:
- [x] `source-alpaca`
- [x] `source-dynamodb`
- [x] `source-firestore`
- [x] `source-hello-word`
- [x] `source-kinesis`
- [x] `source-mongodb`
- [x] `source-bigquery-batch`
- [x] `source-mysql-batch`
- [x] `source-postgres-batch`
- [x] SQL Captures: https://github.com/estuary/connectors/pull/1101
  - `source-mysql`
  - `source-postgres`
  - `source-sqlserver`
- [x] Filesource captures:
  - `source-gcs`
  - `source-google-drive`
  - `source-http-file`
  - `source-s3`
  - `source-sftp`
 - [x] `materialize-databricks`

Evaluated with no updates needed:
- [x] `source-http-ingest` - Doesn't persist any state.
- [x] `source-salesforce-next` - Since the push topics the connector uses can only go back 24 hours anyway, there doesn't seem to be a strong motivation to support `backfill` causing these reset explicitly. A re-backfill would need to be done with the historical connector.
- [x] `materialize-s3-parquet` - Incrementing the `backfill` will reset the checkpoint for a binding and it will start processing data from the beginning, while continuing to produce files with increasingly incremented file names from where it left off previously since [these counts are checkpointed](https://github.com/estuary/connectors/blob/cb7dee3ca34d0c91c4b1473e0033dfc257e88e68/materialize-s3-parquet/checkpoint/driver_checkpoint.go#L16). I don't know that it would be an obvious improvement to do anything differently than that. Also I am noting that this current strategy seems incompatible with ever removing bindings from the spec (they would shift down in position and assume a different position in `NextSeqNumList`), so it would be worth a deeper look & modifications if we want to make any changes.
 - [x] `materialize-firebolt` - Minimal usage, still "works" without state keys although there may be problems with specific scenarios related to persisted queries. Can revisit if the connector gets more usage.
 - [x] `source-kafka` - Moved to a separate issue: https://github.com/estuary/connectors/issues/1211

PR #1147, opened 2023-12-20
sources: first round of capture state keys

williamhbaker:
**Description:**

Updates a few captures to use the new runtime-provided `stateKey`s for tracking their bindings states. This will allow a single publication of a `backfill` counter increment to reset the state for these bindings from the beginning.

These first few connectors are some of the lesser used and generally least complicated ones. `source-dynamodb` is in here as a PoC for a more complicated connector, and that update is mostly just replacing `table` with `stateKey`.

Of note is that they all store binding state in this form:
```json
{
  "bindingStateV1": {
    "aResource": "someState"
  }
}
```

We don't have immediate plans to implement runtime pruning of unused state keys, but using the consistent top-level object of `bindingStateV1` is intended to allow that to be possible in the future if desired.

See issue https://github.com/estuary/connectors/issues/1146

**Workflow steps:**

Run pre-existing tasks with these updates and they will continue from where they left off (well, except for `source-hello-world`, which will start over with the new state shape), and increment the `backfill` for a binding and it will restart from the beginning.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1147)


source-boilerplate/boilerplate.go
```diff
@@ -29,6 +29,11 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+// StateKey is used to key binding-specific state a connector's driver checkpoint. It's just a type
+// alias for a string to help differentiate any old string from where a specific state key from the
+// runtime should be used.
+type StateKey string
```
williamhbaker, line 35: I originally had this as a temporary measure to help with making sure I didn't use the wrong string anywhere. It seemed useful enough to leave here after going through that on some of the initial conversions though.

willdonnelly (approved):
LGTM

This all seems reasonable to me. I like the `boilerplate.StateKey` type declaration, it makes the rest of the code using those keys very clear.

PR #1149, opened 2023-12-21
sqlcapture: Downgrade scary discovery warnings

willdonnelly:
**Description:**

These are all things which should not result in the user being shown scary messages. The excluded view/table logic is totally normal and not an issue at all, and while the unhandled datatype stuff often indicates something we could improve in our connector or test coverage, we're not systematically collecting this info and showing it to the user as a warning is basically never going to be helpful.

(In fact if we wanted to make a list of column types which encounter issues in production, the obvious approach would be to query the live_specs table in Supabase and look for those `"using catch-all schema"` description strings, so the log message part of this is really not helpful at all)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1149)


williamhbaker (approved):
LGTM

PR #1151, opened 2023-12-22
source-mongodb: use `stateKey` for tracking binding state

williamhbaker:
**Description:**

Updates the connector to use `stateKey` for each binding to track its state.

This includes a migration for existing captures. Once all existing captures have started up with this new connector and we are confident that no additional old captures will be re-enabled, we can remove the migration code.

I manually tested this by starting a MongoDB capture with the current `dev` image, set it up to capture across several databases and collections within those databases, and then simulated a migration by updating to the image built from this code and observed the binding states continued from where they left off.

**Workflow steps:**

Increment `backfill` counters on `source-mongodb` captures and those collections will re-backfill starting from the beginning. Increment _all_ of the `backfill` counters and the entire capture will start fresh, with a new stream resume token as well. 

**Notes for reviewers:**


source-mongodb/pull.go
```diff
@@ -204,6 +209,88 @@ type captureState struct {
...
+
+		stateFromOld, ok := state.OldResources[resourceId(res)]
+		if !ok {
+			// This may happen if the connector has never emitted any checkpoints with data for this
+			// binding.
+			ll.Warn("no state found for binding while migrating state")
+			continue
+		}
+
+		state.Resources[boilerplate.StateKey(b.StateKey)] = stateFromOld
```
jgraettinger, line 247: nit: cast of StateKey isn't needed

jgraettinger (approved):
LGTM

source-mongodb/pull.go
williamhbaker: Hmm I'm finding that it is: `source-mongodb/pull.go:247:19: cannot use b.StateKey (variable of type string) as boilerplate.StateKey value in map index`

I may be missing something, but since the `migrateState` function is not intended to be permanent I'm not going to worry too much more about it right now.

PR #1156, opened 2024-01-03
materialize-mongodb: remove _id property from loaded docs

psFried:
Removes the `_id` property before returning loaded documents. This is to prevent that property from causing schema validation errors for collections that use a top level `reduce: { strategy: merge }`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1156)

PR #1157, opened 2024-01-03
Revert "materialize-mongodb: remove _id property from loaded docs"

psFried:
Reverts estuary/connectors#1156

We need to re-think this. As implemented here, we'd break use cases where the users data legitimately includes an `_id` field (such as if they're capturing from mongodb 🤦 ).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1157)

### DIFF
--- a/materialize-mongodb/transactor.go
+++ b/materialize-mongodb/transactor.go
@@ -280,14 +280,6 @@ func (t *transactor) storeWorker(
 }
 
 func sanitizeDocument(doc map[string]interface{}) map[string]interface{} {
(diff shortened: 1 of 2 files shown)


psFried:
False alarm.
The reason this is safe is that there's two possibilities if the source collection documents contain an `_id` property. Either:

- The collection key is `[ /_id ]`: in which case, we're guaranteed that the _incoming_ document will have an `_id` property that will get encoded as the `_id` on the stored document.
- The collection key is anything else: in which case, the `_id` on the source documents will be overwritten by the connector, anyway.

PR #1178, opened 2024-01-16
`flow-sdk` updates

### LINKED ISSUE #1179: python connectors: respect per-binding state key

jshearer:
**Description:**

* Fix bug where Singer shim was not emitting schema for `/_meta/row_id` for bindings that don't have a defined key field
* Proper state support for Singer and Airbyte. We now emit state messages with the standard `bindingStateV1` key containing an object whose keys correspond to binding `stateKey`s.
* Persist `rowId` counters inside `bindingStateV1`, and re-hydrate them if available upon connector restarts. These live inside the same `stateKey`-keyed fields as the actual connector state, so they'll get blown away as well upon re-backfill
* Change Singer connectors to use `["/tap_stream_id"]` as the `resourcePathPointers` instead of `/stream`, as the docs claim that it's possible to have multiple streams with identical `stream` values, but `tap_stream_id` is unique
* Update Airbyte shim to store the value of `namespace` alongside `stream` in the `resourceConfig`, as it's part of the resource path
* Bump `flow-sdk` version to `0.1.5`, publish to PyPi
* Update test snapshots to reflect schema changes (and apparently Asana changed something minor in their API, so that's also reflected)

Closes #1179

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1178)


python/flow_sdk/shim_singer_sdk.py
```diff
@@ -92,14 +168,36 @@ def discover(self, discover: request.Discover) -> response.Discovered:
                     for p in entry.key_properties
                 ]
             else:
+                logger.info(f"Stream {stream.name} does not have a primary key, using '/_meta/row_id' as the key.") 
+                if entry.replication_method == REPLICATION_INCREMENTAL:
+                    logger.warn(f"Stream {stream.name} is using {REPLICATION_INCREMENTAL} replication without a primary key. If this is not an append-only data source, updates to existing documents will be captured as inserts with new keys. Switch to {REPLICATION_FULL_TABLE} to fix this.")
```
jgraettinger, line 173: 🤔 I think it's very likely we'll want to tone down this warning.

What's a connector supposed to do to not be scary if it _wants_ this behavior?

ETA: Oh, I suppose it could use `/_meta/row_id` as the key explicitly? But for connector ergonomics, I'm not sure we should be asking that of users / connector developers.

python/flow_sdk/shim_singer_sdk.py
```diff
@@ -92,14 +168,36 @@ def discover(self, discover: request.Discover) -> response.Discovered:
...
+                
                 key = ["/_meta/row_id"]
+                row_id_type = Property("row_id", IntegerType, required=True)
+
+                if entry.schema.properties is not None:
```
jgraettinger, line 178: Shouldn't we schematize `/_meta/row_id` in *all* cases? It's always there, right?

python/flow_sdk/shim_singer_sdk.py
```diff
@@ -250,8 +355,9 @@ def sync_streams(self, delegate: singer_sdk.Tap):
 resource_config_schema = {
     "type": "object",
     "properties": {
+        # "tap_stream_id": {"type": "string"},
```
jgraettinger, line 358: nit: left-over ?

source-asana/tests/snapshots/source_asana_tests_test_snapshots__capture__capture.stdout.json
```diff
@@ -216,11 +216,11 @@
         "gid": "1205846102972015"
       },
       "gid": "1205846058555984",
-      "html_text": "<b><a href=\"https://app.asana.com/0/profile/1205846104557056\">David Yaffe</a></b> changed the due date to Nov 2",
+      "html_text": "<b><a href=\"https://app.asana.com/0/profile/1205846104557056\">David Yaffe</a></b> changed the due date to Nov 2, 2023",
```
jgraettinger, line 219: lol I'm guessing it's just because the year rolled over.

python/flow_sdk/shim_airbyte_cdk.py
```diff
@@ -108,6 +153,7 @@ def discover(self, discover: request.Discover) -> response.Discovered:
 
             config: StreamResourceConfig = {
                 "stream": stream.name,
+                "namespace": stream.namespace,
```
jgraettinger, line 156: We were already setting this just below, if it evaluated as truthy. Do we _need_ to always explicitly set it?

[Resource path pointers resolution treats undefined and null identically.](https://github.com/estuary/flow/blob/f4c93998569fa3681b99144b4880ce663ef609c9/crates/agent/src/discovers/specs.rs#L67-L87)

jgraettinger (commented):
source-zoom got bundled into this change -- it should be its own import commit right?

jshearer:
🤦 that's right, I thought I had stashed it but I guess not

python/flow_sdk/shim_singer_sdk.py
jshearer: I guess it seemed like a decent sized footgun to me in that if you accidentally'd the key it would appear to work but generate incorrect data at runtime, and only after a connector restart at that, making it hard to track down. 

But yeah, not having a way to silence the warning is annoying. If you think it's common enough, we can just document it somewhere and get rid of the warning.

python/flow_sdk/shim_singer_sdk.py
jshearer: Well, sort of. I couldn't decide if we should expose both "stream" and "tap_stream_id" or not. Honestly we shouldn't really be exposing either one as editable, since any edit would break the binding. But it might be informative to have exposed, for some reason I can't think of.

source-asana/tests/snapshots/source_asana_tests_test_snapshots__capture__capture.stdout.json
jshearer: Oh duh, that makes perfect sense

python/flow_sdk/shim_singer_sdk.py
jshearer: Removed

python/flow_sdk/shim_airbyte_cdk.py
jshearer: Oop, tunnel vision strikes again. Removed 👍

jgraettinger (approved):
LGTM

jgraettinger:
FYI test failures appear to be real

jshearer:
Fixed the tests (forgot to update snapshots). Airtable is still failing, I'll do that swap right nowish

PR #1180, opened 2024-01-16
source-firestore: use stateKey for tracking binding state

williamhbaker:
**Description:**

Updates the connector to use `stateKey` for each binding to track its state.

This includes a migration for existing captures. Once all existing captures have started up with this new connector and we are confident that no additional old captures will be re-enabled, we can remove the migration code.

I tested this by creating a capture with the "old" connector version, and then switching it to the new version. It picked up where it left off, and incrementing the `backfill` counter could trigger a re-backfill of a binding.

So far I've run all the semi-manual tests and updated the test snapshots. I'm also running the really huge datasynth test just for grins and will update the snapshot for that when it is done, but that will take many hours more than likely. All that should change is the checkpoint representation, in a similar way as the other test snapshots have changed since the connector state is keyed off of state keys now.

See issue https://github.com/estuary/connectors/issues/1146

**Workflow steps:**

Increment backfill counters on `source-firestore` captures and those collections will re-backfill starting from the beginning.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1180)

### DIFF
--- a/source-firestore/pull.go
+++ b/source-firestore/pull.go
@@ -67,24 +67,27 @@ func (driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) error 
 		}
 	}
 
-	var resourceBindings []resource
-	for _, binding := range open.Capture.Bindings {
-		var res resource
-		if err := pf.UnmarshalStrict(binding.ResourceConfigJson, &res); err != nil {
-			return fmt.Errorf("parsing resource config: %w", err)
-		}
-		resourceBindings = append(resourceBindings, res)
+	if err := migrateState(&prevState, open.Capture.Bindings); err != nil {
+		return fmt.Errorf("migrating previous state: %w", err)
 	}
 
-	updatedResourceStates, err := initResourceStates(prevState.Resources, resourceBindings)
+	updatedResourceStates, err := initResourceStates(prevState.Resources, open.Capture.Bindings)
 	if err != nil {
 		return fmt.Errorf("error initializing resource states: %w", err)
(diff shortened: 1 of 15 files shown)


source-firestore/pull.go
```diff
@@ -210,8 +258,10 @@ func (s *captureState) Validate() error {
 func (s *captureState) BindingIndex(resourcePath string) (int, bool) {
 	s.RLock()
 	defer s.RUnlock()
-	if state := s.Resources[resourcePath]; state != nil {
-		return state.bindingIndex, true
+	for _, state := range s.Resources {
```
jgraettinger, line 261: just sanity checking that this function (and friends) are not called super frequently ?

jgraettinger (approved):
LGTM

source-firestore/pull.go
williamhbaker: Hmmm they actually are called very frequently, once per document in fact. I had been thinking of it in terms of not that many bindings to iterate through, but I guess that wouldn't necessarily be the case. I think we could hydrate a lookup table to use instead of looking through the entire list, so I'll do that.

williamhbaker: Alright I added a map to lookup the state key (for getting values from the state) from the resource path (which is what we get from captured documents). It should be more efficient if there are a lot of bindings. I think there's probably a more elegant way to do it, but I couldn't come up with anything that didn't seem like it would take a pretty deep reworking of things.

PR #1184, opened 2024-01-17
fix: Forgot to include `VERSION` files for `source-airtrable` and `source-asana`.

jshearer:
**Description:**

Setting `source-airtable`'s version to `:v2` because we don't want to overwrite the existing connector tag at `:v1`


jshearer:
Screen Shot 2024-01-17 at 16 53 15

See `v1` showing up there, this fixed the issue

jgraettinger (approved):
LGTM

PR #1185, opened 2024-01-17
source-mongodb: better error message when url is for online-archive

psFried:
Mongodb has an "online archive" service, which looks a lot like mongodb but doesn't work with our connector.  This adds an explicit check for URLs that appear to be for that service, and returns a clearer error message in case someone uses such a URL.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1185)


mdibaiee (approved):
Thanks Phil!

PR #1187, opened 2024-01-18
materialize-sql: fixes for column validations and nullability

williamhbaker:
**Description:**

- For `materialize-postgres`, we create columns as "bigint", but pre-created tables may have columns that are just "integers", and we want to support those.
- In general, columns with default values can be NOT NULL even if the Flow collection field is not required, since
the default value will take the place of a null value for the field.

This works for our stats postgres materialization, which uses pre-created tables with `integer` column types (not `bigint`) and also has `NOT NULL` columns with default values where we are materializing nullable fields. I also ran the tests for all other materializations that consider columns with default values.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1187)

### DIFF
--- a/materialize-postgres/sqlgen.go
+++ b/materialize-postgres/sqlgen.go
@@ -53,7 +53,7 @@ var pgDialect = func() sql.Dialect {
 	}
 
 	columnValidator := sql.NewColumnValidator(
-		sql.ColValidation{Types: []string{"bigint"}, Validate: sql.IntegerCompatible},
+		sql.ColValidation{Types: []string{"bigint", "integer"}, Validate: sql.IntegerCompatible},
 		sql.ColValidation{Types: []string{"double precision"}, Validate: sql.NumberCompatible},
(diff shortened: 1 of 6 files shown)


materialize-postgres/sqlgen.go
```diff
@@ -54,6 +54,7 @@ var pgDialect = func() sql.Dialect {
 
 	columnValidator := sql.NewColumnValidator(
 		sql.ColValidation{Types: []string{"bigint"}, Validate: sql.IntegerCompatible},
+		sql.ColValidation{Types: []string{"integer"}, Validate: sql.IntegerCompatible},
```
jgraettinger, line 57: nit: Types is an array -- could this also be a single row of []string{"integer", "bigint"} ?

jgraettinger (approved):
LGTM

Love how easy it is to tweak logic like this now.

materialize-postgres/sqlgen.go
williamhbaker: Indeed it can...thanks!

PR #1189, opened 2024-01-18
fix: Provide primary key `_airtable_id` for `source-airtable`

jshearer:
This was being set by the `airbyte-to-flow` override in the old connector, that never got transferred over to the python connector

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1189)


jgraettinger (approved):
lgtm

PR #1190, opened 2024-01-18
Fix: Make `_airtable_id` not nullable, since it's a primary key

jshearer:
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1190)


psFried (approved):
LGTM

PR #1196, opened 2024-01-19
Update `source-asana` and `source-airtable`'s endpoint config schemas

jshearer:
They were still missing a couple of fields causing the UI to error when authenticating with OAuth

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1196)


jgraettinger (approved):
LGTM

PR #1202, opened 2024-01-23
source-sftp: ssh key authentication

mdibaiee:
**Description:**

- Support SSH key authentication
- Tested by creating a server, enabling SSH key authentication for the server and using the private key in the config of this connector to capture data

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-sftp/main.go
+++ b/source-sftp/main.go
@@ -27,9 +27,10 @@ import (
 type config struct {
 	Address    string         `json:"address" jsonschema:"title=Address" jsonschema_extras:"order=0"`
 	Username   string         `json:"username" jsonschema:"title=Username" jsonschema_extras:"order=1"`
-	Password   string         `json:"password" jsonschema:"title=Password" jsonschema_extras:"secret=true,order=2"`
-	Directory  string         `json:"directory" jsonschema:"title=Directory" jsonschema_extras:"order=3"`
-	MatchFiles string         `json:"matchFiles,omitempty" jsonschema:"title=Match Files Regex" jsonschema_extras:"order=4"`
+	Password   string         `json:"password,omitempty" jsonschema:"title=Password" jsonschema_extras:"secret=true,order=2"`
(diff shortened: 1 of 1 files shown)


source-sftp/main.go
```diff
@@ -27,9 +27,10 @@ import (
 type config struct {
 	Address    string         `json:"address" jsonschema:"title=Address" jsonschema_extras:"order=0"`
 	Username   string         `json:"username" jsonschema:"title=Username" jsonschema_extras:"order=1"`
-	Password   string         `json:"password" jsonschema:"title=Password" jsonschema_extras:"secret=true,order=2"`
-	Directory  string         `json:"directory" jsonschema:"title=Directory" jsonschema_extras:"order=3"`
-	MatchFiles string         `json:"matchFiles,omitempty" jsonschema:"title=Match Files Regex" jsonschema_extras:"order=4"`
+	Password   string         `json:"password,omitempty" jsonschema:"title=Password" jsonschema_extras:"secret=true,order=2"`
```
williamhbaker, line 30: I think it would be good to have password and SSH key as a "one of" option in the config, to make it clear that you don't need to provide both of them. This would probably require writing out the spec by hand since IIRC there's not a way to represent that in the struct tags, and also some kind of backward compatibility with existing specs that have "password" in the top-level. I wish I would have anticipated needing to add another authentication method and originally made this a "one of" to begin with.

It's probably fine having them both as top-level options until/if we need more than 2 authentication options, so I'll leave it up to you as to wether its worth the effort or not to make it a "one of" right now. If nothing else, we should update the descriptions to say "may be left blank if using the other thing".

williamhbaker (approved):
LGTM % updating the spec test snapshot so the connector can build

source-sftp/main.go
mdibaiee: @williamhbaker yeah I thought about `oneOf`, but I think there is a lot of effort necessary for that. At the moment I don't see this connector wanting another authentication method so we can keep it this way for now for simplicity's sake.

PR #1209, opened 2024-01-24
materialize-sql: recommend arrays and fields with multiple types

### LINKED ISSUE #1208: Include arrays by default in materializations

williamhbaker:
**Description:**

Users almost always want arrays and fields with multiple types (materialized as JSON columns) to be materialized by default, so this changes them to be recommended instead of optional.

We will merge this on Monday, Jan 29th, as we have given users a heads-up that this change is coming, and new publications of their tasks will cause these newly recommended fields to be added to their materializations.

Closes #1208

**Workflow steps:**

Create SQL materializations and see that array fields and fields with multiple types are included in the recommended fields, and will be materialized by default. Object fields are still optional.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1209)

### DIFF
--- a/materialize-bigquery/.snapshots/TestValidateAndApply
+++ b/materialize-bigquery/.snapshots/TestValidateAndApply
@@ -1,15 +1,15 @@
 Big Schema Initial Constraints:
 {"Field":"_meta/flow_truncated","Type":4,"TypeString":"FIELD_OPTIONAL","Reason":"Metadata fields fields are able to be materialized"}
-{"Field":"arrayField","Type":4,"TypeString":"FIELD_OPTIONAL","Reason":"This field is able to be materialized"}
+{"Field":"arrayField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This field is able to be materialized"}
 {"Field":"boolField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"The projection has a single scalar type"}
 {"Field":"flow_document","Type":2,"TypeString":"LOCATION_REQUIRED","Reason":"The root document must be materialized"}
 {"Field":"flow_published_at","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"The projection has a single scalar type"}
 {"Field":"intField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"The projection has a single scalar type"}
 {"Field":"key","Type":2,"TypeString":"LOCATION_REQUIRED","Reason":"All Locations that are part of the collections key are required"}
-{"Field":"multipleField","Type":4,"TypeString":"FIELD_OPTIONAL","Reason":"This field is able to be materialized"}
+{"Field":"multipleField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This field is able to be materialized"}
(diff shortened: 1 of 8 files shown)


materialize-bigquery/.snapshots/TestValidateAndApply
```diff
@@ -193,14 +193,14 @@ Big Schema Materialized Resource Schema With No Fields Required:
...
+{"Field":"arrayField","Type":4,"TypeString":"FIELD_OPTIONAL","Reason":"Objects fields may be materialized"}
 {"Field":"boolField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"The projection has a single scalar type"}
 {"Field":"flow_document","Type":2,"TypeString":"LOCATION_REQUIRED","Reason":"The root document must be materialized"}
 {"Field":"flow_published_at","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"The projection has a single scalar type"}
 {"Field":"intField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"The projection has a single scalar type"}
 {"Field":"key","Type":2,"TypeString":"LOCATION_REQUIRED","Reason":"All Locations that are part of the collections key are required"}
-{"Field":"multipleField","Type":4,"TypeString":"FIELD_OPTIONAL","Reason":"This field is able to be materialized"}
-{"Field":"nullField","Type":4,"TypeString":"FIELD_OPTIONAL","Reason":"This field is able to be materialized"}
+{"Field":"multipleField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This field is able to be materialized"}
+{"Field":"nullField","Type":4,"TypeString":"FIELD_OPTIONAL","Reason":"Objects fields may be materialized"}
```
mdibaiee, line 203: I find the message here to be confusing since `type: null` is not really an object

mdibaiee (approved):

materialize-bigquery/.snapshots/TestValidateAndApply
williamhbaker: These snapshots are a little bizarre when viewed in isolation, but may make more sense in the overall sequence of how the tests run. The test that produces these does the following:
1. Applies a materialization spec with a whole bunch of fields, covering every data type I could come up with, where the name of the field corresponds with the type of the field. That is above in `Big Schema Initial Constraints`.
2. Applies the exact same spec again in `Big Schema Re-validated Constraints`.
3. Scrambles the types and tries to apply again in `Big Schema Changed Types Constraints`. In that case, `nullField` was originally `null` and didn't have a column created for it, but the changed types had it become an object, which is now able to be materialized.
4. Does a couple more tests with nullabilty changes to fields in `Big Schema Materialized Resource Schema With All Fields Required` and `Big Schema Materialized Resource Schema With No Fields Required`
5. Finally, the snapshot line in question is produced from the scrambled type snapshot, with the backfill counter incremented on the table. It illustrates that although there were previously incompatible changes, incrementing the backfill counter validates the materialization as-new. With the scrambled type spec, `nullField` is actually an object, which is why it has that `Objects fields may be materialized` (also noting that I need to change the plurality of Objects)

Issue #1182, opened 2024-01-17
improve materialization update delay


williamhbaker:
The "update delay" feature of OLAP SQL materializations is currently pretty simple: It will apply the delay if the current transaction is below number of keys as an approximation for when a backfill is occurring, and it will always skip the delay on the first transaction of the connector's invocation.

We have observed this sometimes doesn't work well. The first transaction of an invocation is often very small (which is why the delay is skipped), but then if there is a delay on the _next_ transaction it can make it look like the materialization isn't doing anything. Also, there are occasionally small transactions mixed in with other large backfill transactions for reasons I won't get into.

We should consider requiring some number of initial transactions to pass before applying the delay, and/or requiring a consecutive number of small transactions as the connector is running to start applying the delay.

https://github.com/estuary/connectors/pull/1117 will move the materialization transactor into the connectors repo, so that would make it easier to incorporate this delay into the transactor, if it makes sense to do so. We might also consider cancelling the delay if the number of load requests for the next concurrent transaction is above a threshold, which is what makes me think we might want to put it in the transactor. Note that this would not work for delta-updates materializations, and may not work well considering recent runtime optimizations to not send load requests for keys that are known to not yet exist in the destination.

jgraettinger:
I think it's likely this really belongs in the runtime. Doing it in the connector ... works, but has the downsides you note, and also: the connector cannot tell that the task is trying to restart with a new configuration, but the runtime can. This frequently comes up as users make changes and then wonder why it's not immediately taking effect (current reason: because we're waiting for the non-runtime-cancelable update delay to elapse)

PR #1210, opened 2024-01-24
materializations: make updateDelay consider the last several transactions

williamhbaker:
**Description:**

Applying the additional "updateDelay" utilized by OLAP SQL materializations has been done by considering just the size of the current transaction to estimate if a backfill is in progress. This causes false activations of the delay due to variable sized runtime transactions - sometimes they are very small, even during a backfill, for various reasons.

This changes the delay decision to use the last 5 transaction sizes to decide if the delay should be used. As part of this, the delay logic is being moved to the transactor itself, which simplifies the implementation for connectors that want to use the delay.

Closes [#1182](https://github.com/estuary/connectors/issues/1182)

**Workflow steps:**

**Notes for reviewers:**

This puts a little more complexity into `RunTransactions`, and there's already a lot going on there. But I think the reduction in complexity for individual materializations is still worth it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1210)


go/protocols/materialize/transactor.go
```diff
@@ -132,15 +163,44 @@ func RunTransactions(
...
+		// set in the "round" loop after awaitDoneCh has closed.
+		//
+		// commitStartedAt is the time at which the the commit for the latest round began.
+		commitStartedAt time.Time
+		// storedHistory is a list of the most recent stored counts for transactions.
+		storedHistory [storedHistorySize]int
 	)
 
+	// True only if all historical stored counts are below the threshold, and none of them are 0.
+	// Counts at 0 means that many transactions haven't been completed yet.
```
mdibaiee, line 194: The zero value is strange. Can we not wait until `startCommit` is called and finished before we store `it.Total` in `storedHistory`?

go/protocols/materialize/transactor.go
```diff
@@ -300,6 +378,7 @@ func RunTransactions(
 
 		// `startCommit` may be nil to indicate a no-op commit.
 		var stateUpdate *pf.ConnectorState = nil
+		commitStartedAt = time.Now()
 		if startCommit != nil {
 			stateUpdate, ourCommitOp = startCommit(
 				stream.Context(), runtimeCheckpoint)
```
mdibaiee, line 384: I'm wondering if we can move the line below here so that it is never zero.
```
storedHistory[round%storedHistorySize] = storeIt.Total
```

williamhbaker: `storeIt.Total` will not be zero after `transactor.Store` returns - ref calls to `(*StoreIterator).Next` in `lifecycle.go`. The counts of 0 are from the array's initialization, where it is initialized at N values, but we haven't completed any transactions yet. We don't want it to look like a bunch 0 sized transactions have run, and a transaction of 0 documents is impossible, so they are skipped really just until N transactions have been run and the array buffer is "filled".

[`container/Ring`](https://pkg.go.dev/container/ring) would work for this too, but I don't think the ergonomics of it are great compared to just using a fixed size array.

williamhbaker: I updated some of the comments around this logic which may be helpful. LMK what you think.

mdibaiee (approved):

PR #1217, opened 2024-01-26
fix: Airbyte shim typo

jshearer:
The consequence of this was that we were not persisting state messages (and crashing instead), so incremental captures would fail to be incremental.

Note, this also adds a more clear exception for a scenario I encountered in a customer's task where it was outputting documents for a binding/stream ID that we didn't know about, and crashing. I _think_ we should still crash in this case, but at least now the message will be more clear and make it easier to track down the problem. 

As a fast-follow I would like to add integration test coverage for incremental captures

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1217)

### DIFF
--- a/python/flow_sdk/shim_airbyte_cdk.py
+++ b/python/flow_sdk/shim_airbyte_cdk.py
@@ -259,7 +259,8 @@ class CaptureShim(Connector):
             if record := message.record:
                 entry = index.get((record.namespace, record.stream), None)
                 if entry is None:
-                    raise Exception(f"Document read in unrecognized stream {record.stream} (namespace: {record.namespace})")
+                    logger.warn(f"Document read in unrecognized stream {record.stream} (namespace: {record.namespace})")
+                    continue
 
                 record.data.setdefault("_meta", {})["row_id"] = entry[1]
                 entry[1] += 1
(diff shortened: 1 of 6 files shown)


python/flow_sdk/shim_airbyte_cdk.py
```diff
@@ -257,7 +257,11 @@ def open(self, open: request.Open, emit: t.Callable[[Response], None]) -> None:
             logger, config, catalog, state.to_airbyte_input()
         ):
             if record := message.record:
-                entry = index[(record.namespace, record.stream)]
+                entry = index.get((record.namespace, record.stream), None)
+                if entry is None:
+                    logger.warn(f"Document read in unrecognized stream {record.stream} (namespace: {record.namespace})")
```
jgraettinger, line 262: i assume this is temporary per your update? it has the potential to produce a lot of logs.

jgraettinger (approved):
LGTM

python/flow_sdk/shim_airbyte_cdk.py
jshearer: Yeah, it's temporary. I just really didn't want to introduce a change that lets us accidentally read from disabled streams and then drop those records on the floor without complaining about it.

PR #1218, opened 2024-01-26
materialize-snowflake: unset internal state for empty bindings

mdibaiee:
**Description:**

- Unset the internal state of bindings which have no documents so no queries are run for those in subsequent transactions

**Workflow steps:**

**Notes for reviewers:**


jgraettinger:
This seems to remove the binding state for binding X, while preparing a transaction to commit where binding Y, Z, are updated (but not X). Right?

Why is the post-Acknowledge state clearing mechanism insufficient? Is it because the _runtime_ state has the binding removed per the post-Acknowelge flow, but the connector's view of the state doesn't ? I'm also assuming (but didn't verify) that it's _not_ using merge-patch state updates.

This would be a harder mistake to make, if a) the connector issued merge-patch updates of its state, and 2) issued merge-patch updates like deletions were coded up side-by-side with the corresponding operation updating the connector's internal state view. Agree? Is that an achievable factoring without a ton of effort?

mdibaiee:
@jgraettinger that does make sense actually. We do use merge patch updates, but at the point where we clear the checkpoint of bindings that have been processed, we were not clearing it up from the connector's current state. I just moved the deletion of the binding state to the point where we cleanup the checkpoint

materialize-snowflake/snowflake.go
```diff
@@ -568,6 +568,7 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
 	var checkpointClear = make(checkpoint)
 	for _, b := range d.bindings {
 		checkpointClear[b.target.StateKey] = nil
+		delete(d.cp, b.target.StateKey)
```
williamhbaker, line 571: With this change, I don't think we need the `if len(item.Query) == 0 {` check above anymore, on line 531.

williamhbaker (approved):
This LGTM, assuming @jgraettinger is good with the update per his comment.

I had a thought that constructing an entirely separate "checkpoint" object was even a bit more than we needed to do, and could instead more directly update the internal state to clear the completed queries and then marshal that internal state into the checkpoint to emit. But I guess this would be slightly less efficient, and kind of annoying to deal with the `nil` checkpoints values we'd need for merge patches to work out.

jgraettinger (approved):
LGTM

PR #1220, opened 2024-01-29
source-dynamodb: limit frequency of calls to describeStream

williamhbaker:
**Description:**

Describe stream calls that are needed to get a shard listing are limited to 10 per second, so we need to do requests less often than that to avoid errors. The main place we do this call is when listing shards, which is done periodically for each table being streamed, and so a rate limiter is needed to prevent lots of these shard listings from lining up and causing an API rate limit error.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1220)


williamhbaker:
Added a couple more commits to this PR:
- Remove the state migration code that was used for the inclusion of state keys, since all tasks have been migrated
- Update the connector to use merge patches for more efficient state updates, since we have seen some pretty extreme recovery log usage from dynamo captures with high rates of changes being captured

willdonnelly (approved):
LGTM

Lots of variable renaming noise in the diff, but the changes all look reasonable and the rate limiter parts seem good.

PR #1227, opened 2024-01-31
materialize-snowflake: switch checkpoint format for migration

mdibaiee:
**Description:**

- A small change in the format of checkpoint item which will allow us to distinguish between old v2 checkpoints. In case we do find an old v2 checkpoint, we clean up the checkpoint and start anew.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1227)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -251,6 +251,14 @@ func (d *transactor) UnmarshalState(state json.RawMessage) error {
 	if err := json.Unmarshal(state, &d.cp); err != nil {
 		return err
 	}
+	// TODO: remove after migration
+	for _, item := range d.cp {
+		if len(item.StagedDir) == 0 {
+			log.WithField("state", string(state)).Info("found old checkpoint format, ignoring the checkpoint and starting clean")
+			d.cp = make(checkpoint)
+			break
+		}
+	}
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


materialize-snowflake/snowflake.go
```diff
@@ -446,9 +454,9 @@ func (d *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage)
 }
 
 type checkpointItem struct {
-	Table    string
-	Query    string
-	ToDelete string
```
williamhbaker, line 451: With the way we are using merge patches, I don't think this will ever completely remove the old `ToDelete` field. I think it would technically still work to always have that field in legacy checkpoints since we aren't doing a strict unmarshalling the forbids unknown fields, but I'd feel better if we explicitly deleted the field from checkpoints so it doesn't cause problems down the road. 

We should be able to use something like this:
```go
type checkpointItem struct {
	Table     string
	Query     string
	StagedDir string
	ToDelete  *string `json:",omitempty"`
}
```

And then for the migration part, set `ToDelete` to `nil` so that it gets marshalled as `null` in the merge patch and fully deleted from the checkpoint.

williamhbaker:
A more general concern: How will this migration work for existing tasks that are on `v2` and always were `v2`? Would they have a prior valid checkpoint cleared?

materialize-snowflake/snowflake.go
williamhbaker: I'll actually walk this back - since we are emitting checkpoints at post-ack that have `checkpointItem: nil`, as long as one of _those_ gets emitted, then the checkpoint will be cleared of its old field. And it's quite reasonable to assume that those post-ack checkpoints will be emitted in a timely fashion, since it always does one at startup.

williamhbaker (approved):
LGTM, per slack discussions

PR #1228, opened 2024-01-31
source-sftp: oneOf credentials jsonschema

mdibaiee:
**Description:**

- I attempted to make a more generalised way of creating `oneOf`s without outputting the schema directly, and I realised we soon end up almost re-implementing jsonschema. There probably still is some way to get a middle ground where we reduce boilerplate for connectors but also don't end up maintaining too much code on the side of the abstraction, but for now I gave up on it. My attempt is [here](https://github.com/estuary/connectors/commit/de9cc38121882015614fbe7f5ad5fa582ff8aab0). If I was to do this again, I would probably do this with multiple structs, where each struct represents one of the `oneOf` cases, this allows for more control over specifics of each and would probably also provider a better interface, we just would need a way of parsing the config into these multiple specs which itself can be a challenge.
- Tested this change by testing the previous config syntax (both password and sshKey) and also the new `credentials` formats and all four versions worked.
- Also tested the form on the jsonform test page and it renders as expected
image
image

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1228)

### DIFF
--- a/source-sftp/.snapshots/TestSpec
+++ b/source-sftp/.snapshots/TestSpec
@@ -1,6 +1,5 @@
 {
-  "$schema": "http://json-schema.org/draft/2020-12/schema",
-  "$id": "https://github.com/estuary/connectors/source-sftp/config",
+  "$schema": "http://json-schema.org/draft-07/schema#",
   "properties": {
     "address": {
       "type": "string",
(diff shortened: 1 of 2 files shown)


source-sftp/.snapshots/TestSpec
```diff
@@ -1,33 +1,78 @@
...
+              "title": "Username",
+              "description": "Username for authentication.",
+              "order": 1
+            },
+            "sshKey": {
+              "type": "string",
+              "title": "SSH Key",
+              "description": "SSH Key for authentication",
+              "multiline": true,
+              "order": 3,
```
williamhbaker, line 36: This order should be 2?

source-sftp/.snapshots/TestSpec
```diff
@@ -1,33 +1,78 @@
...
+              "title": "Password",
+              "description": "Password for authentication",
+              "order": 2,
+              "secret": true
+            }
+          },
+          "required": [
+            "type",
+            "username",
+            "sshKey"
```
williamhbaker, line 72: Should be `password` instead of `sshKey` for this object?

source-sftp/main.go
```diff
@@ -127,20 +135,39 @@ func (c config) PathRegex() string {
...
 
-	// For now only password authentication is supported.
-	if cfg.Password != "" {
-		sshConfig.Auth = append(sshConfig.Auth, ssh.Password(cfg.Password))
-	} else if cfg.SSHKey != "" {
-		if signer, err := ssh.ParsePrivateKey([]byte(cfg.SSHKey)); err != nil {
-			return nil, fmt.Errorf("parsing ssh key: %w", err)
+	// Legacy authentication method
+	if cfg.Credentials.Type == "" {
+		if cfg.Password != "" {
```
williamhbaker, line 151: Do we have any tasks with the legacy authentication method that are using SSH key? I'd imagine not yet, and so the legacy authentication could be simplified to only consider the password case.

williamhbaker (approved):
LGTM % comments

source-sftp/.snapshots/TestSpec
mdibaiee: Type is order 1, username is order 2, sshKey is order 3?

williamhbaker: Oh I must have been missing the ordering for the type!

PR #1231, opened 2024-02-01
Fix airbyte state serialization, also fix `source-google-ads`

jshearer:
* Fixes Airbyte shim state serialization/loading logic
* Add temporary (maybe no longer needed?) bypass to just skip invalid state data on startup and re-backfill instead
* Remove noisy log
* Update source-google-ads to have specs and tests

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1231)


python/flow_sdk/shim_airbyte_cdk.py
```diff
@@ -97,18 +92,21 @@ def to_flow_state(self, index: dict[t.Tuple[str | None, str], t.List[int]], bind
...
+
+            index: dict[t.Tuple[str | None, str], t.List[int]] = {}
+            for i, binding in enumerate(bindings):
+                if binding["stateKey"] in parsed_state.bindingStateV1:
+                    namespace = binding["resourceConfig"].get("namespace",None)
+                    stream = binding["resourceConfig"]["stream"]
+                    index[(namespace, stream)] = [i, parsed_state.bindingStateV1[binding["stateKey"]].rowId]
+            return (parsed_state, index)
+
+        except Exception as e:
```
jgraettinger, line 107: AFAIK this is no longer needed, and for that reason I think we should take it out now.

python/flow_sdk/shim_airbyte_cdk.py
```diff
@@ -97,18 +92,21 @@ def to_flow_state(self, index: dict[t.Tuple[str | None, str], t.List[int]], bind
...
-        index: dict[t.Tuple[str | None, str], t.List[int]] = {}
-        for i, binding in enumerate(bindings):
-            if binding["stateKey"] in state.bindingStateV1:
-                namespace = binding["resourceConfig"].get("namespace",None)
-                stream = binding["resourceConfig"]["stream"]
-                index[(namespace, stream)] = state.bindingStateV1[binding["stateKey"]].rowId
-        return (state, index)
+    def from_flow_state(state: dict | None, bindings: t.List[CaptureBinding]):
+        try:
+            parsed_state = state and State(**state) or State()
```
jgraettinger, line 97: oh neat, this will recursively hydrate the structure from raw `dict` ?

jgraettinger (approved):
LGTM % comment on malformed state

python/flow_sdk/shim_airbyte_cdk.py
jshearer: Yep! `AirbyteStateMessage`, `StreamState` and all

python/flow_sdk/shim_airbyte_cdk.py
jshearer: I left the try-catch in so we have a useful error message, but now the exception is re-raised

jgraettinger: Sounds good, thanks.

FYI there's already some top-level code that catches exceptions and logs them out as structured errors.

PR #1233, opened 2024-02-01
materialize-postgres: limit batch size based on document size

williamhbaker:
**Description:**

This implements batch flushing logic based on the size of documents handled, rather than a simple count of documents.

For loads, it uses the size of the packed key. For stores, it uses the size of packed key, packed values, and raw document JSON. This is just an approximation of the memory used for buffering the queries for these documents in a `pgx.Batch`, but it seemed to work pretty well in a synthetic test case: For documents each around 100kb, this version of the connector will use just a little over 100mb of memory per `docker stats` when doing stores, and have batch sizes of about 50 documents. Each materialized row has all of the fields from the document as well as a field with the entire document, so 50 documents is just about right with a batch size limit of 10mb.

For this same synthetic dataset, the prior version of the connector with its 4096 document count limit would reach 1gb of memory consumption and OOM.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1233)

### DIFF
--- a/go/protocols/materialize/lifecycle.go
+++ b/go/protocols/materialize/lifecycle.go
@@ -243,13 +243,14 @@ func ReadFlush(request *pm.Request) error {
 
 // StoreIterator is an iterator over Store requests.
 type StoreIterator struct {
-	Binding   int             // Binding index of this stored document.
-	Exists    bool            // Does this document exist in the store already?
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

PR #1245, opened 2024-02-02
source-mysql: Normalize zero-month/day in MySQL datetimes

willdonnelly:
**Description:**

These are invalid and you have to make sure the `NO_ZERO_IN_DATE` mode flag is cleared in order to insert them, but since that's a thing we can receive from the database we need to handle it. We'll handle them by normalizing the zero component to one instead.

This code is rather ugly and produces off-by-one results for backfills versus replication, but it's sufficient to get things unblocked in production and such invalid datetimes are generally a hacky way of saying something like "July 2023" without additional detail anyway, so the inconsistency doesn't feel like the end of the world. We should still chase that down soon, obviously.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1245)

### DIFF
--- a/source-mysql/.snapshots/TestDatetimeNormalization-replication
+++ b/source-mysql/.snapshots/TestDatetimeNormalization-replication
@@ -1,0 +1,12 @@
+# ================================
+# Collection "acmeCo/test/test_datetimenormalization_24528211": 4 Documents
+# ================================
+{"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"DatetimeNormalization_24528211","cursor":"binlog.000123:56789:123","txid":"11111111-1111-1111-1111-111111111111:111"}},"id":200,"x":"1991-08-31T17:34:57Z"}
+{"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"DatetimeNormalization_24528211","cursor":"binlog.000123:56789:123","txid":"11111111-1111-1111-1111-111111111111:111"}},"id":201,"x":"0001-01-01T00:00:00Z"}
+{"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"DatetimeNormalization_24528211","cursor":"binlog.000123:56789:123","txid":"11111111-1111-1111-1111-111111111111:111"}},"id":202,"x":"2022-11-30T06:00:00Z"}
+{"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"DatetimeNormalization_24528211","cursor":"binlog.000123:56789:123","txid":"11111111-1111-1111-1111-111111111111:111"}},"id":203,"x":"2023-06-30T05:00:00Z"}
+# ================================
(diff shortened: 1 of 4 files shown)


source-mysql/.snapshots/TestDatetimeNormalization-replication
```diff
@@ -0,0 +1,12 @@
+# ================================
+# Collection "acmeCo/test/test_datetimenormalization_24528211": 4 Documents
+# ================================
+{"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"DatetimeNormalization_24528211","cursor":"binlog.000123:56789:123","txid":"11111111-1111-1111-1111-111111111111:111"}},"id":200,"x":"1991-08-31T17:34:57Z"}
+{"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"DatetimeNormalization_24528211","cursor":"binlog.000123:56789:123","txid":"11111111-1111-1111-1111-111111111111:111"}},"id":201,"x":"0001-01-01T00:00:00Z"}
+{"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"DatetimeNormalization_24528211","cursor":"binlog.000123:56789:123","txid":"11111111-1111-1111-1111-111111111111:111"}},"id":202,"x":"2022-11-30T06:00:00Z"}
```
williamhbaker, line 6: I'm understanding that this corresponds to the input `"2023-00-00 00:00:00"` - but I'm not understanding how the `2023` year part gets turned into a `2022`? It seems like we are only adding to the input. Is this just something weird happening with MySQL that I shouldn't worry too much about?

willdonnelly: Yeah, that's the inconsistency I referred to. I have absolutely no idea why we're seeing `2023-00-00` turned into `2022-11-30` (note that this is completely insane, it should _at least_ be `2022-12-31` if things are getting normalized backwards like that) but I've confirmed that it's nothing we're doing -- that's what the literal replication events coming from the database appear to say (though I have not yet ruled out a client library bug in how the binary representation of these datetimes is parsed, so it still might be something we could fix in theory).

willdonnelly: I'm not entirely sure (still haven't finished chasing down the relevant parsing code), but _I think_ that the MySQL replication protocol might actually be representing datetime values as a 4-byte Unix timestamp, which would kind of halfway explain why things get normalized oddly there and make it Not Our Problem, in my opinion.

williamhbaker (approved):
LGTM - My biggest concern would be this special normalization making otherwise valid datetimes different than they were before, but I'm not seeing any way for that to happen. 

I'm not really sure if this apparently sometimes strange conversion is better than just emitting the sentinel value, but also not sure if it's worse, so :shrug: - and maybe it can be improved later.

willdonnelly:
I'm going to go ahead and merge this as-is for now. We have tests sufficient to convince me that this doesn't break the common case, and for now I think unblocking a production capture takes priority.

PR #1247, opened 2024-02-05
source-sftp: set recommended_name when directory is /

psFried:
A user entered `/` for the `directory` in the config, and it resulted in the `recommended_name` of the discovered binding being empty. This ensures that the recommended name is always non-empty by defaulting it to `root-directory`. Feel free to suggest a better default name.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1247)

### DIFF
--- a/source-sftp/main.go
+++ b/source-sftp/main.go
@@ -118,7 +118,12 @@ func (c config) DiscoverRoot() string {
 }
 
 func (c config) RecommendedName() string {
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
I'd probably just call it `root`, to prevent weird handling of hyphens in hypothetical materialized table names of the collection

PR #1249, opened 2024-02-05
materialize-sql: forbidden constraints for fields that are too long

williamhbaker:
**Description:**

Rather than erroring during `Apply`, we will now produce "field forbidden" constraints for Flow collection fields that are longer than the destination system will allow.

This is slightly complicated by that fact that we want to return an error if "required" fields cannot be materialized due to length limits, rather than successfully validate with a "forbidden" constraint. This requires tracking all projections of a location and making sure that at least one of them is not too long if the location is required.

**Workflow steps:**

**Notes for reviewers:**


materialize-bigquery/transactor.go
```diff
@@ -321,21 +321,12 @@ func (t *transactor) commit(ctx context.Context) error {
...
 
 	// This returns a single row with the error status of the query.
-	job, err := t.client.runQuery(ctx, query)
-	if err != nil {
+	if _, err := t.client.runQuery(ctx, query); err != nil {
 		return fmt.Errorf("commit query: %w", err)
 	}
 	log.Info("store: finished commit")
 
-	if err != nil {
```
williamhbaker, line 330: Did some cleanup here since the second `if err != nil { ... }` condition was impossible to reach.

materialize-boilerplate/testdata/generate-spec-proto.sh
```diff
@@ -28,11 +28,9 @@ function cleanup {
 }
 trap cleanup EXIT
 
-flowctl-go api build \
-    --build-id build-id \
-    --build-db ${TEMP_DIR}/build.db \
-    --network flow-test \
-    --source $1 \
-    --log.level info
+flowctl raw build \
```
williamhbaker, line 31: Moving over to `flowctl raw build` for long-term compatibility. This did cause all of the generate proto files to be updated.

materialize-mysql/client.go
```diff
@@ -189,28 +189,3 @@ func (c *client) FetchSpecAndVersion(ctx context.Context, specs sql.Table, mater
...
 	c.db.Close()
 }
-
-func checkIdentifierLengthError(err error) error {
-	if err == nil {
-		return nil
-	}
-
-	// Handling for errors resulting from identifiers being too long.
-	// TODO(whb): At some point we could consider moving this into `Validate` to check
```
williamhbaker, line 199: Finally able to clean this up, since MySQL and MariaDB both have the same limit of 64 characters for a column name.

mdibaiee (approved):

PR #1255, opened 2024-02-09
materializations: don't check for identical resource paths in validate

williamhbaker:
**Description:**

This check was problematic because it didn't allow for "removing" a binding and "adding" a different binding to the same table in the same publication. There is no reason to forbid this, since the actual columns & types of the materialized resource are used for validating that a binding is compatible. So, the source collection of a binding could be changed and still go to the same table, as long as the already materialized table is compatible with the source collection.

There are edge cases where this might be a problem, such as if the previous binding was delta updates and the new one is not and the target table has existing rows in duplicate for certain keys from the delta updates materialization. This same thing could currently happen if the binding were removed, published, and then added back with a different source collection in the same publication, so this change doesn't seem like it makes that situation any worse. And there is a fix for this, via the backfill counter of the binding.

As for the possibility of there being two identical resource paths within the same materialization, we don't need to check that in the connector anymore since the control plane validates that it can't happen.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1255)

### DIFF
--- a/materialize-boilerplate/validate.go
+++ b/materialize-boilerplate/validate.go
@@ -323,8 +323,7 @@ func (v Validator) ambiguousFieldIsSelected(p pf.Projection, fieldSelection []st
 	return false
 }
 
-// findExistingBinding locates a binding within an existing stored specification, and verifies that
-// there are no target path conflicts with a proposed binding that does not already exist.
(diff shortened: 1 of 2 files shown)


psFried (approved):
LGTM

PR #1258, opened 2024-02-09
materializations: find existing bindings based only on resource path

williamhbaker:
**Description:**

Matching on both the collection name + resource path doesn't work well if the backfill counter for a binding is incremented simultaneously with changing the source collection. This isn't terribly common, but will automatically happen via the schema evolutions handler when a collection key changes. In this case, the current behavior will consider the binding brand new, and will not re-create the table.

Matching on just the resource path is consistent with how the runtime computes state keys for bindings, which are only based on the resource path. This will allow the materialization to consider a binding with a different source collection for the same resource path to be a continuation of that binding, and will react to backfill counter changes accordingly.

This also resolves edge cases noted in 3b616f2, where changing the source collection for a table simultaneously with switching from delta updates to standard updates would cause problems. We can now prevent that switch from happening as well.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1258)

### DIFF
--- a/materialize-boilerplate/apply.go
+++ b/materialize-boilerplate/apply.go
@@ -112,7 +112,7 @@ func ApplyChanges(ctx context.Context, req *pm.Request_Apply, applier Applier, i
 		// The existing binding spec is used to extract various properties that can't be learned
 		// from introspecting the destination system, such as the backfill counter and if the
 		// materialization was previously delta updates.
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

PR #1261, opened 2024-02-13
materialize-redshift: compress persisted spec bytes

williamhbaker:
**Description:**

The VARBYTE column type we are using to store specs has a maximum size of 1MB. This is not big enough to store some of the larger specs that we have seen in the wild for materializations with fairly reasonable numbers of bindings. This change stores those specs with compression, which in some quick tests I did with large specs reduces the their size by 5x or more.

This is fairly simple change and with a small scope, compared to something more involved like using a VARIANT column with multiple fields to chunk up the spec. But it is pretty low risk and I'm hoping it will push off larger changes until we start providing the "last validated" spec in `Validate` calls, and don't need to persist a spec at all.

Tested by running the unit & integration tests which exercise spec persisting and loading, and also manually doing a migration scenario where an older task is re-applied with the new task, which loads an uncompressed spec and persists it compressed.

**Workflow steps:**

**Notes for reviewers:**

Theoretically this could be made a higher-level concern of `materialize-sql`, but since we haven't seen spec bytes size restriction problems anywhere else I wanted to limit the blast radius if something doesn't work out right for now. It is something we could consider for the future though.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1261)

### DIFF
--- a/materialize-redshift/client.go
+++ b/materialize-redshift/client.go
@@ -1,12 +1,15 @@
 package main
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
 	stdsql "database/sql"
 	"encoding/base64"
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"path"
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

PR #1266, opened 2024-02-15
materializations: various enhancements for handling problematic field names

williamhbaker:
**Description:**

Various enhancements for handling problematic field names in materializations that manage the equivalent of "columns" in their destination systems.

A few highlights:
- `materialize-elasticsearch` will sanitize field names by appending `_flow` to them, in cases where they conflict with ES pre-deefined "Metadata" fields and would not otherwise be able to be materialized
- `materialize-databricks` will sanitize field names that contain one of the set of characters that are always disallowed for column names, no matter how hard to you try to quote or escape them
- `materialize-mysql` will sanitize field names under some less common circumstances...characters with high code points, and trailing spaces
- Fixed a systemic issue with escaping identifier quotes within quoted identifiers for SQL materializations
- Better handling for "ambiguous" field names, where the destination system preserves capitalization but still doesn't allow columns to co-exist that differ only in capitalization - this is actually pretty common

I also did some manual testing with the [absurd schema](https://github.com/estuary/connectors/pull/1266/files#diff-c8390ca516f0802c9eda06a86c34a70193e5bbd6fed0b3bb2588e3a3b138b434R1-R48) used for the validate & apply tests to make sure that we could actually materialize data to tables with these columns for all materializations (including storing and loading). And I did a bonus test where a resource path was configured with a table name of `" ,;{}().- problematic#Table � 𐀀 嶲 "` and all materializations worked with that except `materialize-mysql`, which may suggest that there is still some limitation with how MySQL itself or the client driver handles weird characters. It doesn't seem worth spending a ton of time trying to figure out why this is right now, but we'll probably have to handle it eventually.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1266)


materialize-elasticsearch/driver.go
```diff
@@ -304,19 +305,24 @@ func normalizeIndexName(index string, truncateLimit int) string {
...
-		// Strip disallowed prefixes that may be present in a Flow collection name. A '.' is also a
-		// bad prefix, but those were already replaced with '_' above.
-		if !afterPrefix && (r == '-' || r == '_') {
-			continue
+		// Replace disallowed characters with underscore as long as we aren't still at the beginning
+		// of the string. Most of the characters in this list are named in the docs, but ES will
+		// also truncate index names that contain a '#' character to drop everything including &
+		// after that '#' character, so we'll normalize those too.
+		if slices.Contains([]rune{'*', '<', '"', ' ', '\\', '/', ',', '|', '>', '?', ':', '#'}, r) {
+			if !afterPrefix {
```
mdibaiee, line 318: nit: is this on a separate line due to line length?

materialize-elasticsearch/driver_test.go
```diff
@@ -118,7 +118,13 @@ func TestNormalizeIndexName(t *testing.T) {
 			name:       "dots",
 			byteLength: maxByteLength,
 			input:      ".some..collection",
-			want:       "some__collection",
+			want:       "some..collection",
```
mdibaiee, line 121: shouldn't this be `some__collection`?

materialize-snowflake/sqlgen.go
```diff
@@ -112,14 +112,16 @@ var snowflakeDialect = func(configSchema string) sql.Dialect {
...
 		Literaler: sql.LiteralFn(sql.QuoteTransform("'", "''")),
 		Placeholderer: sql.PlaceholderFn(func(_ int) string {
 			return "?"
 		}),
-		TypeMapper:      mapper,
-		ColumnValidator: columnValidator,
+		TypeMapper:             mapper,
+		ColumnValidator:        columnValidator,
+		MaxColumnCharLength:    255,
+		CaseInsensitiveColumns: false,
```
mdibaiee, line 124: I'm under the impression that in Snowflake unless the identifier is quoted, the fields will be case-insensitive. Does that not mean we should set this to `true`?

materialize-elasticsearch/driver.go
williamhbaker: It's a separate condition, since index names can't begin with an underscore and if a proposed index name has a disallowed character at the beginning it won't be valid to replace it with an underscore that would be at the beginning.

But looking at this now I think there's a simpler way to factor this that removes the second nested check of `if !afterPrefix` so I'll push that change.

materialize-snowflake/sqlgen.go
williamhbaker: `CaseInsensitiveColumns` (which is a proxy for `caseInsensitiveFields` in the `materialize-boilerplate` validator) means that even if the fields are quoted, the destination system will still not allow more than one column with the same name that differs only in capitalization. Snowflake does allow `"thisField"` vs `"thisfield"`, so having it be `false` represents that. Admittedly I'm finding it hard to come up with a variable name that succinctly represents what this toggle does, so let me know if you have suggestions on something that would be more clear.

mdibaiee (approved):

PR #1267, opened 2024-02-15
materialize-snowflake: use snowpipe for delta update bindings + jwt auth

mdibaiee:
**Description:**

- Support a new form of authentication using a username and a private key. We generate a JWT token and sign it using the private key to authenticate with Snowflake and Snowpipe REST API
- When using the JWT authentication mode, delta updates bindings are processed using Snowpipe
- Tested erroring out at the end of Acknowledge phase, and confirmed that we maintain idempotency by not replacing the pipe before sending requests again to the pipe.

One part that I still need to test: I need to test missing files again, I think given that snowpipe requires explicit specification of file names, it might break if we delete some files and then end up retrying those files. Need to verify

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1267)

### DIFF
--- a/materialize-snowflake/config.go
+++ b/materialize-snowflake/config.go
@@ -59,11 +59,11 @@ func (c *config) ToURI(tenant string) string {
 	return dsn
 }
 
-func (c *config) privateKey() (*rsa.PrivateKey, error) {
-	if c.Credentials.AuthType == JWT {
+func (c *credentialConfig) privateKey() (*rsa.PrivateKey, error) {
+	if c.AuthType == JWT {
 		// When providing the PEM file in a JSON file, newlines can't be specified unless
 		// escaped, so here we allow a escape hatch to parse these PEM files
-		var pkString = strings.ReplaceAll(c.Credentials.PrivateKey, "\\n", "\n")
+		var pkString = strings.ReplaceAll(c.PrivateKey, "\\n", "\n")
 		var block, _ = pem.Decode([]byte(pkString))
 		if key, err := x509.ParsePKCS8PrivateKey(block.Bytes); err != nil {
 			return nil, fmt.Errorf("parsing private key: %w", err)
@@ -104,7 +104,7 @@ func (c *config) asSnowflakeConfig(tenant string) sf.Config {
 		conf.Authenticator = sf.AuthTypeJwt
 		// We run this as part of validate to ensure that there is no error, so
 		// this is not expected to error here
-		if key, err := c.privateKey(); err != nil {
+		if key, err := c.Credentials.privateKey(); err != nil {
 			panic(err)
 		} else {
 			conf.PrivateKey = key
@@ -167,10 +167,6 @@ func (c *config) Validate() error {
 		return err
 	}
 
-	if _, err := c.privateKey(); err != nil {
-		return err
-	}
-
 	return validHost(c.Host)
 }
 
@@ -211,6 +207,10 @@ func (c *credentialConfig) validateUserPassCreds() error {
 func (c *credentialConfig) validateJWTCreds() error {
 	if c.PrivateKey == "" {
 		return fmt.Errorf("missing private_key")
+	}
+
+	if _, err := c.privateKey(); err != nil {
+		return err
 	}
 
 	return nil
--- a/materialize-snowflake/config.go
+++ b/materialize-snowflake/config.go
@@ -62,7 +62,7 @@ func (c *config) ToURI(tenant string) string {
 func (c *credentialConfig) privateKey() (*rsa.PrivateKey, error) {
 	if c.AuthType == JWT {
 		// When providing the PEM file in a JSON file, newlines can't be specified unless
-		// escaped, so here we allow a escape hatch to parse these PEM files
+		// escaped, so here we allow an escape hatch to parse these PEM files
 		var pkString = strings.ReplaceAll(c.PrivateKey, "\\n", "\n")
 		var block, _ = pem.Decode([]byte(pkString))
 		if key, err := x509.ParsePKCS8PrivateKey(block.Bytes); err != nil {
@@ -161,6 +161,7 @@ func (c *config) Validate() error {
 
 		c.Credentials.AuthType = UserPass
 		c.Credentials.Password = c.Password
+		c.Credentials.User = c.User
 	}
 
 	if err := c.Credentials.Validate(); err != nil {
@@ -197,6 +198,9 @@ func (c *credentialConfig) Validate() error {
 }
 
 func (c *credentialConfig) validateUserPassCreds() error {
+	if c.User == "" {
+		return fmt.Errorf("missing user")
+	}
 	if c.Password == "" {
 		return fmt.Errorf("missing password")
 	}
@@ -205,6 +209,9 @@ func (c *credentialConfig) validateUserPassCreds() error {
 }
 
 func (c *credentialConfig) validateJWTCreds() error {
+	if c.User == "" {
+		return fmt.Errorf("missing user")
+	}
 	if c.PrivateKey == "" {
 		return fmt.Errorf("missing private_key")
 	}
--- a/materialize-snowflake/config.go
+++ b/materialize-snowflake/config.go
@@ -1,0 +1,284 @@
+package main
+
+import (
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"github.com/iancoleman/orderedmap"
+	"github.com/invopop/jsonschema"
+	"regexp"
+	"strings"
+
+	m "github.com/estuary/connectors/go/protocols/materialize"
+	sf "github.com/snowflakedb/gosnowflake"
+)
+
+// config represents the endpoint configuration for snowflake.
+// It must match the one defined for the source specs (flow.yaml) in Rust.
+type config struct {
+	// TODO(mahdi): the Host and Account config is very confusing since Snowflake has multiple ways and very specific requirements
+	// for specifying the url of an instance. Ideally we should just accept the URL directly and use that, that would save us and
+	// users some headache
+	// See: https://docs.snowflake.com/en/user-guide/admin-account-identifier#non-vps-account-locator-formats-by-cloud-platform-and-region
+	Host      string `json:"host" jsonschema:"title=Host URL,description=The Snowflake Host used for the connection. Must include the account identifier and end in .snowflakecomputing.com. Example: orgname-accountname.snowflakecomputing.com (do not include the protocol)." jsonschema_extras:"order=0,pattern=^[^/:]+.snowflakecomputing.com$"`
+	Account   string `json:"account" jsonschema:"title=Account,description=The Snowflake account identifier." jsonschema_extras:"order=1"`
+	User      string `json:"user" jsonschema:"-"`
+	Password  string `json:"password" jsonschema:"-"`
+	Database  string `json:"database" jsonschema:"title=Database,description=The SQL database to connect to." jsonschema_extras:"order=4"`
+	Schema    string `json:"schema" jsonschema:"title=Schema,description=Database schema for bound collection tables (unless overridden within the binding resource configuration) as well as associated materialization metadata tables." jsonschema_extras:"order=5"`
+	Warehouse string `json:"warehouse,omitempty" jsonschema:"title=Warehouse,description=The Snowflake virtual warehouse used to execute queries. Uses the default warehouse for the Snowflake user if left blank." jsonschema_extras:"order=6"`
+	Role      string `json:"role,omitempty" jsonschema:"title=Role,description=The user role used to perform actions." jsonschema_extras:"order=7"`
+
+	Credentials credentialConfig `json:"credentials" jsonschema:"title=Authentication"`
+
+	Advanced advancedConfig `json:"advanced,omitempty" jsonschema:"title=Advanced Options,description=Options for advanced users. You should not typically need to modify these." jsonschema_extras:"advanced=true"`
+}
+
+type advancedConfig struct {
+	UpdateDelay string `json:"updateDelay,omitempty" jsonschema:"title=Update Delay,description=Potentially reduce active warehouse time by increasing the delay between updates. Defaults to 30 minutes if unset.,enum=0s,enum=15m,enum=30m,enum=1h,enum=2h,enum=4h"`
+}
+
+// ToURI converts the Config to a DSN string.
+func (c *config) ToURI(tenant string) string {
+	// Build a DSN connection string.
+	var configCopy = c.asSnowflakeConfig(tenant)
+	// client_session_keep_alive causes the driver to issue a periodic keepalive request.
+	// Without this, the authentication token will expire after 4 hours of inactivity.
+	// The Params map will not have been initialized if the endpoint config didn't specify
+	// it, so we check and initialize here if needed.
+	if configCopy.Params == nil {
+		configCopy.Params = make(map[string]*string)
+	}
+	configCopy.Params["client_session_keep_alive"] = &trueString
+	dsn, err := sf.DSN(&configCopy)
+	if err != nil {
+		panic(fmt.Errorf("building snowflake dsn: %w", err))
+	}
+
+	return dsn
+}
+
+func (c *config) privateKey() (*rsa.PrivateKey, error) {
+	if c.Credentials.AuthType == JWT {
+		// When providing the PEM file in a JSON file, newlines can't be specified unless
+		// escaped, so here we allow a escape hatch to parse these PEM files
+		var pkString = strings.ReplaceAll(c.Credentials.PrivateKey, "\\n", "\n")
+		var block, _ = pem.Decode([]byte(pkString))
+		if key, err := x509.ParsePKCS8PrivateKey(block.Bytes); err != nil {
+			return nil, fmt.Errorf("parsing private key: %w", err)
+		} else {
+			return key.(*rsa.PrivateKey), nil
+		}
+	}
+
+	return nil, fmt.Errorf("only supported with JWT authentication")
+}
+
+func (c *config) asSnowflakeConfig(tenant string) sf.Config {
+	var maxStatementCount string = "0"
+	var json string = "json"
+
+	var conf = sf.Config{
+		Account:     c.Account,
+		Host:        c.Host,
+		Database:    c.Database,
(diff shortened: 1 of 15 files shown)


materialize-snowflake/config.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	return dsn
+}
+
+func (c *config) privateKey() (*rsa.PrivateKey, error) {
+	if c.Credentials.AuthType == JWT {
+		// When providing the PEM file in a JSON file, newlines can't be specified unless
+		// escaped, so here we allow a escape hatch to parse these PEM files
+		var pkString = strings.ReplaceAll(c.Credentials.PrivateKey, "\\n", "\n")
```
mdibaiee, line 66: I have checked that in the UI when we submit a multi-line secret through the upload form, newlines are present as `\n` in the string, so this should work with the UI form

materialize-snowflake/sqlgen.go
```diff
@@ -266,6 +289,14 @@ WHEN NOT MATCHED THEN
 		r.{{$key.Identifier -}}
 	{{- end -}}
 );
+{{ end }}
+
+{{ define "copyHistory" }}
+SELECT FILE_NAME, STATUS, FIRST_ERROR_MESSAGE FROM TABLE(INFORMATION_SCHEMA.COPY_HISTORY(
+  TABLE_NAME=>'{{ $.Table.Identifier }}',
+  START_TIME=>DATEADD(MINUTE, -5, TO_TIMESTAMP_LTZ('{{ $.StartTime }}'))
```
mdibaiee, line 297: Even though I am fetching the start time from Snowflake itself, and I'm doing that before we even begin sending the pipe requests, I hit this issue a few times where the start time ended up being after the start time of the pipe load requests we sent... so I ended up with this -5 minutes. It's a bit gross though.

materialize-snowflake/snowflake.go
```diff
@@ -518,31 +485,75 @@ func (d *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
...
+			pipes[item.PipeName] = &pipeRecord{
+				files:     item.PipeFiles,
+				dir:       item.StagedDir,
+				binding:   d.bindingByStateKey(stateKey),
+				startTime: item.PipeStartTime,
+			}
 		}
 	}
 
+	//var hasPipes = len(pipes) > 0
```
mdibaiee, line 555: I use this along with a line below to test errors during ack phase

materialize-snowflake/snowflake.go
```diff
@@ -553,6 +564,113 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
...
+					if err := rows.Scan(&fileName, &status, &firstErrorMessage); err != nil {
+						return nil, fmt.Errorf("scanning copy history row: %w", err)
+					}
+
+					log.WithFields(log.Fields{
+						"fileName":          fileName,
+						"status":            status,
+						"firstErrorMessage": firstErrorMessage,
+						"pipeName":          pipeName,
+					}).Info("snowpipe: copy history row")
```
mdibaiee, line 622: Since this is not supposed to happen frequently, I think it's fine to leave this as info so we can troubleshoot when this happens

materialize-snowflake/snowflake.go
```diff
@@ -573,6 +691,10 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
 		return nil, fmt.Errorf("creating checkpoint clearing json: %w", err)
 	}
 
+	/*if hasPipes {
+		return nil, fmt.Errorf("artificial error")
+	}*/
```
mdibaiee, line 696: left here for testing purposes, will clean up before merging

materialize-snowflake/pipe.go
```diff
@@ -0,0 +1,267 @@
...
+		return nil, err
+	}
+
+	var claims = &jwt.RegisteredClaims{
+		IssuedAt: jwt.NewNumericDate(time.Now()),
+		// TODO: automatically refresh the JWT token
+		// JWT tokens for Snowflake can live up to an hour
+		ExpiresAt: jwt.NewNumericDate(time.Now().Add(59 * time.Minute)),
+		Issuer:    fmt.Sprintf("%s.%s", qualifiedUser, fingerprint),
+		Subject:   qualifiedUser,
```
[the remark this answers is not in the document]
williamhbaker: Pondering this some more: I'd really like to get rid of the `account` configuration input completely, and just require the `host` config input. It seems like you actually do need the `account` here, and it would be nice if the [`CURRENT_ACCOUNT_NAME`](https://docs.snowflake.com/en/sql-reference/functions/current_account_name) (or maybe `CURRENT_ACCOUNT`?) function could be used for getting that with the `host`-based connection, rather than requiring it as configuration.

williamhbaker (commented):
Made a few comments & questions - let me know what you think, and I'll take another look after you've made updates that make sense.

materialize-snowflake/config.go
```diff
@@ -0,0 +1,300 @@
...
+// It must match the one defined for the source specs (flow.yaml) in Rust.
+type config struct {
+	// TODO(mahdi): the Host and Account config is very confusing since Snowflake has multiple ways and very specific requirements
+	// for specifying the url of an instance. Ideally we should just accept the URL directly and use that, that would save us and
+	// users some headache
+	// See: https://docs.snowflake.com/en/user-guide/admin-account-identifier#non-vps-account-locator-formats-by-cloud-platform-and-region
+	Host      string `json:"host" jsonschema:"title=Host URL,description=The Snowflake Host used for the connection. Must include the account identifier and end in .snowflakecomputing.com. Example: orgname-accountname.snowflakecomputing.com (do not include the protocol)." jsonschema_extras:"order=0,pattern=^[^/:]+.snowflakecomputing.com$"`
+	Account   string `json:"account" jsonschema:"title=Account,description=The Snowflake account identifier." jsonschema_extras:"order=1"`
+	User      string `json:"user" jsonschema:"-"`
+	Password  string `json:"password" jsonschema:"-"`
```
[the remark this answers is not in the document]
mdibaiee: The default rendering of the form is:
image

Since credentials is required, it is "open" by default in the form and the user must fill them before they can submit the form, so I think it should be clear for our users that they need to enter their user and password here now. That said, a heads up won't hurt!

materialize-snowflake/pipe.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+type InsertFilesResponse struct {
+	Status    string `json:"responseCode"`
+	RequestId string `json:"requestId"`
+}
+
+const insertFilesRawTpl = "https://{{ $.Base }}/v1/data/pipes/{{ $.PipeName }}/insertFiles?requestId={{ $.RequestId }}"
+const contentType = "application/json;charset=UTF-8"
+const userAgent = "Estuary Technologies Flow"
```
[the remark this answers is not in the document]
mdibaiee: I don't think it is necessary, I'm mostly providing it for troubleshooting purposes if at some point Snowlake support needs to be involved, so they can find our requests. I don't think adding the tenant name to the user agent would really help with that purpose, since the tenant is implicit in the URL of the request

materialize-snowflake/pipe.go
```diff
@@ -0,0 +1,267 @@
...
+		insertFilesTpl:  insertFilesTpl,
+		insertReportTpl: insertReportTpl,
+	}, nil
+}
+
+type insertFilesURLTemplate struct {
+	Base     string
+	PipeName string
+	// Used to track the request, we generate a random uuid for this
+	RequestId string
```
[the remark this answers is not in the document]
mdibaiee: the request id is useful for troubleshooting when contacting support. The thought is if we hit some weird issue that we can't understand / need help from Snowflake support for, we can give them the request id (which is logged in our debug level) so they can find the request we are talking about

materialize-snowflake/pipe.go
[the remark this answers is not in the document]
williamhbaker: We do get credit for the Snowflake partner program for including the `Application: <TenantName>_EstuaryFlow` parameter to the SQL connection string. If there was a way to set that for these API requests that would be good, since that will be the primary mechanism that credits are consumed with delta updates + snowpipe. No idea if there's actually a way to do that or not, but it would be good to know.

materialize-snowflake/sqlgen.go
```diff
@@ -266,6 +290,14 @@ WHEN NOT MATCHED THEN
 		r.{{$key.Identifier -}}
 	{{- end -}}
 );
+{{ end }}
+
+{{ define "copyHistory" }}
+SELECT FILE_NAME, STATUS, FIRST_ERROR_MESSAGE FROM TABLE(INFORMATION_SCHEMA.COPY_HISTORY(
+  TABLE_NAME=>'{{ $.Table.Identifier }}',
+  START_TIME=>DATEADD(MINUTE, -30, TO_TIMESTAMP_LTZ('{{ $.StartTime }}'))
```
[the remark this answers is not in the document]
mdibaiee: This makes sense, I switched to always searching the last 14 days, but I also checked the behaviour in case of asking for a `START_TIME` that's more than 14 days ago, and there is no error, just that no records older than 14 days will be present in the response.

materialize-snowflake/sqlgen.go
```diff
@@ -212,6 +215,27 @@ SELECT * FROM (SELECT -1, CAST(NULL AS VARIANT) LIMIT 0) as nodoc
 {{ end -}}
 {{ end }}
 
+{{ define "pipe_name" -}}
```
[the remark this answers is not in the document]
mdibaiee: I think ideally we would have a single pipe among different shards, since that would allow optimal utilisation of the pipe's queue. The main issue here is that when we `CREATE OR REPLACE` a pipe, we will clobber up the pipe that might be used by another shard...

It is not possible to change the `COPY INTO` statement of a pipe, which means it is not possible to _update_ a pipe in response to changes in projections of a collection, we need to recreate the pipe.

Re-creating a pipe before we run `Acknowledge` would break the `Acknowledge` recovery phase, this means we can't create the pipe during `Apply` phase as we create the tables.

We are creating pipes during `Store` phase, and in this phase I don't think there is a way for us to ensure recreation of the pipe without risking clobbering some other shard's work. Let me know if you can think of a way to do this.

With that, we may have to create shard-specific pipes similar to what we do in Databricks 🤔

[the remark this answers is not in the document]
williamhbaker: I also think that ideally there'd be a single pipe among different shards. This concept of coordination could be optimal in other ways too, such as a standard updates materialization where there could be sharded "workers" that only serialize data and upload files and then the "leader" (the one with responsible for the key range starting at 0 perhaps) is the one that issues the `MERGE` queries, rather than each shard queuing up its separate, smaller merge queries on the same table concurrently.

I could imagine achieving this using a table in Snowflake to coordinate between shards, somehow...it would take some thought. For a standard updates materialization this would make more sense as it would involve waking up the warehouse and they'd do that anyway, but a purely snowpipe materialization not so much. I don't think I know enough about how existing transactions are coordinated between shards to say much more. I do recall that there is some element of them committing all around the same time. So maybe there's something we could do there, and use driver checkpoints in a clever way.

It would seem fine for now to use shard-specific pipes for the sake of getting this done (and sharded materializations are not really functional atm for other reasons), although could be something interesting to discuss more.

materialize-snowflake/snowflake.go
```diff
@@ -553,6 +561,120 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
...
+			}
+
+			var hasItemsInProgress = false
+
+			// If the files have already been loaded, when we submit a request to
+			// load those files again, our request will not show up in reports.
+			// One way to find out whether the files were successfully loaded is to check
+			// the COPY_HISTORY to make sure they are there. If they are not, then something
+			// is wrong.
+			if len(report.Files) == 0 {
```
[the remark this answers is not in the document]
mdibaiee: I can't quite trace the sequence you are describing with the code. There has been some refactoring now, and in this factoring we do not allow any progress until all files have been successfully processed. If any of the files have an error we bail out. On subsequent runs of the connector, we still wait until we find results for all the filenames that we have in the checkpoint, which includes the failed ones. If they don't show up in the report for some reason, we look for it among the copyHistory results and only continue if all the files we expect are found

materialize-snowflake/pipe.go
```diff
@@ -0,0 +1,311 @@
...
+	req.Header.Add("Authorization", fmt.Sprintf("BEARER %s", c.token))
+	req.Header.Add("Content-Type", contentType)
+	req.Header.Add("User-Agent", userAgent)
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("insertFiles request: %w", err)
+	}
+
+	var respBuf = new(strings.Builder)
```
williamhbaker, line 190: Decoding the response here could be simplified using something like this:

```go
	resp, err := c.httpClient.Do(req)
	if err != nil {
		return nil, fmt.Errorf("insertFiles request: %w", err)
	}

	var response InsertFilesResponse
	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
		return nil, fmt.Errorf("parsing response of insertFiles failed: %w", err)
	}
	resp.Body.Close()
```

materialize-snowflake/pipe.go
```diff
@@ -0,0 +1,311 @@
...
+	req.Header.Add("Content-Type", contentType)
+	req.Header.Add("User-Agent", userAgent)
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("insertReport request: %w", err)
+	}
+
+	var respBuf = new(strings.Builder)
+	_, err = io.Copy(respBuf, resp.Body)
```
williamhbaker, line 289: See above about decoding the response using a `json.NewDecoder`

materialize-snowflake/pipe.go
mdibaiee: I intentionally first write to a string buffer and then decode, since in case of non-200 status code we use the response body as part of the error message.

materialize-snowflake/snowflake.go
```diff
@@ -552,6 +611,108 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
...
+					} else {
+						return nil, fmt.Errorf("unexpected status %q for files in pipe %q: %s", row.status, pipeName, row.firstErrorMessage)
+					}
+				}
+
+				// If items are still in progress, we continue trying to fetch their results
+				if hasItemsInProgress {
+					tries--
+					time.Sleep(2 * time.Second)
+					continue
```
williamhbaker, line 681: I think these means we'll loop around and try the REST API again. Since the REST API will have demonstrated that it isn't going to give us anything by this point, would it make sense to retry querying `COPY_HISTORY` again (after a delay) instead?

materialize-snowflake/snowflake.go
```diff
@@ -552,6 +611,108 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
...
+			// If there are items still in progress, we continue retrying until those items
+			// resolve to another status
+			var hasItemsInProgress = false
+
+			// If the files have already been loaded, when we submit a request to
+			// load those files again, our request will not show up in reports.
+			// One way to find out whether the files were successfully loaded is to check
+			// the COPY_HISTORY to make sure they are there. The COPY_HISTORY is much more
+			// reliable. If they are not there, then something is wrong.
+			if len(report.Files) == 0 {
```
williamhbaker, line 644: Is there a scenario where the REST API gives us a partial list of files because some of the files have timed out of its retention, but some haven't? Is that covered by `report.CompleteResult`? I think we'd want to try `COPY_HISTORY` in that case.

williamhbaker (commented):
A couple more questions, mostly on the control flow for handling REST API responses and COPY_HISTORY queries. I think I'm better understanding what's going on from the refactored code, but am not yet totally confident that I fully grok it.

materialize-snowflake/snowflake.go
mdibaiee: @williamhbaker we will retry the REST API, and if it doesn't give us any results again, we will try the COPY_HISTORY again. Note that `tries--` will mean that on the next iteration we will still run `COPY_HISTORY` if no files are given

materialize-snowflake/snowflake.go
```diff
@@ -552,6 +611,108 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
...
+				if len(pipe.files) > 0 {
+					return nil, fmt.Errorf("snowpipe: could not find reports of successful processing for all files of pipe %v", pipe)
+				}
+
+				// All files have been processed for this pipe, we can skip to the next pipe
+				d.deleteFiles(ctx, []string{pipe.dir})
+				break
+			}
+
+			tries = 0
```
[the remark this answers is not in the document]
mdibaiee: The comparison below is extra. The way I have it in my mind is this: we will loop forever to find items using the REST API, and if we start getting no results from REST API, then we retry `maxTries` times again asking the REST API, and if that doesn't give us results, we then ask `COPY_HISTORY`.

If we do get new results from the REST API, we reset the counter for the retries necessary to prompt us to ask `COPY_HISTORY`.

materialize-snowflake/snowflake.go
mdibaiee: In that scenario eventually all files will time out and go out of retention period of REST API (10 minutes maximum), and then will ask `COPY_HISTORY`. It is very likely that files that we are checking as part of the same transaction will all time out close to each other, so if some of them have timed out, all of them are expected to time out soon.

mdibaiee: Also note that, once we have seen the partial list of files, we will persist the `beginMarker`, which means on the next run of `insertReport`, we will not get the same partial list of files. We will get a list of no values if the other files have expired, so that expediates our falling back to `COPY_HISTORY`.

williamhbaker (approved):
LGTM, thanks!

materialize-snowflake/snowflake.go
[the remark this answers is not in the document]
mdibaiee: I added a bunch of comments to explain this a bit more. It is a complicated way of doing things.

PR #1268, opened 2024-02-16
materialize-elasticsearch: allow materializing string values as keywords

williamhbaker:
**Description:**

Adds the option to materialize string values as keywords via the field configuration in field selection. Previously we would only materialize string fields as keywords if they were part of the collection key.

I had thought about doing more comprehensive validation on this to disallow setting the "keyword" field config value for non-string fields, but that was surprisingly difficult with the way the code is currently structured. This is going to be a pretty advanced feature currently since the only way to edit the field config is through the JSON of the spec, so it seems alright for now to keep it a little unpolished for the sake for expedience.

**Workflow steps:**

Use a spec like this to materialize a string value field `stringField` as a `keyword` mapping instead of a `text` mapping:

```json
{
  "bindings": [
    {
      "resource": {
        "index": "myIndex",
        "delta_updates": false
      },
      "source": "prefix/source/collection",
      "fields": {
        "include": {
          "stringField": {
            "keyword": true
          }
        },
        "recommended": true
      }
    }
  ]
}
```

The `materialize-elasticsearch` docs should be updated according to this change, which I will do shortly.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1268)

### DIFF
--- a/materialize-elasticsearch/type_mapping.go
+++ b/materialize-elasticsearch/type_mapping.go
@@ -31,8 +31,8 @@ type property struct {
 	Index  *bool               `json:"index,omitempty"`
 }
 
-func propForField(field string, binding *pf.MaterializationSpec_Binding) property {
-	return propForProjection(binding.Collection.GetProjection(field))
+func propForField(field string, binding *pf.MaterializationSpec_Binding) (property, error) {
+	return propForProjection(binding.Collection.GetProjection(field), binding.FieldSelection.FieldConfigJsonMap[field])
 }
 
 var numericStringTypes = map[boilerplate.StringWithNumericFormat]elasticPropertyType{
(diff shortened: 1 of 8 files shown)


materialize-elasticsearch/type_mapping.go
```diff
@@ -145,30 +160,20 @@ func (constrainter) NewConstraints(p *pf.Projection, deltaUpdates bool) *pm.Resp
 	return &constraint
 }
 
-func (constrainter) Compatible(existing boilerplate.EndpointField, proposed *pf.Projection, _ json.RawMessage) (bool, error) {
-	prop := propForProjection(proposed).Type
-
-	if strings.EqualFold(existing.Type, string(elasticTypeText)) {
-		// Allow any of the "formatted string" types to be materialized into an existing text
```
williamhbaker, line 152: I'm removing this compatibility since it is not consistent with other materializations (ex: SQL) and I think it was an oversight in the first place. No ES materializations that I'm aware of are relying on this behavior right now.

mdibaiee (approved):

materialize-elasticsearch/type_mapping.go
mdibaiee: So our position now is that we don't want customers to bring their own table / index if it doesn't conform to the exact schema that we have for the collection? In case of keyword we are allowing this if they tell us about it in their field selection. So for further compatibility between these types we would need similar configurations in field selection?

williamhbaker: To the first question, yes, and this is really because we want re-backfills to be as unsurprising as possible. If a pre-existing index had a `text` mapping that was being materialized as a `format: xxxxx` string collection field, a re-backfill would re-create that mapping as the formatted type rather than the original `text` mapping, which could be surprising. Similar behavior exists in SQL materializations, where pre-existing text columns could by all rights be materialized into from `type: string, format: xxxxx` fields, but we prohibit that so re-backfills don't alter columns that were previously validated as compatible.

For the second question, we could perhaps formalize a field configuration that would ignore the format of strings, although I'm not sure if that's the right thing to do right now. This the `keyword` vs. `text` field config option is controlling behavior in the materialized system that we can't otherwise represent in the source collection schema via a `format` annotation.

PR #1269, opened 2024-02-16
source-*-batch: Implement "daily at HH:MMZ" polling schedules

willdonnelly:
**Description:**

This PR retcons the `"poll"` config property of the capture config and resource specs. Previously this property was a "polling interval" specified as a Go duration string, and now it's a "polling schedule" which accepts Go duration strings as a subset of the allowed specifications.

The upshot of that redefinition is that we can implement other types of polling schedule, most notably `"daily at 6:00Z"` is now a valid schedule description as well. This has the expected behavior -- regardless of restart or publication times or execution delays, we'll always do our best to poll the binding once a day at the specified time (modulo any delays which might arise from other ongoing polling operations).

**Workflow steps:**

If you type something like `"daily at 01:00Z"` in the `"Polling Schedule"` property of a resource binding, it will be polled at 1AM (UTC) every day.

This change is entirely undocumented at the moment, but at some point should be added to:

- https://docs.estuary.dev/reference/Connectors/capture-connectors/bigquery-batch/
- https://docs.estuary.dev/reference/Connectors/capture-connectors/mysql-batch/
- https://docs.estuary.dev/reference/Connectors/capture-connectors/postgres-batch/

### DIFF
--- a/source-bigquery-batch/driver.go
+++ b/source-bigquery-batch/driver.go
@@ -14,6 +14,7 @@ import (
 
 	"cloud.google.com/go/bigquery"
 	"github.com/estuary/connectors/go/encrow"
+	"github.com/estuary/connectors/go/schedule"
 	schemagen "github.com/estuary/connectors/go/schema-gen"
 	boilerplate "github.com/estuary/connectors/source-boilerplate"
 	pc "github.com/estuary/flow/go/protocols/capture"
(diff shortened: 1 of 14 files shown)


source-bigquery-batch/driver.go
```diff
@@ -49,7 +50,7 @@ type Resource struct {
 	Name         string   `json:"name" jsonschema:"title=Name,description=The unique name of this resource." jsonschema_extras:"order=0"`
 	Template     string   `json:"template" jsonschema:"title=Query Template,description=The query template (pkg.go.dev/text/template) which will be rendered and then executed." jsonschema_extras:"multiline=true,order=3"`
 	Cursor       []string `json:"cursor,omitempty" jsonschema:"title=Cursor Columns,description=The names of columns which should be persisted between query executions as a cursor." jsonschema_extras:"order=2"`
-	PollInterval string   `json:"poll,omitempty" jsonschema:"title=Poll Interval,description=How often to execute the fetch query (overrides the connector default setting)" jsonschema_extras:"order=1"`
+	PollSchedule string   `json:"poll,omitempty" jsonschema:"title=Polling Schedule,description=When and how often to execute the fetch query (overrides the connector default setting)" jsonschema_extras:"order=1"`
```
williamhbaker, line 53: Might be nice to have an example of the different inputs you can use here - also, if possible, a regexp to provide feedback in the UI if the string is valid, although I don't really know if that's practical.

williamhbaker (approved):
LGTM

source-bigquery-batch/driver.go
willdonnelly: Done. If we're lucky the regexp might even be correct!

PR #1270, opened 2024-02-16
fix: Python packages broke when introducing `source-hubspot`, this fixes all python packages

jshearer:
**Description:**

Because `source-python` doesn't use a sub-package like all other python packages do, it needed different treatment. This different treatment broke everything, so this PR undoes it and fixes the problem a different way.

tl;dr, the upper `source-python` package was a special [namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/), and Airbyte CDK was using a specific function to load the `spec.yaml` file that [doesn't work on namespace packages](https://bugs.python.org/issue28424), apparently. So the fix ended up being to move `spec.yaml` up a directory and add a blank `__init__.py` to `source-python` to make it not a namespace package... k ¯\_(ツ)_/¯

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1270)


python/Dockerfile
```diff
@@ -59,7 +59,6 @@ WORKDIR /connector
 COPY --from=builder /builder/connector ./$CONNECTOR_NAME
 COPY --from=builder /builder/python ./python
 
-ENV PYTHONPATH="/connector/$CONNECTOR_NAME"
```
jshearer, line 62: This is the change that broke things -- all python packages (except `source-hubspot`) are now complaining that 

```
File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/connector/source-airtable/__main__.py", line 2, in <module>
    from source_airtable import SourceAirtable
ImportError: cannot import name 'SourceAirtable' from 'source_airtable' (unknown location)
```

dyaffe (approved):
LGTM

Issue #1253, opened 2024-02-08
materialize-bigquery: better handling for partitioned tables


williamhbaker:
- Add docs about partitioned tables and how they can be pre-created or created from already existing tables, since partitioning can't be added after the fact and the connector doesn't automatically partition tables
- Make validation fail if a backfill counter is incremented on a partitioned table, since that won't work

See https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1707402955934499

PR #1274, opened 2024-02-20
materializations: always delete resources before re-creating them

williamhbaker:
**Description:**

This gets rid of the "replace resource" action of materializations in favor of a more simple "delete resource" action, which is then always followed by a "create resource" to accomplish a resource replacement.

Only a few systems could really take advantage of the "replace resource" action, by doing a "CREATE OR REPLACE TABLE" type of thing, or a "DROP TABLE" followed by "CREATE TABLE" with transactional control. The "CREATE OR REPLACE TABLE" statement was actually not fully functional for BigQuery, which does not allow a table to be replaced with different partitioning or clustering, which was making it impossible to change the keys of a collection and re-backfill a table.

Rather than continue to maintain the excess complexity of the "replace resource" that is not generally applicable, we'll just do a simple "delete" on the resource and then re-use the "create" resource action that already has to be defined. With the way materialization `Apply` now works, there's not much justification for allowing finer control over these actions, for example dropping a table and then failing to re-create it for some reason will be resolved on the next `Apply` when the materialization sees that the table doesn't exist and creates it anew anyway.

Closes https://github.com/estuary/connectors/issues/1253

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1274)


jgraettinger (approved):
LGTM ❤️

PR #1276, opened 2024-02-21
Phil/webhook streams

psFried:
**Description:**

Discussed in [this slack thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1708529350478799), the idea is to allow our webhook connector to more easily work in scenarios where you want to accept webhooks at multiple different paths and bind each one to a different collection.

This also rolls up a few other commits, most notably adding support for capturing url query parameters. I recommend a commit by commit review.

**Workflow steps:**

- Existing captures should be unaffected, as long as their endpoint config doesn't get edited. I tested upgrading an existing task and confirmed that discovers still return the same results.
- For either new captures (or editing the endpoint config of an existing one), the `paths` property of the connector determines which collections will be "discovered".
- If `paths` is empty, then discovers will return a binding that matches what the connector outputs today, thus preserving compatibility
- If `paths` is non-empty, then discovers will output a collection spec for each path item, with the `path` and `stream` properties set appropriately. These will _not_ match what the connector outputs today.

Still working on some updates to [this page](https://docs.estuary.dev/reference/Connectors/capture-connectors/http-ingest/), but should have a PR ready shortly.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1276)

### DIFF
--- a/source-http-ingest/Cargo.lock
+++ b/source-http-ingest/Cargo.lock
@@ -379,7 +379,7 @@ dependencies = [
 [[package]]
 name = "doc"
 version = "0.0.0"
-source = "git+https://github.com/estuary/flow#91e9f39aa5ac11bb55c41861bd0bddd398d8fe22"
+source = "git+https://github.com/estuary/flow#d7cb2f8b2f459cbf2950eee491a37cb50c42ca57"
 dependencies = [
  "base64 0.13.1",
  "bumpalo",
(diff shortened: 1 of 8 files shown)


williamhbaker (approved):
LGTM

Mind bumping the `h2` version while you're in here? Dependabot has a PR open for that [here](https://github.com/estuary/connectors/pull/1194) and I've hesitated to blindly merge it without any kind of testing.

PR #1277, opened 2024-02-22
source-mongodb: log buildInfo on startup

williamhbaker:
**Description:**

Log some basic information about the database that is being connected to, including its version and maximum object size setting, if we are able to.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1277)

### DIFF
--- a/source-mongodb/pull.go
+++ b/source-mongodb/pull.go
@@ -68,6 +68,17 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
 	if err != nil {
 		return fmt.Errorf("connecting to database: %w", err)
 	}
+
+	// Log some basic information about the database we are connected to if possible. buildInfo is
+	// an administrator command so it may not be available on all databases / configurations.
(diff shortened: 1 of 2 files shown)


source-mongodb/pull.go
```diff
@@ -797,3 +809,19 @@ func idToString(value interface{}) string {
 	}
 	return string(j)
 }
+
+type buildInfo struct {
+	Version       string `bson:"version"`
+	MaxObjectSize int    `bson:"maxBsonObjectSize"`
```
mdibaiee, line 815: This is always 16MB, no? From their docs:

> BSON Document Size
> The maximum BSON document size is 16 megabytes.
https://www.mongodb.com/docs/manual/reference/limits/#bson-documents

williamhbaker: On extremely old versions it was 4MB, but those are probably so old they would have other problems running. We'd get the size limit on any "document too large" errors, so I'll remove this from the logging.

PR #1278, opened 2024-02-24
materializations: better unsatisfiable constraint validation messages

williamhbaker:
**Description:**

The changes materializations to provide more actionable messages when producing unsatisfiable constraints.

Particularly for SQL materializations, which will currently report the existing column type and proposed JSON schema of the field, the message will now also report the type of column that is needed to accommodate the proposed JSON schema.

Overall the objective is to make it easier to understand how a user could migrate their existing table to be compatible with a changed collection schema.

**Workflow steps:**

Previously a constraint message for an unsatisfiable field change would look like this:
```
Field 'boolField' is already being materialized as endpoint type 'boolean' and cannot be changed to type '[number,string] format: integer content-encoding: base64'
```

Now, it will look like this:
```
Field 'boolField' is already being materialized as endpoint type 'BOOLEAN' but endpoint type 'JSON' is required by its schema '{ type: [number, string], format: integer, content-encoding: base64 }'
```

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1278)

### DIFF
--- a/materialize-bigquery/.snapshots/TestValidateAndApply
+++ b/materialize-bigquery/.snapshots/TestValidateAndApply
@@ -79,24 +79,24 @@ Big Schema Re-validated Constraints:
 Big Schema Changed Types Constraints:
 {"Field":"_meta/flow_truncated","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
 {"Field":"arrayField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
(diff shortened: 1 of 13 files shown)


mdibaiee (commented):
LGTM, but this has broken a lot of tests it seems, need to update the snapshots I think

williamhbaker:
> LGTM, but this has broken a lot of tests it seems, need to update the snapshots I think

Got the snapshots updated - also a few small fixes to `materialize-starburst` so that the full validate and apply set of tests could run.

mdibaiee (approved):

PR #1288, opened 2024-02-27
Fix Portifolio and Event streams on Asana connector

Z33DD:
**Description:**

According to [this issue](https://github.com/airbytehq/airbyte/pull/31114) and [this one](https://github.com/airbytehq/airbyte/pull/31114), those streams will throw errors if the account doesn't pay for an Asana's paid plan.
The problem with the events' stream is the different pagination style that demands the first request to return a 'Precondition Failed' failed error (412).

The main issue is the lack of error handling for 402 Payment Required errors and the missing `owner` query parameter in the request. While the Asana API reference says `owner` is an optional parameter, the actual API returns an error when it is not present in the request.

About the Events stream, it is broken because of pagination, so it is not related to the paid account problem. I tested this flow with both paid and free accounts, the error is the `sync` token. I should fix this in a further PR.

Streams to fix:
- [x] portfolios
- [x] portfolios_compact
- [x] portfolios_memberships

Asana API Reference:
- [Events](https://developers.asana.com/reference/getevents)
- [Portfolios](https://developers.asana.com/reference/getportfolios)

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1288)


jshearer (approved):
LGTM, Thanks for making the changes we talked about! :)

PR #1289, opened 2024-02-27
source-redshift-batch: Initial implementation

willdonnelly:
**Description:**

This PR adds a new connector `source-redshift-batch` which is a gratuitous copy-paste of `source-postgres-batch` [1] with various things renamed. Most notably the default query template and polling interval have been modified to use a user-specified cursor and default to polling every 24 hours (just like BigQuery and MySQL batch connectors do) because Redshift doesn't expose the `xmin` system column which `source-postgres-batch` uses for automatic incremental captures.

[1] Redshift is based on Postgres internally and is thus compatible with the Postgres client protocol.

There are a handful of other tweaks, including:

- The spec response gives `https://go.estuary.dev/source-redshift-batch` as the documentation URL so we can have Amazon Redshift specific documentation.
- A bunch of new test logic making sure that things actually work as expected when running against Redshift.

https://go.estuary.dev/source-redshift-batch needs to be created and should describe at a minimum:

- How to make a Redshift database accessible from the outside world and whitelist the Estuary IPs, plus where to find its hostname/port info.
- How to create a `flow_capture` user with the necessary table read permissions.
- How to configure cursor columns after running discovery.

**Notes for reviewers:**

In principle this could probably have been implemented by adding some autodetection logic to `source-postgres-batch` so that when connecting to a Redshift database the generated resource specs would use cursors rather than `xmin`, along with the existing `DOCS_URL` environment override trick we've done for other connectors so we can have useful Redshift-specific docs. However this would have presented some issues regarding how to set the connector-level default polling interval, and I think there are probably a couple of other details like that which are harder to change automagically, so it was easier to just copy and edit.

I feel that it's justifiable to just copy-paste things for now, because I'm planning in the near future to tackle a series of refactorings to pull out more of the common logic between all of these Batch SQL captures, so most of the duplicated code this copying introduces should be removed in short order, likely right before the SQL Server or Oracle batch connectors get written.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1289)


source-redshift-batch/driver.go
```diff
@@ -0,0 +1,791 @@
...
+	}
+
+	var state captureState
+	if open.StateJson != nil {
+		if err := pf.UnmarshalStrict(open.StateJson, &state); err != nil {
+			return fmt.Errorf("parsing state checkpoint: %w", err)
+		}
+	}
+
+	migrated, err := migrateState(&state, open.Capture.Bindings)
```
williamhbaker, line 452: We shouldn't need the migration stuff for a brand new capture. Doesn't specifically hurt anything to leave it in though if that makes the copy/paste easier, and we plan to refactor the commonality of these captures soon-ish.

willdonnelly: Yeah, the entire file driver.go is currently identical between Redshift and Postgres, and I'd prefer to keep it that way if possible to make the refactor just that much simpler.

PR #1290, opened 2024-02-27
source-{mysql,postgres,redshift}-batch: Schema filtering

willdonnelly:
**Description:**

Adds a new advanced config property for specifying a list of schemas to discover tables within.

If the property is unset then tables all schemas will still be discovered. If one or more values are specified for this property then only tables from those schemas will be returned and everything else will be ignored.

Note that the discovery queries still fetch information about all tables in the entire database and the filtering is applied within the connector to the query results. In general this should be fine, but if it becomes an issue in the future it should be possible to push down the logic into `WHERE` clauses in the discovery queries themselves.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1290)


williamhbaker (approved):
LGTM

PR #1295, opened 2024-02-29
Fixing source-hubspot for long duration captures

Luishfs:
**Description:**
When running long duration captures in source-hubspot, a issue might occur where a new `ACCESS_TOKEN` has not yet refreshed, triggering a backoff.

The maximun try value set for this backoff is 5. Given that this can occur multiple times on a long duration stream, this PR aims to fix this by upgrading the backoff max value to 20.

**Tests**
Two tests were made on `contacts_lists_membership` stream:

1. The first one consisted on lowering the backoff max_try value to 1 ( forcing the issue to be raised ). This happened two hours after the stream initialization.

1. The second one consisted on raising the max_try value to 5 ( default ). Two backoffs were triggered: the first after 3 hours and the second after 6 hours after the stream initialization. Prints:

Screenshot from 2024-02-29 03-56-07

To my knowledge, the standard duration of a `ACCESS_TOKEN` is 30 minutes. Given the time proximity between both backoffs triggers, this might be caused by a pattern. Given the priority of this task and the amount of time necessary to efectivly test, no pattern search was conducted.

This PR lead to [the modification](https://github.com/estuary/flow/pull/1399) of Hubspot.md connector docs. 

**Notes for reviewers:**
Tests were made using Estuary dashboard, on a Hubspot > SnowflakeDB pipeline, to guarantee materialization of this stream. No issues were found.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1295)

### DIFF
--- a/source-hubspot/source_hubspot/streams.py
+++ b/source-hubspot/source_hubspot/streams.py
@@ -206,8 +206,8 @@ class API:
 
         return response.json()
 
-    @retry_connection_handler(max_tries=5, factor=5)
-    @retry_after_handler(max_tries=3)
+    @retry_connection_handler(max_tries=10, factor=5)
+    @retry_after_handler(max_tries=5)
     def get(
         self, url: str, params: MutableMapping[str, Any] = None
     ) -> Tuple[Union[MutableMapping[str, Any], List[MutableMapping[str, Any]]], requests.Response]:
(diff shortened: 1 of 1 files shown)


jshearer (approved):
Let's get this in if it works, but FWIW I feel like if we need to make 19 failing auth requests only for the 20th to work, there's something broader wrong here...

PR #1296, opened 2024-02-29
materialize-databricks: quote schema and volume identifier

mdibaiee:
**Description:**

- We were not quoting the schema and volume identifier when creating the volume, leading to this error in case of a schema with a special character:
```
Exec(CREATE VOLUME IF NOT EXISTS flow_staging;): databricks: execution error: failed to execute query: unexpected operation state ERROR_STATE: [PARSE_SYNTAX_ERROR] Syntax error at or near 'VOLUME'.(line 1, pos 7) == SQL == CREATE VOLUME IF NOT EXISTS vcp-galleries.flow_staging; -------^^^
```

Updated the code to quote the schema name in `CREATE VOLUME`. I also updated our integration tests to use a schema name with a special character so we will always test this case to avoid regression in the future.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-databricks/driver.go
+++ b/materialize-databricks/driver.go
@@ -216,7 +216,7 @@ func newTransactor(
 	}
 
 	// Create volume for storing staged files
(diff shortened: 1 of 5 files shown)


psFried (approved):
LGTM

PR #1300, opened 2024-03-01
source-bigquery-batch: Translate NaN to `"NaN"`

willdonnelly:
**Description:**

Unlike most SQL databases BigQuery allows NaNs to be stored into a floating-point column. The Go JSON serializer rejects NaNs, as they're not actually representable as JSON numbers.

There's no strictly right answer here, but in the past (see for instance `source-firestore`) we've mapped NaN floats to the string `"NaN"` in captures, so let's do that here too.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1300)

### DIFF
--- a/source-bigquery-batch/main.go
+++ b/source-bigquery-batch/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #1319, opened 2024-03-05
source-gladly: don't emit a checkpoint with no documents every time

williamhbaker:
**Description:**

Only emit an updated checkpoint when there are no documents if the current cursor is sufficiently old. Emitting a checkpoint with no documents will trigger an immediate re-invocation of the incremental task, so it shouldn't be done every time.

**Workflow steps:**

**Notes for reviewers:**


source-gladly/source_gladly/api.py
```diff
@@ -60,6 +60,15 @@ async def fetch_events(
                 f"not updating cursor since last_ts ({last_ts}) <= log_cursor ({log_cursor})"
             )
             return
+        elif last_ts - log_cursor < timedelta(hours=6):
```
jgraettinger, line 63: nit: i suspect this control flow could be cleaned up further, but this is correct :shrug:

PR #1320, opened 2024-03-05
Asana | Fix events stream

Z33DD:
Description:

According to https://github.com/airbytehq/airbyte/pull/31114 and https://github.com/airbytehq/airbyte/pull/31114, those streams will throw errors if the account doesn't pay for an Asana's paid plan.
The problem with the events' stream is the different pagination style that demands the first request to return a 'Precondition Failed' failed error (412).

About the Events stream, it is broken because of pagination, so it is not related to the paid account issue. I tested this flow with both paid and free accounts, the error is the sync token. I also added a condition in the `next_page_token` method to stop reading if the API returns `has_more` as false.

[Pagination for HTTP API connectors](https://docs.airbyte.com/connector-development/cdk-python/http-streams#pagination)

Asana API Reference:
- [Events](https://developers.asana.com/reference/getevents)
- [Portfolios](https://developers.asana.com/reference/getportfolios)


Z33DD:
The PR checks are failing for some other connectors. As my code doesn’t affect them, I don’t know how to proceed.
In the case of the Asana source, it's a problem in the way the workflow is configured.

jgraettinger (approved):
LGTM

PR #1328, opened 2024-03-06
Making table name readOnly

jonwihl:
**Description:**

Table name in the resource config for sql captures should be a read only field

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1328)


travjenkins (approved):
lgtm - do want to call out I think this would impact configs through the CLI as well. I think that is fine but not 100% sure if no user has ever wanted to change this.

PR #1331, opened 2024-03-06
materialize-mysql: reduce buffer use for data loading

williamhbaker:
**Description:**

Create transactor-scoped buffers and local file readers for uploading data across all bindings, rather than a buffer per binding. Also flush buffers after each time the binding changes during loads to take advantage of the natural grouping of load keys per binding, even though they aren't strictly increasing like for stores.

This greatly reduces memory use of the connector, and allows it to use a constant amount of memory rather than being proportional to the number of bindings.

Closes [#1332](https://github.com/estuary/connectors/issues/1332)

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1331)

### DIFF
--- a/materialize-mysql/infile.go
+++ b/materialize-mysql/infile.go
@@ -1,0 +1,109 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	stdsql "database/sql"
+	"encoding/csv"
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/go-sql-driver/mysql"
+)
+
+const (
+	// Limit the maximum size of buffered data before flushing a batch.
+	batchSizeThreshold = 10 * 1024 * 1024
+
+	// File name representations used both here and in the SQL templates.
+	loadInfileName   = "flow_batch_data_load"
+	insertInfileName = "flow_batch_data_insert"
+	updateInfileName = "flow_batch_data_update"
+)
+
+type infile struct {
+	w    *csv.Writer
+	buff *bytes.Buffer
+}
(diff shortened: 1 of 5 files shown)


materialize-mysql/infile.go
```diff
@@ -0,0 +1,109 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/go-sql-driver/mysql"
+)
+
+const (
+	// Limit the maximum size of buffered data before flushing a batch.
+	batchSizeThreshold = 10 * 1024 * 1024
```
williamhbaker, line 17: This was previously `33554432`, but I am lowering it here. This should still work for large documents, as we don't limit the size of the buffer and just use this limit to account for when the buffer should be drained. As before, a really large document will be serialized into the buffer and then the buffer will immediately be drained. But the lower limit will work better in the common case of small documents to keep the connector memory usage low.

materialize-mysql/sqlgen.go
```diff
@@ -163,7 +163,7 @@ type templates struct {
 	updateLoad        *template.Template
 	updateReplace     *template.Template
 	updateTruncate    *template.Template
-	storeLoad         *template.Template
+	insertLoad        *template.Template
```
williamhbaker, line 166: It's a bit of extra term, but the variable name `storeLoad` was absolutely breaking my brain.

jgraettinger (approved):
LGTM

PR #1336, opened 2024-03-07
source-mysql: Support enum primary keys whose cases are in non-alphabetical order

willdonnelly:
**Description:**

This PR fixes the connector backfill behavior when capturing from tables whose primary key includes a column of enum type whose cases are listed in non-alphabetical order.

The issue in this scenario was previously that if asked to `ORDER BY category` (where `category` is an enum column), MySQL will sort the rows based on the integer representation of those values. However when we ask it to return rows `WHERE category > 'foo'` for some string value `'foo'` the database will return those rows where that column's string representation is lexicographically greater than the string `'foo'`. Which is arguably correct, we need to cast that string to an enum to get the comparison behavior we actually want. But it makes things tricky for us here, because that cast is nontrivial to write generically.

So instead what we've done is the moral equivalent of going `WHERE category > 2` and comparing to the integer representation of the enum. To make this work we have to store the integer index of the enum value in our backfill row key serialization, instead of the string value as we previously did. Since the only things we use the row keys for are ordering comparisons and feeding back into the database, this is better than the current behavior in all cases.

**Workflow steps:**

Backfills in the rather niche corner case described above can now work in precise mode like they're supposed to.

I believe this change can safely be rolled out without any special effort at coordinating things. The change to row key serialization is technically incompatible, but this will only be an issue in practice if there's an ongoing backfill of a table with an enumeration primary key which gets restarted before the backfill completes, and if that were to happen it would show up in the task failures log and we can fix it by re-backfilling the impacted table(s) from the top.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1336)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -381,7 +381,9 @@ func getColumns(ctx context.Context, conn *client.Conn) ([]sqlcapture.ColumnInfo
 
 func parseDataType(typeName, fullColumnType string) any {
 	if typeName == "enum" {
-		return &mysqlColumnType{Type: "enum", EnumValues: append(parseEnumValues(fullColumnType), "")}
+		// Illegal values are represented internally by MySQL as the integer 0. Adding
+		// this to the list as the zero-th element allows everything else to flow naturally.
+		return &mysqlColumnType{Type: "enum", EnumValues: append([]string{""}, parseEnumValues(fullColumnType)...)}
 	} else if typeName == "set" {
 		return &mysqlColumnType{Type: "set", EnumValues: parseEnumValues(fullColumnType)}
 	}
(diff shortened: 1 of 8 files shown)


source-mysql/discovery.go
```diff
@@ -433,6 +431,20 @@ func (t *mysqlColumnType) translateRecordField(val interface{}) (interface{}, er
 	return val, fmt.Errorf("error translating value of complex column type %q", t.Type)
 }
 
+func (t *mysqlColumnType) encodeKeyFDB(val any) (tuple.TupleElement, error) {
```
williamhbaker, line 434: Do we think this method may eventually allow more types? Since it only allows `enum` here it seems kind of unnecessary vs. handling it in `encodeKeyFDB`, but it would make sense if there was an expectation to extend it in the future.

williamhbaker (approved):
LGTM

PR #1346, opened 2024-03-08
ci: make usage_rate a string in matrix definition

psFried:
**Description:**

Fixes how the `usage_rate` is passed in the GH actions for the new python connectors. Previously, this was being passed as an empty string due to an error in the workflow definition.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1346)


dyaffe (approved):
LGTM

PR #1348, opened 2024-03-08
materialize-snowflake: create schema if doesn't exist

mdibaiee:
**Description:**

- If the user-specified schema does not exist, attempt to create it automatically

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1348)

### DIFF
--- a/materialize-snowflake/client.go
+++ b/materialize-snowflake/client.go
@@ -22,7 +22,7 @@ type client struct {
 func newClient(ctx context.Context, ep *sql.Endpoint) (sql.Client, error) {
 	cfg := ep.Config.(*config)
 
-	db, err := stdsql.Open("snowflake", cfg.ToURI(ep.Tenant))
+	db, err := stdsql.Open("snowflake", cfg.ToURI(ep.Tenant, false))
 	if err != nil {
 		return nil, err
 	}
@@ -56,6 +56,12 @@ func (c *client) PutSpec(ctx context.Context, updateSpec sql.MetaSpecsUpdate) er
 }
 
 func (c *client) CreateTable(ctx context.Context, tc sql.TableCreate) error {
+	if len(tc.Path) > 1 {
+		var schemaName = tc.Path[0]
+		if _, err := c.db.ExecContext(ctx, fmt.Sprintf("CREATE SCHEMA IF NOT EXISTS %q", schemaName)); err != nil {
+			return err
+		}
+	}
 	_, err := c.db.ExecContext(ctx, tc.TableCreateSql)
 	return err
 }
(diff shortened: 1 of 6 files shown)


williamhbaker (changes requested):
I think we need to look at this more holistically, by incorporating the the concept of a resource "schema" into [InfoSchema](https://github.com/estuary/connectors/blob/main/materialize-boilerplate/info_schema.go) and handling the creation of schemas through `Apply`. That way it could be implemented for all SQL materializations, and all materializations that have a concept of a "schema", since I don't think we'll want to have it just be a snowflake-exclusive feature.

mdibaiee:
@williamhbaker I changed the implementation to fit into the existing Apply. I think this way this can be implemented for other materializations as well.

materialize-snowflake/snowflake.go
```diff
@@ -62,9 +62,6 @@ func schemasEqual(s1 string, s2 string) bool {
 }
 
 func (c tableConfig) Path() sql.TablePath {
-	if c.Schema == "" || schemasEqual(c.Schema, c.endpointSchema) {
```
williamhbaker, line 65: This bit is weird and cumbersome but I believe it is very important for backward compatibility: There was a time not so long ago that we did not allow resource-level schemas to even be set, see https://github.com/estuary/connectors/pull/980. Prior to this change, all resource paths were just the single `[]string{c.Table}`. If we were to start saying that all resource paths are now the 2-element `[]string{c.Schema, c.Table}`, it would change the connector's representation of pre-existing bindings' resource paths and would cause them to start over materializing from an empty checkpoint, which would not be good.

materialize-snowflake/client.go
```diff
@@ -56,6 +56,12 @@ func (c *client) PutSpec(ctx context.Context, updateSpec sql.MetaSpecsUpdate) er
 }
 
 func (c *client) CreateTable(ctx context.Context, tc sql.TableCreate) error {
+	if len(tc.Path) > 1 {
```
williamhbaker, line 59: I think this is OK for a quick-fix since it sounds like there are users that have an immediate need for automatic schema creation.

As a broader feature, I think we will want `InfoSchema` to represent which schemas already exist, and materialization `(*client).InfoSchema` implementations to populate that listing of existing schemas. Then we could add a `CreateSchema` or similar to the `boilerplate.Applier` interface and have the boilerplate figure out if there are schemas that don't exist but need to & create them and call `CreateSchema` if so.

williamhbaker (commented):
See comments - generally I think this will work for a immediate-term solution, but we need to resolve the concern about pre-existing resource paths.

materialize-snowflake/client.go
mdibaiee: @williamhbaker I don't think schemas are a universal-enough of a concept among materialization technologies for us to incorporate this into `InfoSchema`, which is part of `materialize-boilerplate`, not `materialize-sql`.

If this was a concept in `materialize-sql` I think it would make sense, since most of our current materialize-sql connectors do have schemas (I think Firebolt doesn't)

williamhbaker (approved):
LGTM

PR #1352, opened 2024-03-11
Source-hubspot: Removing max amount of auth retries

Luishfs:
**Description:**
As of now, Hubspot Source still fails due to  re-authentication limits.
This happens due to Hubspot `access_token` short lifespan ( about 30 minutes ), and the connector failing to re-authenticate properly.
Previous attemps set the number of retries to 20, but it seems the amount of user data can be higher than estimated and no issues would arise from removing the limit.

More information on the commit.
**Notes for reviewers:**
### New tests ( second commit, reviewing )
Tests were made locally running for about 2 hours and 15 hours

Token rotation is marked with a **GOT NEW TOKEN** info logger message.
Screenshot from 2024-03-22 14-51-57

as one can validate, each new token now rotates on about 28 minutes run, not 30 ( lifetime of the hubspot access_token)

Fully stats after runing for about 15 hours:
Screenshot from 2024-03-24 11-25-16

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1352)

### DIFF
--- a/source-hubspot/source_hubspot/streams.py
+++ b/source-hubspot/source_hubspot/streams.py
@@ -371,7 +371,7 @@ class Stream(HttpStream, ABC):
         return json_schema
 
 
-    @retry_401(max_tries=20)
+    @retry_401()
     def handle_request(
         self,
         stream_slice: Mapping[str, Any] = None,
--- a/source-hubspot/source_hubspot/streams.py
+++ b/source-hubspot/source_hubspot/streams.py
@@ -14,6 +14,7 @@ from typing import Any, Dict, Iterable, List, Mapping, MutableMapping, Optional,
 import backoff
 import pendulum as pendulum
 import requests
+from copy import deepcopy
 from airbyte_cdk.entrypoint import logger
 from airbyte_cdk.models import SyncMode
 from airbyte_cdk.sources import Source
@@ -371,7 +372,7 @@ class Stream(HttpStream, ABC):
         return json_schema
 
 
-    @retry_401()
+    @retry_401(max_tries=5)
     def handle_request(
         self,
         stream_slice: Mapping[str, Any] = None,
@@ -443,6 +444,8 @@ class Stream(HttpStream, ABC):
         pagination_complete = False
 
         next_page_token = None
+        self.authenticator._token_expiry_date = self.authenticator._token_expiry_date.subtract(minutes=2)
+        old_expiry_date = deepcopy(self.authenticator._token_expiry_date)
         try:
             while not pagination_complete:
                 properties = self._property_wrapper
@@ -461,6 +464,11 @@ class Stream(HttpStream, ABC):
                     )
                     records = self._transform(self.parse_response(response, stream_state=stream_state, stream_slice=stream_slice))
 
+                if old_expiry_date != self.authenticator._token_expiry_date:
+                    # this means that we got a new token
+                    self.logger.info(" GOT NEW TOKEN ")
+                    self.authenticator._token_expiry_date = self.authenticator._token_expiry_date.subtract(minutes=2)
+                    old_expiry_date = deepcopy(self.authenticator._token_expiry_date)
                 if self.filter_old_records:
                     records = self._filter_old_records(records)
                 yield from records
--- a/source-hubspot/source_hubspot/streams.py
+++ b/source-hubspot/source_hubspot/streams.py
@@ -444,6 +444,9 @@ class Stream(HttpStream, ABC):
         pagination_complete = False
 
         next_page_token = None
+        # Here we define both the new expiry date and its copy
+        # This will be used to compare wether the token expiry date has rotated
+        # if the token has a new expiry date, it means we have a new token
         self.authenticator._token_expiry_date = self.authenticator._token_expiry_date.subtract(minutes=2)
         old_expiry_date = deepcopy(self.authenticator._token_expiry_date)
         try:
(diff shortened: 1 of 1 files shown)


dyaffe (approved):
LGTM

jgraettinger (changes requested):
Hold up.

I'd like to understand _why_ removing the limit helps here. Is there a cached, global count or something?

This introduces a new risk of silent failure where it spins forever, retrying.

Luishfs:
@jgraettinger 
> This introduces a new risk of silent failure where it spins forever, retrying.

That's why i initially set the number of max retries to 20. 
Max retries = 20 basically means that, when running a capture, if the `access_token`  lifespan ends while on the request function ( causing hubspot to return a 401 HTTP response ), a backoff will trigger re-writing the request with a new `access_token`. The only time this backoff triggers is when a 401 HTTP response is delivered by Hubspot

Even after setting the max_retry number to 20, some long captures still failed, and if we are running everything single-thread, i believe that the backoff is running on a global count on the whole capture.

I did thought about setting up a absurd number, like 600, which would mean that even if every token of every stream would break, the connector would still run for about 300 hours, but since the backoff was only responsible for handling 401 responses ( meaning the access_token went off ), i've set it to inifinite retries.

Would you rather have the 600/absurd number?

jgraettinger:
> i believe that the backoff is running on a global count on the whole capture.

This sounds very plausible given your testing.

It's simply surprising to me because my definite assumption naively reading that decorator, is that it would apply to a _single_ invocation and would effectively reset the retry counter after a successful request.

Are you able to identify / link to where in the Airbyte CDK it has this behavior of keeping a global count?

Luishfs:
@jgraettinger  i think its being handled by the backoff lib [here](https://github.com/estuary/connectors/blob/main/source-hubspot/source_hubspot/streams.py#L73), but i did some searching and backoff strategies seem to be implemented directly by each connector ( [example](https://docs.airbyte.com/connector-development/connector-builder-ui/error-handling) )

Luishfs:
But on second thought, i think adding a really big number might really be the best option. A 401 that does not comply with re-auth is really unlikely but not impossible. If we set a number as high as 2k i believe it could cover most of the cases, and, if something happens, it should take about 1s for each requests to be completed, so it shouln't take long for the connector to fail. Do you agree?

jgraettinger:
They're using the `backoff` module. Digging into its code, [this](https://github.com/litl/backoff/blob/d82b23c42d7a7e2402903e71e7a7f03014a00076/backoff/_sync.py#L31-L34) is where tries and retries are actually calculated.

This sure looks to me like it's a per-paged-call retry counter -- in which case 20 should be _plenty_, assuming it's actually re-requesting a token after the first 401 --  and not a global count that can be exhausted.

Luishfs:
The behaviour we saw during a client capture was that the backoff was exhausted and let the exception escape ( with 20 )
I dont have access to logs so i can confirm it exactly, what i've been trying to do is assume the issue is being caused by the amount of expired `access_tokens`. which proved to be the case when runing all previous tests. 
Or the 20 retries are being "used" somewhere else, or its a global count by some matter.

jgraettinger:
What this tells us, unless I'm missing something, is that the bug is not in the number of retries but rather that it's not properly handling a retried 401 by immediately obtaining a new access token before the next attempt. **That** seems to be the bug.

That it happened to work with infinite retries is, I suspect, an accident of giving it _such_ a long time to retry in a tight loop that eventually it did turn over the token and get a new one.

source-hubspot/source_hubspot/streams.py
```diff
@@ -461,6 +464,11 @@ def read_records(
                     )
                     records = self._transform(self.parse_response(response, stream_state=stream_state, stream_slice=stream_slice))
 
+                if old_expiry_date != self.authenticator._token_expiry_date:
```
mdibaiee, line 467: I don't understand how this works. A few lines above we are doing a deep-copy of `old_expiry_date` and `self.authenticator._token_expiry_date`.

Either the deep copy is equal to the original value, in which case we will always have `old_expiry_date == self.authenticator._token_expiry_date` or it is not equal, and we will always have the opposite, since I don't see either of these values changing independently from each other. They are always a deep copy of each other. I might be missing some nuance about how the authenticator works, if so, can you please add a comment explaining how this works? Thank you

Luishfs: The `self.authenticator` handles a class (`Oauth2Authenticator`) that manages token rotation and the `._token_expiry_date` property. When a token rotates, this property is modified and thus will hold a new value, thats why i deep copy the value and compare it with a probable new one. New comments were added in e10bdb7f5a7456886d7d65d7709a3b1c21a2ff9f hopefully making it better to understand, would really appreciate with your feedback if these new comments help to understand better how the changes work.
Also, i've futher tested this change to 15 hours tests, and updated that on the PR description along with the running stats

Luishfs: The `self.authenticator` handles a class (`Oauth2Authenticator`) that manages token rotation and the `._token_expiry_date` property. When a token rotates, this property is modified and thus will hold a new value, thats why i deep copy the value and compare it with a probable new one. New comments were added in e10bdb7f5a7456886d7d65d7709a3b1c21a2ff9f hopefully making it better to understand, would really appreciate with your feedback if these new comments help to understand better how the changes work.
Also, i've futher tested this change to 15 hours tests, and updated that on the PR description along with the running stats

mdibaiee: Thanks, that makes sense now, the comments are really helpful!

source-hubspot/source_hubspot/streams.py
```diff
@@ -371,7 +372,7 @@ def get_json_schema(self) -> Mapping[str, Any]:
         return json_schema
 
 
-    @retry_401(max_tries=20)
+    @retry_401(max_tries=5)
```
mdibaiee, line 375: Should we keep this at 20, or do we also have a reason for reducing this?

Luishfs: The reason for reducing it is to change it back to the original value
This value should not matter anymore, as this backoff should not be triggered again anytime.

Reducing it back to 5 was my way of making sure future reviewers wouldnt think this is being used ( since it would show on their diff this value was modified sometime in the past)

mdibaiee (approved):

jgraettinger (approved):

williamhbaker:
This seems good to merge, so I am going to merge it.

PR #1361, opened 2024-03-12
source-mysql: Fix enum decoding off-by-one bug by migrating old metadata

willdonnelly:
**Description:**

In https://github.com/estuary/connectors/pull/1336 I (unnecessarily) refactored `source-mysql` to have the enum cases ordered like ["", "A", "B"] where previously they were ["A", "B", ""] and updated the indexing to match. But preexisting captures from before that change have the old style ordering in their metadata, and that's not updated anywhere, so they decode replication events incorrectly.

We can't even revert the changed indexing behavior at this point, because by now there are important captures which have been re-backfilled using the new ordering in their metadata, which would be incorrect if we simply reverted the change.

In order to fix this, we need to roll things forward and implement some metadata update logic which will prepend `""` to the list of options for any enum which doesn't already have it.

Once all production captures have run this code and updated their metadata it will be safe to remove the migration code and the associated test case.

After this change is live we should:

1. Ensure that all `source-mysql` capture tasks are restarted ASAP to make sure they pick up this fix.
2. Identify the scope of potentially-incorrect data which might have been captured over the past 4 days. Capture tasks which were restarted in the last 4 days and which captured replication events into their output collection from a table with an enum column type could potentially have incorrect values and will need to either be re-backfilled or otherwise fixed [1].

[1] It is theoretically possible to identify any incorrectly captured documents in the output collections which have not since been updated again, locate the enum case ordering which was in use at the time, construct an edited document, and append that into the collection. This would be a ton of work and it's almost certainly better to re-backfill, but if it's really important and we find a collection which cannot be recaptured this could be done.

**Notes for reviewers:**

Please look at this closely and let me know if you think of any edge cases I might have overlooked, I already screwed this up once with the original refactoring and I'd really prefer not to make things worse with a failed bugfix.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1361)

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -769,6 +769,36 @@ func (rs *mysqlReplicationStream) ActivateTable(ctx context.Context, streamID st
 				}
 			}
 		}
+
+		// This is a temporary piece of migration logic added in March of 2024. In the PR
+		// https://github.com/estuary/connectors/pull/1336 the ordering of enum cases in this
+		// metadata was changed to simplify the decoding logic. Specifically the order of the
+		// cases was changed from ["A", "B", ""] to ["", "A", "B"] to more directly mirror how
+		// MySQL represents the illegal-enum value "" as integer 0. However this introduced a bug
+		// when the new indexing code was used with older metadata from before that change, and
+		// by the time this was discovered there were captures in production with the new ordering
+		// in their checkpointed metadata, so a simple revert was not an option.
+		//
+		// The solution is to push forward and add the missing migration step, so that upon table
+		// activation any metadata containing the old ordering will be updated to the new. Once all
+		// metadata in production has been so updated, it will be safe to remove this logic and the
+		// associated 'TestEnumDecodingFix' test case.
+		for column, columnType := range metadata.Schema.ColumnTypes {
+			if enumType, ok := columnType.(*mysqlColumnType); ok && enumType.Type == "enum" {
+				if len(enumType.EnumValues) > 0 && enumType.EnumValues[0] != "" {
+					var logEntry = logrus.WithField("column", column)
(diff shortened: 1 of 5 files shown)


source-mysql/replication.go
```diff
@@ -769,6 +769,36 @@ func (rs *mysqlReplicationStream) ActivateTable(ctx context.Context, streamID st
...
+		// MySQL represents the illegal-enum value "" as integer 0. However this introduced a bug
+		// when the new indexing code was used with older metadata from before that change, and
+		// by the time this was discovered there were captures in production with the new ordering
+		// in their checkpointed metadata, so a simple revert was not an option.
+		//
+		// The solution is to push forward and add the missing migration step, so that upon table
+		// activation any metadata containing the old ordering will be updated to the new. Once all
+		// metadata in production has been so updated, it will be safe to remove this logic and the
+		// associated 'TestEnumDecodingFix' test case.
+		for column, columnType := range metadata.Schema.ColumnTypes {
```
jgraettinger, line 786: What's the mode of action by which the updated `rs.tables.metadata[streamID]` gets persisted back into the connector state? Will it happen if the DB is otherwise idle? (I think I know the answers, just want to explicitly have the discussion)

jgraettinger (approved):
LGTM % questions on state persistence

source-mysql/replication.go
willdonnelly: Tables which are already active (that is, we've previously started capturing them) at startup will run this logic before replication streaming starts. So while the main replication loop does only send metadata updates before blocking for the next replication event, these updates should be ready to go before that and thus take effect immediately even if the database is idle.

Also we generate a watermark change every couple minutes which would ensure we make progress even if the above wasn't the case.

PR #1366, opened 2024-03-13
Added new connector: source-google-analytics-data-api

Luishfs:
**Description:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1366)


source-google-analytics-data-api/tests/snapshots/snapshots__discover__capture.stdout.json
```diff
@@ -16,20 +27,19 @@
...
-      ],
       "type": [
+        "null",
         "object"
       ]
     },
     "key": [
-      "/date",
-      "/property_id"
+      "/property_id",
```
jgraettinger, line 41: It's crucially important that we not change around the ordering of keys, and also the schematization of keyed fields like `property_id` here.

There is a good reason why we had previously made them non-nullable (downstream materializations typically cannot support nullable primary keys).

Fixing this will require tweaking the schemas and keys of the connector streams. This is essentially re-applying patches we had previously used in airbyte-to-flow.

(Getting this right is the **central** reason I've been asking for this snapshot -> import -> update snapshots -- so that we can verify in review that nothing is changing unintentionally).

Luishfs: Got it, thanks for the info, this made everything more clear, will close the PR

PR #1367, opened 2024-03-13
Added new connector: Source-Salesforce

Luishfs:
**Description:**

**Notes for reviewers:**


source-salesforce/test.flow.yaml
```diff
@@ -0,0 +1,4810 @@
+---
+captures:
+  acmeCo/source-salesforce:
+    endpoint:
+        connector:
+          image: "ghcr.io/estuary/source-salesforce:v1"
+          config: config.yaml
+    bindings:
```
jgraettinger: nit: can leave this empty, as `bindings: []`.

Rationale: when editing / testing locally it's trivial to restore these with `flowctl raw discover`. Since we don't have a snapshot test that verifies captured documents, these aren't necessary so probably better not to have them.

source-salesforce/tests/snapshots/snapshots__spec__capture.stdout.json
```diff
@@ -8,52 +8,44 @@
...
               "type": "string"
             }
-          },
-          "required": [
-            "client_id",
-            "client_secret",
-            "refresh_token"
-          ],
-          "type": "object",
-          "x-oauth2-provider": "salesforce"
```
jgraettinger: This x-oauth2-provider bit is important and needs to be preserved.

source-salesforce/tests/snapshots/snapshots__discover__capture.stdout.json
jgraettinger, line 1: The amount of change in this file is concerning -- it's too large to even review in GitHub. Can you discuss the nature of the changes here? If you re-create this snapshot, after having switched to the estuary-cdk, are there fewer changes (because it was re-orderings introduced by the older `flow-sdk` that have since reverted to their original snapshot order)?

jgraettinger (commented):
Looking close. My remaining concerns are around the differences introduced in the discovery and spec snapshots. See comments below

PR #1370, opened 2024-03-14
ci: only refresh tags if connector is updated

mdibaiee:
**Description:**

- Previously we would refresh the connector tags of all connectors even if they were not updated in the commit
- With this change, we only refresh connectors which have had their code updated
- One caveat is updating of a SDK / dependencies (go.mod) / shared files will not trigger a refresh unless a connector has also been directly updated. We can also include those shared resources in our condition

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1370)

### DIFF
--- a/.github/actions/deploy/action.yaml
+++ b/.github/actions/deploy/action.yaml
@@ -60,8 +60,15 @@ runs:
       shell: bash
       run: sudo apt install postgresql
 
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM,  neat solution

Issue #1344, opened 2024-03-08
source-mysql: Handle CDC events with a `DB_ROW_HASH_1` column


willdonnelly:
Apparently there's a system column in MySQL / MariaDB called `DB_ROW_HASH_1` which is computed behind the scenes when you use unique constraints, at least some of the time. And apparently under some circumstances (one random comment I found says this started in MariaDB 10.4), MariaDB will write out this column into the binlog as part of a row change event.

When our connector encounters this, currently it instantly fails the capture because there's a column present in the change stream which we have no knowledge of. We should fix this, in one or more of the following ways:

- Delete fields named `DB_ROW_HASH_1` from the change event before processing it further, so that we never have to deal with these values and they never make it into the JSON output and whatnot. If we do this, we should probably confirm that MySQL won't allow you to create your own columns with this name.
- Hard-code a special case so that the metadata type lookup logic knows what type `DB_ROW_HASH_1` should have even if there is no tracked metadata corresponding to that field name.
- Modify the value translation logic so that when we're lacking metadata we just do a best-effort value-directed translation, since _most_ of the time the column type actually isn't important anyway and the appropriate handling is self-evident from the value.

willdonnelly:
Steps to reproduce this issue:
1. Run MariaDB version 10.4 or greater (confirmed that it still reproduces in `latest=11.3.2` at present)
2. Run the server with command line flags `["--log-bin=binlog", "--binlog-format=ROW", "--binlog_row_metadata=FULL"]`. The `--binlog_row_metadata=FULL` setting is required so that we get column names and can tell that the unexpected `DB_ROW_HASH_1` is present. Without that flag things will fail earlier when we get two column values in the change event while only expecting one, which is also suboptimal but may not be something we can handle.
3. Create a test table like `CREATE TABLE foobar (x VARCHAR(32) NOT NULL, UNIQUE INDEX x (x) USING hash)`.
4. Set up a capture and let it start up, then insert a row into the test table. The resulting change event in the binlog will contain a `DB_ROW_HASH_1` column value and will be received by the capture which then promptly fails.

[...]

PR #1372, opened 2024-03-14
source-mysql: Omit unexpected `DB_ROW_HASH_1` column values

willdonnelly:
**Description:**

See https://github.com/estuary/connectors/issues/1344 for the full explanation, but this commit adds a special-case to our document value translation logic to exclude columns by the name of `DB_ROW_HASH_1` for which we have no type information.

There is no accompanying test because this can only be reproduced on MariaDB and we use MySQL for our default test setup, but I have tested this locally and confirmed that it fixes what would otherwise be a capture error.

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -171,6 +171,15 @@ func (db *mysqlDatabase) translateRecordFields(columnTypes map[string]interface{
 		return nil
 	}
 	for id, val := range f {
+		// MariaDB versions 10.4 and up include a synthetic `DB_ROW_HASH_1` column in the
+		// binlog row change events for certain tables with unique hash indices (see issue
+		// https://github.com/estuary/connectors/issues/1344). In such cases we won't have
+		// any type information for the column, but we also don't want it anyway, so as a
+		// special case we just delete the 'DB_ROW_HASH_1' property from the document.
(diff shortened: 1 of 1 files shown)


source-mysql/discovery.go
```diff
@@ -171,6 +171,15 @@ func (db *mysqlDatabase) translateRecordFields(columnTypes map[string]interface{
 		return nil
 	}
 	for id, val := range f {
+		// MariaDB versions 10.4 and up include a synthetic `DB_ROW_HASH_1` column in the
+		// binlog row change events for certain tables with unique hash indices (see issue
+		// https://github.com/estuary/connectors/issues/1344). In such cases we won't have
+		// any type information for the column, but we also don't want it anyway, so as a
+		// special case we just delete the 'DB_ROW_HASH_1' property from the document.
+		if id == "DB_ROW_HASH_1" && columnTypes[id] == nil {
```
mdibaiee, line 179: Is it possible to somehow end up actually having `columnTypes[id]` here?

willdonnelly: I haven't tried but I suspect some supported versions of MySQL (most likely older ones and/or non-MariaDB) may prrmit a user to create a column by that name.

In such a case we would want to still capture that value, though really this is probably just theoretical and nobody will ever have an actual column by that name.

mdibaiee (approved):

PR #1378, opened 2024-03-15
materialize-snowflake: use dialect identifier transformer for schema

mdibaiee:
**Description:**

- We were quoting all identifiers and this was leading to "simple" identifiers being quoted which would then distinguish them from then unquoted, upper-case relatives

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1378)


williamhbaker (approved):
LGTM

Issue #1342, opened 2024-03-08
source-{postgres,redshift}-batch: Use `pg_catalog` tables for discovery


willdonnelly:
Currently the batch Postgres and Redshift connectors query `information_schema` views when performing table discovery. Unfortunately, these views are prone to permission-related setup issues. What's worse is that permission-related issues at this point manifest as a silent failure as though the tables or primary key information simply didn't exist.

This is especially awful when it comes to primary key discovery, because among other things we need to consult the `information_schema.table_constraints` view, and this view has the most insane access control rule imaginable. While most of these `information_schema` views require that the user simply have "some permission on the table", `table_constraints` contains:

> all constraints belonging to tables that the current user owns or has some privilege other than SELECT on.

We don't have nearly as much trouble with `source-postgres` because that connector consults various `pg_catalog` tables directly, and that either works and tells us about the tables/keys, or we get permission denied on the `pg_catalog` table access.

So I think at this point it's clear that we ought to be using the same discovery queries in these batch connectors. The only reason we don't already use those is that I originally hoped that sticking closer to standard SQL with `information_schema` might make future batch captures that much easier, but in practice it looks like we're always going to need per-database discovery functions with appropriate database-specific query strings, so there's no real benefit to using `information_schema` here and we've discovered plenty of downside.

PR #1379, opened 2024-03-15
source-{postgres,redshift}-batch: Use 'pg_catalog' for discovery

willdonnelly:
**Description:**

This PR replaces the table and primary key discovery queries used by the batch Postgres and Redshift captures. Previously these queries read from various `information_schema` views, but in practice getting permission grants set up so those views return the needed information could be quite difficult.

The fix is to drop down a level and consult the 'pg_catalog' tables directly. This ended up being a bit more complicated than expected, because Redshift is based on a really old version of Postgres and also changed some things around, but I think the new queries that I ended up with should work well enough for now.

This fixes https://github.com/estuary/connectors/issues/1342

**Workflow steps:**

There should be no changes that the user needs to care about, save that discovery will actually produce useful results in more situations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1379)

### DIFF
--- a/source-postgres-batch/.snapshots/TestBasicCapture-Discovery
+++ b/source-postgres-batch/.snapshots/TestBasicCapture-Discovery
@@ -15,7 +15,8 @@ Binding 0:
       "read_schema_json": {
         "type": "object",
         "required": [
-          "_meta"
+          "_meta",
+          "id"
         ],
         "properties": {
           "_meta": {
@@ -39,13 +40,15 @@ Binding 0:
               "polled",
               "index"
             ]
+          },
+          "id": {
+            "type": "integer"
           }
         },
         "x-infer-schema": true
       },
       "key": [
-        "/_meta/polled",
-        "/_meta/index"
+        "/id"
       ],
       "projections": null
     },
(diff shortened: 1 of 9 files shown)


williamhbaker (approved):
LGTM

Any concerns about re-discovery of existing captures changing the key of their collections?

willdonnelly:
There shouldn't be any issues with rediscovery. The representation of type names is changed (like `smallint` to `int2`) but I have updated the database->JSON type mapping table to match so the results should be unchanged.

Note that our type discovery is doing as little as possible here and exists solely to tell us if our proposed collection key is an integer, boolean, or string.

williamhbaker:
I was thinking of a case where we had previously discovered the collection key as the `_meta` fields since there wasn't any primary key information, but now there would be, so the collection would need to get re-created. It's probably an improvement though.

willdonnelly:
Oh, that's a very good point, thanks for bringing it up.

But per the discussion thread on Slack it's probably either acceptable or actively desirable to update most collections where the source table has a primary key and the destination collection currently doesn't, so I'm going to go ahead and merge this.

PR #1380, opened 2024-03-15
source-mysql: Omit `BINARY` qualifier when doing unfiltered backfill

willdonnelly:
**Description:**

Currently `source-mysql` has no logic to automatically select the unfiltered backfill mode, so we basically always do either a keyless or a precise keyed backfill. (This is because the whole precise/unfiltered distinction was added to deal with SQL Server text collation behavior and has not yet been fully extended to other situations where it might apply).

But in order to make precise backfills work, the MySQL backfill queries slap a `BINARY` qualifier on any textual primary key columns, and this basically always forces a full-table sort which can be prohibitively expensive in some situations.

It's possible to explicitly set the backfill mode for a binding to `"Normal"` in the spec, and it would be nice if doing this resulted in dropping the `BINARY` qualifier since we don't need the key ordering invariants this produces in the unfiltered mode.

So that's what this commit does -- if you explicitly select the `"Normal"` backfill mode we end up performing an unfiltered backfill and the query generation logic just omits the `BINARY` when the backfill mode isn't precise.

**Workflow steps:**

This shouldn't change the default behavior at all, but if you explicitly go and select the `"Normal"` backfill mode on a binding it will omit the `BINARY` ordering clauses in the backfill query, which should in some cases allow us to run much faster.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1380)


williamhbaker (approved):
LGTM

PR #1382, opened 2024-03-18
Add source-jira from Airbyte

Z33DD:
**Description:**

 Add Jira connector from the Airbyte repository.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1382)


williamhbaker:
For the import commit: Any reason not to use the license path of `airbyte-integrations/connectors/source-jira/metadata.yaml`?  Per the airbyte license in `./LICENSE`, "A `license` property defined in the `metadata.yaml` configuration file" takes precedence.

source-jira/tests/snapshots/test_snapshots/test_discover/capture.stdout.json
```diff
@@ -0,0 +1 @@
...log record.', 'readOnly': True}, 'issueId': {'type': 'string', 'description': 'The ID of the issue this worklog is for.', 'readOnly': True}, 'properties': {'type': 'array', 'description': 'Details of properties for the worklog. Optional when creating or updating a worklog.', 'items': {'type': 'object', 'properties': {'key': {'type': 'string', 'description': 'The key of the property. Required on create and update.'}, 'value': {'description': 'The value of the property. Required on create and update.'}}}}, '_meta': {'type': 'object', 'properties': {'row_id': {'type': 'integer'}}, 'required': ['row_id']}}, 'additionalProperties': True, 'description': 'Details of a worklog.', 'x-infer-schema': True}, 'key': ['/id']}, {'recommendedName': 'jira_settings', 'resourceConfig': {'stream': 'jira_settings', 'syncMode': 'full_refresh'}, 'documentSchema': {'$schema': 'http://json-schema.org/draft-07/schema#', 'type': 'object', 'properties': {'id': {'type': 'string', 'description': 'The ID of the application property. The ID and key are the same.'}, 'key': {'type': 'string', 'description': 'The key of the application property. The ID and key are the same.'}, 'value': {'type': 'string', 'descriptio
```
williamhbaker, line 1: These snapshot outputs need to follow the canonical structure (see [source-asana, for example](https://github.com/estuary/connectors/tree/main/source-asana/tests/snapshots)) rather than each being in a subdirectory, and should also be formatted to be human-readable, that way we can easily see and understand if there are any differences from subsequent changes.

source-jira/tests/snapshots/test_snapshots/test_capture/capture.stdout.json
```diff
@@ -0,0 +1 @@
+[]
```
williamhbaker, line 1: This snapshot does not seem useful. Is there supposed to be something in it?

source-jira/connector_config.yaml
```diff
@@ -0,0 +1,32 @@
+api_token_sops: ENC[AES256_GCM,data:S2H7YakJ+VJlB10Q8zHAH79Lkm7tBGNE44AxZug0SlVidrq4QPQZTfvmFvDGGQ8Eaixr9peu3fuSn8gjbeckEoHHr8JiyLF8+WX7GQzpa3JvP+w3yMFbwoj/3WniaN/gdaB0ruPsPAwAswg9nvbYb1luEWLcnE+ksn4seIa3OROp7iBKwXal6n/OOoxNkDTC9fZjV6UHDr8mx6o5fIeHhBIBOv3LAaxeVKcbsuL1LNSz69AHaS8nBuYVNyAUVtPd,iv:Refwby8Rzy77ipyCM/Zf2rI0r46g6r1NIXSuJTOy56o=,tag:0zru+Trf/RDUtX9MfwOoXg==,type:str]
```
williamhbaker, line 1: I do not think these are the test credentials we want to use. Can we set up an estuary test account and use that instead?

source-jira/source_jira/__main__.py
```diff
@@ -0,0 +1,24 @@
...
+from source_jira import SourceJira
+
+
+def wrap_with_braces(body: str, count: int):
+    opening = "{" * count
+    closing = "}" * count
+    return f"{opening}{body}{closing}"
+
+
+def urlencode_field(field: str):
```
williamhbaker, line 14: Do we need `urlencode_field` and `wrap_with_braces`? I don't see them being used anywhere.

source-jira/README.md
```diff
@@ -0,0 +1,91 @@
+# Jira source connector
```
williamhbaker, line 1: We need to update the github workflow so that tests run in CI and the connector image gets built.

See [this commit](https://github.com/estuary/connectors/commit/8e69ce33fa8671087d7ad6d6ecae5cbcc94f4f88) for a reference.

source-jira/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+This is the repository for the Jira source connector, written in Python.
```
williamhbaker, line 4: General comment: We need to get rid of this file and others that are not applicable for our imported connector. For example, this airbyte-specific `README` is not useful for us to check in to our repository.

Z33DD:
> For the import commit: Any reason not to use the license path of `airbyte-integrations/connectors/source-jira/metadata.yaml`? Per the airbyte license in `./LICENSE`, "A `license` property defined in the `metadata.yaml` configuration file" takes precedence.

No, no reason, should I change it?

williamhbaker:
> > For the import commit: Any reason not to use the license path of `airbyte-integrations/connectors/source-jira/metadata.yaml`? Per the airbyte license in `./LICENSE`, "A `license` property defined in the `metadata.yaml` configuration file" takes precedence.
> 
> No, no reason, should I change it?

Yes, since the import commit saying `License Path: ./LICENSE` is incorrect. Was just wondering if there was something I was missing.

source-jira/connector_config.yaml
Z33DD: Actually, Dave gave me those credentials.

williamhbaker: Ah I see, would you mind double-checking with Dave that these were the credentials we should check in with the code and running as part of CI?

source-jira/README.md
Z33DD: I have a question: should I do this in a separated commit? If not, in which commit should I include it?

Z33DD: I figured out that poetry demands a README.md file for each project. So we will need to keep but rewrite it.

williamhbaker: A separate commit would be fine for now.

Are you sure we need a README.md? Other connectors like [source-asana](https://github.com/estuary/connectors/tree/main/source-asana) don't have one.

williamhbaker (commented):
See inline comments, and also these more general ones:

- We need a commit history that shows the snapshots of the pre-CDK version of the connector, and then a separate commit that shows the diff of the snapshots produced by the connector running on the CDK. I see in 0e41544757cf11b3bcc1c0aca1b11924a2329883 the snapshots were generated, but it appears that those were from the CDK connector. For an example, see [this PR](https://github.com/estuary/connectors/pull/1407) where it was done for `source-notion`. We need to understand all the changes in this diff, since that is part of how we will know that existing tasks are safe to switch over to the new connector (see more on that in the next bullet point).
- There are a number of tasks currently running in production on the pre-CDK version of the connector, and we don't want those to break when the connector changes over to this version. So, please test out the migration scenario and add a note to the PR description on how it goes (hopefully it works without issue). See [here](https://github.com/estuary/connectors/blob/wb/connector-import-ref/estuary-cdk/importing_airbyte_source_connectors.md#migration-testing) for more information.

source-jira/pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+airbyte-cdk = "^0.52"
+
+[tool.poetry.scripts]
+source-jira = "source_jira.run:run"
+
+[tool.poetry.group.dev.dependencies]
+pytest = "==6.2.5"
+requests-mock = "^1.9.3"
+pytest-snapshot = "^0.9.0"
+responses = "^0.25.0"
```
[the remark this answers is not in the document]
Z33DD: `pytest-snapshot` does the same thing but with `--snapshot-update`. The main reason I use it is that it is easier to configure.

source-jira/tests/snapshots/spec.stdout.json
```diff
@@ -2,124 +2,104 @@
...
+        "order": 1
       },
       "email": {
-        "description": "The user email for your Jira account which you used to generate the API token. This field is used for Authorization to your account by BasicAuth.",
-        "order": 2,
+        "type": "string",
         "title": "Email",
-        "type": "string"
-      },
-      "enable_experimental_streams": {
```
williamhbaker: It seems like there are some fields that are being removed in the new spec, but it's hard to tell from this diff. Removal of fields from the spec is not backward compatible and is something we'll want to think carefully about.

Can you comment more on what is changing in the spec snapshot? A link to a [jsondiff](https://github.com/zgrossbart/jdd?tab=readme-ov-file#load-my-json-data-from-the-internet) that is easier to visually parse would probably suffice here.

williamhbaker (commented):
This is looking pretty good. My feedback at this point is about the changes in the snapshotted connector output. There's an inline comment for the `spec` output. Also for the `discover` output...the diff is too large to even review in the github web UI. We need to (somehow) understand what is changing between the pre- and post-import version of the connector. Can you comment on what is changing in the discovery output, and perhaps provide a jsondiff for that as well as the `spec` output?

Z33DD:
# Snapshot updates

As the diff are too big ever to show in a screenshot, I downloaded the HTML files from jsondiff.com for each of them and you can see in this zip file:
[JSON Diff - .zip](https://github.com/estuary/connectors/files/14911600/JSON.Diff.-.zip)

williamhbaker:
I couldn't get those HTML files to show up correctly in my browser.

Something like this works though, where the left side is the existing connector and the right side is the new one: https://jsondiff.com/?left=https://raw.githubusercontent.com/estuary/connectors/b0f03c4a1ea9b60e247f579dc218f4283334b022/source-jira/tests/snapshots/discover.stdout.json&right=https://raw.githubusercontent.com/estuary/connectors/5f9f0f4f0065a090f24d9c7adf4ae574f851c149/source-jira/tests/snapshots/discover.stdout.json

With this, I am able to see that there are a large number of changes, and some of them might not be backward compatible. Since these imports are intended to be drop-in replacements, I can't see us deploying this without more investigation to make sure it won't break existing tasks.

At this point my ask of you is this: Take a close look at the scheme changes yourself and let me know if I'm wrong about them not being backward compatible. Maybe they all are. If they are, just add an explanation for how that is. If there are changes that aren't backward compatible, is there anything we can do to resolve that? Maybe there are airbyte-to-flow patches that we haven't applied here, for example. Basically, you've provided the diff I asked for in [this comment](https://github.com/estuary/connectors/pull/1382#pullrequestreview-1964851908), but not the explanation for the diff, which is what I'm still needing to see.

Z33DD:
> I couldn't get those HTML files to show up correctly in my browser.
> 
> Something like this works though, where the left side is the existing connector and the right side is the new one: https://jsondiff.com/?left=https://raw.githubusercontent.com/estuary/connectors/b0f03c4a1ea9b60e247f579dc218f4283334b022/source-jira/tests/snapshots/discover.stdout.json&right=https://raw.githubusercontent.com/estuary/connectors/5f9f0f4f0065a090f24d9c7adf4ae574f851c149/source-jira/tests/snapshots/discover.stdout.json
> 
> With this, I am able to see that there are a large number of changes, and some of them might not be backward compatible. Since these imports are intended to be drop-in replacements, I can't see us deploying this without more investigation to make sure it won't break existing tasks.
> 
> At this point my ask of you is this: Take a close look at the scheme changes yourself and let me know if I'm wrong about them not being backward compatible. Maybe they all are. If they are, just add an explanation for how that is. If there are changes that aren't backward compatible, is there anything we can do to resolve that? Maybe there are airbyte-to-flow patches that we haven't applied here, for example. Basically, you've provided the diff I asked for in [this comment](https://github.com/estuary/connectors/pull/1382#pullrequestreview-1964851908), but not the explanation for the diff, which is what I'm still needing to see.

Outside the formatting changes and the patches, the change is to remove the `xml` fields in the schemas. I did this because the current connector does not support it and I don't think personally this is a relevant field for our customer as it just shows the same information in a different format.

williamhbaker:
> Outside the formatting changes and the patches, the change is to remove the `xml` fields in the schemas. I did this because the current connector does not support it and I don't think personally this is a relevant field for our customer as it just shows the same information in a different format.

I don't understand this comment. In the diff I linked, I can see other changes other than `xml` fields. I'm not sure which `xml` fields you are referring to. One example: The discovered `application_roles` collection for the new version of the connector does not have the collection key `/key` as required, whereas the prior version's output does have this key field as required. Collection keys should always be required, so this is a regression in the new connector and needs to be fixed. I need you to look at the differences in these connector outputs and minimize the differences where possible and explain areas where they can't be eliminated.

PR #1386, opened 2024-03-19
facebook: use default empty string for primary keys instead of required

mdibaiee:
**Description:**

- Use a default string so we can capture these collections even though sometimes the keys do not exist in certain documents. In these cases those documents will get reduced together (this is what has happened previously as well, since these fields were not even marked as a key)

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1386)


jgraettinger:
This is a little too broad, because it's applied to `format: date` fields as well. Default values must pass all schema validations. If you tried to actually preview a collection with these schemas I believe they'd failed with a build error.

PR #1387, opened 2024-03-19
Enable primary_key for custom queries in source-google-ads

Z33DD:
**Description:**

We have a customer who is trying to add custom queries to the source-google-ads connector. The errors were caused by the lack of primary keys for the `CustomQueryMixin` stream, this PR enables it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1387)

### DIFF
--- a/source-google-ads/source_google_ads/custom_query_stream.py
+++ b/source-google-ads/source_google_ads/custom_query_stream.py
@@ -16,13 +16,7 @@ class CustomQueryMixin:
 
     @property
     def primary_key(self) -> str:
-        """
-        The primary_key option is disabled. Config should not provide the primary key.
-        It will be ignored if provided.
-        If you need to enable it, uncomment the next line instead of `return None` and modify your config
-        """
-        # return self.config.get("primary_key") or None
-        return None
+        return self.config.get("primary_key") or None
 
     @property
     def name(self):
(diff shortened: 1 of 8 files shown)


source-google-ads/source_google_ads/custom_query_stream.py
```diff
@@ -17,12 +17,14 @@ def __init__(self, config, **kwargs):
...
+        The primary_key option is disabled. Config should not
+        provide the primary key.
         It will be ignored if provided.
-        If you need to enable it, uncomment the next line instead of `return None` and modify your config
+        If you need to enable it, uncomment the next line instead of
+        `return None` and modify your config
         """
-        # return self.config.get("primary_key") or None
-        return None
+        return self.config.get("primary_key") or None
```
williamhbaker, line 26: Can you update the comment above this line, since changing this code has invalidated it?

williamhbaker (commented):
Since we're going to expose the `primary_key` configuration, we need to update the spec to expose that configuration otherwise there will be no way to set in in the UI. So I'd expect the `spec.json` to need an update, and also the spec snapshot test output.

source-google-ads/source_google_ads/custom_query_stream.py
Z33DD: Done!

williamhbaker:
Closing this per https://github.com/estuary/connectors/issues/1486, where broader work for enabling this is tracked

PR #1389, opened 2024-03-19
source-boilerplate: Buffer documents to amortize write overhead

willdonnelly:
**Description:**

This change is fairly small but impactful. Johnny's profiling earlier today apparently found that we're spending a lot of time blocked on write syscalls, which makes sense since we currently output every document to Flow as a separate message send.

This commit wraps a 1MiB `bufio.Writer` around stdout at the very base level, and adds flush logic to the `Send()` method so that anything *except* a document flushes that buffer. The end result should be that our behavior is mostly unchanged except that we'll accumulate capture documents into batches which get flushed only when the buffer fills up or we reach the next checkpoint.

This should hopefully improve connector throughput, though I can't say by how much.

This is a very low-level change, so it isn't exercised by most of our CI test suite. However our integration tests should run the new logic, so if CI passes we've probably got it right. Also Johnny wrote a very similar implementation in parallel before he had to leave for his vacation (see https://github.com/estuary/connectors/pull/1388) so comparing the two also gives me some confidence this is correct.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1389)

### DIFF
--- a/source-boilerplate/boilerplate.go
+++ b/source-boilerplate/boilerplate.go
@@ -29,6 +29,8 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
+const outputWriteBuffer = 1 * 1024 * 1024 // 1MiB output buffer to amortize write syscall overhead
+
 // StateKey is used to key binding-specific state a connector's driver checkpoint. It's just a type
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM

PR #1391, opened 2024-03-20
Add source-stripe from Airbyte

Z33DD:
**Description:**

This pull request introduces a new connector, source-stripe, into Estuary's connectors repository derived from Airbyte. The Stripe connector allows Estuary users to easily and securely ingest data from their Stripe accounts in real time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1391)


source-stripe/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM python:3.9-slim
```
williamhbaker, line 1: There are leftover files in the root directory from the import that we don't need anymore.

source-stripe/tests/snapshots/capture.stdout.json
```diff
@@ -0,0 +1,5893 @@
+[
```
williamhbaker, line 1: There are several duplicated snapshot outputs for these tests. There should only be one snapshot output file for each of the capture/discover/spec commands, and they should be in the canonical file structure.

williamhbaker (commented):
Made a couple of specific comments here. Also see comments on https://github.com/estuary/connectors/pull/1382 for applicability here.

PR #1393, opened 2024-03-20
cdk: Fix unneccesary escaping on key components in Airbyte shim

jshearer:
**Description:**

Flow key components are strings that represent JSON pointers. So a key component `/meta/row_id` would mean that the key for document `{"meta": {"row_id": 1234}}` would be `1234`. But we could also have a document like `{"pizza/party": 1234}`, and if we wanted to use that field `"pizza/party"` as the key, that’s where we’d want to escape it as `"pizza~1party"`. I did some digging, and it looks like airbyte handles nested keys a little differently:
```python
    @property
    @abstractmethod
    def primary_key(self) -> Optional[Union[str, List[str], List[List[str]]]]:
        """
        :return: string if single primary key, list of strings if composite primary key, list of list of strings if composite primary key consisting of nested fields.
          If the stream has no primary keys, return None.
        """
```

So this PR updates our key escaping to be more nuanced and handle translating Airbyte's nested-list representation to our escaped-string representation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1393)

### DIFF
--- a/estuary-cdk/estuary_cdk/shim_airbyte_cdk.py
+++ b/estuary-cdk/estuary_cdk/shim_airbyte_cdk.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass
 from logging import Logger
 from pydantic import Field
-from typing import Any, ClassVar, Annotated, Callable, Awaitable, Literal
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM, thanks!

PR #1397, opened 2024-03-21
source-redshift-batch: Discover all columns and omit fallback key

willdonnelly:
**Description:**

This PR modifies the `source-redshift-batch` connector so that the generated schemas for discovered collections will include all columns with appropriate JSON schema types, rather than just the primary-key columns as it did previously.

In addition `source-redshift-batch` will no longer suggest a fallback collection key of `[/_meta/polled, /_meta/offset]` for tables without a primary key. This means that users will be forced to pick their desired collection key, which is why it was important that we adjust the discovered schema so that they have appropriate options to choose from.

**Workflow steps:**

I haven't given a lot of thought to how the change in discovery output will impact existing uses of the Redshift connector. The omission of a collection key for keyless tables should be benign, but I'm not sure what the removal of the `x-infer-schema` property and the sudden appearance of more properties in the connector's discovered schemas will do. Probably it will all work out correctly, I think?

New users will find that when doing the capture creation flow in the UI any tables lacking primary keys (which, in a data warehouse like Redshift, is most of them) they are prompted to select a collection key using one or more columns of  the source table, and must do so before they can finish creating the capture.

Existing users will find that if they go and manually reversion their collection with a more appropriate key, it no longer gets blown away by the automatic discovery merging behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1397)

### DIFF
--- a/source-redshift-batch/.snapshots/TestBasicCapture-Discovery
+++ b/source-redshift-batch/.snapshots/TestBasicCapture-Discovery
@@ -39,11 +39,13 @@ Binding 0:
               "index"
             ]
           },
+          "data": {
+            "type": "string"
+          },
           "id": {
             "type": "integer"
           }
-        },
-        "x-infer-schema": true
+        }
       },
       "key": [
         "/id"
--- a/source-redshift-batch/.snapshots/TestBasicDatatypes-Discovery
+++ b/source-redshift-batch/.snapshots/TestBasicDatatypes-Discovery
@@ -39,11 +39,28 @@ Binding 0:
               "index"
             ]
           },
+          "a_bool": {
+            "type": "boolean"
+          },
+          "a_date": {
+            "type": "string",
+            "format": "date-time"
+          },
+          "a_real": {
+            "type": "number"
+          },
+          "a_ts": {
+            "type": "string",
+            "format": "date-time"
+          },
+          "a_tstz": {
+            "type": "string",
+            "format": "date-time"
+          },
           "id": {
             "type": "integer"
           }
-        },
-        "x-infer-schema": true
+        }
       },
       "key": [
         "/id"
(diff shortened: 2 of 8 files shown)


willdonnelly:
A note regarding the "impact to existing users" thing, there is actually only one noteworthy capture using this connector in production at present and this change will improve things for them. So never mind, doesn't matter what the theoretical impact is on other hypothetical captures that don't actually exist.

psFried (approved):
The code here looks good to me.

> The omission of a collection key for keyless tables should be benign, but I'm not sure what the removal of the x-infer-schema property and the sudden appearance of more properties in the connector's discovered schemas will do. Probably it will all work out correctly, I think?

I looked into [the agent code](https://github.com/estuary/flow/blob/4424b02d8890c97f1ad290ef3ff7486d0a22499d/crates/agent/src/discovers/specs.rs#L256), and I think what will happen is that existing captures will continue to use schema inference unless the collections are either re-created or manually updated. I don't know whether that's what we want, but I also don't really know what we could realistically do differently. The agent assumes that if a collection specifies a `readSchema` then it shouldn't be modified by discovers after it's created. That still seems like the correct behavior to me.

willdonnelly:
> I looked into [the agent code](https://github.com/estuary/flow/blob/4424b02d8890c97f1ad290ef3ff7486d0a22499d/crates/agent/src/discovers/specs.rs#L256), and I think what will happen is that existing captures will continue to use schema inference unless the collections are either re-created or manually updated. I don't know whether that's what we want, but I also don't really know what we could realistically do differently. The agent assumes that if a collection specifies a readSchema then it shouldn't be modified by discovers after it's created. That still seems like the correct behavior to me.

That sounds like an acceptable behavior, especially because I think all the production collections we actually care about will get re-versioned manually right after this change is live.

PR #1398, opened 2024-03-21
source-redshift-batch: Omit empty cursors in discovered resources

willdonnelly:
**Description:**

The cursor values will always be empty, and we want that to be represented by omitting the property from the discovered resource spec entirely rather than with an explicit null.

Also makes the same change to `source-postgres-batch`. It doesn't matter there because discovered resources always have the default cursor `["txid"]` but the fact that I left off the `omitempty` in that case is why the Redshift connector got this wrong.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1398)

### DIFF
--- a/source-postgres-batch/.snapshots/TestSpec
+++ b/source-postgres-batch/.snapshots/TestSpec
@@ -109,8 +109,7 @@
     "type": "object",
     "required": [
       "name",
-      "template",
-      "cursor"
+      "template"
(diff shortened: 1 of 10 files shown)


willdonnelly:
Merging without review because this is a fairly trivial bugfix for a rough edge when using the connector in the UI.

PR #1401, opened 2024-03-21
source-redshift-batch: Run discovery queries in parallel

willdonnelly:
**Description:**

The parallel execution is done using a fairly clunky and verbose pattern (running each discovery query in a separate `errgroup` worker thread, using a single-element buffered channel for each thread's result) because it was what I could confidently bang out in 15 minutes.

We should consider replacing the whole mess with a suitable 'promise' abstraction helper package if one exists.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1401)


willdonnelly:
Merging without review because it seems to work fine and the worst that's likely to happen if there were a bug is Redshift discovery timing out.

PR #1402, opened 2024-03-22
materialize-sql: close client in Validate

mdibaiee:
**Description:**

- We previously applied this change of short-lived database connections elsewhere in the connector, but we got reports that they still see leftover connections _sometimes_. I found that we had this case where we would open a connection in Validate as part of initialising the `client`, but we were not closing it.
- Decreased file upload concurrency until https://github.com/databricks/databricks-sql-go/pull/197 is merged
**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1402)

### DIFF
--- a/materialize-databricks/staged_file.go
+++ b/materialize-databricks/staged_file.go
@@ -16,7 +16,7 @@ import (
 )
 
 const fileSizeLimit = 128 * 1024 * 1024
-const uploadConcurrency = 5 // that means we may use up to 1.28GB disk space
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

It would not seem unreasonable to me to lower the `uploadConcurrency` down to 1 just to be sure we don't still get OOMs until the proper fix is merged, but if 3 works that's fine too.

PR #1406, opened 2024-03-22
source-gladly: add base_url advanced configuration option

williamhbaker:
**Description:**

Adds an advanced option to specify an alternate base URL to use for connecting to the Gladly API. An alternate base URL is sometimes useful when connecting to a sandbox account, for example.

Manually tested this with both a regular production config, and a sandbox config.

**Workflow steps:**

`source-gladly` docs should be updated with this new config option.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1406)


jshearer (approved):
Do we know if there are non-https URLS people are using? Ideally I feel like this would be something like `domain` rather than `base_url` as otherwise it's hard to know whether you need to include the scheme, a trailing slash or not, etc. The description/example does help clarify though, and there probably is some weird edge case that having this additional configurability will help with, so I think it's fine to ship this as is too. 👍

williamhbaker:
Its only HTTPS as far as I know. For sandbox accounts, they give you the address to use beginning with an HTTPS, so accepting that input is probably the most direct way users will interact with this. To your point, we could add more verification that it looks correct, and maybe we will if there's much usage to this, but I'm being lazy for now and punting on that.

PR #1408, opened 2024-03-22
source-redshift-batch: Fix handling of column nullability

willdonnelly:
**Description:**

Currently most discovery flows are broken because we treat columns as always being non-null due to a dumb oversight in how the boolean nullability value is (not) plumbed through to schema generation, so then the moment you actually start capturing data you get a ton of schema violations wherever nulls actually exist.

This PR fixes that oversight.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1408)

### DIFF
--- a/source-redshift-batch/.snapshots/TestBasicCapture-Discovery
+++ b/source-redshift-batch/.snapshots/TestBasicCapture-Discovery
@@ -39,7 +39,10 @@ Binding 0:
             ]
           },
           "data": {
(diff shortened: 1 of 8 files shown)


willdonnelly:
Merging without review because I've verified that the snapshot changes look correct, the test suite passes, and the current state of things without this fix is basically unusable on real-world datasets that tend to contain nulls.

PR #1410, opened 2024-03-25
materialize-databricks: close db connections after use in client

mdibaiee:
**Description:**

- The previous attempt to fix this issue does not seem to have worked, and I tried to spot other areas where we might be failing to close the connection by not calling `client.Close`, but couldn't find anything. According to the user, this is usually followed by a `select 1` query which I think is the `db.PingContext` call in `Validate` RPC.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1410)

### DIFF
--- a/materialize-databricks/client.go
+++ b/materialize-databricks/client.go
@@ -22,7 +22,6 @@ import (
 )
 
 type client struct {
-	db  *stdsql.DB
 	cfg *config
 	ep  *sql.Endpoint
 }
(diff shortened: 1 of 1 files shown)


williamhbaker:
I think this is implied from the PR description but since we're calling `client.Close()` wherever a client is created I'm not able to see how this would change anything. Have you been able to reproduce the issue that is being reported?

PR #1411, opened 2024-03-25
Mahdi/bigquery idempotent apply

mdibaiee:
**Description:**

- Switches BigQuery to idempotent apply pattern, this means no more fencing and that would help alleviate the write contentions we experience with this connector
- Also updated the way the integration test snapshot is taken so it is more informative now and includes duplicate keys and delta updates
- During Store phase, we upload the data to gcs and prepare the queries to be run as part of the checkpoint. This includes a JobID for the query job to be run. During Acknowledge we try to create a job and run the query with the given ID. If we find that the job already exists, it means we have been able to successfully kickstart the query before, and at this point we only need to check the status of the job and if it has errored, we raise that error for the query.
- A JobID can only be created once. If the "creation" of a job itself is successful, no matter whether the actual query succeeds or not, the JobID has been allocated and cannot be re-allocated. On the other hand, if the creation of a Job fails, we can always create that job with the JobID since that JobID is free to take.
- This means once we have executed a SQL query once, we cannot retry it if it fails. This is a limitation in face of queries that may not be deterministic. I have not hit this error myself, but I can imagine a case of a SQL query failing due to a hiccup in reading the content of files from gcs (I think of this as unlikely)
- However we are guaranteed to be able to run the SQL query with the generated JobID once, since if creating the job fails we can retry creating the job with the same ID

**Workflow steps:**

- Tested erroring out after Store phase but before Ack phase can run, and on a subsequent restart the queries are run as expected
- Tested erroring out after Ack phase has run queries but before it has cleaned up the checkpoint, and on restart the recovery step checks the job status and does not lead to duplicate rows

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1411)

### DIFF
--- a/materialize-bigquery/client.go
+++ b/materialize-bigquery/client.go
@@ -3,7 +3,6 @@ package connector
 import (
 	"context"
 	dbSql "database/sql"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"net/http"
@@ -237,42 +236,7 @@ func (c *client) ExecStatements(ctx context.Context, statements []string) error 
 }
 
 func (c *client) InstallFence(ctx context.Context, _ sql.Table, fence sql.Fence) (sql.Fence, error) {
-	var query strings.Builder
-	if err := tplInstallFence.Execute(&query, fence); err != nil {
-		return fence, fmt.Errorf("evaluating fence template: %w", err)
-	}
-
-	log.Info("installing fence")
-	job, err := c.query(ctx, query.String())
-	if err != nil {
-		return fence, err
-	}
-	var bqFence struct {
-		Fence      int64  `bigquery:"fence"`
-		Checkpoint string `bigquery:"checkpoint"`
-	}
-	log.Info("reading installed fence")
-	if err = c.fetchOne(ctx, job, &bqFence); err != nil {
-		return fence, fmt.Errorf("read fence: %w", err)
-	}
-
-	checkpoint, err := base64.StdEncoding.DecodeString(bqFence.Checkpoint)
-	if err != nil {
-		return fence, fmt.Errorf("base64.Decode(checkpoint): %w", err)
-	}
-
-	fence.Fence = bqFence.Fence
-	fence.Checkpoint = checkpoint
-
-	log.WithFields(log.Fields{
-		"fence":            fence.Fence,
-		"keyBegin":         fence.KeyBegin,
-		"keyEnd":           fence.KeyEnd,
-		"materialization":  fence.Materialization.String(),
-		"checkpointsTable": fence.TablePath,
-	}).Info("fence installed successfully")
-
-	return fence, nil
+	return sql.Fence{}, nil
 }
 
 func (c *client) Close() {
--- a/materialize-bigquery/descriptor_proto.go
+++ b/materialize-bigquery/descriptor_proto.go
@@ -1,0 +1,49 @@
+package connector
+
+import (
+	"fmt"
+
+	sql "github.com/estuary/connectors/materialize-sql"
+	"google.golang.org/protobuf/types/descriptorpb"
+)
+
+func getTableDescriptorProto(table sql.Table) descriptorpb.DescriptorProto {
+	var cols = table.Columns()
+
+	var dp = descriptorpb.DescriptorProto{
+		Name:  &table.Identifier,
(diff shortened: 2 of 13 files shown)


williamhbaker:
Making sure I understand this limitation:
>This means once we have executed a SQL query once, we cannot retry it if it fails. This is a limitation in face of queries that may not be deterministic. I have not hit this error myself, but I can imagine a case of a SQL query failing due to a hiccup in reading the content of files from gcs (I think of this as unlikely)

We do see these random intermittent failures somewhat regularly, so much so that we have this [whole bit of ugly code](https://github.com/estuary/connectors/blob/6db7e4a98da653a2ea91d4dac60d544028c66456/materialize-bigquery/query.go#L47-L131) to retry under various circumstances. If one of these intermittent failures occurs, will the materialization be forever stuck with this change, only able to proceed if a binding is re-backfilled? If so I don't think that's viable, since these do happen on a not super-rare basis.

mdibaiee:
> We do see these random intermittent failures somewhat regularly, so much so that we have this [whole bit of ugly code](https://github.com/estuary/connectors/blob/6db7e4a98da653a2ea91d4dac60d544028c66456/materialize-bigquery/query.go#L47-L131) to retry under various circumstances. If one of these intermittent failures occurs, will the materialization be forever stuck with this change, only able to proceed if a binding is re-backfilled? If so I don't think that's viable, since these do happen on a not super-rare basis.

If one of these intermittent issues happens and the connector restarts before it is able to succeed, it will be stuck in a forever loop of reporting that error and not retrying

williamhbaker:
> > We do see these random intermittent failures somewhat regularly, so much so that we have this [whole bit of ugly code](https://github.com/estuary/connectors/blob/6db7e4a98da653a2ea91d4dac60d544028c66456/materialize-bigquery/query.go#L47-L131) to retry under various circumstances. If one of these intermittent failures occurs, will the materialization be forever stuck with this change, only able to proceed if a binding is re-backfilled? If so I don't think that's viable, since these do happen on a not super-rare basis.
> 
> If one of these intermittent issues happens and the connector restarts before it is able to succeed, it will be stuck in a forever loop of reporting that error and not retrying

Is it worse than that even? [Here](https://github.com/estuary/connectors/blob/6db7e4a98da653a2ea91d4dac60d544028c66456/materialize-bigquery/query.go#L53C3-L53C28) where we retry queries, right now we create a new job for each retry. If we have set a query ID, I would think that would not work either, since a new job can't be created with the same ID. The big uncertainty I've had with this strategy is if there is a way to retry jobs with the same ID on these types of intermittent failures, and it sounds like there isn't?

mdibaiee:
> If we have set a query ID, I would think that would not work either, since a new job can't be created with the same ID.

Yes as it is written right now, that's the case, but we can workaround that by changing the JobID for these client-side retries, which would only apply for the first run of Acknowledge and not for recovery (because then the initial JobID would exist and we would not retry the "Already Exists" error)

> he big uncertainty I've had with this strategy is if there is a way to retry jobs with the same ID on these types of intermittent failures, and it sounds like there isn't?

No. I initially had this worry as well, and then later I convinced myself that if the creation of a SQL query job can be retried in case we are not able to actually create the job, and if we can track the status of that job, with the assumption that intermittent failures are rare, we can live with this, but now I see that intermittent failures are not uncommon, in which case this would be a disaster as we would be crash looping...


------------

I've considered the Storage Write API for idempotent apply pattern as well, and I think with that we can handle delta updates bindings and for normal bindings we would need to use `MERGE` queries as we do right now, and since the `MERGE` query is idempotent in and of itself, we can retry by creating a job with a different ID.

Basically this is the process for Storage Write API:

1. During Store phase we create a write stream and checkpoint its ID
2. During `Acknowledge`, we read from the files and stream the JSON records to the write stream using `AppendRows` RPC calls, and we also provide the `offset` which is basically telling the stream which record number the current record holds. If the record with this offset has already been written by Storage API, it is not written again. This allows us to retry writes while keeping it idempotent. The main challenge here is us having to stream records through GCS to Snowpipe streaming. We need to have the records "somewhere" so we can send them using RPC to Snowpipe Streaming since Snowpipe Streaming does not have a way of reading records from files directly. Hopefully this won't be too slow...

mdibaiee:
Committed my partial work for Storage Write API, going to close this pull-request as we have decided not to go ahead with this for now.

PR #1414, opened 2024-03-25
source-mysql: Silently ignore a couple of query events after all

willdonnelly:
**Description:**

Turns out that in https://github.com/estuary/connectors/pull/1412 I was overly optimistic about the rarity of certain non-DDL query events in the binlogs.

The two most egregiously spammy forms are:

    "BEGIN"
    "# Dummy event replacing event type 160 that slave cannot handle.             "

But I'm still skeptical about the `ignoreQueriesRe` in general and suspect that one or more of those query event regexps is responsible for us missing out on DDL operations we care about.

So I'm introducing a quick-and-dirty third tier here, a regexp for only the most spammy of query events which uses more conservative patterns so we're *really* sure anything matching that can be ignored without even logging it (at `INFO` level).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1414)

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -452,6 +452,7 @@ func decodeRow(streamID string, colNames []string, row []interface{}) (map[strin
 // with the binlog Query Events for some statements like GRANT and CREATE USER.
 // TODO(johnny): SET STATEMENT is not safe in the general case, and we want to re-visit
 // by extracting and ignoring a SET STATEMENT stanza prior to parsing.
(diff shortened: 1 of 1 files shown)


willdonnelly:
I intend to merge this without review as soon as the CI build completes, because this is really quite a bit spammier than I expected and I'd like to put a stop to that ASAP, and I'm pretty confident that this copy-paste of our existing logic with a much smaller regexp is correct.

PR #1430, opened 2024-03-27
stdsql: close connection used to ping database in exec

mdibaiee:
**Description:**

- I found the root cause of the leftover connection in Databricks, we were creating this additional connection here but not closing it.
- I was able to reproduce the original issue and can confirm with this patch it doesn't happen anymore

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1430)

### DIFF
--- a/materialize-sql/std_sql.go
+++ b/materialize-sql/std_sql.go
@@ -46,11 +46,13 @@ func StdSQLExecStatements(ctx context.Context, db *sql.DB, statements []string) 
 	// through a single connection. This allows a driver to explicitly run
 	// `BEGIN;` and `COMMIT;` statements around a transactional operation.
 	var conn, err = db.Conn(ctx)
-	if err == nil {
-		err = conn.PingContext(ctx)
-	}
 	if err != nil {
 		return fmt.Errorf("connecting to DB: %w", err)
+	}
+	defer conn.Close()
+
+	if err = conn.PingContext(ctx); err != nil {
+		return fmt.Errorf("ping DB: %w", err)
 	}
 
 	for _, statement := range statements {
@@ -59,7 +61,8 @@ func StdSQLExecStatements(ctx context.Context, db *sql.DB, statements []string) 
 		}
 		logrus.WithField("sql", statement).Debug("executed statement")
 	}
-	return conn.Close() // Release to pool.
+
+	return nil
 }
 
 // StdInstallFence is a convenience for Client implementations which
--- a/materialize-sql/std_sql.go
+++ b/materialize-sql/std_sql.go
@@ -49,7 +49,9 @@ func StdSQLExecStatements(ctx context.Context, db *sql.DB, statements []string) 
 	if err != nil {
 		return fmt.Errorf("connecting to DB: %w", err)
 	}
-	defer conn.Close()
+	defer func() {
+		err = conn.Close()
+	}()
 
 	if err = conn.PingContext(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


materialize-sql/std_sql.go
```diff
@@ -46,12 +46,14 @@ func StdSQLExecStatements(ctx context.Context, db *sql.DB, statements []string)
...
-	if err == nil {
-		err = conn.PingContext(ctx)
-	}
 	if err != nil {
 		return fmt.Errorf("connecting to DB: %w", err)
 	}
+	defer conn.Close()
+
+	if err = conn.PingContext(ctx); err != nil {
+		return fmt.Errorf("connecting to DB: %w", err)
```
psFried, line 55: Might help to have a different error message, so we can tell where the errors are coming from.

psFried (approved):
Left one optional suggestion, but LGTM

williamhbaker:
We're already closing the conn at the end of the function (`return conn.Close()`) so I'm not 100% sure how this helps? Was it not getting to that point because of the early return ?

williamhbaker:
So the thought is that we were erroring for some reason and returning early, thus not closing the connection. In that case, wouldn't the connector exit as well? Is the connection left open even if the connector exits, unless we specifically close it?

Put another way, generally I would expect the previous function to have always closed the connection, unless the entire connector was getting to a particular point and crashing with an error. So, maybe that was happening, just wanting to make sure I'm understanding it.

mdibaiee:
@williamhbaker okay after more checking, here is what I have found:

1. When sending a `Ping` to a stopped warehouse that is "about to wake up", we wake it up with our ping, and then when we continue with our code, we end up with a leaked connection. See https://github.com/databricks/databricks-sql-go/issues/198
2. This does not happen if the warehouse is already awake when we send the ping, at least in the case of Servleress warehouses, with the new code I have implemented, I am not able to reproduce the leaked connection, while without this patch I can consistently reproduce the leaked connection
3. On "Pro" warehouses however, even with this code, I see a leaked connection, even though we send Ping after the warehouse is up and running. I haven't been able to resolve this one yet.

The customer that is facing this issue is using Serverless warehouses so I would like to get this out to them and see if we can follow-up with Databricks to fix the underlying issue.

materialize-sql/std_sql.go
```diff
@@ -46,20 +46,23 @@ func StdSQLExecStatements(ctx context.Context, db *sql.DB, statements []string)
 	// through a single connection. This allows a driver to explicitly run
 	// `BEGIN;` and `COMMIT;` statements around a transactional operation.
 	var conn, err = db.Conn(ctx)
-	if err == nil {
```
williamhbaker, line 49: Do we still need these changes? The changes in `materialize-databricks` generally make sense. This change seems unnecessary, since we we already doing `conn.Close()` (and actually returning errors from it) unless there was some other error that would crash the connector.

williamhbaker (approved):

materialize-sql/std_sql.go
mdibaiee: @williamhbaker I think the best practice is to use `defer` in case of an early return to make sure we explicitly cleanup resources. That said, I found a way to return the error from `conn.Close` in defer so we can keep the best of both worlds.

Initially when I opened this pull-request I saw changes in how the warehouse waking up and the leaked connections behaved by addition of this, although I have a hard time reproducing that again... so I'd like to keep these.

PR #1431, opened 2024-03-27
source-sqlserver: Query job status during replication diagnostics

willdonnelly:
**Description:**

Since SQL Server replication works by issuing polling queries and most ways that can fail result in errors, a "silent failure" mode seems most likely to come about if the CDC agent job whose job it is to tail the WAL and copy changes into change tables simply isn't running.

The new diagnostic logic is untested, but it will only be run in the event that replication is already not working as intended, so the risk of breakage is minimal here and I intend to try it out in production shortly after it's live.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1431)


williamhbaker (approved):
LGTM

PR #1438, opened 2024-04-01
Adding source-hubspot-native connector

Luishfs:
**Description:**

Adding `source-hubspot-native` connector

https://github.com/estuary/flow/pull/1440

**Notes for reviewers:**

Tests were made using a custom Hubspot account.
Each stream was tested one at a time with each stream. In order
to allow for the cursor to catch recent data, my `cut_off`/`start_date` variables
were set using ` - timedelta(days=365)` .

Pro streams data could not be validated. Because of that, they were passed on the base V3 
model which should handle all of their pagination cases. ( workflows needed a special pagination function
and resource).
Pro streams are:

- goal_targets
- feedback_submissions
- workflows

## Streams Interval

|Stream  | Interval  |
|--|--|
|Companies  | 30 seconds  |
|Contacts  | 30 seconds  |
|Deals  | 30 seconds  |
|Engagements  | 30 seconds  |
|Contact Lists  | 1 minute |
|Contact Lists Membership  | 1 minute  |
|Subscription Changes  | 1 minute  |
|Email Events  | 1 minute  |
|Ticket Pipelines  | 1 minute  |
|Deal Pipelines  | 1 minute  |
|Campaigns  | 1 minute  |
|Engagements Calls | 30 seconds  |
|Engagements Emails  | 30 seconds  |
|Engagements Tasks  | 30 seconds  |
|Engagements Notes | 30 seconds  |
|Goal Targets  | 30 seconds  |
|Line Items  | 30 seconds  |
|Tickets  | 30 seconds  |
|Email Subscriptions  | 30 seconds  |
|Marketing Forms  | 30 seconds  |
|Owners  | 30 seconds  |
|Properties  | 1 day  |
|Feedback submissions  | 30 seconds  |
|Marketing Emails  | 30 seconds  |
|Workflows  | 30 seconds  |
|All Custom objects  | 1 minute  |

Hubspot imposes a API rate limit of 100 requests per 10 seconds.
V3 Objects can require more than one request per stream, this happens because of the "Associations" and "batch" options, which uses more API requests than usual.
Because of that, a interval of 30 seconds was added, so that smaller streams ( like Companies or Campaigns ) don't add-up to the rate limit.

V1 Objects that can require a longer capture time were set to 60 seconds so that on general their capture can happen after the quicker V3 Object and dont trigger 429 responses that often. Since they can usually take hours to occur, real-time transfers are already blocked, but a longer wait time period can harm smaller clients that dont require long hours of capture.

Finally, Properties were set to 1 day. They are a really small stream that does not require real-time processing and wont affect usage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1438)


source-hubspot-native/test.flow.yaml
```diff
@@ -7,35 +7,121 @@ captures:
...
-          # - "--listen"
-          # - "0.0.0.0:5678"
-          # - "--wait-for-client"
           - "-m"
           - source_hubspot_native
         config: config.yaml
     bindings:
       - resource:
           name: companies
+          interval: PT420S
```
williamhbaker, line 16: What's the reason for adding this explicit interval to the test configuration?

source-hubspot-native/tests/snapshots/source_hubspot_native_tests_test_snapshots__capture__stdout.json
```diff
@@ -1,5574 +1,2081 @@
 [
   [
-    "acmeCo/companies",
+    "acmeCo/marketing_emails",
     {
+      "activeDomain": "45386308.hs-sites.com",
```
williamhbaker, line 5: There's a lot changing in this snapshot. Understandably the new resources will add to it, but the churn on the pre-existing outputs is hard to digest. Is any of the existing output changing, or is just all from shifting the ordering of the things around?

source-hubspot-native/tests/snapshots/source_hubspot_native_tests_test_snapshots__discover__stdout.json
```diff
@@ -747,10 +751,3088 @@
...
+          "title": "Hubspotobject",
+          "type": "string"
+        }
+      },
+      "title": "Property",
+      "type": "object",
+      "x-infer-schema": true
+    },
+    "key": [
+      "/_meta/row_id"
```
williamhbaker, line 3828: `_meta/row_id` needs to be a required property for this to work well with most materializations. So `_meta` needs to be required,  and in `_meta` `row_id` needs to be required.

Technically there is a default value for it set so it might work without this, but I think we might as well set the field as required since we know it will always be there.

source-hubspot-native/tests/test_snapshots.py
```diff
@@ -20,19 +20,6 @@ def test_capture(request, snapshot):
     assert result.returncode == 0
     lines = [json.loads(l) for l in result.stdout.splitlines()[:50]]
 
-    for l in lines:
-        _collection, record = l[0], l[1]
-
-        for m in ["properties", "propertiesWithHistory"]:
-            for prop in [
-                "hs_time_in_lead",
```
williamhbaker, line 28: I'm curious as to why we don't need to do this anymore?

source-hubspot-native/source_hubspot_native/resources.py
```diff
@@ -1,16 +1,19 @@
 from datetime import datetime, UTC, timedelta
-from typing import AsyncGenerator, Awaitable, Iterable
+from typing import AsyncGenerator, Awaitable, Iterable, Dict
 from logging import Logger
 import functools
 
 from estuary_cdk.flow import CaptureBinding
 from estuary_cdk.capture import Task
-from estuary_cdk.capture.common import Resource, LogCursor, PageCursor, open_binding
+#from estuary_cdk.capture.common import Resource, LogCursor, PageCursor, open_binding
```
williamhbaker, line 8: Why is this commented out?

source-hubspot-native/source_hubspot_native/resources.py
```diff
@@ -22,56 +25,344 @@
...
+    return common.Resource(
+        name=cls.NAME,
+        key=cls.PRIMARY_KEY,
+        model=cls,
+        open=open,
+        initial_state=ResourceState(
+            inc=ResourceState.Incremental(cursor=started_at),
+            backfill=ResourceState.Backfill(next_page=None, cutoff=started_at),
+        ),
+        initial_config=ResourceConfig(name=cls.NAME, interval=timedelta(minutes=7)),
```
williamhbaker, line 165: I asked about this elsewhere in a snapshot output as well - I'd like to know more about the default time delta of 7 minutes.

source-hubspot-native/source_hubspot_native/resources.py
```diff
@@ -22,56 +25,344 @@
...
+        schema_inference=True,
+    )
+
+def subscription_object(
+    cls: type[V1CRMObject], http: HTTPSession, fetch_recent: FetchRecentFn
+) -> common.Resource:
+
```
williamhbaker, line 348: The comment for `workflow_object` is nice - as I'm reading this I am wondering why `subscription_object` exists, so a similar comment explaining that would be appreciated.

source-hubspot-native/source_hubspot_native/resources.py
```diff
@@ -115,14 +406,47 @@ def open(
...
+            name=Names.properties, interval=timedelta(minutes=7)
         ),
         schema_inference=True,
     )
+
+def custom_objects(
+    cls: type[CRMObject], http: HTTPSession
+) -> common.Resource:
+    
+    def open(
```
williamhbaker, line 425: A comment (is that what they are called? docstrings is maybe the right term?) on this function would be nice too.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -51,16 +83,27 @@ async def fetch_page(
...
+        }
+        if len(cls.ASSOCIATED_ENTITIES) == 0:
+            del input['associations']
+    else: 
+        properties = await fetch_properties(log, cls, http)
+        property_names = ",".join(p.name for p in properties.results if not p.calculated)
+
+        input = {
+            "associations": ",".join(cls.ASSOCIATED_ENTITIES),
+            "limit": 2,  # 50, # Maximum when requesting history. TODO(johnny).
```
williamhbaker, line 101: A limit of 2 seems really small? Not sure if it was set at that before though.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -201,6 +355,41 @@ async def fetch_changes(
 
     if recent:
         yield recent[-1][0]
+    else:
+        return
+
+
+async def fetch_changes_no_batch(
```
williamhbaker, line 362: Can you comment more on what this function is for? Coming in without a deep understanding of the pre-existing code, I am struggling to understand why this is needed as an alternative to `fetch_changes`. I see that it does things a little differently and that it doesn't have that final `fetch_batch_with_associations` part, but I need a little help understanding what the reasoning for that is.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -271,14 +460,260 @@ async def fetch_recent_tickets(
...
+
+
+    result = V1CustomCRMObject2.model_validate_json(
+        data
+    )
+
+    for item in result.lists:
+        item['listId'] = str(item['listId'])
+
+    if result.hasMore is False:
```
williamhbaker, line 491: Other similar seeming functions use `result.has_more and result.time_offset` as the return value for the `PageCursor`, seemingly as an alternative for this if/else. Would that work for these too?

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -271,14 +460,260 @@ async def fetch_recent_tickets(
...
+        (r.updatedAt, str(r.id))
+        for r in result.results
+    ), None and None
+
+async def fetch_recent_contacts_lists(
+    log: Logger, http: HTTPSession, since: datetime, page: PageCursor
+) -> tuple[Iterable[tuple[datetime, str, Any]], PageCursor]:
+
+    url = f"{HUB}/contacts/v1/lists"
+    params = {"count": 100, "offset": page} if page else {"count": 250}
```
williamhbaker, line 479: We're using `{"count": 1}` elsewhere if there is no pre-existing page cursor. Why 250 here?

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -271,14 +460,260 @@ async def fetch_recent_tickets(
     log: Logger, http: HTTPSession, since: datetime, cursor: PageCursor
 ) -> tuple[Iterable[tuple[datetime, str]], PageCursor]:
 
-    url = f"{HUB}/crm-objects/v1/change-log/tickets"
-    params = {"timestamp": int(since.timestamp() * 1000) - 1}
```
williamhbaker, line 275: Is there not a way to pass the timestamp parameter anymore? Do we need to filter the result on `since` as we will now I assume get every single ticket each time?

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -271,14 +460,260 @@ async def fetch_recent_tickets(
...
+        data
+    )
+
+    for item in result.lists:
+        item['listId'] = str(item['listId'])
+
+    if result.hasMore is False:
+        return (
+            (_ms_to_dt(r['updatedAt']), str(r['listId']), r)
+            for r in result.lists if _ms_to_dt(r['updatedAt']) > since
```
williamhbaker, line 494: This filtering conceptually makes since to me...since we want things after `since`, only emit documents that come after `since`. But I don't know why we don't have that on others, like `fetch_recent_contacts`? It seems they are doing pretty much the same thing. How does that work for `fetch_recent_contacts`, but it wouldn't work here?

williamhbaker (commented):
Made some initial comments!

source-hubspot-native/test.flow.yaml
Luishfs: If i got it right, when a user does not generate their own bindings, we generate then by the default `common.Resource` resource, and the default wait value for estuary is 7 minutes. That's why this specific value at this place ( since discovery would see that no interval was set, i've added that ) .
Does that make sense?

source-hubspot-native/tests/snapshots/source_hubspot_native_tests_test_snapshots__capture__stdout.json
Luishfs: Lots of shifiting around since this snapshot was first added when we had only 4 streams

source-hubspot-native/tests/snapshots/source_hubspot_native_tests_test_snapshots__discover__stdout.json
Luishfs: To solve the second requirement, i did a modification at estuary-cdk in this commit 344be6d90aa7b0f286f210242c7afa9f77c1279d

The first one ( add `_meta/row_id` ) seem a bit harder since its already required by the `BaseDocument` model, and i don't really know how to make both of then required since the models listed in models.py already require the BaseDocument model, could you help me to solve it?

source-hubspot-native/source_hubspot_native/resources.py
Luishfs: This was for code standardization, i saw that google-sheets-native was using `from estuary_cdk.capture import common` and decided that it would be best to keep everything stardard from this initial point forward, nothing important

source-hubspot-native/source_hubspot_native/resources.py
Luishfs: I believe to have answered that on one of the more initial comments, please tell me if that explanation is enough

source-hubspot-native/source_hubspot_native/resources.py
Luishfs: A comment for this specific function was added here a484280098b57da301ffc5bc09ab89fa835fc8d0

source-hubspot-native/source_hubspot_native/resources.py
Luishfs: A comment for this specific function was added here a484280098b57da301ffc5bc09ab89fa835fc8d0

source-hubspot-native/source_hubspot_native/api.py
Luishfs: They are being handled by the fetch_changes_no_batch function, more on how it works in their docstring added here a484280098b57da301ffc5bc09ab89fa835fc8d0

source-hubspot-native/source_hubspot_native/api.py
Luishfs: Hubspot limits the amount of lists by 250 ( if you have Pro/Enterprise it seems to handle more), and data from this endpoint is really small. If no page is given, instead of creating a pagination object with 100 data and have it pass 3 times, i just saw more simpler to handle all 250 lists at once.

If there are more lists, pagination will occur normally

source-hubspot-native/source_hubspot_native/api.py
Luishfs: So, when developing this i had 2 main thoughts in my head ( Hubspot [docs](https://legacydocs.hubspot.com/docs/methods/tickets/get-ticket-changes) for reference):

1. If im using this endpoint to trigger which tickets have changed, i will not get all tickets. They do not specify what is a change, and i dont think a  ticket creation would reach this.
2. Even if it did, when i select the timestamp i would never allow for a """full_refresh""" pattern where i get all data. From what i understood, the since is a way of us understading where the last data parsed was set in history, not as means of getting a specific timeframe

Luishfs: > 2\. the `since` is a way of us understading where the last data parsed was set in history, not as means of getting a specific timeframe

just further explaining this quote, i believe the `timestamp` parameter would return to me 'timestamp - 24 >= data => timestamp`  as is cited in the docs

source-hubspot-native/source_hubspot_native/api.py
Luishfs: This is a really clever idea from johnny, were we call this [endpoint](https://legacydocs.hubspot.com/docs/methods/lists/get_list_contacts_recent)  where it formats the results in order of more recent to less recent. Then, the `fetch_recent_contacts` method gets when this data has been updated and its Id and sents it to the `fetch_changes` function. This function is responsible to filter these timestamps with our `cut_off` / `since` value, and the ones that passes are then collected using batch functions

source-hubspot-native/tests/test_snapshots.py
Luishfs: Im not sure what this did, but it was made to remove these specific timestamps specifically from v3 objects properties, i didnt deem it necessary as there were no comments explaining why ( my idea was that this was probably breaking tests? )

source-hubspot-native/source_hubspot_native/api.py
Luishfs: Yes, raised in 0b549e9f4cdfafdc3aadf8255959108dfb071aab to be the default value. Will leave johnny comments for future reference in case the limit number needs to be revised

Luishfs:
@williamhbaker  I believe i've answer all the initial comments!

source-hubspot-native/tests/snapshots/source_hubspot_native_tests_test_snapshots__discover__stdout.json
williamhbaker: I don't think we want `row_id` to be required in `_meta` across the board for all connectors, since some of them won't have a concept of a `row_id` and will just use `_meta` to report `op`.

So, given that there is not a straightforward way to set this field as required and it _does_ have a default value which will allow it to be materialized (I actually tested this with our `source-google-sheets-native` connector, which has the same kind of schema), I'll retract my statement that this needs to be changed, and we should leave it alone for now and back out the change in https://github.com/estuary/connectors/commit/344be6d90aa7b0f286f210242c7afa9f77c1279d

source-hubspot-native/tests/test_snapshots.py
williamhbaker: OK yeah, I guess if the tests work repeatedly without it then it shouldn't be a problem.

source-hubspot-native/source_hubspot_native/resources.py
williamhbaker: OK, having it removed entirely seems good, just didn't want to leave a mysterious line of commented code in. But it looks good now.

source-hubspot-native/test.flow.yaml
williamhbaker: Yeah, that does make sense. I am wondering how/why these bindings were originally set to have no explicit `interval`. I don't 100% know what happens in that case, and have raised the question in the #saas-connectors slack channel.

And 7 minutes is a pretty weird value, lol. We use 30 seconds in a lot of other places, or maybe 5 minutes, but I'm not sure about a default wait value for estuary of 7 minutes?

Luishfs: I dont know why but in my head 7 minutes was the default interval. Really don't remember where that came from ( probably need my meds ). @williamhbaker Should i leave the default interval of 30 seconds?

source-hubspot-native/tests/snapshots/source_hubspot_native_tests_test_snapshots__discover__stdout.json
Luishfs: Got it!
Changes were reverted here 051e02e8056c6f40d56685a485d3ba9fc8674bed

source-hubspot-native/test.flow.yaml
williamhbaker: 30 seconds seems like a reasonable default interval for bindings that do not otherwise have restrictive rate limits, and are incremental in nature. We wouldn't want to re-fetch a huge snapshot every 30 seconds for example, so bindings like that should have larger intervals. I'll leave it to you to decide on the intervals based on your knowledge of how hubspot works, since I'm not very familiar with it other than these general statements.

williamhbaker: Another note is that I don't know why these pre-existing bindings had no interval at all, and were set to just re-fetch as fast as possible. Maybe there's a good reason for that?

dyaffe: I think because hubspot lets you and we can get real-time updates this way.

Luishfs: @dyaffe @williamhbaker  Hubspot limits us by 100 requests every 10 seconds, and given that we have 27+ streams, i think having a 30+ seconds breather for each stream would really help to not reach the limit. I will test with some values and get more data on this matter

Luishfs: @williamhbaker Added intervals in 4171ca4368a0cd29e8af7b77533686cb6c80a249

williamhbaker: I'm still seeing this `PT420S` interval in `test.flow.yaml`, which I don't think is what we want.

Luishfs: @williamhbaker Forgot to run discover, rebased the commit here da3558d73fb91d5fa30662cc2462b84f6fe451fb

Luishfs: @williamhbaker Forgot to run discover, rebased the commit here fda33783a560759e3bdc40251a9633277d602c9f

Luishfs:
@williamhbaker Tests are now pasing

source-hubspot-native/test.flow.yaml
dyaffe: @Luishfs just one more push here.  We spoke about having 27 streams and some of them have 3 api calls per stream.  The worst case scenario seems to me that every single one has 3 calls.  If that's the case, we could lower the limit from 30s to 10s with room to spare since 10s would be at max 81 API requests every 10s.

Let me know if I'm wrong for any reason.

source-hubspot-native/test.flow.yaml
```diff
@@ -7,35 +7,118 @@ captures:
...
+          name: owners
+          interval: PT30S
+        target: acmeCo/owners
+      - resource:
+          name: workflows
+          interval: PT30S
+        target: acmeCo/workflows
       - resource:
           name: properties
-          interval: P1D
```
williamhbaker, line 36: Just spot checking: The table in the PR description says that this is supposed to be 1 day. But its 30 seconds here. Which one is it supposed to be?

williamhbaker (approved):
LGTM. See comment about interval on one of the streams. I didn't go through and check every single one line by line against the table in the PR description, but I did find that discrepancy at a glance. I trust that you'll make sure things match up before merging.

source-hubspot-native/test.flow.yaml
Luishfs: Updated at ae65e51c85616092acd08ece0c3ace515985e4d3

dyaffe (approved):

PR #1440, opened 2024-04-01
source-firestore: Error message improvements

willdonnelly:
**Description:**

We fairly often see a fatal error message from `source-firestore` which reads simply:

    unexpected TargetChange.REMOVE: The datastore operation timed out, or the data was temporarily unavailable

And we should really be mentioning the name of the collection on which this error occurred. This commit should fix that, and also makes sure that we mention the collection name on backfill errors as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1440)


williamhbaker (approved):
LGTM

PR #1444, opened 2024-04-03
m-snowflake: increase retry count and delay

mdibaiee:
**Description:**

- Allow more time for Snowpipe results to show up
- We have a task which we have seen that requests sent at 11:06 show up in results only at 11:10, that's 4 minutes later. Here I am allowing up to 5 minutes.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1444)

### DIFF
--- a/tests/materialize/materialize-snowflake/setup.sh
+++ b/tests/materialize/materialize-snowflake/setup.sh
@@ -14,7 +14,7 @@ export SNOWFLAKE_AUTH_TYPE="${SNOWFLAKE_AUTH_TYPE}"
 # if auth type is user_password
 export SNOWFLAKE_USER="${SNOWFLAKE_USER:-}"
 export SNOWFLAKE_PASSWORD="${SNOWFLAKE_PASSWORD:-}"
-if [ -z "${${SNOWFLAKE_PRIVATE_KEY+x}}" ]; then 
+if [ -n "${SNOWFLAKE_PRIVATE_KEY:-}" ]; then 
   # if auth type is jwt
   export SNOWFLAKE_PRIVATE_KEY="$(cat ${SNOWFLAKE_PRIVATE_KEY:-} | jq -sR . | sed -e 's/^"//' -e 's/"$//')"
 fi
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -619,8 +619,9 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
 	// until they have all been successful, or an error has been thrown
 
 	// If we see no results from the REST API for `maxTries` iterations, then we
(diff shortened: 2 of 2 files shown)


tests/materialize/materialize-snowflake/setup.sh
```diff
@@ -14,7 +14,7 @@ export SNOWFLAKE_AUTH_TYPE="${SNOWFLAKE_AUTH_TYPE}"
 # if auth type is user_password
 export SNOWFLAKE_USER="${SNOWFLAKE_USER:-}"
 export SNOWFLAKE_PASSWORD="${SNOWFLAKE_PASSWORD:-}"
-if [ -z "${${SNOWFLAKE_PRIVATE_KEY+x}}" ]; then
```
mdibaiee, line 17: My shell was complaining about this syntax, so I changed it to a different syntax but with the same meaning I think

psFried: Just a drive by comment:
This syntax really is confusing. I have no idea what the `+x` is even doing here. The [bash docs](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) cover `${var:+var}`, but the `:` seems important.

The change from `-z` to `-n` seems like it might change the meaning, though. `-z` means "is zero-length" and `-n` means "is _not_ zero-length".

mdibaiee: I'm not sure about the `+` syntax, but my general understanding of the intent behind this `if` block is to only load the private key if the variable is set, and that's what the new `-n` condition does: only run that block if `SNOWFLAKE_PRIVATE_KEY` is not zero-length

williamhbaker: I'm also remembering that I tried to get the tests to run by having `SNOWFLAKE_PRIVATE_KEY` set but set to an empty string, and then some later part of the test mechanism failing. At any rate, I'd like us to just have a way to run these tests locally using username/password auth, so whatever works for that I think is fine.

williamhbaker (approved):
This seems fine for now, although I increasingly suspect this strategy of using snowpipe isn't going to be 100% rock solid, and we will occasionally see failures due to the retry not being long enough, and we'll have to balance that with the retry being excessively long.

I've had the thought that we actually could use snowpipe streaming, at least in principal, with the post-commit apply pattern. We'd stage a file like we do now, and then the "apply" part would involve reading that file back in and sending the rows through the snowpipe streaming API. The rows would have offsets calculated by the connector that would enable exactly-once delivery on a row-by-row basis. In this way the transactions would be idempotent not from the runtime, but from the connector staging these files like it does now. Not at all suggesting we should do that right now, but I think it could be interesting to explore depending on how snowpipe works out.

PR #1445, opened 2024-04-03
source-firestore: Only retry TargetChange.REMOVE after 60s

willdonnelly:
**Description:**

This *probably* isn't necessary, but one of the challenges with Firestore change streaming is that it's actually not super hard to end up in a permanent failure loop where you keep reading the same documents over and over and accidentally rack up large bills.

In other places where we retry after a failure status we generally wait 60 seconds before trying again, and it's probably prudent to do that for the newly added `TargetChange.REMOVE` retry logic as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1445)


willdonnelly:
Going to just go ahead and merge this because I don't think adding a couple of calls to `time.Sleep()` in the retry path is a remotely controversial change.

PR #1447, opened 2024-04-03
source-firestore: Handle catchup failures after any time interval

willdonnelly:
**Description:**

When streaming replication events from Firestore, sometimes the streaming for a particular high-volume collection gets stuck in a state where we open the stream, and then before becoming caught up Firestore sends us a `TargetChange.REMOVE` event. In such cases it is unlikely that we will ever succeed on a subsequent retry, so we have logic to give up and restart streaming from "now" and go back and re-backfill the collection later.

In the past we've always observed this happening after exactly 10 minutes, so the condition for triggering this recovery behavior includes spending >5m streaming change events before getting cut off. However now we're seeing basically the exact same pattern of behavior occurring after less than two minutes, and the desired response is the same.

So the appropriate fix here is to just remove the time threshold entirely. After this commit, any time we receive a `TargetChange.REMOVE` event before the stream catches up, we'll immediately give up and jump ahead to the latest changes for the collection.

### DIFF
--- a/source-firestore/pull.go
+++ b/source-firestore/pull.go
@@ -678,7 +678,6 @@ func (c *capture) StreamChanges(ctx context.Context, client *firestore_v1.Client
 	var listenClient firestore_pb.Firestore_ListenClient
 	var numRestarts, numDocuments int
 	var isCurrent, catchupStreaming bool
-	var catchupStarted time.Time
 	for {
 		if listenClient == nil {
 			var err error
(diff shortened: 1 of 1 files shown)


willdonnelly:
Going to YOLO-merge this one too. I'm pretty confident in my analysis of the problem, the actual change is trivial, and until this is merged we've got production captures pulling down redundant change data at high speed.

PR #1448, opened 2024-04-03
materializations: don't try to drop nullability if the projection has a default value

williamhbaker:
**Description:**

In e852e80, materializations were made aware of the possibility for default values in the materialized columns, and that we should not try to drop "not null" constraints on columns like this, under the assumption that the destination system is fine with us sending a "null" since it will automatically be populated (in a sense) with that default value.

This didn't take into consideration the fact that Flow projections can also have default values, and in this case we will never see a "null" for these fields from the Flow runtime, and instead will always see that default value. So here too we do not need to drop nullability constraints - the materialized columns can stay "not null" and as long as there is that default value annotation things will work fine.

This is particularly important for materializing nullable collection keys with default values. This is allowed (see 5d53313 as well), and some systems like postgres consider primary key fields non-nullable. Prior to this commit, the connector would try to drop a "not null" constraint on a primary key column that was created from a projection with a default value if the materialized column didn't have a default value. We don't currently create columns with default values based on projections (though maybe we will, later) so this would cause the materialization to fail the next time it tries to do an `Apply`. But now, the materialization will not try to drop a "not null" on a materialized key field as long as the projection has a default value.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1448)

### DIFF
--- a/materialize-boilerplate/apply.go
+++ b/materialize-boilerplate/apply.go
@@ -178,10 +178,12 @@ func ApplyChanges(ctx context.Context, req *pm.Request_Apply, applier Applier, i
 					}
 
 					newRequired := projection.Inference.Exists == pf.Inference_MUST && !slices.Contains(projection.Inference.Types, pf.JsonTypeNull)
-					if !existingField.Nullable && !newRequired && !existingField.HasDefault {
-						// The existing field is not nullable and does not have a default value, but
-						// the proposed projection for the field is nullable. The existing field
-						// will need to be modified to be made nullable.
+					newlyNullable := !existingField.Nullable && !newRequired
+					projectionHasDefault := projection.Inference.DefaultJson != nil
+					if newlyNullable && !existingField.HasDefault && !projectionHasDefault {
+						// The field has newly been made nullable and neither the existing field nor
+						// the projection has a default value. The existing field will need to be
+						// modified to be made nullable since it may need to hold null values now.
 						params.NewlyNullableFields = append(params.NewlyNullableFields, existingField)
 					}
 				} else {


materialize-boilerplate/apply.go
```diff
@@ -178,10 +178,12 @@ func ApplyChanges(ctx context.Context, req *pm.Request_Apply, applier Applier, i
 					}
 
 					newRequired := projection.Inference.Exists == pf.Inference_MUST && !slices.Contains(projection.Inference.Types, pf.JsonTypeNull)
-					if !existingField.Nullable && !newRequired && !existingField.HasDefault {
-						// The existing field is not nullable and does not have a default value, but
-						// the proposed projection for the field is nullable. The existing field
-						// will need to be modified to be made nullable.
+					newlyNullable := !existingField.Nullable && !newRequired
+					projectionHasDefault := projection.Inference.DefaultJson != nil
+					if newlyNullable && !existingField.HasDefault && !projectionHasDefault {
```
mdibaiee, line 183: why do we check for whether the existing field has a default value? If someone updates a collection and projection to make a field nullable, but at the same time add a projection with a default value, we should still allow them to keep their column as non-nullable, no?

williamhbaker: `existingField` refers to the actual column in the destination system. So if they have altered or pre-created their database table with a column that is not null and has a default value, we allow them to keep that column as non-nullable even if the Flow collection field is nullable and does not have a default value.

The situation you're describing is allowed - if someone updates a collection and makes a field nullable but adds a default value to it, we won't drop nullability on the column. We only will drop nullability on a column if the column itself doesn't have a default value, and our Flow projection doesn't have a default value.

mdibaiee (approved):

PR #1450, opened 2024-04-04
Add source-azure-blob-storage

Z33DD:
**Description:**

This implements Azure Blob Storage using the [filesource](https://github.com/estuary/connectors/tree/main/filesource) framework and the [Azure Go SDK](https://github.com/Azure/azure-sdk-for-go).

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1450)


source-azure-blob-storage/go.mod
```diff
@@ -0,0 +1,49 @@
+module github.com/estuary/source-azure-blob-storage
```
williamhbaker, line 1: Our convention has been to have all go connectors use the same top-level `go.mod` file: https://github.com/estuary/connectors/blob/main/go.mod

I'm not necessarily saying that is better or worse than each connector defining its dependencies separately, but its what we've been doing and I don't want to diverge from that right now. So you should be able to add the necessary dependencies to the main `go.mod`.

For example, see [this commit](https://github.com/estuary/connectors/commit/28ee7a8cfcafb123d059359625f8e574fda84973) where we added a filesource connector and updated `go.mod` with what it needed.

source-azure-blob-storage/test.flow.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
williamhbaker, line 1: From what I understand this isn't currently working, so we can get rid of it here.

source-azure-blob-storage/main.go
```diff
@@ -0,0 +1,212 @@
...
+	AzureClientSecret   string         `json:"azureClientSecret"`
+	AzureTenantID       string         `json:"azureTenantID"`
+	AzureSubscriptionID string         `json:"azureSubscriptionID"`
+	StorageAccountName  string         `json:"storageAccountName"`
+	ContainerName       string         `json:"containerName"`
+	Parser              *parser.Config `json:"parser"`
+}
+
+func (c config) Validate() error {
+	if c.AzureTenantID == "" {
```
williamhbaker, line 34: This can be cleaned up by doing something like this: https://github.com/estuary/connectors/blob/cd7c1b12ed2aa450e24953e2d5df18143f12e4ad/materialize-snowflake/config.go#L145-L155

source-azure-blob-storage/main.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/estuary/connectors/filesource"
+	"github.com/estuary/flow/go/parser"
+	pf "github.com/estuary/flow/go/protocols/flow"
+)
+
+//go:embed schema.json
+var schema string
+
+func handleError(err error) {
```
williamhbaker: We need to return errors so that they are bubbled up to the higher level handlers here: https://github.com/estuary/connectors/blob/cd7c1b12ed2aa450e24953e2d5df18143f12e4ad/source-boilerplate/boilerplate.go#L94-L96.

The whole `if err != nil { return err }` thing is one of the most horrible parts of Go, but doing it this way makes our error presentation a lot better and more consistent.

source-azure-blob-storage/main.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+type azureBlobStore struct {
+	client *azblob.Client
+	cfg    *config
+}
+
+// TODO: Test this
+func newAzureBlobStore(cfg config) (*azureBlobStore, error) {
+	os.Setenv("AZURE_CLIENT_ID", cfg.AzureClientID)
```
williamhbaker, line 84: Setting up authentication by setting env vars is pretty weird and I hope there is a cleaner way to do it. Seems like this is relying on the azure equivalent of AWS & friends' "application default credentials". But we don't use that for S3 and are instead able to do something like [this](https://github.com/estuary/connectors/blob/cd7c1b12ed2aa450e24953e2d5df18143f12e4ad/source-s3/main.go#L78-L107), and ideally we could configure the azure client in a similar way.

source-azure-blob-storage/schema.json
```diff
@@ -0,0 +1,69 @@
+{
```
williamhbaker, line 1: I actually like using `go:embed` for this quite a bit, but for consistency I think we should just have the schema definition as a in-line string like we do for [others](https://github.com/estuary/connectors/blob/cd7c1b12ed2aa450e24953e2d5df18143f12e4ad/source-s3/main.go#L279-L348)

_Really_ what we should do is use our [schema-gen](https://github.com/estuary/connectors/tree/main/go/schema-gen) to generate the schema from the config struct tags (see [example](https://github.com/estuary/connectors/blob/cd7c1b12ed2aa450e24953e2d5df18143f12e4ad/source-dynamodb/main.go#L16-L28)), but schema-gen doesn't support things like `oneOf` configurations for auth methods so we haven't done that generally. But that is the eventual goal, and we could then remove all of the hand-written schemas, so best to have them all be consistently represented for now.

Z33DD: My argument for having this in a separated file is because it makes it easier to see the changes coming in git. I did an extra logic to update the `parser` property.

source-azure-blob-storage/.gitignore
```diff
@@ -0,0 +1 @@
+.env
```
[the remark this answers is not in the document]
Z33DD: I'm using dot env files for testing, see `source-azure-blob-storage/main_test.go`

source-azure-blob-storage/schema.json
williamhbaker: There should be a spec test that snapshots the output of the spec, and that is how we will see changes. You've got that test already so that's great.

In this case, I'd like us to follow the convention of all the other connectors that need to use a hand-written schema because they are beyond the current capabilities of `schema-gen`, which is to have the schema defined as a string [like this](https://github.com/estuary/connectors/blob/a54358b5ef852d5103a421e7ad30da3ef3e7190e/source-sftp/main.go#L441-L552). Another viable option would be to extend `schema-gen` to include the features we need; namely `oneOf` configs, so that we could define the schema as part of the configuration. That would be the best long-term solution, but I don't think its reasonable to require that as a condition of this pr unless you are feeling really ambitious 😄 . But I don't want us to introduce _another_ way of declaring these schemas that is basically still a workaround for `schema-gen`.

source-azure-blob-storage/.gitignore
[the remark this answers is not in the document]
williamhbaker: Ideally we would use a local emulator to run these test, so that they can run automatically in CI and not incur any charges that would be associated with a more permanent "real" azure setup.

Could we [Azurite](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=docker-hub%2Cblob-storage) as a local emulator for tests? We use localstack like this for `source-s3`, and have an [advanced endpoint config option](https://github.com/estuary/connectors/blob/a54358b5ef852d5103a421e7ad30da3ef3e7190e/source-s3/main.go#L38) for it. The tests get run [here](https://github.com/estuary/connectors/tree/main/tests/source-s3).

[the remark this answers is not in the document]
Z33DD: Yes I can. I didn't know that Azure has an emulator, thanks for letting me know.

source-azure-blob-storage/main.go
```diff
@@ -0,0 +1,342 @@
...
+// since these are the permissions required by the connector.
+func (az *azureBlobStore) check(ctx context.Context) error {
+	// All we care about is a successful listing rather than iterating on all objects
+
+	maxResults := int32(1)
+	listingOptions := azblob.ListBlobsFlatOptions{MaxResults: &maxResults}
+
+	pager := az.client.NewListBlobsFlatPager(az.cfg.ContainerName, &listingOptions)
+	if !pager.More() {
+		log.Printf("The container '%q' is empty", az.cfg.ContainerName)
```
williamhbaker, line 127: Use `logrus` for logging - we use it everywhere else. Also don't log this line at all because it's probably somewhat common and it would confuse users.

source-azure-blob-storage/main.go
```diff
@@ -0,0 +1,342 @@
...
+	if c.Credentials.ConnectionString != "" {
+		return nil
+	}
+	var requiredProperties = [][]string{
+		{"AzureTenantID", c.Credentials.AzureTenantID},
+		{"AzureClientID", c.Credentials.AzureClientID},
+		{"AzureClientSecret", c.Credentials.AzureClientSecret},
+		{"AzureSubscriptionID", c.Credentials.AzureSubscriptionID},
+		{"StorageAccountName", c.StorageAccountName},
+		{"ContainerName", c.ContainerName},
```
williamhbaker, line 48: Do we need a storage account name and container name if a connection string is specified?

source-azure-blob-storage/main.go
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		return filesource.ObjectInfo{}, err
+	}
+
+	if page == nil {
+		return filesource.ObjectInfo{}, io.EOF
+	}
+
+	for _, blob := range page.Segment.BlobItems {
+		fmt.Println(*blob.Name)
```
williamhbaker, line 203: Get rid of this

williamhbaker (commented):
See inline comments. Also we need a `VERSION` file, which all the other connectors have.

source-azure-blob-storage/main.go
Z33DD: In that case, we need just the container name, thanks for pointing out.

source-azure-blob-storage/main.go
```diff
@@ -0,0 +1,355 @@
...
+	if c.Credentials.ConnectionString == "" {
+		requiredProperties = [][]string{
+			{"AzureTenantID", c.Credentials.AzureTenantID},
+			{"AzureClientID", c.Credentials.AzureClientID},
+			{"AzureClientSecret", c.Credentials.AzureClientSecret},
+			{"AzureSubscriptionID", c.Credentials.AzureSubscriptionID},
+			{"StorageAccountName", c.StorageAccountName},
+			{"ContainerName", c.Credentials.ContainerName},
+		}
+	} else {
```
williamhbaker, line 54: We never will reach this `else` because of the check on the line 41

Z33DD: True, just fixed.

williamhbaker:
Have you run this locally? Loading it up in a local stack the credentials controls don't look right:

image

Also, if I put in a connection string, I get this error when trying to publish:

image

I think the credentials `anyOf` should be a `oneOf`, and maybe other changes to get this working.

Z33DD:
> Have you run this locally? Loading it up in a local stack the credentials controls don't look right:
> image
> 
> Also, if I put in a connection string, I get this error when trying to publish:
> image
> 
> I think the credentials `anyOf` should be a `oneOf`, and maybe other changes to get this working.

I just ran the test, no QA test done. But I figured out this is because the spec was outdated, I already fixed it, but I'm doing a proper QA test before sending it to upstream.

source-azure-blob-storage/.snapshots/TestAzureBlobStore_getConfigSchema
```diff
@@ -0,0 +1,110 @@
...
+		"type": "object",
+		"properties": {
+			"credentials": {
+				"type": "object",
+				"title": "Credentials",
+				"description": "Azure credentials used to authenticate with Azure Blob Storage.",
+				"order": 0,
+				"anyOf": [
+					{
+						"title": "OAuth2 Credentials",
```
williamhbaker, line 13: Ah so is this really an OAuth2 flow kind of authentication? If so we'd probably want to get an OAuth app approved and have it work via our UI. If that's the case, I'm thinking it isn't worth the effort right now and we should only support the "connection string" auth, which is TBH the only one of the auth methods I could figure out myself. We'd still want this to be a `oneOf` (you've got it as `anyOf` right now, which I think we should change before merging to be like [this](https://github.com/estuary/connectors/blob/6ec57528da06bec6846c307fc7b48c651e2ac9c6/source-sftp/main.go#L456)) so that we could expand it in the future by adding to the list, even though there'd only be one option.

williamhbaker (approved):
LGTM % comment about the auth

Z33DD:
No, we don't need to setup our own Estuary app. You can see the docs PR for clarification.

Sent from Proton Mail mobile

-------- Original Message --------

source-azure-blob-storage/.snapshots/TestAzureBlobStore_getConfigSchema
Z33DD: About the OAuth2 authentication, we don't need to create an app by ourselves. The customer needs to. This is the recommended way in the docs, you can see how to set up it and more information on the docs PR here: https://github.com/estuary/flow/pull/1441/
https://github.com/estuary/flow/blob/1c2b31df606ce41956c9e848ed6d5506b749251e/site/docs/reference/Connectors/capture-connectors/azure-blob-storage.md

williamhbaker: Oh I see, thanks! 

Should we make this a `oneOf`? I think we should. since technically it would not be value for a config to match more than one of the options.

PR #1452, opened 2024-04-04
source-mysql: Correctly handle enum values when processing DDL query events

willdonnelly:
**Description:**

When processing DDL query events involving enum (and set) columns we're relying on the Vitess SQL parser giving us the correct list of options for the column types and using that to update our metadata. However, apparently we've been getting single-quoted (and potentially backslash-escaped) strings back from the parser and so our metadata ends up in an incorrect state which disagrees with our own discovery logic and the results of a subsequent backfill.

The behavior on Vitess's part appears to be deliberate. The single-quoted strings are actually tokenized into plain-old strings and then explicitly turned back into single-quoted strings via the `encodeSQLString()` helper function when the parse tree is built. I have no idea why this would ever be desirable, but whatever.

The fix is simple, we just need to apply the same single-quoted-string-unquoting function to these enum values here as we already do when parsing discovery results involving enums.

**Workflow steps:**

Future DDL queries involving enum columns will result in correct metadata updates (no more incorrect `'foo'` single-quotes around the actual options).

Any captures which previously processed a DDL query involving an enum column may have incorrect metadata for that column, however they should generally not capture incorrect data as a result because the first time they actually process a change event for that column the resulting document will fail validation.

(Actually, I guess that implicitly assumes they're using automatic discovery so that the collection schema knows the correct options so the incorrect document fails validation. Captures which don't automatically update their schema to have the new enum column could hypothetically allow an incorrect value to slip past)

I intend to go spend some time tomorrow investigating whether any production captures (other than the two which have already encountered errors and been re-backfilled manually) have incorrect metadata as a result of this. I suspect not, because DDL queries involving enum columns seem like they ought to be a reasonably rare edge case, but it's worth taking the time to be sure.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1452)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -463,7 +463,7 @@ var enumValuesRegexp = regexp.MustCompile(`'((?:''|\\.|[^'])+)'(?:,|$)`)
 
 // enumValueReplacements contains the complete list of MySQL string escapes from
 // https://dev.mysql.com/doc/refman/8.0/en/string-literals.html#character-escape-sequences
-// plus the `”` repeated-single-quote mechanism.
+// plus the `'​'` repeated-single-quote mechanism.
 var enumValueReplacements = map[string]string{
 	`''`: "'",
 	`\0`: "\x00",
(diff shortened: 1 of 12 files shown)


source-mysql/discovery.go
```diff
@@ -492,13 +492,22 @@ func parseEnumValues(details string) []string {
...
-		var opt = match[1]
+		opts = append(opts, decodeMySQLString(match[1]))
+	}
+	return opts
+}
+
+// decodeStringMySQL decodes a MySQL-format single-quoted string (including
+// possible backslash escapes) and returns it in unquoted, unescaped form.
+func decodeMySQLString(qstr string) string {
+	if strings.HasPrefix(qstr, "'") && strings.HasSuffix(qstr, "'") {
```
williamhbaker, line 503: probably dumb question: is it possible to have inputs like `'this''`? Where the decoded version should be `this'` but here we'd return `this`

williamhbaker (approved):
LGTM

source-mysql/discovery.go
willdonnelly: That specific input shouldn't be possible. The strings we're "decoding" here are actually fully decode within the Vitess tokenizer and then _re-encoded_ using their `encodeSQLString()` operation which will only ever produce correctly formed outputs (in this case the extra single-quote would be backslash-escaped so it'd be `'this\''`).

Even if that input did occur, we only strip off a single instance of the leading and trailing single-quote before decoding any escapes within the string, so it would product the correct result (inasmuch as a malformed input has a correct result at all).

PR #1457, opened 2024-04-05
materialize-snowflake: round-robin try pipes

mdibaiee:
**Description:**

- Instead of retrying 5 minutes for each binding before moving on to the next binding, retry in a round-robin fashion across bindings by looping through bindings for each retry attempt. Reduce the time between retry attempts and reduce the total time spent retrying

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1457)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -620,10 +620,16 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
 
 	// If we see no results from the REST API for `maxTries` iterations, then we
 	// fallback to asking the `COPY_HISTORY` table. We allow up to 5 minutes for results to show up in the REST API.
-	var maxTries = 60
-	var retryDelaySeconds = 5 * time.Second
-	for pipeName, pipe := range pipes {
-		for tries := 0; tries < maxTries; tries++ {
+	var maxTries = 20
+	var retryDelaySeconds = 3 * time.Second
+	for tries := 0; tries < maxTries; tries++ {
+		for pipeName, pipe := range pipes {
+			// if the pipe has no files to begin with, just skip it
+			// we might have processed all files of this pipe previously
+			if len(pipe.files) == 0 {
+				continue
+			}
+
 			// We first try to check the status of pipes using the REST API's insertReport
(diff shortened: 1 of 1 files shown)


materialize-snowflake/snowflake.go
```diff
@@ -620,10 +620,16 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
 
 	// If we see no results from the REST API for `maxTries` iterations, then we
 	// fallback to asking the `COPY_HISTORY` table. We allow up to 5 minutes for results to show up in the REST API.
-	var maxTries = 60
-	var retryDelaySeconds = 5 * time.Second
-	for pipeName, pipe := range pipes {
-		for tries := 0; tries < maxTries; tries++ {
+	var maxTries = 20
+	var retryDelaySeconds = 3 * time.Second
+	for tries := 0; tries < maxTries; tries++ {
```
williamhbaker, line 625: How are we getting out of this `tries` loop?

mdibaiee: Scenarios for exiting the loop:
1. We retry every pipe `maxTries` times and don't find anything, in that case the loop naturally exits with `tries < maxTries` becoming false
2. We successfully find the files we are looking for for all pipes, in that case the inner pipes loop will just `continue` through all pipes and we iterate until `tries < maxTries` no longer holds and exit. We could put a shortcut for this where if we recognise that the inner loop exit without processing any pipes, we just break out of the tries loop.
3. We have some pipes that are giving us results through the API but not all of the files we are looking for. In that case each time this happens we reset the `tries` counter to zero to keep trying until we find the files we are looking for. However this does not happen repeatedly, since once we have seen some results from the API, we paginate to the next page, which won't recur with the same records.
4. Another case of us looping more than `maxTries` is when we hit no results in the API until `maxTries - 1` tries, and then on the last try we ask for results from `COPY_HISTORY` table. In this case, if we find that the files are there, but their status is "Load in progress", we will add an additional retry by decrementing `tries` by one and trying again. I don't expect this to happen for long either, files that are being loaded are supposed to succeed/fail at some point and so will we with them.

williamhbaker (approved):

PR #1458, opened 2024-04-08
CDK updates

jshearer:
**Description:**

Updates to the CDK that arose out of Netsuite work:

* Relax `PageCursor` to accept negative ints. This is just a typing change, but it was required because some page cursor values in Netsuite are genuinely negative ints
* Teach `FetchChangesFn` and `FetchPageFn` to accept documents for different bindings by way of yielding `AssociatedDocuments`. This is needed in order to enable loading associated documents, which we need to do for e.g `TransactionAccountingLine`
* Pass `open()` information about all bindings in order for it to find the bindings of its associations (specifically so that it knows which binding index to emit documents for when loading associated docs)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1458)

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -276,6 +276,12 @@ class Resource(Generic[_BaseDocument, _BaseResourceConfig, _BaseResourceState]):
             int,
             ResourceState,
             Task,
+            list[
+                tuple[
+                    CaptureBinding[_ResourceConfig],
+                    "Resource[_BaseDocument, _ResourceConfig, _ResourceState]",
+                ]
+            ]
         ],
         None,
     ]
@@ -394,6 +400,7 @@ def open(
                 index,
                 state,
                 task,
+                resolved_bindings
             )
 
     return (response.Opened(explicitAcknowledgements=False), _run)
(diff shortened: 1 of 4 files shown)


estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -552,6 +568,10 @@ async def _binding_backfill_task(
             if isinstance(item, BaseDocument):
                 task.captured(binding_index, item)
                 done = True
+            elif isinstance(item, AssociatedDocument):
+                # TODO(jshearer): Assert that the binding index exists (and is enabled?)
```
jgraettinger, line 572: nit: disabled bindings are not represented in the built CaptureSpec and cannot be present in `resolved_bindings`

jgraettinger (approved):
LGTM

estuary-cdk/estuary_cdk/capture/common.py
jshearer: What does the runtime do when it encounters a document for a binding index that doesn't exist? You could theoretically just `yield AssociatedDocument(doc=..., binding=999999)`. I suppose if the runtime errors if it sees an invalid binding index then extra validation here is unneccesary

jgraettinger: https://github.com/estuary/flow/blob/master/crates/runtime/src/capture/protocol.rs#L371-L375

PR #1464, opened 2024-04-09
Adding Source-Twilio  

Luishfs:
**Notes for reviewers:**

## Capture diff
No capture diff was created (lack of a official account + a lot of data would need to be redacted since it could be sensible, i think it would be best for future maintanance to always test capture and dont rely on the capture snapshot to raise errors). Capture tests were executed in a test account and end-to-end test was also validated.

## Discover Jsondiff
[jsondiff link](https://jsondiff.com/?left=https://raw.githubusercontent.com/estuary/connectors/defe4e4708ba86544eae7d06caa8b5a7aaa37c42/source-twilio/tests/snapshots/snapshots__discover__capture.stdout.json&right=https://raw.githubusercontent.com/estuary/connectors/a8c5065c72f087a04297fe98ce11c3d3a62c00ea/source-twilio/tests/snapshots/snapshots__discover__capture.stdout.json) to capture snapshot ( left = image snapshot, right = PR/connector snapshot)

## Spec gitdiff
[gitdiff](https://github.com/estuary/connectors/pull/1464/commits/10c771b8a67b5c6199090ef75c2402693570279e#diff-742884931cf98303bb12bf01915f11815b55612a7a84053a072e440265ff99dd) link

## Changes to the connector
One change was made from the original connector [here](https://github.com/estuary/connectors/blob/cf39e267efa54cfd360afe5a2e0cc28012ffcf05/source-twilio/source_twilio/streams.py#L232-L249). The UsageRecords stream endpoint will return a 404 in case no data was found on that date. To prevent this, i've added a simple try/except waiting for a `HTTPError` raise, if so, the stream will move to the next stream_slice.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1464)


williamhbaker:
There are discovery output artifacts included here (all the bindings and schemas in `acmeCo/`), but there are no capture output snapshot tests. We need to either:

1) Add capture snapshot output tests, including the diff between the existing connector's output and this new one's output or
2) Don't check in all of the discovery output stuff, and explain why we aren't adding capture snapshot output tests, and how the outputs where manually verified to be compatible between the existing connector and this one

Luishfs:
@williamhbaker The PR description is now updated and the import connector is already generating diff snapshots, also tested it end-to-end and no issue was found.
Could you please review it againt?

source-twilio/source_twilio/streams.py
```diff
@@ -0,0 +1,683 @@
...
+        try:
+            for record in super().read_records(sync_mode, cursor_field, stream_slice, stream_state):
+                record[self.cursor_field] = pendulum.parse(record[self.cursor_field], strict=False).to_iso8601_string()
+                unsorted_records.append(record)
+            sorted_records = sorted(unsorted_records, key=lambda x: x[self.cursor_field])
+            for record in sorted_records:
+                if record[self.cursor_field] >= self.state.get(self.cursor_field, self._start_date):
+                    self._cursor_value = record[self.cursor_field]
+                    yield record
+        except HTTPError:
```
williamhbaker, line 246: Should we be ignoring _all_ HTTP errors here, or just the `20404` code? I think it would be better to minimally ignore errors, except for the known case, so that other unexpected conditions would be loudly announced to us. So I'd prefer only passing on the `20404` error case rather than all HTTP errors in general unless there's a good reason not to do that.

Luishfs: Fixed in 96cfc2e427ad787ad8a7f5f560963b3ae181dbbd

Just to give my thought process in that:
I've read throught their [error docs](https://www.twilio.com/docs/api/errors) ( for about the 29999 code errors, after that it gets really product specific ). i know airbyte-cdk will handle 429 and 5XX error responses, so the 4xx were still not handled and to the point where the incremental streams starts to run (where this check is made), all errors related to invalid authentication  or similar ( like 20008 and on ) should have already been called, so i thought passing all HTTPError exceptions would mostly be used on 4XX TCP/Data not found errors

Luishfs: But i agree it would be good to raise all other errors and see if that is the case

williamhbaker (approved):
LGTM - Thanks!

Alex-Bair:
Rebased onto main. Added a `VERSION` file at the "add to CI" commit.

PR #1466, opened 2024-04-10
source-mongodb: incremental backfills

### LINKED ISSUE #1491: MongoDB: fix backfills to support individual bindings while keeping up with change stream

### LINKED ISSUE #1492: MongoDB: allow connector to work with more restricted permissions and ops logs

williamhbaker:
**Description:**

Implements incremental backfills for `source-mongodb`, that allow it to process backfills and change stream reading in a pseudo-concurrent fashion. The primary benefit to this is that we won't require long oplog retention periods for backfilling large collections, since we can always ensure that the change stream is "caught up" and not lose its position. The also means we don't need access to the `oplog.rs` collection, which frequently trips users up or completely prevents us from capturing their database due to access restrictions.

The individual commit messages provide further details. The commit history is a bit of an evolutionary tale, since I had to re-work an initial bit of implementation, and it didn't seem worth it to try to re-write the history to erase all evidence of its existence.

The general highlights are this:
- The incremental backfill part, with some nice enhancements about change stream resumption and simplifying the connector to only ever use database-level change streams. It will now work very similarly to `source-dynamodb`, and also similarly to SQL capture imprecise backfills, sans doing watermark writes.
- Some horrible migration code to switch existing connectors from a deployment-level change stream to exclusively database-level change streams. I'm probably the most worried about this part, so I tested it extensively. Basically we will read from the deployment-level change stream and use either its post-batch resume token if there are no new events available, or the operation time of an event that is available, to initialize the database-level streams. It is simple to describe but there is a lot of handling for making sure we don't lose any events in the process. Having only database-level streams makes the rest of the connector much simpler, so I think the migration is worth it.
- Some small-ish improvements to CPU and memory use. A note about memory use: The previous version of the code had the possibility for concurrent backfill bindings to trample each others documents and checkpoints. This one doesn't, but it means we have to batch up some documents for a single binding and emit them atomically with a checkpoint. That fundamentally requires more memory to do, and so some optimization was necessary. There are "other" strategies we could explore for this, but this batching approach works for now and is simple to implement and understand.
- There will be a version bump with this change. Mostly to manage risk to existing tasks considering the extent of the changes, but also because it technically isn't backward-compatible with tasks that have a backfill that is in-progress. There's quite a lot of existing MongoDB captures so I'll have to work through upgrading them all over the days++ to come.

Closes #1491
Closes #1492

**Workflow steps:**

The `source-mongodb` documentation should be updated since there are some simplifications here, and I'll work on getting that in order before merging this.

**Notes for reviewers:**

I manually tested these scenarios:
- Migrate from an existing deployment-wide change stream (there's a number of unit tests for this too)
- Migrate from an existing database-specific change stream
- Add a new binding to an existing capture
- Capture a huge database, restarting the connector a bunch of times in the process and make sure it gets everything correctly
- A general load test of capturing a large database with many documents written per second to make sure nothing blows up in the connector

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1466)


source-mongodb/pull.go
```diff
@@ -279,477 +223,47 @@ func updateResourceStates(prevState captureState, bindings []bindingInfo) (captu
...
 	return nil
 }
 
 type backfillState struct {
-	Done bool `json:"done,omitempty"`
-
-	StartedAt time.Time     `json:"started_at,omitempty"`
-	LastId    bson.RawValue `json:"last_id,omitempty"`
+	Done           bool           `json:"done,omitempty"`
+	LastId         *bson.RawValue `json:"last_id,omitempty"`
```
williamhbaker, line 261: Made this a pointer since `omitempty` doesn't really work with non-pointer structs

source-mongodb/migrate.go
```diff
@@ -0,0 +1,333 @@
...
+		}
+
+		if gotEvent {
+			return fmt.Errorf("application error: got an event but resume token didn't change")
+		}
+
+		log.WithFields(log.Fields{
+			"globalToken": globalToken,
+			"newToken":    newToken,
+		}).Info("v2 state migration will re-poll since resume token didn't change")
```
mdibaiee, line 204: Am I right that if there is no event, `newToken.String() != globalToken.String()` is guaranteed to be true, so if we have passed that condition, we definitely have an event, and in that case this log will never be printed since it's after a guaranteed `break`

williamhbaker: Practically speaking I think that's true, but I was able to produce scenarios in test setups where a post-batch resume token was the same as the global token (which will always be associated with an event), which I think has to do with how MongoDB advances its global operation time. What we're doing with this re-poll is waiting for the global operation time to advance enough where we get a post-batch resume token that is either not associated with an event and can be used as-is for database change streams, or a token for a _new_ event where we know the optime of the event.

source-mongodb/pull.go
```diff
@@ -79,120 +89,103 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
...
-}
-
-func (c *capture) outputCheckpoint(state *captureState, merge bool) error {
-	cp, err := json.Marshal(state)
-	if err != nil {
-		return fmt.Errorf("marshalling checkpoint: %w", err)
-	} else if err = c.Output.Checkpoint(cp, merge); err != nil {
-		return fmt.Errorf("outputting checkpoint: %w", err)
+	if didBackfill {
+		log.Info("backfill complete")
```
mdibaiee, line 161: can we move this log into the block above where we set `didBackfill = true`?

source-mongodb/backfill.go
```diff
@@ -0,0 +1,304 @@
...
+	for cursor.Next(ctx) {
+		docCount++
+		totalDocsCaptured++
+
+		var doc bson.M
+		if lastId, err = cursor.Current.LookupErr(idProperty); err != nil {
+			return fmt.Errorf("looking up idProperty: %w", err)
+		} else if err = cursor.Decode(&doc); err != nil {
+			return fmt.Errorf("backfill decoding document: %w", err)
+		}
```
mdibaiee, line 215: Is there a reason to prefer `LookupErr(idProperty)` over `doc[idProperty]`?

source-mongodb/pull.go
williamhbaker: It would log on every backfill "round" then. Having it here makes it log "backfill complete" only when the backfills for the collection are complete done. The extra `didBackfill` setting is so that it doesn't log it every time the capture re-starts after completing its backfill.

source-mongodb/change_stream.go
```diff
@@ -0,0 +1,404 @@
...
+			ts := time.Now()
+			logEntry := log.WithField("database", s.db)
+			logEntry.Info("started catching up stream for database")
+			defer func() {
+				logEntry.WithField("took", time.Since(ts).String()).Info("finished catching up stream for database")
+			}()
+
+			if err := c.tryStream(groupCtx, s, &opTime); err != nil {
+				return fmt.Errorf("catching up stream for %q: %w", s.db, err)
+			} else if err := s.ms.Close(groupCtx); err != nil {
```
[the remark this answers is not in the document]
williamhbaker: Yeah probably not a bad idea. I do want to explicitly do the `Close` here and check for an error since we will cycle through this over and over again while backfilling, but having it as a `defer` also is probably not a bad idea since `Close` is idempotent and the symmetry is nice.

source-mongodb/change_stream.go
```diff
@@ -0,0 +1,404 @@
...
+type changeStream struct {
+	ms *mongo.ChangeStream
+	db string
+}
+
+// makeEventDocument processes a change event and returns a document if the change event is for a
+// tracked collection and operation type. The binding index is also returned if a document is
+// available from the change event. If no document is applicable to the change event, the boolean
+// output will be false.
+func makeEventDocument(ev changeEvent, collectionBindingIndex map[string]bindingInfo) (map[string]any, int, bool) {
```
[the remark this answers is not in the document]
williamhbaker: I went back and fourth a bit on this. I personally like the explicit boolean but it may be more idiomatic to do `nil` instead, so I'll update it to be that.

source-mongodb/pull.go
```diff
@@ -33,6 +49,7 @@ func (d *driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) err
 	}
 
 	var bindings = make([]bindingInfo, len(open.Capture.Bindings))
+	var collectionBindingIndex = make(map[string]bindingInfo)
```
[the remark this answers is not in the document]
williamhbaker: 👍 the prior name was really a holdover from the previous code. I'll rename it to `resourceBindingInfo`

source-mongodb/backfill.go
williamhbaker: It's a really nice way to get a `bson.RawValue` that works for checkpointing. We could get a `primitive.Binary` by `doc[idProperty]` and I'm not sure what we'd have to do with that to make sure it serializes and deserializes from a checkpoint JSON. We used to decode twice; first into a `map[string]bson.RawValue` to get the `bson.RawValue` of the id, and then into a `bson.M`, see [here](https://github.com/estuary/connectors/blob/3c0b7d31744f19af05359ad9326b62f6ddca208e/source-mongodb/pull.go#L684-L688), seemingly getting the `bson.RawValue` for good reason.  Using `Lookup` is much faster.

`Lookup` traverses into nested fields if you provide more than one key. So `Lookup("some", "nested")` would find `"value"` in `{"some": {"nested": "value"}}` as a simplified example. `LookupErr` will return an error if the `idProperty` is not present in the top-level of the document, which is where we need to find it.

I ran this as a sanity check:
```go
func TestLookupErr(t *testing.T) {
	val := bson.D{{"some", bson.D{{"nested", "value"}}}}

	b, err := bson.Marshal(val)
	require.NoError(t, err)

	raw := bson.Raw(b)

	_, err = raw.LookupErr("nested")
	require.Error(t, err)

	got, err := raw.LookupErr("some")
	require.NoError(t, err)
	fmt.Println(got) // {"nested": "value"}

	got, err = raw.LookupErr("some", "nested")
	require.NoError(t, err)
	fmt.Println(got) // {"value"}
}
```

PR #1467, opened 2024-04-10
m-snowflake: recreate pipes on version change only

mdibaiee:
**Description:**

- We have realised that `CREATE OR REPLACE PIPE` is eventually consistent, and so re-creating a pipe on each connector restart was leading to a race where we would send our file ingestion requests to the old pipe (even though we made the request after having run the `CREATE OR REPLACE` query) while that old pipe was going to be deleted and we would then request status updates from the new pipe which did not know about the old pipe. This lead to maximum retries when trying to fetch status of pipe and then a restart for us to attempt to ingest the file again...
- With this change, we only re-create a pipe if the spec version has changed. We also version the pipe name so that when this change is happening, the new pipe and the old pipe do not collide in names, leading to races

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1467)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -129,13 +129,6 @@ type transactor struct {
 	db         *stdsql.DB
 	pipeClient *PipeClient
 
-	// We initialise pipes (with a create or replace query) on the first Store
-	// to ensure that queries running as part of the first recovery Acknowledge
-	// do not create duplicates. Queries running on a pipe are idempotent so long
-	// as the pipe is not removed or replaced. The replace clause of the query
-	// will wipe the history of the pipe and so lead to duplicate documents
-	pipesInit bool
-
 	// map of pipe name to beginMarker cursors
 	pipeMarkers map[string]string
 
@@ -153,8 +146,12 @@ type transactor struct {
 	updateDelay time.Duration
 	cp          checkpoint
 
-	// this shard's range spec, used to key pipes so they don't collide
-	_range *pf.RangeSpec
+	// this shard's range spec and version, used to key pipes so they don't collide
+	_range  *pf.RangeSpec
+	version string
+
+	// list of pipe names that need to be cleaned up
+	pipeCleanups []string
 }
 
 func (t *transactor) AckDelay() time.Duration {
@@ -164,14 +161,6 @@ func (t *transactor) AckDelay() time.Duration {
 func (d *transactor) UnmarshalState(state json.RawMessage) error {
 	if err := json.Unmarshal(state, &d.cp); err != nil {
 		return err
-	}
-	// TODO: remove after migration
-	for _, item := range d.cp {
-		if len(item.StagedDir) == 0 {
-			log.WithField("state", string(state)).Info("found old checkpoint format, ignoring the checkpoint and starting clean")
-			d.cp = make(checkpoint)
-			break
-		}
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


materialize-snowflake/snowflake.go
```diff
@@ -452,14 +429,37 @@ func (d *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
...
 					Query:     mergeIntoQuery,
 					StagedDir: dir,
+					Version:   d.version,
 				}
 			}
 		} else if b.pipeName != "" {
+			var pipeNameParts = strings.Split(b.pipeName, ".")
+
+			// Check to see if a pipe for this version already exists
+			var query = fmt.Sprintf("SHOW PIPES LIKE '%s';", pipeNameParts[len(pipeNameParts)-1])
```
mdibaiee, line 439: We could probably use some caching here to avoid sending this query on every transaction for every binding. Once per version should be sufficient

williamhbaker: Could we re-use a similar mechanism as the prior `d.pipesInit? Basically just setting it to `true` once we have done the check once, and then not running this branch if it is `true`.

materialize-snowflake/snowflake.go
```diff
@@ -452,14 +429,37 @@ func (d *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("finding pipe %q: %w", b.pipeName, err)
+			}
+			var pipeExists = rows.Next()
+
+			// Only create the pipe if it doesn't exist. Since the pipe name is versioned by the spec
+			// it means if the spec has been updated, we will end up creating a new pipe
+			if !pipeExists {
+				log.WithField("name", b.pipeName).Info("store: creating pipe")
+				if createPipe, err := RenderTableShardVersionTemplate(b.target, d._range.KeyBegin, d.version, d.templates.createPipe); err != nil {
```
williamhbaker, line 450: There was some concern about `CREATE PIPE` being eventually consistent as well. Since we will ignore "Pipe not found" errors in Acknowledge now, after getting the success response for creating the pipe should we query `SHOW PIPES` to make sure it does show up as existing? For materializations tailing a collection that don't have an update delay set, the time between creating the pipe and doing the acknowledge may be very short.

williamhbaker (approved):
LGTM % comments

mdibaiee:
@williamhbaker I've updated the pull-request, but I'm also waiting on Snowflake support since they said they are working on potentially making `CREATE OR REPLACE PIPE` strongly consistent. If they do that we can simplify our code.

williamhbaker:
> @williamhbaker I've updated the pull-request, but I'm also waiting on Snowflake support since they said they are working on potentially making `CREATE OR REPLACE PIPE` strongly consistent. If they do that we can simplify our code.

Interesting. SGTM

materialize-snowflake/snowflake.go
mdibaiee: @williamhbaker I'm actually going to back down on this. I would prefer if we always check the existence of the pipes. `SHOW PIPES` does not wake up the warehouse, so I think it doesn't hurt in that sense.

The reason is that if for any reason a pipe has been deleted (let's say the user accidentally dropped it), then we would go through transactions, ignoring the fact that the pipe does not exist.

PR #1493, opened 2024-04-15
Modified OAuth2 and Access Token specs in estuary_cdk & source-hubspot-native fixes

Luishfs:
# Tests

## Local Stack

Tests were made using a flow local stack along with a free-tier version of supabase for materialization.

In order to fix OAuth2 & Access Token issues, both were granted a `secret=True` setting within their sensible data fields.
Also, both `credentials_title` and `credentials` fields were assigned basic types, so that Flow UI would not display warnings. This modified both spec snapshots from `source-hubspot-native` and `source-google-sheets-native`, and they were also updated in this PR. 

The following Streams were validated using the newly built ":local" source-hubspot-native inside Flow UI:
- Companies
- Contact Lists
- Contacts
- Contacts Lists Subscription
- Email Events
- Email Subscriptions
- Marketing Emails
- Properties
- Subscription Changes
- Custom Objects ( Venues is our generated example )

The following Streams were not able to be validated using the newly built connector inside Flow UI, but were validated using flowctl preview:
- Campaigns
- Deal Pipelines
- Deals
- Engagements
- Engagements Calls
- Engagements Emails
- Engagements Meetings
- Engagements Notes
- Engagements Tasks
- Feedback Submissions
- Marketing Forms
- Owners
- Products
- Tickets

The following Streams were not able to be validated (No data was ever returned by them, probably due to the lack of data inside testing accounts):
- Workflows
- Goal Targets
- Line Items
- Ticket Pipelines

## Production Stack

As this connector is already deployed, tests were executed on the production version, also using a free-tier version of supabase for materialization.
Initial tests displayed schema validation errors, showing missing properties on "Contacts" Stream, along with schema errors on Deals and Engagements streams. Missing properties on the "Contacts" Stream seem to be caused by newly added properties. To allow for better maintenance of this connector, PropertiesWithHistory fields were set to be dinamic and allow for extra inputs.

With the new changes, capture snapshots are now more complete and hold similar data to the imported `source-hubspot` version

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1493)

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api.py
+++ b/source-hubspot-native/source_hubspot_native/api.py
@@ -349,7 +349,7 @@ async def fetch_changes(
 
     recent.sort()  # Oldest updates first.
 
-    for batch_it in itertools.batched(recent, 100):
+    for batch_it in itertools.batched(recent, 50):
         batch = list(batch_it)
 
         documents: BatchResult[CRMObject] = await fetch_batch_with_associations(
@@ -449,11 +449,18 @@ async def fetch_recent_deals(
 ) -> tuple[Iterable[tuple[datetime, str]], PageCursor]:
 
     url = f"{HUB}/deals/v1/deal/recent/modified"
-    params = {"count": 100, "offset": page} if page else {"count": 1}
+    params = {"count": 100, "offset": page, "since": int(since.timestamp())} if page else {"count": 100}
 
     result = OldRecentDeals.model_validate_json(
         await http.request(log, url, params=params)
     )
+    if page:
+        if page == 9800:
+            return (
+        (_ms_to_dt(r.properties.hs_lastmodifieddate.timestamp), str(r.dealId))
+        for r in result.results
+    ), None
+
     return (
         (_ms_to_dt(r.properties.hs_lastmodifieddate.timestamp), str(r.dealId))
         for r in result.results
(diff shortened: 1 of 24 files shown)


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -349,7 +349,7 @@ async def fetch_changes(
 
     recent.sort()  # Oldest updates first.
 
-    for batch_it in itertools.batched(recent, 100):
+    for batch_it in itertools.batched(recent, 50):
```
jonwihl, line 352: Whats the reason for the change?

Luishfs: For context: This method receives a tuple consisting of IDs and a datetime field that marks the log_cursor. After that, it checks wether the datetime field is older than the older log_cursor, if so, it saves that ID to be used in a batch API endpoint.

The batch API endpoint size limit for the Deals endpoint seems to be limited by 50 by this error i've received when testing locally using a test account:
Screenshot from 2024-04-16 03-04-00
The only streams using this method are Companies, Contacts, Deals and engagements. Never actually stumbled upon this error before, but since they basically use the same endpoint in the V3 API, i figured instead of checking if the Deals streams is being used and impose the limit, already set the limit for all of the 4 streams.
Since it also runs on the batch endpoint and i never had any issues with speed, i figured cutting the size in half didnt really change much performance wise

jonwihl (approved):
LGTM

PR #1494, opened 2024-04-16
source-mysql: Log raw column type definition for enums

willdonnelly:
**Description:**

It's possible that the enum edge-case we've been seeing (where we see replicated enum values one greater than any valid option) is caused by us parsing the original enum type definition wrong. To figure out if that's the case we need to be able to see what the raw enum definition says before we parse it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1494)


williamhbaker (approved):
LGTM

PR #1499, opened 2024-04-17
Add Source Dropbox

Z33DD:
**Description:**

This connector uses OAuth for Authentication, so we need to create an app on the Dropbox console like in the image bellow:
image

After setting the Estuary's branding, go to the permission tab and make sure that `account_info.read`, `files.metadata.read` and `files.content.read` are selected.

On the overview tab, copy the OAuth credentials and set the Redirect URIs.

Now, apply to production and wait.
image

**Workflow steps:**

Add docs for this connector: https://github.com/estuary/flow/pull/1458

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1499)


source-dropbox/main.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func newDropboxStore(ctx context.Context, cfg config) (*dropboxStore, error) {
+	httpClient, err := GetOAuth2HTTPClient(ctx, dropboxConfig)
+	if err != nil {
+		return &dropboxStore{}, err
+	}
+	config := dropbox.Config{
+		Client:   httpClient,
+		LogLevel: dropbox.LogInfo, // if needed, set the desired logging level. Default is off
```
williamhbaker, line 73: We should probably set this log level to debug, since we typically want user-facing logs at the info level to be carefully crafted so they don't cause confusion.

source-dropbox/main.go
```diff
@@ -0,0 +1,220 @@
...
+
+func (s *dropboxStore) Read(ctx context.Context, obj filesource.ObjectInfo) (io.ReadCloser, filesource.ObjectInfo, error) {
+	meta, reader, err := s.client.Download(&files.DownloadArg{Path: obj.Path})
+	if err != nil {
+		return nil, filesource.ObjectInfo{}, err
+	}
+
+	obj.ModTime = meta.ClientModified
+	obj.ContentType = meta.ExportInfo.ExportAs
+	obj.Size = int64(meta.Size) // Convert uint64 to int64
```
williamhbaker, line 124: The comment here isn't necessary - its very obvious what is happening from the code without the comment

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/dropbox/dropbox-sdk-go-unofficial/v6/dropbox"
+	"golang.org/x/oauth2"
+)
+
+// TODO: Replace the client ID, client secret and redirect URL with the real ones
```
williamhbaker, line 13: These will need to come from the `connectors` table. We won't be able to put these values directly into the code. I'm a little hazy on how this is all threaded together, but it should work similar to `materialize-google-sheets`, or maybe even a better example is `source-google-drive`.

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,53 @@
...
+func GetOAuth2HTTPClient(ctx context.Context, config *oauth2.Config) (*http.Client, error) {
+
+	// token := new(oauth2.Token)
+	// tokenSource := config.TokenSource(ctx, token)
+
+	// use PKCE to protect against CSRF attacks
+	// https://www.ietf.org/archive/id/draft-ietf-oauth-security-topics-22.html#name-countermeasures-6
+	verifier := oauth2.GenerateVerifier()
+
+	// TODO: Handle the code in the UI
```
williamhbaker, line 37: I think most of what is described as being done here is automatically done via the UI already. Again, checkout out `source-google-drive` etc. and see how that works. The connector's responsibility is to produce a valid oauth spec, but it shouldn't need to actually navigate the oauth flow.

source-dropbox/oauth2.go
Z33DD: Okay, I'm just trying to understand how do I access the OAuth2.0 Token from the code so I can use it in the client instantiation.

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, req := range requiredProperties {
+		if req[1] == "" {
+			return fmt.Errorf("missing '%s'", req[0])
+		}
+	}
+	return nil
+}
+
+func (c *Credentials) GetAccessToken(ctx context.Context) (string, error) {
+	return c.AccessToken, nil
```
williamhbaker, line 71: I don't think this is going to work, since the access token will expire in a short time. It would mean that a user would need to come back to the system and re-authenticate when the access token expired. I believe the default is 4 hours so that wouldn't work well at all. We need to be getting a long-lived refresh token, and using that to get a new access token every time the connector starts up.

In the [rclone source](https://github.com/rclone/rclone/blob/5bf70c68f1ed0255be8592abe428b9d50ce37280/backend/dropbox/dropbox.go#L410) the `Client` property of the dropbox config is set to an oauth client. This seems a bit hacky since that field is documented as "for testing purposes only", but it seems like the best way to make this work and since it is implemented that way in rclone I feel decent about it.

Relatedly, I think the `AccessTokenResponseJsonMap` and fields of `Credentials` needs to be re-worked. We should only need `ClientID`, `ClientSecret`, and `RefreshToken`. It should look very similar to the google oauth in that regard. The fact that both systems are using Oauth2 and yet this implementation looks so different should be a strong indicator to you that this one is not right.

Since I was a little hazy on how the oauth stuff worked for connectors, I pushed a branch to prove out what I am saying. Check it out here: https://github.com/estuary/connectors/commit/3c421fe6ff8ac96892c291fc8d0b1482d544b57e

I think we should make updates along these lines to use the refresh token.

source-dropbox/main.go
```diff
@@ -0,0 +1,272 @@
...
+	log.Debug(l.files.Entries)
+
+	if len(l.files.Entries) == 0 {
+		return filesource.ObjectInfo{}, io.EOF
+	}
+	if !l.files.HasMore && l.index > 0 {
+		log.Debug("No more files to list")
+		return filesource.ObjectInfo{}, io.EOF
+	}
+	entry, ok := l.files.Entries[l.index].(*files.FileMetadata)
```
williamhbaker, line 151: We need to handle subdirectories. When I tested this configured with a folder containing a subdirectory it crashed. There's also a "deleted file" metadata type that we may need to handle. We will need to recursively list the contents of the subdirectory I'd imagine. I think we do this in `source-s3` etc. by returning an object with `IsPrefix: true` for directories, see https://github.com/estuary/connectors/blob/2c4ad1bcedde33800518f7125d9ef94f4e498c72/source-s3/main.go#L214-L217

source-dropbox/main.go
```diff
@@ -0,0 +1,272 @@
...
+func (dbx *dropboxStore) check() error {
+	_, err := dbx.client.ListFolder(&files.ListFolderArg{Path: dbx.config.Path})
+	if err != nil {
+		return fmt.Errorf("failed to list files: %w", err)
+	}
+	return nil
+}
+
+func (dbx *dropboxStore) List(ctx context.Context, query filesource.Query) (filesource.Listing, error) {
+	log.Debug("Listing files in path: ", dbx.config.Path)
```
williamhbaker, line 101: Something isn't working with this file listing logic. I tested this on my dropbox account by adding 10,000 files to a directory and it only captured a single file. There's a lot going on with `ListFolderContinue` and other pagination-looking methods so I wonder if something is being missed there. The lexical ordering of file paths also matters.

williamhbaker (changes requested):
See inline comments. Please address them, and test the updated connector against a dropbox account with a non-trivial number of files in a directory structure containing subdirectories.

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,141 @@
...
+
+	pf "github.com/estuary/flow/go/protocols/flow"
+	"golang.org/x/oauth2"
+)
+
+func GetScopes() []string {
+	return []string{
+		"files.metadata.read",
+		"files.content.read",
+		// "account_info.read",  // Not sure if we need this
```
williamhbaker, line 22: Remove this line

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,141 @@
...
+	"mime/multipart"
+	"net/http"
+	"strings"
+	"time"
+
+	pf "github.com/estuary/flow/go/protocols/flow"
+	"golang.org/x/oauth2"
+)
+
+func GetScopes() []string {
```
williamhbaker, line 18: There's no reason to have this be a function, and certainly not an exported one. You can just have it be a global variable: `var scopes = []string{"files.metadata.read", "files.content.read"}`

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,141 @@
...
+		AuthUrlTemplate:           fmt.Sprintf(AUTH_URL_TEMPLATE_FORMAT_STR, strings.Join(scopes, " ")),
+		AccessTokenUrlTemplate:    "https://api.dropboxapi.com/oauth2/token",
+		AccessTokenHeadersJsonMap: map[string]json.RawMessage{"Content-Type": json.RawMessage(`"application/x-www-form-urlencoded"`)},
+		AccessTokenBody:           "client_id={{{ client_id }}}&client_secret={{{ client_secret }}}&code={{{ code }}}&grant_type=authorization_code&redirect_uri={{{ redirect_uri }}}",
+		AccessTokenResponseJsonMap: map[string]json.RawMessage{
+			"access_token":  json.RawMessage(`"/access_token"`),
+			"refresh_token": json.RawMessage(`"/refresh_token"`),
+			"token_type":    json.RawMessage(`"/token_type"`),
+			"expires_in":    json.RawMessage(`"/expires_in"`),
+			"scope":         json.RawMessage(`"/scope"`),
```
williamhbaker, line 45: We do not need nor want to have all of these extra fields as part of the config. See my changes [here](https://github.com/estuary/connectors/commit/3c421fe6ff8ac96892c291fc8d0b1482d544b57e#diff-9206f1d552e9c838a5bdd3e4d39d1c505f55b63545f622cd4161d5bb0ced9189L35-R41) for what will work.

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,141 @@
...
+	ExpireTime   int    `json:"expires_in"`
+	Scope        string `json:"scope"`
+	UID          string `json:"uid"`
+	AccountID    string `json:"account_id"`
+	ClientID     string `json:"client_id"`
+	ClientSecret string `json:"client_secret"`
+}
+
+func (c *Credentials) Validate() error {
+	requiredProperties := [][]string{
```
williamhbaker, line 67: we need ClientID and ClientSecret here too

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	for _, req := range requiredProperties {
+		if req[1] == "" {
+			return fmt.Errorf("missing '%s'", req[0])
+		}
+	}
+	return nil
+}
+
+func (c *Credentials) GetOauthClient(ctx context.Context) (*http.Client, error) {
```
williamhbaker, line 78: This function doesn't need to return an error

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return oauth2.NewClient(ctx, ts), nil
+}
+
+type TokenSource struct {
+	cred *Credentials
+	ctx  context.Context
+}
+
+func (ts *TokenSource) Token() (*oauth2.Token, error) {
```
williamhbaker, line 89: All of this is handled internally by `oauth2.Config.TokenSource` and `oauth2.Token`. We just need to provide it with the refresh token. See https://github.com/estuary/connectors/commit/3c421fe6ff8ac96892c291fc8d0b1482d544b57e#diff-9206f1d552e9c838a5bdd3e4d39d1c505f55b63545f622cd4161d5bb0ced9189R58-R71

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,141 @@
...
+	return []string{
+		"files.metadata.read",
+		"files.content.read",
+		// "account_info.read",  // Not sure if we need this
+	}
+}
+
+// Most of this code was converted from the Google Auth specification updating with
+// the info on https://developers.dropbox.com/oauth-guide
+func OAuth2Spec() *pf.OAuth2 {
```
williamhbaker, line 28: Function doesn't need to be exported

source-dropbox/main.go
```diff
@@ -0,0 +1,312 @@
...
+func (c config) ParserConfig() *parser.Config {
+	return c.Parser
+}
+
+func (c config) PathRegex() string {
+	return c.MatchKeys
+}
+
+func newDropboxStore(ctx context.Context, cfg config) (*dropboxStore, error) {
+	log.Debug("Creating Dropbox client")
```
williamhbaker, line 64: I think we should remove the debug logs throughout this file. They seem like they were probably useful for testing but I don't think they should be included in the final version.

source-dropbox/main.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	return &store, nil
+}
+
+type dropboxStore struct {
+	client files.Client
+	config config
+}
+
+// check checks the connection to the Azure Blob Storage container.
```
williamhbaker, line 91: copy/paste error in this comment

source-dropbox/main.go
```diff
@@ -0,0 +1,312 @@
...
+// It returns an error if the container is not found, the account is disabled, or if there is an authorization failure.
+// If the listing is successful, it returns nil.
+func (dbx *dropboxStore) check() error {
+	resp, err := dbx.client.ListFolder(&files.ListFolderArg{Path: dbx.config.Path})
+	if err != nil {
+		log.Error("Failed to list files")
```
williamhbaker, line 97: This log isn't useful, just return the error

source-dropbox/main.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+		log.Debug("File metadata: ", meta)
+		defer reader.Close()
+		content := make([]byte, meta.Size)
+		if _, err := reader.Read(content); err != nil {
+			log.Error("Failed to download file: ", fileEntry.PathDisplay)
+			return err
+		}
+
+		// log.Debug("File content: ", string(content))
```
williamhbaker, line 118: Remove commented line

source-dropbox/main.go
```diff
@@ -0,0 +1,312 @@
...
+			continue
+		}
+		meta, reader, err := dbx.client.Download(&files.DownloadArg{Path: fileEntry.PathDisplay})
+		if err != nil {
+			log.Error("Failed to download file: ", fileEntry.PathDisplay)
+			return err
+		}
+		log.Debug("File metadata: ", meta)
+		defer reader.Close()
+		content := make([]byte, meta.Size)
```
williamhbaker, line 112: This will OOM the connector if the file it happens to download is very large. We do our best to make sure the connector is configured in a way that will work, but it's not always possible to 100% verify everything. Is being able to get the metadata for a file sufficient to verify that we can read data?

source-dropbox/main.go
```diff
@@ -0,0 +1,312 @@
...
+		break
+	}
+	log.Debug("Successfully connected to Dropbox")
+	return nil
+}
+
+func (dbx *dropboxStore) List(ctx context.Context, query filesource.Query) (filesource.Listing, error) {
+	log.Debug("Listing files in path: ", dbx.config.Path)
+	log.Debug("Query: ", query)
+	files, err := dbx.client.ListFolder(&files.ListFolderArg{Path: query.Prefix, Recursive: query.Recursive})
```
williamhbaker, line 128: The `files.ListFolderResult` returned here has a `hasMore` property to indicate that more paged results are available, and there's a `ListFolderContinue` on the `Client` interface that seems to be for the purposes of getting the next page of results. Do we need to be using these capabilities to make sure we get the full listing? It seems like we probably do.

source-dropbox/main.go
```diff
@@ -0,0 +1,312 @@
...
+
+	if len(l.files.Entries) == 0 {
+		return filesource.ObjectInfo{}, io.EOF
+	}
+	if l.index > len(l.files.Entries)-1 {
+		log.Debug("No more files to list")
+		return filesource.ObjectInfo{}, io.EOF
+	}
+
+	fileEntry, ok := l.files.Entries[l.index].(*files.FileMetadata)
```
williamhbaker, line 178: We should make this check and the one below it into a type switch.

williamhbaker:
In addition to the inline comments, please clean up the commit history prior to submitting this for review again. We'd like the merged set of commits to be useful for future readers of the code, rather than a chronology of all the changes made during the review process.

source-dropbox/main.go
Z33DD: As we are using two scopes, "files.metadata.read" and "files.content.read", reading the metadata just verifies the first one. As you said, we couldn't verify 100% without acctually reading a file, so I will just check the metadata.

source-dropbox/main.go
```diff
@@ -0,0 +1,274 @@
...
+
+	"github.com/dropbox/dropbox-sdk-go-unofficial/v6/dropbox"
+	"github.com/dropbox/dropbox-sdk-go-unofficial/v6/dropbox/files"
+	"github.com/estuary/connectors/filesource"
+	"github.com/estuary/flow/go/parser"
+	pf "github.com/estuary/flow/go/protocols/flow"
+)
+
+type config struct {
+	Path        string         `json:"path" yaml:"path"`
```
Z33DD, line 19: Those YAML tags are necessary for reading test credentials

Z33DD:
Docs PR: https://github.com/estuary/flow/pull/1458

Issue #1495, opened 2024-04-16
MySQL: Fix enum out-of-range decoding bug


willdonnelly:
We have observed a capture failing in production when it receives a just-barely out-of-range enum value.

For context, MySQL enums are stored internally as integers and when we receive them via replication we receive the integer index and have to look up the appropriate string corresponding to that index. Enum cases are numbered from 1 to N, so for instance in `ENUM('Y', 'N')` we have `Y = 1` and `N = 2`. And what we've observed is that it's apparently possible to receive a value one greater than the last enum option, which in this two-option example would be the integer `3`.

Re-backfilling the table doesn't fix this, so it's unlikely to be an unlogged DDL change coming in and screwing up our metadata accuracy. This leaves a handful of options which I still need to investigate:

1. Perhaps there's a way to insert arbitrary integers into an enum column and tell MySQL to just shut up and allow it. This would be dumb, but no dumber than a lot of the other "just shut up and do it" flags that MySQL really does have. However, I have looked a little bit and didn't find any mode settings which seemed like they might do this.
2. Perhaps there's some mechanism for explicitly defining the integer-value mapping which can be used to get larger index values than the number of defined options.
3. Perhaps there's some other subtlety to the enum definition which we're failing to discover correctly.

The production capture in question is running MySQL version 8.0.35 so not exactly an obscure or outdated release. In fact that happens to be the version our CI builds are currently pinned on.

PR #1500, opened 2024-04-17
source-mysql: Fix handling of enums with explicit `''` option

willdonnelly:
**Description:**

It is possible to define an enum in MySQL with an explicit case for the empty string `''`, for instance `enum('', 'Y', 'N')` is a valid enum definition.

When this is done the empty string is both the text representation of the error value zero *and also* a valid enum option with some particular index (in the above option that index is 1 but it is possible to place that anywhere in the options list), but other than that everything continues to operate as normal.

So to fix that our enum parsing logic needs to accept the empty string, which turns out to only require a single-character fix to the parsing regex because apparently I did a dumb thing and wrote it in such a way that it excluded the empty string before.

This fixes https://github.com/estuary/connectors/issues/1495

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1500)

### DIFF
--- a/source-mysql/.snapshots/TestEnumEmptyString-backfill
+++ b/source-mysql/.snapshots/TestEnumEmptyString-backfill
@@ -1,0 +1,16 @@
+# ================================
+# Collection "acmeCo/test/test_enumemptystring_29144777": 8 Documents
+# ================================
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM

Issue #1451, opened 2024-04-04
materialize-sql: tables can exist with no columns, which confuses InfoSchema


williamhbaker:
It is possible for tables to already exist, but to have no columns at all, which currently makes `InfoSchema` think the tables don't exist, since it is only indexing columns. We need to have `InfoSchema` separately enumerate existing tables so that materializations can handle pre-existing tables that have no correctly. Currently a pre-existing table with no columns will cause a `CREATE TABLE IF NOT EXISTS` query, instead of adding all of the columns to the table.

Issue #1256, opened 2024-02-09
materialization-sql: better validation handling for "schemas"


williamhbaker:
Validation currently finds all tables in the relevant schemas per the configuration, but it does not specifically enumerate which schemas exist. This means that if a binding is set with an alternate schema that doesn't exist, the materialization will pass validation since it thinks the table doesn't exist, but fail to apply since the schema doesn't exist. We could also consider automatically creating schemas that don't exist as part of this.

williamhbaker:
Related: https://github.com/estuary/connectors/issues/1451

PR #1503, opened 2024-04-18
materializations: automatically create schemas and handle pre-existing tables with no columns

williamhbaker:
**Description:**

Two main changes in this pull request:
1. Make `InfoSchema` aware of pre-existing resources that have no columns, via a new `PushResource` method. Materialization connectors are responsible for populating resources (tables, in the common case) even if they don't have any columns. This closes https://github.com/estuary/connectors/issues/1451
2. Implement automatic schema creation as a general feature for SQL materializations. They can all do this now, except for `materialize-sqlserver` and `materialize-mysql` which do not support setting schemas. The driver for this was to eliminate runtime errors when a user sets an alternate schema that doesn't exist and we don't check for that in validation, which closes https://github.com/estuary/connectors/issues/1256

There are some other small fixes and adjustments that were made along the way; see individual commit messages.

I tested these changes by doing the following for all the updated materializations:
 - Ran all the unit tests, even the ones that you have to run manually
 - Manually set up a materialization with a pre-existing table that has no columns. It still breaks for BigQuery, see commit/comment for that, but at least we'll have a better error message. Other than that we'll be able to add columns to the existing tables successfully on systems that actually support creating a table with no columns.
 - Manually set up a materialization where both the endpoint config schema and a resource config schema don't exist. Admittedly most cases won't work if the endpoint config schema doesn't already exist, and I stopped short of re-working things enough to make that work in general. But resource config alternate schema creation does work for everything.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1503)

### DIFF
--- a/materialize-databricks/client.go
+++ b/materialize-databricks/client.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	stdsql "database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
@@ -18,6 +19,7 @@ import (
 	boilerplate "github.com/estuary/connectors/materialize-boilerplate"
 	sql "github.com/estuary/connectors/materialize-sql"
 	pf "github.com/estuary/flow/go/protocols/flow"
+	log "github.com/sirupsen/logrus"
 
 	_ "github.com/databricks/databricks-sql-go"
 )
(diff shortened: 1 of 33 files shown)


materialize-databricks/client.go
```diff
@@ -12,98 +13,111 @@ import (
...
+				var colMeta columnMeta
+				if err := json.Unmarshal([]byte(c.TypeJson), &colMeta); err != nil {
+					return nil, fmt.Errorf("unmarshalling column metadata: %w", err)
+				}
+
+				is.PushField(boilerplate.EndpointField{
+					Name:               c.Name,
+					Nullable:           c.Nullable,
+					Type:               string(c.TypeName),
+					CharacterMaxLength: 0, // TODO(whb): Currently not supported by us, although we could parse the metadata for VARCHAR columns.
```
mdibaiee, line 116: Is this a regression from the last version given that we used to have this information? I'm not sure what does this property do / what are the implications of not having it now

materialize-databricks/driver.go
```diff
@@ -188,15 +188,24 @@ func newTransactor(
...
-	// Create volume for storing staged files
-	if _, err := db.ExecContext(ctx, fmt.Sprintf("CREATE VOLUME IF NOT EXISTS `%s`.`%s`;", cfg.SchemaName, volumeName)); err != nil {
-		return nil, fmt.Errorf("Exec(CREATE VOLUME IF NOT EXISTS %s;): %w", volumeName, err)
+	// Create a volume for storing staged files for all schemas in the scope of the materialization.
+	for sch := range schemas {
+		log.WithFields(log.Fields{
+			"schema":     sch,
+			"volumeName": volumeName,
+		}).Debug("creating volume for schema if it doesn't already exist")
+		if _, err := db.ExecContext(ctx, fmt.Sprintf("CREATE VOLUME IF NOT EXISTS `%s`.`%s`;", sch, volumeName)); err != nil {
```
mdibaiee, line 206: Are we using these separate volumes now? I think we would need to tweak the staged files machinery to specify the volume to which we upload files, otherwise by default we upload to the volume that resides in the connection's default schema which is `cfg.SchemaName`.

materialize-databricks/client.go
williamhbaker: As far as I can tell databricks would always say 0 for the "CharacterMaximumLength" even from querying it out of `INFORMATION_SCHEMA.COLUMNS`. A way I could see us getting this information would be to parse it out of the "metadata" JSON returned by this API, which will tell you if a field is for example a `VARCHAR(20)`.

But we only use the `CharacterMaxLength` for `materialize-redshift` and `materialize-mysql` right now, for them to dynamically resize columns, because they don't support an efficient string column of arbitrary length. So since the `CharacterMaxLength` is not used by our Databricks materialization I didn't think it was worth the effort to implement the parsing now.

materialize-databricks/driver.go
williamhbaker: When I tested this, I found that unless volumes were created for the alternate schemas of the bindings, the materialization would fail because the staged file and queries were trying to use a volume specific to that alternate schema already, but it didn't exist.

[Here](https://github.com/estuary/connectors/blob/dd5b4a206360c8ac79783ed5f8f0a90e1cb56075/materialize-databricks/driver.go#L234) is where the staging file path is set, and it uses `target.Path[0]` which is the binding's schema. The staging files are then initialized [a little later](https://github.com/estuary/connectors/blob/dd5b4a206360c8ac79783ed5f8f0a90e1cb56075/materialize-databricks/driver.go#L244-L245), and [use that file path](https://github.com/estuary/connectors/blob/dd5b4a206360c8ac79783ed5f8f0a90e1cb56075/materialize-databricks/staged_file.go#L176) containing the binding's schema to upload files. That path is also used when building [load](https://github.com/estuary/connectors/blob/dd5b4a206360c8ac79783ed5f8f0a90e1cb56075/materialize-databricks/driver.go#L284-L291) and [store](https://github.com/estuary/connectors/blob/dd5b4a206360c8ac79783ed5f8f0a90e1cb56075/materialize-databricks/driver.go#L403-L419) queries.

So from what I found, this volume creation for each binding's schema was necessary for bindings with alternate schemas to work.

PR #1516, opened 2024-04-18
Updating hubspot-native V3 sync types

Luishfs:
Some V3 objects did not have incremental sync correctly implemented and only ran using backfill, this commit updates all other V3 objects to run incremental using the objects stream and backfill using batch

Also, backfill states were being set by default, triggering backfill runs unintentionally, this PR fixes that

**Notes for reviewers:**

`Contacts` stream still have a known issue where batch queries fail to return a formatted json

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1516)


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -122,6 +122,121 @@ async def fetch_page(
...
+        }
+        if len(cls.ASSOCIATED_ENTITIES) == 0:
+            del input['associations']
+
+    recent_log_cursors = []
+    while True:
+
+        _cls: Any = cls  # Silence mypy false-positive.
+        result: PageResult[CRMObject] = PageResult[_cls].model_validate_json(
+            await http.request(log, url, method="GET", params=input)
```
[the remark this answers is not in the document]
Luishfs: Pagination works a few lines after that in
```python
if result.paging:
            input["after"] = result.paging.next.after
        else:
            break

    if len(recent_log_cursors) > 0:
        recent_log_cursors.sort()
        yield recent_log_cursors[-1] 
    else:
        return
```
If `result.paging` is not empty, it means we are still searching, so the `page` parameter is updated with the next paging result.
After we've paginated,the loop break, and the most recent log cursor is yielded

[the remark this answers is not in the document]
williamhbaker: Ah thanks, I missed that

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -122,6 +122,121 @@ async def fetch_page(
         yield result.paging.next.after
 
 
+async def fetch_page_incremental(
+    cls: type[CRMObject],
+    http: HTTPSession,
+    log: Logger,
+    log_cursor: LogCursor,
```
williamhbaker, line 129: Can we use this `log_cursor` as the initial input value for `after`? The idea being that otherwise every time this function is called, it will restart from the "beginning" of the API. But we will already know that we have read up to the time of the `log_cursor` so we don't need to get anything older than that.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -122,6 +122,121 @@ async def fetch_page(
...
+
+    if cls.IGNORE_PROPERTY_SEARCH is True:
+        input = {
+        "associations": ",".join(cls.ASSOCIATED_ENTITIES),
+        "limit": 10,  # 50, # Maximum when requesting history. TODO(johnny).
+        }
+        if len(cls.ASSOCIATED_ENTITIES) == 0:
+            del input['associations']
+    else: 
+        properties = await fetch_properties(log, cls, http)
```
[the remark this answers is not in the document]
Luishfs: I believe it really could be a issue for medium size users, i dont remember actually getting this as an error, but as most of them are already not being searched due to some of issues with properties, im removing this check from the fetch_page methods.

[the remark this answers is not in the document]
Luishfs: Removed in 234055b8b93870329c954b86d9c59f0d6049d7b1

source-hubspot-native/source_hubspot_native/resources.py
```diff
@@ -293,14 +299,13 @@ def open(
         open=open,
         initial_state=ResourceState(
             inc=ResourceState.Incremental(cursor=started_at),
-            backfill=ResourceState.Backfill(next_page=None, cutoff=started_at),
```
williamhbaker, line 296: For this change, if there was no `fetch_page` function, how does the initial `backfill` state matter?

Luishfs: Oh, i didnt, i actually forgot to remove that

source-hubspot-native/source_hubspot_native/api.py
Luishfs: The V3 endpoint only allows for page tokens to be inserted on `after`. The stream would always re-start and bigger users would be affected without a actual "full_refresh".
The stream backoff is really fast, setting it up as default in the ResourceState should take care of real time data for bigger users @williamhbaker ?

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -86,25 +86,13 @@ async def fetch_page(
...
-            del input['associations']
+    input = {
+    "associations": ",".join(cls.ASSOCIATED_ENTITIES),
+    "limit": 10,  # 50, # Maximum when requesting history. TODO(johnny).
+    }
+    if len(cls.ASSOCIATED_ENTITIES) == 0:
+        del input['associations']
+
     if page:
         input["after"] = page
```
williamhbaker, line 97: Is `after` a page counter or a date? I would kind of expect `cutoff` to be used here, without knowing what the API expects.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -122,6 +110,98 @@ async def fetch_page(
...
+    log: Logger,
+    log_cursor: LogCursor,
+) -> AsyncGenerator[CRMObject | LogCursor, None]:
+    # Runs V3 API queries and formats the execution to match
+    # the CDK's fetch_changes requirements 
+    assert isinstance(log_cursor, datetime)
+
+    url = f"{HUB}/crm/v3/objects/{cls.PROPERTY_SEARCH_NAME}"
+
+    input = {
```
williamhbaker, line 125: Can we use the `log_cursor` value here to initialize `input["after"]`, so that it doesn't start over from the beginning each round?

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -122,6 +110,98 @@ async def fetch_page(
...
+    http: HTTPSession,
+    log: Logger,
+    log_cursor: LogCursor,
+) -> AsyncGenerator[CRMObject | LogCursor, None]:
+
+    # Runs V3 API queries and formats the execution to match
+    # the CDK's fetch_changes requirements 
+    url = f"{HUB}{cls.ENFORCE_URL}"
+ 
+    input = {
```
williamhbaker, line 170: Same question here about using `log_cursor` to initialize an `after` value

williamhbaker (commented):
A few more comments, sorry about the scattered review, I think this is it!

source-hubspot-native/source_hubspot_native/api.py
williamhbaker: Ah ok so after is an integer-ish page token?

Luishfs: @williamhbaker yes, its a random string

williamhbaker (approved):
LGTM

My remaining questions are related to my lack of understand of how these API's work, so I'll leave it up to you to consider if they are useful or not.

PR #1520, opened 2024-04-19
Oracle capture

mdibaiee:
**Description:**

- Oracle capture connector. It works by combining backfills using the [psuedo-column `ROWID`](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ROWID-Pseudocolumn.html) and incremental updates using [Flashback](https://docs.oracle.com/en/database/oracle/oracle-database/19/adfns/flashback.html#GUID-603A2020-1F8E-4C2C-BAE6-73454C4CFD7E)
- Introduced tuples of string or integers as a possible `LogCursor` type to allow this connector to use the string `ROWID` as a log cursor for the cutoff point of backfill
- For tests, I have a free tier autonomous Oracle database running which the tests initialise and then write to while running a capture. Credentials for this database are available in config.yaml (sops-encrypted)

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1520)


source-oracle/source_oracle/api.py
```diff
@@ -0,0 +1,265 @@
...
+                doc.meta_ = Document.Meta(op='c', source=source)
+                for (k, v) in row.items():
+                    if k == rowid_column_name:
+                        continue
+                    setattr(doc, k, v)
+
+                yield doc
+
+    if last_rowid is not None:
+        yield last_rowid
```
mdibaiee, line 154: Currently I'm emitting this checkpoint at the end, but I think we can emit this every N rows to make sure we don't lose a lot of progress in case we fail mid-way through a backfill

source-oracle/source_oracle/resources.py
```diff
@@ -0,0 +1,149 @@
...
+            key=[f"/{c.column_name}" for c in t.primary_key],
+            model=t.create_model(),
+            open=functools.partial(open, t),
+            initial_state=ResourceState(
+                backfill=backfill,
+                inc=ResourceState.Incremental(cursor=current_scn),
+            ),
+            initial_config=ResourceConfig(
+                name=t.table_name,
+                interval=timedelta(seconds=0),
```
mdibaiee, line 144: I'm not sure of a good value for this yet, but definitely not zero to avoid overwhelming the server. Perhaps this should be configurable?

source-oracle/source_oracle/api.py
```diff
@@ -0,0 +1,267 @@
...
+      AND ROWID <= '{{ max_rowid }}'
+    ORDER BY ROWID ASC
+""",
+    'inc': """
+SELECT VERSIONS_STARTSCN, VERSIONS_OPERATION, {% for c in table.columns -%}
+{%- if not loop.first %}, {% endif -%}
+{{ c | cast }}
+{%- endfor %} FROM {{ table.table_name }}
+    VERSIONS BETWEEN SCN {{ cursor }} AND MAXVALUE
+    ORDER BY VERSIONS_STARTSCN ASC
```
mdibaiee, line 263: Incremental query

source-oracle/source_oracle/api.py
```diff
@@ -0,0 +1,267 @@
...
+    # as a column identifier, however other casings of the same word
+    # can be used as a column name.
+    'backfill': """
+SELECT ROWID, {% for c in table.columns -%}
+{%- if not loop.first %}, {% endif -%}
+{{ c | cast }}
+{%- endfor %} FROM {{ table.table_name }}
+    WHERE ROWID >{% if is_first_query %}={% endif %} '{{ rowid }}'
+      AND ROWID <= '{{ max_rowid }}'
+    ORDER BY ROWID ASC
```
mdibaiee, line 255: Backfill query

mdibaiee:
CI tests fail because we currently only support TLS auth which requires explicit allow-listing of the IPs. I'm going to add support for wallet authentication which will allow CI to connect using mTLS

source-oracle/source_oracle/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+        resources = await all_resources(log, self, discover.config, pool)
+        return common.discovered(resources)
+
+    async def validate(
+        self,
+        log: Logger,
+        validate: request.Validate[EndpointConfig, ResourceConfig],
+    ) -> response.Validated:
+        pool = create_pool(validate.config)
+        await validate_flashback(log, validate.config, pool)
```
mdibaiee, line 64: Do we always run validate before Open and Discover?

source-oracle/tests/db_seeds/create_test_all_types.sql
```diff
@@ -0,0 +1,21 @@
...
+	nvchar2 NVARCHAR2(4000),
+	vchar2 VARCHAR2(4000),
+	single_nchar NCHAR,
+	vchar VARCHAR(4000),
+	num NUMBER (38, 9),
+	small_int SMALLINT,
+	double_precision DOUBLE PRECISION,
+	float_126 FLOAT(126),
+	real_num REAL,
+	datetime DATE,
```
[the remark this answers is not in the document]
mdibaiee: it actually is date + time-of-day, see https://docs.oracle.com/en/database/oracle/oracle-database/19/nlspg/datetime-data-types-and-time-zone-support.html#GUID-3A1B7AC6-2EDB-4DDC-9C9D-223D4C72AC74

The only difference really between DATE and TIMESTAMP is that DATE does not have timezone and it does not support fractional seconds while TIMESTAMP can do both.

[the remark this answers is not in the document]
willdonnelly: Huh, interesting

source-oracle/source_oracle/__init__.py
willdonnelly: I think that we don't need to include it in discovery, but the main open RPC might still want to run this validation as well. That's how we handle prerequisites in the SQL CDC connectors for instance, running the checks on Validate() and Pull() both.

It's kind of a judgement call, the concern is the scenario where the user has a long-running capture which has been ticking along happily, and then they make some dangerous changes to their database settings so that it no longer passes our validation checks. Is it better to keep things running and potentially not surface the new issue to the user at all, or is it better for the capture to immediately (or at least on the next task restart) begin failing so that the user can potentially be alerted? I have no idea what the right answer here is.

mdibaiee: I think that makes sense to me, will add it to Open as well 👍🏽

source-oracle/source_oracle/models.py
```diff
@@ -0,0 +1,261 @@
...
+        return self.data_type.startswith(self.Type.INTERVAL)
+
+    is_pk: bool = Field(default=False, alias="COL_IS_PK", description="Calculated by join on all_constraints and all_cons_columns")
+
+
+@ dataclass
+class Table:
+    """Table is an extracted representation of a table built from its OracleColumns"""
+
+    table_name: str
```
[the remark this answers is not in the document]
mdibaiee: My understanding is that schemas are basically users... so each user has a schema of their own and that's where we operate, so yeah we don't really discover schemas as such, they don't mean the same thing in Oracle

>    A schema is collection of database objects, including logical structures such as tables, views, sequences, stored procedures, synonyms, indexes, clusters, and database links.
    A user owns a schema.
    A user and a schema have the same name.
    The CREATE USER command creates a user. It also automatically creates a schema for that user.
    The CREATE SCHEMA command does not create a "schema" as it implies, it just allows you to create multiple tables and views and perform multiple grants in your own schema in a single transaction.
    For all intents and purposes you can consider a user to be a schema and a schema to be a user.

source-oracle/source_oracle/models.py
```diff
@@ -0,0 +1,261 @@
...
+
+    pass
+
+
+class OracleTable(BaseModel, extra="forbid"):
+    """
+    Oracle all_tables records. docs: https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/ALL_TABLES.html
+    """
+
+    tablespace_name: str = Field(alias="TABLESPACE_NAME")  # "SYSTEM", or "AcmeCo"
```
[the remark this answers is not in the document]
mdibaiee: No we don't really use this, going to remove it 👍🏽

source-oracle/source_oracle/resources.py
```diff
@@ -0,0 +1,151 @@
...
+            task: Task,
+            _all_bindings,
+        ):
+            common.open_binding(
+                binding,
+                binding_index,
+                state,
+                task,
+                fetch_page=functools.partial(fetch_page, table, pool),
+                fetch_changes=functools.partial(fetch_changes, table, pool),
```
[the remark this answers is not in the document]
mdibaiee: Yep, they run concurrently, and the coordination between them is through the `cutoff` parameter of backfill which is supposed to separate the documents the backfill and incremental changes are captured.

The initial state of backfill has a cutoff set to the last `ROWID` of the table we operate on, and the starting point of the incremental capture is set to the current SCN of the database.

Thinking about this now, I wonder if there is still risk of race between them. I think there is a chance that while we are processing the documents from the backfill query, some document changes, the incremental task outputs the change, but we later process the same document in the backfill task and emit an older version of it 🤔

source-oracle/source_oracle/resources.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+        resources_list.append(common.Resource(
+            name=t.table_name,
+            key=[f"/{c.column_name}" for c in t.primary_key],
+            model=t.create_model(),
+            open=functools.partial(open, t),
+            initial_state=ResourceState(
```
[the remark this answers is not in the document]
mdibaiee: My understanding is that this `initial_state` is used for bindings which have been recently added and have no prior state. So when a binding is first added, these values are used for it, but later when the binding emits its own state, we use that

source-oracle/source_oracle/models.py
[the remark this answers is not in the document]
willdonnelly: Hmm. Is it ownership or access rights? I still don't really understand how this works in practice for tables which aren't created in the user-schema corresponding to our capture user. Like, to create a new user and grant access to a table (the internet tells me this is how it looks in Oracle), I do something like:

```
CREATE USER flow_capture IDENTIFIED BY secret1234;
GRANT CONNECT TO flow_capture;
GRANT SELECT ON foo.products TO flow_capture;
```

Does our discovery logic then see that as a `flow_capture.products` table, or are we discovering the fully-qualified name `foo.products`, or what?

source-oracle/source_oracle/resources.py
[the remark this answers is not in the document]
willdonnelly: Yeah. This is the exact problem that the whole watermarking approach we use in the other SQL CDC connectors is intended to solve. We assume that if we write a watermark, read some rows from a table and then write a second watermark, then by the time we observe the watermark via replication we must necessarily also have observed all changes which resulted in the backfilled state of that portion of the table, plus zero or more changes beyond that. And by emitting the backfilled rows immediately but hold off on emitting a checkpoint update until we observe that next watermark change (only when there's a backfill chunk like this, other replication streaming doesn't do this because that could result in massive transactions) we guarantee that the observable state is both correct and respects transaction boundaries.

There may be some heuristic you can use to simplify things here, but in general if you don't trust there to be any particular causal relationship between the stream of changes and the backfilled row states you need to do something to establish that causality.

source-oracle/source_oracle/api.py
```diff
@@ -0,0 +1,285 @@
...
+    log: Logger,
+    page: str | None,
+    cutoff: Tuple[str],
+) -> AsyncGenerator[Document | str, None]:
+    is_first_query = False
+    if page is None:
+        is_first_query = True
+        async with pool.acquire() as conn:
+            with conn.cursor() as c:
+                await c.execute(f"select min(ROWID) from {table.table_name}")
```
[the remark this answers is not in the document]
mdibaiee: This does require a full-table scan according to the execution plan. I've changed this code to just use the string `AAAAAAAAAAAAAAAAAA` which is the minimum possible base64 encoded string of length 18 (ROWID is of length 18), so this can serve as the minimum value for us instead of us having to scan for it.

source-oracle/source_oracle/models.py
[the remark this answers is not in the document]
mdibaiee: It is ownership, I've updated our discovery logic and now we capture the table from your example as `foo.products`

source-oracle/source_oracle/resources.py
mdibaiee: Unfortunately the way the `common` module is wired up doesn't really allow for that at the moment. I've set a default interval in the config but that's only applicable for the first time a binding is being discovered and afterwards to update the value across the board needs a Find and Replace.

source-oracle/source_oracle/resources.py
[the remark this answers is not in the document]
mdibaiee: @willdonnelly so am I right that we can handle this by having the state of the backfill task available to fetch_changes, and filtering events of fetch_changes which have `ROWID`s that have been processed by the backfill task based on the state. So whenever the backfill task emits N rows, it also updates its last processed ROWID, and the fetch_changes task will only process events which have a ROWID below the backfilled one. If the ROWID is higher, we pause the fetch_changes task (well, we loop in it until the last ROWID has progressed). Once the backfill task has finished, we then don't need to filter anymore in fetch_changes.

Implementing this requires some tweaks in the estuary_cdk 🤔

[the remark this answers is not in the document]
willdonnelly: Hmm, I'm having trouble thinking this through right now because it's always a really complicated subject to get right, but reasoning by analogy the ROWID comparison you propose seems like it would serve the same purpose as the primary-key comparisons we do in the precise backfill mode. And the reason we have that is just to ensure that the first time a document is observed in the capture's output, it shows up as an insertion (whether from replication or a "pretend" insert event from the backfill) and not an update/deletion of the row. The watermarking is a different piece of the puzzle, so I suspect that the ROWID comparison alone wouldn't do the trick anyway.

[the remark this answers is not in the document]
willdonnelly: Ah, I just read your comment a bit more closely and I think there's another more fundamental issue with the approach you describe. The bit where "if the ROWID is higher, we wait until the ROWID is increased by the backfill job then we emit the changes" would effectively mean that a new row inserted at the end of the table would block change streaming until the backfill completed.

And of course "do the whole backfill and only after it completes begin streaming changes from a point in time prior to the backfill start" _is_ a reliable policy which ensures a consistent ordering of all change events, so that suggests that your proposed logic would actually be correct. It's just also not super helpful since most tables will have writes that either append to the end or are uniformly distributed throughout the table, and either of those ends up approximating the lazy "don't even start streaming changes until we're done" policy so closely that we might as well just do that instead and avoid the extra complexity.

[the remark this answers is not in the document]
willdonnelly: Thinking this through step by step:

- The primary concern is about eventual consistency. It's nice to make guarantees about intermediate states where possible, but the really important thing is just that if the user doesn't change a particular row and our capture is allowed to get caught up, the row will eventually have the correct state in the collection.
- If we consider change streaming in isolation without any filtering, it's trivially correct in this sense for any row modified after the capture started. The issue is how to bring in backfilled row states in a way that doesn't compromise correctness.
- If we just blindly run the two in parallel, the concern is that if a row is changed concurrently with our backfill of that portion of the table, it's possible for the backfill query to observe the old value but the change event get emitted before the backfilled version, which means that the row is now in an incorrect state which will persist unless/until it gets modified again in the source DB.
- Blocking change streaming until we've backfilled up to a given row ID is a valid solution to avoid this, but in the worst case it reduces to simply waiting for the backfill to complete before processing any changes, and almost any concurrent-updates scenario will closely approximate that worst-case.
- Filtering out change events with a ROWID greater than the last backfilled one works to avoid the backfill-after-replication concurrency issue, but introduces a new failure mode where we could theoretically filter out change events which ought to have been emitted.
- This whole concurrency issue is what we use watermarks to deal with. Once we see the next watermark we know we have definitely processed any changes which could have occurred concurrently with our backfill query, and so when we emit the backfilled rows followed by all changes since the previous watermark we guarantee the "changes after backfill" ordering invariant. I've just realized that this is in some sense very similar to a "backfill the whole table, then stream all changes since the start of the backfill" policy just carried out on a much smaller scale.
- The thing that I keep coming back to is that in this Oracle capture the backfill and change stream threads aren't completely uncoordinated. They're both executing queries synchronously against the database, and it feels like it might be possible to use that instead of watermarks to get the same sort of "backfill then changes" ordering.
- Hand-waving a bunch here, but suppose that we removed the watermarking behavior from other SQL CDC captures and instead replaced it with a "stream changes for 5 minutes" policy. So long as 5 minutes is enough for us to process all outstanding change events and get caught up to "now" _and_ for the database to not do any weird stale-read concurrency stuff in practice, this would work just as well as actually writing and waiting for a watermark.
- And it doesn't actually have to be 5 minutes, that's just an example of the dumbest possible solution that would work in most cases. More likely what we'd do is some sort of heuristic which could figure out when we're basically caught up on change streaming by querying the system max LSN and streaming to there, plus another second or two at most.
- But something like that would still hinge on some sort of blocking coordination between the backfill and change streaming processes, which would iteratively backfill a chunk and then stream changes until fully caught up to "now" on the source database.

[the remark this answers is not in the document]
willdonnelly: I think there are still only three options here that actually preserve correctness in all possible edge cases:

- The laziest possible option: just run the entire backfill and only then begin streaming all changes since the backfill first began. Since we're requiring 7 days of change retention anyway that's a fairly generous amount of time to finish the backfill.
- Use a watermarks table just like other SQL CDC captures, in order to establish whether we've become caught up to "now" while processing replication events. It's heavyweight and complex, but there's a **_reason_** that's the design we arrived at when we first sat down to figure out how to do accurate SQL CDC without a full snapshot or a backfill-everything-then-stream arrangement.
- Essentially the same logic as watermarking, but using some sort of "get current max SCN of the database and stream to that" instead of actual watermark table writes. This would be breaking new ground and requires the same exact pattern of coordination inside the connector, but would avoid the requirement for a writable watermarks table in the Oracle DB.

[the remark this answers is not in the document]
mdibaiee: Here is a sketch of what might work:
1. We start a backfill and a changes task concurrently
2. Since we have identified the table requires backfilling, the changes task will not proceed since the rowid of any changes captured will be larger than the starting rowid of `AAAAAAA...`
3. So we are guaranteed that the first task to run is a backfill task. Meanwhile runs of the changes task will not proceed, since there is still no checkpoint from backfill task. A backfill task runs and emits 1000 documents and a checkpoint, along with an integer counter which is used to synchronise backfill and changes tasks.
4. At this point the changes task will fetch any changes from the time of step (1) to the last transaction of the database (`BETWEEN {cursor} and MAXVALUE`). While this is happening, the backfill task will not run due to the counter being an even number (we count 0 as even). Once the incremental task is done, we increment the counter by one, at which point it becomes odd and the backfill task is then allowed to run.
5. This ensures a ping-pong between backfill and changes tasks, so we backfill up to a certain point, then fetch all changes since the last time we fetched changes until the last transaction of the database

source-oracle/source_oracle/api.py
```diff
@@ -0,0 +1,332 @@
...
+# thus being missed by the backfill process, however since the incremental changes are captured based on transaction IDs, and not ROWIDs
+# the incremental changes task will capture those updated rows, so we will eventually have all the documents
+
+# In order to avoid races between backfill and incremental tasks, we implement a ping-pong between the two,
+# meaning that they don't really run concurrently ever, but rather they run one after another. We start by running
+# a backfill task (sync_counter = 0), backfilling CHECKPOINT_EVERY documents, and then passing the ball to the
+# incremental task (sync_counter = 1), which will fetch updates since the last seen transaction ID up until the latest transaction
+# of the database. The ball is then passed back to backfill (sync_counter = 2). This ensures
+# that we do not capture incremental updates to documents and emit them while a backfill task may be reading and emitting an older
+# version of the same row afterwards. Incremental updates always come after a backfill operation.
```
mdibaiee, line 133: See these comments for more information on how the backfill and incremental tasks work together to ensure consistency

source-oracle/source_oracle/resources.py
```diff
@@ -0,0 +1,158 @@
...
+        t = build_table(config, ot.owner, ot.table_name, columns)
+
+        max_rowid = None
+        async with pool.acquire() as conn:
+            with conn.cursor() as c:
+                await c.execute(f"SELECT max(ROWID) FROM {t.owner}.{t.table_name}")
+                max_rowid = (await c.fetchone())[0]
+        # if max_rowid is None, that maens there are no rows in the table, so we
+        # skip backfill
+        backfill = ResourceState.Backfill(cutoff=(max_rowid,)) if max_rowid is not None else None
```
willdonnelly, line 117: I'm going to be a bit pedantic here, but is it actually correct to store the maximum RowID in the cutoff field here? According to the class docstrings the cutoff value is supposed to be "[The] LogCursor at which incremental replication began" while RowID values are page cursors.

It feels like either the docstring and type of the cutoff property ought to be modified, or else a new property should be added to `ResourceState.Backfill` to explicitly model the concept of a backfill only proceeding up to a particular page cursor.

source-oracle/source_oracle/api.py
```diff
@@ -0,0 +1,332 @@
...
+                elif current_transaction is not None and scn > current_transaction:
+                    yield current_transaction + 1
+                    current_transaction = scn
+
+    if first_transaction is not None and current_transaction is None:
+        yield first_transaction + 1
+    elif current_transaction is not None and current_transaction > scn:
+        yield current_transaction + 1
+
+    if full_state.backfill is not None:
```
willdonnelly, line 272: Is it actually necessary to check the backfill state here? As the early-return logic is written, if there's no backfill going on then any `sync_counter` value is fine. Alternatively the early-return logic doesn't need to check if the backfill state is `None` if the counter is going to stop incrementing while it's on an acceptable value for change streaming.

willdonnelly (approved):
LGTM

I think that the "ping-ponging" approach you've implemented here avoids any concurrent-update issues, provided that there is no way for a backfill query issued _after_ we've already run a `VERSIONS BETWEEN SCN ... AND MAXVALUE` query to return an older version of a row than the `VERSIONS` query gave us.

Which seems like a reasonable assumption.

(Although strictly speaking I think it's not guaranteed by the Oracle docs or the SQL standard (since transaction isolation guarantees only exist within a single transaction, and for some strange reason even when SQL databases sit down and specify their read semantics they don't tend to give any sort of "we're not going to suddenly run your next query against a snapshot of the table state from five minutes ago" liveness guarantees. But that's just me venting, in practice no database does that because they're all just using writeahead logging and MVCC)

source-oracle/source_oracle/api.py
mdibaiee: @willdonnelly it is possible that we start the connector in a state in which it does not require a backfill at all (think: an empty table with no maximum rowid), in that case the sync counter will still start at 0...  we can probably in that scenario just set the sync counter to 1 and remove this condition, but I find more comfort in knowing that this function will definitely run if there is no backfill for any reason

.github/workflows/python.yaml
```diff
@@ -15,6 +15,7 @@ on:
       - "source-hubspot/**"
       - "source-notion/**"
       - "source-linkedin-pages/**"
+      - "source-oracle/**"
```
jgraettinger, line 18: needs to be updated to source-oracle-flashback

estuary-cdk/estuary_cdk/__init__.py
```diff
@@ -152,6 +152,8 @@ def dump_all_tasks(signum, frame):
             for base in reversed(self.__class__.__bases__):
                 if exit := getattr(base, "_mixin_exit", None):
                     await exit(self, log)
+            if exit := getattr(self.__class__, "exit", None):
```
jgraettinger, line 155: what's this?

estuary-cdk/estuary_cdk/capture/__init__.py
```diff
@@ -10,6 +10,7 @@
 import tempfile
 import traceback
 import xxhash
+import orjson
```
jgraettinger, line 13: duplicate import.

These look to be cosmetic -- remove this file from changeset?

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -456,7 +459,6 @@ async def closure(task: Task):
                 binding_index,
                 fetch_snapshot,
                 state.snapshot,
```
jgraettinger, line 461: why is this removed?

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -587,8 +589,10 @@ async def _binding_backfill_task(
                     f"Implementation error: FetchPageFn yielded PageCursor None. To represent end-of-sequence, yield documents and return without a final PageCursor."
                 )
             else:
-                state.next_page = item
-                task.checkpoint(connector_state)
+                # Avoid unnecessary checkpointing if the page_cursor has not changed
```
jgraettinger, line 592: does this come up somehow for oracle ?

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -13,12 +13,13 @@
...
     Literal,
     TypeVar,
+    Tuple,
 )
-
 from pydantic import AwareDatetime, BaseModel, Field, NonNegativeInt
 
 from ..flow import (
     AccessToken,
+    BasicAuth,
```
jgraettinger, line 22: not used?

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -47,7 +47,8 @@
 CONSISTENCY_HORIZON_DELTA: timedelta = timedelta(minutes=5)
 
 
-properties_cache: dict[str, Properties]= {}
+properties_cache: dict[str, Properties] = {}
```
jgraettinger, line 50: this file seems to be cosmetic and unrelated changes that can be backed out, although the types of `page` and `cutoff` are also changed (and shouldn't be?)

jgraettinger (commented):
Some comments before merging.

I didn't get into source-oracle-flashback broadly as I know @willdonnelly has been doing so. The ssh tunneling bits look reasonable

jgraettinger:
There are some other connectors that are broken by the changed signature in the _binding_snapshot_task call

estuary-cdk/estuary_cdk/__init__.py
mdibaiee: @jgraettinger this allows connectors to specify a "cleanup" function in their class so they can clean up resources before exit

jgraettinger: A better way to do that would be to define a method an `exit` method on BaseConnector with a default implementation that's called.

But I would first check whether the resources to be cleaned up can be factored into a base class of my connector that's a subclass of `Mixin`, and then resource acquisition and cleanup can be done through the existing `_mixin_enter`/ `_mixin_exit` mechanism.

jgraettinger: An example mixin for an HTTP session: https://github.com/estuary/connectors/blob/main/estuary-cdk/estuary_cdk/http.py#L209-L222

jgraettinger:
source-gladly is broken by this PR FYI

jgraettinger (approved):
LGTM

PR #1523, opened 2024-04-22
source-{mysql,postgres}: Default to imprecise backfills for text primary keys

willdonnelly:
**Description:**

As discussed last week, this is the PR which changes the default backfill behavior for MySQL and Postgres when capturing tables with textual primary keys so that they no longer force a full-table sort on every backfill chunk but instead just default to using imprecise backfill ordering semantics.

**Workflow steps:**

A user who just blindly clicks through discovery and creates a capture will now be autoselecting imprecise backfill ordering for any tables whose primary keys are collated text. They can explicitly change the 'Backfill Mode' setting of the resource binding back to `"Precise"` if they want those semantics, and this will work as it did before by applying the `BINARY` or `COLLATE "C"` ordering qualifiers and potentially being slow on massive tables.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1523)


williamhbaker (approved):
LGTM

PR #1524, opened 2024-04-22
materialize-elasticsearch: don't index long strings in flattened mappings

williamhbaker:
**Description:**

We store object types as "flattened" index mappings, and strings with byte-lengths larger than 32766 cause errors when trying insert into those mappings, which is primarily a problem for the `flow_document` field.

There is an option `ignore_above` for this which can be configured on mappings to not index or store these long string fields in Lucene (which is what has the length limit), but still have them available in the `_source` field when retrieving documents. This allows objects with very large strings to be materialized and have the values of those strings retrieved via the `_source` field, which is what we use for querying load documents.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1524)

### DIFF
--- a/materialize-elasticsearch/type_mapping.go
+++ b/materialize-elasticsearch/type_mapping.go
@@ -26,9 +26,10 @@ const (
 )
 
 type property struct {
-	Type   elasticPropertyType `json:"type"`
-	Coerce bool                `json:"coerce,omitempty"`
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #1525, opened 2024-04-22
source-firestore: Sanitize float infinities into strings

willdonnelly:
**Description:**

We already sanitize `NaN` in this way, I'm just copy-pasting the same logic we apply in other connectors for positive and negative infinity as well.

Since previously this would fail JSON serialization we know that the connector has never successfully processed float infinities before.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1525)


williamhbaker (approved):
LGTM

PR #1530, opened 2024-04-24
Adding source-salesforce connector

Luishfs:
**Description:**
Importing the source-salesforce connector with a new modification on how stream discover works.

i'm creating this as a draft PR since the changes proposed in this current version **will affect user experience**

The way the original connector discovery works is by calling the "describe" endpoint in Salesforce API, which describes every callable data endpoint from salesforce and generates streams based on that. The problem is that 900+ streams are generated, and discovery can take longer to occurr.

Also, the original import from this connector disabled most of streams by inserting the `disabled` tag after the discovery process happened. To do this without needing to "inject" this value after discovery, i figured i would probably need to modify how resources are generated within the shim_airbyte_cdk and i didn't spend a lot of time thinking on it since this solution also fixed the long wait for discovery.

It works by adding a new field where users can select which streams they would like to use ( similar to how the meltano version works):
Screenshot from 2024-04-24 09-22-02

If no stream is selected, the connector will generate all streams based on the Describe endpoint mentioned earlier. All streams will be listed in the docs.

Question is: Is this a better alternative to the current version, or should we leave discovery as it is?

Documentation links to be added after "approval" of the Draft pull request

**Notes for reviewers:**

Waiting for approval on the changes made to the connector before opening the pull request

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1530)


Luishfs:
This connector will not be imported. Closing...

PR #1532, opened 2024-04-24
remove state key migrations

williamhbaker:
**Description:**

Generally removes the state key migration code from technical connectors.

It's pretty unlikely that there are any dormant tasks that have been disabled since state keys were implemented and will be re-enabled at some point in the future, but if that happens they will fail on startup with an unmarshalling error. If that happens we can decide what to do - probably create a new capture. I figured it was high time to clean up this code since the chances of having it around still being useful seem very low.

Closes https://github.com/estuary/connectors/issues/1263

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1532)


willdonnelly (approved):
LGTM

PR #1533, opened 2024-04-24
source-google-sheets-native: checking if sheet is empty 

Luishfs:
**Description:**

Added a check before model eval to see if a sheet actually contains any data.
If not, the connector will raise a ` RuntimeError` citing the empty sheet 

**Notes for reviewers:**

Tested with both flowctl preview and end-to-end.
Raise example:
Screenshot from 2024-04-24 15-52-34

### DIFF
--- a/source-google-sheets-native/source_google_sheets_native/api.py
+++ b/source-google-sheets-native/source_google_sheets_native/api.py
@@ -4,6 +4,7 @@ from estuary_cdk.http import HTTPSession
 from logging import Logger
 from typing import Iterable, Any
 from zoneinfo import ZoneInfo
+import json
 
 from .models import (
     NumberType,
@@ -38,8 +39,15 @@ async def fetch_rows(
         "fields": "properties,sheets.properties,sheets.data(rowData.values(effectiveFormat(numberFormat(type)),effectiveValue))",
     }
 
+    # checking if sheet is empty before model eval
+    # since the rowData field wont be returned 
+    sheet_info = json.loads(await http.request(log, url, params=params))
+    if len(sheet_info["sheets"][0]["data"][0]) == 0:
+        raise RuntimeError(f"Spreedsheet sheet '{sheet.properties.title}' is empty!")
+
+
     spreadsheet = Spreadsheet.model_validate_json(
-        await http.request(log, url, params=params)
+        json.dumps(sheet_info)
     )
 
     if len(spreadsheet.sheets) == 0:


source-google-sheets-native/source_google_sheets_native/api.py
```diff
@@ -38,8 +39,16 @@ async def fetch_rows(
         "fields": "properties,sheets.properties,sheets.data(rowData.values(effectiveFormat(numberFormat(type)),effectiveValue))",
     }
 
+    # checking if sheet is empty before model eval
+    # since the rowData field wont be returned 
+    sheet_info = json.loads(await http.request(log, url, params=params))
```
williamhbaker, line 44: This seems like a pretty bad way to do this check. Can `"sheets"` be zero-length? The `if len(spreadsheet.sheets) == 0` check below implies it can. Beside that, deserializing and re-serializing the sheet info looks really hacky.

What actually happens in a spreadsheet contains no rows? Is `sheet.data` `None`? We have an `assert sheet.data` a little below, so maybe we can handle the absence of data more gracefully there?

Luishfs: > Can "sheets" be zero-length? The if len(spreadsheet.sheets) == 0 check below implies it can.

Based on the query we are making, the given `sheet` would need to at least exist, given that the negative return from the below query is a 400:

```python
url = f"{API}/v4/spreadsheets/{spreadsheet_id}"
params = {
    "ranges": [sheet.properties.title],
    "fields":"properties,sheets.properties,sheets.data(rowData.values(effectiveFormat(numberFormat(type)),effectiveValue))",
    }

```
We are already validating the sheet name using properties.title, so i dont know how it could return a sheet that passes that query and does not actually "exists".  It can not be validated by the `spreadsheet.sheets` under a special condition i cant really find, but we would at least receive the API payload to be able to do the data check.

> What actually happens in a spreadsheet contains no rows? Is sheet.data None? 

No. `sheet.data` does not passes model eval.  `sheet.data` Requires a specific field called rowData thats created inside the sheets["data"] field. The issue is, when the sheet has no data at all, API calls return data as:
```json
"data":  [{}]
```
So the `assert sheet.data` would never trigger that, because the code would raise at `Spreadsheet.model_validate_json` by not having the `rowData` inside `data`

> maybe we can handle the absence of data more gracefully there?

I did thought about modifing the pydantic model to allow for data to be empty and then add a check there, but i dont personally see it as "more efficient" than just receiving the payload and checking there, at the end of the day, if there is no data in that sheet, i think we should't spend a second more of processing it.
I dont have any ideas as of now, but if you have any ideas or more feedback i would love to hear it!

Luishfs:
@williamhbaker Updated the fix in 3ec0cb5d82410dd88aa999efa5bed3292bc92be4
Also added a brief explanation of the "Add data > No data > Change data"  test in the PR description

williamhbaker (approved):
LGTM

PR #1538, opened 2024-04-29
source-hubspot: Fixing tests

Luishfs:
# Why this fix was created?
Errors in [source-hubspot](https://github.com/estuary/connectors/actions/runs/8883481172/job/24390442518) build
# Fixes
Updated discovery snapshots, as a new field was added in companies Just a note on this: 
i **think** its not a new field added by hubspot, but a property field that came in a different order when calling the properties API. The code will limit the amount of properties, and given that then can come in different orders, may cause this issue. I saw this behaviour when working with source-hubspot-native , but i still neeed some time to evaluate.

Also added a check on the authenticator type in streams.py. Tests were using TokenAuthenticator and calling the read_records method, causing the code to fail.

**Notes for reviewers:**

All changes were tested with end-to-end tests on a hosted supabase

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1538)

### DIFF
--- a/source-hubspot/source_hubspot/streams.py
+++ b/source-hubspot/source_hubspot/streams.py
@@ -447,8 +447,9 @@ class Stream(HttpStream, ABC):
         # Here we define both the new expiry date and its copy
         # This will be used to compare wether the token expiry date has rotated
         # if the token has a new expiry date, it means we have a new token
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #1539, opened 2024-04-29
source-firestore: Add 'Extra Collections' config option and some other improvements

willdonnelly:
**Description:**

This PR bundles together a bunch of fairly small tweaks to the Firestore connector. A quick rundown:

- Some lint warnings and test logic fixes
- Adding a `/_meta/backfill` boolean so that one can tell the difference between backfilled and streamed documents.
- Adding an 'Extra Collections' advanced config option which allows the user to explicitly specify one or more collections which should be returned from discovery regardless of what our document-sampling approach finds.
- Adding a 'Skip Automatic Discovery' advanced option (intended to be used in conjunction with 'Extra Collections') which just short-circuits past the document sampling entirely.
  - This allows the 'Extra Collections' input to also serve as an explicit whitelist if desired, which is one of those things that _theoretically_ we shouldn't do and should just guide users to use a hand-written spec but in practice it's useful to be able to leave all of the control-plane automation active and just make sure the connector won't do anything unexpected.
- Fixing a bug in the calculation of backfill restart delays which could (in certain edge involving repeated task restarts and stream catchup failures) cases push the backfill further and further into the future.
- Adds some very experimental debug-logging around streaming resume tokens. There's a (somewhat slim) chance that the server could actually be sending us updated resume tokens even before the first consistent point in the stream is reached, and if it is that could theoretically be used to make change streaming more reliable. But it's hard to know for sure because we have no visibility into that in production and it's really hard to reproduce any of this stuff outside of prod.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1539)


source-firestore/discovery.go
```diff
@@ -68,6 +68,7 @@ type documentMetadata struct {
 	CreateTime *time.Time `json:"ctime,omitempty" jsonschema:"title=Create Time,description=The time at which the document was created. Unset if the document is deleted."`
 	UpdateTime *time.Time `json:"mtime" jsonschema:"title=Update Time,description=The time at which the document was most recently updated (or deleted)."`
 	Deleted    bool       `json:"delete,omitempty" jsonschema:"title=Delete Flag,description=True if the document has been deleted, unset otherwise."`
+	Backfill   bool       `json:"backfill,omitempty" jsonschema:"title=Backfill Flag,description=True if this document was captured via a backfill."`
```
williamhbaker, line 71: We use the term `snapshot` elsewhere: https://github.com/estuary/connectors/blob/e516b35943186765f73945b07651b1a5569f32f2/sqlcapture/interface.go#L35

Might be good to use the same terminology here for consistency?

williamhbaker (approved):
LGTM

source-firestore/discovery.go
willdonnelly: Yeah, you're probably right. It's a little ambiguous because the only reason we call it `snapshot` in other connectors is because that's how Debezium names that bit of metadata, and in most of our system and docs we're pretty consistent about naming that concept 'backfill' instead. But we're still using the name in newer connectors such as `source-snowflake` so I guess we ought to here as well.

willdonnelly: Done

PR #1540, opened 2024-04-30
sqlcapture: don't discover materialized tables

williamhbaker:
**Description:**

Don't discover materialized tables, since that is almost never what is intended, and causes problems with synthetic projection names. A column of "flow_published_at" is used as a sentinel for guessing if the table is from a materialization or not. Although not 100% accurate (the user could exclude this field), it will work in the vast majority of cases.

Manually tested this by materializing a table to a Postgres database and trying to capture from that same database. On the current connector, we'd discover the materialized table and have no easy way to exclude it and avoid getting collection spec errors. With this change, we won't discover the materialized table, but will discover all other non-materialized tables like before.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1540)

### DIFF
--- a/sqlcapture/discovery.go
+++ b/sqlcapture/discovery.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

PR #1542, opened 2024-05-01
source-sqlserver: Improved replication diagnostics

willdonnelly:
**Description:**

This PR adds a couple more queries to the automated replication diagnostics for SQL Server.

It also rewords the `"got row"` log message for all three of Postgres/MySQL/MSSQL -- it was usually clear from context that these were diagnostic query results, but when there are lots of complex result rows that's less obvious so hopefully `"got diagnostic row"` will be clearer in such cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1542)


williamhbaker (approved):
LGTM

PR #1543, opened 2024-05-01
source-sqlserver: Another replication diagnostic query

willdonnelly:
**Description:**

Adds another query (`SELECT * FROM msdb.dbo.cdc_jobs;`) to SQL Server replication diagnostics.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1543)


willdonnelly:
Merging unreviewed because I'd like to see what it says about a production task and this is too trivial to bother waiting

PR #1545, opened 2024-05-02
materialize-sql: process deletion events as part of MERGE queries

mdibaiee:
**Description:**

- Add an advanced config to the following connectors to support hard deletions:
    - bigquery
    - databricks
    - snowflake
    - sqlserver

The following connectors still need work, as they were not as straightforward to do:
- redshift (does not support multiple `WHEN MATCHED` predicates in `MERGE INTO` queries)
- mysql, postgres, motherduck: these don't use MERGE INTO queries, so need to delete the rows directly
- mongodb, and others: need specific solutions of their own for deletions, might get to them later

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1545)


materialize-bigquery/bigquery.go
```diff
@@ -31,6 +31,8 @@ type config struct {
 
 type advancedConfig struct {
 	UpdateDelay string `json:"updateDelay,omitempty" jsonschema:"title=Update Delay,description=Potentially reduce compute time by increasing the delay between updates. Defaults to 30 minutes if unset.,enum=0s,enum=15m,enum=30m,enum=1h,enum=2h,enum=4h"`
+
+	HardDelete bool `json:"hardDelete,omitempty" jsonschema:"title=Hard Delete,description=If this option is enabled items deleted in the source will also be deleted from the destination. By default is disabled and _meta/op in the destination will signify whether rows have been deleted (soft-delete).,default=false"`
```
williamhbaker, line 35: I'm wondering if this should be an advanced option, or if it should be a regular option. We've discussed and generally agreed that `UpdateDelay` should _not_ be an advanced option, but just haven't gotten around to re-working it to be in the regular section of the configuration. IMO hard deletes are something that should be in the regular configuration section, since we've seen people have trouble with the update delay config being in the advanced section.

materialize-bigquery/sqlgen.go
```diff
@@ -239,7 +239,7 @@ ON {{ range $ind, $key := $.Keys }}
 	l.{{$key.Identifier}} = r.c{{$ind}}
 {{- end}}
 {{- if $.Document }}
-WHEN MATCHED AND r.c{{ Add (len $.Columns) -1 }} IS NULL THEN
+WHEN MATCHED AND JSON_TYPE(PARSE_JSON(r.c{{ Add (len $.Columns) -1 }}))='null' THEN
```
williamhbaker, line 242: Is there any performance impact when using these extra functions? I remember a time when we tried to allow primary keys to be nullable, and we did a (completely different) kind of `null` checking in load query key equality that was very detrimental for performance.

materialize-redshift/.snapshots/TestSpecification
```diff
@@ -1,165 +0,0 @@
-{
```
williamhbaker, line 1: Is deleting this snapshot file intentional?

materialize-bigquery/sqlgen.go
williamhbaker: It seems like we are serializing a json string with the value of `"null"` to indicate deletions, so is there a more direct way to check for that here? Something more along the lines of `r.cN = 'null'`? Or does that not work for some reason?

materialize-bigquery/transactor.go
```diff
@@ -247,7 +249,11 @@ func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
 		var b = t.bindings[it.Binding]
 		b.storeFile.start()
 
-		converted, err := b.target.ConvertAll(it.Key, it.Values, it.RawJSON)
+		var flowDocument = it.RawJSON
+		if t.cfg.Advanced.HardDelete && it.Delete {
+			flowDocument = json.RawMessage("null")
```
williamhbaker, line 254: I think it's kind of confusing to have the sentinel value for issuing a deletion be the string `"null"`. I assume we're not doing `json.RawMessage(nil)` and checking for the actual `NULL` value here because the `flow_document` column is non-nullable - that seems alright, because it would be a pain to have a special case for that if `HardDelete` is enabled. Could we use a string value of `"delete"` instead for indicating that the row should be deleted?

materialize-bigquery/bigquery.go
```diff
@@ -26,6 +26,8 @@ type config struct {
 	BucketPath       string `json:"bucket_path,omitempty" jsonschema:"title=Bucket Path,description=A prefix that will be used to store objects to Google Cloud Storage's bucket." jsonschema_extras:"order=5"`
 	BillingProjectID string `json:"billing_project_id,omitempty" jsonschema:"title=Billing Project ID,description=Billing Project ID connected to the BigQuery dataset. Defaults to Project ID if not specified." jsonschema_extras:"order=6"`
 
+	HardDelete bool `json:"hardDelete,omitempty" jsonschema:"title=Hard Delete,description=If this option is enabled items deleted in the source will also be deleted from the destination. By default is disabled and _meta/op in the destination will signify whether rows have been deleted (soft-delete).,default=false"`
```
williamhbaker, line 29: It may be good to have an `order` specified for these configurations, since without it I'm not sure where the will show in the ordering.

williamhbaker (approved):
LGTM

PR #1549, opened 2024-05-03
source-hubspot-native: add email events and custom objects

williamhbaker:
**Description:**

Adds bindings for email events and custom objects to `source-hubspot-native`, as well as some other fixes along the way to get the connector working.

The APIs used for both of these bindings have been noted elsewhere to have consistency issues, so the data from them will be delayed by 5 minutes with the hopes that such a delay is long enough to give us consistent results. Of course this is a total guess, based on almost nothing, and we may need to adjust it based on real-world observations.

I've manually tested the connector built from these changes and compared it with the existing ATF based connector to verify that the outputs are equivalent. I've also run it against a large hubspot account and verified that it captures data that looks reasonably correct.

**Workflow steps:**

We should update the hubspot native docs per these changes, and note that the data from the new bindings is delayed by 5 minutes.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1549)

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api.py
+++ b/source-hubspot-native/source_hubspot_native/api.py
@@ -1,8 +1,18 @@
 import asyncio
 import itertools
-from datetime import UTC, datetime
+import time
+from datetime import UTC, datetime, timedelta
 from logging import Logger
-from typing import Any, AsyncGenerator, Awaitable, Callable, Iterable
+from typing import (
+    Any,
+    AsyncGenerator,
+    Awaitable,
+    Callable,
+    Dict,
+    Iterable,
+    Optional,
+    Tuple,
+)
 
 from estuary_cdk.capture.common import (
     LogCursor,
@@ -12,9 +22,12 @@ from estuary_cdk.http import HTTPSession
 from pydantic import TypeAdapter
 
 from .models import (
+    EMAIL_EVENT_HORIZON_DELTA,
     Association,
     BatchResult,
     CRMObject,
+    EmailEvent,
+    EmailEventsResponse,
     OldRecentCompanies,
     OldRecentContacts,
     OldRecentDeals,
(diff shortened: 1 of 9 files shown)


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -280,5 +352,171 @@ async def fetch_recent_tickets(
...
+    cursor: PageCursor,
+) -> tuple[Iterable[tuple[datetime, str]], PageCursor]:
+    
+    url = f"{HUB}/crm/v3/objects/{object_name}/search"
+
+    # The search API has known inconsistencies with very recent data.
+    horizon = datetime.now(tz=UTC) - CONSISTENCY_HORIZON_DELTA
+
+    if horizon <= since:
+        return [], None
```
jgraettinger, line 371: is it somehow possible to reach this condition?

jgraettinger (approved):
LGTM!

Curious for your thoughts on CDK / Python at this point -- this logic would have been annoying in _any_ language, but is there anything that felt missing or could have helped?

supernit: there's a lot of vertical spacing (single SLOC's broken up by empty lines) which IMO makes it harder to read / less code fits on screen at once.

source-hubspot-native/source_hubspot_native/api.py
williamhbaker: When the connector runs for the first time, `since` will be "now", and the horizon will be prior to that.

williamhbaker:
>supernit: there's a lot of vertical spacing (single SLOC's broken up by empty lines) which IMO makes it harder to read / less code fits on screen at once.

👍 I can see that in some areas and have removed some excessive empty lines...I might like empty lines more than the average person, but hopefully this helps.

PR #1550, opened 2024-05-06
source-mongodb: adjust logging

williamhbaker:
**Description:**

Reduces some of the logging noise by adjusting logs to debug level, and makes the frequency of change stream progress updates 5 minutes instead of 1 minute.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1550)

### DIFF
--- a/source-mongodb/backfill.go
+++ b/source-mongodb/backfill.go
@@ -141,7 +141,7 @@ func (c *capture) doBackfill(
 		"collection":         binding.resource.Collection,
 		"estimatedTotalDocs": estimatedTotalDocs,
 	})
-	logEntry.Info("starting backfill for collection")
+	logEntry.Debug("starting backfill for collection")
 
 	// By not specifying a sort parameter, MongoDB uses natural sort to order documents. Natural
 	// sort is approximately insertion order (but not guaranteed). We hint to MongoDB to use the _id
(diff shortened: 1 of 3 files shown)


source-mongodb/backfill.go
```diff
@@ -141,7 +141,7 @@ func (c *capture) doBackfill(
 		"collection":         binding.resource.Collection,
 		"estimatedTotalDocs": estimatedTotalDocs,
 	})
-	logEntry.Info("starting backfill for collection")
+	logEntry.Debug("starting backfill for collection")
```
mdibaiee, line 144: Is this only logged once per task startup? If yes I'd keep it at INFO

source-mongodb/change_stream.go
```diff
@@ -137,7 +137,7 @@ func (c *capture) initializeStreams(
 			return nil, fmt.Errorf("initializing change stream on database %q: %w", db, err)
 		}
 
-		logEntry.Info("intialized change stream")
+		logEntry.Debug("intialized change stream")
```
mdibaiee, line 140: Same comment here

mdibaiee (approved):
thanks! LGTM % personal take

source-mongodb/backfill.go
williamhbaker: These get logged every 5 minute cycle so it's pretty spammy. We get the info log about having progressed the backfill for the collection at the end of every cycle, and an info log when a collection backfill completes, so it was a lot of noise.

source-mongodb/change_stream.go
williamhbaker: In practice I found these logs to be quite redundant with the "catching up streams" info logging. We always initialize a change stream for every database when this happens so its not super useful.

Issue #1273, opened 2024-02-20
materialize-redshift: handle clusters with `enable_case_sensitive_identifier` set to `true`


williamhbaker:
Redshift has an option for enabling case sensitivity for identifiers: https://docs.aws.amazon.com/redshift/latest/dg/r_enable_case_sensitive_identifier.html

Currently we assume this option is `false` and that all identifiers will be lower-cased, but if a destination system has a global setting for `true` we will encounter problems since various assumptions in the connector no longer hold.

A few possible options:
- Produce an error on validate and say that the cluster or user needs to be set with `enable_case_sensitive_identifier = FALSE`, as these settings can be configured on a per-user basis: https://docs.aws.amazon.com/redshift/latest/dg/cm_chap_ConfigurationRef.html
- Explicitly set `enable_case_sensitive_identifier = FALSE` for every connection/session we use the connector
- Query for the value of `enable_case_sensitive_identifier` and use that to configure the behavior of the connector dynamically

PR #1551, opened 2024-05-06
materialize-redshift: support clusters with `enable_case_sensitive_identifier` set to ON

williamhbaker:
**Description:**

The `enable_case_sensitive_identifier` configuration makes it so Redshift does not always lowercase all identifiers, and causes it to preserve the casing of quoted identifiers.

Previously we would always assume that this parameter was OFF and that we should treat identifiers as lowercase, but this breaks materializations where the cluster is configured with the parameter to ON.

The changes here allow us to support clusters where `enable_case_sensitive_identifier` is set to ON.

Tested by running all the unit and e2e tests, and manually ran a materialization with mixed capitalization fields with `enable_case_sensitive_identifier` both on and off to make sure it produced the expected output.

Closes https://github.com/estuary/connectors/issues/1273

**Workflow steps:**

The [Naming conventions](https://docs.estuary.dev/reference/Connectors/materialization-connectors/amazon-redshift/#naming-conventions) section in our documentation should be updated to reflect this change.

**Notes for reviewers:**

I've verified that all existing Redshift materializations either have `enable_case_sensitive_identifier` set to OFF, or are broken because we can't handle clusters where it's ON. For existing systems with `enable_case_sensitive_identifier` set to OFF, there should be no change in behavior of the connector. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1551)

### DIFF
--- a/materialize-redshift/driver.go
+++ b/materialize-redshift/driver.go
@@ -243,18 +243,22 @@ func newRedshiftDriver() *sql.Driver {
 			}
 
 			var caseSensitiveIdentifierEnabled = strings.EqualFold(caseSensitiveIdentifier, "on")
+			var dialect = rsDialect(caseSensitiveIdentifierEnabled)
+			var templates = renderTemplates(dialect)
 
-			log.WithField("caseSensitiveIdentifierEnabled", caseSensitiveIdentifierEnabled).Info("detected value for enable_case_sensitive_identifier")
+			if caseSensitiveIdentifierEnabled {
+				log.WithField("caseSensitiveIdentifierEnabled", caseSensitiveIdentifierEnabled).Info("detected value for enable_case_sensitive_identifier")
+			}
 
 			return &sql.Endpoint{
 				Config:              cfg,
-				Dialect:             rsDialect,
+				Dialect:             dialect,
 				MetaSpecs:           &metaSpecs,
 				MetaCheckpoints:     &metaCheckpoints,
 				NewClient:           newClient,
-				CreateTableTemplate: tplCreateTargetTable,
+				CreateTableTemplate: templates.createTargetTable,
 				NewResource:         newTableConfig,
-				NewTransactor:       newTransactor,
+				NewTransactor:       prepareNewTransactor(templates, caseSensitiveIdentifierEnabled),
 				Tenant:              tenant,
 				ConcurrentApply:     true,
 			}, nil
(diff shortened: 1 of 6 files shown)


materialize-redshift/driver.go
```diff
@@ -359,23 +385,22 @@ func (t *transactor) addBinding(
...
+		{&b.copyIntoMergeTableSQL, fmt.Sprintf("flow_temp_table_%d", bindingIdx), target.Columns(), b.storeFile},
+		{&b.copyIntoTargetTableSQL, target.Identifier, target.Columns(), b.storeFile},
 	} {
 		var sql strings.Builder
-		if err := tplCopyFromS3.Execute(&sql, copyFromS3Params{
-			Target:          m.target,
-			Columns:         m.columns,
-			ManifestURL:     m.stagedFile.fileURI(manifestFile),
-			Config:          t.cfg,
-			TruncateColumns: m.truncateColumns,
```
williamhbaker, line 378: We don't need this `TRUNCATECOLUMNS` option anymore since the runtime automatically keeps strings to less than 64k in length redshift materializations.

materialize-redshift/driver.go
```diff
@@ -662,14 +687,6 @@ func (d *transactor) commit(ctx context.Context, fenceUpdate string, hasUpdates
 	}
 	defer conn.Close(ctx)
 
-	// Truncate strings within SUPER types by setting this option, since these have the same limits
```
williamhbaker, line 665: See note above about not needing this truncation anymore, since it is done by the runtime automatically now.

materialize-redshift/sqlgen.go
```diff
@@ -2,23 +2,17 @@ package main
...
 
 	sql "github.com/estuary/connectors/materialize-sql"
 	"github.com/estuary/flow/go/protocols/fdb/tuple"
 )
 
-// Identifiers matching the this pattern do not need to be quoted. See
-// https://docs.aws.amazon.com/redshift/latest/dg/r_names.html. Identifiers (table names and column
-// names) are case-insensitive, even if they are quoted. They are always converted to lowercase by
-// default.
-var simpleIdentifierRegexp = regexp.MustCompile(`(?i)^[a-z_][a-z0-9_]*$`)
```
williamhbaker, line 19: Previously we wouldn't quote column names with uppercase characters, which never really made any difference since we always assumed that column names would be treated as lowercase. Now with the potential for handling case sensitivity, we can use the standard `sql.IsSimpleIdentifier` method which will quote identifiers with uppercase characters.

mdibaiee (approved):

PR #1562, opened 2024-05-07
materialize-databricks: custom DDL specification in field selection

mdibaiee:
**Description:**

- We support capturing uint64 values, which do not fit in a `BIGINT`. So we are allowing users to switch to `DECIMAL(20)` by specifying a custom DDL for their columns using field selection config

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1562)


materialize-databricks/sqlgen.go
```diff
@@ -50,12 +50,34 @@ var databricksDialect = func() sql.Dialect {
...
+		}
+	}
+
 	// https://docs.databricks.com/en/sql/language-manual/sql-ref-datatypes.html
 	var mapper sql.TypeMapper = sql.ProjectionTypeMapper{
 		sql.ARRAY:    jsonMapper,
 		sql.BINARY:   sql.NewStaticMapper("BINARY"),
 		sql.BOOLEAN:  sql.NewStaticMapper("BOOLEAN"),
-		sql.INTEGER:  sql.NewStaticMapper("BIGINT"),
+		sql.INTEGER:  customDDLMapper(sql.NewStaticMapper("BIGINT")),
```
mdibaiee, line 80: At the moment I have only added this mapper to integers, just for the sake of "this will be hard to take back, so let's not give it away prematurely"

materialize-databricks/sqlgen.go
```diff
@@ -198,23 +220,9 @@ SELECT -1, ""
...
-{{- else if Contains $ "BOOLEAN" -}}
-	::BOOLEAN
-{{- else if Contains $ "BINARY" -}}
-	::BINARY
-{{- else if Contains $ "STRING" -}}
-	::STRING
-{{- end -}}
-{{ end }}
+{{ define "cast" -}}
+::{{- First (Split $ " ") -}}
```
mdibaiee, line 224: This may not work with every possible custom DDL (e.g. `INTERVAL` or `MAP` or `ARRAY` might fail to work with this cast), but I didn't want to invest more time in this for now

materialize-databricks/sqlgen.go
```diff
@@ -50,12 +50,34 @@ var databricksDialect = func() sql.Dialect {
...
+	var customDDLMapper = func(mapper sql.TypeMapper) sql.FieldConfigMapper {
+		return sql.FieldConfigMapper{
+			Delegate: mapper,
+			Map: func(m *sql.MappedType, config json.RawMessage) error {
+				type customDDLFieldConfig struct {
+					DDL string `json:"DDL,omitempty"`
+				}
+
+				var fieldConfig customDDLFieldConfig
+				if err := json.Unmarshal(config, &fieldConfig); err != nil {
```
jgraettinger, line 62: nit: use `pf.UnmarshalStrict` here?

materialize-databricks/sqlgen.go
jgraettinger: Perhaps just a TODO comment then?

materialize-sql/type_mapping.go
```diff
@@ -193,6 +193,33 @@ func WithElementConverter(converter ElementConverter) StaticMapperOption {
 	}
 }
 
+// FieldConfigWrapper wraps a TypeMapper to specify the type of the column in
```
jgraettinger, line 196: This comment seems to be copy / pasted from somewhere?

materialize-sql/type_mapping.go
```diff
@@ -193,6 +193,33 @@ func WithElementConverter(converter ElementConverter) StaticMapperOption {
...
+
+var _ TypeMapper = FieldConfigMapper{}
+
+func (m FieldConfigMapper) MapType(p *Projection) (mapped MappedType, err error) {
+	delegated, err := m.Delegate.MapType(p)
+	if err != nil {
+		return MappedType{}, err
+	}
+
+	if p.RawFieldConfig != nil {
```
jgraettinger, line 214: nit: an improvement would be to have `FieldConfigMapper` have an argument function which returns a new instance of the field config to parse into.

Then the deserialization and error check logic can live here, and the `Map` function can receive a pre-validated instance of it's custom field config type.

jgraettinger (approved):
LGTM % comments

Comments are possible improvements but this seems sound to me as-is.

jgraettinger:
Looks like your update to the common materialization fixture -- which is definitely a Good Idea -- is surfacing some other materialization connectors that have this issue.

materialize-sql/type_mapping.go
mdibaiee: Updated it to use a generic type which allows it to handle the parsing of the config and pass it to the map function (without a need to later do type assertions on an `interface`

mdibaiee:
@jgraettinger separated the test for unsigned-bigint into a collection of its own and enabled it for databricks, snowflake and materialize-mysql for now, the first two work as expected, while mysql clamps the value to a maximum that it supports.

tests/materialize/materialize-databricks/snapshot.json
```diff
@@ -800,6 +809,17 @@
   ],
   "table": "`main`.`some-schema`.formatted_strings"
 }
+{
+  "rows": [
+    {
+      "flow_document": "{\"_meta\":{\"uuid\":\"fb0cc8b0-20e1-11ee-990b-ffd12dfcd47f\"},\"id\":1,\"unsigned_bigint\":18446744073709551615}",
+      "flow_published_at": "2023-07-12T18:29:04.671352Z",
+      "id": 1,
+      "unsigned_bigint": 18446744073709552000
```
mdibaiee, line 818: This inaccuracy seems to be due to `jq` not supporting unsigned uint64s properly... I've verified the value comes correctly from our `fetch-data`, but gets clamped by `jq`

materialize-sql/type_mapping.go
jgraettinger: This is even better, I keep forgetting that Go generics are now a thing :+1:

PR #1563, opened 2024-05-07
materialize-databricks: materialize base64 encoded strings as BINARY columns

williamhbaker:
**Description:**

Adds the capability to materialize base64 encoded strings as BINARY columns to materialize-databricks.

This change is not fully backward compatible, since existing materializations with base64 encoding annotated string fields will fail validations with their previously created STRING columns. Also the new casting from base64 to bytes will result in the raw bytes being inserted into the STRING column, which technically doesn't break the materialization, but is a departure from previous behavior.

As such, this change has a version bump of the connector, and tasks can be migrated over to it by re-backfilling any bindings with base64 strings in their source collections.

Manual testing done, in addition to running all of the unit and integration tests:
- Materialize a collection with base64-encoded strings using the existing v1 version
- See that the v1 version of the task creates string columns for these
- Try to update the task to the v2 image built from these changes and see that it fails validation
- Re-backfill the tables as part of updating to v2 and see that it works
- For good measure, add some more data to the collection so that it does more inserts and merges and see that works too

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1563)


materialize-databricks/.snapshots/TestSQLGeneration
```diff
@@ -21,29 +21,29 @@ SELECT 0, `a-schema`.target_table.flow_document
 	USING (
 		(
 			SELECT
-			key1, key2, boolean, integer, `unsigned-integer`, number, string, flow_document
+			key1::BIGINT, key2::BOOLEAN, unbase64(binary)::BINARY as binary, boolean::BOOLEAN, integer::BIGINT, `unsigned-integer`::DECIMAL(20), number::DOUBLE, string::STRING, flow_document::STRING
```
williamhbaker, line 24: I think this is one of the more interesting changes: Doing the casting in the `SELECT` and then not doing it in the rest of the merge query seems to work, and it simplifies the weird stuff I had to do for the casting template for handling binary columns, which need their base64 base unbase64'd into bytes to store correctly in binary columns. Just calling this out since conceptually I think it works, and it seems to work in tests, but I know we have learned lessons about casting that I might be missing with this.

materialize-databricks/driver.go
```diff
@@ -213,10 +213,6 @@ type binding struct {
 	// into the target table. Note that in case of delta updates, "needsMerge"
 	// will always be false
 	needsMerge bool
-
-	mergeInto string
```
williamhbaker, line 217: This variables weren't used anywhere, so this is some opportunistic cleanup.

mdibaiee (approved):
LGTM, thanks!

PR #1564, opened 2024-05-07
source-mysql: Correctly handle unsigned integer values

willdonnelly:
**Description:**

Previously there was a bug where values of an unsigned integer column which were received via replication would incorrectly be cast to the bit-equivalent signed integer value. This applied to all widths of integer -- a `TINYINT UNSIGNED` column could demonstrate this with value `222` being incorrectly captured as `-34`.

This applied only to replication, backfilled values were and are still handled correctly.

The root of the issue is that (unlike the value encoding used for backfill query results) the MySQL binlog value representation doesn't distinguish based on the signedness of integer values, only their bit-width. Consequently the client library parsed all such values as signed integers of the appropriate width and we serialized them unmodified.

The fix was to keep track of an additional signed/unsigned flag as part of the column type metadata for integer columns. When the column is unsigned we cast the value to an unsigned integer and where necessary (primarily for the 24-bit `MEDIUMINT` type but we'll handle any weirdness that might theoretically come up by going `intN -> int64 -> uint64 -> mask`) trim off any sign-extended upper bits.

**Workflow steps:**

No user action is required, except in cases where we believe that unsigned integer values might previously have been captured incorrectly. In those cases we should re-backfill the impacted bindings, ideally discarding old data and materializing into a fresh destination table.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1564)

### DIFF
--- a/source-mysql/.snapshots/TestAlterTable_AddUnsignedColumn-discover1
+++ b/source-mysql/.snapshots/TestAlterTable_AddUnsignedColumn-discover1
@@ -1,0 +1,130 @@
+Binding 0:
+{
+    "recommended_name": "test_altertable_addunsignedcolumn_57413089",
+    "resource_config_json": {
+      "namespace": "test",
+      "stream": "AlterTable_AddUnsignedColumn_57413089"
+    },
+    "document_schema_json": {
+      "$defs": {
+        "TestAlterTable_AddUnsignedColumn_57413089": {
+          "type": "object",
+          "required": [
+            "id"
+          ],
+          "$anchor": "TestAlterTable_AddUnsignedColumn_57413089",
+          "properties": {
+            "data": {
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "id": {
+              "type": "integer"
+            }
+          }
+        }
+      },
+      "allOf": [
+        {
+          "if": {
+            "properties": {
+              "_meta": {
+                "properties": {
+                  "op": {
+                    "const": "d"
+                  }
+                }
+              }
+            }
+          },
+          "then": {
+            "reduce": {
+              "delete": true,
+              "strategy": "merge"
+            }
+          },
+          "else": {
+            "reduce": {
+              "strategy": "merge"
+            }
+          },
+          "required": [
+            "_meta"
+          ],
+          "properties": {
+            "_meta": {
+              "type": "object",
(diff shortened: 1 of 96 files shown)


source-mysql/discovery.go
```diff
@@ -471,10 +475,50 @@ func (t *mysqlColumnType) translateRecordField(val interface{}) (interface{}, er
 			return string(bs), nil
 		}
 		return val, nil
+	case "tinyint":
+		if sval, ok := asInt64(val); ok && t.Unsigned {
+			return uint64(sval) & 0xFF, nil
+		}
```
mdibaiee, line 481: This seems to work, but I don't grok why we need to cast to int64 and then to uint64, instead of directly from int8 to uint8?

mdibaiee (approved):
LGTM, one comment for me to understand how this works, and probably a good idea to have that answer as a comment

source-mysql/discovery.go
willdonnelly: Yeah, on reflection I think the `asInt64` helper is an unnecessary bit of complexity. I'll replace it with type assertions for the appropriate integer sizes in each case (int8 for `tinyint`, int16 for `smallint`, etc) and only apply a mask to `mediumint` values.

willdonnelly:
> Also, can you speak to how migration of the state checkpoints work, from "foo": "int" => "foo": {"type": "int"} ?
> 
> Has this dynamic "maybe it's a string or object" always been supported, and you're just using it here?

Yes, we previously implemented support for struct-valued type information in MySQL so that we could store the list of enum/set options, and we're just reusing that same capability here to store metadata about integer signedness.

source-mysql/.snapshots/TestAlterTable_AddUnsignedColumn-discover1
```diff
@@ -0,0 +1,130 @@
...
+          "$anchor": "TestAlterTable_AddUnsignedColumn_57413089",
+          "properties": {
+            "data": {
+              "type": [
+                "string",
+                "null"
+              ]
+            },
+            "id": {
+              "type": "integer"
```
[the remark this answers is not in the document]
willdonnelly: Done

mdibaiee (approved):

PR #1565, opened 2024-05-08
sqlcapture: Fix backfill termination on slow, high-volume tables

willdonnelly:
**Description:**

Historically we have detected when a backfill reaches the end of a table when we receive zero new rows from a backfill query. This usually works fine, but can misbehave when a table is both heavily loaded with new rows constantly inserted at the end *and also* the individual backfill queries are relatively slow.

When this happens, the backfill can keep on issuing queries for up to <chunkSize> rows, and receive a few dozen each time, which keeps the backfill from terminating. And normally that's also benign, so we haven't bothered to fix it up until now.

However there are still a few edge cases where backfill queries can get fairly expensive. The one we're seeing in production right now is queries against a very large keyless table in MySQL, but there are a few other scenarios like that where we would very much like to ensure the backfill ends as soon as it can.

Really any number of result rows from a backfill query less than the chunk size means we've just read the final chunk that we need, so it's conceptually simple enough to switch the condition from `count == 0` to `count < chunkSize`. The only complexity is that there's no easy way for the top-level sqlcapture logic to see what chunk size the capture was configured with. So in order to do it that way I had to add a bit of plumbing so that the `ScanTableChunk` function can determine for itself when the backfill is complete.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1565)

### DIFF
--- a/source-mysql/backfill.go
+++ b/source-mysql/backfill.go
@@ -26,7 +26,7 @@ func (db *mysqlDatabase) WatermarksTable() string {
 	return db.config.Advanced.WatermarksTable
 }
 
-func (db *mysqlDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.DiscoveryInfo, state *sqlcapture.TableState, callback func(event *sqlcapture.ChangeEvent) error) error {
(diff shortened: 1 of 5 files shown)


mdibaiee (approved):
LGTM

The source-sqlserver CI task is failing but it seems unrelated to your change

PR #1567, opened 2024-05-08
source-mysql: Stop ignoring query events with leading comments

willdonnelly:
**Description:**

This PR stops the MySQL replication DDL parsing logic from ignoring queries which begin with a comment.

There were reasons this was expedient in the past, but it's causing us pain at the moment and the Vitess SQL parser ought to be able to cope with the comments themselves -- IIRC the issue in the past was specific queries which began with a comment and used some unsupported statement type or another.

But right now this is the number one issue with our DDL tracking logic, and unlike the second biggest issue (DDL queries which don't get logged at all) it's unarguably our fault we're not handling it well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1567)


williamhbaker (approved):
LGTM

PR #1569, opened 2024-05-08
source-mysql: Don't specify `minimum: 0` for now

willdonnelly:
**Description:**

This broke several things in production and we should get those fixed with correct values of unsigned integer columns before we start enforcing that in the schema.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1569)


willdonnelly:
Merging without review because enforcing `minimum: 0` has caused some important stuff to fail in production and I'd like to get it unblocked ASAP. We can then go and fix the underlying data at our leisure and eventually re-enable this.

PR #1574, opened 2024-05-10
snowpipe: don't use pagination parameter at all

mdibaiee:
**Description:**

- After some back and forth with Snowflake support team, they said if we don't provide the `beginMarker` at all, we will get _all_ pipe statuses for the past 15 minutes (my impression was that we only get the first page if we don't provide `beginMarker`), which makes it much easier for us to work with the API, since the paginated results required us to keep track of status of every pipe as we paginate. This should resolve a lot of the issues we have been having with pipe statuses not showing up (those were mostly due to us paginating to a new page, and still expecting to see status of pipes that were not LOADED, but they only show up in later pages if there is an update to them).
- This also means we are much less likely to hit a case where the report has zero files, which would trigger us to ask the COPY_HISTORY table for data, so saving costs as well.

Tested locally by running integration tests, but also ran tests on production with logs to make sure reports do show all statuses for long-running queries.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1574)


williamhbaker (approved):
LGTM

PR #1580, opened 2024-05-15
source-hubspot: Fixing hubspot snapshot

Luishfs:
**Description:**

Simply updating the source-hubspot snapshot

**Notes for reviewers:**

This error still seem to be related to the source limitation, will write a fix propose in the comments

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1580)

### DIFF
--- a/source-hubspot/tests/snapshots/snapshots__discover__capture.stdout.json
+++ b/source-hubspot/tests/snapshots/snapshots__discover__capture.stdout.json
@@ -14856,6 +14856,12 @@
                 "string"
               ]
             },
+            "hs_latest_message_visible_to_visitor": {
+              "type": [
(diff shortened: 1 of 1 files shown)


Luishfs:
@williamhbaker Are you in favor of limiting the amount of response data from the properties API to 3 or 5 so we can bypass this constant snapshot error?

Luishfs:
@williamhbaker For reference, i think this error happens because the atf connector hard-limits the amount of properties, so probably it leaves or adds new properties to the "main" if a new property comes in

Luishfs:
This is an old PR to fix a small snapshot issue, the changes made here do not reflect the current state of the connector. Closing...

PR #1581, opened 2024-05-16
sqlcapture: handle u/int8 values in fdb tuple packing

mdibaiee:
**Description:**

- Added the ability to test datatypes as primary keys to our test harness and used it to test the int8 case
- Prior to this pull-request we would get this error if a `tinyint` was used as [part of] a primary key
```
error encoding row key for \"test.datatypes_5_tinyint_10000005\": error serializing FDB tuple: unencodable element at index 0 (124, type int8)
```

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1581)

### DIFF
--- a/source-mysql/datatype_test.go
+++ b/source-mysql/datatype_test.go
@@ -31,10 +31,10 @@ func TestDatatypes(t *testing.T) {
 		{ColumnType: "text", ExpectType: `{"type":["string","null"]}`, InputValue: "hello", ExpectValue: `"hello"`},
 
 		// Integer Types
-		{ColumnType: "tinyint", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, PrimaryKeyValue: 124, PrimaryExpectValue: `124`},
-		{ColumnType: "smallint", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, PrimaryKeyValue: 124, PrimaryExpectValue: `124`},
-		{ColumnType: "mediumint", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, PrimaryKeyValue: 124, PrimaryExpectValue: `124`},
-		{ColumnType: "int", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, PrimaryKeyValue: 124, PrimaryExpectValue: `124`},
+		{ColumnType: "tinyint", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, AsPrimaryKey: true},
+		{ColumnType: "smallint", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, AsPrimaryKey: true},
+		{ColumnType: "mediumint", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, AsPrimaryKey: true},
+		{ColumnType: "int", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, AsPrimaryKey: true},
 		{ColumnType: "bigint", ExpectType: `{"type":["integer","null"]}`, InputValue: -1234567890123456789, ExpectValue: `-1234567890123456789`},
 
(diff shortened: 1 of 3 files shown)


source-mysql/datatype_test.go
```diff
@@ -31,10 +31,10 @@ func TestDatatypes(t *testing.T) {
 		{ColumnType: "text", ExpectType: `{"type":["string","null"]}`, InputValue: "hello", ExpectValue: `"hello"`},
 
 		// Integer Types
-		{ColumnType: "tinyint", ExpectType: `{"type":["integer","null"]}`, InputValue: 123, ExpectValue: `123`},
-		{ColumnType: "smallint", ExpectType: `{"type":["integer","null"]}`, InputValue: 123, ExpectValue: `123`},
-		{ColumnType: "mediumint", ExpectType: `{"type":["integer","null"]}`, InputValue: 123, ExpectValue: `123`},
-		{ColumnType: "int", ExpectType: `{"type":["integer","null"]}`, InputValue: 123, ExpectValue: `123`},
+		{ColumnType: "tinyint", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, PrimaryKeyValue: 124, PrimaryExpectValue: `124`},
```
willdonnelly, line 34: Shouldn't the `InputValue` and `ExpectValue` settings be removed from these test cases, if the intention is to exercise the capture behavior when it's used as a primary key?

sqlcapture/tests/datatypes.go
```diff
@@ -19,6 +19,9 @@ type DatatypeTestCase struct {
 	ExpectType  string
 	InputValue  interface{}
 	ExpectValue string
+
+	PrimaryKeyValue    interface{}
```
willdonnelly, line 23: If it's an either/or choice between primary-key and just-a-value cases, why do we need separate `PrimaryKeyValue` and `PrimaryExpectValue` fields? Couldn't there just be a `PrimaryKey bool` which toggles the behavior while using the same test case fields for input and expected output?

willdonnelly (commented):
The datatype change and adding it to `TestScanKeyTypes` looks good, but I'm not entirely clear what purpose adding the whole `PrimaryKeyValue` bit of infrastructure to `TestDatatypes` serves, and also the implementation of that change is very confusing to me.

sqlcapture/tests/datatypes.go
mdibaiee: It's not an either-or, it's both. When you provide `PrimaryKeyValue`, we test the value both as a normal column and as a primary key. I wasn't sure if there is something to be tested about normal columns that wouldn't apply to primary keys, so I went with the cautious path of doing both.

If doing a primary key check alone is sufficient, I can refactor this to just be a boolean flag.

willdonnelly: Perhaps I just need more coffee and I'm missing something obvious, but from my reading of the code it really doesn't look like it tests both options. Just as a single example, only one table is created which either has the column of interest as a key or a value. If it were running both options I would expect to see either that or an extra `t.Run()` call somewhere to run separate "as primary key" and "as ordinary value" subtests.

Regardless, I don't think that we need to test both at the same time in the first place. In general there is no difference between how we handle values as primary keys and non primary-keys, and if there were such a difference we could still test it with just a boolean flag by having a pair of test cases like:

```
{ColumnType: "int", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`},
{ColumnType: "int", ExpectType: `{"type":"integer"}`, InputValue: 123, ExpectValue: `123`, PrimaryKey: true},
```

mdibaiee: You are right actually, now I remember why I did it this way: we are inserting to the same table from the backfill and the replication parts of the test, but we couldn't insert the same value for a primary key, so I introduced a second value to be unique... However I just refactored this to use a compound key when testing as a primary key, that way we it works with the same value for both backfill and replication and simplifies things a lot

willdonnelly: Ah, that makes sense. With that change this all looks good to me.

willdonnelly (approved):
LGTM

PR #1586, opened 2024-05-17
estuary-cdk: use fixed restart interval of 24 hours and make backfills respect it too

williamhbaker:
**Description:**

Currently when any incremental task stops yielding documents after the resource-specific interval
has elapsed, the task will exit and remain idle until all other tasks have completed and the
connector exits. This is a problem when there are long-running backfills which make take longer to
complete than the incremental streams can allow because of log cursor expiration, etc.

To address this, backfill tasks are being made aware of the "stopping" signal, and they will return
when this signal is set, even if the backfill isn't 100% complete. This will cause the capture to do
its periodic restart after all incremental tasks have caught up, and not require all backfills to be
complete as well.

To prevent excessive restarting that may negatively impact throughput during backfills a fixed
interval for these restarts of 24 hours is being set.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1586)

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/__init__.py
+++ b/estuary-cdk/estuary_cdk/capture/__init__.py
@@ -255,14 +255,14 @@ class BaseCaptureConnector(
 
             stopping = Task.Stopping(asyncio.Event())
 
-            async def stop_on_elapsed_interval(interval: int) -> None:
-                await asyncio.sleep(interval)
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

PR #1600, opened 2024-05-23
Add support for optionally bypassing Pydantic

jshearer:
I found that Pydantic serialization was causing some significant slowdowns, especially with many small documents. This PR exposes the ability for connectors to yield plain old dicts instead of Pydantic documents, and serialize them quickly with orjson.

It also appears that no `poetry.lock` file was committed to the estuary CDK folder yet, so I added it here. Is there a reason it wasn't included? 

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1600)


jgraettinger (approved):
LGTM

I don't believe `poetry.lock` belongs, because this is a library and not a complete application -- instead, the connector that _uses_ the CDK has a `poetry.lock`.

PR #1605, opened 2024-05-28
source-firestore: Add 'Restart Cursor Path' resource config property

willdonnelly:
**Description:**

This PR implements a new resource config property 'Restart Cursor Path' which allows the user to specify a JSON pointer to some known monotonically-increasing property of their dataset which has an explicit index configured on it. When set, the connector will do the following things:

- It will keep track of some recent observed value of that property, persisted in the state checkpoint.
- It will order backfills of that resource according to the specified path. Note that this requires the user to manually configure an index in the Firestore console.
- In the event of the change stream breaking, the "re-establish consistency" backfill will begin from the checkpointed value rather than the start of the dataset.
- In the event of the change stream breaking, the task will restart much sooner than if no such path were set.

The upshot of all this is we're basically admitting that Firestore change streaming for high-volume collections is fundamentally unreliable and if your change rate is sufficiently high it's a question of _when_ and not _if_ it will eventually fail. And currently our recovery strategy when this fails is to go and re-backfill the entire collection, which is the best we can do in a very generic sense but if the user's dataset has _any property at all_ which increases monotonically (such as a serial integer ID or a document-level `updated_at` timestamp) then we can use that to dramatically reduce the cost of such re-backfills, and then because they're so much cheaper we can do them much faster as well.

We should really update https://docs.estuary.dev/reference/Connectors/capture-connectors/google-firestore/ to more accurately describe how the connector operates these days.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1605)


source-firestore/pull.go
```diff
@@ -118,6 +132,15 @@ type resourceState struct {
 	// to be performed to re-establish consistency.
 	Inconsistent bool `json:"Inconsistent,omitempty"`
 
+	// The `RestartCursorValue` property is updated periodically during change streaming
+	// (if the binding has a `RestartCursorPath` setting) so it holds an arbitrarily selected
+	// restart-cursor value which was recently observed in the dataset.
+	RestartCursorValue any `json:"RestartCursorValue,omitempty"`
+
+	restartCursorPath           []string  // The parsed path to the restart-cursor property. Not serialized as part of the state, it's just copied from the resource spec for convenience.
+	pendingRestartCursorTimeout time.Time // The timestamp after which the pending restart-cursor value may be promoted to active. Zero if there is no pending value.
```
williamhbaker, line 141: I'm likely missing something here, but why do we need to wait until some amount of time has passed before promoting a restart cursor?

source-firestore/main.go
```diff
@@ -21,11 +22,14 @@ const (
...
+	BackfillMode      backfillMode `json:"backfillMode" jsonschema:"title=Backfill Mode,description=Configures the handling of data already in the collection. Refer to go.estuary.dev/source-firestore for details or just stick with 'async'. Has no effect if changed after a binding is added.,enum=async,enum=none,enum=sync"`
+	InitTimestamp     string       `json:"initTimestamp,omitempty" jsonschema:"title=Initial Replication Timestamp,description=Optionally overrides the initial replication timestamp (which is either Zero or Now depending on the backfill mode). Has no effect if changed after a binding is added."`
+	RestartCursorPath string       `json:"restartCursorPath,omitempty" jsonschema:"title=Restart Cursor Path,description=Optionally specifies a JSON pointer to some document property which increases monotonically and can be used as a restart cursor to optimize backfill behavior when streaming consistency is lost. Generally this only matters for collections with very high write volumes.,pattern=^(/([^/~]|~[01])+)*$"`
 }
 
+var jsonPointerRegexp = regexp.MustCompile(`^(/([^/~]|~[01])+)*$`)
```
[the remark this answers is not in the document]
willdonnelly: No reason for it to be different, it just didn't occur to me to check there for a regexp I could copy so I just wrote my own here. I'll see about swapping this out for that one in the name of consistency.

source-firestore/pull.go
willdonnelly: The point of the whole "restart cursor" thing isn't precision so much as it is cutting the bulk of the unnecessarily redundant backfill work we would otherwise have to do when the stream breaks. And I can hardly overstate how little I trust Firestore to behave in a remotely sane manner, so we don't want to assume that the changes we observe via streaming will occur in a reliable order.

The failure mode which could potentially occur without the whole "pending value" thing would look like this:

- After a task restart we begin streaming changes, and fail to catch up to that initial consistent point so the stream breaks, and then we have to go re-backfill it.
- But it's entirely possible that Firestore could have sent us those changes in some arbitrary order which doesn't match the order the changes were made in.
  - This would be valid from their standpoint because the way the change streaming protocol is designed we're clearly intended to buffer the entire "initial snapshot" set of changes and get to a consistent point before doing anything with them
  - That's what the official client library does for example, and avoiding that is why we have to drop down and use the raw gRPC protocol directly.
- So if we simply saved the most recent value we observed and then began backfilling from that value after the stream broke, we could potentially skip over some number of changes with slightly smaller values which would have been sent to us later on if the stream hadn't broken.

That would be pretty bad since the entire point of re-backfilling the collection is to make sure it's consistent with the source dataset, so we're deliberately biasing things so that we accept a small amount of overlap between the documents observed by change streaming and the documents we'll fetch when re-backfilling, in order to minimize the chance that we miss anything.

williamhbaker: Makese sense; thanks!

williamhbaker (approved):

PR #1607, opened 2024-05-28
source-{mysql,postgres,sqlserver}: Sort secondary indices by name

willdonnelly:
**Description:**

This commit alters the secondary index selection heuristic (which is used during discovery, if a table has no primary key, to try and provide a useful alternative to use as the collection/backfill key).

Previously we picked the first suitable index, where "first" was defined in terms of iterating over a map from index name to the list of columns making up the index. When stated like that the flaw ought to be obvious -- Go map iteration order is random, so if there were multiple suitable indices we could pick a different one every time discovery gets run.

The fix is to make a list of suitable indexes, sort the list by name, and then pick the first one from that list. This is still not perfect (as it could still change if a new index is created on the source database) but the selection will at least no longer keep changing every time we run discovery.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1607)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"regexp"
 	"slices"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


psFried (approved):
LGTM

PR #1609, opened 2024-05-28
source-google-sheets-native: ignoring null rows after empty spreadshe…

Luishfs:
…et validation

**Description:**

One bug report showed that there can be cases where a row simply has a Null value inside its 'values' parameters. This doesnt seem to be covered in the docs and i dont fully understand what happened to that row, but since we are now validating if a spreadsheet is empty or not, adding a simple check to the row generator in `fetch_rows` solves the issue
and allow for the streams to run

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1609)


jonwihl (approved):
LGTM

PR #1613, opened 2024-05-29
materialize-mysql: support hard deletions

mdibaiee:
**Description:**

- Support hard deletions for materialize-mysql

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1613)


materialize-mysql/driver.go
```diff
@@ -734,21 +769,27 @@ func (d *transactor) Store(it *m.StoreIterator) (_ m.StartCommitFunc, err error)
 			defer txn.Rollback()
 
 			for _, b := range d.bindings {
-				if !b.mustMerge {
-					// Keys that don't already exist are inserted directly into the target table.
-					continue
-				}
+				if b.mustDelete {
```
williamhbaker, line 772: Does this work if there are deletions and merge updates in the same transaction for the same binding?

mdibaiee: Good point! I initially made the same mistake on postgres as well and fixed it later...
So I just updated the deletions fixture to include a transaction during which we delete, update and insert at the same time to make sure this doesn't happen

williamhbaker (approved):
Seems that the test fixture updates might require updates to all the other snapshots, but LGTM other than that

PR #1618, opened 2024-05-31
new parquet and csv materializations

williamhbaker:
**Description:**

This PR adds new streaming encoders for writing rows of materialized values to Parquet and CSV files, and creates materializations for writing such files to GCS and S3.

As part of this, a `filesink` framework was created for (hopefully) easily creating new "file-based" materializations using other formats and writing data other systems. I haven't done any JSON materializations with this since I'm not sure if there's any demand for them (you can just use your normal Flow storage bucket), and I also haven't done any Azure blob storage materialization yet - the latter is something we probably will want to do at some point soon.

See individual commit messages for further descriptions.

Closes [#1626](https://github.com/estuary/connectors/issues/1626)

**Workflow steps:**

New/updated documentation will be needed for all 4 of these materializations.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1618)


materialize-boilerplate/stream-encode/parquet_schema.go
```diff
@@ -0,0 +1,241 @@
...
+		out = append(out, projToSchemaElement(*p))
+	}
+
+	return out
+}
+
+func projToSchemaElement(p pf.Projection) ParquetSchemaElement {
+	out := ParquetSchemaElement{
+		Name:     p.Field,
+		Required: p.Inference.Exists == pf.Inference_MUST,
```
williamhbaker, line 173: Self-review comment: This isn't exactly right; it needs to also consider the possibility for the existence of a `null` value.

filesink/filesink.go
```diff
@@ -0,0 +1,370 @@
...
+		constraint.Reason = "The root document should usually be materialized"
+	case len(p.Inference.Types) == 0 || slices.Equal(p.Inference.Types, []string{"null"}):
+		constraint.Type = pm.Response_Validated_Constraint_FIELD_FORBIDDEN
+		constraint.Reason = "Cannot materialize a field with no types"
+	case p.Field == "_meta/op":
+		constraint.Type = pm.Response_Validated_Constraint_LOCATION_RECOMMENDED
+		constraint.Reason = "The operation type should usually be materialized"
+	case strings.HasPrefix(p.Field, "_meta/"):
+		constraint.Type = pm.Response_Validated_Constraint_FIELD_OPTIONAL
+		constraint.Reason = "Metadata fields fields are able to be materialized"
```
mdibaiee, line 358: Do we have an established convention for not including _meta fields by default? I feel like we usually do include them as recommended (e.g. uuid and flow_published_at, etc.)

materialize-boilerplate/stream-encode/parquet.go
```diff
@@ -0,0 +1,701 @@
...
+	case Gzip:
+		propOpts = append(propOpts, parquet.WithCompression(compress.Codecs.Gzip))
+	default:
+		panic(fmt.Sprintf("unknown compression setting: %d", cfg.compression))
+	}
+
+	return file.WithWriterProps(parquet.NewWriterProperties(propOpts...))
+}
+
+// Encode encodes a row of data by buffering it in the encoder'ss buffer, and if thresholds are
```
mdibaiee, line 214:
```suggestion
// Encode encodes a row of data by buffering it in the encoder's buffer, and if thresholds are
```

materialize-boilerplate/stream-encode/parquet.go
```diff
@@ -0,0 +1,701 @@
...
+
+	if e.scratch.sizeBytes >= e.cfg.rowGroupByteLimit ||
+		e.scratch.rowCount >= e.cfg.rowGroupRowLimit ||
+		e.scratch.columnChunkCount >= maxScratchColumnChunkCount {
+		if e.scratch.columnChunkCount >= maxScratchColumnChunkCount {
+			log.WithFields(log.Fields{
+				"scratchSizeBytes":        e.scratch.sizeBytes,
+				"scratchRowCount":         e.scratch.rowCount,
+				"scratchColumnChunkCount": e.scratch.columnChunkCount,
+			}).Debug("flushing scratch file based on column chunk count")
```
mdibaiee, line 247: for my learning: why do we specificaly log this one?

materialize-boilerplate/stream-encode/parquet.go
```diff
@@ -0,0 +1,701 @@
...
+
+// columnBatchWriter is a generic wrapper for writing a batch of values having type T to a column.
+// This interface is satisfied by any of the typed column writers from the Apache parquet package.
+type columnBatchWriter[T parquetValue] interface {
+	// WriteBatch writes a batch of repetition levels, definition levels, and values to the column.
+	// We don't currently support nested data structures (typed arrays being the only ones we
+	// reasonably could), so repLvls is always nil. The number of values in defLvls must equal the
+	// number of conceptual rows that are being written. The vals slice may contain a number of
+	// values less than or equal to the number of rows, as null values are omitted. Since we don't
+	// currently support nested data structures, a repLvl of 0 means the row at that corresponding
```
mdibaiee, line 412: did you mean to say "a defLevel of 0 means the row at that corresponding..." instead of "repLevel"? I don't see repLevel being used here in the code

materialize-gcs-csv/.snapshots/TestSpec
```diff
@@ -0,0 +1,98 @@
...
+        "order": 1,
+        "secret": true
+      },
+      "uploadInterval": {
+        "type": "string",
+        "enum": [
+          "5m",
+          "15m",
+          "30m",
+          "1h"
```
mdibaiee, line 26: So we don't allow a value of 0 here?

materialize-motherduck/.snapshots/TestValidateAndApply
```diff
@@ -113,7 +113,7 @@ Big Schema Changed Types Constraints:
 {"Field":"stringUriField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
 {"Field":"stringUriReferenceField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
 {"Field":"stringUriTemplateField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
-{"Field":"stringUuidField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
+{"Field":"stringUuidField","Type":6,"TypeString":"UNSATISFIABLE","Reason":"Field 'stringUuidField' is already being materialized as endpoint type 'UUID' but endpoint type 'VARCHAR' is required by its schema '{ type: [string] }'"}
```
mdibaiee, line 116: Is this expected?

mdibaiee (approved):
LGTM, got some minor comments

materialize-boilerplate/stream-encode/parquet_schema.go
williamhbaker: Nevermind, I actually accounted for this a little lower in the `if t == "null"` check where it sets `Required` to `false`.

materialize-boilerplate/stream-encode/parquet.go
williamhbaker: It would mean that we'd be producing row groups that were smaller than the configured row group byte limit or row count limit. So it could be a troubleshooting mechanism in case somebody wonders why their row groups are too small.

materialize-boilerplate/stream-encode/parquet.go
williamhbaker: Yep! Thanks.

materialize-gcs-csv/.snapshots/TestSpec
williamhbaker: For the moment at least I'm not exposing that through the UI dropdown menu since having a huge amount of tiny files is probably not what users are going to want. But it would actually be possible to set it manually via the CLI if somebody has a reason to do so.

materialize-motherduck/.snapshots/TestValidateAndApply
williamhbaker: Yeah, the particular snapshot output of this test is kind of weird, as it is changing various types and trying to re-validate against an existing table. In this case it was a `type: string, format: uuid` field originally that created a `UUID` column, and the test is verifying that when the `format: uuid` is removed (so it's just `type: string`) validation fails.

PR #1621, opened 2024-06-03
CDK Updates

jshearer:
**Description:**

* Specs with many incremental tasks were struggling to start backfilling on startup as all of the incremental tasks were eagerly starting. This way, if an incremental task has run less than `interval` ago, it'll delay until that interval has fully elapsed before starting.
* Pydantic automatically serializes Decimals as strings, but orjson doesn't know about that. In order to handle this, we must tell `orjson` how to handle Decimals. [Their docs on this](https://github.com/ijl/orjson?tab=readme-ov-file#default)
* Pass the current Task to all of the `FetchXXFns` so that they can directly write to the buffer if need be. **This is a breaking change**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1621)

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -614,11 +614,29 @@ async def _binding_incremental_task(
     connector_state = ConnectorState(
         bindingStateV1={binding.stateKey: ResourceState(inc=state)}
     )
+
+    sleep_for = timedelta()
+
     task.log.info(f"resuming incremental replication", state)
 
+    if isinstance(state.cursor, datetime):
+        lag = datetime.now(tz=UTC) - state.cursor
+
+        if lag < binding.resourceConfig.interval:
+            sleep_for = binding.resourceConfig.interval - lag
+            task.log.info("incremental task ran recently, sleeping until `interval` has fully elapsed", {"sleep_for": sleep_for, "interval": binding.resourceConfig.interval})
+
     while True:
-        # Yield to the event loop to prevent starvation.
-        await asyncio.sleep(0)
+        try:
+            if not task.stopping.event.is_set():
+                await asyncio.wait_for(
+                    task.stopping.event.wait(), timeout=sleep_for.total_seconds()
+                )
+
+            task.log.debug(f"incremental replication is idle and is yielding to stop")
+            return
+        except asyncio.TimeoutError:
+            pass  # `sleep_for` elapsed.
 
         checkpoints = 0
         pending = False
@@ -657,8 +675,6 @@ async def _binding_incremental_task(
                 "Implementation error: FetchChangesFn yielded a documents without a final LogCursor",
             )
 
-        sleep_for: timedelta = binding.resourceConfig.interval
-
         if not checkpoints:
             # We're idle. Sleep for the full back-off interval.
             sleep_for = binding.resourceConfig.interval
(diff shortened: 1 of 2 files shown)


estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -182,7 +182,7 @@ class AssociatedDocument(Generic[_BaseDocument]):
     doc: _BaseDocument
     binding: int
 
-FetchSnapshotFn = Callable[[Logger], AsyncGenerator[_BaseDocument | dict, None]]
+FetchSnapshotFn = Callable[[Logger, Task], AsyncGenerator[_BaseDocument | dict, None]]
```
jgraettinger, line 185: You don't need to add Task here (and, if you're doing so, this PR needs to fix all other connectors which were broken).

Instead, just close over Task in the code path that's calling open_binding()

jshearer: That's what I thought too, and I was confused when it didn't seem to be emitting any documents: it's the wrong Task. The Task [passed to `resource.open`](https://github.com/estuary/connectors/blob/3888fc81a225bf8d38d8780cdd06713f511dffd0/estuary-cdk/estuary_cdk/capture/common.py#L399-L405) is actually the _parent_ of the Task that actually runs the generators. It's actually `open_binding` itself that [creates](https://github.com/estuary/connectors/blob/main/estuary-cdk/estuary_cdk/capture/common.py#L438) [the](https://github.com/estuary/connectors/blob/main/estuary-cdk/estuary_cdk/capture/common.py#L448) [final](https://github.com/estuary/connectors/blob/main/estuary-cdk/estuary_cdk/capture/common.py#L463) Tasks that contain the buffers we want to write to here (and it has to, as we need separate a separate Task/buffer per generator type -- backfill, incremental, snapshot)

jgraettinger: :man_facepalming: d'oh, you're totally right.

Okay. Trying to figure out how to not make this more confusing for the common case.

One option would be to replace the Logger parameter with Task, since the Logger is trivially accessed as `task.log` (and this PR does then need to update all call sites). That's ... okay, but if I'm writing a connector, I'm wondering what this Task is and what I'm supposed to do with it ("do I call `captured` myself?").

Or, could the netsuite connector just re-implement open_binding itself? If you did that and called _binding_backfill_task / _binding_incremental_task directly it could work?

jshearer: >Or, could the netsuite connector just re-implement open_binding itself?

Yep, good point. Removed this commit in favor of re-implementing `open_binding` in the connector

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -611,16 +611,28 @@ async def _binding_incremental_task(
     connector_state = ConnectorState(
         bindingStateV1={binding.stateKey: ResourceState(inc=state)}
     )
-    task.log.info(f"resuming incremental replication", state)
 
     while True:
+        if isinstance(state.cursor, datetime):
```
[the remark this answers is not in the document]
jshearer: Done

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -665,12 +673,14 @@ async def _binding_incremental_task(
...
                 continue
             else:
                 # We're idle. Sleep until the cursor is `interval` old.
                 sleep_for = binding.resourceConfig.interval - lag
         else:
             # We're not idle. Attempt to fetch the next changes.
+            sleep_for = timedelta()
             continue
 
         task.log.debug("incremental task is idle", {"sleep_for": sleep_for, "cursor": state.cursor})
```
jgraettinger, line 686: You have to move the below sleep up to the top, or you're sleeping twice.

jshearer: 🤦 I had that and it got lost when rebasing, thanks. Updated

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -611,14 +611,32 @@ async def _binding_incremental_task(
...
+        try:
+            if not task.stopping.event.is_set():
+                await asyncio.wait_for(
+                    task.stopping.event.wait(), timeout=sleep_for.total_seconds()
+                )
+
+            task.log.debug(f"incremental replication is idle and is yielding to stop")
+            return
+        except asyncio.TimeoutError:
+            pass  # `interval` elapsed.
```
jgraettinger, line 634: nit: update comment to `sleep_for` elapsed.

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -611,14 +611,32 @@ async def _binding_incremental_task(
...
 
         checkpoints = 0
         pending = False
+
```
jgraettinger, line 638: this will log every iteration and be quite noisy on logs (hubspot drives this many times a second), instead of once at startup as it had been. Move outside of loop?

jgraettinger (approved):
LGTM % comments -- need to fix log volume

estuary-cdk/estuary_cdk/capture/common.py
jshearer: moved

estuary-cdk/estuary_cdk/capture/common.py
jshearer: fixed

Issue #1576, opened 2024-05-13
source-postgres: Detect connection poolers during validation


willdonnelly:
At least one connection pooler in common use (the Supabase one, specifically) will silently ignore the `replication=database` connection parameter and just give clients a normal connection. Which obviously won't work for our purposes, and currently results in a cryptic `syntax error at or near "IDENTIFY_SYSTEM" (SQLSTATE 42601)` error message later on.

We should see if it's possible to detect the use of a connection pooler and provide a more useful error message during validation, so that users will be confronted with the problem and given a more concise "we need a direct connection URL" message while they're still in the UI setting up a capture and it's easy to go get the right value and change it.

One obvious way of detecting this might be to add `replication=database` to the connection config we use for prerequisite validation, and then issue that same `IDENTIFY_SYSTEM` command to make sure it's supported.

PR #1622, opened 2024-06-03
source-postgres: Special-case check for Supabase connection pooler

willdonnelly:
**Description:**

Adds a special case check for Postgres database addresses ending in `.pooler.supabase.com:6543` and `.pooler.supabase.com` (that latter address is _also_ invalid since the pooler isn't running  on the default Postgres port, but the user intent is clear so we might as well present the appropriate error message then too).

We should still try to add generic connection pooler detection logic in the future (https://github.com/estuary/connectors/issues/1576) but even after such logic exists it's probably still useful to have special cases for common ones like this so that we can provide an even more tailored error message. This one, for instance, will tell the user as clearly as I could manage that `address must be a direct connection: address %q is using the Supabase connection pooler, uncheck the 'Display connection pooler' option` and will use the "UserError" plumbing to strip out as much error prefixing as possible.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1622)

### DIFF
--- a/source-postgres/main.go
+++ b/source-postgres/main.go
@@ -121,6 +121,20 @@ func (c *Config) Validate() error {
 		}
 	}
 
+	// Connection poolers like pgbouncer or the Supabase one are not supported. They will silently
+	// ignore the `replication=database` option, meaning that captures can never be successful, and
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #1623, opened 2024-06-04
source-postgres-batch: Correctly handle XID wraparound

willdonnelly:
**Description:**

This PR modifies the `source-postgres-batch` connector in two related ways:

- The polling query is tweaked so that it should work correctly even when the XID epoch wraps around (provided that the source table is updated more frequently than that wraparound happens -- if that's an issue I have ideas and believe it's fixable, but it felt out of scope for the current goal).
- The resource config now leaves the template empty and expects the connector to supply the default behavior. Since the template used to hard-code in a table name, there are now two new resource config fields for the schema/table name, which simply get plumbed into the template. This means that if there's some minor issue with the polling query behavior we can fix it in the future by rolling out a new connector build, without having to go and update old captures which have "baked in" the query template initially provided by discovery.

I would have liked to have tested this PR more than I did, but exercising XID wraparound behavior in small-scale local testing is damnably hard. I have substituted a small amount of semi-manual testing involving `pg_resetwal` plus a whole bunch of thinking about the problem, so hopefully I didn't overlook anything important.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1623)

### DIFF
--- a/source-postgres-batch/main.go
+++ b/source-postgres-batch/main.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
+	"text/template"
 
 	"github.com/estuary/connectors/go/schedule"
 	schemagen "github.com/estuary/connectors/go/schema-gen"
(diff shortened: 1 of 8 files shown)


source-postgres-batch/main.go
```diff
@@ -105,46 +106,55 @@ func connectPostgres(ctx context.Context, cfg *Config) (*sql.DB, error) {
...
+// to integers (by way of text) and then reimplement the desired ordering behavior ourselves.
+//
+// Given a particular "cursor" value obtained from a previous polling query execution, we can
+// begin by assuming that there are fewer than 2^31 "live" XIDs greater than that cursor. The
+// issue is that with XID wraparound some of these values may be numerically smaller. Given a
+// wrapping uint32 wrapping subtraction operator the expression `xmin - cursor` would compute
+// an orderable count of "how far past the previous cursor" a given row is.
+//
+// We can emulate wrapping uint32 subtraction using PostgreSQL int64 values by simply writing
+// the expression `((x::bigint - y::bigint)<<32)>>32`.
```
mdibaiee, line 126: I don't really understand how this is a wrapping subtraction, can you give a numerical example of a wrapped `xid` value and a `cursor` value and show how this calculation unfolds in practice please?

willdonnelly: Suppose we have a table with four rows:
* Row A has `xmin = 0xFFFFC000`
* Row B has `xmin = 0xFFFFF000`. This is our cursor value, in this example.
* Row C has `xmin = 0xFFFFFFFF`
* Row D has `xmin = 0x00001000`

The correct comparison operation on XIDs is the u32 comparison-via-subtraction-and-sign-check function `x - y > 0`. In normal non-wrapping cases it should be obvious that this works because it's equivalent to `x > y`, and in wrapping cases we get the following results:

- Row A `0xFFFFC000 - 0xFFFFF000 = -0x3000`
- Row B `0xFFFFF000 - 0xFFFFF000 = 0`
- Row C `0xFFFFFFFF - 0xFFFFF000 = 0xFFF`
- Row D `0x00001000 - 0xFFFFF000 = 0x2000`

The `(x<<32)>>32` operation takes the bottom 32 bits of an int64 quantity, shifts them up to the top of the int64 so that the 32-bit MSB becomes the int64's MSB, and then shifts them back down to the normal position. This is a fairly common idiom for "sign-extend an N-bit quantity into an M-bit one (where M > N)".

willdonnelly: Working out the math on that final one in more detail, we have something like:

```
(((0x00001000::bigint - 0xFFFFF000::bigint)<<32)>>32)::int
(((0x0000000000001000 - 0x00000000FFFFF000)<<32)>>32)::int
((0xFFFFFFFF00002000<<32)>>32)::int
(0x0000200000000000>>32)::int
0x0000000000002000::int
0x00002000
```

But the final cast back to `::int` is irrelevant and unnecessary for our purposes (we just want to filter for positive results and order by the result value) so we skip it.

willdonnelly:
Merging without LGTM because we've got a user anxious to try this out. I'm reasonably confident that this PR is correct, and thanks to the changes around default template handling it should be less of an issue if we have to issue a followup to fix something.

PR #1630, opened 2024-06-05
source-dynamodb: deletion annotation in discovered schemas

mdibaiee:
**Description:**

- Add deletion annotation to discovered schemas of source-dynamodb

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1630)

### DIFF
--- a/source-dynamodb/.snapshots/TestDiscovery-additional_table_without_stream_enabled
+++ b/source-dynamodb/.snapshots/TestDiscovery-additional_table_without_stream_enabled
@@ -21,6 +21,28 @@ Binding 0:
       },
       "allOf": [
         {
+          "if": {
+            "properties": {
+              "_meta": {
+                "properties": {
+                  "op": {
+                    "const": "d"
+                  }
+                }
+              }
+            }
+          },
+          "then": {
+            "reduce": {
+              "delete": true,
+              "strategy": "merge"
+            }
+          },
+          "else": {
+            "reduce": {
+              "strategy": "merge"
+            }
+          },
           "required": [
             "_meta"
           ],
@@ -78,9 +100,6 @@ Binding 0:
                 "strategy": "merge"
               }
             }
-          },
-          "reduce": {
-            "strategy": "merge"
           }
         },
         {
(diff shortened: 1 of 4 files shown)


source-dynamodb/.snapshots/TestDiscovery-additional_table_without_stream_enabled
```diff
@@ -78,9 +100,6 @@ Binding 0:
                 "strategy": "merge"
               }
             }
-          },
```
williamhbaker, line 81: Why is this top-level merge going away? It is something that actually works with dynamodb, in contrast to say mongodb, because we produce explicit `null`'s for deleted fields: https://github.com/estuary/connectors/blob/ff661d58a92447b8f87708058408c234e26453cf/source-dynamodb/capture.go#L425-L432

mdibaiee: @williamhbaker the top-level merge is replaced by one in the "else" clause. In my prior testings I've realised the top-level reduce overrides the ones in "if" and "else", so even for deleted documents we end up with the top-level reduce stanza.

So this doesn't change the workings of the merge reduction strategy for the normal case, and adds a delete annotation to the merge strategy for the deleted document case.

----

I don't fully grok why we are setting fields to null though, is it necessary?

williamhbaker: The fields get set to `null` if they are completely absent in updated documents but were present in prior documents so that they are removed from a materialized table rather than being left there from the merge reduction, which treats explicit `null` differently from their absence.

For example, reducing two documents like this into a materialized table with a top-level `merge` strategy will result in columns with values `"first": "updated", "second": "value"` instead of `"first": "updated", "second": null`.
```json
{
  "first": "value",
  "second": "value"
}
```

```json
{
  "first": "updated",
}
```

To get materialized columns with `"first": "updated", "second": null`, we need the second document to be this:
```json
{
  "first": "updated",
  "second": null
}
```

williamhbaker (approved):

Issue #1637, opened 2024-06-07
materialize-databricks: doesn't respect "hard delete" config


danthelion:
With "hard delete" enabled, delete operations still show up in Databricks as separate records.

mdibaiee:
Fixed in #1640

PR #1638, opened 2024-06-07
materialize-*: hard deletion when doc created & deleted in one txn

mdibaiee:
**Description:**

- If a document is created and deleted in the same transaction, we do not want to insert that document. Prior to this pull-request we would end up inserting this document because it didn't already exist in the target table, but with this pull-request I'm filtering out documents which do not exist in the target table and are marked as deleted.
- This scenario is specially prominent during backfills
- Added a case in the integration test to account for this scenario
- Fixes https://github.com/estuary/connectors/issues/1637

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1638)


tests/materialize/fixture.json
```diff
@@ -73,6 +73,9 @@
 
     ["tests/deletions", { "id": 1, "_meta": {"op": "d"} }],
     ["tests/deletions", { "id": 2, "_meta": {"op": "u"} }],
-    ["tests/deletions", { "id": 3, "_meta": {"op": "c"} }]
+    ["tests/deletions", { "id": 3, "_meta": {"op": "c"} }],
+
+    ["tests/deletions", { "id": 4, "_meta": {"op": "c"} }],
```
williamhbaker, line 78: Is it now possible to get store requests for the same document key twice in the same transaction in the scenario that it is created and deleted within the same flow transaction? This implies that it must be, but I'm a bit surprised by that. In the case of non-hard deletes, it would seem to result in incorrect behavior for merge queries where we'd end up inserting a row for both the creation record and the deletion record and have two rows for the same key in the target table.

mdibaiee: This doesn't lead to two store requests for the same document. The confusion comes, I think, from the difference in behavior of flowctl in driving fixtures and the previous implementation. The previous implementation's fixtures were actual store requests but in this implementation the fixtures represent documents from a collection prior to reduction. In this example it means there are two documents that are being reduced in one transaction before being passed to the materialization. So in the end there is only one document for id: 4 in this transaction.

williamhbaker: Ah ok, that makes sense. But now I'm confused about the general premise. If a store request is received for a document and it's a deletion, that means that the document is known to exist in the destination, having been received in a loaded response, right? The change here seems like it would mitigate the scenario where we get deletion store request for a document that _doesn't_ exist in the destination, which as far as I understand it isn't possible.

Extending on this, I'm not sure what happens if a capture emits an `op: d` document prior to having ever emitted a "creation" document for that key. My assumption has been that the runtime would send a load request for the document key with the `op: d`, not get a loaded response back, and then...not send anything to the connector for the store of that document? And maybe that's the crux of my misunderstanding: Does the runtime actually send a store request for a document key that is a deletion, but the document _doesn't_ exist in the destination?

mdibaiee: The symptoms tell me that the runtime does send documents with op: d as store requests even if they don't already exist in the destination. @jgraettinger can you comfirm?

williamhbaker: Thinking it through a little more: The runtime probably does send `op: d` events even if they are not known to already exist in the destination, otherwise delta updates wouldn't work, and also it wouldn't be backward-compatible with materializations that don't have hard deletions enabled, which the runtime knowns nothing about.

So I'd guess that we would get a store request for a document with `op: d`, but with `Request_Store.Delete` as `false`. That would result in us inserting a document into the destination with `op: d`, even though hard deletion is enabled, and that would even happen with the change proposed here I think. But I haven't tested this myself or thoroughly reviewed the runtime code for how its expected to behave in the this circumstance.

mdibaiee: @williamhbaker that's not what's happening with this example. We are getting a request with `Delete: true` and that's why the integration tests work as they do right now.

Dani had seen this more explicitly in a materialization he deployed today which had documents inserted in his tables with `flow_document: "delete"`, which means `Delete: true` was set, but we inserted the document because the filter wasn't there

mdibaiee: Will close this pull-request. Turns out it is possible to have iterator instances with `it.Delete = true` and `it.Exists = false`, in which case we can ignore staging the files from the beginning to filter these items out for hard deletions. Will address in a new pull-request.

PR #1641, opened 2024-06-10
fix: orjson serialization of `decimal.Decimal`

jshearer:
**Description:**

Pydantic automatically serializes Decimals as strings, but orjson doesn't know about that. In order to handle this, we must tell `orjson` how to handle Decimals. [Their docs on this](https://github.com/ijl/orjson?tab=readme-ov-file#default)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1641)


jshearer:
Merging into one branch for easier PR review

PR #1660, opened 2024-06-17
source-oracle: logminer

mdibaiee:
**Description:**

New Oracle database connector that uses Logminer to capture change events from the database.

Notes:
- Due to the nature of Logminer, we need to parse Oracle queries in order to construct rows of data. I could not find a good Oracle SQL parser, so instead I forked https://github.com/vitessio/vitess into https://github.com/mdibaiee/vitess and made some changes to make it compatible with Oracle SQL. Note that the range of SQL queries that Logminer constructs are somewhat limited, since the queries are constructed by Logminer based on internal data it has about the change events, so they are not arbitrary user SQLs. I have written tests trying to cover potential edge cases that could arise in the SQL parsing part (e.g. long strings that don't fit in a single Oracle string literal, unicode characters, special characters and escaped single quotes, etc.)
- The default integer, smallint and other numeric types of Oracle are larger than what fits in an integer, so by default they end up as `type: "string", format: "number"`. This may not be what users expect intuitively, although it is the _correct_ format.
- Two dictionary modes are supported: `extract` which extracts the dictionary from log files (which takes longer and more resources) supporting schema changes, and `online` which is faster and less resource-intensive but does not handle all schema changes gracefully. There are tests covering both cases.
- The network tunnel changes are necessary when using network tunnel in tests: we would previously not clean up the network tunnel process, assuming that the docker container running the connector would be killed anyway, but in tests the process would become a zombie.
- Some changes in testsuite were necessary to allow Oracle to use them: using shorter names in table names (Oracle Logminer doesn't support names longer than 30 characters) and default uppercase names
- I have not yet tested thoroughly on multitenant oracle databases and their various configurations

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1660)


sqlcapture/capture.go
```diff
@@ -832,5 +835,5 @@ type StreamID = string
 
 // JoinStreamID combines a namespace and a stream name into a dotted name like "public.foo_table".
 func JoinStreamID(namespace, stream string) StreamID {
-	return strings.ToLower(namespace + "." + stream)
+	return namespace + "." + stream
```
williamhbaker, line 838: To ask the obvious question: Are we certain changing this doesn't break some assumption elsewhere in a SQL capture about the `StreamID` being lower-case? Spot checking I can't find anything obvious, but I didn't do a deep dive.  I'm assuming the reason for making this change is that Oracle names default to uppercase, and I see above that we have to do an extra check for the capitalized `"WATERMARK"`.

I'm wondering if the Oracle capture could set its `event.After` to have `"watermark"` as a lower case, and then we could keep the Stream ID creation unchanged?

williamhbaker (commented):
I did a small initial review looking at some of the peripheral changes and have a couple of questions. Haven't looked at the actual bulk of the Oracle capture code yet, and will need to come back to that a little later. But I thought these comments might be useful for now.

go.mod
```diff
@@ -1,140 +1,158 @@
...
 	github.com/jackc/pgx/v5 v5.6.0
-	github.com/jmoiron/sqlx v1.3.3
-	github.com/klauspost/compress v1.17.8
-	github.com/marcboeker/go-duckdb v1.7.0
+	github.com/jmoiron/sqlx v1.4.0
+	github.com/klauspost/compress v1.17.9
+	github.com/marcboeker/go-duckdb v1.5.6
 	github.com/mattn/go-sqlite3 v2.0.3+incompatible
-	github.com/microsoft/go-mssqldb v0.19.0
+	github.com/mdibaiee/vitess v0.15.7
```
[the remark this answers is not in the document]
mdibaiee: Vitess is pretty large, although the sqlparser part may be smaller. It may be possible to extract the sqlparser part only. What we could do is fork vitess into an estuary repository, and strip it down to the sqlparser only.

Note: you will never lose access to my fork even if I make it private, see: https://trufflesecurity.com/blog/anyone-can-access-deleted-and-private-repo-data-github

sqlcapture/capture.go
mdibaiee: Yes this is to accommodate Oracle names being uppercase by default... I'm also not completely easy with this change.

Is there a reason for us to lowercase all streamIDs though, other than to avoid duplicate streamIDs (which I think should be forbidden before sqlcapture gets involved anyway)? @willdonnelly perhaps you may know something about this?

go.mod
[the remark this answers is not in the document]
mdibaiee: I tried stripping the repo down to sqlparser, but sqlparser depends on so many packages in the repository, and they subsequently depend on others, so I left the repo as-is but forked it by estuary: github.com/estuary/vitess

source-oracle/backfill.go
```diff
@@ -0,0 +1,383 @@
...
+
+func scanToMap(rows *sql.Rows, cols []string, colTypes map[string]oracleColumnType) (map[string]any, error) {
+	var fieldsArr = make([]any, len(cols))
+	var fieldsPtr = make([]any, len(cols))
+	for idx, col := range cols {
+		if col == "ROWID" {
+			var rowid = ""
+			fieldsArr[idx] = &rowid
+		} else {
+			fieldsArr[idx] = reflect.Zero(colTypes[col].t).Interface()
```
williamhbaker, line 163: Does the oracle driver require the placeholder values to have the appropriate concrete type for the interfaces that will receive their values? So they can't be just an `interface{}`? Or get them via `rows.ColumnTypes`?

source-oracle/main_test.go
```diff
@@ -0,0 +1,289 @@
...
+			return
+		}
+	}
+	defer rows.Close()
+	// Log the response, doing a bit of extra work to make it readable
+	cols, err := rows.Columns()
+	if err != nil {
+		t.Fatalf("unable to get columns: %v", err)
+	}
+	colTypes, err := rows.ColumnTypes()
```
williamhbaker, line 176: `colTypes` isn't use here

source-oracle/discovery.go
```diff
@@ -0,0 +1,384 @@
...
+	var pks = make(map[string][]string)
+	var columns []sqlcapture.ColumnInfo
+	var sc sqlcapture.ColumnInfo
+
+	var ownersMap = make(map[string]bool)
+	for _, t := range tables {
+		ownersMap[t.Schema] = true
+	}
+	var owners []string
+	for k, _ := range ownersMap {
```
williamhbaker, line 294: Can drop the `_` here

williamhbaker (commented):
I got through the interesting oracle-specific code and left some comments!

source-oracle/prerequisites.go
```diff
@@ -0,0 +1,102 @@
...
+					"columnName": colName,
+				}).Warn("column names longer than 30 characters are not supported for capturing, column will be ignored.")
+			}
+		}
+	}
+
+	return nil
+}
+
+func (db *oracleDatabase) SetupTablePrerequisites(ctx context.Context, owner, tableName string) error {
```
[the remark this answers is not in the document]
mdibaiee: This is `SetupTablePrerequisites` which does not have a slice pattern, but `SetupPrerequisites` does

source-oracle/main.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+func (db *oracleDatabase) FallbackCollectionKey() []string {
+	return []string{"/_meta/source/row_id"}
+}
+
+func encodeKeyFDB(key any, colType oracleColumnType) (tuple.TupleElement, error) {
+	if colType.jsonType == "integer" {
+		return key.(int64), nil
+	} else if colType.jsonType == "number" {
```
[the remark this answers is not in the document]
mdibaiee: @williamhbaker it should not be possible, in `discovery.go` primary keys are discovered as `format: "integer"` if they can't be `type: "integer"`, this is a sanity check here

source-oracle/discovery.go
```diff
@@ -0,0 +1,390 @@
...
+		// so here we ignore primary key information for tables we don't care about.
+		var info, ok = tableMap[streamID]
+		if !ok {
+			continue
+		}
+		logrus.WithFields(logrus.Fields{
+			"table": streamID,
+			"key":   key,
+		}).Trace("queried primary key")
+		info.PrimaryKey = key
```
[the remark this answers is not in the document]
mdibaiee: It is not possible for us to discover a primary key with `format: "number"`, the check is done in `getColumns`, lines 309 to 320:

```
		if dataType == "NUMBER" && (dataScale.Valid && dataScale.Int16 == 0 && precision <= 18) {
			t = reflect.TypeFor[int64]()
			jsonType = "integer"
		} else if slices.Contains([]string{"NUMBER", "DOUBLE", "FLOAT"}, dataType) {
			if isPrimaryKey || precision > 18 {
				t = reflect.TypeFor[string]()
				format = "integer"
				jsonType = "string"
			} else {
				t = reflect.TypeFor[float64]()
				jsonType = "number"
			}
```

source-oracle/main_test.go
mdibaiee: It is used a few lines below in
```
			fields[col] = reflect.New(colTypes[idx].ScanType()).Interface()
```

source-oracle/backfill.go
mdibaiee: Huh, I don't know why I thought it is necessary to have the concrete type for each column, I switched to using `new(any)` to simplify the code (and improve performance, reflect is time-consuming)

source-oracle/replication.go
```diff
@@ -0,0 +1,899 @@
...
+	s.tables.keyColumns[streamID] = keyColumns
+	s.tables.discovery[streamID] = discovery
+	s.tables.Unlock()
+	return nil
+}
+
+// Acknowledge informs the ReplicationStream that all messages up to the specified
+// SCN have been persisted
+func (s *replicationStream) Acknowledge(ctx context.Context, cursor string) error {
+	logrus.WithField("cursor", cursor).Debug("advancing acknowledged SCN")
```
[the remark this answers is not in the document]
williamhbaker: The `startCursor` is from the connector's persisted state (see [here](https://github.com/estuary/connectors/blob/051a0503d1f4023c6bda58a635684f185b6976b3/sqlcapture/capture.go#L163) for the invocation of `ReplicationStream`).

I believe the only current SQL capture that does anything with runtime acknowledgements is `source-postgres`, which [uses the runtime acknowledgements](https://github.com/estuary/connectors/blob/051a0503d1f4023c6bda58a635684f185b6976b3/source-postgres/replication.go#L664-L701) that indicate that the checkpoint (and its documents) has been durably committed to Flow to advance the postgres transaction log to a specific LSN, basically telling postgres that the connector will never need to return to any prior LSN, which is known since it has gotten confirmation from the runtime that the commit up until that LSN was durably committed. Other SQL captures (ex: [sqlserver](https://github.com/estuary/connectors/blob/051a0503d1f4023c6bda58a635684f185b6976b3/source-sqlserver/replication.go#L161-L163), [mysql](https://github.com/estuary/connectors/blob/051a0503d1f4023c6bda58a635684f185b6976b3/source-mysql/replication.go#L882-L890)) do not have anything in particular to do with runtime acknowledgements.

As the code is written here the `cursor` is being set to `s.ackSCN`, which appears to be similar to `source-postgres`, but then `s.ackSCN` is never accessed anywhere else. I would say at least as its written this could instead be a no-op like sqlserver/mysql/etc., unless there is actually something oracle should be doing in response to runtime acknowledgements, like informing the server that it can delete files or something. If it's possible or necessary to do that I don't see it happening anywhere in the code yet.

source-oracle/backfill.go
```diff
@@ -0,0 +1,383 @@
...
+		columnTypes[name] = column.DataType.(oracleColumnType)
+	}
+
+	// Compute backfill query and arguments list
+	var query string
+	var args []any
+	switch state.Mode {
+	case sqlcapture.TableModeKeylessBackfill:
+		// A is the lexicographically smallest character in base64 encoding, so this is the smallest possible base64 encoded string
+		var afterRowID = "AAAAAAAAAAAAAAAAAA"
```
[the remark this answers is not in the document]
mdibaiee: According to this page, yes, it is 18 characters: https://www.orafaq.com/wiki/ROWID

"AAAAAAAAAAAAAAAAAA" would stand for a row which is the first row, in the first block inside the first file in the tablespace of the first data object. I think it is safe to assume the first data object is going to be a system data object and data objects of the USERS tablespace usually have much higher data object IDs.

I checked data object IDs using this query to confirm the pattern:
```
select * from dba_objects WHERE data_object_id is not null;
```

source-oracle/capture_test.go
```diff
@@ -0,0 +1,369 @@
...
+			t.Errorf("materialized view returned by catalog discovery")
+		}
+	}
+}
+
+func TestAllTypes(t *testing.T) {
+	var unique = "18110541"
+	var tb, ctx = oracleTestBackend(t), context.Background()
+	var typesAndValues = [][]any{
+		[]any{"nvchar2", "NVARCHAR2(2000)", "nvarchar2 value with unicode characters ❤️ \\ 🔥️'')"},
```
williamhbaker, line 44: The `[]any` type specification for each item in the `[][]any` is redundant and can be removed - flagging this since `gopls` flags it.

source-oracle/replication.go
```diff
@@ -0,0 +1,902 @@
...
+			val += string(c)
+		}
+	}
+
+	return val, nil
+}
+
+// decode SQL values extracted from the sql queries returned by logminer and parsed by sqlparser, into values
+// to be used in the document
+func decodeValue(colInfo sqlcapture.ColumnInfo, expr sqlparser.Expr, colName string, originalSql string) (any, error) {
```
williamhbaker, line 380: `colInfo`, `colName`, and `originalSql` aren't used anywhere in this function - should they be? Or maybe the can be removed as inputs?

source-oracle/replication.go
```diff
@@ -0,0 +1,902 @@
...
+			return val, nil
+		// Used for DATE values
+		case "TO_DATE":
+			var val = v.Exprs[0].(*sqlparser.Literal).Val
+			if t, err := time.Parse(PARSE_DATE_FORMAT, val); err != nil {
+				return nil, fmt.Errorf("invalid date %q: %w", val, err)
+			} else {
+				return t.Format(OUT_DATE_FORMAT), nil
+			}
+			return val, nil
```
williamhbaker, line 418: Line 418 is not reachable

williamhbaker (approved):
LGTM % outstanding comment about the `JoinStreamID`. Being extra careful about that one since it effects all existing SQL captures rather than being a net-new thing just for oracle. 

Also left a few more small nits. And I understand that there is more performance tuning going on, so let me know if anything significant comes out of that.

sqlcapture/capture.go
willdonnelly: There used to be good reason, but we've been slowly working towards fixing that, so we're almost at the point where they can be made case-sensitive now.

The last major roadblock, IIRC, was resolved in the state key migration. It'd be a good idea to double check, but I think that with that changed we're no longer persisting StreamID values across connector runs, so it might be as simply as just removing it. We'd want to double-check pretty thoroughly that it doesn't break anything in production before rolling that out, but it might be time to bite that bullet.

Also if/when we make it case sensitive I'd really like to try redefining the StreamID type as something like `struct { Schema, Table string }` so it actually has proper type safety and isn't just an alias for a string.

mdibaiee:
Okay I had to go through a bunch of changes to:
1. Improve performance of replication: This is done by refactoring how `receiveMessages` and `decodeMessages` work: we stream messages from `receiveMessages` to a channel, with a `decodeMessageWorker` decoding the messages and emitting them, no longer waiting until all messages are stored in an array.
2. Another improvement to performance comes from filtering the tables we really need to process using their object and data object IDs. Note that we don't filter with table name since if there is a missing dictionary, we want to still receive the event with a missing dictionary and error. With this filter, we avoid receiving a lot of messages which are not relevant to us (and in my tests, there can be tons of these messages on system tables with missing dictionaries, congesting our query).
3. Handle cases with a large SCN gap between two runs of the connector (when we need to replicate a large SCN range we used to choke up): This is done by handling a specific SCN range at a time, instead of trying to replicate all of `stateSCN...lastSCN`, which can sometimes be huge (I've seen cases with more than 500 archive files). This way we limit the amount of data that needs to be processed at once. The current configured default for this SCN range is probably not the most efficient for all cases. We may want to somehow "adapt" this value depending on the target database's traffic.
4. Handle cases where dictionary files may be missing from the logminer session, leading to `Dictionary Mismatch` errors: This is done by ensuring we always have a dictionary log file before or at the startSCN, and one at or after the endSCN.

You can see the diff from the last time you reviewed up until now here: https://github.com/estuary/connectors/compare/58f6f38...mahdi/oracle-sqlcapture

source-oracle/replication.go
```diff
@@ -0,0 +1,680 @@
...
+		// it is possible for there to be multiple rows with the same SCN
+		// and to be cautious we use the SCN as-is. This may lead to duplicate events being captured
+		// on restart, but it saves us from missing events
+		s.lastTxnEndSCN = endSCN
+
+		s.decodeCh <- logminerMessage{Op: opFlush, SCN: s.lastTxnEndSCN}
+	}
+}
+
+func (s *replicationStream) decodeMessageWorker(ctx context.Context) error {
```
williamhbaker, line 384: I'm curious how much of an improvement you saw from running the message decoding on a separate goroutine? I've tried things like this before and haven't been able to get much of a performance increase, possibly due to extra GC + channel overhead, and it increases memory usage and OOM potential if documents are very large, particularly if the `decodeCh` is buffered.

williamhbaker (approved):
Looked through the most recent set of commits & the updated code overall as best I could and the changes look reasonable. Just had a couple of questions. My biggest hangup is still the changes to `JoinStreamID`; we'll either want to really test/study that thoroughly or kick the can I think and figure out a way for oracle to work with it being case-insensitive to avoid changing it now. I don't think we should merge until that comment is resolved.

willdonnelly:
> My biggest hangup is still the changes to `JoinStreamID`; we'll either want to really test/study that thoroughly or kick the can I think and figure out a way for oracle to work with it being case-insensitive to avoid changing it now.

Just as a general style note, the way stream IDs are used within other SQL CDC connectors is strictly one-way -- you can generate an ID from a schema and table name but nowhere [1] do we ever split them back into those components, they're only human readable for clarity. The fact that case-sensitivity is required (for something other than supporting identically-named-except-for-casing tables) may be a sign that something else is wrong in there.

[1] IIRC there actually is one spot (SQL server prerequisite validation) where we do as an unprincipled hack, but that's bad and we probably shouldn't be doing it.

mdibaiee:
To avoid the change in `JoinStreamID` potentially breaking things, I introduced a temporary middle-way solution where the user of `sqlcapture` can ask for case-sensitive StreamIDs, and this is only applicable to Oracle now. We can do thorough testing for each connector step by step after that and migrate them over. This makes the migration also easier, as it can be done incrementally.

source-oracle/replication.go
mdibaiee: I'll do another round of testing with and without the `decodeCh` to see if there is still an improvement in performance, previously I had noticed there was too much time wasted on I/O blocking CPU processes, so I moved decode to its own worker. The decoding process is one of the most CPU-intensive parts of the program (parsing sql is CPU intensive). I'll see if that still applies after I changed how `receiveMessages` works

mdibaiee: I did some tests with two versions of the connector, one with `decodeCh` and one without.

The one with the channel shows ~10% better performance on 1 million documents (500MB) in a single SCN span (one run of `receiveMessages`). However when working on larger data in a single SCN, the `decodeCh` version does indeed OOM while the normal version doesn't.

I think not OOMing is more important than the 10% performance improvement for now. Will have to optimise CPU utilisation of this connector at some stage and will address performance then. For now I'm going to revert back the `decodeCh` changes 👍🏽

sqlcapture/capture.go
willdonnelly: Okay, I've finally had some time to read through the latest code and now I'm even more confused why exactly the stream IDs need to be case sensitive for Oracle specifically.

As far as I can tell no stream ID is ever being split back into schema or table name components, so additional normalization should be fine unless we're specifically worried about capturing two tables with identical-except-capitalization names. Is that the only issue we're concerned with here, and if so is there some reason that's a bigger issue with Oracle than other databases?

mdibaiee: @willdonnelly The main issue arises here:
https://github.com/estuary/connectors/blob/main/sqlcapture/main.go#L167-L170

The bindings are mapped keyed on the StreamID, but during discovery I want the table names to retain their original casing (uppercase by default in case of Oracle). If I do discovery with uppercase names, then I end up with this error due to the discrepancy:

```
error updating capture state: table "flow_test_logminer.t95911555" is a configured binding of this capture, but doesn't exist or isn't visible with current permissions
```

willdonnelly: I'm afraid I'm still confused, what exactly do you mean by "retain their original casing"?

The original schema and table names are present in the `DiscoveryInfo.Schema` and `DiscoveryInfo.Name` properties, which then become the resource spec `Namespace` and `Stream` properties. All of the generic discovery logic should be based on that and not care at all about the stream ID used for a table. I've just double-checked the relevant discovery code to be sure and I still see nowhere that the stream IDs should be making it into the actual discovered resources or schemas, but if that is happening it's a bug.

The bit of logic you linked should also work correctly if stream IDs are normalized to lowercase. The appropriately-capitalized schema and table name come from the resource spec, and a lowercase stream ID is generated from them and used to look up a value from the discovery output. The key of the discovery map is also a lowercased stream ID generated from the appropriately-capitalized schema/table names that discovery just found. Where does the problem come from?

Like, in principle we ought to be able to just come along one day and swap out the implementation of `JoinStreamID` for something that hashes the names and everything else would keep going without caring. The actual representation of these IDs isn't supposed to matter and if it does that implies a bug somewhere else.

So the only thing that should be broken by normalizing stream IDs to lowercase is that we can't support tables with the exact same name but differently capitalized. I'd like to make them case sensitive at some point so that stops being an issue for all SQL CDC connectors, but it doesn't sound like that's the problem you're trying to avoid here? The fact remains that **if anything else is broken by this, the IDs are being used incorrectly somewhere**.

mdibaiee: @willdonnelly okay that's a good perspective on how streamID is supposed to be handled. I dug deeper and I found what was causing this issue:

1. The `DiscoverTables` method that is to be implemented by each connector is expected to have `StreamID` keys, and in many places in `sqlcapture` we use an `StreamID` to check for existence of a table
3. In case of Oracle, the `DiscoverTables` is using `sqlcapture.JoinStreamID`, so the discover tables is keyed on `StreamID`s as expected
4. However, in the case of the watermarks table, the check is done like so: https://github.com/estuary/connectors/blob/main/sqlcapture/capture.go#L174-L177 which expects the value returned by `WatermarksTable()` method to match one of the keys in the discovery output. In this case, the Oracle connector was outputting an all-uppercase `<USER>.FLOW_WATERMARKS` table name, which would not match the lowercased keys of discovery.
5. This was causing the following error:
```
error activating replication for watermarks table %q: table was not observed by autodiscovery
```
6. The issue this arises is that if the user enters a watermarks table that is case-sensitive, we will fail with the same error, since the case-sensitive table name will not match its lowercased counterpart. This seems to affect all connectors, but this is probably more common among Oracle users to want to enter their watermarks table in uppercase. So to _workaround_ this limitation we will need to also lowercase the `WatermarksTable()` output of Oracle

So I think the issue here is that we are using `JoinStreamID` for keys of the discovery, but not for checking the existence of `WatermarksTable()`, so there is a possibility for inconsistency here.

willdonnelly: Oh, interesting. Yeah, it makes sense it'd be watermarks table handling that causes the problem there, since that's the one spot we're exposing something like a stream ID directly to the user.

I have a vague notion of how to fix this (basically recognizing that the watermarks table name is its own thing which is not necessarily a stream ID, and modifying the appropriate interface bits to split `cfg.Advanced.WatermarksTable` explicitly and then feed the schema/table name components into JoinStreamID) but I'll have to sit down and try it to be sure if that would work.

mdibaiee: I think for now I'll take this limitation of case-sensitive watermarks tables, put a `strings.ToLower` around the watermarks table configuration of Oracle. We should definitely revisit this though, I think your idea sounds good both for addressing the watermarks table, and perhaps to generally move from the current `StreamID` into an actual struct

willdonnelly: Yeah, that's basically the end goal of making sure that nothing else cares about the internal representation of a stream ID -- if it doesn't matter then we can just swap it over to being a struct and nothing will care, at which point we can be less vigilant against guarding against any sort of "misuse" of the values.

But none of the other connectors handle watermark table names correctly at this spot either, so just lowercasing seems fine for now, and I'll put the same "split using a custom function and then recombine with JoinStreamID" behavior in the Oracle connector when I do the others.

PR #1665, opened 2024-06-19
source-hubspot-native: emit `null` instead of empty string for date and datetime

williamhbaker:
**Description:**

This will allow schema inference on the collection to infer the appropriate format, rather than the `""` string forcing it to be a string with no format.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api.py
+++ b/source-hubspot-native/source_hubspot_native/api.py
@@ -137,7 +137,7 @@ async def fetch_page(
 
     for doc in output:
         if doc.updatedAt < cutoff:
-            yield doc
+            yield await _fix_formats(log, http, object_name, doc)
 
     if result.paging:
         yield result.paging.next.after
@@ -267,7 +267,7 @@ async def fetch_changes(
             log, cls, http, object_name, [id for _, id in batch]
         )
         for doc in documents.results:
-            yield doc
+            yield await _fix_formats(log, http, object_name, doc)
 
     if recent:
         yield recent[-1][0]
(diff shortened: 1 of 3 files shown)


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -484,6 +484,21 @@ async def fetch_recent_email_events(
     if yielded_event:
         yield max_ts + timedelta(milliseconds=1) # startTimestamp is inclusive.
 
+async def _fix_formats(
+    log: Logger,
+    http: HTTPSession,
+    object_name: str,
+    doc: CRMObject,
+) -> CRMObject:
+    property_types = {p.name: p.type for p in (await fetch_properties(log, http, object_name)).results}
```
jgraettinger, line 493: this looks ... expensive. I recognize it's cached, but we're still building a coroutine and new dict entry for every property of every document

Stepping back, can we do the really basic thing of filtering out any empty property value from the documents properties, regardless of its type? This seems just as much a possible problem for properties which are string-formatted numbers, or emails, or other things.

Is it a problem for regular string properties having empty values to map to NULL ?

williamhbaker: The only slightly weird thing that could happen if we completely filter a property from the properties object is that it would still show up in the property history as an empty string. I don't know if that is a problem or not...probably not.

And I guess I'm not specifically aware of any problem that would arise from filtering all empty string properties regardless of their types, but I can't say with 100% confidence that is true. I was thinking of a case where there is some difference to the user between `""`, `"something"`, and the complete absence of the thing. But I do think you are right that formatted numbers would also have this problem. So, doing the really basic thing you suggest is probably fine, and we can revisit later if it's an issue.

jgraettinger (approved):
LGTM

dyaffe:
For someone to take advantage of this that previously had the hubspot connector running, would they have to re-backfill everything from the source and destination with a `notBefore`?

williamhbaker:
> For someone to take advantage of this that previously had the hubspot connector running, would they have to re-backfill everything from the source and destination with a `notBefore`?

I believe the most direct way would be to re-capture into new collections that would have a new schema inferred, and then materialize those. Only re-backfilling the capture into the same collection would not work (without other interventions) since the inferred schema would still be `type: string` with no format for these fields.

dyaffe:
Okay -- sounds like theoretically this is ready and I can suggest that.

PR #1668, opened 2024-06-20
source-oracle-flashback: fix network tunnel config rendering

mdibaiee:
**Description:**

- When explicitly specifying the type as `NetworkTunnel | None`, we end up with an `anyOf: [{ networkTunnel: ... }, { type: null }]` which ends up rendering as two tabs in the UI. I realised it is possible in pydantic to not specify the type as a union with None, and still provide a `default=None` to mark the field as not required, so this is the solution.
- Add network-tunnel binary to the Dockerfile for source-oracle-flashback. Since tests were running on the host machine I had not realised the network-tunnel is missing from the Dockerfile. Updated CI so it allows for custom Dockerfiles for python connectors. Tested this using a local run of the connector with SSH tunnelling using the docker image

Tested using JSONForm test renderer:
image

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1668)


source-oracle-flashback/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    openssh-client
+
+LABEL FLOW_RUNTIME_PROTOCOL=${CONNECTOR_TYPE}
+LABEL FLOW_RUNTIME_CODEC=json
+LABEL dev.estuary.usage-rate=${USAGE_RATE}
+
+COPY --from=builder /opt/$CONNECTOR_NAME /opt/$CONNECTOR_NAME
+COPY --from=builder /opt/estuary-cdk /opt/estuary-cdk
+COPY --from=builder /opt/venv /opt/venv
+COPY --from=ghcr.io/estuary/network-tunnel:dev /flow-network-tunnel /usr/bin/flow-network-tunnel
```
williamhbaker, line 42: Can we add this to copy in the network tunnel into the base python connector image instead? Most connectors won't need it but others in the future probably will and that way they wouldn't all need to have separate docker files and it keeps the github workflow simpler. It doesn't seem like it would be a problem for connectors to have the network tunnel binary backed in to their running dockerfile even if they didn't end up needing it.

mdibaiee: I think it is okay to add network tunnel to the base image, but I think we will end up needing custom Dockerfiles anyway, so I kept that part but commented it

williamhbaker (approved):

PR #1678, opened 2024-06-25
sqlcapture: Tweak table activation error messages

willdonnelly:
**Description:**

These messages were originally written before we had prerequisite validation for a lot of the basic errors, which is why the error for the watermarks table not showing up via discovery is "table doesn't exist" -- because that used to be by far the most likely reason we'd see that.

But now we already check for the basics so we should really give the wordier and more complicated but actually precise description of what's gone wrong there.

Then while I was at it I tweaked a couple of adjacent messages for consistency.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1678)


williamhbaker (approved):
LGTM

PR #1684, opened 2024-06-27
source-mysql: Support for MySQL v8.4

willdonnelly:
**Description:**

This PR fixes the `source-mysql` connector to support MySQL 8.4, which was released a couple of months ago. This involved a couple of trivial incompatibilities and a version bump to the CI database dockerfile, and then the main thing was that MySQL 8.4 went and fully removed the deprecated `SHOW MASTER STATUS` command.

This command was deprecated in MySQL 8.0.22 in favor of a new `SHOW BINARY LOG STATUS` which does the exact same thing. Since we want to support MySQL <8.0.22 we used the old query exclusively, but now that's no longer sufficient and there is no single query that works on all versions of MySQL. So to handle this we have to actually keep track of the database version and issue the appropriate query depending on what server we're connected to.

This was done by factoring out most of the version string querying/parsing logic from the `prerequisiteVersion` check into a helper function which gets run immediately upon connecting and which saves the resulting version info into some fields of the database struct, and then factoring out the `SHOW MASTER STATUS` query into another helper function which selects the appropriate query to run based on those properties.

This seems to work, although the refactor of the version-checking logic has me feeling a bit cautious so I'll probably do a careful rollout tomorrow where I'll force a restart of all production MySQL capture tasks and make sure none of them suddenly start failing, while I'm still sitting at my desk and thinking about it so it can be reverted quick if anything goes wrong.

(Honestly this keeps coming up, it might be time to write a helper script that can query the set of live tasks running a particular connector, figure out which ones are currently green, issue a forced restart to all of them, and then complain if any of them start failing during some short window after the restart. I know I'll want it for the Postgres "slot validation redux" PR whenever I get around to merging that...)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1684)

### DIFF
--- a/source-mysql/database.go
+++ b/source-mysql/database.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/estuary/connectors/sqlcapture"
+	"github.com/sirupsen/logrus"
+)
+
+// queryDatabaseVersion examines the server version string to figure out what product
+// and release version we're talking to, and saves the results for later use.
+func (db *mysqlDatabase) queryDatabaseVersion() error {
+	var version string
+	var results, err = db.conn.Execute(`SELECT @@GLOBAL.version;`)
+	if err != nil {
+		return fmt.Errorf("unable to query database version: %w", err)
+	} else if len(results.Values) != 1 || len(results.Values[0]) != 1 {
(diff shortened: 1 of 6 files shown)


williamhbaker (approved):
LGTM

PR #1686, opened 2024-06-28
materialize-iceberg: table identifiers are always lowercase

williamhbaker:
**Description:**

At least with the AWS Glue catalog (which is the only one we use right now), table identifiers are always lowercase. So if you have a resource with a table that has uppercase letters, right now we can't find that table. This fix will have use always treat table identifiers with lowercase, which should fix that problem.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1686)


willdonnelly (approved):
LGTM

PR #1688, opened 2024-07-01
source-oracle-flashback: skip flashback_on check on RDS instances

mdibaiee:
**Description:**

- RDS instances report flashback as OFF because it cannot be enabled on the whole database, but it is enabled for tables. On RDS instances we skip the flashback_on check.
- Users will still end up needing to skip Flashback checks since RDS instances usually have small undo retentions which are problematic, but at least the error is not misleading

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1688)

### DIFF
--- a/source-oracle-flashback/config.yaml
+++ b/source-oracle-flashback/config.yaml
@@ -21,8 +21,8 @@ sops:
     azure_kv: []
     hc_vault: []
     age: []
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

PR #1690, opened 2024-07-01
Revert "source-postgres: Additional replication slot validation, take two"

willdonnelly:
Reverts estuary/connectors#1679

I have a suspicion that something is going wrong here, probably related to the assumption that our checkpointed restart LSN should always be greater than or equal to the LSN we've confirmed with Postgres. Reverting out of caution to hopefully avoid any lasting prod breakage, we can always roll this forward again if it turns out I'm wrong.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1690)


willdonnelly:
Merging without review for expediency (and because this is just an automated revert of a PR merged earlier today)

PR #1691, opened 2024-07-02
source-oracle-flashback: fix backfill_chunk_size iterations

mdibaiee:
**Description:**

- When backfills had more documents than the backfill_chunk_size, we would loop and run the same query again and again with the same `ROWID` filters, which meant we emitted duplicate documents until a new sessions was started with the correct new `ROWID`. This was a regression in how we form the query, and was missed since the tests wouldn't go over the backfill_chunk_size. Added a new test to cover this case.
- Meanwhile added a test for incrementally capturing multiple transactions in a single session, which also uncovered a bug in that routine.

**Workflow steps:**

**Notes for reviewers:**


williamhbaker (approved):
LGTM

PR #1692, opened 2024-07-02
source-klaviyo: Adding event_name field to Events stream

Luishfs:
**Description:**

A user requested the "event_name" field to be included on the "Events" stream schema.
This PR adds a call to the metrics endpoint in order to get the event_name value, since the Events stream originally does not return it.   

**Notes for reviewers:**


jonwihl (approved):
LGTM

PR #1694, opened 2024-07-03
source-oracle-flashback: filter out schemas at query time

mdibaiee:
**Description:**

- Previously we would filter out schemas in code after having queried tables from all schemas. This was not the most efficient solution and would cause the error below if there were more than 1000 schemas (even if they specified advanced.schemas config):
```
ORA-01795: maximum number of expressions in a list is 1000
```
- A customer actually hit this error, which was due to the filtering we do in `fetch_columns` on schemas.
- Move the filtering logic to the `fetch_tables` query so that subsequently `fetch_columns` will also only filter on the limited schemas

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1694)

### DIFF
--- a/source-oracle-flashback/source_oracle_flashback/api.py
+++ b/source-oracle-flashback/source_oracle_flashback/api.py
@@ -9,7 +9,7 @@ import time
 import tempfile
 import json
 import pytz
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #1695, opened 2024-07-03
source-klaviyo: fixing event_name field

Luishfs:
**Description:**

event_name field was set to be created with a None/NULL value. Instead, it should've been created only with a string value, and if requirements to create it are not met, it should not be created at all. This commit fixes that, along with updating tests and schema.

**Notes for reviewers:**

Tested using flowctl and a local flow setup with a supabase DB as destination.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1695)


jonwihl (approved):
LGTM

PR #1697, opened 2024-07-03
source-postgres: Replication slot validation, take three

willdonnelly:
**Description:**

This is a more cautious attempt to get out the replication slot LSN checking.

The check was previously merged in https://github.com/estuary/connectors/pull/1653, then reverted because of a query failure on older database versions in which replication slots don't have a `wal_status` column. This was fixed and the check was merged again in https://github.com/estuary/connectors/pull/1679, which was then reverted because we saw failures in production which suggested that there might be something wrong with our assumption that the connector's restart checkpoint LSN would always be greater than or equal to the latest LSN confirmed with the database.

So this time we're taking it even slower. This PR adds the same `queryReplicationSlotInfo` helper function, a test case covering the behavior when a slot is dropped and then recreated, and the basic logic for the LSN comparison check, but the check itself has been nerfed so it will only ever log warnings. In addition the more comprehensive changes to the "does the slot exist" validation check have been omitted entirely, since they're not the most important part of this anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1697)

### DIFF
--- a/source-postgres/.snapshots/TestCaptureAfterSlotDropped-capture1
+++ b/source-postgres/.snapshots/TestCaptureAfterSlotDropped-capture1
@@ -1,0 +1,10 @@
+# ================================
+# Collection "acmeCo/test/test_captureafterslotdropped_46115540": 2 Documents
+# ================================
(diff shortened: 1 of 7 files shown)


williamhbaker (approved):
LGTM

PR #1698, opened 2024-07-05
source-oracle-flashback: handle TRUNCATEs by backfilling automatically

mdibaiee:
**Description:**

- Automatically trigger a backfill when a table is TRUNCATEd
- Still throw an error if the schema has changed (a re-discover is necessary in that case)

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1698)

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -1,4 +1,5 @@
 import abc
+from enum import Enum, auto
 import asyncio
 from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
@@ -45,6 +46,10 @@ These cursors are predominantly an opaque string or an internal offset integer.
 None means "begin a new iteration" in a request context,
 and "no pages remain" in a response context.
 """
+
+
+class Triggers(Enum):
+    BACKFILL = auto()
 
 
 class BaseDocument(BaseModel):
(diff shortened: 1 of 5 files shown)


estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -648,6 +653,12 @@ async def _binding_incremental_task(
             elif isinstance(item, AssociatedDocument):
                 task.captured(item.binding, item.doc)
                 pending = True
+            elif isinstance(item, Triggers) and item == Triggers.BACKFILL:
+                task.log.info("triggering an automatic backfill")
+                task.checkpoint(ConnectorState(
+                    bindingStateV1={binding.stateKey: None}
```
williamhbaker, line 659: This seems like it would race with an in-progress backfill in [`_binding_backfill_task`](https://github.com/estuary/connectors/blob/1b9fac9d75776cb6f2d924ca1bb19370d871f1be/estuary-cdk/estuary_cdk/capture/common.py#L581-L595), where the backfill state might be cleared here but then clobbered by a checkpoint emitted there.

I don't immediately know of an easy way to accomplish what I think we're wanting here; for the incremental task to be able to modify the state and processing of the backfill task. Perhaps an `event` that the incremental task would `set` to communicate to an in-progress backfill that it should reset its state and start over...but it would also need to work if the backfill is already complete.

Stepping back a bit I'm also wondering about the motivation for this. Would the user expect the destination to be truncated as well? Or deletion events to be emitted for the truncated documents? We don't generally support truncation well in other captures so I'm wondering if we need to do anything more sophisticated than throw an error here like we have already been doing.

PR #1702, opened 2024-07-08
source-klaviyo: ignoring 503 HTTP error for Profiles and Global Exclu…

Luishfs:
…sion streams

**Description:**

Both `Profiles` and `Global Exclusions `streams are raising a 503 HTTP error due to pagination issues. Pagination still believe its returning data, but when accessing the next url for the Profiles search, Klaviyo returns a 503 error and capture breaks. To fix this for now, im ignoring raises on htpp and retries on 503 responses for both streams.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1702)


jonwihl (approved):
LGTM

PR #1705, opened 2024-07-09
source-postgres: Add check for active replication slot

willdonnelly:
**Description:**

One easily-overlooked way that one might end up with a task's restart LSN being less than the slot's confirmed flush LSN is if there are multiple captures from the same database trying to use the same replication slot.

We should at least try to avoid giving a misleading "the slot was probably dropped and recreated" message in this case, which can be done by checking whether the slot is already active before we've started replication.

Note that this isn't bulletproof, since it's possible for the other of the two misconfigured captures to be momentarily offline when we're doing this check, but it's still better than nothing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1705)

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -69,6 +69,8 @@ func (db *postgresDatabase) ReplicationStream(ctx context.Context, startCursor s
 		// completely unable to produce a fatal error, and double-checking to make extra sure we don't
 		// accidentally dereference a null pointer doesn't really hurt anything.
 		logrus.WithField("slot", slot).Warn("replication slot has no metadata (and probably doesn't exist?)")
+	} else if slotInfo.Active {
+		logrus.WithField("slot", slot).Warn("replication slot is already active (is another capture already running against this database?)")
 	} else if slotInfo.WALStatus == "lost" {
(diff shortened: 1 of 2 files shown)


source-postgres/replication.go
```diff
@@ -69,6 +69,8 @@ func (db *postgresDatabase) ReplicationStream(ctx context.Context, startCursor s
 		// completely unable to produce a fatal error, and double-checking to make extra sure we don't
 		// accidentally dereference a null pointer doesn't really hurt anything.
 		logrus.WithField("slot", slot).Warn("replication slot has no metadata (and probably doesn't exist?)")
+	} else if slotInfo.Active {
+		logrus.WithField("slot", slot).Warn("replication slot is already active (is another capture already running against this database?)")
```
jgraettinger, line 73: should we be error-ing out when this happens? A warning will be easy to ignore and I imagine this is going to give incorrect results.

willdonnelly: This whole section of logic is deliberately only logging warnings for now. Eventually all of these warnings will be upgraded to errors once I'm confident they're working correctly in production.

Also in the specific case of this "is the slot active?" check, we know that the attempt to actually start replication a few lines further down in this function will fail anyway.

jgraettinger (approved):
LGTM

PR #1708, opened 2024-07-09
materialize-motherduck: support for standard update materializations

williamhbaker:
**Description:**

Motherduck now supports multi-statement transactions, and is able to do bulk in-place updates using a combination of delete + insert operations, so we can do a standard updates materialization in that way.

Querying loaded documents was a little more difficult than normal here: The most obvious way of running the query with the SQL driver and iterating through the results apparently loads the entire query result into memory, which won't work for large load queries that would have results taking up a lot of memory. Saving the results to an S3 bucket does work though, and we can read the resulting rows efficiently from there.

Closes [#1707](https://github.com/estuary/connectors/issues/1707)

**Workflow steps:**

Update the `materialize-motherduck` docs, since it currently says we only support delta updates.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1708)


materialize-motherduck/driver.go
```diff
@@ -229,26 +234,155 @@ func newTransactor(
...
+	}
+
+	var subqueries []string
+	for _, b := range d.bindings {
+		if !b.loadFile.started {
+			// Pass.
+		} else if delete, err := b.loadFile.flush(); err != nil {
+			return fmt.Errorf("load.stage(): %w", err)
+		} else {
+			defer delete(ctx)
```
jgraettinger, line 268: this could defer multiple calls to `delete(ctx)`; is that intended?

jgraettinger (approved):
LGTM % question about `delete`

materialize-motherduck/driver.go
williamhbaker: It should be one `delete(ctx)` for every binding `b` where `b.loadFile.started` is `true`, so it will defer a call to delete the staged file(s) for every binding that had keys to load documents for.

jgraettinger: :man_facepalming: I see it now; `delete` is a closure returned by `flush()`.

Might be worth a comment simply because it tripped up my "avoid defer in for loops" pattern recognizer.

PR #1710, opened 2024-07-09
source-stripe-native: removing schema from pydantic models

Luishfs:
**Description:**

This PR removes data fields related to all pydantic models in source-stripe-native.
This means that all data models allow extra fields and contain only data related to the primary key and excluded fields

**Notes for reviewers:**

Tested using flowctl and a local flow instance, with a supabase DB as a destination

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1710)


jonwihl (approved):
LGTM

PR #1711, opened 2024-07-09
source-mongodb: request pre-images for change events if available

williamhbaker:
**Description:**

Recent versions of MongoDB make pre-images for update, replace, and delete events available if the MongoDB collections are configured for it. This is useful to capture if it is there, so we will capture these pre-images if they are available.

We'll probably need to implement support for splitting large events at some point, so we will only request these pre-images on versions of MongoDB that support `$changeStreamSplitLargeEvent`. I think we are safe to not do that right now, since it will be fairly complicated and it is not common for documents to be that large - and if there are hopefully they don't have pre-images enabled on their collections.

The existing unit and integration tests confirm the connector output is identical after this change for collections that don't have pre-images enabled, since the tests only create them without pre-images enabled. I manually tested scenarios of collections that do have pre-images enabled and verified that the `before` documents are captured for update, replace, and delete events.

**Workflow steps:**

`source-mongodb` documentation will need updated for this change.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1711)

### DIFF
--- a/source-mongodb/.snapshots/TestDiscover
+++ b/source-mongodb/.snapshots/TestDiscover
@@ -44,6 +44,11 @@ Binding 0:
               ],
               "title": "Change Operation",
               "description": "Change operation type: 'c' Create/Insert 'u' Update 'd' Delete."
+            },
+            "before": {
+              "type": "object",
+              "title": "Before Document",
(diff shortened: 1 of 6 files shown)


jgraettinger (approved):
LGTM :bowing_man:

PR #1714, opened 2024-07-11
source-shopify: ordering "Orders" stream and substreams

Luishfs:
"Orders" stream takes a long time to sync and ofter breaks. Because of that its state would be written and trigger the incremental step in the capture, instead of finishing backfill. Now that the stream is ordered and the incremental step was limited, the connector will only enter its "incremental" step after synced the whole stream since start_date

**Notes for reviewers:**

Tested using a local flow setup, with materialization to a supabase DB

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1714)

### DIFF
--- a/source-shopify/acmeCo/orders.schema.yaml
+++ b/source-shopify/acmeCo/orders.schema.yaml
@@ -864,6 +864,18 @@ properties:
           type:
             - string
             - "null"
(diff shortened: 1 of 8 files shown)


jonwihl (approved):
LGTM

PR #1717, opened 2024-07-12
source-klaviyo: changing Events stream API && new field

Luishfs:
Changed Events stream API to stable version
And added new field called campaign_id

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1717)


jonwihl (approved):
LGTM

PR #1725, opened 2024-07-16
materialize-snowflake: make account field optional

Alex-Bair:
**Description:**

Snowflake does not require an `account` query parameter in the connection string, so this change makes the `account` input optional. The `go-snowflake` driver's `sf.DSN` method for building the connection string can error out when `account` isn't provided, so instead we build the connection string ourselves.

**Workflow steps:**

The [Snowflake materialization docs](https://github.com/estuary/flow/blob/master/site/docs/reference/Connectors/materialization-connectors/Snowflake.md) will need updated to reflect that `account` is not required.

**Notes for reviewers:**

Tested locally with:
- both password & JWT authentication.
- both flavors of Snowflake `host` URL formats 
  - `<orgname>-<account_name>.snowflakecomputing.com`
  - `<accountlocator>.<region>.<cloud>.snowflakecomputing.com`

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1725)

### DIFF
--- a/materialize-snowflake/config.go
+++ b/materialize-snowflake/config.go
@@ -3,8 +3,10 @@ package main
 import (
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/pem"
 	"fmt"
+	"net/url"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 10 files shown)


materialize-snowflake/config.go
```diff
@@ -42,24 +40,63 @@ type advancedConfig struct {
 	UpdateDelay string `json:"updateDelay,omitempty" jsonschema:"title=Update Delay,description=Potentially reduce active warehouse time by increasing the delay between updates. Defaults to 30 minutes if unset.,enum=0s,enum=15m,enum=30m,enum=1h,enum=2h,enum=4h"`
 }
 
-// ToURI converts the Config to a DSN string.
-func (c *config) ToURI(tenant string) string {
-	// Build a DSN connection string.
-	var configCopy = c.asSnowflakeConfig(tenant)
+// toURI manually builds the DSN connection string
```
williamhbaker, line 43: Nit: Comments should usually be complete sentences, including punctuation at the end. That's clearly not a universally adhered to rule but I'd suggest working in that direction, which will require your new comments to have periods at the end.

materialize-snowflake/pipe.go
```diff
@@ -72,7 +72,11 @@ func generateJWTToken(key *rsa.PrivateKey, user string, accountName string) (str
 func NewPipeClient(cfg *config, accountName string, tenant string) (*PipeClient, error) {
 	httpClient := http.Client{}
 
-	var dsn = cfg.ToURI(tenant)
+	var dsn, err = cfg.toURI(tenant) // revisit
```
williamhbaker, line 75: Probably don't need this comment?

materialize-snowflake/config_test.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+	} {
+		t.Run(name, func(t *testing.T) {
+			require.NoError(t, cfg.Validate())
+			uri, err := cfg.toURI("mytenant")
+			require.NoError(t, err)
+			cupaloy.SnapshotT(t, uri)
+		})
+	}
+}
```
williamhbaker, line 48: Need a newline at the end of this file - `gofmt` should take care of this automatically for you. Setting up for editor to automatically run that on save would be a good idea.

materialize-snowflake/config.go
Alex-Bair: I made most of my new comments be complete sentences with punctuation.

materialize-snowflake/pipe.go
Alex-Bair: Good catch. I removed it.

Issue #1720, opened 2024-07-16
materialize-snowflake: use LTZ/TZ columns for datetime fields instead of NTZ


williamhbaker:
See https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1718736502716669

PR #1728, opened 2024-07-17
materialize-snowflake: use timezone containing columns for datetime fields

williamhbaker:
**Description:**

Closes https://github.com/estuary/connectors/issues/1720

Historically we have created columns for `format: datetime` fields as `TIMESTAMP`. There is no `TIMESTAMP` column in Snowflake, and this actually ends up being a `TIMESTAMP_LTZ`, `TIMESTAMP_TZ`, or `TIMESTAMP_NTZ` timestamp column depending on how the Snowflake database is configured. The default is to use `NTZ` which is just about the worst choice since it stores the timestamps without any time zone information, and without converting the timestamps to UTC.

With this change we will respect the database setting for timestamp column types if it is configured for `LTZ` or `TZ` columns, both of which will preserve time zone information. But if the database is set to use `NTZ` columns, we'll create the columns as `LTZ` instead since we want to be storing time zones.

There will be no behavior change to existing materialized tables, since we stage `datetime` fields to load as their timestamp string and that continues to work for any kind of `TIMESTAMP_*` column. The change will only apply for _new_ tables: Either new materializations/bindings, or (notably) re-backfills of existing tables that result in the tables being dropped & re-created. Because of this latter fact, we'll want to give users some advanced notice before this change goes live.

**Workflow steps:**

We'll want to update the snowflake docs for this change.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1728)


materialize-snowflake/snowflake.go
```diff
@@ -661,7 +704,7 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
 	// If we see no results from the REST API for `maxTries` iterations, then we
 	// fallback to asking the `COPY_HISTORY` table. We allow up to 5 minutes for results to show up in the REST API.
 	var maxTries = 40
-	var retryDelaySeconds = 3 * time.Second
+	var retryDelay = 3 * time.Second
```
williamhbaker, line 707: Taking care of a linter complaint

materialize-snowflake/.snapshots/TestValidateAndApply
```diff
@@ -89,7 +89,7 @@ Big Schema Changed Types Constraints:
 {"Field":"numField","Type":6,"TypeString":"UNSATISFIABLE","Reason":"Field 'numField' is already being materialized as endpoint type 'FLOAT' but endpoint type 'BOOLEAN' is required by its schema '{ type: [boolean] }'"}
 {"Field":"objField","Type":6,"TypeString":"UNSATISFIABLE","Reason":"Field 'objField' is already being materialized as endpoint type 'VARIANT' but endpoint type 'STRING' is required by its schema '{ type: [string] }'"}
 {"Field":"stringDateField","Type":6,"TypeString":"UNSATISFIABLE","Reason":"Field 'stringDateField' is already being materialized as endpoint type 'DATE' but endpoint type 'STRING' is required by its schema '{ type: [string] }'"}
-{"Field":"stringDateTimeField","Type":6,"TypeString":"UNSATISFIABLE","Reason":"Field 'stringDateTimeField' is already being materialized as endpoint type 'TIMESTAMP_NTZ' but endpoint type 'STRING' is required by its schema '{ type: [string] }'"}
+{"Field":"stringDateTimeField","Type":6,"TypeString":"UNSATISFIABLE","Reason":"Field 'stringDateTimeField' is already being materialized as endpoint type 'TIMESTAMP_LTZ' but endpoint type 'STRING' is required by its schema '{ type: [string] }'"}
```
williamhbaker, line 92: Note: This test is snapshotting the constraint that is produced when changing a `type: string, format: datetime` field to a `type: string` field without the `format: datetime`.

jgraettinger (approved):
LGTM

Any concern of LTZ cast being used on stored data within a merge query, that's loading into what's actually a NTZ column?

williamhbaker:
> Any concern of LTZ cast being used on stored data within a merge query, that's loading into what's actually a NTZ column?

Nothing comes to mind, since we will continue to serialize datetime fields using RFC3339Nano for storing data into columns for all `TIMESTAMP_*` columns. And the data that we are loading from snowflake for reductions comes from the `flow_document` which is a JSON stored in a variant column, and the [timestamps are preserved as strings there](https://docs.snowflake.com/en/user-guide/semistructured-considerations#storing-semi-structured-data-in-a-variant-column-vs-flattening-the-nested-structure).

PR #1729, opened 2024-07-17
source-mysql: Log 'SHOW VARIABLES' at `DEBUG` on startup

willdonnelly:
**Description:**

This is a fairly large log message, but our debug-level logs are already pretty spammy these days so this shouldn't be a big deal, and it's useful to be able to go and check the complete settings of any capture on demand.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1729)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -267,6 +267,11 @@ func (db *mysqlDatabase) connect(ctx context.Context) error {
 		return fmt.Errorf("unable to connect to database: %w", err)
 	}
 
+	// Debug logging hook so we can get the server config variables when needed
+	if err := db.logServerVariables(); err != nil {
+		logrus.WithField("err", err).Warn("failed to log server variables")
+	}
+
 	if db.config.Timezone != "" {
 		// The user-entered timezone value is verified to parse without error in (*Config).Validate,
 		// so this parsing is not expected to fail.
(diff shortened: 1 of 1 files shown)


source-mysql/main.go
```diff
@@ -314,6 +319,38 @@ func (db *mysqlDatabase) connect(ctx context.Context) error {
 	return nil
 }
 
+func (db *mysqlDatabase) logServerVariables(level logrus.Level) error {
+	if !logrus.IsLevelEnabled(level) {
```
williamhbaker, line 323: Having the log level as an argument here was a little confusing since we only ever call it with the debug level from line 271 and I was looking for other places it was called with different log levels and couldn't find any. Maybe we could hard-code the check here to return early on `!logrus.IsLevelEnabled(logrus.DebugLevel)` and make the logging on line 350 always at the `debug` level?

williamhbaker (approved):
LGTM, one non-blocking comment

source-mysql/main.go
willdonnelly: Hmm, good point. Previously I had the "is it enabled" check wrapped around the call to `logServerVariables` over in the connect function, and so the level was a parameter to avoid stating it in two separate places and getting them out of sync. But yeah, now that this check is also part of the function it's unnecessary.

PR #1733, opened 2024-07-18
materialize-elasticsearch: create the meta index with the server default number of shards

williamhbaker:
**Description:**

Rather than always create the meta index with 1 shard, create it with the server default number of shards by leaving the "shards" parameter unspecified.

We might not really need more than 1 shard for this index, but there are near-term plans to not create the meta index at all and instead use the "last spec" available in Validate and Apply requests. This change should also allow our connector to work with serverless Elasticsearch, which does not allow setting the number of shards or replicas for indices.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1733)


willdonnelly (approved):
LGTM

PR #1734, opened 2024-07-18
materialize-elasticsearch: don't waitForActiveShards when connecting to serverless clusters

williamhbaker:
**Description:**

This option isn't supported when connecting to serverless clusters (they don't have shards), so we need to detect when we are connecting to a serverless cluster and not set it if we are.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1734)

### DIFF
--- a/materialize-elasticsearch/client.go
+++ b/materialize-elasticsearch/client.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strings"
 
 	elasticsearch "github.com/elastic/go-elasticsearch/v8"
 	"github.com/elastic/go-elasticsearch/v8/esutil"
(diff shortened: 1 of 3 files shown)


materialize-elasticsearch/client.go
```diff
@@ -238,3 +239,27 @@ func (c *client) infoSchema(ctx context.Context) (*boilerplate.InfoSchema, error
 
 	return is, nil
 }
+
+func (c *client) serverless(ctx context.Context) (bool, error) {
```
willdonnelly, line 243: IIRC typical Go style is to name the method something like `isServerless`

willdonnelly (approved):
LGTM

PR #1735, opened 2024-07-18
source-postgres: Lower replication buffer from 16k to 16

willdonnelly:
**Description:**

I'm making an educated guess that this is responsible for a user's capture OOMing, and I suspect that 16k is much too large anyway.

As we know from queuing theory, in the steady state a buffer is always either empty or full. And I have a strong suspicion that this buffer is basically always emptyish in most production capture tasks and the only purpose it actually serves is helping smooth out the work passing so that both the replication thread and the main capture serialization thread can keep busy.

Based on that intuition, I figure that 16 should actually be plenty to keep the threads from blocking on each other most of the time, while being much less likely to cause OOMing if it does fill up temporarily.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1735)

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -211,10 +211,10 @@ const standbyStatusInterval = 10 * time.Second
 // In normal use it's a constant, it's just a variable so that tests are more
 // likely to exercise blocking sends and backpressure.
 //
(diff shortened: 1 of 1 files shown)


psFried (approved):
LGTM

PR #1736, opened 2024-07-18
source-mysql: Support partial row images properly

willdonnelly:
**Description:**

Makes partial row images work correctly by omitting the unspecified columns from the resulting change document rather than filling them out with implicit nulls.

Also adds a test verifying the correct behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1736)

### DIFF
--- a/source-mysql/.snapshots/TestPartialRowImages-delete
+++ b/source-mysql/.snapshots/TestPartialRowImages-delete
@@ -1,0 +1,9 @@
+# ================================
+# Collection "acmeCo/test/test_partialrowimages_16824726": 1 Documents
+# ================================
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

I'll just acknowledge that not emitting a full representation of the row may be a little unexpected, but it does seem like the best we can reasonably do, and it seems like it is pretty rare for a database to be configured in a way allowing these partial rows.

willdonnelly:
Yeah, the intent is that eventually we'll improve this further by having discovery put appropriate default-value information in the generated schemas. But for now omitting the properties is a pure improvement over including them with explicit null values.

PR #1737, opened 2024-07-18
source-postgres: Try lowering replication buffer size to one event

willdonnelly:
**Description:**

Followup to https://github.com/estuary/connectors/pull/1735

Just to make absolutely sure that this buffer isn't the issue here. If this doesn't fix the OOMs we're seeing in production, we should revert it and figure out what else is going wrong.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1737)


willdonnelly:
Merging without review. This shouldn't be able to break anything (other than indirectly by theoretically lowering replication throughput) and I want to see if this helps ASAP and then revert it if it doesn't.

PR #1741, opened 2024-07-19
source-mysql: Simplify secondary index discovery

willdonnelly:
**Description:**

Previously the secondary-index discovery query had a bug in its join against `information_schema.table_constraints` which would crop up if there were multiple constraints on the same table, with the same constraint name, but different constraint types. For instance, a `UNIQUE` constraint and a `FOREIGN KEY` one.

But after rereading the relevant discovery query I'm not sure why we're even looking at that table in the first place. All we used that join for was a `tco.constraint_type != 'PRIMARY KEY'` filter clause, but as we can see in the primary-key discovery query we can also just check whether `stat.index_name = 'PRIMARY'` so the join should be entirely unnecessary.

This commit also adds a subtest to `TestSecondaryIndexDiscovery` which reproduces the scenario that previously failed. After the query fix in this commit the new subtest also passes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1741)

### DIFF
--- a/source-mysql/.snapshots/TestSecondaryIndexDiscovery-index_and_fk
+++ b/source-mysql/.snapshots/TestSecondaryIndexDiscovery-index_and_fk
@@ -1,0 +1,130 @@
+Binding 0:
+{
+    "recommended_name": "test_secondaryindexdiscovery_index_and_fk_g16025",
+    "resource_config_json": {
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

Issue #1724, opened 2024-07-16
source-hubspot-native: automatically use search API if legacy API limits are reached


williamhbaker:
See https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1721141812973129

PR #1743, opened 2024-07-22
source-hubspot-native: search API fallback for certain bindings

williamhbaker:
**Description:**

Closes https://github.com/estuary/connectors/issues/1724

The main change here is adding the ability for the `deals` and `companies` bindings to fall back to using the [HubSpot search API](https://developers.hubspot.com/docs/api/crm/search) when there are more than 10,000 changed records to fetch. The "legacy" APIs we use for listing changes of these objects do not let you page back further than 10,000 items, so this is necessary to avoid needing to completely re-backfill the binding.

One assumption here is that by using the legacy API to get the ~10,000 records that we can, we won't run into issues with the Search API returning bogus results for very "recent" data, as it is known to do. For example, if a source HubSpot account has 20,000 updates to `deals` all right about at the same time, we'll see that happen right away with the legacy API and then after having paged through ~10,000 records with the legacy API, start using the search API. Theoretically this could be too recent for the search API to produce consistent results and we might miss data. I am assuming that the time it takes to get 10,000 records from the legacy API and the inner workings of HubSpot _will_ produce consistent results when using the Search API in this way, but it is an area of uncertainty and would be the first thing I'd suspect if missing data is reported for these cases. Hopefully it all just works though.

Another adjustment is to make the `email_events` resource have a time horizon considerably longer than 5 minutes, due to observations that a 5 minute horizon will still end up missing data to API inconsistencies. I was able to reproduce these inconsistencies locally with 5 minutes, and tested that a 1 hour horizon did not produce them. So we'll go with 1 hour to not miss any data.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1743)

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api.py
+++ b/source-hubspot-native/source_hubspot_native/api.py
@@ -1,4 +1,5 @@
 import asyncio
+import functools
 import itertools
 import json
 from datetime import UTC, datetime, timedelta
@@ -235,6 +236,7 @@ async def fetch_changes(
     fetch_recent: FetchRecentFn,
     http: HTTPSession,
     object_name: str,
+    fallback_to_search_api: bool,
     # Remainder is common.FetchChangesFn:
     log: Logger,
     log_cursor: LogCursor,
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

PR #1751, opened 2024-07-23
source-dropbox: new connector

Alex-Bair:
**Description:**

This is a relatively straightforward, filesource-type connector using the ["unofficial" Dropbox Go SDK](https://github.com/dropbox/dropbox-sdk-go-unofficial). A Dropbox App must be setup & configured for OAuth2 to work.

**Workflow steps:**

Documentation will need created. I'm working on creating that documentation, and I'll add a reference to that PR here when it's ready.

**Notes for reviewers:**

Tested locally with with a large number of JSON, CSV, and nested directories. To get OAuth to work locally, I set up my own Dropbox App per this [guide](https://developers.dropbox.com/oauth-guide), made sure to select the `files.metadata.read` and `files.content.read` permissions, & added the app key and secret to my local database.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1751)


go.mod
```diff
@@ -121,6 +121,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dgraph-io/ristretto v0.1.1 // indirect
 	github.com/dnephin/pflag v1.0.7 // indirect
+	github.com/dropbox/dropbox-sdk-go-unofficial/v6 v6.0.5 // indirect
```
williamhbaker, line 124: You should run `go mod tidy`, since this import shouldn't be `indirect`, as you are using it directly.

source-dropbox/main.go
```diff
@@ -0,0 +1,300 @@
+package main
+
+import (
+	"cmp"
+	"context"
+	_ "embed"
```
williamhbaker, line 6: I don't think this is needed anywhere

source-dropbox/main.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/dropbox/dropbox-sdk-go-unofficial/v6/dropbox"
+	"github.com/dropbox/dropbox-sdk-go-unofficial/v6/dropbox/files"
+	"github.com/estuary/connectors/filesource"
+	"github.com/estuary/flow/go/parser"
+	pf "github.com/estuary/flow/go/protocols/flow"
+)
+
+type config struct {
+	Path        string         `json:"path"`
+	Credentials *Credentials   `json:"credentials"`
```
williamhbaker, line 22: Credentials here doesn't need to be a pointer. Parser doesn't really either but I know we do that elsewhere so that's fine, and I guess it makes the `ParserConfig()` method work arguably a little cleaner.

source-dropbox/main.go
```diff
@@ -0,0 +1,300 @@
...
+func (c config) Validate() error {
+	requiredProperties := [][]string{
+		{"Path", c.Path},
+	}
+	for _, req := range requiredProperties {
+		if req[1] == "" {
+			return fmt.Errorf("missing '%s'", req[0])
+		}
+	}
+	return nil
```
williamhbaker, line 41: We should validate `Credentials` somehow here too. It has a `Validate()` already that isn't being called anywhere so you could call that here?

source-dropbox/main.go
```diff
@@ -0,0 +1,300 @@
...
+	Advanced    advancedConfig `json:"advanced"`
+}
+
+type advancedConfig struct {
+	AscendingKeys bool `json:"ascendingKeys,omitempty"`
+}
+
+func (c config) Validate() error {
+	requiredProperties := [][]string{
+		{"Path", c.Path},
```
williamhbaker, line 34: Since `Path` needs to start with a `/` it would be good to check that here.

source-dropbox/main.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	return nil
+}
+
+func (c config) DiscoverRoot() string {
+	return c.Path
+}
+
+func (c config) RecommendedName() string {
+	return c.Path[1:]
```
williamhbaker, line 49: I think it would be more clear to use `strings.Trim` here, since I believe the intent is to return the path with the leading slash removed.

On most other filesource captures we are using `strings.Trim(c.DiscoverRoot(), "/")`.

source-dropbox/main.go
```diff
@@ -0,0 +1,300 @@
...
+
+func (c config) PathRegex() string {
+	return c.MatchKeys
+}
+
+func newDropboxStore(ctx context.Context, cfg config) (*dropboxStore, error) {
+	oauthClient := cfg.Credentials.GetOauthClient(ctx)
+	config := dropbox.Config{
+		Client:   oauthClient,
+		LogLevel: dropbox.LogDebug,
```
williamhbaker, line 68: What kind of logs does this produce? Are they useful? Having debug logging enabled seems a little surprising. Without actually looking at what kind of logging we are getting, I would suggest using `LogOff` if the configured connector log level is `Info` or "higher", and only set this to `LogDebug` if the connector log level is `Debug`. Usually we want to have more direct control over logs since they are user-facing, and the SDK logs can be confusing for users.

source-dropbox/main.go
```diff
@@ -0,0 +1,300 @@
...
+	// Get all files at the specified path from Dropbox.
+	arg := &files.ListFolderArg{Path: query.Prefix, Recursive: query.Recursive}
+	f, err := dbStore.client.ListFolder(arg)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving folder: %w", err)
+	}
+
+	entries := f.Entries
+
+	for f.HasMore {
```
williamhbaker, line 110: A comment here about why we need to do a full listing and sort in-memory would be helpful

source-dropbox/main.go
```diff
@@ -0,0 +1,300 @@
...
+		arg := files.NewListFolderContinueArg(f.Cursor)
+
```
williamhbaker, line 112: Nit: There is a newline between `arg := ...` and `f, err =` here but not on line 102

source-dropbox/main.go
```diff
@@ -0,0 +1,300 @@
...
+				objects = append(objects, obj)
+			}
+
+		default:
+			// Ignore all other metadata types (like DeletedMetadata).
+			continue
+		}
+	}
+
+	// Remove files that are lexicographically less than the query's StartAt property.
```
williamhbaker, line 150: We could do the conversion and filtering in a single pass, something like this:

```go
	for _, entry := range entries {
		var obj filesource.ObjectInfo

		switch e := entry.(type) {
		case *files.FileMetadata:
			obj = filesource.ObjectInfo{
				Path:       e.PathLower,
				IsPrefix:   false,
				ContentSum: e.ContentHash,
				Size:       int64(e.Size),
				ModTime:    e.ServerModified,
			}

		case *files.FolderMetadata:
			if query.Recursive {
				continue
			}

			obj = filesource.ObjectInfo{
				Path:     e.PathLower,
				IsPrefix: true,
			}

		default:
			// Ignore all other metadata types (like DeletedMetadata).
			continue
		}

		if query.StartAt < obj.Path {
			// Remove files that are lexicographically less than the query's StartAt property.
			objects = append(objects, obj)
		}
	}
```

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,69 @@
...
+	return &spec
+}
+
+type Credentials struct {
+	ClientID     string `json:"client_id"`
+	ClientSecret string `json:"client_secret"`
+	RefreshToken string `json:"refresh_token"`
+}
+
+func (c *Credentials) Validate() error {
```
williamhbaker, line 41: I don't think this method needs to be exported

source-dropbox/oauth2.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	for _, req := range requiredProperties {
+		if req[1] == "" {
+			return fmt.Errorf("missing '%s'", req[0])
+		}
+	}
+	return nil
+}
+
+func (c *Credentials) GetOauthClient(ctx context.Context) *http.Client {
```
williamhbaker, line 55: This one also doesn't need to be exported

williamhbaker (commented):
Looking good - made a few small comments.

source-dropbox/main.go
Alex-Bair: Agreed, I wasn't sure if combining them would make it a little harder to follow the code. But looking at it, what's happening is still pretty clear. I'll make that change.

williamhbaker:
We'll also need to add the connector to the CI process so it will actually be built and pushed, see https://github.com/estuary/connectors/commit/49914085bdaad47c68f03e2612fbd6c2dc5ceb1a for an example

source-dropbox/main.go
Alex-Bair: Nice catch, I'll remove the newline here.

source-dropbox/main.go
Alex-Bair: Agreed, I"ll remove it.

source-dropbox/oauth2.go
Alex-Bair: Wait, I got mixed up with the `Validate` in `main.go`. Yeah, I agree I don't think this needs exported either.

williamhbaker: `config.Validate(&` does need to be exported, but I don't think Credentials.Validate()` does. And now that I'm looking at it, the `Credentials` struct also does not need to be exported.

source-dropbox/main.go
Alex-Bair: I agree, I'll change it to use `strings.Trim`.

source-dropbox/main.go
Alex-Bair: I'll add a check using `strings.HasPrefix`, similar to how it's done in the [sftp source connector](https://github.com/estuary/connectors/blob/ad38756f847ba5637d122271a0470276469a80ec/source-sftp/main.go#L103-L105).

source-dropbox/main.go
Alex-Bair: Good idea. What do you think about adding this comment? I could put it as the first comment within the `List` method.
> The Dropbox API returns pages of entries sorted by their modification date and time. However for the connector to work with the `filesource` package correctly, the listing returned by `List` must contain all entries sorted in lexicographic, monotonic order. To achieve this, we paginate through all entries, convert them to `filesource.ObjectInfo`s, and sort them based on their path.

williamhbaker: Yeah that seems good. The bit about how the API returns entries sorted differently is great.

source-dropbox/main.go
Alex-Bair: I actually haven't been able to find any logs produced by the SDK. I've tried using both `LogDebug` and `LogInfo` as the `LogLevel`, and providing a `Logger` to the `dropbox.Config`, but I'm not seeing any SDK generated logs show up. I'm leaning toward removing that line since it's not adding any value.

source-dropbox/main.go
Alex-Bair: I'll add that comment. I'll also clarify the comments for the pagination since it's not obvious that the Dropbox API's `ListFolderContinue` fetches the next page of entries.

source-dropbox/main.go
williamhbaker: Yeah that sounds good

Alex-Bair:
Thank you for the review & comments! I believe I've addressed them all. Also, in `.github/workflows/ci.yaml`, I only added to the `build_connectors` section since this connector doesn't have any integration tests (like the SFTP one does).

source-dropbox/main.go
```diff
@@ -0,0 +1,308 @@
...
+
+		default:
+			// Ignore all other metadata types (like DeletedMetadata).
+			continue
+		}
+
+		// Only add files and folders that are lexicographically 
+		// greater than or equal to the query's StartAt property.
+		if query.StartAt < obj.Path {
+			objects = append(objects, obj)
```
williamhbaker, line 166: What happens in the case of `*files.FolderMetadata` and when `query.Recursive` is `true`? I guess `obj.Path` would be `""` and so the empty object wouldn't be appended?

Alex-Bair: That's what I'm thinking & seeing when I test it out. But that's not super clear, so what do you think about re-writing the `*files.FolderMetadata` case like below?
```golang
case *files.FolderMetadata:
	if query.Recursive {
		continue
	}
	obj = filesource.ObjectInfo{
		Path:     e.PathLower,
		IsPrefix: true,
	}
```

Then we wouldn't reach `if query.StartAt < obj.Path` when `query.Recursive` is `true`.

williamhbaker: That seems better!

Alex-Bair: Updated!

williamhbaker (approved):
LGTM

PR #1753, opened 2024-07-25
source-postgres: Add 'MinimumBackfillXID' advanced option

willdonnelly:
**Description:**

This is completely untested except to make sure it doesn't break anything else in our automated test suite, but the XMIN filtering logic is taken from our batch Postgres connector and this is a pretty straightforward bit of plumbing directly from the advanced connector config into the relevant backfill query generation, so it'll either work or it won't.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1753)


jgraettinger (approved):
LGTM

PR #1755, opened 2024-07-25
source-postgres: Bump pgx to v5 and fix JSON column unmarshalling

willdonnelly:
**Description:**

The original motivation for all this work was fixing JSON column unmarshalling. Previously `source-postgres` would receive values of `JSON` or `JSONB` columns, parse those into a `map[string]any` immediately upon receipt, and then serialize them back into JSON as part of the value translation logic. This is wasteful and unnecessary and we believe that even relatively modest JSON columns (in the tens of MiB range) can cause the connector to OOM.

Fixing that, however, required hooking into the pgx client library's value decoding process a bit more deeply than we previously did. And it seemed a little bit wasteful to do that while we were still using the v4 release of pgx, since that version is a couple of years old by now and there's been a total overhaul of the entire value encoding/decoding process in v5.

So this PR is mostly a version bump so that `source-postgres` will use PGX v5 instead, and then almost as an afterthought I've tacked on the JSON unmarshalling fix. It's in the same PR as the version bump because if it turns out I missed something and the version bump breaks stuff and has to be reverted, there's no way the JSON fix would keep working so we'd have to revert it as well anyway, so we might as well just keep all of that in the same PR for simplicity.

**Notes for reviewers:**

For the most part I've tried to ensure that nothing about the connector's observable behavior changed as a result of the package version bump, however that was more complicated than it sounds because the all-new value decoding stuff in PGX v5 will in many cases produce values of a different datatype entirely for the same data coming from the DB.

You can see for yourself that most of the test cases in `datatype_test.go` are unchanged, meaning that the ultimate output of our connector for a given input to Postgres is unchanged (for the cases we test, at least). For those that have changed, I've tried to summarize how they're different and why that's okay here:

* The stringification of `decimal` and `numeric` values now produces a string like `"123.456"` where before it would be `"123456e-3"`. This should be fine, as the value is still correct and should still satisfy the string-format-number validation.
* The translation of `timestamp with time zone` values has been fixed (using some of the new datatype customization plumbing which was the whole goal of migrating to v5) so that they are consistently written in the server time zone.
  * Previously the values were correct, but the time zone in which they were reported would vary between backfills and replication, because backfills would receive a Unix microsecond timestamp in UTC and then represent that in the local time zone of the connector, whereas the replicated values would arrive as text strings in the server's time zone and ultimately we would serialize them in that same time zone.
  * In retrospect we could probably have addressed this by simply having them always be printed out in UTC, but having them be in the server's configured time zone is probably even better for most users.
* The stringification of `interval` values now omits the `.000000` suffix when the interval is a whole number of seconds. We have no string-format validation for this type, and the change seems small enough to just allow it.
* The stringification of `lseg` values is now wrapped in square brackets, which seems like probably a bugfix given that now it matches how we represent the test value for input. We have no string-format validation for this type, and the change seems small enough to just allow it.
* The handling of some date/timestamp array/range types has been fixed so that the individual values are now stringified into RFC3339 values which are consistent with the handling of scalar equivalents.
  * In the case of `date array` values the old way couldn't have possibly worked as it didn't satisfy our own JSON schema for such columns.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1755)


source-postgres/datatype_test.go
```diff
@@ -64,25 +57,30 @@ func TestDatatypes(t *testing.T) {
...
 		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06 PST'`, ExpectValue: `"04:05:06-08:00"`},
 		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06 UTC'`, ExpectValue: `"04:05:06Z"`},
 		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06.123 UTC'`, ExpectValue: `"04:05:06.123Z"`},
 		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06.123+0330'`, ExpectValue: `"04:05:06.123+03:30"`},
 		{ColumnType: `time with time zone`, ExpectType: `{"type":["string","null"],"format":"time"}`, InputValue: `'04:05:06.123+03:30:50'`, ExpectValue: `"04:05:06.123+03:30:50"`},
-		{ColumnType: `interval`, ExpectType: `{"type":["string","null"]}`, InputValue: `'2 months 1 day 5 minutes 6 seconds'`, ExpectValue: `"2 mon 1 day 00:05:06.000000"`},
+		{ColumnType: `interval`, ExpectType: `{"type":["string","null"]}`, InputValue: `'2 months 1 day 5 minutes 6 seconds'`, ExpectValue: `"2 mon 1 day 00:05:06"`},
```
williamhbaker, line 69: Ideally maybe might turn these into an ISO8601 duration like [these](https://github.com/estuary/flow/blob/3aee84e71909330af093f68d2e739ffe22449529/crates/json/src/schema/formats.rs#L251-L254), though probably not worth the effort / breaking change right now.

williamhbaker (approved):
LGTM

source-postgres/datatype_test.go
willdonnelly: Yeah, that would probably be a good move in the future if/when we ever break compatibility for some of these value translations, but unless we can verify that nobody's currently using them and relying on this behavior we probably shouldn't do it now.

It looks like ISO8601 durations support fractional seconds with a decimal point (and so does our validation regexp) so it ought to work at least as well as the current behavior.

PR #1757, opened 2024-07-26
source-mysql: Truncate strings (and JSON) over 8MiB

willdonnelly:
**Description:**

Pretty much just copy-pasted from the Postgres equivalents, but the logic is pretty straightforward so there shouldn't be much that can go wrong here.

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -18,6 +18,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	truncateColumnThreshold = 8 * 1024 * 1024 // Arbitrarily selected value
+)
+
 func (db *mysqlDatabase) DiscoverTables(ctx context.Context) (map[string]*sqlcapture.DiscoveryInfo, error) {
 	var tableMap = make(map[string]*sqlcapture.DiscoveryInfo)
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

source-mysql/discovery.go
```diff
@@ -350,11 +363,18 @@ func (db *mysqlDatabase) translateRecordField(columnType interface{}, val interf
 				return t.UTC().Format(time.RFC3339Nano), nil
 			}
 		}
+		if len(val) > truncateColumnThreshold {
```
jgraettinger, line 366: under what conditions does this fire -- what does it mean for columnType to not be a `string`?

jgraettinger (approved):
LGTM % my question below

source-mysql/discovery.go
willdonnelly: The logic could definitely be made a bit more clear here, it just kind of grew like this over time. The `columnType` should always be a string at this point (because the only non-string types are handled up at the top of this function where we delegate to a separate `columnType.translateRecordField(val)` method for enums/sets/integers).

This return here is the default case for when the `switch typeName` doesn't have a special case for a particular column type.

PR #1759, opened 2024-07-26
source-postgres: Fix empty array serialization after PGX v5

willdonnelly:
**Description:**

The package upgrade to PGX v5 caused a lot of churn around how we translate various datatypes, and the array handling had to be rewritten entirely. Looks like I missed an edge case because the dimensions list is serialized as `null` when it should be `[]` for an empty array coming from the DB.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1759)


willdonnelly:
Merging unreviewed because this is a trivial 2-line PR (of which one line is a new test case verifying that the fix works as expected) and I'd like to get the production capture that hit this fixed ASAP

jgraettinger (commented):
LGTM

PR #1760, opened 2024-07-26
source-mongodb: split large change stream event documents v2

williamhbaker:
**Description:**

This is a modification to the way we check for pre-image support, and a re-application of the commit https://github.com/estuary/connectors/commit/5f66809a67e39f53a73030069fd081e4382f3e5a where the `$changeStreamSplitLargeEvent` pipeline stage was added.

The problem from that initial attempt was that the split event pipeline stage is not supported on somewhat-commonly used versions of MongoDB Atlas. Unfortunately there's no way to know if we are connected to one of those or not without just trying to create a change stream with that pipeline stage enabled.

The first commit here updates the check for pre-image support (which by extension is checking for support of `$changeStreamSplitLargeEvent`) to verify that the split event stage works by trying to create a change stream with it enabled and examining the error if there is one. The second commit is the re-application of https://github.com/estuary/connectors/commit/5f66809a67e39f53a73030069fd081e4382f3e5a, which should now work without causing existing tasks that don't support the split event stage to fail.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1760)

### DIFF
--- a/source-mongodb/.snapshots/TestCaptureSplitLargeDocuments
+++ b/source-mongodb/.snapshots/TestCaptureSplitLargeDocuments
@@ -1,0 +1,14 @@
+# ================================
+# Collection "acmeCo/test/testCollection": 6 Documents
+# ================================
+{"_id":"smallDocument","_meta":{"op":"c"},"baz":"qux","foo":"bar"}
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM

PR #1761, opened 2024-07-27
source-klaviyo: capturing campaign_id via  message field

Luishfs:
**Description:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1761)

### DIFF
--- a/source-klaviyo/source_klaviyo/streams.py
+++ b/source-klaviyo/source_klaviyo/streams.py
@@ -350,7 +350,7 @@ class Events(IncrementalKlaviyoStream):
         super().__init__(**kwargs)
 
         self.campaign_data = None
-        self.record_amount = 500000 # bypass failing snapshot tests, since current credentials break _prepare_campaign
+        self.record_amount = 100000 # bypass failing snapshot tests, since current credentials break _prepare_campaign
 
     def path(self, **kwargs) -> str:
         return "events"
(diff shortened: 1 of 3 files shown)


source-klaviyo/source_klaviyo/streams.py
```diff
@@ -345,30 +345,61 @@ class Events(IncrementalKlaviyoStream):
...
-
             record['datetime'] = record['attributes']['datetime'].replace(" ","T")
             record['attributes']['datetime'] = record['attributes']['datetime'].replace(" ","T")
-            record["attributes"]['event_properties'] = None
+
+            if self.campaign_data is not None:
+                if type(record["attributes"]['event_properties'].get("$message")) is str and record["attributes"]['event_properties']["$message"] in self.campaign_data:
+                    record["campaign_id"] = record["attributes"]['event_properties']["$message"]
+            else:
+                if type(record["attributes"]['event_properties'].get("$message")) is str and len(record["attributes"]['event_properties']["$message"]) >= 20:
```
jonwihl, line 398: This second check seems redundant. All campaign_ids in the >2023 format are going to be found in the first check. I also don't think this 'else' block will ever be reached since we only move into this block when campaign_data == None

Luishfs: The idea behind this if statement is in the case some error happens when capturing campaign data
We are doing relatively simple requests to klaviyo's API, but in case something goes 500 for some reason, the users capture wouldn't break, and the else statement would make sure we are at least getting campaign_id values for >= 2023. Not ideal, but at least user data or incremental run wouldnt be that much affected

Luishfs: @jonwihl Would you rather have the capture stop in case we dont get campaign_id data?

Luishfs:
@jonwihl Just made all the changes and validated them

jonwihl (approved):
LGTM

PR #1767, opened 2024-07-31
source-postgres: LSN commit logging tweaks

willdonnelly:
**Description:**

Hopefully this will be enough to show us what, if anything, is going wrong in rare cases to cause the 'confirmed_flush_lsn is greater than resume LSN' warning to fire.

The most important thing here is the introduction of a new INFO level debug message which will output the confirmed LSN immediately before we send it to the server in a "Standby Status Update", if it's different from the previous update. Since we send these updates every 10 seconds this could be moderately spammy, producing up to 6 log messages each minute where previously our steady-state streaming behavior was to emit 2 log messages each minute, a 4x increase. But I think it's worth doing that temporarily across the board so we can actually be 100% certain, the next time we see that warning trigger, whether we had actually sent a confirmation with that value. Once this debugging is all done this message will probably be lowered back to DEBUG level.

There are also a couple of other tweaks to the DEBUG level logging which should make it a bit easier to see what the actual Flow acknowledgement pattern looks like if I dig into a specific task more deeply, but I think that would be much too spammy to have on by default.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1767)

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -74,13 +74,13 @@ func (db *postgresDatabase) ReplicationStream(ctx context.Context, startCursor s
 	} else if startLSN < slotInfo.ConfirmedFlushLSN {
 		logrus.WithFields(logrus.Fields{
 			"slot":         slot,
-			"confirmedLSN": slotInfo.ConfirmedFlushLSN,
-			"startLSN":     startLSN,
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #1768, opened 2024-08-01
sqlcapture: Add source column type information to JSON schemas

willdonnelly:
**Description:**

This commit modifies the generic SQL CDC discovery logic so that it will include type information like `(source type: int)` or `(source type: non-nullable enum)` in the description of every column property in the discovered JSON schemas.

In the event that there's already a description, for example in the case of Postgres discovery with comments, the source type info will be appended after that comment.

The goal of this change is to make it much easier to see exactly what sort of source column types we're actually dealing with, as the JSON schema types are a more restricted set and often there may be multiple database types mapped to identical JSON schema types. Currently to see this information we have to go use a low-level `flowctl-go shards edit` command to toggle debug logging on for a capture, search through the massive volume of debug output for the relevant column of the correct table to see what it was, and then go toggle the log level back to `INFO`. This is a pain and I'm also not a big fan of having to restart a task just to see what column types it's handling.

**Workflow steps:**

When looking at a collection in the UI, the "code view" of the collection spec should now tell us what source database type corresponds to each property.

**Notes for reviewers:**

This is in some sense a fairly trivial change, but it produces a pretty large diff since it impacts every property of every discovery snapshot. I also had to refactor a bit of the Postgres discovery logic so that this would work nicely with enums and other user-defined data types, and in practice the biggest risk of this change is if I could have made a mistake there which wasn't caught by our tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1768)

### DIFF
--- a/source-postgres/discovery.go
+++ b/source-postgres/discovery.go
@@ -452,7 +452,7 @@ type postgresCompositeType struct{}
 
 func (t postgresCompositeType) String() string { return "composite" }
 func (t postgresCompositeType) toColumnSchema(_ sqlcapture.ColumnInfo) (columnSchema, error) {
-	return columnSchema{jsonType: "string"}, nil
+	return columnSchema{}, nil
 }
 
 // Query copied from pgjdbc's method PgDatabaseMetaData.getPrimaryKeys() with
--- a/source-postgres/discovery.go
+++ b/source-postgres/discovery.go
@@ -197,8 +197,12 @@ func (db *postgresDatabase) TranslateDBToJSONType(column sqlcapture.ColumnInfo) 
 		if !ok {
 			return nil, fmt.Errorf("unable to translate PostgreSQL type %q into JSON schema", columnType)
 		}
-	} else if colSchema, ok = column.DataType.(columnSchema); ok {
-		// Nothing else to do since the columnSchema was already determined.
+	} else if dataType, ok := column.DataType.(postgresComplexType); ok {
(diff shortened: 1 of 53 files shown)


source-postgres/discovery.go
```diff
@@ -401,21 +405,56 @@ func getColumns(ctx context.Context, conn *pgx.Conn) ([]sqlcapture.ColumnInfo, e
...
+
+func (t postgresRangeType) toColumnSchema(_ sqlcapture.ColumnInfo) (columnSchema, error) {
+	return columnSchema{jsonType: "string"}, nil
+}
+
+type postgresCompositeType struct{}
+
+func (t postgresCompositeType) String() string { return "composite" }
+func (t postgresCompositeType) toColumnSchema(_ sqlcapture.ColumnInfo) (columnSchema, error) {
+	return columnSchema{jsonType: "string"}, nil
```
williamhbaker, line 455: Do these always come through as a string now? Just double-checking since we used to use the catch-all schema for them by the looks of it.

williamhbaker (approved):
LGTM % one question

source-postgres/discovery.go
willdonnelly: Good catch, I didn't mean to make the type any narrower than it was before here.

willdonnelly: Should be fixed by the commit I just added

PR #1770, opened 2024-08-01
source-mysql: Ignore `CREATE/ALTER/DROP EVENT` queries

willdonnelly:
**Description:**

Ignores `CREATE EVENT`, `ALTER EVENT`, and `DROP EVENT` queries encountered in the binlog. We don't need to care about any of those and the Vitess SQL parser doesn't recognize those statements.


mdibaiee (approved):

williamhbaker (approved):
LGTM

PR #1771, opened 2024-08-01
source-google-pubsub: new connector

williamhbaker:
**Description:**

Creates a new capture connector to capture documents from Google PubSub topics.

This connector works more like a webhook capture than a typical capture that keeps track of the messages it has read via checkpoints. This is because of how Google PubSub works: A request to receive messages does not include any kind of offset or indication of where the recipient wants to resume from. Once the receipt channel is opened, PubSub starts sending any unacknowledged messages to the recipient, which must then process them and acknowledge their receipt. Among other things, this means that there is no "backfill" mechanism for the connector, or any kind of state tracking at all.

Discovery works by finding all of the available PubSub topics and representing them as bindings. For all enabled bindings, a subscription with a deterministic but unique name is created, and that subscription is used to receive messages.

Managing Flow runtime acknowledgements and coordinating acknowledgements of messages to PubSub is probably the most interesting part of the connector code. In short, documents to emit to the runtime are sequenced with signals that the runtime has acknowledge the checkpoints associated with those documents, and those signals are used to send the acknowledgement of the message to PubSub.

With this initial implementation only messages that are encoded as JSON are supported, which should work well for BigQuery continuous queries.

**Workflow steps:**

Create and use the connector to capture from PubSub topics.

New docs will need to be created.

**Notes for reviewers:**

I ran various manual tests with the capture to make sure it worked, with lots of topics and large documents produced at a high rate. I haven't included an integration test with this PR in the interest of time, but may add it later or as a follow-up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1771)


mdibaiee (approved):
LGTM!

PR #1776, opened 2024-08-02
source-stripe-native: adding a pre-process step for 'discounts' fields

Luishfs:
**Description:**

The idea behind this PR is to add a pre-process step to both `InvoiceItems` and `SubscriptionItems`  records.
Stripe API seems to deliver them in a variety of types, and for some reason that broke down the schema inference from the connector. Untill the cause is verified, this should allow the connector to run

**Notes for reviewers:**

Tested end-to-end on a local flow setup. Also, the write-schema was modified to allow for the `discounts` field.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1776)

### DIFF
--- a/source-stripe-native/acmeCo/InvoiceItems.schema.yaml
+++ b/source-stripe-native/acmeCo/InvoiceItems.schema.yaml
@@ -30,11 +30,19 @@ properties:
   id:
     title: Id
     type: string
+  discounts:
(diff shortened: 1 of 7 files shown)


Luishfs:
This issue was not replicated, and expressive connector changes make this PR useless, closing....

PR #1777, opened 2024-08-02
source-sqlserver: Fix CI builds consistently failing

willdonnelly:
**Description:**

Apparently in late July, Microsoft pushed an update to the `2022-latest` Docker image we use for SQL Server in CI builds. (I didn't notice this locally because how often does one run a `docker pull` for a relatively stable test dependency, but for the past couple of weeks all SQL Server CI builds were failing.)

The update contained two notable changes:
- The directory `/opt/mssql-tools` which contains the `sqlcmd` binary was renamed to `/opt/mssql-tools18`
- The `sqlcmd` binary was upgraded and the new version defaults to stricter certificate validation that a self-hosted server in a CI build fails.

The fix was very simple, just changing the path to the binary and adding a `-C` flag which means "trust the server certificate". I'm also leaving in place a couple of the tweaks I had to make as part of figuring out what the issue was, because I think they're useful changes for CI build debuggability in the future.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1777)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -181,6 +181,7 @@ jobs:
             "source-mongodb"
             ]'), matrix.connector)
         run: |
+          set +e # Show logs even if startup fails
           docker compose --file ${{ matrix.connector }}/docker-compose.yaml up --wait
           docker logs ${{ matrix.connector }}-db-1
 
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

williamhbaker:
Although it looks like the sql server tests are failing for a different reason now?

willdonnelly:
It's the same reason, I just debugged it and fixed it for the database setup script and then forgot to apply the same fix to the integration test script.

PR #1779, opened 2024-08-05
source-zendesk-support: update to Estuary CDK

Alex-Bair:
**Description:**

Imports `source-zendesk-support` to the Estuary CDK.

_Current ATF connector outputs vs. new imported connector outputs_:

- **Spec snapshot changes**: Just the expected resource config changes seen for all of these connectors. See [jsondiff](https://jsondiff.com/?left=https://raw.githubusercontent.com/estuary/connectors/513fae26b52107ffbca4e1898cffc4d54e99a188/source-zendesk-support/tests/snapshots/snapshots__spec__capture.stdout.json&right=https://raw.githubusercontent.com/estuary/connectors/48af972315fafe2241c43dc73651493d657a02ab/source-zendesk-support/tests/snapshots/snapshots__spec__capture.stdout.json).

- **Discovery snapshot changes**: Primary changes are the expected addition of `x-infer-schema` and `_meta` properties for each binding. For a few full refresh streams, the `cursorField` was removed since it isn't used when doing a full refresh. A few [previously disabled streams](https://github.com/estuary/airbyte/commit/73d551c28bdb61b348e0f1010893e968c46fdf98) have been fixed as part of this import & no longer have a `disable` property. See [jsondiff](https://jsondiff.com/?left=https://raw.githubusercontent.com/estuary/connectors/513fae26b52107ffbca4e1898cffc4d54e99a188/source-zendesk-support/tests/snapshots/snapshots__discover__capture.stdout.json&right=https://raw.githubusercontent.com/estuary/connectors/48af972315fafe2241c43dc73651493d657a02ab/source-zendesk-support/tests/snapshots/snapshots__discover__capture.stdout.json).

- **Capture snapshot changes**: Other than the addition of `op` and `row_id` properties in the `_meta` objects, the snapshots are the same. See [jsondiff](https://jsondiff.com/?left=https://raw.githubusercontent.com/estuary/connectors/513fae26b52107ffbca4e1898cffc4d54e99a188/source-zendesk-support/tests/snapshots/snapshots__capture__capture.stdout.json&right=https://raw.githubusercontent.com/estuary/connectors/48af972315fafe2241c43dc73651493d657a02ab/source-zendesk-support/tests/snapshots/snapshots__capture__capture.stdout.json).

**Workflow steps:**

Docs will need updated to reflect all of the supported bindings. See estuary/flow#1543.

**Notes for reviewers:**

A manual migration test on a local stack was performed to ensure the ATF and imported connector are compatible. No noteworthy differences were found.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1779)


source-zendesk-support/source_zendesk_support/source.py
```diff
@@ -127,20 +138,25 @@ def streams(self, config: Mapping[str, Any]) -> List[Stream]:
...
+        # but the Zendesk API does not provide a public API to get user's subscription plan. That's why we try to read 
+        # at least one record and expose this stream in case of success or skip it otherwise
         ticket_forms_stream = TicketForms(**args)
-        # TicketForms stream is only available for Enterprise Plan users but Zendesk API does not provide
-        # a public API to get user's subscription plan. That's why we try to read at least one record and expose this stream
-        # in case of success or skip it otherwise
         try:
             for stream_slice in ticket_forms_stream.stream_slices(sync_mode=SyncMode.full_refresh):
                 for _ in ticket_forms_stream.read_records(sync_mode=SyncMode.full_refresh, stream_slice=stream_slice):
-                    streams.append(ticket_forms_stream)
```
williamhbaker: Is it possible to have an Enterprise plan and have 0 records come back from `TicketForms`? In that case `TicketForms` wouldn't be added, and neither would `AccountAttributes` or `AttributeDefinitions`.

source-zendesk-support/source_zendesk_support/__main__.py
```diff
@@ -0,0 +1,44 @@
...
+            ),
+            accessTokenUrlTemplate=(f"https://estuarysupport.zendesk.com/oauth/tokens"),
+            accessTokenResponseMap={
+                "access_token": "/access_token",
+            },
+            accessTokenHeaders={
+                "content-type": "application/x-www-form-urlencoded",
+            },
+        ),
+        schema_inference=True,
```
williamhbaker, line 42: I don't think we want to use schema inference here, since the schemas for the different streams are fully known.

source-zendesk-support/VERSION
```diff
@@ -0,0 +1 @@
+v3
```
williamhbaker, line 1: The existing connector is `v1` I believe, and I think it makes sense to do a version bump, so this should probably be `v2` instead of `v3`.

williamhbaker (commented):
Looking really good, just a few relatively minor comments!

source-zendesk-support/VERSION
Alex-Bair: You're right that the existing connector is `v1`. When we briefly talked about how we were going to roll out the imported connector, I thought you mentioned bringing the `v1` connector into the `connectors` repo & renaming it to `v2`. Then the imported connector would be `v3`, and we could roll back to `v2` if there were issues with `v3`. Am I misremembering? Or maybe there's a better roll out strategy?

williamhbaker: We can bring the `v1` connector into the connectors repo as `v1` also, so we'll only need to go to `v2`

Alex-Bair: Got it, I'll change this one to `v2`. Thanks!

source-zendesk-support/source_zendesk_support/source.py
Alex-Bair: It is not possible to have an Enterprise plan & have no `TicketForms` records. Every Enterprise plan starts with a default ticket form, and the default ticket form can't be deleted. I tested & confirmed this behavior with API calls.

williamhbaker (approved):

PR #1781, opened 2024-08-06
source-mysql: Fix zero-valued time bug by upgrading MySQL client library

willdonnelly:
**Description:**

Apparently there was a dumb copy-paste bug in the MySQL client library which resulted in zero-valued `time` columns being decoded as `"0000-00-00"` instead of `"00:00:00"` as they ought to be. This bug only occurred in query results when using prepared statements and/or query parameters (as either of these cause the database to use a binary result format in which the zero-valued time has a special-cased representation), and only impacted that one value (because the bug was basically just `if n == 0 { return "0000-00-00" }` when it should have returned `"00:00:00"` instead).

Amazingly this bug had been present for the past 11 years, but was fixed two months ago, so we can fix it by simply upgrading the MySQL client library. Unfortunately that fix hasn't made it into a tagged release version yet, so this upgrades us to use the latest mainline commit hash. Once a new release is tagged we can move to using that instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1781)


williamhbaker (approved):
LGTM

PR #1784, opened 2024-08-07
source-notion: remove pattern_descriptor from config

williamhbaker:
Removes the "pattern_descriptor" property from the configuration, since I don't believe it was actually doing something and seems to have recently cause the endpoint config to fail to work with config encryption for unknown reasons.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1784)


jonwihl (approved):
LGTM

PR #1786, opened 2024-08-09
materialize-s3-iceberg: add ignoreStringFormat field configuration

williamhbaker:
**Description:**

Adds the ignoreStringFormat field configuration which when set will materialize string fields as regular strings, disregarding any format annotation they might have.

Manually tested by first materializing a collection without the field config set and seeing the column as a timestamptz, then adding the field configuration which causes an unsatisfiable constraint. Increasing the backfill counter allows the materialization to proceed by re-creating the table with a string type for the column, and string values are added to the column

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1786)


jgraettinger (approved):
LGTM

PR #1789, opened 2024-08-09
source-zendesk-support: Fix pagination issues with `tickets` and `users` streams

Alex-Bair:
**Description:**

`tickets` and `users` streams were previously using incremental timestamp-based pagination, resulting in duplicated data. It was also possible for the connector to get stuck making the same request if 1,000+ documents were modified at the same time. By moving to [incremental cursor pagination](https://developer.zendesk.com/documentation/ticketing/managing-tickets/using-the-incremental-export-api/#cursor-based-incremental-exports), these issues are resolved.

The `test_check_stream_state` unit test is no longer applicable, since the `cursorField` is not longer within each document. The `cursorField` is instead only in the response body, and the `IncrementalMixin`'s `state` setters and getters are used to checkpoint cursors between requests. The [`ClientSideIncrementalStream`](https://github.com/estuary/connectors/blob/14f92e7f20590cfac389bfee712d8656f8d131c6/source-hubspot/source_hubspot/streams.py#L794C7-L794C34) and [`IncrementalStream`](https://github.com/estuary/connectors/blob/14f92e7f20590cfac389bfee712d8656f8d131c6/source-hubspot/source_hubspot/streams.py#L906C7-L906C24) from the `source-hubspot` connector were referenced for determining how to checkpoint state use the Airbyte CDK.

Discover snapshot changes: Only the expected `cursorField` changes.

**Workflow steps:**

Docs will need updated to reflect the changed `cursorField` for the `tickets` and `users` streams.

**Notes for reviewers:**

Tested on a local stack and using a Zendesk account with a non-trivial amount of `tickets` and `users` data. Confirmed that cursors are checkpointed & the connector resumes with the last checkpointed cursor after recovering from a failure.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1789)

### DIFF
--- a/source-zendesk-support/tests/snapshots/snapshots__discover__capture.stdout.json
+++ b/source-zendesk-support/tests/snapshots/snapshots__discover__capture.stdout.json
@@ -3885,7 +3885,7 @@
       "stream": "tickets",
       "syncMode": "incremental",
       "cursorField": [
-        "updated_at"
+        "after_cursor"
       ]
     },
     "documentSchema": {
@@ -4468,7 +4468,7 @@
       "stream": "users",
       "syncMode": "incremental",
       "cursorField": [
-        "updated_at"
+        "after_cursor"
       ]
     },
     "documentSchema": {
--- a/source-zendesk-support/source_zendesk_support/streams.py
+++ b/source-zendesk-support/source_zendesk_support/streams.py
@@ -22,6 +22,7 @@ import pytz
 import requests
 from airbyte_cdk.models import SyncMode
 from airbyte_cdk.sources.streams.availability_strategy import AvailabilityStrategy
+from airbyte_cdk.sources.streams.core import IncrementalMixin
 from airbyte_cdk.sources.streams.http import HttpStream, HttpSubStream
 from airbyte_cdk.sources.streams.http.auth.core import HttpAuthenticator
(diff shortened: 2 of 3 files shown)


source-zendesk-support/tests/snapshots/snapshots__discover__capture.stdout.json
```diff
@@ -3885,7 +3885,7 @@
       "stream": "tickets",
       "syncMode": "incremental",
       "cursorField": [
-        "updated_at"
```
williamhbaker, line 3888: I assume this doesn't actually matter, since the `"after_cursor"` value is hard coded in the connector

williamhbaker (approved):
LGTM

source-zendesk-support/tests/snapshots/snapshots__discover__capture.stdout.json
Alex-Bair: Yep. Before, the `cursorField` was being used to retrieve a timestamp from one of the streamed documents. But since these streams' cursors are now being stored outside of the documents, we don't need `"updated_at"` to know where to look in the documents any more. 

The new `cursorField` is just the key we use when accessing the persisted `state`, and I gave it the same name as the response body's property `after_cursor` that contains the cursor.

PR #1797, opened 2024-08-13
materialize-bigquery: don't start the staged file unless there is data to write to it

williamhbaker:
**Description:**

If hard deletions are enabled it is possible that a transaction will contain only a single record for a document that is being deleted and doesn't already exist. In this case we won't write any data to the staged file or table.

This scenario uniquely impacts materialize-bigquery, since BigQuery requires there to actually be at least one file present when doing the equivalent of `SELECT * FROM gs://bucket/*.json`, and the rare case of a binding with only deletion of non-existent keys with hard deletions enabled will result in there not being any data to load.

There is likely a more principled way to accomplish this fix, but moving the "starting" of the staged file down a few lines gets it done quickly and easily, and in a way that is assured to not cause any other problems. This particular edge case is not a wide-spread issue, so I'm keeping the change isolated to BigQuery specifically.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1797)

### DIFF
--- a/materialize-bigquery/transactor.go
+++ b/materialize-bigquery/transactor.go
@@ -253,7 +253,6 @@ func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
 	log.Info("store: starting encoding and uploading of files")
 	for it.Next() {
 		var b = t.bindings[it.Binding]
-		b.storeFile.start()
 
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #1798, opened 2024-08-13
materialize-s3-iceberg: don't advance checkpoints in persisted state for bindings with no data

williamhbaker:
**Description:**

If a binding has no data for a transaction, its table will not be appended to and the "checkpoint" property of that table will not be updated.

The "PreviousCheckpoint" for that binding must also not be advanced in this case, otherwise the next append to the table will fail since the "PreviousCheckpoint" persisted in the state will not match the "checkpoint" property of the table.

By leaving the "PreviousCheckpoint" unchanged in the persisted connector state for bindings with no data, the next append files operation for that table will be able to transition the table from its last applied checkpoint to the current checkpoint.

This also includes a change that will update the persisted connector state after `Acknowledge` is complete to clear out the list of files that need to be appended once they are successfully appended. This is a best-effort and is not necessary for correctness, but does cut down on some of the work the connector has to do when it restarts sometimes.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1798)

### DIFF
--- a/materialize-s3-iceberg/transactor.go
+++ b/materialize-s3-iceberg/transactor.go
@@ -48,7 +48,7 @@ type bindingState struct {
 	// storage overhead of table snapshot files.
 	PreviousCheckpoint string   `json:"previousCheckpoint,omitempty"`
 	CurrentCheckpoint  string   `json:"currentCheckpoint,omitempty"`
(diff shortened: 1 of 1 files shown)


mdibaiee (approved):
LGTM

PR #1799, opened 2024-08-13
source-zendesk-support: fix `ticket_audits` schema and pagination

Alex-Bair:
**Description:**

Schema violations were occurring because the `value` in a `ticket_audits` document could be an integer. Also, `ticket_audits` were taking a significant amount of time to complete reading documents. This is because the API endpoint does not support filtering results by a start date, so we were reading all documents available. To fix this, logic to filter client-side has been added for the `ticket_audits` stream

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack with a non-trivial amount of documents. Ensured the document that previously caused a schema violation no longer causes an issue. Confirmed pagination works, and stream still gets documents incrementally after the initial backfill.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1799)

### DIFF
--- a/source-zendesk-support/acmeCo/ticket_audits.schema.yaml
+++ b/source-zendesk-support/acmeCo/ticket_audits.schema.yaml
@@ -202,6 +202,7 @@ properties:
               additionalProperties: true
             - type: string
             - type: "null"
+            - type: integer
         author_id:
           type:
             - "null"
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

PR #1570, opened 2024-05-09
materialize-databricks: support decimal column validation

mdibaiee:
**Description:**

- Tested this using my other branch https://github.com/estuary/connectors/pull/1571 which also runs Validate as part of integration tests. Creating the tables, not deleting them, and running integration tests again, verified an error happens without this pull-request complaining about unknown type `DECIMAL`, whereas with this pull-request Validate passes 👍🏽 
- Separated the other pull-request since it requires more time put into it to update all tests for all materialization connectors

**Workflow steps:**

(How does one use this feature, and how has it changed)

**Documentation links affected:**

(list any [documentation links](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing) that you created, or existing ones that you've identified as needing updates, along with a brief description)

**Notes for reviewers:**

(anything that might help someone review this PR)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1570)

PR #1801, opened 2024-08-13
materialize-sql: generalized custom DDL and handling of huge numbers

williamhbaker:
**Description:**

The main functional changes here, for materializations using the `materialize-sql` framework:
- Adds field configuration to set user-defined "custom" DDL as a general capability
- Similarly adds a field configuration to materialize any field a string, converting its value to a string if needed. This may be useful for various situations, but the one I have in mind right now is for dealing with very large numeric values that would not have any other way to be materialized.
- New handling for projections with numeric inference where the minimum/maximum are outside the ranges of a signed 64-bit integer, and handling for projection with numeric formats where the string lengths are extremely long. This enhancement will be available to materialized collections using schema inference, which isn't all collections but it should be most of the ones where we most often see issues.
- A few other opportunistic removals of backward-compatibility shims that we've had in place for a long time. Since the above changes will come with a version bump, I want to take this opportunity to clean up these things too.

The major changes are in the `materialize-sql` package itself, specifically `type_mapping.go`. Then every SQL materialization was updated to match that.

Reviewing commit-by-commit will probably be easiest. There's a lot of churn in SQL generation tests in the first commit, where I normalized all of tests to use the same fixtures. Subsequent commits make smaller updates to the snapshot outputs, illustrating the functional code changes. Also see commit messages for more detail, and some in-line comments I have made while doing a self-review.

There are runtime updates in progress to add numeric inference to the Flow protocol that this change depends on. Once those are landed, I will update the Flow pin. This will be needed to get all of the snapshot tests to pass. I will also do a final round of manual QA tests once that runtime updates are out prior to merging.

The main reason for bumping all of the connector versions is so that they can be migrated to the new type mapping definitions such that they will generate `UNSATISFIABLE` constraints if an previously exist columns that _did_ pass validation now don't. I expect there to be some of these, but hopefully not too many, and the migration will set the `DDL` field configuration to get things working in that case. I'd rather get all of these kludges and shims in the code cleaned up now while it is still practical to do so (although it will be quite tedious) instead of continuing to maintain them forever going forward. My plan is to migrate all existing tasks to the new versions over the course of a couple of days once this lands.

**Workflow steps:**

I'm going to put together what will probably end up being a rather large documentation update as part of this which will mostly fill in gaps in our current documentation. I'd like for each SQL materialization to have a table of Flow field types -> column types, like [this](https://docs.estuary.dev/reference/Connectors/materialization-connectors/amazon-s3-iceberg/#iceberg-column-types). I will also write documentation for how to use these common field configurations, and how a user could manually migrate a destination table's column type to be compatible with a change in a collection schema, since that is something that has been possible for a while but we have not had documentation for.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1801)


materialize-motherduck/driver_test.go
```diff
@@ -137,27 +136,6 @@ func TestPrereqs(t *testing.T) {
 			cfg:  func(cfg config) *config { return &cfg },
 			want: nil,
 		},
-		{
-			name: "token is well formed but wrong",
```
williamhbaker, line 141: Some of these tests were broken and of questionable use, so I am removing them.

go.mod
```diff
@@ -29,7 +29,7 @@ require (
 	github.com/databricks/databricks-sql-go v1.5.5
 	github.com/dropbox/dropbox-sdk-go-unofficial/v6 v6.0.5
 	github.com/elastic/go-elasticsearch/v8 v8.9.0
-	github.com/estuary/flow v0.3.12-0.20240118233153-7681c66d4e1f
+	github.com/estuary/flow v0.4.1-0.20240814191845-c228194b3e2f
```
williamhbaker, line 32: TODO: Update this pin once runtime updates are merged. This will be needed to get the tests passing.

materialize-databricks/sqlgen.go
```diff
@@ -48,64 +26,36 @@ var databricksDialect = func() sql.Dialect {
...
-
-	mapper = sql.NullableMapper{
-		NotNullText: "NOT NULL",
-		Delegate:    mapper,
-	}
-
-	columnValidator := sql.NewColumnValidator(
-		sql.ColValidation{Types: []string{"string"}, Validate: stringCompatible},
-		sql.ColValidation{Types: []string{"boolean"}, Validate: sql.BooleanCompatible},
-		sql.ColValidation{Types: []string{"long", "decimal"}, Validate: sql.IntegerCompatible},
```
williamhbaker, line 104: Note that "decimal" was added here in https://github.com/estuary/connectors/pull/1570 to support a specific use case. The new custom DDL capability handles this by allowing fields with custom DDL to pass validation with any column type, why is the only practical thing we can really do.

materialize-mysql/sqlgen_test.go
```diff
@@ -18,115 +15,40 @@ import (
...
-			addColumns: addCols,
-		},
-		{
-			name:         "alter table drop not nulls",
-			dropNotNulls: dropNotNulls,
+		sql.TestTemplates{
+			TableTemplates: []*template.Template{
+				templates.createTargetTable,
+				templates.createLoadTable,
+				templates.createUpdateTable,
```
williamhbaker, line 33: `createUpdateTable` was not previously included in the snapshot tests

materialize-postgres/sqlgen.go
```diff
@@ -10,62 +10,43 @@ import (
...
 		},
-	}
-
-	mapper = sql.NullableMapper{
-		NotNullText: "NOT NULL",
-		Delegate:    mapper,
-	}
-
-	columnValidator := sql.NewColumnValidator(
-		sql.ColValidation{Types: []string{"bigint", "integer"}, Validate: sql.IntegerCompatible},
```
williamhbaker, line 56: IIRC there was a single case of a pre-existing table where a column type of `integer` needed to be allowed. If this task is still around it an use a custom DDL field config.

materialize-bigquery/sqlgen.go
```diff
@@ -120,32 +116,11 @@ var bqDialect = func() sql.Dialect {
 			return "?"
 		}),
 		TypeMapper:             mapper,
-		ColumnValidator:        columnValidator,
 		MaxColumnCharLength:    300,
 		CaseInsensitiveColumns: true,
 	}
 }()
 
-// bignumericCompatible allows either integers or numbers, since we used to create number columns as
```
williamhbaker, line 129: As noted here, we used to create `type: number` columns as bignumeric but have been creating them as float64 for quite a long time. I'm removing the ability to validate pre-existing bignumeric columns for `type: number` fields, unless there is a user-defined DDL set. There are probably some existing tasks that will need to have this DDL set for as part of the migration.

materialize-postgres/sqlgen.go
williamhbaker: Note to self: Our stats materialization actually will require this. That might have been the only case where this is needed.

materialize-snowflake/sqlgen.go
```diff
@@ -63,45 +48,33 @@ func (m timestampTypeMapping) valid() bool {
...
-				"number": sql.PrimaryKeyMapper{
-					PrimaryKey: sql.NewStaticMapper("STRING"),
-					// https://docs.snowflake.com/en/sql-reference/data-types-numeric#special-values
-					Delegate: sql.NewStaticMapper("DOUBLE", sql.WithElementConverter(sql.StdStrToFloat("NaN", "inf", "-inf"))),
+	mapper := sql.NewDDLMapper(
+		sql.FlatTypeMappings{
+			sql.ARRAY:    sql.MapStatic("VARIANT", sql.UsingConverter(sql.ToJsonBytes)),
+			sql.BINARY:   sql.MapStatic("TEXT"),
+			sql.BOOLEAN:  sql.MapStatic("BOOLEAN"),
+			sql.INTEGER:  sql.MapStatic("INTEGER", sql.AlsoCompatibleWith("number")), // INTEGER DDL is actually an alias for NUMBER
```
williamhbaker, line 56: We could actually use the `NUMBER` DDL to create these columns and not need the "number" alias but it's already confusing enough that snowflake calls a column "number" when you create it as an "integer" so we'll try to have our code and test snapshots remain somewhat comprehensible.

materialize-sql/type_mapping.go
```diff
@@ -24,15 +20,17 @@ type FlatType string
...
+	ARRAY          FlatType = "array"
+	BINARY         FlatType = "binary"
+	BOOLEAN        FlatType = "boolean"
+	INTEGER        FlatType = "integer"
+	MULTIPLE       FlatType = "multiple"
+	NEVER          FlatType = "never"
+	NUMBER         FlatType = "number"
+	OBJECT         FlatType = "object"
+	STRING         FlatType = "string"
+	STRING_INTEGER FlatType = "string_integer"
```
williamhbaker, line 32: Adding these are separate `FlatType`'s helped clean up their handling. `STRING_INTEGER` and `STRING_NUMBER` are only generated for fields that aren't primary keys, so each materialization's type mapping configuration doesn't need to account for that.

materialize-sql/type_mapping.go
```diff
@@ -142,379 +144,169 @@ type MappedType struct {
...
-		out, ok := i.SetString(str, 10)
-		if !ok {
-			return nil, fmt.Errorf("could not convert %q to big.Int", str)
-		}
-		return out, nil
-	})
+type StringMappings struct {
+	Fallback        MapProjectionFn
+	WithFormat      map[string]MapProjectionFn
+	WithContentType map[string]MapProjectionFn
```
mdibaiee, line 278: A comment explaining what ContentType means would be useful here, it took me some research to find out what it is

materialize-sql/type_mapping.go
```diff
@@ -142,379 +144,169 @@ type MappedType struct {
...
-type PrimaryKeyMapper struct {
-	PrimaryKey TypeMapper
-	Delegate   TypeMapper
+type fieldConfig struct {
+	// CastToString will materialize the field as a string representation of its
+	// value.
+	CastToString bool `json:"castToString"`
+	// IgnoreStringFormat is a legacy configuration used as an alias for
+	// castToString.
+	IgnoreStringFormat bool `json:"ignoreStringFormat"`
```
mdibaiee, line 338: Not sure if we generate a jsonschema for these options to be included somewhere, if we do we should probably hide this field from the jsonschema

mdibaiee (commented):
Reviewed the first two commits so far, will review the rest after lunch

materialize-sql/type_mapping.go
williamhbaker: We don't current generate a jsonschema for these. It'll be something to keep in mind if we ever want to show field config options in the UI, which seems like a possibility.

materialize-sql/type_mapping.go
williamhbaker: Added some additional comments here!

PR #1802, opened 2024-08-14
source-zendesk-support: handle bools in `ticket_skips` documents

Alex-Bair:
**Description:**

Very rarely, we get boolean values in a `TicketSkips` stream's custom field values. This change coerces those bools to strings, when present. Using the `parse_response` method for this purpose is similar to its use in other streams, like `TicketComments`.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that boolean values `True` and `False` are changed to strings & that when the value is not a boolean it's left unchanged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1802)

### DIFF
--- a/source-zendesk-support/source_zendesk_support/streams.py
+++ b/source-zendesk-support/source_zendesk_support/streams.py
@@ -976,6 +976,18 @@ class TicketSkips(SourceZendeskSupportCursorPaginationStream):
 
         return params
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #1804, opened 2024-08-14
source-zendesk-support: change streams to cursor pagination

Alex-Bair:
**Description:**

Zendesk's API allows the use of both offset pagination and cursor pagination to get documents. Previously, we were using offset pagination, which is limited to only getting the first 100 pages of documents. Cursor pagination does not have this limit, so all streams using offset pagination have been changed to use cursor pagination.

Since many streams inherit from common classes, some logic was pushed down the inheritance chain to streams that don't use cursor pagination.

Additionally, the `TicketAudits` stream was previously retrieving all pages of documents. This has been changed to only get pages containing documents after the start date/last seen date.

**Workflow steps:**

**Notes for reviewers:**

Tested all streams on a local stack.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1804)


williamhbaker (approved):
LGTM

PR #1805, opened 2024-08-14
source-*-batch: Improve poll scheduling around interruption

willdonnelly:
**Description:**

If the connector gets restarted partway through a table poll we want it to retry immediately, which means that we shouldn't have updated the 'Last Polled' timestamp state yet.

This commit changes that for all four batch SQL connectors, so now we'll have an unconditional checkpoint after the result set processing ends (where previously we only had that if there were still uncommitted documents) and before that checkpoint we update the 'Last Polled' timestamp to the time at which we *started* the query.

I also tweaked a bit of logging that I thought could be better and added a couple of missing error checks for Postgres, since I'm already touching that part of the code and I noticed the after-query check on `rows.Err()` was missing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1805)

### DIFF
--- a/source-bigquery-batch/driver.go
+++ b/source-bigquery-batch/driver.go
@@ -579,6 +579,7 @@ func (c *capture) poll(ctx context.Context, binding *bindingInfo, tmpl *template
 	log.WithFields(log.Fields{
 		"name": res.Name,
 		"poll": pollScheduleStr,
+		"prev": state.LastPolled.Format(time.RFC3339Nano),
 	}).Info("waiting for next scheduled poll")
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

Seems like we should probably be emitting partial checkpoints for postgres too and I'm not seeing that anywhere, but it was like that already so not necessarily something that needs addressed now.

willdonnelly:
That's a good point. That's a trivial change while I'm already thinking about it, but there's no reason it needs to be in this PR so I'll upload that as a separate thing in a minute.

PR #1806, opened 2024-08-14
source-postgres-batch: Emit partial-progress checkpoints

willdonnelly:
**Description:**

It's somewhat debatable how useful partial-progress checkpoints are nowadays since there are no hard limits on Flow transaction sizing or anything, but every other Batch SQL capture has them, so it's probably good for `source-postgres-batch` to have them as well for consistency if nothing else.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1806)

### DIFF
--- a/source-postgres-batch/driver.go
+++ b/source-postgres-batch/driver.go
@@ -23,6 +23,14 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #1807, opened 2024-08-15
source-zendesk-support: redact fields in capture snapshot

Alex-Bair:
**Description:**

Zendesk automatically archives some data, which alters the capture snapshot even though nothing in the connector has changed. Redacting these fields makes the capture snapshot test a little more robust. Some other fields may need redacted in the future if we find they're changing too, but this should be a good start.

Additionally, the posts schema was made more permissive due to schema violations we've encountered.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1807)

### DIFF
--- a/source-zendesk-support/tests/test_snapshots.py
+++ b/source-zendesk-support/tests/test_snapshots.py
@@ -29,6 +29,16 @@ def test_capture(request, snapshot):
             unique_stream_lines.append(line)
             seen.add(stream)
 
+    for l in unique_stream_lines:
+        _, rec = l[0], l[1]
+
+        rec['_meta']['row_id'] = 0
+        if "updated_at" in rec:
+            rec["updated_at"] = "redacted"
+        if "last_login_at" in rec:
+            rec["last_login_at"] = "redacted"
+
+
     assert snapshot("capture.stdout.json") == unique_stream_lines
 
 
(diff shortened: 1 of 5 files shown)


source-zendesk-support/tests/test_snapshots.py
```diff
@@ -29,6 +29,41 @@ def test_capture(request, snapshot):
...
+        "acmeCo/schedules",
+        "acmeCo/ticket_forms",
+        "acmeCo/account_attributes",
+        "acmeCo/attribute_definitions"
+    ]
+
+    for l in unique_stream_lines:
+        type, rec = l[0], l[1]
+
+        rec['_meta']['row_id'] = 0
```
williamhbaker, line 60: It would probably be alright to simplify this along the lines of:

```
if "updated_at" in rec:
    rec['updated_at'] = 'redacted'
if "last_login_at" in rec:
    rec['last_login_at'] = 'redacted'
```

williamhbaker (approved):
LGTM

One non-blocking suggestion

source-zendesk-support/tests/test_snapshots.py
Alex-Bair: That's a lot simpler, I'll make that change. Thanks for the suggestion!

PR #1812, opened 2024-08-15
source-postgres: Support NaN values in float4/float8 columns

willdonnelly:
**Description:**

Adds support for NaN (and Infinity and -Infinity) values in float4/float8 columns. Previously these didn't work because we serialize them as JSON numbers and there's no legal JSON representation of a floating-point NaN or infinity. Now we stringify just those three special values, which required also updating the discovery output for those column types to `{type: [string, number], format: number}` but I am assured that our materializations should handle that correctly so it should basically be a no-op for any dataflows that don't encounter NaNs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1812)

### DIFF
--- a/source-postgres/.snapshots/TestDiscoveryComplex
+++ b/source-postgres/.snapshots/TestDiscoveryComplex
@@ -44,8 +44,12 @@ Binding 0:
               "description": "(source type: non-nullable text)"
             },
             "real_": {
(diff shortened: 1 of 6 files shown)


jgraettinger (approved):
LGTM

PR #1815, opened 2024-08-15
source-klaviyo: updating schemas to minimal schema

Luishfs:
**Description:**

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1815)


source-klaviyo/source_klaviyo/schemas/events.json
```diff
@@ -4,80 +4,6 @@
   "additionalProperties": true,
   "required": ["id"],
   "properties": {
-    "type": { "type": "string" },
-    "id": { "type": "string" },
-    "datetime": { "type": "string", "format": "date-time" },
```
Alex-Bair, line 9: `datetime` is the cursor field for the `events` stream, right? I don't think omitting it here will cause an issue, but it leaving the cursor field in shouldn't cause a problem since we're relying on it being in every document.

Alex-Bair (approved):
LGTM. One comment about cursor fields in the schema files, but I don't think it should break anything to leave them out.

source-klaviyo/source_klaviyo/schemas/events.json
Luishfs: @Alex-Bair There actually was a error with cursor fields being defined in the schema, which caused this [PR](https://github.com/estuary/connectors/pull/1983) to be created. I think it was more related to their schema being set as nullable, but in order to be sure i just simply remove then now

PR #1826, opened 2024-08-19
source-pendo: import to Estuary CDK

Alex-Bair:
**Description:**

Imports `source-pendo` to the Estuary CDK. This connector uses Airbyte's [Low-Code CDK](https://docs.airbyte.com/connector-development/config-based/low-code-cdk-overview), where almost everything is specified in `manifest.yaml`.  See the [Airbyte docs](https://docs.airbyte.com/connector-development/config-based/understanding-the-yaml-file/yaml-overview) for more specific information on what can be included in `manifest.yaml`.

**Workflow steps:**

Documentation will need created since this is a new connector. Docs PR is [here](https://github.com/estuary/flow/pull/1568).

**Notes for reviewers:**

No test account for Pendo with API access was available, so the only checks performed were making sure the connector builds successfully and renders correctly in the UI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1826)


source-pendo/source_pendo/manifest.yaml
```diff
@@ -0,0 +1,164 @@
+version: 0.78.5
+
+type: DeclarativeSource
+
+check:
+  type: CheckStream
```
williamhbaker, line 6: I think we can leave the metadata streams included, and have the check still be based on the account info stream. It seems like this would work since the key would default to the row_id.

Alex-Bair: Will do, I'll add them back in.

source-pendo/acmeCo/flow.yaml
```diff
@@ -0,0 +1,110 @@
+---
```
williamhbaker, line 1: We'd usually not have the `acmeCo` directory for connectors that aren't doing a capture snapshot test of the capture's output. Just leave `bindings: []` (empty) in the `test.flow.yaml`.

williamhbaker (approved):
LGTM % Comment

PR #1828, opened 2024-08-19
sqlcapture: Handle context cancellation during change streaming

willdonnelly:
**Description:**

Currently if the context passed to 'streamToWatermarkWithOptions' is cancelled but the events channel doesn't close, the connector will continue to wait for a watermark which may never arrive.

Most of the time when the context is cancelled it's at the top level of the capture so this doesn't come up, but the indefinite streaming state is currently modelled as a process of running the change streaming process in a worker thread with a second thread which will wait for a reasonable period of time and then write the watermark.

The problem is, if this watermark write fails we want the work context cancellation to cause the stream-until-watermark thread to also shut down, and as currently written it doesn't.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1828)

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -528,47 +528,51 @@ func (c *Capture) streamToWatermarkWithOptions(ctx context.Context, replStream R
 	var eventCount int
 	defer func() { logrus.WithField("events", eventCount).Info("processed replication events") }()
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #1829, opened 2024-08-20
dbt: DBT job trigger in warehouse materializations

mdibaiee:
**Description:**

- Trigger DBT jobs after transactions which materialize data

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1829)

### DIFF
--- a/materialize-databricks/.snapshots/TestSpecification
+++ b/materialize-databricks/.snapshots/TestSpecification
@@ -118,6 +118,57 @@
         "title": "Sync Schedule",
         "description": "Configure schedule of transactions for the materialization."
       },
+      "dbt_job_trigger": {
+        "properties": {
+          "job_id": {
+            "type": "string",
+            "title": "Job ID",
+            "description": "DBT job ID"
+          },
+          "account_id": {
+            "type": "string",
+            "title": "Account ID",
+            "description": "DBT account ID"
+          },
+          "account_prefix": {
+            "type": "string",
+            "title": "Account Prefix",
+            "description": "DBT account prefix"
+          },
+          "api_key": {
+            "type": "string",
+            "title": "API Key",
+            "description": "DBT API Key",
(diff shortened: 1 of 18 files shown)


dyaffe:
This option only existing if a sync schedule is specified makes sense to me.

Also, would it be possible to have our first version include all warehouse?  That would involve adding support in Snowflake, Bigquery and Redshift.

mdibaiee:
@dyaffe just had a chat with Dani regarding the main concern, now going to add an option to choose between three different modes of treating already-running jobs: `skip` triggering a job if one is already running, `replace` the running job with a new one, or `ignore` the running job and trigger the job anyway.

After that will add to all warehouses

materialize-databricks/.snapshots/TestSpecification
```diff
@@ -138,7 +189,8 @@
         "type": "object",
         "title": "Advanced Options",
         "description": "Options for advanced users. You should not typically need to modify these.",
-        "advanced": true
+        "advanced": true,
+        "order": 6
```
williamhbaker, line 193: A few of these have `order` set in their configurations but others don't - just wondering if there is a specific reason for that?

williamhbaker (approved):
LGTM

At some point in the future it could be worth considering moving some of this behavior to the transactor, like we have done for the sync schedule. Right now it's only for SQL materializations but I could see others (Iceberg) also wanting to use this which would make it interesting as a general feature there.

materialize-databricks/.snapshots/TestSpecification
mdibaiee: I haven't set order for `dbt_job_trigger` at all, some connectors' configurations generally have `order` while others don't... the diff is for other properties of those connectors

PR #1832, opened 2024-08-21
source-zendesk-support: fix oauth

Alex-Bair:
**Description:**

The imported `source-zendesk-support` connector's OAuth did not work. The OAuth code in `__main__.py` was updated to work with the OAuth app created in our Zendesk testing account.

I also changed the subdomain of our testing account to "d3v-estuary" since it might be viewable if someone inspects HTTP requests during the OAuth process., Although the subdomain change is reflected in the capture snapshot, I don't anticipate we will need to change this again, so the affected snapshotted properties don't need redacted.

**Workflow steps:**

When creating a Zendesk Support capture, enter the subdomain, then click the OAuth button. Follow the prompts to login to your Zendesk account, then allow Estuary access.

Docs will need updated - authentication can be done with OAuth now.

**Notes for reviewers:**

Tested on a local stack. Confirmed streams still work as before with our Zendesk testing account when authenticated with OAuth.

The `{{{ config.subdomain }}}` syntax in `__main__.py` comes from our [docs](https://github.com/estuary/connectors/tree/main/python#the-url) and gets the user-entered subdomain in the UI. This means OAuth will fail if the user hasn't entered a subdomain prior to initiating the OAuth process. We currently don't require the subdomain before pressing the OAuth button, so that failure could be unexpected if a user tries to complete OAuth before filling out earlier fields. I think this is unlikely to happen though, and we don't handle this situation with other connectors (like `source-shopify`), so I don't think this needs addressed right now.

In order for the OAuth app in our Zendesk testing account to work with other Zendesk accounts, I had to request a [global OAuth client](https://developer.zendesk.com/documentation/marketplace/building-a-marketplace-app/set-up-a-global-oauth-client/#set-up-a-global-oauth-client). The request was made 21AUG2024, and the expected turnaround is about a week. I'll hold off merging into main until Zendesk finishes their part of the process.

The client id and client secret in the production database will need updated when this PR is merged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1832)

### DIFF
--- a/source-zendesk-support/config.yaml
+++ b/source-zendesk-support/config.yaml
@@ -3,7 +3,7 @@ credentials:
     email: alexb@estuary.dev
     api_token_sops: ENC[AES256_GCM,data:WxXkpbpDG5dfxToEkz41WopKbmj1OrCt43X3Nzhrk9bs3XNkHR1jpg==,iv:2V30uIHCj/fzt6YekkvxqLrIm7cJ4VqeRvp5h+8hqx4=,tag:u8MBtzFdbMER9r4PoLFFFA==,type:str]
 start_date: "2024-07-26T00:00:00Z"
-subdomain: d3v-estuary6996
+subdomain: d3v-estuary
 sops:
     kms: []
     gcp_kms:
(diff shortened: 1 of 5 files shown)


Alex-Bair:
Update: Zendesk has finished making our Zendesk OAuth app into a global one. I tested on a local stack & confirmed the OAuth process succeeds with a Zendesk account that's not ours!

williamhbaker (approved):
LGTM

PR #1841, opened 2024-08-22
source-oracle: fix first run on redo log files with no dictionary

mdibaiee:
**Description:**

- the first run will not be able to find any archive logs with dictionaries, so it will yield NULL for `MAX(NEXT_CHANGE#) from V$ARCHIVELOG`. In these instances we just find the first log file that contains the SCN range we are looking for.
- SCNs are actually 48 bit in size, so we should store them in int64s, updated that
- Turns out if there is an `ANYDATA` column in a table, the logminer query will use a table alias (presumable to use another alias for the anydata in case its properties need to be updated). As I wrote a test for unsupported data types, also found this issue, which is now resolved by parsing the left hand side of comparison expressions in `WHERE` statements as `sqlparser.ColName` and using the column name without its alias qualification.
- Also remove columns which are not part of the discovered columns from the resulting document to avoid ambiguous behavior for those columns

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1841)

### DIFF
--- a/source-oracle/.snapshots/TestUnsupportedTypes-backfill
+++ b/source-oracle/.snapshots/TestUnsupportedTypes-backfill
@@ -1,0 +1,9 @@
+# ================================
+# Collection "acmeCo/test/flow_test_logminer_t18110541": 1 Documents
+# ================================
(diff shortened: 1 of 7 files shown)


williamhbaker (approved):
LGTM

PR #1850, opened 2024-08-22
materialize-sql: remove CheckedInt64 converter

williamhbaker:
**Description:**

This removes the CheckedInt64 converter, the only purpose of which was to evaluate if a value would fit into a signed 64-bit integer column, and produce a nice error message if it didn't.

It was flawed because it did not account for values transmitted as floats when they are small and have no fractional component.

I may revisit this in the future. For now this just takes out the check.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1850)


willdonnelly (approved):
LGTM

PR #1851, opened 2024-08-23
materialize-databricks: do not validate string integers as LONG columns

williamhbaker:
**Description:**

Although this would appear to be a backward compatible transition, the explicit
casting to `NUMERIC(38,0)` that is done in Store queries does not work with
pre-existing `LONG` columns.

Any materializations with pre-existing `LONG` columns for `type: string, format:
integer` fields will need the `LONG` DDL field configuration to be upgraded.
There are a few more of these than I wanted to deal with manually but it
shouldn't be _that_ bad, and it will set us up better for the future.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1851)

### DIFF
--- a/materialize-databricks/sqlgen.go
+++ b/materialize-databricks/sqlgen.go
@@ -42,9 +42,7 @@ var databricksDialect = func() sql.Dialect {
 			sql.OBJECT:   jsonMapper,
 			sql.MULTIPLE: jsonMapper,
 			sql.STRING_INTEGER: sql.MapStringMaxLen(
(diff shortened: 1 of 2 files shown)


mdibaiee (approved):

williamhbaker:
Added on a small fix to the `castToString` DDL configuration when it is used for fields that are already strings, but might have some format annotation that woudl cause them to end up as a different column anyway.

PR #1863, opened 2024-08-27
source-pendo: rebuild in Estuary CDK and add event streams

Alex-Bair:
**Description:**

The `source-pendo` connector has been rebuilt with our CDK, making the connector more extensible than its previous version that utilized Airbyte's low-code framework.

Five incremental event streams have been added by querying Pendo's [Aggregation API](https://engageapi.pendo.io/#7c8479b8-3843-403c-94a9-04cbdf542db9):
1. `pageEvents`
2. `featureEvents`
3. `trackEvents`
4. `guideEvents`
5. `pollEvents`

Authorization with Pendo's API requires placing the API key in a custom `x-pendo-integration-key` header. Our CDK did not support custom headers, so I modified the `HTTPMixin` class to allow custom authorization headers.

**Workflow steps:**

Pendo docs will need updated to reflect the added streams. Docs PR will be linked when its ready.

**Notes for reviewers:**

Tested end-to-end on a local stack. Manually confirmed pre-existing streams' schemas are the same after the rebuild. Confirmed all new event streams are incremental and work as intended.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1863)


.github/workflows/python.yaml
```diff
@@ -137,7 +137,7 @@ jobs:
             usage_rate: "1.0"
           - name: source-pendo
             type: capture
-            version: v1
+            version: v2
```
williamhbaker, line 140: Any reason to do a version bump here? I wouldn't think we'd need to based on these changes the current usage of the connector.

source-pendo/source_pendo/models.py
```diff
@@ -0,0 +1,76 @@
...
+        title="API Token",
+        json_schema_extra={"secret": True},
+    )
+
+
+ConnectorState = GenericConnectorState[ResourceState]
+
+
+class Resource(BaseDocument, extra="allow"):
+    id: str = ""
```
williamhbaker, line 54: There are a lot of fields in these models that have default values. Usually they shouldn't, since that is actually schematized with a JSON default. Especially for `id` which I'd imagine we always want to have present, and it should be an error if it isn't there. And for `id` in particular setting a default value makes it's schema not have it be a required field, and having the collection key be a require field is pretty much universally required in order for the collection to be materialized.

So I'd generally recommend getting rid of the default values for these models. You _could_ use the `| None` union to represent optional fields, but probably best for those to just not even explicitly schematize them.

source-pendo/source_pendo/resources.py
```diff
@@ -0,0 +1,181 @@
...
+    fetch_resources,
+    fetch_events,
+    fetch_metadata,
+)
+
+def resources(
+        log: Logger, http: HTTPMixin, config: EndpointConfig
+) -> list[common.Resource]:
+
+    async def snapshot(http: HTTPMixin, entity: str, log: Logger) -> AsyncGenerator[Metadata, None]:
```
williamhbaker, line 34: I know this seems like a pattern in other connectors with snapshot bindings, but it seems like we should be able to have `fetch_resources` (etc.) itself be a generator, and then use it directly (with partial function application) as the `fetch_snapshot` parameter. Does that not work for some reason?

source-pendo/source_pendo/resources.py
```diff
@@ -0,0 +1,181 @@
...
+            task,
+            fetch_changes=functools.partial(
+                fetch_events,
+                http,
+                entity,
+                timestamp_field
+            )
+        )
+
+    start_date = config.startDate
```
williamhbaker, line 149: Usually we would allow an empty start date configuration, and have some default like start capturing X days ago. Would that make sense to do here?

source-pendo/source_pendo/models.py
```diff
@@ -0,0 +1,76 @@
...
+]
+
+# Supported metadata types for the Pendo API and their corresponding resource name.
+METADATA_TYPES: list[tuple[str, str]] = [
+    ("account", "AccountMetadata"),
+    ("visitor", "VisitorMetadata")
+]
+
+# Supported event types for the Pendo Aggregation API, their corresponding resource name, their associated identifying field, and their "last updated" timestamp field.
+EVENT_TYPES: list[tuple[str, str, str]] = [
```
williamhbaker, line 26: This tuple needs to have 4 strs

source-pendo/source_pendo/api.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+async def fetch_resources(
+        http: HTTPSession,
+        entity: str,
+        log: Logger,
+) -> Iterable[Resource]:
+    url = f"{API}/{entity}"
+
+    response = json.loads(await http.request(log, url))
```
williamhbaker, line 20: The `json.loads` and `json.dumps` indicate a mismatch in how the API is presenting data vs. how we are modeling it. Fortunately there's a pretty easy way to resolve this I think...if we're expecting a list of `Resource`, I think you can use a pydantic `TypeAdapter` to clean this up, something like:

```
TypeAdapter(list[Resource]).validate_json(await http.request(...))
```

source-pendo/source_pendo/api.py
```diff
@@ -0,0 +1,123 @@
...
+
+    resources: list[Resource] = []
+    for id, resource in enumerate(response):
+        r = Resource.model_validate_json(json.dumps(resource))
+        r.meta_ = Resource.Meta(op="c", row_id=id)
```
williamhbaker, line 25: For snapshot bindings, I don't think we should need to or want to set the metadata manually like this, since it should handled here: https://github.com/estuary/connectors/blob/602f4bb3a139f944dfad68cff890bf98b8771759/estuary-cdk/estuary_cdk/capture/common.py#L523-L525

source-pendo/source_pendo/api.py
```diff
@@ -0,0 +1,123 @@
...
+        resources.append(r)
+
+    log.info(f"Read {len(resources)} documents from stream {entity}.")
```
williamhbaker, line 28: Not sure if these logs are intended to be kept, but I think it would be best to remove them - this information is available from stats actually

source-pendo/source_pendo/api.py
```diff
@@ -0,0 +1,123 @@
...
+        },
+        "request": {
+            "pipeline": [
+                {
+                    "source": {
+                        entity: None,
+                        "timeSeries": {
+                            "period": "hourRange",
+                            "first": f"{last_ts}",
+                            "last": "now()"
```
williamhbaker, line 83: Main conceptual question here: For these aggregate events, does the API return "partial" hour aggregates? For example, if we request the value for the current hour and the current hour isn't complete, are we going to get an incomplete aggregation and then never see it again?

Another thought is that we should probably bound the size of these responses. If the configured start time is far in the past, I'd imagine we could end hitting API limits or OOMing the connector on the response size. Could we use `count` to get 24 hours at a time or something like that?

.github/workflows/python.yaml
Alex-Bair: The main reason I did a version bump was because the endpoint config has `startDate` as a new required field. I agree that with how many tasks currently exist for this connector, we could keep it as `v1`. I'll undo the version bump.

source-pendo/source_pendo/api.py
Alex-Bair: I'll make sure to remove them after addressing all comments. They're helpful for me when stats are visible locally.

source-pendo/source_pendo/api.py
Alex-Bair: My understanding from the docs & testing is that aggregates returned by the API are as up to date as they can be, and a timestamp field is used to indicate when that aggregation was last updated. So for the current hour, we'll get an "incomplete" aggregation. But if that aggregation is updated within the current hour, it will be returned in a later API response since it's "last updated" timestamp (either `lastTime` or `guideTimestamp`) will be updated. The filter in the request body makes sure we get all aggregations with a "last updated" timestamp after the last aggregate we saw.

To bound the response sizes, there's no option to specify a certain number of days. We can use `limit` to limit how many lines we get though. I'll test that out & try to empirically find a reasonable number.

source-pendo/source_pendo/resources.py
Alex-Bair: Yep, I think that makes sense. I actually used Pydantic's [`default_factory`](https://docs.pydantic.dev/latest/concepts/fields/#default-values) this morning to do just that, but the default value wasn't populated into the config. I think I could have something like `config.startDate = config.startDate` in the `all_resources`  method to make sure that default value get populated. I'll keep looking into it.

source-pendo/source_pendo/models.py
Alex-Bair: For the various event streams, I use the `| None` union for fields that exist on one stream, but don't exist on other streams (ex: `pageId` only exists on `pageEvents` but not `featureEvents`). To avoid having those fields be optional for the other streams, I think I'll have to make separate models for each one. I'll give it a try.

williamhbaker: You could consider just omitting the fields entirely too since they will be determined by schema inference. Having one model that only has the fields that are common to all event streams would be totally fine.

source-pendo/source_pendo/resources.py
Alex-Bair: Making `fetch_resources` be the generator itself does work. I'll update both it and `fetch_metadata` so we can get rid of the `snapshot` functions.

source-pendo/source_pendo/api.py
Alex-Bair: `TypeAdapter` works, thank you!

source-pendo/source_pendo/models.py
Alex-Bair: I ended up having separate models for each of the event streams - each use a different field as a key. And no optional or default values for any models.

source-pendo/source_pendo/api.py
Alex-Bair: While digging into this, I learned that not all of the event endpoints are actually aggregates. For `pollEvents` and `guideEvents`, a document is a single, distinct event. For `pageEvents`, `featureEvents`, and `trackEvents`, a document is an aggregate of multiple events within an hour. I ended up handling events & aggregated events separately due to their differences.

For the aggregated events, I added another condition to the `filter` operation in the API request. Aggregates are included in the response if its `hour` (the bottom of the hour, like 6:00 if the aggregate was updated at 6:15) is less than the bottom of the current hour. So if the current time is 6:30, we won't get aggregates for the ongoing 6:00 hour, but we will for 5:00, 4:00, etc.

I also noticed that Pendo's API isn't updated in real-time; it took ~30 minutes for the past hour's aggregates to show up in API calls. So this strategy might not be _completely_ necessary right now, but it is the right way to do it & covers us in case the API becomes closer to real-time.

estuary-cdk/estuary_cdk/http.py
```diff
@@ -237,7 +238,9 @@ async def _request_stream(
             await asyncio.sleep(cur_delay)
 
             headers = {}
-            if _with_token and self.token_source is not None:
+            if self.auth_headers:
```
williamhbaker, line 241: I had made a comment here, but I think it got lost somehow (I must have forgot to add it): I think we should modify `TokenSource` to have an optional `authorization_header` parameter that defaults to `Authorization`, but could allow for an alternate header to be used in cases like this. That would be more consistent with the usage of `TokenSource`.

source-pendo/source_pendo/api.py
```diff
@@ -0,0 +1,185 @@
...
+                        }
+                    }
+                },
+                # Only return aggregates that have been updated since the last aggregate 
+                # we've seen AND aren't "in-progress" during the current hour.
+                {
+                    "filter": f"lastTime > {last_ts} && hour < {current_hour}"
```
williamhbaker, line 149: I don't think this is 100% bullet-proof, and it's because it involves comparisons between distributed clocks. In an extreme case, imagine the connector's system clock was 5 minutes fast. At 2:01 it could think it was fine to start getting results for the 1:00 hour, but the actual time is 1:56 and that would get an incomplete result for that hour.

Would it work to just always ignore the most recent aggregation result, and only output documents for the "N-1"'th aggregates? This would cause additional delay in outputting a document for the "latest" hour, but considering how this API seems to behave I don't think that would be a huge issue.

source-pendo/source_pendo/models.py
```diff
@@ -0,0 +1,116 @@
...
+class Metadata(BaseDocument, extra="allow"):
+    pass
+
+
+# Event streams have slightly different required fields, so we have models for each event type.
+class PageEvent(BaseDocument, extra="allow"):
+    appId: int
+    hour: int
+    remoteIp: str
+    lastTime: int
```
williamhbaker, line 48: We should have these be actual datetime strings rather than millisecond ints. That's more useful for the consumer, but will complicate the connector code a little bit. For reference, checkout of how `EmailEvent` works in source-hobspot-native, [here](https://github.com/estuary/connectors/blob/da81a628064499560ebca50e828976f89650e3aa/source-hubspot-native/source_hubspot_native/models.py#L373) and [here](https://github.com/estuary/connectors/blob/da81a628064499560ebca50e828976f89650e3aa/source-hubspot-native/source_hubspot_native/api.py#L513-L518) in particular.

source-pendo/source_pendo/resources.py
```diff
@@ -0,0 +1,222 @@
...
+        common.Resource(
+            name=resource_name,
+            key=["/appId", "/guideTimestamp", "/remoteIp", identifying_field],
+            model=model,
+            open=functools.partial(open, entity, model),
+            initial_state=ResourceState(
+                inc=ResourceState.Incremental(cursor=to_millisecond_timestamp(config.startDate))
+            ),
+            initial_config=ResourceConfig(
+                name=resource_name, interval=timedelta(minutes=5)
```
williamhbaker, line 157: Are there any API limits / concerns to be worried about here? If we can get individual events in near real-time, we could leave this as a 0 interval to try to get them as fast as possible.

Alex-Bair: No documented or observed API limits or other concerns that would prevent us from using a 0 interval here. I'll update it.

source-pendo/source_pendo/models.py
Alex-Bair: Thanks for the references, those were helpful! I've changed these int timestamps over to datetime strings. It's interesting how Pydantic handles converting the millisecond int properties in the response to datetime strings with just the `model_validate_json` method.

source-pendo/source_pendo/api.py
Alex-Bair: I think I understand the issue. So in that extreme case, it's possible we emit aggregate documents more than once per hour for a specific aggregate.  Since these streams are for multiple different aggregate, IP address, and page/feature/track combos, not outputting the "N-1"th document won't address this issue. To solve this, we'd have to keep track of each unique aggregate combo we see in an hour & avoid emitting the N-1 doc for each of those combos. 

That's really tedious/difficult to implement. Since this is unlikely to be an issue, I just added a 15 minute delay when we check for the previous hour's aggregates - ex: at 2:15 we check for the 1:00 hour aggregates.

williamhbaker (approved):
LGTM

PR #1870, opened 2024-08-28
source-mysql: Reduce binlog event cache from 10240 to 256

willdonnelly:
**Description:**

It turns out that the MySQL replication client library we use has an internal event buffer in addition to the one we maintain. We try to keep buffers small so that we never run out of memory, so this commit reduces the size of that buffer from 10,240 events to 256 events.

This should put us at a significantly smaller risk of OOMing, and should still be plenty high enough to keep everything running smoothly without unnecessary blocking on send/receives.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1870)

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -24,11 +24,17 @@ import (
 	"vitess.io/vitess/go/vt/sqlparser"
 )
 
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

Issue #1861, opened 2024-08-27
source-hubspot-native: address inconsistencies


jgraettinger:
HubSpot APIs are inconsistent (not monotonically increasing with respect to timestamps) when requesting very recent data. Because of these we will miss documents if trying to read these as a stream.

A potential solution is to add a second "delayed" stream of documents that is held back by some conservative time horizon, but is 100% correct.

PR #1871, opened 2024-08-28
source-hubspot-native: read a delayed stream of documents in addition to the realtime stream

williamhbaker:
**Description:**

Adds a secondary stream of documents for reading "delayed" changes that is assured to get all of the data that will run occasionally in tandem with the "real time" document reading which we know produces inconsistent results.

In many cases this delayed stream can use the v3 Search API, but not all. And even when using the Search API, there are annoying nuances for some object types. Overall this tends to push the code into a more gnarly structure than I'd like, but it's as good as I could reasonably get it for now and I _hope_ it is fairly easy to extend with additional object types in the future.

Closes https://github.com/estuary/connectors/issues/1861

**Workflow steps:**

**Notes for reviewers:**


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -64,7 +64,7 @@ async def fetch_properties(
     return properties_cache[object_name]
 
 
-async def fetch_page(
+async def fetch_page_with_assocations(
```
mdibaiee, line 67: typo
```suggestion
async def fetch_page_with_associations(
```

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -216,137 +216,404 @@ async def fetch_batch_with_associations(
...
-            # we will capture some of the same records twice.
-            fetch_recent = functools.partial(fetch_recent_search_objects, object_name, min(r[0] for r in recent))
-
     recent.sort()  # Oldest updates first.
 
     for batch_it in itertools.batched(recent, 50):
         batch = list(batch_it)
 
+        # Enable lookup of datetimes for IDs from the result batch.
+        dts = {r[1]: r[0] for r in batch}
```
mdibaiee, line 414: I think this can be rewritten as below for better readability
```suggestion
        dts = {id: dt for dt, id in batch}
```

mdibaiee (approved):
LGTM, a couple of nits above

Issue #1426, opened 2024-03-26
source-mixpanel: import to Estuary python CDK


williamhbaker:

williamhbaker:
This is currently noted as being blocked because the API is a paid feature. @Luishfs do you know anything more about that?

Luishfs:
Yes, when running tests (initial snapshot) on this connector on a recent created account, i receive a error claiming i need a paid account to actually use those endpoints.

[They claim](https://mixpanel.com/pricing/) that **Raw data export API** is free, but formatted is not, and it seems the connector runs on Formatted

PR #1873, opened 2024-08-29
source-mixpanel-native: various stream improvements

Alex-Bair:
**Description:**

The imported `source-mixpanel-native` connector queries different endpoints depending on what credentials are provided. Each endpoint puts slightly different properties in the response, so making the schemas less restrictive should help avoid schema violations.

The `cohort_members` stream was also checking for members of empty cohorts. This would essentially waste an API query, which really hurts due to Mixpanel's API limits. The logic has been updated to only check for members of non-empty cohorts.

Using a large date window for the `export` stream causes OOM issues since there's so much data returned for a single day. We now only use a maximum of 30 days for this stream to avoid OOMs.

Links to issue #1426 

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1873)


jonwihl (approved):
LGTM

PR #1875, opened 2024-08-30
build: run on ubuntu-24.04

williamhbaker:
**Description:**

For compatibility with the built Flow artifacts.

Needed to fix connector builds, which rely on `flowctl` build from the Flow repo master branch, among other things.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1875)


materialize-sqlserver/docker-compose.yaml
```diff
@@ -2,19 +2,19 @@ version: "3.7"
 
 services:
   sqlserver:
-    image: 'mcr.microsoft.com/mssql/server:2017-latest'
+    image: 'mcr.microsoft.com/mssql/server:2022-latest'
```
williamhbaker, line 5: Couldn't get the 2017 version to work on the newer ubuntu runner. The snapshot change is because this version uses `varchar` instead of `nvarchar` by default.

tests/materialize/materialize-databricks/snapshot.json
```diff
@@ -598,7 +598,7 @@
     {
       "flow_published_at": "1970-01-01T00:00:28Z",
       "id": 1,
-      "unsigned_bigint": 18446744073709552000
+      "unsigned_bigint": 18446744073709551615
```
jgraettinger, line 601: is this meaningful ?

jgraettinger (approved):
LGTM

tests/materialize/materialize-databricks/snapshot.json
williamhbaker: I don't totally understand why it was the way it was before. The value for the test fixture is `18446744073709551615` so getting `18446744073709551615` back seems accurate. For some reason it was getting approximated on the older runner. I have always gotten `18446744073709551615` back when running these tests locally (with the same images) so this seems like an improvement.

Some tests (bigquery for example) have always had `18446744073709551615` instead of the approximated version in the snapshot, so I am pretty confident it is just an artifact of how we are extracting the data from the systems for snapshotting, and not something related to the connector.

Issue #1859, opened 2024-08-27
source-mysql: support for replica-only capture


willdonnelly:

willdonnelly:
The main PR (https://github.com/estuary/connectors/pull/1878) addressing this issue is now out for review. Once that's released, we will need to:

- Update various connector documentation to remove all mention of the watermarks table and possibly also call out read-replica captures as a thing we explicitly support.
- Wait a few days to make sure everything is running happily in production, and then finish removing the deprecated `watermarks_table` and `heartbeat_interval` config properties.

willdonnelly:
PR #1878 has been live for a few hours and nothing has exploded so far, which means that MySQL read replica captures are now something that works in production. I'll get the aforementioned documentation updates out soon and then next week I'll finish the job of removing the deprecated config properties.

PR #1878, opened 2024-08-30
source-mysql: Support captures from read-only replicas

willdonnelly:
**Description:**

This PR replaces the watermarking logic previously used by `source-mysql` with a new implementation based solely on binlog positions. This in turn means that the capture can now successfully run against a read-only replica, where previously it had to run against a writable database in order to issue watermark upserts.

This is accomplished in essentially three steps:
1. A major refactoring of the interface between the generic SQL CDC machinery and the database-specific logic, which pushes down the details of watermarking into the database-specific code while renaming the same basic idea to a more generic "fences".
2. A second major refactoring of the SQL CDC interface, which replaces the channel-based event stream interface with a synchronous callback passed into a new `StreamToFence` method. As part of this the `WriteWatermark` / `RequestFence` method is removed and built into the streaming primitive instead. In practice the only ways we ever used this before were either "write a watermark and then stream to it" or "wait a bit, then write a watermark and stream to it" so the new interface more closely matches the actual usage pattern and thus is probably preferable anyway.
3. Replacing the MySQL `StreamToFence` implementation (which after the first two refactorings now handles the entire process of writing a watermark and then processing change events until that watermark is observed) with a new one based on binlog offsets. This step was nontrivial because there are a lot of possible edge cases given that we need separate "time-based" and "fence-pased" streaming phases plus a fast-path so that we don't block forever on an idle database, but I believe that the new implementation is finally correct in all of those edge cases.

In addition to those core changes, the MySQL connector has been modified to remove almost all prerequisite validation and discovery logic related to the watermarks table (with the sole exception of the code that filters out the watermarks table from discovered catalogs, because preexisting users will still have one and we don't want to suddenly start discovering it as a table to be captured) and to change the descriptions of the `watermarks_table` and `heartbeat_interval` config properties but still allow them to be present. I intend to clean up those last vestiges, but only after this PR has been released and working properly in production for a while.

As we previously discussed over Slack, I believe that there is no value in keeping the old watermark-based implementation around as an option for MySQL so it's been removed in its entirety. The one negative side-effect of this is that any users who were relying on a regular cadence of "heartbeat" writes generated by our connector for their own _non-Flow_ purposes will no longer have these. For the hypothetical user who is broken by this, the MySQL Event Scheduler can be used as a drop-in replacement with something like the following statement:

```sql
CREATE EVENT IF NOT EXISTS flow.watermarks_heartbeat
  ON SCHEDULE EVERY 1 MINUTE
  COMMENT 'Generates writes to the Flow watermarks table once per minute'
  DO REPLACE INTO flow.watermarks (slot, watermark)
           VALUES (0, CONCAT('Heartbeat: ', UTC_TIMESTAMP(), ' UTC'));
```

The one other setup worth noting is captures from tables using the MyISAM storage engine. Because MyISAM is non-transactional it doesn't write "XID Events" into the binlog on commit. We require database commit events to drive the Flow transaction checkpointing logic, so these tables previously behaved suboptimally: the row change events would be captured correctly, but not actually committed into Flow until a change occurred on an InnoDB table. Watermark writes to a watermarks table explicitly created on InnoDB previously served as the escape hatch for this configuration.

Since we're removing watermark writes it was necessary to solve this another way, so this PR also contains proper support for capturing from non-transactional MyISAM tables. With this support added, our entire test suite still passes after being modified to create all test tables on the MyISAM storage engine, so I believe this feature should be reasonably robust.

This PR addresses the bulk of https://github.com/estuary/connectors/issues/1859 but there will be a bit of cleanup work around the advanced config settings to tackle in a week or two once we're sure this is working nicely in production.

**Workflow steps:**

Existing users shouldn't have to do anything. The goal is really for this all to just work without any fuss.

Obviously, new users will no longer need to set up a `flow` schema or `flow.watermarks` table, grant the capture user access to that table, or grant the capture user sufficient permissions to set it up for them.

More importantly, new users will now be able to provide the address of a read-only replica of their database. This replica will still need binary logging enabled and an appropriate capture user needs to exist and have the necessary replication and backfill permissions, but otherwise this configuration should just work.

We will need to update the connector reference docs for `source-mysql` and variants to remove all mention of watermarks setup requirements:

- https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/
- https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/amazon-rds-mysql/
- https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/google-cloud-sql-mysql/

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1878)

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -119,6 +119,7 @@ func (db *mysqlDatabase) ReplicationStream(ctx context.Context, startCursor stri
 	stream.tables.discovery = make(map[string]sqlcapture.DiscoveryInfo)
 	stream.tables.metadata = make(map[string]*mysqlTableMetadata)
 	stream.tables.keyColumns = make(map[string][]string)
+	stream.tables.nonTransactional = make(map[string]bool)
 	return stream, nil
 }
 
@@ -163,16 +164,21 @@ type mysqlReplicationStream struct {
 	gtidTimestamp time.Time // The OriginalCommitTimestamp value of the last GTID Event
 	gtidString    string    // The GTID value of the last GTID event, formatted as a "<uuid>:<counter>" string.
 
+	// A count of how many changes have been made on non-transactional tables since
+	// the last commit was processed.
+	nonTransactionalChanges int
+
 	// The active tables set and associated metadata, guarded by a
 	// mutex so it can be modified from the main goroutine while it's
 	// read from the replication goroutine.
 	tables struct {
 		sync.RWMutex
-		active        map[string]struct{}
-		discovery     map[string]sqlcapture.DiscoveryInfo
-		metadata      map[string]*mysqlTableMetadata
-		keyColumns    map[string][]string
-		dirtyMetadata []string
+		active           map[sqlcapture.StreamID]struct{}
+		discovery        map[sqlcapture.StreamID]sqlcapture.DiscoveryInfo
+		metadata         map[sqlcapture.StreamID]*mysqlTableMetadata
+		keyColumns       map[sqlcapture.StreamID][]string
+		nonTransactional map[sqlcapture.StreamID]bool
+		dirtyMetadata    []sqlcapture.StreamID
 	}
 }
 
@@ -255,6 +261,8 @@ func (rs *mysqlReplicationStream) run(ctx context.Context, startCursor mysql.Pos
 				return fmt.Errorf("unknown key columns for stream %q", streamID)
 			}
 
+			var nonTransactionalTable = rs.isNonTransactional(streamID)
+
 			switch event.Header.EventType {
 			case replication.WRITE_ROWS_EVENTv1, replication.WRITE_ROWS_EVENTv2:
 				for rowIdx, row := range data.Rows {
@@ -283,6 +291,9 @@ func (rs *mysqlReplicationStream) run(ctx context.Context, startCursor mysql.Pos
 						Source:    sourceInfo,
 					}); err != nil {
 						return err
+					}
+					if nonTransactionalTable {
+						rs.nonTransactionalChanges++ // Keep a count of uncommitted non-transactional changes
 					}
 				}
 			case replication.UPDATE_ROWS_EVENTv1, replication.UPDATE_ROWS_EVENTv2:
@@ -324,6 +335,9 @@ func (rs *mysqlReplicationStream) run(ctx context.Context, startCursor mysql.Pos
 						}); err != nil {
 							return err
 						}
+						if nonTransactionalTable {
+							rs.nonTransactionalChanges++ // Keep a count of uncommitted non-transactional changes
+						}
 					}
 				}
 			case replication.DELETE_ROWS_EVENTv1, replication.DELETE_ROWS_EVENTv2:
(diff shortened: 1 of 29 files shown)


source-mysql/replication.go
```diff
@@ -387,7 +405,18 @@ func (rs *mysqlReplicationStream) run(ctx context.Context, startCursor mysql.Pos
 		case *replication.PreviousGTIDsEvent:
 			logrus.WithField("gtids", data.GTIDSets).Trace("PreviousGTIDs Event")
 		case *replication.QueryEvent:
-			if err := rs.handleQuery(ctx, string(data.Schema), string(data.Query)); err != nil {
+			if string(data.Query) == "COMMIT" {
```
mdibaiee: Do we need to handle possible `ROLLBACK` queries?

This line from MySQL docs got me worried:

> Every transaction (including [autocommit](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_autocommit) transactions) is recorded in the binary log as though it starts with a [BEGIN](https://dev.mysql.com/doc/refman/5.7/en/commit.html) statement, and ends with either a [COMMIT](https://dev.mysql.com/doc/refman/5.7/en/commit.html) or a [ROLLBACK](https://dev.mysql.com/doc/refman/5.7/en/commit.html) statement. This is even true for statements affecting tables that use a nontransactional storage engine (such as [MyISAM](https://dev.mysql.com/doc/refman/5.7/en/myisam-storage-engine.html)).

https://dev.mysql.com/doc/refman/5.7/en/replication-features-transactions.html

mdibaiee (commented):
Thanks Will! One main question regarding `ROLLBACK` for MyISAM and a small request for clarification on something

source-mysql/replication.go
```diff
@@ -345,7 +324,7 @@ func (rs *mysqlReplicationStream) run(ctx context.Context) error {
 						}
 						var sourceInfo = &mysqlSourceInfo{
 							SourceCommon: sourceCommon,
-							EventCursor:  fmt.Sprintf("%s:%d:%d", rs.cursor.Name, rs.cursor.Pos, rowIdx/2),
+							EventCursor:  fmt.Sprintf("%s:%d:%d", cursor.Name, cursor.Pos, rowIdx/2),
```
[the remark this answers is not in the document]
willdonnelly: The row images of an update event come in pairs: before-image, then after-image. Up at the top of this loop we have an `if rowIdx%2 == 1` check so that we only process each complete update / pair of row images once, so dividing by two here just makes things look slightly nicer.

source-mysql/replication.go
willdonnelly: There is no such thing as a `ROLLBACK` query modifying a MyISAM table for the same reason there's no such thing as transactional changes to a MyISAM table.

I can see how the phrasing in the docs might be confusing. A clearer way of restating what they're saying might be:
1. Every change is recorded in the binlog as a full transaction and ends with either a commit or a rollback.
2. Even when you're using MyISAM, each change operation is recorded as though it ends in a commit even though MyISAM doesn't have any concept of commits or rollbacks.

mdibaiee (approved):
LGTM

PR #1886, opened 2024-09-03
materialize-s3-iceberg: add REST catalog support

williamhbaker:
**Description:**

The main change here is adding support for the Iceberg REST catalog. I haven't been able to get ahold of a "hosted" account like Snowflake's Polaris catalog to test with, but this works with self-hosted catalogs like the [hansetag/iceberg-catalog](https://github.com/hansetag/iceberg-catalog) that the integration tests have been updated to run against. 

Also included is an enhancement/fix to how checkpoint tracking works for individual tables. The previous strategy of using a property of the latest snapshot was flawed; this changes to using a table property instead. And the table and file location logic has been improved to use a more standard location for data files, which is something that has been requested by users.

The version is being bumped to v2 since these changes a technically incompatible with previously existing tasks. 

Issue https://github.com/estuary/connectors/issues/1885

**Workflow steps:**

The materialization docs will need to be updated to reflect these changes.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1886)


mdibaiee (approved):
LGTM

PR #1887, opened 2024-09-03
source-mixpanel-native: Remove state checkpoint property for `export`

Alex-Bair:
During the connector import, the [`state_checkpoint_interval` property](https://docs.airbyte.com/connector-development/cdk-python/incremental-stream#interval-based-checkpointing) was added to the [`export` stream](https://developer.mixpanel.com/reference/raw-event-export), which checkpoints state after every N documents. This assumes that documents are in ascending order of the cursor field. However, `export` does not return documents in ascending order, so this strategy does not work. I also suspect it interfers with `get_updated_state` - the other older way to checkpoint state. This could explain why the `export` stream for an existing task is missing documents.

Also, sometimes the API returns an `export` document with a `insert_id�` property instead of `insert_id`. Logic to strip off the extra bytes was added.

**Description:**

**Workflow steps:**

**Notes for reviewers:**

Tested locally with a smaller date range (~MAR2020 - AUG2020). A significant number of `insert_id`s (the primary key for `export`) were present in the `export` collection that were not there before.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1887)

### DIFF
--- a/source-mixpanel-native/source_mixpanel_native/streams/base.py
+++ b/source-mixpanel-native/source_mixpanel_native/streams/base.py
@@ -27,13 +27,6 @@ class MixpanelStream(HttpStream, ABC):
     DEFAULT_REQS_PER_HOUR_LIMIT = 60
 
     @property
-    def state_checkpoint_interval(self) -> int:
-        # to meet the requirement of emitting state at least once per 15 minutes,
-        # we assume there's at least 1 record per request returned. Given that each request is followed by a 60 seconds sleep
-        # we'll have to emit state every 15 records
-        return 10000
-
-    @property
     def url_base(self):
         prefix = "eu." if self.region == "EU" else ""
         return f"https://{prefix}mixpanel.com/api/2.0/"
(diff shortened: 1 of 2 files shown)


source-mixpanel-native/source_mixpanel_native/streams/base.py
```diff
@@ -26,13 +26,6 @@ class MixpanelStream(HttpStream, ABC):
 
     DEFAULT_REQS_PER_HOUR_LIMIT = 60
 
-    @property
```
williamhbaker, line 29: Just noting that this will impact all the mixpanel streams and not just the Exports one - but I assume that if anything that's probably a good thing, since other APIs may not have the sorting properties needed for it to work correctly.

williamhbaker (approved):

source-mixpanel-native/source_mixpanel_native/streams/base.py
Alex-Bair: Some of the other streams (`engage`, `cohort_members`, `funnels`) are overriding the `state_checkpoint_interval` from their parent, so they shouldn't be affected by this change. 

The streams that no longer have a `state_checkpoint_interval` property are `annotations` and `cohorts`. Both of these only work as full refresh, so they shouldn't have `state_checkpoint_interval` anyway.

PR #1891, opened 2024-09-04
materialize-databricks: batch merge queries into smaller sections

mdibaiee:
**Description:**

- Tested by running integration tests, and by reducing the staged file size threshold and running a task with that configuration. With 128 files per merge query, the length of the query string itself is small enough to fit databricks limits, but it means we may process up to ~16GB of data in a single query, which I also think should be okay.
- I didn't apply the same batching on load queries, since for load queries, if we consider each ID to be 36 bytes (that's a uuid, I think this is one of the larger common key types, whereas most keys will be ~8 bytes as int64 or similar), we  can fit ~3M ids in a single staged file for load phase, and before we hit 128 files to batch up the query, we will have ~477M ids in those 128 files.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1891)

### DIFF
--- a/materialize-databricks/driver.go
+++ b/materialize-databricks/driver.go
@@ -326,7 +326,8 @@ func (d *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage) 
 }
 
 type checkpointItem struct {
-	Query    string
+	Query    string   `json:",omitempty"` // deprecated, kept for backward compatibility
+	Queries  []string `json:",omitempty"`
 	ToDelete []string
 }
 
@@ -346,6 +347,8 @@ func (d *transactor) deleteFiles(ctx context.Context, files []string) {
 		}
 	}
 }
+
+const queryBatchSize = 128
 
 func (d *transactor) Store(it *m.StoreIterator) (_ m.StartCommitFunc, err error) {
 	ctx := it.Context()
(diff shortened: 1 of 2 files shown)


materialize-databricks/driver.go
```diff
@@ -395,24 +398,41 @@ func (d *transactor) Store(it *m.StoreIterator) (_ m.StartCommitFunc, err error)
...
+		var queries []string
 		// In case of delta updates or if there are no existing keys being stored
 		// we directly copy from staged files into the target table. Note that this is retriable
 		// given that COPY INTO is idempotent by default: files that have already been loaded into a table will
 		// not be loaded again
 		// see https://docs.databricks.com/en/sql/language-manual/delta-copy-into.html
 		if b.target.DeltaUpdates || !b.needsMerge {
-			if query, err = RenderTableWithFiles(b.target, toCopy, b.rootStagingPath, tplCopyIntoDirect); err != nil {
-				return nil, fmt.Errorf("copyIntoDirect template: %w", err)
+			for i := 0; i < len(toCopy); i += queryBatchSize {
```
williamhbaker, line 408: Would be a nice place to use [slices.Chunk](https://pkg.go.dev/slices#Chunk), but that's not available until go1.23 which we haven't upgrade to yet, but maybe something to refactor when do we.

williamhbaker (approved):
LGTM

materialize-databricks/driver.go
mdibaiee: That's nice, I did look into `slices` but didn't see Chunk. I'll add a comment so we do that later

PR #1893, opened 2024-09-04
source-mixpanel-native: `export` stream enhancements

Alex-Bair:
**Description:**

In the `export` stream, events can have either a lowercased `id` property or a capitalized `Id` property. These properties represent the same thing, and having different cases can cause issues when materializing an `export` collection, like having multiple columns with IDs split between them instead of a single column with all IDs. The check for extra bytes at the end of the `insert_id` property has been updated to be very slightly more performant with string slicing instead of using a regex match.

Since a single `export` request can take a while to process during backfills, the sleep logic in `base.py` to avoid the API rate limit has been updated to only sleep if processing the request took less than 60 seconds.

Some of the unused code leftover from the import was removed. I mainly focused on removing the low-hanging fruit - stuff I quickly & easily identified as unused.

**Workflow steps:**

**Notes for reviewers:**

Tested locally with a small date range of a couple months. Confirmed streams still work & the timer logic works as intended.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1893)


williamhbaker (approved):
LGTM

PR #1895, opened 2024-09-05
source-zendesk-support: fix `ticket_metrics` and `ticket_metric_events`

Alex-Bair:
**Description:**

Tasks have had issues with `ticket_metrics` and `ticket_metric_events` taking a very long time to backfill. Fixes are detailed below.

## `ticket_metrics`
The endpoint we were using before wasn't appropriate for capturing changes to ticket metrics; it was returning metrics in descending order of when the associated ticket was created, _not_ when the ticket was updated. 

To fix this, the stream was changed to use the same incremental logic & endpoint as the `tickets` stream, and sideload in the metrics for each ticket. This gives us all the benefits of the recent `tickets` improvements (emitting state every 1k documents, good cursor-based incremental sync).

## `ticket_metric_events`
This stream functions correctly & behaves incrementally, but it's extremely dense and takes a long time to complete the initial backfill. 

The number of documents per page was increased from 100 to 1,000, so that should help speed up backfills some. Logging was also added after each successfully parsed response to show how far along the stream has caught up.

**Workflow steps:**

Zendesk Support docs need updated to reflect the new cursor field for `ticket_metrics`.

**Notes for reviewers:**

Tested on a local stack with a number of Zendesk accounts. Confirmed streams completed & were incremental after the initial backfill.

Existing tasks `ticket_metrics` binding will need backfilled very soon after the PR is merged due to the `ticket_metrics` cursor field change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1895)

### DIFF
--- a/source-zendesk-support/source_zendesk_support/streams.py
+++ b/source-zendesk-support/source_zendesk_support/streams.py
@@ -805,8 +805,27 @@ class TicketForms(SourceZendeskSupportCursorPaginationStream):
         return {}
 
 
-class TicketMetrics(SourceZendeskSupportCursorPaginationStream):
-    """TicketMetric stream: https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_metrics/"""
+class TicketMetrics(Tickets):
+    """TicketMetrics stream: https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_metrics
+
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

PR #1896, opened 2024-09-06
tests: use flowctl raw discover instead of flowctl-go in capture tests

williamhbaker:
**Description:**

~~Most captures already test discovery as part of their unit testing, and
`flowctl-go` doesn't work for this anymore. `flowctl raw discover` doesn't work
quite the same way and I couldn't come up with a very satisfying way to replace
it with that.~~

~~Eventually we should rework these tests to use `flowctl preview` along with
`flowctl raw discover`. I've made a couple of attempts at that and have found it
to be more of an effort that I want to take on right now, so this does the
simple thing and just disables the non-working `flowctl-go api discover`.~~

Edit: Actually ended up getting this working with `flowctl raw discover`; see updated commit description.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1896)

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -77,19 +77,6 @@ DATA_PLANE_PID=$!
 # Arrange to stop the data plane on exit.
 trap "kill -s SIGTERM ${DATA_PLANE_PID} && wait ${DATA_PLANE_PID}" EXIT
 
-# Get the spec from the connector and ensure it's valid json.
-cat >"$TESTDIR/spec.yaml" <<EOF
-captures:
-  tests/${CONNECTOR}/from-source:
-    endpoint:
-      connector:
-        image: "${CONNECTOR_IMAGE}"
-        config: {}
-    bindings: []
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

Do you mind documenting slightly more detail on the impedance mismatch with `flowctl raw discover`? Are there changes to its behavior you'd suggest?

williamhbaker:
> Do you mind documenting slightly more detail on the impedance mismatch with `flowctl raw discover`? Are there changes to its behavior you'd suggest?

I wouldn't suggest any changes in behavior, and it was really just my inability to get `flowctl raw discover` dropped in and work the same way as our cumbersome test scripts previously used `flowctl-go`.

But against my better judgement, I've continued to tinker with this and have finally gotten it working. I pushed the fix and updated the PR title/description to reflect that.

PR #1899, opened 2024-09-06
source-postgres: Suport MAC addresses as primary keys

willdonnelly:
**Description:**

Previously MAC addresses as primary keys didn't work. Or possibly this was broken as part of the PGX v5 upgrade and we didn't notice because we didn't have a test case for MAC address PKs.

Either way, this PR adds a couple of test cases and the necessary FDB key serialization special case to make them work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1899)


williamhbaker (approved):
LGTM

PR #1900, opened 2024-09-07
source-mysql: No flushing TableMapEvents

willdonnelly:
**Description:**

These events are paired with one or more immediately-following row changes. Checkpointing after the TableMapEvent means that a capture could potentially restart in between those events and be unable to decode the row changes due to missing context.

This may have happened in production, but I haven't proven this was the issue -- it's just definitely wrong regardless.


jgraettinger (approved):
LGTM

PR #1903, opened 2024-09-09
sqlcapture: Implement an ugly but functional replication cursor replacement mechanism

willdonnelly:
**Description:**

As described in the comment, this is an ugly but functional solution to the problem of "sometimes it would be really useful to just adjust the replication cursor value instead of backfilling an entire multi-terabyte dataset".

Replacements are scoped to only take effect for a capture task with a specific name and specific cursor value, so unless I've made a very dumb error in the implementation this should have zero impact on any tasks other than the one we're trying to fix. The task name is hashed so that entries in this table don't leak any information about the customer. In theory we might want to obscure the binlog filename as well, but in practice everyone seems to have a default of `binlog` or `mysql-bin` and we're adding entries manually so it's not actually a big deal for now.

**Workflow steps:**

Hashing the task name with `echo -n "acmeCo/whatever/source-mysql" | sha256sum` ought to work, but also this PR adds a debug log containing the necessary hash value so that should be what we use when adding future entries. Given that hash, add a new replacement table entry for that hash which maps the old cursor value to the desired new value, and then upload and merge the PR. Once the new CI build goes out the replacement will take effect and fix the incorrect cursor position.

**Notes for reviewers:**

I've started with a replacement entry for one task in production that we would really like to fix. Assuming this works I'll have a followup PR with a couple more, and then a final PR later this afternoon which removes all of the non-example entries.

While it would be nice if this worked on the first try, I'm more concerned about not accidentally breaking things even worse, so if you see any way this could possibly cause unrelated captures to fail or for the cursor replacement to take effect on an unintended task please bring it up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1903)

### DIFF
--- a/sqlcapture/main.go
+++ b/sqlcapture/main.go
@@ -2,6 +2,7 @@ package sqlcapture
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #1904, opened 2024-09-09
sqlcapture: Two more production cursor replacements

willdonnelly:
**Description:**

And deletes the previous one now that it's taken effect on the intended capture.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1904)

PR #1905, opened 2024-09-09
sqlcapture: Remove cursor replacements

willdonnelly:
**Description:**

Added in https://github.com/estuary/connectors/pull/1904, these have had their intended effect in production and it's good practice to clean them up immediately now that they're no longer needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1905)


williamhbaker (approved):
LGTM

Issue #1909, opened 2024-09-10
source-mysql: unexplained delays in reading some binlogs


willdonnelly:
The PR (https://github.com/estuary/connectors/pull/1878) which made watermark-less CDC from MySQL possible added a new watchdog timeout to the `StreamToFence` operation, such that if we're actively streaming to an established fence position _and_ we receive no change events for over 5 minutes, the operation errors out because something is seriously wrong.

We're seeing this check trigger in at least one production task. This issue is for debugging that and fixing any bugs that might be uncovered in the process.

willdonnelly:
One obvious thought which comes to mind is that the "replication idle" timeout is reset by change events at a point _downstream_ of where we filter out changes on tables which aren't being captured. So it's theoretically possible that a sufficiently long run of changes to a non-captured table could cause an arbitrarily long delay as we work through them, and if this delay is >5m then we'd hit the timeout, fail the task, and then enter a failure loop because we didn't advance the cursor so it keeps happening each time the task restarts.

In practice this should be fairly rare because commit events always produce checkpoints, it's just the row change events we're filtering out here. So that means that the 5 minutes worth of data has to be in a single transaction, and a transaction with more data than we can consume in 5 minutes is quite large (unless, hypothetically, we were constrained by DB->Connector bandwidth...). But it's definitely a theoretical possibility.

PR #1910, opened 2024-09-10
source-mysql: Send a 'Keepalive Event' for inactive table changes

willdonnelly:
**Description:**

This introduces a new replication event type `KeepaliveEvent` which is effectively a no-op which non-change non-commit WAL events can be translated to, and modifies the MySQL replication logic to use that when filtering out changes from non-captured tables.

In theory I could have kept this entirely within the MySQL code, but I've elected to make it part of the generic machinery since we'll need something very similar if/when we implement read-only replica support for Postgres or SQL Server, and also because in connectors which send a `KeepaliveEvent` this finally resolves the annoying thing where the `processed replication events` logging only counted changes from active bindings.

This is likely to fix https://github.com/estuary/connectors/issues/1909 (if my theory about massive transactions on non-captured tables is correct). Even if it doesn't fix that issue this is definitely still the fix for that theoretical edge case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1910)

### DIFF
--- a/source-mysql/replication.go
+++ b/source-mysql/replication.go
@@ -241,8 +241,14 @@ func (rs *mysqlReplicationStream) run(ctx context.Context, startCursor mysql.Pos
 		case *replication.RowsEvent:
 			var schema, table = string(data.Table.Schema), string(data.Table.Table)
 			var streamID = sqlcapture.JoinStreamID(schema, table)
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

Issue #1923, opened 2024-09-11
source-mysql-batch: Failures when trying to capture from SingleStore


willdonnelly:
SingleStore is "MySQL wire protocol compatible" so in theory we should be able to capture from it. In practice it's failing with an error message saying:

```
unable to connect to database: <nil>
```

I suspect there are at least two issues here: one with the error reporting and one which is actually causing it to fail.

willdonnelly:
I've confirmed that there are in fact three errors going on here:

- Incorrect variable shadowing in the error handling resulting in <nil> where there should have been a useful error message.
- If connecting with SSL fails we retry without SSL, which in this case fails because SSL is required. This means that the original error is yet again hidden from us.
- The actual original error in this case  appears to be `unable to connect to database: handleAuthResult: readAuthResult: ERROR 1046 (3D000): In order to use shared tier you need to select a database by default`

willdonnelly:
We've confirmed that the real "you need to select a database" error can be addressed by setting the `/advanced/dbname` property when connecting to SingleStore. Everything else is just some error reporting jank which I've got a PR out to fix.

PR #1924, opened 2024-09-11
source-mysql-batch: Improve connection error reporting

willdonnelly:
**Description:**

Restructures some of the error reporting around MySQL database connection attempts. Previously there were two issues:

- Error messages sometimes got swallowed and reported as `unable to connect to database: <nil>` because of incorrect variable shadowing.
- We try to connect with both TLS and not-TLS, but in some cases the server will require TLS and the non-TLS connection attempt will fail for that reason, but we didn't report the reason the with-TLS connection attempt failed. Now we report the TLS failure reason when both fail.

Also this commit fixes the "incorrect username/password" special case since that got broken in the upgrade to v1.8.1 of the MySQL client library.

This fixes https://github.com/estuary/connectors/issues/1923

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1924)

### DIFF
--- a/source-mysql-batch/main.go
+++ b/source-mysql-batch/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

Issue #1912, opened 2024-09-10
source-postgres: JSON columns string-ified in replication events


willdonnelly:
Reports that JSON column values may be captured as strings-containing-JSON in some circumstances.

This would not be immediately caught by JSON schema validation because our schema for a JSON column is (as it should be) the catch-all schema, but that means that this could easily happen if we're not careful.

Datatype validation should catch this though, so another question is why, if this is happening, the tests are passing.

willdonnelly:
This is a weird one. I've confirmed that it's definitely happening on our own production capture from our own Supabase instance, but only for one table. The columns which are getting stringified are a non-nullable `jsonb` and a nullable `json` column, so neither JSONB/JSON or nullability appear to matter.

There is nothing especially interesting about how the columns are declared in the table definitions, and for some reason the values are correct in backfills and incorrect via replication. Most bizarrely, we have other tables with nullable `json` columns which are coming through just fine as objects.

I am attempting to reproduce the issue with a locally modified connector with additional debug logging.

willdonnelly:
I have dug into the issue and established that a `jsonb` column which is getting captured incorrectly has values identified as OID 31280 in the replicated change event, whereas the normal JSONB OID is 3802. Given that clue I've managed to reproduce basically all of the things I've observed by declaring a test table as:

```
CREATE DOMAIN UserDomain AS JSONB;
CREATE TABLE foobar (id INTEGER PRIMARY KEY, data UserDomain NOT NULL);
```

I will investigate how exactly we could handle this better, and why the values are captured correctly in backfills.

PR #1926, opened 2024-09-11
source-postgres: Register domain type OIDs as aliases

willdonnelly:
**Description:**

As explained in the added code, a user-defined domain type (which can be created via `CREATE DOMAIN name AS base_type` and which is basically just a type alias with optional constraints) ought to be decoded the same as a value of the base type.

We accomplish this by processing the `TypeMessage` event and when the namespace is omitted (which implicitly means "pg_catalog", if the Postgres docs can be trusted here) we look up the type name in the type map and register the user type OID as having the name and codec of the base type (if found).

Also adds a test case which exercises this behavior. If not for the fix in this commit, snapshot `TestCaptureDomainJSONB-capture2` would have its JSONB values emitted as strings rather than as raw JSON.

This fixes https://github.com/estuary/connectors/issues/1912

**Notes for reviewers:**

I considered making the fix a bit more narrowly-targeted to just `json` and `jsonb` aliases, but I believe that the logic as written does the right thing for domain types which alias any builtin Postgres data type, and there are plenty of other cases where we haven't registered our own custom decode function but we'd still like to use the same decode logic for these aliases as the PGX client library would use for the base type.

Since the "register as an alias" logic only kicks in when the named type is in `pg_catalog` (which, per the docs, is what an empty schema means for this message) and it does nothing other than emitting a warning if the type name isn't found when we look it up, I believe it's safe to just do this generically for all `TypeMessage` events with an empty schema.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1926)

### DIFF
--- a/source-postgres/.snapshots/TestCaptureDomainJSONB-capture1
+++ b/source-postgres/.snapshots/TestCaptureDomainJSONB-capture1
@@ -1,0 +1,10 @@
+# ================================
+# Collection "acmeCo/test/test_capturedomainjsonb_40925847": 2 Documents
+# ================================
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"capturedomainjsonb_40925847","loc":[11111111,11111111,11111111]}},"data":{"foo":"bar"},"id":1}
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

Issue #1927, opened 2024-09-11
source-sql-server: large text columns come through as NULL in deletion events


willdonnelly:
According to [SQL Server documentation](https://learn.microsoft.com/en-us/sql/relational-databases/system-tables/cdc-capture-instance-ct-transact-sql?view=sql-server-ver16#large-object-data-types):

> Columns of data type **image**, **text**, and **ntext** are always assigned a NULL value when __$operation = 1 or __$operation = 3.

Operation 1 is a delete, so what this is saying is that image/text/ntext column values are omitted in a delete event by setting that column to `NULL`. (Operation 3 is irrelevant at present, it contains update row preimages and currently the capture would halt if it ever received one)

Currently the connector emits these nulls as though they were real values, but the usual handling of omitted values is to omit the corresponding property so having an explicit null is incorrect. Usually this is a pretty benign incorrectness, but if a user has a `TEXT NOT NULL` column and deletes a row, we emit a null value which fails schema validation.

As far as I can tell, when processing a deletion event for a nullable text (or image/ntext) column there's no way for us to tell whether this `NULL` value is the real preimage of the row or an omitted value. Since we don't have enough information to tell the difference, we might have to just always omit `NULL` values of a text/image/ntext column when capturing deletions, but that would probably be fine.

PR #1929, opened 2024-09-12
source-sqlserver: Omit null text/image/ntext values on deletes

willdonnelly:
**Description:**

According to MS docs, these columns always hold null values in the CDC table for a delete op, and blindly emitting those is incorrect and can cause schema validation failures if the text column is non-nullable.

The fix is to omit the property when the value is nil, the type is text/image/ntext, and the operation is a delete. In theory
this is always fine, because if the preimage value is a real null then we already have that in the LHS document that the delete will be merged into.

We could probably be even more aggressive in omitting properties here, honestly, but there's no real motivation to do that so I have gone with the minimal fix that only impacts the datatypes for which the default behavior is definitely wrong.

This fixes https://github.com/estuary/connectors/issues/1927

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1929)

### DIFF
--- a/source-sqlserver/.snapshots/TestDeletedTextColumn-init
+++ b/source-sqlserver/.snapshots/TestDeletedTextColumn-init
@@ -1,0 +1,10 @@
+# ================================
+# Collection "acmeCo/test/test_deletedtextcolumn_41823101": 2 Documents
+# ================================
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

Issue #1930, opened 2024-09-12
source-mysql: Copy improved connection error reporting from batch MySQL


willdonnelly:
In https://github.com/estuary/connectors/pull/1924 I modified the Batch MySQL connection logic to improve error reporting and fix the "incorrect username/password" special-case so it works again. Since the MySQL CDC connector has basically the same connection logic I should copy those changes over.

PR #1932, opened 2024-09-12
source-mysql: Fix connection error reporting

willdonnelly:
**Description:**

This copies over the fixes which I made to `source-mysql-batch` connections in https://github.com/estuary/connectors/pull/1924 which made it so that when TLS and non-TLS connection attempts both fail we will report the with-TLS error, and also so that the "incorrect username/password" special case works again on the latest version of the MySQL client library.

This fixes https://github.com/estuary/connectors/issues/1930


williamhbaker (approved):
LGTM

PR #1933, opened 2024-09-12
source-mongodb: merge deletion events

williamhbaker:
**Description:**

This adds the (conditional) reduction strategy to deletion events to merge them into existing documents. For materializations using soft deletions, this will have the effect of only updating the `_meta/op` column to `"d"` for the materialized row, instead of removing all of the values from other non-key columns.

This will make soft deletions for MongoDB-captured collections work the same way as other CDC connectors, while still using the top-level "last write wins" reductions strategy for other events, most notably updates, where we don't emit explicit `null` values for removed fields (yet).

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1933)

### DIFF
--- a/source-mongodb/.snapshots/TestDiscover
+++ b/source-mongodb/.snapshots/TestDiscover
@@ -20,7 +20,7 @@ Binding 0:
       "then": {
         "reduce": {
           "delete": true,
-          "strategy": "lastWriteWins"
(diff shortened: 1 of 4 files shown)


willdonnelly (approved):
LGTM

PR #1935, opened 2024-09-13
source-impact-native: adding new connector + estuary_cdk changes

Luishfs:
**Description:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1935)


estuary-cdk/estuary_cdk/http.py
jonwihl, line 1: Whats this change? Why are we modifying the CDK?

Luishfs: The default `http.request` only handles Authorization headers, and did now allow for a custom header to be set. 
HTTPMixin already implements a async `_request_stream` method, so in order to append header data into the default `request` method, i've added the `headers` variable support.
This was required since the impact-native API only returns data in JSON format when specified in the header, but i also believe to made a similar change to a past PR in order to support headers

williamhbaker: This seems like a reasonable change, and it is even noted in the comment [here](https://github.com/estuary/connectors/blob/37150fb708c708f06fbbebb057d381a8a7498f3c/estuary-cdk/estuary_cdk/http.py#L96-L98) that such a thing may eventually be needed.

One thing for the sake of type-checking is that I don't think we'd want to accept a `None` value for headers. It seems like it should be `headers: dict[str, Any] = {}` on line 42.

Luishfs: Thanks for the review @williamhbaker !!! 
Just made this change in 2f2d7e3711a7b17a88bd8374ec5457e1ff6e1350

jonwihl (approved):
LGTM

PR #1936, opened 2024-09-13
materialize-s3-iceberg: don't use context manager for transaction

williamhbaker:
**Description:**

The transaction context manager will actually commit the transaction even if an exception is raised by any of its operations. This means that setting the table property and then attempting to add a data file where the data file addition operation fails would still update the table checkpoint, resulting in Flow transaction erroneously being acknowledged even when data wasn't added to the table.

Using the explicit transaction handle and committing it does seem to work as expected though, where if there is an exception before the `commit` none of the operations will be applied.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1936)


Alex-Bair (approved):
LGTM

PR #1942, opened 2024-09-13
source-stripe-native: query param bug fix & minor refactors

### LINKED ISSUE #1748: source-stripe-native: Fix schema issues
Connector should use minimal schemas and rely on schema inference to resolve differences. Current failure is the result of string and array values resolving to the same field

### LINKED ISSUE #1790: source-stripe-native: Customer X does not have card with ID Y
source-stripe-native connector is returning errors for Customer streams, similar to the following:

```
runTransactions: readMessage: Task capture.Cards.incremental: Encountered HTTP error status 400 which cannot be retried.
URL: https://api.stripe.com/v1/customers/cus_XXXXXXXXXXXXX
Response:
{
  "error": {
    "code": "resource_missing",
    "doc_url": "https://stripe.com/docs/error-codes/resource-missing",
    "message": "Customer cus_XXXXXXXXXXXX does not have card with ID card_YYYYYYYYYYYYYYYYY",
    "param": "starting_after",
    "request_log_url": "https://dashboard.stripe.com/logs/req_XXXXXXXXXXXXXXXXXXX",
    "type": "invalid_request_error"
  }
}
```

Alex-Bair:
**Description:**

Various small fixes/refactors bundled together in this PR:
- Unnecessary fields were removed from models in order to take full advantage of schema inference & avoid schema violations. The remaining fields are either primary keys or are necessary to process child streams (like with `SubscriptionItems` and `UsageRecords`). Closes #1748.
- All query parameters are specified in `api.py` instead of split between class definitions and `fetch_*` methods. This addresses a bug where the last page token of a previous pagination cycle was retained in the `PARAMETERS` class variable, causing us to make an invalid API request on the next pagination cycle & crash the connector. Closes #1790 
- The `parent_id` field is no longer added in child streams. This field was redundant with existing fields the API already provides.
- Snapshot file names were updated to not include the file location. There were a few changes to the discover snapshot, but they are all expected from removing unnecessary fields from models.
- Some minor formatting & refactoring.

**Workflow steps:**

**Notes for reviewers:**

Tested all streams on a local stack with a small date range & confirmed no errors. Also, confirmed that we no longer get the "Customer X does not have card Y" errors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1942)


williamhbaker (approved):
LGTM

PR #1947, opened 2024-09-16
Phil/http ingest logs

psFried:
**Description:**

Minor updates to the logging in the `source-http-ingest` connector to add some useful information and reduce redundant spans.

Now with debug logging enabled, you'll get logs like this:

```json
{"timestamp":"2024-09-16T17:58:46.055975789Z","level":"DEBUG","message":"close","time.busy":"229µs","time.idle":"236µs","span":{"handler_time_ms":"0","path":"/another.json","published":1,"status_code":"200 OK","name":"http-request"}}
```

where the span covers the entire time it took to process the request, including sending the response. The `handler_time_ms` is the time it took to validate and publish the documents, and `published` is the number of documents that were captured.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1947)

### DIFF
--- a/source-http-ingest/Cargo.lock
+++ b/source-http-ingest/Cargo.lock
@@ -2410,6 +2410,7 @@ dependencies = [
  "tokio-util",
  "tower-layer",
  "tower-service",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

PR #1948, opened 2024-09-16
source-impact-native: removing logcursor check for Actions streams

Luishfs:
The Actions stream response seems to be ignoring the creation date order, and it also seems to not be returning data as in order of last updated record, only of impact creation. in order to support that, im removing the smaller logcursor check which would cause the connector to not return every record in incremental sync

The Actions stream response seems to be ignoring the creation date order, and it also seems to not be returning
data as in order of last updated record, only of impact creation. in order to support that, im removing the smaller logcursor check
which would cause the connector to not return every record in incremental

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1948)

### DIFF
--- a/source-impact-native/source_impact_native/api.py
+++ b/source-impact-native/source_impact_native/api.py
@@ -143,10 +143,7 @@ async def fetch_incremental_actions(
                     max_ts = _s_to_dt(results[f"CreationDate"])
                     doc = _cls.model_validate_json(json.dumps(results))
                     yield doc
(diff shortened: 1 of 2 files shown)


Luishfs:
@jonwihl Added both logcursor fix and tombstone fix for linkedin-ads-v2

jonwihl (approved):
LGTM

PR #1956, opened 2024-09-17
source-mongodb: batch mode for capture bindings

williamhbaker:
**Description:**

**Workflow steps:**

**Notes for reviewers:**


source-mongodb/main.go
```diff
@@ -236,44 +298,93 @@ func (d *driver) Validate(ctx context.Context, req *pc.Request_Validate) (*pc.Re
...
+
+		// Validate changes to the resource spec.
+		for idx, lastResource := range lastResources {
+			if lastResource.Database == res.Database && lastResource.Collection == res.Collection {
+				// Changing the capture mode is not allowed, although
+				// theoretically going from change stream to a snapshot could
+				// work. It's just easier to not allow it for now unless the
+				// binding is being reset (backfill counter incremented) as
+				// well.
+				if lastResource.getMode() != res.getMode() && binding.Backfill == lastBackfillCounters[idx] {
```
Alex-Bair, line 341: Should the backfill counter comparisons here be `binding.Backfill <= lastBackfillCounters[idx]`? Or is the control plane making sure the backfill counter is never decremented & we don't need to check that in the connector?

Alex-Bair (approved):
LGTM % minor question.

Alex-Bair:
Also looks like the [check](https://github.com/estuary/connectors/actions/runs/10908662896/job/30275196447?pr=1956) for `materialize-mongodb` failed. I'm guessing those snapshots need updated, assuming it uses the same testing dataset as `source-mongodb`?

source-mongodb/main.go
williamhbaker: Neither the control plane nor the connector checks that backfill counters are never decremented in general. Materializations do have some code for checking this which I added as a curiosity. It is perhaps debatable if this condition should be enforced. As it is, if you decrement the backfill counter for a task for some reason, the outcome is undefined - it's probably not a good idea, but has actually come in handy sometimes. So I think when designing connectors they should assume that backfill counters always increase, even if that isn't strictly true, and if one happens to decrease we really can't predict what is going to happen and shouldn't try.

PR #1957, opened 2024-09-17
source-http-ingest: fix handling of acknowledgements

psFried:
**Description:**

The source-http-ingest connector was never updated to account for the fact that the runtime can acknowledge multiple checkpoints with a single Acknowledge request. This manifested as a bug where it would wait for a long time on acknowledgements that would never come, and clients would experience many timeouts waiting for responses.

The fix is to expect a non-zero `checkpoints` as part of each acknowledgement, and acknowledge that number of transactions.

I've tested this by setting up a capture in production and generating load by having multiple shells that each use curl to send requests in a loop. Then killing all but one of those scripts would result in a deadlock of the remaining one. That's no longer happening with this branch. I did observe a different deadlock, though, which seems to be in the TLS negotiation with data-plane-gateway. I'm ignoring that for now, since we're working on removing DPG anyway and I'm guessing that it primarily affects clients such my bash scripts that are rapidly creating new connections.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1957)

### DIFF
--- a/source-http-ingest/src/transactor.rs
+++ b/source-http-ingest/src/transactor.rs
@@ -15,12 +15,15 @@ use tokio::{
 
 pub struct Acknowledgements(io::BufReader<io::Stdin>);
 impl Acknowledgements {
-    pub async fn next_ack(&mut self) -> anyhow::Result<anyhow::Result<()>> {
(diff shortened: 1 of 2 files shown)


mdibaiee (approved):
LGTM

PR #1960, opened 2024-09-17
source-stripe-native: only discover accessible streams

Alex-Bair:
**Description:**

Stripe API keys that start with `rk_` are restricted keys that may not have permissions for all API endpoints. If a restricted key is provided, we have to check each stream's accessibility during a discover.

Most child streams' accessibilities are the same as their parent stream. But `UsageRecords` is the exception - it can be inaccessible when its parent is accessible. This requires some special handling, like the `discoverPath` property within the `STREAMS` data structure.

When `discoverPath` is present on a child stream, an invalid parent id is used. This works since the Stripe API checks accessibilty before checking whether the requested resource exists, so we can key off the returned error code to determine accessibility. I did this instead of finding a valid parent id to both reduce the number of API requests during discover & to cover the situation where no parent id exists yet.

A check to confirm the `/events` endpoint is accessible was also added, since almost all of the streams rely on the `/events` endpoint for incremental replication.

**Workflow steps:**

**Notes for reviewers:**

Tested locally with both restricted and non-restricted API keys. Confirmed:
- When `/events` is not accessible, an error is thrown during discovery.
- Only accessible streams are discovered.
- Streams backfill & replicate incrementally after task creation.

Discover snapshot differences are expected since the order that streams are discovered has changed.

Within `resource.py`, the `issuing_object` function was removed since it was exactly the same as the `base_object` function.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1960)

### DIFF
--- a/source-stripe-native/source_stripe_native/resources.py
+++ b/source-stripe-native/source_stripe_native/resources.py
@@ -2,139 +2,111 @@ from datetime import datetime, UTC, timedelta
 from typing import AsyncGenerator, Awaitable, Iterable
 from logging import Logger
 import functools
+import re
 
 from estuary_cdk.flow import CaptureBinding
 from estuary_cdk.capture import Task
 from estuary_cdk.capture.common import Resource, LogCursor, PageCursor, open_binding, ResourceConfig, ResourceState
-from estuary_cdk.http import HTTPSession, HTTPMixin, TokenSource
+from estuary_cdk.http import HTTPSession, HTTPMixin, TokenSource, HTTPError
 
-from .api import (fetch_incremental,
-                  fetch_backfill,
-                  fetch_incremental_substreams,
-                  fetch_backfill_substreams,
-                  fetch_incremental_no_events,
-                  fetch_backfill_usage_records,
-                  fetch_incremental_usage_records,
-                  )
+from .api import (
+    API,
+    fetch_incremental,
+    fetch_backfill,
+    fetch_incremental_substreams,
+    fetch_backfill_substreams,
+    fetch_incremental_no_events,
+    fetch_backfill_usage_records,
+    fetch_incremental_usage_records,
+)
 
 from .models import (
-    Accounts,
-    Authorizations,
-    Bank_Accounts,
-    BalanceTransactions,
-    Cards,
-    CardHolders,
-    Coupons,
-    Customers,
-    CustomerBalanceTransaction,
-    Charges,
-    CheckoutSessions,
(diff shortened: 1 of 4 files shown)


source-stripe-native/source_stripe_native/resources.py
```diff
@@ -2,140 +2,102 @@
...
+) -> bool:
+    """
+    Returns a boolean representing whether the provided stream is accessible (True if accessible, False if inaccessible).
+
+    The Stripe API response has a 200 or 404 status code if the stream is accessible.
+
+    404 status codes are received when checking a child stream that can be inaccessible when its parent stream is accessible.
+    """
+    code = requests.get(url=url, auth=HTTPBasicAuth(access_token, "")).status_code
+    return code == 200 or code == 404
```
williamhbaker, line 43: I think we need to explicitly check for a specific code that means the stream is _not_ accessible, like a `400` code. Otherwise intermittent errors may falsely claim that the stream is not accessible, and the binding could get disabled from an autodiscovery.

I'd also suggest using the `http` object we pass around everywhere for this - that way its retying mechanisms can be used.

Alex-Bair: I'll update to check for the status code returned when a stream isn't accessible (IIRC the Stripe API returns a `403`, but I'll confirm).

I initially didn't use the `http` object we pass around because the [error it raises](https://github.com/estuary/connectors/blob/164f6284e2ec7742f053328bd3206828049ce475/estuary-cdk/estuary_cdk/http.py#L279-L281) doesn't provide the status code as a separate property, but rather lists it the error message. I didn't like the idea of keying off of a message that could change at a later time. 

What do you think about me updating those lines in our CDK to raise a custom error like the one below instead? It'd provide the status code as a dedicated property instead of stuffed inside the error message.
```python
class HTTPError(RuntimeError):
    def __init__(self, message: str, code: int):
        super().__init__(message)
        self.code = code
```

williamhbaker: Yeah that seems great!

williamhbaker (approved):
LGTM

Alex-Bair:
I did a little additional checking before merging this morning, and I missed that the status codes for inaccessible streams are different between `rk_test` and `rk_live` API keys. `rk_test` returns `403` and `rk_live` returns `401`. I updated the `check_accessibility` function to account for this:
```python
is_permission_blocked = err.code == 401 or err.code == 403
```

PR #1962, opened 2024-09-18
source-stripe-native: fix incremental logic

Alex-Bair:
**Description:**

In the incremental replication functions, the logic for emitting documents and updating the cursor were coupled; whenever we emit a document, the cursor would be updated. This would work if the API returned documents in increasing order, but the Stripe API returns documents in _decreasing_ order. This means that the cursor was being updated to the timestamp for the oldest document we haven't seen, not the most recent document.

The incremental logic for updating the cursor is now decoupled from emitting documents, and the cursor is now updated to be the most recent document's timestamp.

Variable names were also updated to more accurately reflect the associated entities & improve readability (ex: `events` or `resources` instead of `result`).

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack for a base stream, a child stream, and `UsageRecords` for a short duration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1962)

### DIFF
--- a/source-stripe-native/source_stripe_native/api.py
+++ b/source-stripe-native/source_stripe_native/api.py
@@ -40,22 +40,30 @@ async def fetch_incremental(
     _cls: Any = cls  # Silence mypy false-positive
 
     while iterating:
-        result = EventResult.model_validate_json(
-        await http.request(log, url, method="GET", params=parameters)
-    )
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #1966, opened 2024-09-19
source-postgres: Discover OIDs as integers and add tests

willdonnelly:
**Description:**

Previously OID values "worked" (they've always been captured as integer values and it was possible to use them as the backfill key as far as the connector was concerned), but were discovered as the catch-all schema `{}` which meant that OID primary keys would produce a collection with an invalid collection key.

This commit fixes that by changing OID discovery to say they're integers, and adds some test cases to make sure OIDs work as we intend both as values and as primary keys.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1966)

### DIFF
--- a/source-postgres/.snapshots/TestScanKeyTypes-OID
+++ b/source-postgres/.snapshots/TestScanKeyTypes-OID
@@ -1,0 +1,88 @@
+####################################
+### Capture from Start
+####################################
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

PR #1975, opened 2024-09-23
source-postgres: Exclude `pglogical` schema from discovery

willdonnelly:
**Description:**

This is a "system schema" added by the pglogical extension [1] and I don't believe the typical user who happens to have that extension installed wants these tables captured in the first place.

[1] https://www.2ndquadrant.com/en/resources-old/pglogical/pglogical-docs/

**Workflow steps:**

Discovery will no longer produce bindings for tables in the `pglogical` schema in cases where previously it did.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1975)


dyaffe (approved):
LGTM

PR #1976, opened 2024-09-23
source-stripe-native: fix streams' incremental replications

### LINKED ISSUE #1965: source-stripe-native: some streams' incremental replications aren't happening
### Summary

Some streams for `source-stripe-native` are not replicating incrementally despite data being created/updated in Stripe. For example:

- `ApplicationFees`: The incorrect event type is being checked. The connector is checking for `application_fee.refunded` events when it should be checking for both `application_fee.created` and `application_fee.updated` events.
- `Refunds`: No `refund` events are generated by Stripe. Despite the Stripe API documentation listing [`refund.created`](https://docs.stripe.com/api/events/types#event_types-refund.created) and [`refund.updated`](https://docs.stripe.com/api/events/types#event_types-refund.updated) events, these events are not present on the `/events` endpoint when a refund is created/updated.

Other streams may have these issues too, but they haven't popped up yet or are only checking for a subset of the appropriate events (i.e. only checking for `.updated` but not `.created`). It wouldn't be a bad idea to audit all events we're checking & confirm they are actually being created in Stripe. We should also make sure we're checking for all appropriate events for each stream.

A possible solution includes the following:
- Checking for different event types between parent streams vs. child streams. Right now, they always check the same event type.
- Checking for multiple event types using an array of types as query params (specified like `types[0]`, `types[1]`, etc.) instead of checking for only a single type with the `type` param.
- Switching any streams that don't actually have create/update events over to the existing "no event" resource type.

Alex-Bair:
**Description:**

There were various issues preventing some streams from incrementally replicating:
- The incorrect event types were being listened to (ex: `ApplicationFees`)
- The Stripe API does not generate certain events (ex: `Payouts`).

The connector has been updated to let us check for multiple different event types with the `EVENT_TYPES` class variable. This variable references a dictionary whose keys are Stripe API event types and whose values are the appropriate `meta_.op` value. `EVENT_TYPES` can be different between parent and child streams, enabling "split" streams - streams that backfill like child streams but incrementally replicate like standard streams with associated events.

Notes:
- There are a number of streams whose events I couldn't confirm are actually generated by the Stripe API. These are noted in the code with comments.
- The `Payouts` stream was switched over to be a no-events stream since the associated events aren't generated by Stripe. It seems like most payout docs aren't updated after they are created, so this should work for most use cases. But if that assumption doesn't hold & we need to capture updated payouts, we'll need to rethink this strategy.

Closes #1965.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed what streams I could access are now performing incremental replication. 

Discover snapshot changes are expected due to description updates and the addition of the `object` attribute to all Stripe documents' schemas.


williamhbaker (approved):
LGTM

Lots of little nuances to the Stripe API to work through, nice job!

PR #1978, opened 2024-09-24
source-stripe-native: make created optional for Accounts

Alex-Bair:
**Description:**

Sometimes, `account.updated` events do not have a `created` field in the document, causing failures when we validate the document against its model. Making `created` optional will avoid these failures and reflect the API's actual response. The default value of `None` is removed from the schema to keep it aligned with how we understand the `created` field: it's either present as an `int` or it's not present at all. In other words, it's an optional, non-nullable field.

Note: I've only noted that this occurs for `account.updated` events for when the user's own Stripe account is created. I couldn't confirm if all `account.updated` events have `created` missing in the document, or if `Accounts` documents really don't have `created` fields at all. If this becomes an issue later, we will need to re-evaluate how we backfill `Accounts` since we rely on a `created` field being present.

Discover snapshot change is expected due to switching `created` from a required field to an optional field for `Accounts`.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed `Accounts` does not cause shard failures when `account.updated` events do not have a `created` field in the document.

Existing tasks will need to be re-discovered since we're changing `created` from required to optional for `Accounts`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1978)

### DIFF
--- a/source-stripe-native/source_stripe_native/models.py
+++ b/source-stripe-native/source_stripe_native/models.py
@@ -151,6 +151,12 @@ class Accounts(BaseStripeObjectWithEvents):
         "account.updated": "u",
     }
 
+    # Accounts docs returned in account.updated events may not have a created field.
+    created: int = Field(
+        default=None,
+        # Don't schematize the default value.
+        json_schema_extra=lambda x: x.pop('default') # type: ignore
+    )
 
 # Could not verify Persons events are generated in test mode, but suspect
 # they are generated in Stripe's live mode.
--- a/source-stripe-native/tests/snapshots/snapshots__discover__stdout.json
+++ b/source-stripe-native/tests/snapshots/snapshots__discover__stdout.json
@@ -59,8 +59,7 @@
       },
       "required": [
         "id",
-        "object",
-        "created"
+        "object"
       ],
       "title": "Accounts",
       "type": "object",


source-stripe-native/source_stripe_native/models.py
```diff
@@ -151,6 +151,8 @@ class Accounts(BaseStripeObjectWithEvents):
         "account.updated": "u",
     }
 
+    # Accounts docs returned in account.updated events may not have a created field.
+    created: int | None = None
```
williamhbaker, line 155: Is there a way to get this to schematize as "not required to be present, but if it is present, it must be `int`"? I think that's what we want: The field is either present as an int, or not present at all. It shouldn't ever be present as an actual `null` value.

The schema coming from this is saying it _can_ be a `null`, and also giving it a default `null` value (we don't want that default, for sure). Ideally the schema would be this:

```
created: {
  "type": "integer",
  "title": "created"
}
```

...and it would not be present in the list of `required` fields.

Alex-Bair: I've looked through a handful of Pydantic discussions, and I haven't found any straightforward way to have an optional, non-nullable field. I've looked in Pydantic's [docs](https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields), and it looks like optional fields always have to have a default value.

To avoid saying `created` can be `null` in the schema, we could have a default value like `-1`, then overwrite `created` with the event's timestamp whenever we see that it's `-1`. That's not 100% ideal, but I'm not seeing a better way to handle this right now. What do you think about that approach?

williamhbaker: I have used [this hack](https://github.com/estuary/connectors/blob/911128ff1037dbaecc7bc18082fb085c13717291/source-hubspot-native/source_hubspot_native/models.py#L400) for not schematizing the default value before - does that help here?

I think we need a way to represent this correctly. If it the field is either present and is some integer, or is not present at all, we need to have the schema say that. A default of `null` is not ideal, but a default of `-1` is even worse since that is not what the value of the field is, and since it's a unix timestamp might be problematic in other ways.

Alex-Bair: That code you linked did help - thank you! I'll get the PR updated.

williamhbaker (approved):
LGTM

Issue #1951, opened 2024-09-16
source-mysql: Fix replication of text columns with non-Unicode character sets


willdonnelly:
When a text column uses a non-Unicode charset (such as the `latin1` default used by MySQL 5.7 and below), values are written into the binlog using that encoding and it's the responsibility of the reader to decode them appropriately. Currently our replication value decoding doesn't do that and we just assume the bytes are a valid UTF-8 string.

(Backfill queries appear to handle text with non-Unicode character sets just fine, for what it's worth, the bug here is solely with how we handle text values in binlog events)

We get away with this most of the time because the default character set in modern MySQL / MariaDB versions is a multi-byte UTF-8 encoding, and because in practice most text is ASCII. But when both of those assumptions are violated at the same time, even in a relatively central case like a MySQL 5.7 database with the most basic of accented Latinate characters (as in French/Spanish/etc text), we're pretty blatantly mangling the data.

Note that the capture of these columns isn't exactly "incorrect" in the usual way we use the word, as the text we capture is still a consistent representation of the original source data. The strings will just have all non-ASCII characters replaced with the `U+FFFD REPLACEMENT CHARACTER` and the ASCII code points captured correctly. The problem here is that this is basically never what the user actually wants and we have all the information we need to faithfully translate the non-ASCII code points to the appropriate Unicode equivalents.

We should improve on this by:

- Obtaining the column character set or collation name for each text column during discovery.
- Using this information to do a charset-aware `[]byte -> string` conversion at the point where we currently just assume they're valid UTF-8 and cast to a string.
- So long as we don't support the complete set of all MySQL charset / collation names, we should throw an error somewhere along the line when trying to capture a table with an unsupported collation rather than mistranslating the text.

PR #1979, opened 2024-09-24
source-mysql: Decode text correctly in non-UTF8 character sets

willdonnelly:
**Description:**

This PR fixes `source-mysql` to correctly decode text columns with non-UTF8 character sets. This fixes https://github.com/estuary/connectors/issues/1951

Previously backfilled values were correct, because MySQL always sends us those in UTF-8, but replicated values were occasionally incorrect (or at least, sub-optimal) because the bytes stored in the binlog are in whatever character set the column uses, and at a certain point we just cast `[]byte -> string` which results in any non-ASCII bytes getting replaced with `U+FFFD REPLACEMENT CHARACTER`.

The fix is:
1. Keep track of the character set of each column (this is the hard part, there's a lot of subtleties to making it work correctly even in oddball cases like a DDL `ALTER TABLE` query occurring after the table backfill which adds a new column where the character set could be explicit, could be implied by an explicit collation, or could be the table default).
2. Modify the decoding logic to apply an appropriate `[]byte -> string` decode function to text column bytes received via replication, based on the character set of the column. This part is reasonably straightforward, but required adding an extra parameter to the value translation code to distinguish between backfilled values (which are always UTF-8) and replicated ones (where we need to do all this fanciness).
3. Implement a handful of decoding functions. Currently we've got UTF-8, latin-1, and UCS-2 decoders. There are a couple more which might theoretically come up in the future but which I didn't think warranted worrying about right now. Honestly even UCS-2 is unlikely to ever get used in the real world, I just added it because as a double-byte character set it was a good way of exercising some of the edge cases and clearly demonstrating working vs non-working behavior.

I have opted to make the connector default to a UTF-8 decoder when faced with an unknown or unrecognized charset rather than failing the capture. This would probably be the wrong choice if the connector were new, but right now we don't know if there are any captures in production using some other character set which I overlooked, so the safest behavior is to keep doing the same thing as before this PR (decoding the bytes by casting to a UTF-8 string directly) while logging an error. Later on I can go check to see if that error actually occurred in production anywhere, add support for the offending character sets, and consider making "unknown charset" a fatal error at that time.

**Workflow steps:**

Most captures don't need to care or do anything. Captures will continue operating as normal for preexisting bindings (because their metadata for all tables and columns has no charset information, which means an implicit default of UTF-8). New captures, or new bindings or backfills on existing captures, will reinitialize that metadata and pick up the appropriate charset information. In most cases this will have no impact because the charset really is UTF-8.

Captures of latin-1 or other non-Unicode text columns which actually contain non-ASCII code points currently being captured as `U+FFFD REPLACEMENT CHARACTER` can be fixed to capture the correct values for these non-ASCII characters by re-backfilling the binding.


jgraettinger (approved):
LGTM

PR #1980, opened 2024-09-25
source-impact-native: switching snapshots pk to _meta/row_id

Luishfs:
after fixing the tombstone value with the correct class, the connector still broke due to tombstone failures. The newer bugs happened because there were still a pk required that was not being fed to the newer tombstone model. In order to fix that, i've removed all previous pk from snapshot models and fixated all pks to _meta/row_id ( the default strategy )

**Notes for reviewers:**

All changes were tested using a local flow instance, and materialized to a free-tier neon POSTGRES db
All snapshot streams were tested and materialized

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1980)


jonwihl (approved):
LGTM

jonwihl (approved):
LGTM

PR #1982, opened 2024-09-26
source-stripe-native: handle responses for children of deleted resources

Alex-Bair:
**Description:**

Child streams, like `InvoiceLineItems` perform incremental replication by watching for events on their associated parent resource, then requesting all child resources for that parent when an event happens. However, since there can be some time between when we process an event & what actually happens in Stripe, it's possible for us to request child resources of a deleted parent. Stripe returns a `404` code and somewhat unique text to indicate this. We can safely skip requesting these resources when this happens.

This situation can happen if a creation/update is quickly followed by a deletion. If the connector is stuck for a while and a backlog of events we haven't processed builds up, there's even more opportunity for us to request deleted resources.

I noticed this issue happening with `InvoiceLineItems`, so I also removed the "invoice.deleted" event from that stream's `EVENT_TYPES` since we'll just get an error back if we request a deleted invoice's line items.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that requesting children of a deleted parent resource no longer crashes the connector & that the connector continues processing after handling the error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1982)

### DIFF
--- a/source-stripe-native/source_stripe_native/api.py
+++ b/source-stripe-native/source_stripe_native/api.py
@@ -3,13 +3,16 @@ from decimal import Decimal
 from estuary_cdk.http import HTTPSession
 from logging import Logger
 from typing import Iterable, Any, Callable, Awaitable, AsyncGenerator, Literal
-import json
+import json, re
 
 from estuary_cdk.capture.common import (
     BaseDocument,
     PageCursor,
     LogCursor,
 )
+
+
+from estuary_cdk.http import HTTPError
 
 
 from .models import (
@@ -25,6 +28,8 @@ from .models import (
 
 API = "https://api.stripe.com/v1"
 MAX_PAGE_LIMIT = 100
+
+MISSING_RESOURCE_REGEX = r"resource_missing.+No such.+"
 
 def add_event_types(params: dict[str, str | int], event_types: dict[str, Literal["c", "u", "d"]]):
     """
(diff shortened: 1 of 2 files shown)


source-stripe-native/source_stripe_native/api.py
```diff
@@ -328,9 +333,21 @@ async def _capture_substreams(
...
+        try:
+            result_child = ListResult[cls_child].model_validate_json(
+                await http.request(log, child_url, method="GET", params=parameters)
+            )
+        except HTTPError as err:
+            # It's possible for us to process events for deleted parent resources, making
+            # the requests for the associated child resources fail. Stripe returns a 404
+            # error & a message containing "resource_missing" and "No such" when this happens.
+            if err.code == 404 and bool(re.search(MISSING_RESOURCE_REGEX, err.message, re.DOTALL)):
+                log.debug(f"Missing resource error for URL {child_url}. Skipping to the next resource.", err)
```
psFried, line 346: IDK how common it is to get 404 responses, but it might be helpful to log at info or warn levels as long as it wouldn't be too spammy. Feel free to ignore this suggestion if it'd result in a bunch of log spam though.

psFried (approved):
LGTM

source-stripe-native/source_stripe_native/api.py
Alex-Bair: Thanks for the suggestion, Phil! I don't anticipate we'd encounter this error all that often. I'll switch it over to warn.

PR #1985, opened 2024-09-26
materialize-redshift: Switch to writing checkpoints using LZ4 compression

jshearer:
The motivation for this change is to enable more bindings in a redshift materialization. It should be backwards-compatible in that it supports loading checkpoints in either base64 or LZ4 formats, but will always write new checkpoint data in LZ4.

### 🚨 Do not merge yet 🚨

I have not tested this yet. I need to figure out how to set up a redshift account, run the connector, test that it upgrades from base64 to lz4 correctly, etc.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1985)


williamhbaker:
Closing in favor of https://github.com/estuary/connectors/pull/1999

PR #1986, opened 2024-09-26
source-stripe-native: handle UsageRecords 400 response

Alex-Bair:
**Description:**

Stripe has two billing systems, usage-based and meter-based. Our UsageRecords stream hits the legacy usage-based endpoint. If we try to hit the usage-based endpoint with a subscription item that doesn't actually use usage-based billing, Stripe response with a `400` error and a pretty specific message. Since there are no usage-based usage records for that subscription item, we can safely skip requesting usage records when this happens.

I also fixed the warn level log for my previous commit by passing a `dict` of the error's properties instead of passing the actual `HTTPError`.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed the connector does not crash when requesting usage records for a subscription item that doesn't use Stripe's legacy usage-based billing system and receiving a `400` response. Also confirmed that the warn level logs are actually showing up.

I reviewed the API docs and various API responses, and there's no way to determine if a subscription item is a part of the legacy usage-based billing system before we send a request for the associated usage records.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1986)

### DIFF
--- a/source-stripe-native/source_stripe_native/api.py
+++ b/source-stripe-native/source_stripe_native/api.py
@@ -30,6 +30,7 @@ API = "https://api.stripe.com/v1"
 MAX_PAGE_LIMIT = 100
 
 MISSING_RESOURCE_REGEX = r"resource_missing.+No such.+"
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

PR #1987, opened 2024-09-26
source-http-ingest: Stop using serde_json::Value

psFried:
Removes the use of `serde_json::Value` for parsed documents, and instead use `RawValue` as much as possible. This is done in order to reduce memory usage, as the overhead of `Value` is quite high.

Also changes to using a `String` to represent the fully prepared documents to be captured. This allows us to skip an extra serialization step since the capture protocol requires these documents to be `String`s anyway.

This unfortunately required changing how we validate documents, since the `doc::Validator` can only be used with parsed documents. So we instead use a `json::Validator`, which can validate the `String` directly.

Also changes to using jemalloc instead of the system allocator.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1987)

### DIFF
--- a/source-http-ingest/src/server.rs
+++ b/source-http-ingest/src/server.rs
@@ -6,13 +6,18 @@ use axum::{
     extract::{DefaultBodyLimit, Json, Path, Query, State},
     routing, Router,
 };
+use doc::Annotation;
 use http::status::StatusCode;
-use serde_json::Value;
+use json::validator::Validator;
+use serde_json::{value::RawValue, Value};
 use tower_http::decompression::RequestDecompressionLayer;
 use utoipa::openapi::{self, schema, security, OpenApi, OpenApiBuilder};
 use utoipa_swagger_ui::SwaggerUi;
 
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{BTreeMap, HashMap},
+    sync::Arc,
+};
 use tokio::io;
 use tokio::sync::Mutex;
 
@@ -97,8 +102,9 @@ async fn handle_webhook(
 }
 
 struct CollectionHandler {
+    schema_url: url::Url,
+    schema_index: json::schema::index::Index<'static, doc::Annotation>,
     binding_index: u32,
-    validator: doc::Validator,
     id_header: Option<String>,
 }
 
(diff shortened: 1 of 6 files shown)


source-http-ingest/src/server.rs
```diff
@@ -375,13 +399,27 @@ impl Handler {
...
+                schema_url,
+                ..
+            } = &collection;
+            let mut validator: Validator<doc::Annotation, json::validator::SpanContext> =
+                Validator::new(schema_index);
+            validator
+                .prepare(schema_url)
+                .context("preparing validator")?;
+
+            let mut deserializer = serde_json::Deserializer::from_str(&doc);
```
jgraettinger, line 413: i don't have a useful suggestion, but noting we are trading a gain in memory efficiency at a cost of parsing the JSON document twice.

In the future we could consider using a doc::Parser and parsing into a doc::HeapNode, which then has its bump allocator freed ASAP.

source-http-ingest/src/transactor.rs
```diff
@@ -30,13 +30,9 @@ impl Acknowledgements {
...
-    pub async fn emit_doc(&mut self, binding: u32, doc: &impl Serialize) -> anyhow::Result<()> {
+    pub async fn emit_doc(&mut self, binding: u32, doc_json: String) -> anyhow::Result<()> {
         use tokio::io::AsyncWriteExt;
 
-        // We intentionally don't use `write_capture_response` here because we don't want to wait
-        // for the flush call to complete.
-        let doc_json = serde_json::to_string(doc).context("serializing output document")?;
-
         let resp = Response {
             captured: Some(Captured { binding, doc_json }),
```
jgraettinger, line 37: There's a gotcha here: if the POST body contains newlines, we'll plumb them through and break the newline delimited protocol.

`models::RawValue` takes care of this case to ensure that newlines are removed.

jgraettinger (commented):
Thanks! There's an issue with RawValues that can contain newlines.

I'd suggest either using `models::RawValue`, or consider using doc::Parser with a re-useable doc::Allocator that's parsed / updated-with-meta / validated / serialized in one go, without holding across an `await`.

source-http-ingest/src/transactor.rs
psFried: Good catch! I just switched to `models::RawValue`.

jgraettinger (approved):
LGTM

PR #2007, opened 2024-10-01
source-zendesk-support: fix Groups and Organizations streams

Alex-Bair:
**Description:**

The `Groups` and `Organizations` streams were missing data. This PR addresses the issues causing these streams to miss data.

### `Groups`
Deleted groups were not being retrieved previously. By including the `exclude_deleted` query param and setting it to `false`, the Zendesk Support API is now returning deleted groups.

### `Organizations`
Only 10k organizations were being retrieved previously. The `Organizations` stream was using page numbers to paginate through results. This is an older, but still supported, pagination method for Zendesk Support, and they only allow you to paginate through 10k results even if there are more. Using the page number based pagination method was causing us to only retrieve 10k documents and miss any remaining ones.

I've updated the `Organizations` stream to use Zendesk's [time-based incremental export](https://developer.zendesk.com/documentation/ticketing/managing-tickets/using-the-incremental-export-api/#time-based-incremental-exports) endpoint. This endpoint paginates by providing a start time, and all resources created/updated at or after that start time are returned. This lets us retrieve all resources instead of just a subset. This pagination strategy does have the potential downside of getting stuck in a loop if 1000+ (the max number of results per page) resources are updated at the exact same time, but there aren't currently better ways to retrieve incremental updates for `Organizations`.

Capture snapshot changes are expected. The new time-based incremental export endpoint adds a `deleted_at` property to each organization resource.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack with a Zendesk account that had deleted groups and more than 10k organizations. Confirmed: 
- deleted groups are now retrieved 
- more than 10k organizations are retrieved
- updated tests pass

All `Organizations` bindings for existing tasks need to be backfilled after merging to clear out previous state.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2007)

### DIFF
--- a/source-zendesk-support/source_zendesk_support/streams.py
+++ b/source-zendesk-support/source_zendesk_support/streams.py
@@ -10,7 +10,7 @@ import time
 from abc import ABC
 from collections import deque
 from concurrent.futures import Future, ProcessPoolExecutor
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, UTC
 from functools import partial
 from math import ceil
 from pickle import PickleError, dumps
@@ -73,6 +73,18 @@ def to_int(s):
             return res[0]
     return s
 
+
+def _s_to_dt_str(s: int) -> str:
+    """
+    Converts a UNIX timestamp in seconds to a date-time formatted string.
+    """
+    return datetime.fromtimestamp(s, tz=UTC).isoformat(' ')
+
+def _dt_str_to_s(dt_str: str) -> int:
+    """
+    Converts a date-time formatted string to a UNIX timestamp in seconds.
+    """
+    return int(datetime.fromisoformat(dt_str).timestamp())
 
 class SourceZendeskException(Exception):
     """default exception of custom SourceZendesk logic"""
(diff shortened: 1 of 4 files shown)


source-zendesk-support/source_zendesk_support/streams.py
```diff
@@ -74,6 +74,18 @@ def to_int(s):
...
+    """
+    Converts a millisecond UNIX timestamp to a date-time formatted string.
+    """
+    return datetime.fromtimestamp(ts, tz=UTC).isoformat(' ')
+
+def _dt_to_ts(dt_str: str) -> int:
+    """
+    Converts a date-time formatted string to a millisecond UNIX timestamp.
+    """
+    return int(datetime.fromisoformat(dt_str).timestamp())
```
williamhbaker, line 87: I think this is actually going to return seconds - the float result has the integer part as seconds, and the fractional part as microseconds. We have [similar helpers](https://github.com/estuary/connectors/blob/75c2d46cce5260076043770ee5ae8c1a3be3863c/source-hubspot-native/source_hubspot_native/api.py#L893-L897) elsewhere that account for this to get millisecond epoch times.

source-zendesk-support/source_zendesk_support/streams.py
```diff
@@ -654,6 +734,17 @@ def parse_response(self, response: requests.Response, **kwargs) -> Iterable[Mapp
                     yield event
 
 
+class Organizations(SourceZendeskSupportIncrementalTimeExportStream):
+    """
+    API docs: https://developer.zendesk.com/api-reference/ticketing/ticket-management/incremental_exports/#incremental-organization-export
+
+    Note: This stream theoretically can get stuck in a loop if 1000+ organizations are
```
williamhbaker, line 741: Not a blocker, but per the API docs it kind of implies that this won't actually be possible, since it sounds like it'll keep returning results until one of them has a different timestamp?

> ... The 1000-item limit may be exceeded if items share the same timestamp ...

source-zendesk-support/source_zendesk_support/streams.py
Alex-Bair: You're right, thanks for catching that. I was wrong in that function description. Zendesk's timestamps are in seconds, not milliseconds. I'll update the function description to say "...to a UNIX timestamp in seconds". 

Also, the reason I stored the state as a date-time formatted string instead of an actual `datetime` objects was because when I retrieved the state after restarting the task, the retrieved state wasn't a `datetime` object but was the ISO formatted string. I'll update the helper function names to be clearer that they're accepting/returning a string and not a `datetime` object.

source-zendesk-support/source_zendesk_support/streams.py
Alex-Bair: Yep, the API docs do make it seem like it wouldn't be an issue. But we have seen other streams that used this time-based incremental export method get stuck in a loop if 1000+ resources are updated at the same time (ex: we had to move the `Tickets` stream to a different pagination method because it was stuck requesting the same timestamp - PR was #1789). 

I don't remember the exact number of tickets that were updated at the same time in that instance, but I think it was multiple thousands. I suspect there's a limit to how much higher that 1000-item limit can go. So if there are more resources updated in the same second than 1000 + whatever the flex amount is, the stream will get stuck.

williamhbaker (approved):
LGTM

PR #2012, opened 2024-10-02
source-zendesk-support: switch Groups stream to use cursor pagination

Alex-Bair:
**Description:**

Previously, the `Groups` stream was inheriting from the `BaseSourceZendeskSupportStream` class, meaning it paginated by retrieving the total number of results from a `/count` endpoint, calculated how many pages there should be based off that count, then concurrently sent requests for those pages of results. When I added the `exclude_deleted` param earlier to all `Groups` requests, I didn't ensure that the `/count` endpoint would include deleted groups. It does not, and it doesn't accept the `exclude_deleted` query param. So when the connector was calculating how many pages there should be, it wasn't including deleted groups in the total count, and we were missing results.

Instead of keeping `Groups` as a `BaseSourceZendeskSupportStream` that uses this older method of pagination, I switched `Groups` to use cursor pagination. The stream's behaviour and document schema remained the same, only how it queries the Zendesk API changed. 

We should move `Macros` and `TicketFields` off of `BaseSourceZendeskSupportStream` too since their current pagination method has limitations that don't exist for the newer cursor pagination-based classes, but that can be done at a later time.

**Workflow steps:**

**Notes for reviewers:**

- All results for `Groups` were retrieved.
- All tests pass.

### DIFF
--- a/source-zendesk-support/source_zendesk_support/streams.py
+++ b/source-zendesk-support/source_zendesk_support/streams.py
@@ -841,15 +841,16 @@ class TicketComments(SourceZendeskSupportTicketEventsExportStream):
             yield record
 
 
-class Groups(SourceZendeskSupportStream):
+class Groups(SourceZendeskSupportCursorPaginationStream):
     """Groups stream: https://developer.zendesk.com/api-reference/ticketing/groups/groups/"""
+
     def request_params(
         self,
-        stream_state: Mapping[str, Any] = None,
+        stream_state: Mapping[str, Any],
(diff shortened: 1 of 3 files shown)


Alex-Bair:
In the 30 minutes from when I tested locally & created the PR, Zendesk Support's API stopped including the `configuration.explore_reports` field in custom roles (the same field they added over the weekend & we merged PR #1991 earlier this week to address). I've removed that field from our capture snapshot so our tests can pass.

williamhbaker (approved):
LGTM

PR #2013, opened 2024-10-03
source-oracle-flashback: only validate and operate on enabled bindings

mdibaiee:
**Description:**

- Until now we were doing a full discovery for each Validate and Open call, but normally other connectors only operate on the requested bindings. Here I'm implementing the same logic so that we don't have to do a full discovery every time, instead we only validate tables that match bindings.
- This originally came about because we have tasks which take so long to discover that their Validate request was timing out. Most of the time spent during the request was spent on getting the `MAX_ROWID` of tables. By reducing the number of tables that we query to only the enabled ones, we should be able to run Validates much faster.
- Tested with integration tests and manually running a task with disabled bindings

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2013)

### DIFF
--- a/source-oracle-flashback/source_oracle_flashback/resources.py
+++ b/source-oracle-flashback/source_oracle_flashback/resources.py
@@ -96,14 +96,11 @@ async def validate_flashback(
                 log.warn("We recommend guaranteeing retention of the undo tablespace. See go.estuary.dev/source-oracle for more information.")
 
 
-async def all_resources(
-    log: Logger, http: HTTPMixin, config: EndpointConfig, pool: oracledb.AsyncConnectionPool, is_rds: bool,
+async def tables_to_resources(
+    log: Logger, http: HTTPMixin, config: EndpointConfig, pool: oracledb.AsyncConnectionPool, is_rds: bool, tables: list[Table],
 ) -> list[common.Resource]:
     resources_list = []
-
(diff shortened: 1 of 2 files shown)


source-oracle-flashback/source_oracle_flashback/resources.py
```diff
@@ -96,11 +96,9 @@ async def validate_flashback(
                 log.warn("We recommend guaranteeing retention of the undo tablespace. See go.estuary.dev/source-oracle for more information.")
 
 
-async def all_resources(
-    log: Logger, http: HTTPMixin, config: EndpointConfig, pool: oracledb.AsyncConnectionPool, is_rds: bool,
-) -> list[common.Resource]:
-    resources_list = []
-
+async def get_tables(
```
williamhbaker, line 99: I can't actually find anywhere that this `get_tables` function is being used, am I missing something?

williamhbaker (approved):
LGTM % comment about `get_tables`

source-oracle-flashback/source_oracle_flashback/resources.py
mdibaiee: Forgot to remove this after a refactor 🤦🏽

PR #2023, opened 2024-10-07
source-pendo: separate backfill from incremental

Alex-Bair:
**Description:**

For Pendo accounts with a significant amount of data, the strategy of asking for all data up until the present does not work. Pendo's response time is long enough to cause `TimeoutError`s, and responses that are received end up OOM-ing the connector.

To fix this, there is now a distinct backfill process for events and aggregated events. 
Notable changes include:
- Separating backfills from ongoing incremental replication.
  - Backfills try to get at max a day's worth of data. Incremental replication continues to get all data up to the present.
- The cutoff between backfills and incremental replication is shifted backwards 12 hours due to delays between when an event occurred and when it is available in the API. This is to prevent backfills from mistaking this delay as a series of empty date windows.
- The Pendo API response limit is increased to 50k. This might be able to be increased, but I've done limited testing to confirm 50k won't OOM the connector.
- API responses are now sorted first by timestamp and next by resource ID. This lets us more effectively "paginate" through documents if multiple occur at the same timestamp.
- If all the documents in a given API response have the same timestamp, the connector will fetch the remaining documents with that exact timestamp before incrementing the connector by a millisecond.

Discover snapshot change is expected due to the increased interval for event streams.

**Workflow steps:**

**Notes for reviewers:**

- Events and aggregated events are backfilled a day at a time.
- Backfills are completed once the lower time bound reaches the cutoff date.
- In a small timeframe, all streams make progress with a data-heavy Pendo account & the connector does not crash.

Sometimes, Pendo's API takes too long to response to a given query, even a simple/small query. These cause intermittent `TimeoutError`s, but the connector makes progress after restarting & re-querying the API.

All existing tasks will need their event & aggregated events streams backfilled after merging.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2023)

### DIFF
--- a/source-pendo/source_pendo/api.py
+++ b/source-pendo/source_pendo/api.py
@@ -1,19 +1,157 @@
-from datetime import datetime
+from datetime import datetime, timedelta, UTC
 from logging import Logger
 from typing import AsyncGenerator
 import pendulum
 from pydantic import TypeAdapter
-from estuary_cdk.capture.common import BaseDocument, LogCursor
+from estuary_cdk.capture.common import BaseDocument, LogCursor, PageCursor
 from estuary_cdk.http import HTTPSession
 
 from .models import PageEvent, FeatureEvent, TrackEvent, GuideEvent, PollEvent, AggregatedEventResponse,EventResponse, Resource, Metadata
 
 API = "https://app.pendo.io/api/v1"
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

PR #2031, opened 2024-10-08
source-mysql: More connection error logging

willdonnelly:
**Description:**

I hate this TLS vs non-TLS connection retrying thing and how it results in two distinct errors, let's just make sure to log both so I can at least figure out what went wrong in all cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2031)


psFried (approved):
LGTM

PR #2036, opened 2024-10-09
materialize-sql: only evaluate migrations for selected fields

williamhbaker:
**Description:**

Evaluating migrations of all projections for which a column exists runs into an edge case if the projection has had all of its types removed for some reason, because materializations don't know how to map Flow schema fields with no types to their respective endpoint types.

Instead only evaluate migrations for columns that are included in the field selection. Validation ensures that fields with no types are forbidden from being selected. In a scenario where a projection goes from having types + a column created, to having no types and being excluded from migrations and field selection, to having types again at some point in the future, the migration will be evaluated then when the field for the projection is back in the field selection.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2036)

### DIFF
--- a/materialize-sql/apply.go
+++ b/materialize-sql/apply.go
@@ -6,6 +6,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"strings"
 
 	boilerplate "github.com/estuary/connectors/materialize-boilerplate"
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM

mdibaiee (approved):

PR #2037, opened 2024-10-09
source-klaviyo: fix incremental pagination

Alex-Bair:
**Description:**

Incremental streams were using `>` instead of `>=` when filtering results with a specific datetime. Although subsequent requests use cursors to paginate through responses, it's possible to miss results within the same second on connector restarts (like for schema inference updates).

Snapshot names have also been updated to align with what's generated automatically by `pytest`.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed more data is captured using `greater-or-equal` instead of `greater-than` in the filter query param.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2037)


williamhbaker (approved):
LGTM

There's the obvious consideration that rather than missing records, we'll end up with some duplicates, but that seems better than missing some. And there may be scenarios where the capture could get "stuck" if a huge number of records all have the same timestamp? But I'm not sure how likely that is, and it's probably not something that we can practically address right now.

Alex-Bair:
Agreed, I figured for a quick fix, it'd be better to have some duplicates rather than miss records. And the way the `IncrementalKlaviyoStream` is set up, I don't think it'll get stuck in a loop. When the connector starts up, it uses the checkpointed datetime for the initial request, then it paginates through the remaining records up to the present with [page cursors](https://developers.klaviyo.com/en/reference/api_overview#pagination). Then the datetime of the last emitted record is checkpointed to be used when the connector starts up again.

Changing the checkpointed values to be those page cursors would help reduce duplicates, but threading that through & ensuring nothing breaks is a larger effort that'd take a bit longer than just changing the filter query param.

PR #2038, opened 2024-10-10
materialize-snowflake: retain Load query processing error if delete files succeeds

williamhbaker:
**Description:**

The deferred call to deleting staged load files would clobber any error returned
from the body of the Load function. This could result in load queries failing
and returning incomplete results, but the Load function returning with no error
if the deletion of staged load files succeeded and returned a `nil` error. This
may result in incorrectly reduced or possibly duplicated documents from the
resultant documents to store.

**Workflow steps:**

**Notes for reviewers:**

Also tacked on a couple small updates to fix other connector builds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2038)


Alex-Bair (approved):
LGTM

PR #2039, opened 2024-10-10
materialize-sql: support migration of non-nullable columns

mdibaiee:
**Description:**

**Workflow steps:**

**Notes for reviewers:**


materialize-sql/type_mapping.go
```diff
@@ -475,12 +475,19 @@ func (c constrainter) compatibleType(existing boilerplate.EndpointField, propose
 
 func (c constrainter) migratable(existing boilerplate.EndpointField, proposed *pf.Projection, rawFieldConfig json.RawMessage) (bool, *MigrationSpec, error) {
 	proj := buildProjection(proposed, rawFieldConfig)
+	// We check migratability against nullable DDLs, as we also cast
+	// the type to the nullable form of the DDL, however the final new column will be
+	// created as non-nullable if the proposed projection requests it
+	if proj.Inference.Exists == pf.Inference_MUST {
```
williamhbaker, line 481: `mapped` below already has both the `DDL` and `NullableDDL`. Instead of doing this re-assignment to always seeintg `Inference_MAY`, can we pass `mapped.NullableDDL` in instead of `mapped.DDL` on line 483?

williamhbaker (approved):
LGTM - I don't see a problem with migrated bigquery columns needing to be nullable. As a comparable case, we never even create Redshift columns as non-nullable because it doesn't all them to be made non-nullable, and that hasn't ever been an issue. A manual table re-backfill is always an option as well if a user wants the non-nullable columns.

materialize-sql/type_mapping.go
mdibaiee: ah yes, I wrote this before I realised we have `NullableDDL`, makes sense 👍🏽

PR #2042, opened 2024-10-10
source-postgres: Add 'Discover Only Published Tables' setting

willdonnelly:
**Description:**

This is an advanced option which, when set, instructs the capture to only return discovery results for tables which are already in the publication.

This is useful in cases where users might wish to administer the set of tables which are being captured by adding or removing the table from the publication in the source DB.

I implemented a test case for this, but since our CI test database is currently created with a `FOR ALL TABLES` publication it can't currently be added to the CI test suite.

**Workflow steps:**

By default nothing should change. When the advanced option "Discover Only Published Tables" is enabled, the capture will omit discovery bindings for any tables which aren't already present in the publication. This allows users to control which tables are to be captured by adding or removing them from the publication, and everything should propagate downstream smoothly from there.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2042)

### DIFF
--- a/source-postgres/.snapshots/TestGeneric-SpecResponse
+++ b/source-postgres/.snapshots/TestGeneric-SpecResponse
@@ -83,6 +83,11 @@
             "title": "Discovery Schema Selection",
             "description": "If this is specified only tables in the selected schema(s) will be automatically discovered. Omit all entries to discover tables from all schemas."
           },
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

PR #2043, opened 2024-10-11
materialize: don't use locks in the extended logger

williamhbaker:
**Description:**

The use of a mutex in both the extended logger event handler and the async logger functions was simple, but there was a potential deadlock if the event handler needed to stop the logger and wait for it to stop while holding the lock, and the logger timer ticked over at nearly the same instant making the logger need to get the lock also to finish its last log before exiting.

This updates the extended logger to not use locks at all, since it wasn't far from being able to do that in the first place. The "count" values that are read concurrently by the async loggers have been changed to use atomics, and for good measure the "round" counter is provided as an explicit argument to them, even though it doesn't strictly need to be, it's just easier to understand this way.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2043)

### DIFF
--- a/materialize-boilerplate/logging.go
+++ b/materialize-boilerplate/logging.go
@@ -3,6 +3,7 @@ package boilerplate
 import (
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


materialize-boilerplate/logging.go
```diff
@@ -144,11 +143,7 @@ func newExtendedLogger(ll loggerAtLevel, be *BindingEvents) *extendedLogger {
 // stop it. The asyncLog logging callback is run periodically and while holding
 // the mutex for extendedLogger. The doneLog callback does not acquire the lock.
 func (l *extendedLogger) logAsync(asyncLog func()) (stopLogger func(doneLog func())) {
-	return repeatAsync(func() {
-		l.mu.Lock()
-		asyncLog()
-		l.mu.Unlock()
-	}, loggingFrequency)
+	return repeatAsync(func() { asyncLog() }, loggingFrequency)
```
mdibaiee, line 146: Can be simplified to:
repeatAsync(asyncLog, loggingFrequency)

mdibaiee (approved):
LGTM % small nit

PR #2044, opened 2024-10-11
source-zendesk-support: omit tags from capture snapshot

Alex-Bair:
**Description:**

The [`/tags`](https://developer.zendesk.com/api-reference/ticketing/ticket-management/tags/#list-tags) endpoint returns the most popular tags used in the last 60 days. After 60 days, it's possible for existing tags to not be returned if they weren't used recently. Because we aren't going into Zendesk & using tags every once in a while, the capture snapshot has been failing in CI. Instead of making a reminder to go into Zendesk every 2 months and use a tag, I think it's easier to just remove tags from the capture snapshot.

**Workflow steps:**

**Notes for reviewers:**

Confirmed snapshot tests passed locally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2044)

### DIFF
--- a/source-zendesk-support/tests/snapshots/snapshots__capture__capture.stdout.json
+++ b/source-zendesk-support/tests/snapshots/snapshots__capture__capture.stdout.json
@@ -297,18 +297,6 @@
     }
   ],
   [
-    "acmeCo/tags",
-    {
-      "_meta": {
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #2045, opened 2024-10-14
materialize-sql: migrate to JSON type when possible

mdibaiee:
**Description:**

- Support migration of any type to JSON when the destination actually supports a distinct JSON type.
- There is a caveat here which I had not realised until now: in destinations where there is no distinct JSON type, from the perspective of the migration logic of materialize-sql, migrating to a `sql.STRING` and `sql.MULTIPLE` are the same. Currently we support migration of many types to just plain string, but if the intended type is actually JSON, we will end up with strings without quotations in the psuedo-JSON column. This also applies to the existing "compatible" stanzas. For example if we have a varchar column with a value of `hello`, and we add a boolean type to this field, we want to migrate from a varchar column, to a varchar column: nothing happens, we attempt to insert the boolean value to this varchar field and the destination will automatically cast it (hopefully). However this field cannot be read as a valid JSON, because string values do not have double quotes as they are expected in a JSON scalar string. Databricks and SQLServer are examples of this.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2045)


williamhbaker (approved):
LGTM

Issue #1884, opened 2024-09-03
SQL CDC: handle dropped and renamed tables


jgraettinger:
Currently when a table being captured by a SQL CDC connector is dropped or renamed, we immediately error out and require someone to manually increment the backfill counter or disable the binding before we'll allow progress again. But in practice, 99% of the time a dropped table is immediately recreated and a renamed table has some other staging table renamed to the previous name. So it's actually pretty dumb that we require the user to go push that button manually when we could just do it automatically for them.

We should instead automatically reset a table back to `"Pending"` when we see a drop or rename event which causes it to no longer exist, and then restart the capture task (probably at the end of the current stream-to-fence cycle, so that if multiple tables are dropped/renamed we get them all in a single restart). Once it comes back the capture will get caught up on replication and then start a new backfill of the table as though it were a newly added binding.

If that's all we do, then a dropped and **not** recreated table will still result in a `"table ___ is a configured binding of this capture, but doesn't exist or isn't visible with current permissions"` prerequisite validation error. I think this is probably still correct, because having a state where a configured binding of the capture can just silently not be doing anything seems pretty counterintuitive to the user, but I can see arguments either way.

One possible compromise would be to keep checking table visibility as a prerequisite when validating a task spec, but remove it from the capture and instead check table existence/visibility only when we go to actually start the backfill. This would mean that a missing table would cause regular task failures (which would show up in the UI) but **wouldn't** block ongoing replication. We could even get a step fancier and change the table-activation logic so that we don't error out until/unless we're caught up _and_ all ongoing backfills are complete if we really want to minimize the impact here of a dropped table.

willdonnelly:
One thing which occurs to me: once we have a concept of marking tables as no longer usable, we could fairly easily plug that into some of the MySQL "metadata incorrect" sanity checking paths so that the occasional metadata inconsistencies we see (which we currently believe are caused by explicitly unlogged DDL alteration queries) result in the offending table getting automatically reset and re-backfilled.

willdonnelly:
The PR for this has been merged and is fully rolled-out to all production tasks. Nothing has caught fire despite the major refactoring of the top-level capture loop, so I'm going to be optimistic and say we're done here for now.

PR #2052, opened 2024-10-15
sqlcapture: Support dropped and recreated tables

willdonnelly:
**Description:**

This PR does two closely related things. Firstly it restructures the top level of SQL CDC captures. Previously there was a linear progression from initialization through backfills to an infinite loop of streaming to the next fence, and after this PR the infinite loop now encompasses all of those operations. The upshot of that restructuring is that it's possible in principle for a table to be reset back to pending, get reinitialized, and begin backfilling without an intervening task restart.

The other major thing this PR does is that it introduces a concept of a `TableDropEvent` which is generated when a table is dropped or otherwise goes missing under its original name. In response to such an event, the SQL CDC machinery transitions the binding state to `"Missing"`. Periodically, the connector will re-run discovery, and if any missing tables are found they will be moved back into the `"Pending"` state, and from there the aforementioned changes to initialization and backfill behavior will cause it to be reactivated with a full backfill from the top.

This is then used by the MySQL and SQL Server captures. For MySQL a `TableDropEvent` is emitted any time an active table is either dropped or renamed. For SQL Server it's emitted any time there are no valid CDC instances for a particular source table (which often happens when the table was dropped and then recreated, but could in principle happen in other cases of unrecoverable failure).

We don't do anything special for Postgres because the Postgres logical replication protocol doesn't tell us about dropped or renamed tables (and building our own detection logic would be a ton of effort for no clear gain), but in the case of a simple drop-and-recreate Postgres just keeps on sending us the new changes from after the table is recreated so it's not as important to have a recovery strategy there anyway (although it's a bit annoying IMO that Postgres doesn't bother to even send us a synthetic `TruncateMessage` when this happens).

This fixes https://github.com/estuary/connectors/issues/1884

**Workflow steps:**

MySQL and SQL Server captures will no longer hard-fail when an active binding is dropped and recreated. Instead they will mark the table as missing, and if/when a table by that name reappears they will automatically reinitialize and re-backfill it. Once collection-level truncation is a thing and re-backfills emit the appropriate signals, the dropped-and-recreated table behaviors will automatically get those as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2052)


sqlcapture/capture.go
```diff
@@ -191,224 +203,80 @@ func (c *Capture) Run(ctx context.Context) (err error) {
...
-		if strings.Join(streamState.KeyColumns, ",") != strings.Join(primaryKey, ",") {
-			return fmt.Errorf("stream %q: primary key %q doesn't match initialized scan key %q", streamID, primaryKey, streamState.KeyColumns)
-		}
+		logrus.WithField("stateKey", stateKey).Info("binding added to capture")
+		c.State.Streams[stateKey] = &TableState{Mode: TableStatePending, dirty: true}
 	}
 
-	// Likewise streams may be removed from the catalog, and we need to forget
-	// the corresponding state information.
+	// When a binding is removed we change the table state to "Ignore
```
williamhbaker, line 279: Missing characters at the end of this comment?

williamhbaker (approved):

sqlcapture/capture.go
willdonnelly: Good catch, fixed.

PR #2053, opened 2024-10-15
materialize-mysql: use longtext instead of JSON for mariadb

mdibaiee:
**Description:**

- [MariaDB does not have a distinct `JSON` type, it is instead a `LONGTEXT` alias](https://mariadb.com/kb/en/json-data-type/), so for MariaDB instances from now on we actually create a `LONGTEXT` to avoid confusing the connector over compatibility of MULTIPLE fields.
- Added MariaDB to our tests so we can check that flow_documents is not modified when running an Apply RPC against a MariaDB instance. I ran the tests and can confirm we do not attempt to migrate `flow_document` with this update. MySQL instances will retain their previous behavior (also verified using our tests).

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2053)

### DIFF
--- a/materialize-mysql/sqlgen.go
+++ b/materialize-mysql/sqlgen.go
@@ -42,7 +42,13 @@ func identifierSanitizer(delegate func(string) string) func(string) string {
 	}
 }
 
-var mysqlDialect = func(tzLocation *time.Location, database string) sql.Dialect {
+var mysqlDialect = func(tzLocation *time.Location, database string, product string) sql.Dialect {
+	var jsonType = "JSON"
+	var jsonMapper = sql.MapStatic("JSON", sql.UsingConverter(sql.ToJsonBytes))
+	if product == "mariadb" {
+		jsonType = "LONGTEXT"
+		jsonMapper = sql.MapStatic("LONGTEXT", sql.UsingConverter(sql.ToJsonString))
+	}
 	mapper := sql.NewDDLMapper(
 		sql.FlatTypeMappings{
 			sql.INTEGER: sql.MapSignedInt64(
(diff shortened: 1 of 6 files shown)


materialize-mysql/sqlgen.go
```diff
@@ -51,10 +57,10 @@ var mysqlDialect = func(tzLocation *time.Location, database string) sql.Dialect
 			),
 			sql.NUMBER:   sql.MapStatic("DOUBLE PRECISION", sql.AlsoCompatibleWith("double")),
 			sql.BOOLEAN:  sql.MapStatic("BOOLEAN", sql.AlsoCompatibleWith("tinyint")),
-			sql.OBJECT:   sql.MapStatic("JSON"),
-			sql.ARRAY:    sql.MapStatic("JSON"),
+			sql.OBJECT:   sql.MapStatic(jsonType),
```
williamhbaker, line 60: Should we use the `jsonMapper` for object and array flat types as well? It seems to me like we should, so that objects and arrays materialized as stringified JSON in MariaDB in the same way as multiples type fields.

williamhbaker (approved):
LGTM % one question

materialize-mysql/sqlgen.go
mdibaiee: @williamhbaker I think it is not necessary, since OBJECT and ARRAY types will arrive as `json.RawMessage`, so jsonMapper is effectively no-op for these. Postgres materialization also has a similar mapping as mysql and it is working fine 🤔

PR #2054, opened 2024-10-15
source-postgres: Don't fetch column descriptions in system schemas

willdonnelly:
**Description:**

We already don't fetch the tables themselves from system schemas, so fetching column descriptions is pointless. And empirically, the part of the query expression that actually gets the description often fails with errors like `permission denied for schema cron` in production.

Hopefully if we explicitly filter out those schemas we'll see it succeed more often.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2054)


williamhbaker (approved):
LGTM

Issue #2056, opened 2024-10-15
source-mysql: Support collation `utf8_general_ci`


willdonnelly:
There's some historical reason stuff going on with that collation name, but for our purposes we don't care and should treat it as UTF-8 text just like we do `utf8mb3_foobar` and `utf8mb4_foobar` collations.

In fact that's already what we do, but only because we have a nice default behavior of "if we don't recognize it, assume it's UTF-8 compatible". So really adding this explicitly will just eliminate a benign error log, but that's still worth doing at some point.

willdonnelly:
According to a quick spot check, collations named `utf8_foobar` are the leading cause of MySQL CDC connector errors in production right now, so fixing this will make production error counts a lot less noisy.

PR #2058, opened 2024-10-15
source-mysql: Fix some benign error/warnings that get logged

willdonnelly:
**Description:**

The collation families `utf8_whatever` and `ascii_whatever` are explicitly added to the charsets table so we stop logging errors reading `unknown charset for collation, assuming UTF-8` and just do that without complaining because UTF-8 is in fact correct.

The `replication connected without TLS` log message is downgraded from WARN to INFO to match the main (non-replication) connection logging.

Between these fixes, that's all the main sources of errors and warnings in our production MySQL tasks, so hopefully that will be much less noisy to keep an eye on in the future.

This fixes https://github.com/estuary/connectors/issues/2056

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2058)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -478,7 +478,7 @@ func charsetFromCollation(name string) string {
 	//
 	// We rely on this assumption to identify known charsets based on the decoders table here.
 	for charset := range mysqlStringDecoders {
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #2061, opened 2024-10-16
materialize: fix round counter error in binding events logger

williamhbaker:
**Description:**

The round counter is incremented when the materialization starts its commit, and the general idea of starting the entire commit happens before performing the sub-operations for committing to each resource. So the "round" counter for committing to each resource is actually 1 less than currently tabulated.

The "round" counter offset by 1 is overall pretty convoluted and I may need to think about some kind of "load round" counter instead that stores are associated with, although I'm not sure that would be any better. The async operations of the materialization protocol just makes this a bit tricky

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2061)


Alex-Bair (approved):
LGTM

PR #2063, opened 2024-10-17
materialize: boilerplate validator recommends explicit projections among ambiguous fields

williamhbaker:
**Description:**

A common case for user-defined projections are to change the capitalization of a field. For systems that do not allow identical-except-for-capitalization named fields, this is unfortunate because the user-defined explicit projection is currently getting marked as optional even if it would otherwise be recommended, since the field names only differ in capitalization.

This change has us recommend explicit projections if they would otherwise be recommended even if their field name is ambiguous with others. It is relatively simple to do so since the Flow build process will not allow projected field names that differ only in capitalization, so we know if there is an explicit projection it's the only one with those case-insensitive characters.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2063)


psFried (approved):
LGTM

PR #2064, opened 2024-10-17
source-hubspot-native: custom object path components

Alex-Bair:
**Description:**

Some HubSpot endpoints like `/v3/properties/{objectType}` do not work for every custom object type. However, these endpoints do work if we prepend a `p_` to the beginning of the custom object name and use that in the path instead.

I chose to add `path_component` to the `crm_object_with_associations` function signature mainly so we only prepend `p_` in one spot rather than in two. I'm not dead-set on this approach though, and I can adjust if there's a better/more desireable solution.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed custom objects and their properties are retrieved from HubSpot.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2064)

### DIFF
--- a/source-hubspot-native/source_hubspot_native/resources.py
+++ b/source-hubspot-native/source_hubspot_native/resources.py
@@ -64,25 +64,31 @@ async def all_resources(
     ]
 
     custom_object_names = await list_custom_objects(log, http)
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM, thanks!

PR #2065, opened 2024-10-18
source-hubspot-native: add "FORWARD" to email event types

Alex-Bair:
**Description:**

An existing task is failing because email events it's receiving have a `FORWARD` type, which isn't allow in the model. This PR adds `FORWARD` as an allowed email event type.

Relevant HubSpot docs are [here](https://legacydocs.hubspot.com/docs/methods/email/email_events_overview). Skimming through them, we may need to add more email event types in the future (`PRINT` ? ), but that can be done later if needed.

Discover snapshot changes are expected since a new email event type was added to the `EmailEvent` model.

**Workflow steps:**

**Notes for reviewers:**

I haven't been able to test this locally yet due to issues with `flowctl` I'm hitting this morning, but this feels like a pretty clear cut issue & solution that could probably be safely merged without testing on a local stack. If you think otherwise, LMK and I'll hold off until I can test locally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2065)

### DIFF
--- a/source-hubspot-native/source_hubspot_native/models.py
+++ b/source-hubspot-native/source_hubspot_native/models.py
@@ -390,6 +390,7 @@ class EmailEvent(BaseDocument, extra="allow"):
         "BOUNCE",
         "OPEN",
         "CLICK",
+        "FORWARD",
         "STATUSCHANGE",
         "SPAMREPORT",
         "SUPPRESSED",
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #2067, opened 2024-10-18
source-sqlserver: Add another query to replication diagnostics

willdonnelly:
**Description:**

While debugging a SQL Server replication issue, I noticed that we're not currently capturing this as part of our diagnostics. (I don't think this would give us anything useful for the issue I was just debugging, but in other cases it would be good to have this information).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2067)


williamhbaker (approved):
LGTM

PR #2069, opened 2024-10-18
source-hubspot-native: add auto backfills for delayed streams

Alex-Bair:
**Description:**

Delayed incremental streams should be wrapped in a `try/except` block to trigger automatic backfills if we encounter a `MustBackfillBinding` error.

I kept the recent incremental streams wrapped in the existing `try/except` that serves the same purpose since it's possible for `MustBackfillBinding` exceptions to be raised if a task is disabled long enough for the cursor to fall far enough behind the recent changes available from HubSpot.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that auto backfills are triggered if the recent incremental `engagements` stream falls far enough behind by shifting the cutoff date back ~60 days. I didn't confirm that the same happens for delayed incremental streams since finding the right cutoff shift where the delayed stream raises a `MustBackfillBinding` exception but not the recent stream would some sifting through API responses. Since the `try/except` wrapper for both recent & delayed streams are the same, and the wrapper works for the recent stream, I'm fairly confident it'll work for the delayed stream.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2069)

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api.py
+++ b/source-hubspot-native/source_hubspot_native/api.py
@@ -368,20 +368,24 @@ async def process_changes(
     delayed_emitted = 0
     if delayed_fetch_next_end - delayed_fetch_next_start > delayed_fetch_minimum_window:
         # Poll the delayed stream for documents if we need to.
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2070, opened 2024-10-21
source-zendesk-support: checkpoint during `AuditLogs` stream

Alex-Bair:
**Description:**

For tasks that have a significant number of audit logs in Zendesk Support, it can take more than 24 hours to backfill. With the current `AuditLogs` stream strategy of reading records in descending order then stopping when we've reached a record we already read, we can't checkpoint until the backfill is completed. This means tasks that take more than 24 hours to backfill `AuditLogs` can get stuck restarting & attempting to backfill.

This PR updates the `AuditLogs` stream to read records in ascending order. This lets the connector checkpoint after each response & pick up where it left off when it's restarted.
Some noteable changes include:
- Passing a list containing `start_time` and `end_time` for the `filter[created_at][]` query param to bound the responses' results to the specified timespan.
- Pushing `end_time` 30 seconds in the past to avoid missing records if we query Zendesk in the middle of creating multiple records with the same `created_at`.
- Keeping `AuditLogs` on the older `get_updated_state` method of updating state instead of migrating to the newer `state` property. This is to avoid rebackfilling all existing tasks.

**Workflow steps:**

**Notes for reviewers:**

The upper time bound for the `AuditLogs` stream is moved a little in the past to avoid potential issues where we query Zendesk Support when its in the middle of creating multiple records with the same `created_at`. I don't have evidence that this can actually occur, but it sounds possible to me and a 30 second delay is relatively insignificant when the connector is running on 5 minute intervals.

Existing tasks do not need backfilled. The `AuditLogs` cursor value has remained the same (`created_at` for the most recent record) and the way the connector manages this state is the same.

- Audit log records are received in ascending order.
- State is checkpointed & documents are emitted after each response.
- Checkpointed `AuditLogs` state is used after restarting the connector during a backfill.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2070)

### DIFF
--- a/source-zendesk-support/source_zendesk_support/streams.py
+++ b/source-zendesk-support/source_zendesk_support/streams.py
@@ -754,54 +754,34 @@ class AuditLogs(SourceZendeskSupportCursorPaginationStream):
     
     This endpoint does not respect the start_time param. It requires two query params with the same name to filter by date.
     See https://support.zendesk.com/hc/en-us/community/posts/4859612547866-Audit-Log-API-error.
-    Instead, we mimic the TicketAudits stream strategy of not paginating further when the most recent page contains results we don't need.
     """
 
     response_list_name: str = "audit_logs"
     # audit_logs doesn't have the 'updated_by' field
     cursor_field = "created_at"
+    state_checkpoint_interval = 100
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2072, opened 2024-10-22
source-mysql: Silently ignore COMMIT query events

willdonnelly:
**Description:**

Apparently COMMIT query events (in a context where there aren't any uncommitted changes on non-transactional tables we care about) are the primary source of ops log spam from SQL CDC connectors at the moment.

I'm pretty confident our handling is correct, so let's stop logging those.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2072)


williamhbaker (approved):
LGTM

PR #2073, opened 2024-10-22
source-mysql: Set ReadTimeout/WriteTimeout on connection

willdonnelly:
**Description:**

Without these set, it's possible for the initial server connect to hang indefinitely. We've seen this in production when going through an SSH tunnel (which I assume silently broke during the connection attempt), with a hang of over 8 hours.


willdonnelly:
Note: We already specify a ReadTimeout in the BinlogSyncerConfig for replication (and there is no WriteTimeout setting there), only the "normal" connection was lacking these.

mdibaiee (approved):

PR #2080, opened 2024-10-23
source-mysql: Connection read/write timeouts redux

willdonnelly:
**Description:**

[Previously](github.com/estuary/connectors/pull/2074) when we set read/write timeouts on the MySQL connection we saw a bunch of captures in production consistently fail exactly one timeout-duration after they connected.

I have sinced debugged this (see the comment which makes up the majority of this commit's lines changed) and believe that I have a viable workaround while we try to get the client library fixed.

Note that even with this workaround, we won't have timeouts on subsequent database communications when using TLS until the bug is fixed upstream (and we've upgraded the dependency version). Once it's fixed upstream this workaround is basically a no-op and can be removed at our leisure.

Even without timeouts on subsequent interactions, the timeout is still useful to have since it prevents indefinite hangs during the initial database handshake, which is something we've seen happen pretty frequently in production.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2080)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -217,6 +217,11 @@ func (db *mysqlDatabase) connect(_ context.Context) error {
 		c.SetTLSConfig(&tls.Config{InsecureSkipVerify: true})
 		return nil
 	}
+	var withTimeouts = func(c *client.Conn) error {
+		c.ReadTimeout = 60 * time.Second
+		c.WriteTimeout = 60 * time.Second
+		return nil
+	}
 	// The following if-else chain looks somewhat complicated but it's really very simple.
 	// * We'd prefer to use TLS, so we first try to connect with TLS, and then if that fails
 	//   we try again without.
@@ -224,12 +229,12 @@ func (db *mysqlDatabase) connect(_ context.Context) error {
 	// * Otherwise we report both errors because it's better to be clear what failed and how.
 	// * Except if the non-TLS connection specifically failed because TLS is required then
 	//   we don't need to mention that and just return the with-TLS error.
-	if connWithTLS, errWithTLS := client.Connect(address, db.config.User, db.config.Password, db.config.Advanced.DBName, withTLS); errWithTLS == nil {
+	if connWithTLS, errWithTLS := client.Connect(address, db.config.User, db.config.Password, db.config.Advanced.DBName, withTimeouts, withTLS); errWithTLS == nil {
 		logrus.WithField("addr", address).Info("connected with TLS")
 		db.conn = connWithTLS
 	} else if errors.As(errWithTLS, &mysqlErr) && mysqlErr.Code == mysql.ER_ACCESS_DENIED_ERROR {
 		return cerrors.NewUserError(mysqlErr, "incorrect username or password")
-	} else if connWithoutTLS, errWithoutTLS := client.Connect(address, db.config.User, db.config.Password, db.config.Advanced.DBName); errWithoutTLS == nil {
+	} else if connWithoutTLS, errWithoutTLS := client.Connect(address, db.config.User, db.config.Password, db.config.Advanced.DBName, withTimeouts); errWithoutTLS == nil {
 		logrus.WithField("addr", address).Info("connected without TLS")
 		db.conn = connWithoutTLS
 	} else if errors.As(errWithoutTLS, &mysqlErr) && mysqlErr.Code == mysql.ER_ACCESS_DENIED_ERROR {


mdibaiee:
@willdonnelly your upstream PR has been merged, so I think we can now just update the client pin instead of this workaround? https://github.com/go-mysql-org/go-mysql/pull/925

willdonnelly:
Yeah, I'm going to upload a separate PR to bump the client library to that commit. Once that's live (and we've given it a little while to be sure it didn't break anything else in production that might force us to revert) I'll remove the workaround and we can merge this as just the original change to set the timeouts.

willdonnelly:
This conflicts very slightly with the changes in https://github.com/estuary/connectors/pull/2097 so I'll have to make sure they both land correctly, but otherwise I think we're about ready to try these timeouts again. The upstream client library bug has been fixed, we've updated to the latest commit of that dependency, and nothing caught fire today so it's unlikely to require a revert.

source-mysql-batch/main.go
```diff
@@ -80,24 +80,35 @@ func connectMySQL(ctx context.Context, cfg *Config) (*client.Conn, error) {
...
 	if connWithTLS, errWithTLS := client.Connect(cfg.Address, cfg.User, cfg.Password, cfg.Advanced.DBName, withTLS); errWithTLS == nil {
 		log.WithField("addr", cfg.Address).Info("connected with TLS")
 		conn = connWithTLS
+	} else if errors.As(errWithTLS, &mysqlErr) && mysqlErr.Code == mysql.ER_ACCESS_DENIED_ERROR {
+		return nil, cerrors.NewUserError(mysqlErr, "incorrect username or password")
 	} else if connWithoutTLS, errWithoutTLS := client.Connect(cfg.Address, cfg.User, cfg.Password, cfg.Advanced.DBName); errWithoutTLS == nil {
 		log.WithField("addr", cfg.Address).Info("connected without TLS")
 		conn = connWithoutTLS
+	} else if errors.As(errWithoutTLS, &mysqlErr) && mysqlErr.Code == mysql.ER_ACCESS_DENIED_ERROR {
+		log.WithField("withTLS", errWithTLS).Warn("connecting with TLS failed for an unrelated reason")
```
mdibaiee, line 106: nit: I would reword this a little bit to avoid ambiguity for users about the "unrelated" error, although my suggestion is a bit of a mouthful, feel free to use whatever you think suits
```suggestion
		log.WithField("withTLS", errWithTLS).Warn("connecting with TLS failed for a non-TLS-related reason")
```

source-mysql/main.go
```diff
@@ -211,24 +211,39 @@ func (db *mysqlDatabase) connect(_ context.Context) error {
...
 		logrus.WithField("addr", address).Info("connected with TLS")
 		db.conn = connWithTLS
-	} else if connWithoutTLS, errWithoutTLS := client.Connect(address, db.config.User, db.config.Password, db.config.Advanced.DBName); errWithoutTLS == nil {
+	} else if errors.As(errWithTLS, &mysqlErr) && mysqlErr.Code == mysql.ER_ACCESS_DENIED_ERROR {
+		return cerrors.NewUserError(mysqlErr, "incorrect username or password")
+	} else if connWithoutTLS, errWithoutTLS := client.Connect(address, db.config.User, db.config.Password, db.config.Advanced.DBName, withTimeouts); errWithoutTLS == nil {
 		logrus.WithField("addr", address).Info("connected without TLS")
 		db.conn = connWithoutTLS
+	} else if errors.As(errWithoutTLS, &mysqlErr) && mysqlErr.Code == mysql.ER_ACCESS_DENIED_ERROR {
+		logrus.WithField("withTLS", errWithTLS).Warn("connecting with TLS failed for an unrelated reason")
```
mdibaiee, line 241:
```suggestion
		logrus.WithField("withTLS", errWithTLS).Warn("connecting with TLS failed for a non-TLS-related reason")
```

mdibaiee (approved):
LGTM, left two nits that are up to you

source-mysql-batch/main.go
willdonnelly: That's a good point, though it should probably be addressed in the other PR -- this one just includes that diff because I wanted to get the merge conflict between the two resolved ahead of time.

willdonnelly:
The spurious diffs have been resolved now that the other MySQL connections PR is merged. Going to hold off on this for a little bit while we make sure nothing goes wrong from the other connection changes, and then merge this one.

PR #2082, opened 2024-10-24
dbt: move to materialize-boilerplate, only trigger after delay

mdibaiee:
**Description:**

- Move dbt job trigger logic to materialize-boilerplate where it can be better integrated with sync frequency logic, and makes it easier for materialization connectors to integrate the functionality in general

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2082)

### DIFF
--- a/materialize-boilerplate/transactions_stream.go
+++ b/materialize-boilerplate/transactions_stream.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"slices"
+	"sync/atomic"
 	"time"
 
 	"github.com/estuary/connectors/go/dbt"
@@ -59,7 +60,7 @@ type transactionsStream struct {
 	storedHistory []int
 
 	dbtJobTriggerConfig *dbt.JobConfig
-	dbtJobTriggerCh     chan bool
+	dbtJobTriggerCh     chan struct{}
 }
 
 // newTransactionsStream wraps a base stream MaterializeStream with extra
@@ -71,7 +72,7 @@ func newTransactionsStream(
 	options MaterializeOptions,
 	be *BindingEvents,
 ) (*transactionsStream, error) {
-	s := &transactionsStream{stream: stream, ctx: ctx, dbtJobTriggerCh: make(chan bool)}
+	s := &transactionsStream{stream: stream, ctx: ctx, dbtJobTriggerCh: make(chan struct{})}
 
 	var loggingHandler func(transactionsEvent)
 	if options.ExtendedLogging {
@@ -105,17 +106,12 @@ func newTransactionsStream(
 		s.handler = loggingHandler
 	}
 
-	// NOTE: This delay is set based on the default value of `maxTxnDuration` being 5 minutes, which means that we are guaranteed
-	// to receive a new transaction from the runtime if data exists.
-	// If a task has `maxTxnDuration` higher than 5 minutes, and there has been no delay in the last transaction, we will
-	// trigger a dbt job 5 minutes after the last transaction. To avoid this users can set a different SafetyTriggerDelay to a value
-	// higher than the maxTxnDuration
-	delay := time.Duration(5*time.Minute + 15*time.Second)
-	if s.dbtJobTriggerConfig != nil && s.dbtJobTriggerConfig.SafetyTriggerDelay != "" {
+	delay := time.Duration(30 * time.Minute)
+	if s.dbtJobTriggerConfig != nil && s.dbtJobTriggerConfig.Interval != "" {
 		var err error
-		delay, err = time.ParseDuration(s.dbtJobTriggerConfig.SafetyTriggerDelay)
+		delay, err = time.ParseDuration(s.dbtJobTriggerConfig.Interval)
 		if err != nil {
-			return nil, fmt.Errorf("parsing dbt safety trigger delay: %w", err)
+			return nil, fmt.Errorf("parsing dbt interval: %w", err)
 		}
 	}
 	go s.dbtJobTriggerHandler(ctx, delay)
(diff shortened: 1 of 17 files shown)


materialize-boilerplate/transactions_stream.go
```diff
@@ -199,6 +206,19 @@ func (l *transactionsStream) maybeDelayAcknowledgement() error {
 				return l.ctx.Err()
 			case <-time.After(d):
 			}
+
+			// We trigger dbt job only if the delay is taking effect, to avoid bursting
+			// dbt job triggers during backfills
```
mdibaiee, line 211: One question worth posing here: have cases where a materialization has so much data that it always skips the default delay? In those cases it might seem like dbt job trigger is not working at all, but it is being always postponed... 🤔

williamhbaker: There are definitely active tasks where a materialization has some much continuous data that when combined with the backpressure of the destination system, the transactions are always greater than 1MM documents and the delay would always be skipped.

Fortunately these are also cases where the tasks are configured with an explicit delay of `0s`, so it doesn't really matter. I doubt this will always be the case though. Our current delay triggering mechanism based on the number of documents in a transaction is fundamentally flawed but it is the best we can do right now.

williamhbaker (commented):
Here's a scenario to consider: A user may have a batch capture that runs every 24 hours that produces some large number of documents only once a day but never at any other time. They are materializing from that capture, so the materialization only does any work every 24 hours. When a materialization first starts up, it runs a fixed number of commits (5) without applying any delay, to avoid accidentally tripping the delay due to artificially small commits from the runtime that sometimes happen when a connector restarts. If the materialization connector restarts in that ~24 hour period of idleness, it will have 5 instant commits queued up for when it starts seeing data. If it is able to work through the batch capture's 24 hours worth of data that it captures all at once in 5 or less transactions it won't ever trigger the update delay, and thus would never trigger the dbt job. It might eventually trigger the dbt job during the next 24 hour cycle, or it might not even then if the materialization is restarted again in the idle period.

mdibaiee:
@williamhbaker please see my last commit, this commit is an attempt to cover the case you described of tasks that may run for a short time every day, getting through their data in less than 4 commits.

I wanted to note that I found this case to be hard to reproduce, it only happened when I manually specified a long enough `minTxnDuration` to prevent the runtime from sending a bunch of transactions and only send large ones. But there may be real-world examples where this happens. Still, I would like to question whether those cases are common enough to justify the complexity added here 🤔

Now there is another edge case: this solution is based on the fact that `maxTxnDuration: 5m` by default. If `maxTxnDuration` is higher and the runtime accumulates data for longer than 5 minutes, and the data is so much that it skips delay, this implementation will trigger a dbt job in the background 5 minutes after the last transaction. If the connector tends to consistently skip delay due to large transactions and has a long `maxTxnDuration`, the user should most likely not use a delay at all.

Still, I've added an escape hatch for this case with a configuration that is not exposed by default at the moment. I've hidden it mostly because I have not been able to come up with a concise way to name and describe this feature without being confusing to users setting up dbt. I've called it the Safety Trigger... but that's pretty vague. Open to feedback / suggestions on naming and description and whether we should expose it as a UI form field or keep it as a private, at least for now.

Perhaps I can document this in our docs and explain how it interacts with the Sync Frequency, min and maxTxnDuration and hope that users will not touch the defaults until they hit an issue and we can then point them to docs.

materialize-boilerplate/transactions_stream.go
```diff
@@ -129,6 +190,11 @@ func (l *transactionsStream) storedHistoryHandler(delegate func(transactionsEven
 }
 
 func (l *transactionsStream) Send(m *pm.Response) error {
+	// Every time there is some activity, we send a signal to
+	// the dbt job trigger channel to postpone the trigger another 5 minutes
+	// we only want to trigger the job after minutes of inactivity by the connector
+	l.dbtJobTriggerCh <- true
```
williamhbaker, line 196: I think the dbt job triggering should only occur when the connector is sending an acknowledgement that it has committed some data, so down around line 205, prior to the `maybeDelayAcknowledgement`.

Regarding the edge case about the default 5 minute maximum transaction duration, I think it's pretty important that we aren't restricted by that and have to introduce any kind of explanations for how that interacts with dbt job triggering. Transactions can also be spread out longer than 5 minutes for various other reasons, like if load queries / processing loaded documents takes a long time, or if an actual commit to the destination takes a long time. I am also quite convinced that this dbt job scheduling should not be associated with the delays introduced by the sync frequency at all. Some materializations that support dbt job triggers don't have sync frequency configuration, sometimes users will set it to 0, and ultimately I think the entire ack delay mechanism [should probably live in the runtime](https://github.com/estuary/connectors/issues/1182), although the steps to get there aren't clear right now.

So here's what I'd propose, which I think builds on the async goroutine mechanism you've got here, and would not be constrained by any assumptions about how frequently commits are occuring:

Every time the materialization sends an acknowledgement but _before_ applying an acknowledgement delay, it should send a signal on `dbtJobTriggerCh`, for example between the current line 204 and 205.

Generally the `dbtJobTriggerHandler` would need to do this:
1. Trigger the dbt job right away if one hasn't been run in the current connector's running time. Note that this will cause a dbt job to be triggered the first time a connector starts up. But I actually think this is necessary to prevent cases where a delayed background dbt trigger never gets to run because the connector restarts before it does.
3. Trigger the dbt job right away if the last one was ran more than **dbtRunInterval** ago (I'd suggest hard-coding this at 30 minutes, and potentially make it configurable if users ask for it)
4. If the last dbt job ran within **dbtRunInterval** ago, schedule the next run **dbtRunInterval** after the last one ran, but...
5. ...if there is _already_ a job scheduled, don't do anything.

Specifically I think something like this would work, where the `dbtJobTriggerCh` has been modified to a `chan struct{}`:

```go
func (l *transactionsStream) dbtJobTriggerHandler(ctx context.Context, dbtRunInterval time.Duration) {
	var lastRun time.Time
	var nextRunScheduled atomic.Bool

	runJob := func() {
		lastRun = time.Now()
		if err := l.triggerDBTJob(); err != nil {
			log.WithField("err", err).Error("dbt job trigger failed")
		}
		nextRunScheduled.Store(false)
	}

	for {
		select {
		case <-l.dbtJobTriggerCh:
			if nextRunScheduled.Load() {
				// Another dbt job is set to run already.
			} else if time.Since(lastRun) >= dbtRunInterval {
				// It has been a long time since the last job ran; trigger it
				// now.
				runJob()
			} else {
				// The last job was ran recently. Don't run another job right
				// away, but make sure one does eventually get run.
				nextRunScheduled.Store(true)

				nextRunTime := lastRun.Add(dbtRunInterval)
				go func() {
					select {
					case <-time.After(time.Until(nextRunTime)):
						runJob()
					case <-ctx.Done():
						return
					}
				}()
			}
		case <-ctx.Done():
			return
		}
	}
}
```

williamhbaker (approved):
LGTM

PR #2085, opened 2024-10-24
network-tunnel: exit without logging an error on EOF before READY

mdibaiee:
**Description:**

Sister pull-request of https://github.com/estuary/network-tunnel/pull/9 to surface the underlying SSH client error when network-tunnel fails to establish a SSH tunnel

Example output:

image

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2085)

### DIFF
--- a/go/network-tunnel/network_tunnel.go
+++ b/go/network-tunnel/network_tunnel.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"syscall"
 
+	cerrors "github.com/estuary/connectors/go/connector-errors"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -112,7 +113,7 @@ func (t *SshTunnel) Start() error {
 		// an error that we would like to surface to the user without logging anything further
 		// ourselves
 		if err == io.EOF {
-			os.Exit(1)
+			return cerrors.NewTransparentError(err)
 		}
 		return fmt.Errorf("reading READY signal from network-tunnel: %w", err)
 	} else if !bytes.Equal(readyBuf, ready) {
--- a/go/network-tunnel/network_tunnel.go
+++ b/go/network-tunnel/network_tunnel.go
@@ -108,8 +108,14 @@ func (t *SshTunnel) Start() error {
 	var readyBuf = make([]byte, 5)
 	var ready = []byte("READY")
 	if _, err = io.ReadFull(stdout, readyBuf); err != nil {
+		// EOF means the underlying process exit without writing READY, this is most likely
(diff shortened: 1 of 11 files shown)


go/network-tunnel/network_tunnel.go
```diff
@@ -108,8 +108,14 @@ func (t *SshTunnel) Start() error {
 	var readyBuf = make([]byte, 5)
 	var ready = []byte("READY")
 	if _, err = io.ReadFull(stdout, readyBuf); err != nil {
+		// EOF means the underlying process exit without writing READY, this is most likely
+		// an error that we would like to surface to the user without logging anything further
+		// ourselves
+		if err == io.EOF {
```
williamhbaker, line 114: We're doing something similar in filesource captures where the parser exits with an error and we don't want to clobber that with a less useful connector error, using a specific error type that is handled in the boilerplate, see [here](https://github.com/estuary/connectors/blob/cac47751d4c45fe1c241aa4e50b625e62a3a4129/filesource/filesource.go#L424-L434) for where that happens. I think it would be good to do that here as well.

mdibaiee: @williamhbaker that required us to capture the error line from the underlying process, so I added a bit of extra machinery to capture the last error from stderr of network-tunnel and use that to bubble up a transparent error. I kept the stderr piping just in case for now. The result looks like this now:

image

williamhbaker: :+1: I think it would also be fine to return the EOF error as `cerrors.NewTransparentError(err)` and not worry about capturing the last log line as an `error`. Since it is already being logged to stderr we shouldn't need to also retain its text in the Go program I don't think.

The actual `err` value that is wrapped is not [currently actually used](https://github.com/estuary/connectors/blob/fb99aa2a33ce056ab23f65cd971e35f5a169d1a0/go/connector-errors/errors.go#L57-L80) (maybe it shouldn't be part of TransparentError at all), in the case of the filesource capture it is a generic "Subprocess exited with non-zero code" and we are trusting the subprocess to log the actual useful error to stderr on its own.

williamhbaker (approved):
LGTM

go/network-tunnel/network_tunnel.go
mdibaiee: ah I see, I was testing using `go test -v` but there the `TransparentError` kept coming up, but I just tested using our integration tests and there is no need for us to capture the last error, so I removed those parts and I'm now simply returning the `TransparentError(EOF)`

PR #2086, opened 2024-10-24
source-bigquery-batch: Fix behavior of `DATETIME` cursors with fractional-second precision

willdonnelly:
**Description:**

BigQuery DATETIME columns only support microsecond precision, but the default serialization of the Go `civil.DateTime` struct uses nanosecond precision. Which is fine for output only, and *also* is fine for cursors until the capture restarts, and *also* is fine when the datetime values are whole seconds.

But a fractional-second datetime used as a cursor will be stored as a string in the state checkpoint, and trying to feed that back to BigQuery after a restart will produce an error because there are too many digits of precision.

The fix is to handle the formatting of `civil.DateTime` values into strings ourselves and make sure it only has six digits.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2086)

### DIFF
--- a/source-bigquery-batch/.snapshots/TestDatetimeCursor-Capture
+++ b/source-bigquery-batch/.snapshots/TestDatetimeCursor-Capture
@@ -1,0 +1,11 @@
+# ================================
+# Collection "acmeCo/test/datetime_cursor_132448": 3 Documents
+# ================================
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

Issue #1211, opened 2024-01-25
source-kafka: update connector to support state keys


williamhbaker:
We should update source-kafka to support state keys for re-backfilling the reading of a topic.

We should also think about all the other cross-cutting things we want source-kafka to do with respect to other technical connectors and see if it still makes sense for it to be in rust, where almost every interesting technical connector is in go.

Issue #29, opened 2021-09-08
Support Parsing Avro Documents


saterus:
It's rather common in the Kafka ecosystem to encode Message payloads with Apache Avro. However, Avro documents are not self-describing, they require a external schema definition to be deserialized.

TODO:
- [ ] Support passing a schema definition to the `parser`
- [ ] Parse Avro documents and reformat them as JSON
- [ ] Add support for converting Avro schema definitions to JSON schema definitions

Out of Scope:
- Fetching an Avro schema from the Confluent Schema Registry

PR #2089, opened 2024-10-25
source-kafka: support schema registry and Avro decoding

williamhbaker:
**Description:**

Re-works the capture to support connecting to a Schema registry, which enables parsing Avro messages and discovering collection keys based on an available topic key schema.

The first few commits are a general refactoring of the existing code. The ones after that implement the schema registry, discovery, and message parsing additions. There are a few other improvements thrown in with that, as detailed in the individual commits.

I restructured things pretty comprehensively, so there's not a whole lot left of the original connector. These changes are _not_ backward compatible, but I don't think this will be a real problem considering the existing Kafka captures we have running.

Closes https://github.com/estuary/connectors/issues/29
Closes https://github.com/estuary/connectors/issues/1211
Closes https://github.com/estuary/connectors/issues/2137

**Workflow steps:**

The Kafka source connector documentation needs to be updated to describe the schema registry configuration, and how collections are discovered based on available key schemas.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2089)


willdonnelly (approved):
LGTM

I read through the code and didn't see any obvious issues, but this is such a large refactoring that my LGTM is still more of a rubber-stamp than I'd prefer.

But since you say this is breaking backwards compatibility anyway, and I see a grand total of one Kafka capture in production that's actually green, I think that's sufficient.

PR #2092, opened 2024-10-28
materialize-mongodb: preserve collection `_id` field values as `_flow_id` in materialized documents

williamhbaker:
**Description:**

A collection with a present field of `_id` in its documents would lose that value when materializing to MongoDB, which has its own field for documents called `_id`.

That's not ideal in general, but it would cause specific problems with a standard updates materialization where the loaded document would be missing the `_id` field value that is required by the source collection schema for reductions.

The fix here is to materialize a present `_id` field from the Flow collection under an alternate field name `_flow_id` in stored documents, and reverse that process when loading documents for standard updates by replacing the loaded document's `_id` with the `_flow_id` and then deleting `_flow_id`.

Did a test on a local stack by trying to materialize a collection captured from `source-mongodb` and materialize that with standard updates back to `materialize-mongodb`: Prior to this change, the materialization task would fail with a schema violation error since loaded documents were missing the required `_id` field. But after this change the data flow works as expected, with updates to the source being materialized in the destination.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2092)

### DIFF
--- a/materialize-mongodb/transactor.go
+++ b/materialize-mongodb/transactor.go
@@ -17,7 +17,8 @@ import (
 )
 
 const (
-	idField = "_id"
+	idField    = "_id"
+	idFieldAlt = "_flow" + idField
 
 	batchByteLimit = 5 * 1024 * 1024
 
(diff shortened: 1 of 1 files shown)


mdibaiee (approved):
LGTM, I think we should add this to the docs of materialize-mongodb, perhaps as an FAQ

PR #2095, opened 2024-10-28
estuary-cdk: run `CaptureShim` as a coroutine

Alex-Bair:
**Description:**

Previously, imported Airbyte connectors that utilize the `CaptureShim` class ignored the `SIGQUIT` mechanism used to log stack traces containing all of the currently running coroutines. This was because `CaptureShim` was not _actually_ run as a coroutine.

To resolve this, the `Task.spawn_child` method is used to run a `CaptureShim` as a coroutine and an `await asyncio.sleep(0)` was added to provide a safe stopping point when a `SIGQUIT` signal is set. Also, logic was also added to `CaptureShim._run` to allow imported Airbyte connectors to perform graceful exits triggered by the `periodic_stop` coroutine.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that for imported Airbyte connectors:
- Sending a `SIGQUIT` logs currently running coroutines, kills the container, and fails the shard; all similar to how the `SIGQUIT` mechanism works for native connectors.
- Migrating from a previous version of the CDK to the newer version works without errors & state is maintained.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2095)

### DIFF
--- a/estuary-cdk/estuary_cdk/shim_airbyte_cdk.py
+++ b/estuary-cdk/estuary_cdk/shim_airbyte_cdk.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 from logging import Logger
 from pydantic import Field
 from typing import Any, ClassVar, Annotated, Callable, Awaitable, List, Literal
+import asyncio
 import logging
 import os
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2096, opened 2024-10-28
source-mysql: Normalize invalid dates the same as timestamp date parts

willdonnelly:
**Description:**

This commit factors out the date-normalization we apply to the date portions of timestamps and datetimes, and applies that same logic to the values of date columns.

For valid values this makes no difference. For the zero value this also makes no difference. But MySQL, while it normally rejects dates like `2024-00-00` or `2024-10-00` as invalid, has some mode settings which allows those, and currently we capture them verbatim. Which is a problem because the month or date components of a `format: date` string cannot be zero, and while we don't _currently_ discover `date` columns as a `format: date` string that's only because of compatibility issues and is something we'd like to change in the near future (and also we currently recommend that people add it to the read schema themselves if they need it ASAP), so we should try to satisfy that format anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2096)

### DIFF
--- a/source-mysql/.snapshots/TestDateNormalization
+++ b/source-mysql/.snapshots/TestDateNormalization
@@ -1,0 +1,16 @@
+# ================================
+# Collection "acmeCo/test/test_datenormalization_44310403": 8 Documents
+# ================================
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

Issue #2075, opened 2024-10-22
source-mysql: Better connection error reporting


willdonnelly:
Reporting MySQL connection errors is a giant pain because:

1. We can either attempt to connect with TLS or without TLS, not "preferring TLS".
2. Some server configurations will reject a TLS connection attempt with an error saying that TLS is forbidden, while other server configurations will reject a non-TLS connection attempt with an error saying that TLS is mandatory.

Currently we prioritize reporting the with-TLS error when both attempts fail, but this will fail in some cases. For instance, a server which forbids TLS and a config with the wrong password -- ideally we should report "incorrect username/password" but instead right now we report the "TLS is forbidden" error instead.

To actually do this correctly we need to write logic to detect when a connection error is saying that we have to enable/disable TLS and report the other error, and we haven't written that yet. This is a tracking issue so we can prioritize that work. It's not really that complicated (especially since a best-effort attempt could hardly be worse than the status quo so long as it does something vaguely sensible when the special cases fail), we just need someone to sit down and do it.

PR #2097, opened 2024-10-28
source-mysql: Improve connection error reporting

willdonnelly:
**Description:**

This PR modifies the MySQL connection / error reporting logic so that it should hopefully be a bit clearer in certain edge cases why the connection failed. Specifically:
* We'll now check both the with-TLS and without-TLS error for an "incorrect username/password" condition, where previously it was just the with-TLS error getting checked.
* When none of the special cases apply, we'll report a more verbose error which includes both the with-TLS and without-TLS failure reasons.
* There is an explicit special case for the "TLS is required" error, and only when that case is hit do we report just the with-TLS failure and omit the "TLS required" error.

I'm optimistic that maybe this will be adequate, finally.

This fixes https://github.com/estuary/connectors/issues/2075

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2097)


williamhbaker (approved):
LGTM

PR #2104, opened 2024-10-30
source-mysql: Fix BINARY(N) inconsistency

willdonnelly:
**Description:**

A `BINARY(N)` column holds fixed-length byte data, with zero- padding to the specified length. This all works as expected for backfills, but the actual binlog data stores the value with any trailing zeroes stripped so that's what we get via replication.

This commit adds a new datatype test case verifying the behavior around trailing zero bytes and implicit padding (and just for good measure it also makes the nonzero bytes an invalid UTF-8 byte sequence), modifies discovery to keep track of the length of a `BINARY(N)` column, and modifies the value translation code to zero-pad as necessary.

**Workflow steps:**

Bindings which are initialized after this change will discover and initialize the column metadata with a length property for binary columns, which is used to zero-pad when needed.

Preexisting bindings will not have this length metadata, so the old value translation logic has been left in place as well and will still apply the inconsistent trailing-zeroes behavior unless/until the offending bindings are backfilled again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2104)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -305,6 +305,10 @@ func (db *mysqlDatabase) translateRecordField(isBackfill bool, columnType interf
 				}
 				return acc, nil
 			case "binary", "varbinary":
+				// Unlike BINARY(N) columns, VARBINARY(N) doesn't play around with adding or removing
+				// trailing null bytes so we don't have to do anything to correct for that here. The
+				// "binary" case here is a fallback for captures whose metadata predates the change
+				// to discover binary columns as a complex type with length property.
 				if len(val) > truncateColumnThreshold {
 					val = val[:truncateColumnThreshold]
 				}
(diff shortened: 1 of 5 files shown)


willdonnelly:
Whoops, just realized I forgot the corresponding changes to the replication DDL alteration handling. I'll get that added and then this will be ready for review.

source-mysql/discovery.go
```diff
@@ -305,6 +305,10 @@ func (db *mysqlDatabase) translateRecordField(isBackfill bool, columnType interf
 				}
 				return acc, nil
 			case "binary", "varbinary":
+				// Unlike BINARY(N) columns, VARBINARY(N) doesn't play around with adding or removing
```
williamhbaker, line 308: I'm kind of fuzzy on how the value resolution is all plumbed together, but am a little confused by the comment about a fallback for metadata predating the change to include the length for binary columns. The metadata gets requested anew every time the connector starts up right? Or maybe there's some other persistence mechanism I'm forgetting about right now?

williamhbaker (approved):
LGTM % one question

source-mysql/discovery.go
willdonnelly: Replication decoding metadata is persisted in the state checkpoint between runs. It's originally initialized with the type information from discovery* and then thereafter it's updated only by DDL queries. So the concern here is metadata which predates this change and so doesn't have the binary type with length information, and won't unless/until it's either reinitialized by a backfill or altered in a DDL statement.

* Strictly speaking it is incorrect to do this because there's a data race if a table is altered right around the same time. But there is literally no alternative and in practice it works pretty reliably because people generally aren't altering their tables during that short window, and it only happens once (per binding and backfill counter increment).

Issue #2114, opened 2024-11-01
source-klaviyo: `Profiles` and `GlobalExclusions` streams silently fail when receiving a `400` response


Alex-Bair:
I suspect these [lines ](https://github.com/estuary/connectors/blob/e8da845f4bc6529d7a5fe69d3617c7637f461174/source-klaviyo/source_klaviyo/streams.py#L40-L55) somehow contribute to `Profiles` and `GlobalExclusions` ignoring `400` responses & silently failing, but more investigation is needed to confirm & determine an appropriate fix.

Another lead is these [lines](https://github.com/estuary/connectors/blob/e8da845f4bc6529d7a5fe69d3617c7637f461174/source-klaviyo/source_klaviyo/source.py#L62-L63).

PR #2115, opened 2024-11-01
source-klaviyo: filter `Profiles` with > instead of >=

Alex-Bair:
**Description:**

My earlier commit (dfbe8b5714bf76b053f971d5352e95d26f471d08 in PR #2037 ) was too broad in changing all incremental streams to use `>=` in the `filter` query param. The `/profiles` endpoint used by both the `Profiles` and `GlobalExclusions` streams do not support filtering with `>=`, so they have to use `>`.

For streams that do use `>`, I subtracted one second (the smallest time granularity for Klaviyo timestamps) before the first request to avoid missing records that were updated in the same second but weren't available in the API at the exact same time.

**Workflow steps:**

**Notes for reviewers:**

Although `/profiles` does not support `>=` filtering & the API returns a `400` response when trying to filter with `>=`, existing tasks have failed silently. I'm not sure exactly _why_ tasks didn't fail / an error message wasn't logged, but I made issue #2114 to track this.

- Previously working incremental streams still work incrementally.
- `Profiles` and `GlobalExclusions` streams now work, are incremental, and no longer silently fail.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2115)

### DIFF
--- a/source-klaviyo/source_klaviyo/streams.py
+++ b/source-klaviyo/source_klaviyo/streams.py
@@ -154,6 +154,10 @@ class IncrementalKlaviyoStream(KlaviyoStream, ABC):
         :return str: The name of the cursor field.
         """
 
+    # comparison_operator is used to filter what results we receive from Klaviyo. Some endpoints support "greater-or-equal" (preferred)
+    # and others only support "greater-than". comparison_operator must be specified for each incremental stream.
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2117, opened 2024-11-04
dbt: update text descriptions and allow removing dbt trigger in the UI

mdibaiee:
**Description:**

- Made all fields optional in the JSONSchema (although we still check for their existence)
- Updated title and description of interval

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/go/dbt/trigger.go
+++ b/go/dbt/trigger.go
@@ -12,14 +12,14 @@ import (
 )
 
 type JobConfig struct {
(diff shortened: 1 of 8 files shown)


williamhbaker (approved):
LGTM - I wonder if there's a more general issue with the JSON forms about resetting an optional configuration object even if there has been one of the required fields filled out before, but if that isn't something that can be resolved this seems like the only immediate way to make it possible to remove the dbt job configuration other than using the CLI.

PR #2118, opened 2024-11-04
source-klaviyo: raise errors for `Profiles` and `GlobalExclusions`

Alex-Bair:
**Description:**

The `raise_on_http_errors` property has been suppressing all errors for the `Profiles` and `GlobalExclusions` streams. It looks like this was originally done because Klaviyo sometimes returns a `503` when paginating per the Airbyte comments. I have not been able to replicate this. If that does happen, we should now get an error message, the shard should fail & restart, and the stream should pick up where it last left off paginating.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that HTTP errors (specifically a `400` response) for `Profiles` and `GlobalExclusions` fails the shard and logs out an error message.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2118)

### DIFF
--- a/source-klaviyo/source_klaviyo/streams.py
+++ b/source-klaviyo/source_klaviyo/streams.py
@@ -37,17 +37,6 @@ class KlaviyoStream(HttpStream, ABC):
         return KlaviyoAvailabilityStrategy()
 
 
-    @property
-    def raise_on_http_errors(self) -> bool:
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2123, opened 2024-11-04

willdonnelly:
**Description:**

A production task ran into trouble and needs its binlog cursor manually adjusted using the mechanism we built for doing that.

Ask Will or Dave for further context (or just read the recent Slack thread in #integrations)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2123)


dyaffe (approved):
LGTM

PR #2124, opened 2024-11-04
Revert "sqlcapture: Cursor replacement for production task"

willdonnelly:
Reverts estuary/connectors#2123

The reverted PR was a one-off adjustment to a specific task in production and should be reverted as soon as that adjustment has taken effect.


dyaffe (approved):
LGTM

PR #2127, opened 2024-11-05
source-sqlserver: Support DATETIME as backfill keys

willdonnelly:
**Description:**

By default we serialize these as an RFC3339 timestamp, but SQL Server doesn't support that as an input format for a DATETIME column since DATETIMEs only have three digits of sub-second precision and no time zone.

To make this work we just have to adjust the FDB row key encode operation so that when the column type is `datetime` we apply a different string format. This doesn't change the handling of the actual capture output values, only row key serialization.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2127)


williamhbaker (approved):
LGTM

PR #2128, opened 2024-11-06
materialize-webhook: support custom headers & chunk transactions across multiple webhooks

Alex-Bair:
**Description:**

Changes include:
- Supporting custom headers.
- Requiring collection keys in field selection.
- Chunking transactions across multiple webhooks.
- Webhook bodies are ~100 MiB or less.
- Removing retry mechanism in favor of allowing the connector to fail & retry the webhook when it is automatically restarted.

**Workflow steps:**

The HTTP Webhook [docs](https://docs.estuary.dev/reference/Connectors/materialization-connectors/http-webhook/) should be updated to reflect the configuration options.

**Notes for reviewers:**

- A single webhook only contains documents from a single binding.
- Webhooks are finished & closed once they reach/exceed the 100 MiB cutoff limit.
- Custom headers are specified in webhooks.
- Documents with different primary key values are not reduced by the runtime before being sent as Store requests.

The 1 MiB cutoff was arbitrarily chosen, and it's possible for the webhook to be larger than the cutoff if the last document piped into the webhook body pushes the byte count over the cutoff. This is because the rolling byte count is incremented after piping the document into the request. Taking this approach simplified the connector, but if we absolutely must not send webhooks larger than the cutoff, we can refactor the connector to do so. Hopefully this approach works for now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2128)

### DIFF
--- a/materialize-webhook/driver.go
+++ b/materialize-webhook/driver.go
@@ -126,6 +126,9 @@ func (driver) Validate(ctx context.Context, req *pm.Request_Validate) (*pm.Respo
 			case projection.IsRootDocumentProjection():
 				constraint.Type = pm.Response_Validated_Constraint_LOCATION_REQUIRED
 				constraint.Reason = "The root document must be materialized"
+			case projection.IsPrimaryKey:
+				constraint.Type = pm.Response_Validated_Constraint_LOCATION_REQUIRED
+				constraint.Reason = "Document keys must be included"
 			default:
 				constraint.Type = pm.Response_Validated_Constraint_FIELD_FORBIDDEN
 				constraint.Reason = "Webhooks only materialize the full document"
--- a/materialize-webhook/driver.go
+++ b/materialize-webhook/driver.go
@@ -21,8 +21,27 @@ import (
 // driver implements the pm.DriverServer interface.
 type driver struct{}
 
+type CustomHeader struct {
+	Name  string `json:"name,omitempty" jsonschema:"title=Name"`
+	Value string `json:"value,omitempty" jsonschema:"title=Value"`
+}
+
+func (h CustomHeader) Validate() error {
+	if h.Name == "" {
+		return fmt.Errorf("header name must not be empty")
+	} else if h.Value == "" {
+		return fmt.Errorf("value for header %v must not be empty", h.Name)
+	}
+	return nil
+}
+
+type headers struct {
+	CustomHeaders []CustomHeader `json:"customHeaders,omitempty" jsonschema:"title=Custom Headers"`
+}
+
 type config struct {
 	Address pf.Endpoint `json:"address" jsonschema:"title=Address,description=Base address URL. Must end in a trailing '/'."`
+	Headers headers     `json:"headers,omitempty"`
 }
 
 // Validate returns an error if the config is not well-formed.
@@ -31,6 +50,13 @@ func (c config) Validate() error {
 		return fmt.Errorf("address: %w", err)
 	} else if !strings.HasSuffix(string(c.Address), "/") {
 		return fmt.Errorf("address must end in a trailing '/'")
+	}
+
+	for i := range c.Headers.CustomHeaders {
+		header := c.Headers.CustomHeaders[i]
+		if err := header.Validate(); err != nil {
+			return fmt.Errorf("header: %w", err)
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


materialize-webhook/driver.go
```diff
@@ -161,87 +197,107 @@ func (d *transactor) Load(it *m.LoadIterator, _ func(int, json.RawMessage) error
...
+			if err := finishWebhook(); err != nil {
+				return nil, fmt.Errorf("finishWebhooks: %w", err)
+			}
+		}
+
+		previousAddress = address
+
+		if pipeWriter == nil {
+			startWebhook(address)
+			pipeWriter.Write([]byte("["))
```
williamhbaker, line 272: `Write` returns an error that should be handled

materialize-webhook/driver.go
```diff
@@ -161,87 +197,107 @@ func (d *transactor) Load(it *m.LoadIterator, _ func(int, json.RawMessage) error
...
+		pipeWriter = w
+	}
+
+	finishWebhook := func() error {
+		if pipeWriter == nil {
+			return nil
+		}
+
+		pipeWriter.Write([]byte("]"))
+		pipeWriter.Close()
```
williamhbaker, line 244: `Close` returns an error that should be handled.

materialize-webhook/driver.go
```diff
@@ -161,87 +197,107 @@ func (d *transactor) Load(it *m.LoadIterator, _ func(int, json.RawMessage) error
...
-			}).Error("failed to invoke Webhook (will retry)")
+			byteCount = 0
 		}
 	}
+
+	if err := finishWebhook(); err != nil {
+		return nil, fmt.Errorf("finishWebhook: %w", err)
+	}
+
+	if err := it.Err(); err != nil {
```
williamhbaker, line 293: Probably would be good to move this error check to right after the iterator loop, which is a more typical pattern.

materialize-webhook/driver.go
```diff
@@ -161,87 +197,107 @@ func (d *transactor) Load(it *m.LoadIterator, _ func(int, json.RawMessage) error
...
+				request.Header.Add(header.Name, header.Value)
+			}
+
 			response, err := http.DefaultClient.Do(request)
+
 			if err == nil {
 				err = response.Body.Close()
 			}
+
 			if err == nil && (response.StatusCode < 200 || response.StatusCode >= 300) {
```
williamhbaker, line 227: I think there needs to be a check for `if err != nil` somewhere in here. It's also important to consider what happens if this goroutine errors before we call `wait` on the errgroup - writes to the writer end of the pipe will block forever since there is nothing left reading from the reader end.

You should add some handling to close the read end if the reading process errors, so that the error will be propogated to `write` calls. See https://github.com/estuary/connectors/blob/a492d4573331be75a0ad55e7ea3b8ac1fbec4611/materialize-redshift/staged_file.go#L111-L124 for an example of this. You'll probably ened to do it if `http.NewRequest` fails, or if `http.DefaultClient.Do` fails.

materialize-webhook/driver.go
```diff
@@ -161,87 +197,107 @@ func (d *transactor) Load(it *m.LoadIterator, _ func(int, json.RawMessage) error
...
 
-			if err == nil {
-				body.Reset() // Reset for next use.
-				break
-			} else if attempt == 10 {
-				return nil, fmt.Errorf("webhook failed after many attempts: %w", err)
+			return nil
+		})
+
+		pipeWriter = w
```
williamhbaker, line 235: Nit: Move this assigned up higher, under the `pipeReader, w := io.Pipe()` line.

materialize-webhook/driver.go
```diff
@@ -1,28 +1,52 @@
...
 	pf "github.com/estuary/flow/go/protocols/flow"
 	pm "github.com/estuary/flow/go/protocols/materialize"
-	log "github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 )
 
 // driver implements the pm.DriverServer interface.
 type driver struct{}
 
+type StreamEncoder interface {
```
williamhbaker, line 23: I don't think this interface is used anywhere.

materialize-webhook/driver.go
```diff
@@ -1,28 +1,52 @@
...
+	Encode(row []any) error
+	Written() int
+	Close() error
+}
+
+type CustomHeader struct {
```
williamhbaker, line 29: This struct doesn't need to be exported.

materialize-webhook/driver.go
Alex-Bair: Got it. I used `CloseWithError` in both of those cases plus if `response.Body.Close` fails or if we receive a non-2xx response, so those errors can also be propagated to the `write` calls.

williamhbaker (approved):

PR #2131, opened 2024-11-07
source-{redshift,postgres}-batch: Support NaNs

willdonnelly:
**Description:**

Adds support in the batch Postgres and Redshift captures for the special floating-point values `'NaN'`, `'Infinity'`, and `'-Infinity'`, which are translated into the corresponding strings as we do for other captures.

Since Postgres batch relies on schema inference for non-key property types and Redshift batch already has logic to discover floating-point columns as `type: [number, string], format: number` for other reasons, this doesn't require any discovery changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2131)


williamhbaker (approved):
LGTM

PR #2133, opened 2024-11-08
source-postgres: exclude temp tables from discovery

psFried:
Filters out temporary and unlogged tables during discovery. These tables cause validation to fail because they cannot be added to publications. The filtering is done based on the `pg_class.relpersistence` values, described in the docs: https://www.postgresql.org/docs/current/catalog-pg-class.html#CATALOG-PG-CLASS

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2133)


willdonnelly (approved):
LGTM

Issue #2076, opened 2024-10-23
materializations: don't persist a spec


williamhbaker:
`Apply` and `Validate` RPCs now include the last applied / last validated spec, so we don't have to persist it in the endpoint. This would be a nice simplification for materialization connectors, and may alleviate & prevent future problems with spec persistence (being too big, etc.). So we should stop storing specs in metadata tables. Something to double check is how disabling and re-enabling bindings would work with the new way of getting specs: Currently the materialization completely forgets about the old binding with the persisted spec. How does that work with the ones that are present in the protocol messages?

PR #2139, opened 2024-11-12
materialize-dynamodb: don't require a persisted spec

williamhbaker:
**Description:**

DynamoDB has a limit of 400KB for any single "item" stored, which is problematic for persisting a materialization spec.

We now have the prior materialization spec for validate & apply as part of the protocol messages, so we can use those instead.

This implementation is a short-term hack for unblocking a current task which is stuck trying to persist a spec that is too large, and we should come back and fix it up when we incorporate the prior spec from the validate & apply protocol messages more generally, see issue #2076.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2139)


jgraettinger (approved):
LGTM

PR #2141, opened 2024-11-13

willdonnelly:
**Description:**

A production task ran into trouble and needs its binlog cursor manually adjusted using the mechanism we built for doing that.

Ask Will for further context

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2141)


williamhbaker (approved):
LGTM

PR #2142, opened 2024-11-13
Revert "sqlcapture: Cursor replacement for production task"

willdonnelly:
Reverts estuary/connectors#2141

The reverted PR was a one-off adjustment to a specific task in production and should be reverted as soon as that adjustment has taken effect.


williamhbaker (approved):

willdonnelly:
Cursor replacement has taken effect, cleaning up after myself now

PR #2145, opened 2024-11-14
source-snowflake: Support (incremental) dynamic tables

willdonnelly:
**Description:**

As a general rule, if a connector returns something from discovery then we should be able to capture that thing without errors, and if we can't capture a particular type of entity then it shouldn't be suggested by discovery in the first place. A user should be able to just blindly click 'Next' in the capture creation flow and end up with something that works.

This was not previously the case for dynamic tables in Snowflake. They're tantalizingly close to working without any extra action, but for reasons known only to them Snowflake is very nitpicky about making you acknowledge that a table is dynamic. For instance you can't just say `CREATE STREAM ... ON TABLE`, you have to tell them `CREATE STREAM ... ON DYNAMIC TABLE`.

So dynamic tables didn't work, but our discovery logic returned them, which meant that we're violating that "just clicking Next ought to work" principle I described above. This had to be fixed in one of the two obvious ways, and I figure that once we've got logic to query whether a table is dynamic we might as well just plug that into the appropriate bits of the capture so it can use the necessary dynamic-table variants of the operations.

There turns out to be an additional wrinkle however. Snowflake's documentation won't say this explicitly anywhere, but basically you can create change streams from _incremental_ dynamic tables but you can't create a stream on a _full refresh_ dynamic table. So it turns out we still have to have some discovery filtering logic in order to exclude dynamic tables with `REFRESH_MODE = FULL`.

With these changes, the Snowflake CDC connector should support dynamic tables with `REFRESH_MODE = INCREMENTAL` and also satisfy that "stuff discovery suggests is stuff we can capture" principle described earlier, at least for all of the types of Snowflake table I know about.

**Workflow steps:**

Discovery will no longer return dynamic tables with `REFRESH_MODE = FULL` in the first place. Dynamic tables with `REFRESH_MODE = INCREMENTAL` will still be discovered, and now they will actually work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2145)

### DIFF
--- a/source-snowflake/.snapshots/TestBasicDatatypes-main
+++ b/source-snowflake/.snapshots/TestBasicDatatypes-main
@@ -2,7 +2,7 @@
 # Collection "acmeCo/test/test_basicdatatypes_77528227": 3 Documents
 # ================================
 {"BDATA":"7loRpQ==","BIT":false,"DATA":"DCBA","FDATA":2.64,"ID":3,"X":0,"Y":0,"_meta":{"op":"c","source":{"schema":"PUBLIC","table":"test_BasicDatatypes_77528227","seq":2,"off":0}}}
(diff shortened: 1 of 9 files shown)


williamhbaker (approved):
LGTM

PR #2150, opened 2024-11-14
source-front: rewrite as native connector

Alex-Bair:
**Description:**

The `source-front` connector was not working correctly as an import. Instead of troubleshooting the low-code Airbyte import, I chose to rewrite it as a native connector with only a subset of the streams. I focused on parent streams that seemed most important for the user's use case.

Backwards compatibility with the single existing, broken capture was not a goal.

The version number has been incremented since the connector is changing from an Airbyte import to a native implementation. Last time I rewrote an import as a native connector, I kept the version number the same and ran into issues with the resource path pointers after deployment. Incrementing the version number should avoid the same issue this time.

**Workflow steps:**

Documentation should be created for this connector. I'll link the docs PR here once it's ready.

**Notes for reviewers:**

- all streams capture data & complete.
- incremental streams checkpoint & log every so often to prove they're making progress.

Some streams (primarily `conversations` and `events` can take a while to complete depending on how much data is in the user's Front account. 

Snapshot changes are expected due to this being a full connector rewrite.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2150)


source-front/source_front/__init__.py
```diff
@@ -1,8 +1,63 @@
...
+    BaseCaptureConnector[EndpointConfig, ResourceConfig, ConnectorState],
+    HTTPMixin,
+):
+    def request_class(self):
+        return Request[EndpointConfig, ResourceConfig, ConnectorState]
+
+    async def spec(self, _: request.Spec, logger: Logger) -> ConnectorSpec:
+        return ConnectorSpec(
+            configSchema=EndpointConfig.model_json_schema(),
+            documentationUrl="https://go.estuary.dev/source-pendo",
```
williamhbaker, line 35: Documentation URL needs updated

source-front/source_front/api.py
```diff
@@ -0,0 +1,203 @@
...
+
+        for resource in resources:
+            yield resource
+
+        next_page_token = _extract_page_token(response.pagination.next)
+
+        if next_page_token is None:
+            break
+
+        params.update({
```
williamhbaker, line 71: Any reason for this form vs. a 1-liner like `params["page_token"] = next_page_token`?

source-front/source_front/api.py
```diff
@@ -0,0 +1,203 @@
...
+        })
+
+
+async def _fetch_conversations_between(
+        http: HTTPSession,
+        log: Logger,
+        start: datetime,
+        end: datetime,
+) -> AsyncGenerator[FrontResource, None]:
+    """
```
williamhbaker, line 82: Mega-nit here: I actually would remove this doc comment, since the function name/signature are self-explanatory. Comments have tendency to drift from their associated code as updates are made to the code and corresponding updates are not made to the comments. This isn't one where I think that's likely to happen, but as a general practice I like to keep comments only for annotating non-obvious things; a good example is to document why the code _isn't_ written in a certain way, rather than just re-stating how it is written.

source-front/source_front/api.py
```diff
@@ -0,0 +1,203 @@
...
+
+        for record in records:
+            yield record
+
+        next_page_token = _extract_page_token(response.pagination.next)
+
+        if next_page_token is None:
+            return
+
+        params.update({
```
williamhbaker, line 107: Same as above, could be a one-liner?

williamhbaker (approved):
LGTM % a few minor comments

source-front/source_front/__init__.py
Alex-Bair: :facepalm: Thanks for catching the copy-paste remnants I missed (again)!

source-front/source_front/api.py
Alex-Bair: Nope, no particular reason. I'll update these to be 1-liners.

source-front/source_front/api.py
Alex-Bair: Makes sense, I'll remove it. I'm also thinking that the doc comment I put in for `fetch_conversations` falls into the same category, so I'll remove that one as well.

PR #2153, opened 2024-11-15
materialize: parquet encoder handles integers encoded as floats

williamhbaker:
**Description:**

Integer values can be received as floats with 0 fractional part, even if the value is within the range of a 64-bit integer. This updates the parquet encoder to handle that case.

Integer values can also be received as floats if the value is outside the range of 64-bit integer, and that's not being addressed right now since it doesn't seem like an immediate issue.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2153)

### DIFF
--- a/materialize-boilerplate/stream-encode/parquet.go
+++ b/materialize-boilerplate/stream-encode/parquet.go
@@ -6,6 +6,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #2154, opened 2024-11-18
source-kafka: extend metadata timeout

williamhbaker:
**Description:**

This timeout is used for fetching cluster metadata during the interactive discover and validation operations.

We need some kind of timeout here since its not uncommon for misconfigured configurations to end up hanging forever when trying to connect, but it can also take a little while for all the metadata to be returned from a Kafka cluster if there is a lot of it.

This extends the timeout from a pretty restrictive 5 seconds to a much more permissive 60 seconds. 60 seconds is quite a while for a discover/validate attempt to hang if the configuration is incorrect but that seems like a better direction to err on than failing correct configurations too soon for particularly large clusters.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2154)

### DIFF
--- a/source-kafka/src/discover.rs
+++ b/source-kafka/src/discover.rs
@@ -18,7 +18,7 @@ use crate::{
         RegisteredSchema::{Avro, Json, Protobuf},
         SchemaRegistryClient, TopicSchema,
     },
-    KAFKA_TIMEOUT,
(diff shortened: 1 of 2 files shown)


willdonnelly (approved):
LGTM

PR #2158, opened 2024-11-20
source-mysql: Two bits of production wrangling

willdonnelly:
**Description:**

This PR does two things, which I've combined into one PR because they're both tiny and this was more convenient:

1. Adds another temporary cursor replacement rule to get a production task rewound to an earlier portion of the binlog.
2. Adds a special-case to the "Skip Backfills" advanced config option so that the value `*.*` means "all tables". This doesn't generalize to like `schema.*` or anything right now, it's literally just matching that one string, but we needed a way to stop all tables in a capture from backfilling simultaneously with the cursor reset, and this seemed like the least bad option that could be done quickly.

**Workflow steps:**

The one capture that this is all targeted at is currently disabled. After these changes are live in production, that capture should be modified to have `*.*` in the advanced "Skip Backfills" option and then re-enabled.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2158)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -399,6 +399,18 @@ func decodeKeyFDB(t tuple.TupleElement) (interface{}, error) {
 }
 
 func (db *mysqlDatabase) ShouldBackfill(streamID string) bool {
+	// As a special case, the solitary value '*.*' means that nothing should be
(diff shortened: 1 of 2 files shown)


dyaffe (approved):
LGTM

PR #2159, opened 2024-11-20
source-kinesis: usability updates

williamhbaker:
Reworks some of the source-kinesis code to improve usability. These updates are backward-compatible with existing captures.

Changes of note:
- Upgrade to the v2 AWS Go SDK.
- Implement shard pruning from the state checkpoint on startup.
- Simplify the shard iterator reading retry mechanism.
- Pass all errors from async processes back into the main thread so that fatal errors are surfaced.
- Remove the Flow capture shard splitting handling. There was a flaw in this logic related to Kinesis shard merges: When multiple Kinesis shards are merged, the resulting shard must not be read until all of the parent shards are read to ensure records are in the correct order. With the previous logic it is possible that one of the parent shards of a merged shard is completely outside the range of the Flow capture shard, and so it would have no way to know when that parent was completely read. Implementing capture scale-out may require using a DynamoDB table for coordination similar to other Kinesis reading libraries, and/or some TBD Flow task shard coordination mechanism.

**Description:**

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM based on what I could glean from various resources about how Kinesis works.

PR #2161, opened 2024-11-21

willdonnelly:
**Description:**

A production task ran into trouble and needs its binlog cursor manually adjusted. Ask Will or Dave for further context.

This is the same task we had to poke yesterday and I hadn't gotten around to deleting the previous replacement, so just the specific cursor values need to change this time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2161)


dyaffe (approved):
LGTM

PR #2162, opened 2024-11-22
Bair/facebook marketing account ids

### LINKED ISSUE #2110: source-facebook-marketing: Allow for multiple customer IDs
https://github.com/airbytehq/airbyte/pull/33538/files

Alex-Bair:
**Description:**

The main change in this PR is supporting multiple Facebook Marketing account IDs in a single connector. Multiple account IDs can be provided as comma separated values in the `account_ids` field. This change is backwards compatible with older configs that only have an `account_id` field.

The account ID is used in multiple locations throughout the connector, and many of the changes are for threading the account ID through all streams.

Supporting more than one account ID in the connector also means that streams need to retain a cursor per account ID. The various state management tools (`get_updated_state`, `stream_slices`, the `state` property, etc.) have been updated to retain a cursor value for each account ID.

Closes #2110 

**Workflow steps:**

The `source-facebook-marketing` documentation should be updated to reflect that multiple account IDs are supported. The docs update PR will be linked here once it's ready.

**Notes for reviewers:**

Tested on a local stack with a single account ID and a recent start date (01NOV2024). Confirmed that:
- The same number of records are retrieved for all standard streams before and after these changes.
- Existing tasks switch over to use `account_ids` without error.
- Existing tasks' incremental streams pick up where they left off (i.e. cursor values are preserved through the switchover).

Spec and discover snapshot changes are expected due to adding support for multiple account IDs & updating the documentation URLs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2162)


source-facebook-marketing/source_facebook_marketing/schemas/videos.json
```diff
@@ -1,6 +1,9 @@
 {
   "type": "object",
   "properties": {
+    "account_id": {
+        "type": ["null", "string"]
```
williamhbaker, line 5: For these schemas, can the `account_id` ever really be an explicit `null`? I'd think it would either be present or absent, but always a string if it is present.

williamhbaker (approved):
LGTM

source-facebook-marketing/source_facebook_marketing/schemas/videos.json
Alex-Bair: You're correct. From what I've seen, `account_id` is either present as a string or absent, never an explicit `null`. I've been using `"type": ["null", "someOtherType"]` for fields that can be absent, regardless of whether they can be explicit `null`s since that's consistent with the other Airbyte imports. 

Would it be better to use `anyOf` like below instead? I _think_ this means the field is either present as a string or absent.
```python
"properties": {
    "account_id": {
        "anyOf": [
            { "type": "string"},
            { "type": "null"},
        ]
    }
}
```

williamhbaker: Present or not is controlled by the field being in the list of required fields. So a field that may either be present as a string or completely absent is `type: string` but is not in the list of required fields for the object.

Alex-Bair: Got it, I see now. So `account_id` in these schemas can just be `"type": "string"`. I'll update those.

PR #2164, opened 2024-11-22
source-hubspot-native: bindings for owners, deal pipelines, products, and line items

williamhbaker:
**Description:**

Adds new bindings for owners, deal pipelines, products, and line items.

As noted in the commit messages, for products and line items we don't have data in our test account and I'm not going to add any, so the test snapshots are unchanged for those. I did manually test that the results are as expected on a a HubSpot account that has data for those bindings though.

**Workflow steps:**

HubSpot native docs will need updated.

**Notes for reviewers:**


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -65,6 +70,22 @@ async def fetch_properties(
     return properties_cache[object_name]
 
 
+async def fetch_deal_pipelines(
+    log: Logger, http: HTTPSession
+) -> DealPipelines:
+    url = f"{HUB}/crm-pipelines/v1/pipelines/deals"
```
Alex-Bair, line 76: For my own understanding, was the motivation to use [`/crm-pipelines/v1/pipelines/deals`](https://developers.hubspot.com/beta-docs/reference/api/crm/pipelines/v1) instead of [`/crm/v3/pipelines/deals`](https://developers.hubspot.com/beta-docs/reference/api/crm/pipelines#get-%2Fcrm%2Fv3%2Fpipelines%2F%7Bobjecttype%7D) because the required permissions for the former are simpler?

source-hubspot-native/source_hubspot_native/models.py
```diff
@@ -112,6 +119,22 @@ class Properties(BaseDocument, extra="forbid"):
     results: list[Property]
 
 
+class DealPipeline(BaseDocument, extra="allow"):
+
+    createdAt: AwareDatetime | None
+    updatedAt: AwareDatetime | None
```
Alex-Bair, line 125: Will these fields only ever be `None` for tombstones?

Alex-Bair (approved):
LGTM - I just had a couple questions to solidify my understanding.

source-hubspot-native/source_hubspot_native/api.py
williamhbaker: I didn't really think about it too much. I think they are basically equivalent. The v1 APIs tend to have more lenient rate limits.

source-hubspot-native/source_hubspot_native/models.py
williamhbaker: Yep, assuming the API doesn't do anything strange. But they should always be present in actual API objects.

PR #2166, opened 2024-11-22
source-postgres: Support `_citext` columns

willdonnelly:
**Description:**

Arrays of `citext` are tricky because `citext` is an extension type which means that it has no stable OID, and as a result the PGX client library doesn't have any default type mappings for it. In the absence of any registered codec, values are always captured in their stringified text-protocol format.

However our discovery logic still notices that it's an array and emits the appropriate schema for an array of text values, which means there's a mismatch and we're not satisfying our own schema. Which is bad. While `citext` is an extension type, it's a common one and the citext extension ships as part of the Postgres release, so I believe we should still support it as best we can.

This commit fixes that by extending the "datatype tweaks" logic with a query against `pg_catalog.pg_type` looking for specific extension types we support (which currently is just `citext`) and for each result row we register custom handlers for the base type and arrays of that type.

Since this is all getting fairly complex, I've also gone ahead and added test cases for `citext` and `_citext` datatype support. But of course those don't work against a database without that extension installed, so the CI database setup script has been modified to install it. (The `citext` extension is shipped as part of Postgres itself, it just isn't enabled by default, so we just have to run `CREATE EXTENSION citext`).

**Workflow steps:**

Columns of type `citext[]` should just work going forward.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2166)

### DIFF
--- a/source-postgres/.snapshots/TestCIText-Capture
+++ b/source-postgres/.snapshots/TestCIText-Capture
@@ -1,0 +1,14 @@
+# ================================
+# Collection "acmeCo/test/test_citext_58810479": 6 Documents
+# ================================
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"citext_58810479","loc":[11111111,11111111,11111111]}},"arr":{"dimensions":[2],"elements":["a","b"]},"data":"zero","id":0}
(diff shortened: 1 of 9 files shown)


williamhbaker (approved):
LGTM

PR #2171, opened 2024-11-26
source-postgres: increase connect timeout to 20s

psFried:
We're seeing frequent connection failures for some captures, so kinda just hoping that this might help with those.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2171)


willdonnelly (approved):
LGTM

PR #2172, opened 2024-11-27
Let docker handle the wait and remove docker compose version

surister:
**Description:**

Title

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2172)

### DIFF
--- a/materialize-postgres/docker-compose.yaml
+++ b/materialize-postgres/docker-compose.yaml
@@ -1,5 +1,3 @@
-version: "3.7"
-
 services:
   postgres:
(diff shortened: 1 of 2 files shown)


williamhbaker:
There is no health check in the `materialize-postgres` docker compose file, so I am not sure if the `--wait` flag will consistently have the desired effect here.

Although the 5 second sleep is really hacky, it has worked well enough for a long time that I don't think further investment toward improving this is warranted right now. I'm going to close this PR out, but feel free to open an issue for further discussion if you think there is something we should be doing different!

PR #2173, opened 2024-11-27
source-http-ingest: add CORS support

psFried:
Supports CORS preflight requests on an opt-in basis. The goal is to allow the users to capture data directly from browsers. A list of allowed origins was added to the endpoint config. If left empty (the default), then cors will be disabled entirely. A non-empty list will enable cors for the specific origins listed. As a special case, `*` may be used to permit any origin.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2173)


williamhbaker (approved):
LGTM

PR #2177, opened 2024-12-03
source-braintree-native: new connector

Alex-Bair:
**Description:**

This is a minimal connector for Braintree. All incremental streams only capture incremental creates (not updates) due to limitations with Braintree's APIs, so regular backfills are required to capture updates. I did not include logic to trigger backfills within the connector since we already have a separate mechanism for scheduled backfills.

Braintree provides SDKs and a GraphQL API for retrieving data. Instead of using the GraphQL API and figuring out the quirks of using it for incremental replication, I decided to use the Python SDK to speed up development. This circumvents error handling baked into the Estuary CDK's `HTTPMixin` object, so we may need to improve error handling later.

Braintree's SDK returns custom objects instead of dictionaries, and the `braintree_object_to_dict` function converts these custom objects to dictionaries to feed the Pydantic models. This function may need additional handling/improvement once the connector is used in production and we better understand what types users want.

The `credit_card_verifications` stream is not included in the capture snapshot because I couldn't figure out how to create these in a sandbox Braintree account.

**Workflow steps:**

Documentation will need created for this connector. It should probably clarify that incremental streams only capture creates, not updates, and regular backfills are needed to capture updates. The docs PR will be linked here once it's ready.

**Notes for reviewers:**

- Streams capture data & associated collections can be materialized into PostgreSQL.
- Snapshots test pass.
- Invalid credentials throw an error before task creation.
- Objects returned by Braintree's SDK are converted into Pydantic models appropriately.

I could not figure out how to create a `credit_card_verification` object within Braintree, so I could not confirm that stream captures data.

Supabase, Strapi, and Short.io will need updated to list this new connector in our catalog.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2177)


source-braintree-native/source_braintree_native/api.py
```diff
@@ -0,0 +1,253 @@
...
+TRANSACTION_SEARCH_LIMIT = 50_000
+# Incremental streams use sliding date windows to try & avoid hitting the above search limits.
+WINDOW_SIZE = 15
+
+CONVENIENCE_OBJECTS = [
+    'gateway'
+]
+
+
+def _braintree_object_to_dict(braintree_object):
```
willdonnelly, line 39: Just a sanity check: top-level entities in Braintree are always objects?

source-braintree-native/source_braintree_native/api.py
```diff
@@ -0,0 +1,253 @@
...
+    for object in collection.items:
+        count += 1
+        doc = Transaction.model_validate(_braintree_object_to_dict(object))
+
+        if doc.created_at > log_cursor:
+            yield doc
+            most_recent_created_at = doc.created_at
+
+    if count >= TRANSACTION_SEARCH_LIMIT:
+        log.warning(f"{count} transactions returned in a single search which is greater than or equal to Braintree's documented maximum for a single transaction search.")
```
willdonnelly, line 113: If this happens, what's the user's recourse? If I'm reading this correctly the search is restricted to a specific time window and the solution is to make that window smaller, but it also looks like currently the window size is hard-coded to 15 up above. Should the window size be a configurable setting instead?

willdonnelly, line 113: Sanity checking my own understanding: Is the idea that this is non-fatal because there will be another regular backfill at some point in the future which will get any missing data?

Is there some subtle reason that a fresh backfill of the same resource won't hit this error?

source-braintree-native/source_braintree_native/resources.py
```diff
@@ -0,0 +1,312 @@
...
+            initial_config=ResourceConfig(
+                name=name, interval=timedelta(minutes=5)
+            ),
+            schema_inference=True,
+        )
+        for (name, gateway_property, gateway_response_field) in FULL_REFRESH_RESOURCES
+    ]
+
+
+def transactions(
```
willdonnelly, line 103: Are all of these functions identical except for the `fetch_foobar` function and the resource name? Because it feels like this code could be deduplicated a lot if there's no reason to expect them to diverge in the near future.

willdonnelly (commented):
Looking good overall! I haven't dug very deeply into this, but I gave it a quick read-through and had a few questions, mostly around how search result limits are handled.

source-braintree-native/source_braintree_native/api.py
Alex-Bair: Yep, their SDK returns an object, like the [Transaction](https://developer.paypal.com/braintree/docs/reference/response/transaction/python) object.

source-braintree-native/source_braintree_native/api.py
Alex-Bair: That's correct, if we hit the search limit, the window size should be smaller. I initially didn't include it in the config to make set up easier & more streamlined, but I'm seeing now that it'd be nicer to have it as a configurable setting. I'll change that & add it to the config under the advanced section.

Alex-Bair: No you're right. A backfill will hit the same error. I originally included this log to trigger us to reduce the window size, but it makes more sense to make it a fatal error & let the user configure a different window size when it happens. I'll update this.

source-braintree-native/source_braintree_native/resources.py
Alex-Bair: Yep, you're right. They're all the same except for the `fetch_foobar` functions and resource names. I'll refactor.

source-braintree-native/source_braintree_native/api.py
```diff
@@ -0,0 +1,261 @@
...
+    for object in collection.items:
+        count += 1
+        doc = IncrementalResource.model_validate(_braintree_object_to_dict(object))
+
+        if doc.created_at > log_cursor:
+            yield doc
+            most_recent_created_at = doc.created_at
+
+    if count >= SEARCH_LIMIT:
+        msg = f"{count} customers returned in a single search which is greater than or equal to Braintree's documented maximum for a single customer search. Reduce the window size and backfill this stream."
```
willdonnelly, line 150: Is there a reason this error message is still here instead of using `_search_limit_error_message` like the other generator functions?

willdonnelly (approved):
LGTM % one nit (which I think was just an oversight when refactoring a bunch of nearly-identical functions?)

source-braintree-native/source_braintree_native/api.py
Alex-Bair: Nope, I overlooked it when I was refactoring. Thanks for catching that! I'll update it before merging.

PR #2180, opened 2024-12-04
source-sqlserver: filegroup and role_name settings

willdonnelly:
**Description:**

This commit adds new advanced options for the CDC instance filegroup and role_name arguments. If set, these values will be plugged into the appropriate query parameters whenever the connector tries to autocreate CDC instances. (If unset, they will be omitted so the behavior will be the same as it was prior to this change.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2180)


Alex-Bair (approved):
LGTM!

PR #2182, opened 2024-12-05
source-twilio: improve MessageMedia incremental sync speed

Alex-Bair:
**Description:**

`MessageMedia` was previously checking every message between the config's start date and the present for new media, then filtering out any media created before the last seen cursor value. That made incremental syncs take an extremely long time without any apparent progress; the stream could be searching through the past few years of messages when it usually only needs to search through the past few minutes.

This change makes the `MessageMedia` stream only check messages created since the most recent cursor value, falling back to the config's start date if no cursor value is present. This significantly speeds up the connector during incremental syncs.

This change also increases the date window size used when fetching a message's media from 1 year to 100 years. This reduces the number of API requests needed when backfilling media records over a year old; instead of requesting a single year of media at a time, the connector essentially requests all of a message's media in one request. For example, instead of making two requests spanning NOV2023-NOV2024 and NOV2024-DEC2024, a single request is made for NOV2023 - DEC2024.

It would make more sense to not use a sliding date window strategy for fetching a single message's media, but rewriting the `MessageMedia` stream in a backwards compatible way is a large effort I'd like to avoid, especially when small, targeted changes address the current issue.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that for the `MessageMedia` stream:
- backfills use the config's start date for searching the parent stream `Messages`.
- restarts after a partial/complete backfill use the last seen cursor value for searching the parent stream `Messages`.
- retrieving media from messages over a year old is done in a single request.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2182)

### DIFF
--- a/source-twilio/source_twilio/streams.py
+++ b/source-twilio/source_twilio/streams.py
@@ -557,17 +557,23 @@ class Messages(IncrementalTwilioStream, TwilioNestedStream):
 class MessageMedia(IncrementalTwilioStream, TwilioNestedStream):
     """https://www.twilio.com/docs/sms/api/media-resource#read-multiple-media-resources"""
 
-    parent_stream = Messages
+    parent_stream: type[Messages] = Messages
     data_field = "media_list"
     subresource_uri_key = "media"
     media_exist_validation = {"num_media": "0"}
     lower_boundary_filter_field = "DateCreated>"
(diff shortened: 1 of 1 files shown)


jgraettinger (approved):
LGTM!

PR #2184, opened 2024-12-05
source-asana: don't cache `Tasks` records

Alex-Bair:
**Description:**

A production task that only has the `Tasks` binding enabled is OOM-ing because every `Tasks` record is being cached. I suspect this was done because `Tasks` is used as a parent stream for some other streams. Not caching `Tasks` records will likely slow down those other streams, but currently no one is using those streams & we'd have to address these OOMs due to caching regardless.

I added some logging to give some indication that the connector is doing something while the connector works on the full refresh `Tasks` stream.

The capture snapshot test has been failing for a while, so I've also updated that test to be more resilient. 

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Before this change, the connector's memory usage would steadily increase. I noticed it get up to ~65% after a few minutes, and I assume that if I let it run the connector would get OOM killed. After the change, the connector's memory usage has been steady around 12% for the past 75 minutes.

Capture snapshot changes are expected since I trimmed down the snapshot to only contain a single record per stream, plus I suspect the data within our test Asana account has changed since the snapshot was last updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2184)


jgraettinger (approved):
LGTM

PR #2185, opened 2024-12-05
source-mysql: Treat CDC updates which change the primary key as implicit deletions

willdonnelly:
**Description:**

This PR adds tests to `source-{postgres,mysql,sqlserver}` exercising the capture behavior when a row is updated in a way which modifies its primary key. There are two ways this could be handled:

1. Just transcribe the update literally, with the new row as the collection key and the old primary-key value in the before-state metadata.
2. Generate a synthetic pair of a delete event for the old primary-key value, and an insert event for the new value.

Option (2) is more correct because it results in a deletion of the old row being propagated to any standard / merge update materializations, and also maintains the "the first mention of any previously-nonexisting key should be an insert event" invariant that is sometimes useful. Especially when combined with hard deletions, this means that the destination table will continue to precisely match the source, whereas option (1) would leave behind a bunch of vestigial rows corresponding to old primary-key values which were superseded by later updates.

Previously, SQL Server implemented option (2) while MySQL and Postgres implemented option (1). This was the path of least resistance for each of these captures. The actual meat of this PR is that now `source-mysql` implements option (2) as well by explicitly checking whether the row key is changed between the before and after states of an update event, and emitting the aforementioned synthetic pair of delete/inserts if so.

Postgres is not updated in this PR, because while the actual detection logic would be basically identical there's some additional change event plumbing which assumes that each CDC event from the database will yield at most one change event internally, and fixing that is slightly more involved so I'm punting on it for the moment in order to get MySQL fixed sooner. We should come back and address that soon so all the SQL CDC connectors are consistent.

**Workflow steps:**

Going forward, updates to MySQL tables which modify the primary key of a row will be captured as implicitly deleting the old row-state and inserting the new row-state.

Any preexisting stale rows from primary-key updates which occurred before this change went live will not be cleaned up, remedying that will require manual work or a complete backfill of the impacted dataflows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2185)

### DIFF
--- a/source-mysql/.snapshots/TestPrimaryKeyUpdate
+++ b/source-mysql/.snapshots/TestPrimaryKeyUpdate
@@ -1,0 +1,16 @@
+# ================================
+# Collection "acmeCo/test/test_primarykeyupdate_63510878": 8 Documents
+# ================================
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"PrimaryKeyUpdate_63510878","cursor":"backfill:0"}},"data":"zero","id":0}
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"PrimaryKeyUpdate_63510878","cursor":"backfill:1"}},"data":"one","id":1}
+{"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"PrimaryKeyUpdate_63510878","cursor":"backfill:2"}},"data":"two","id":2}
+{"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"PrimaryKeyUpdate_63510878","cursor":"binlog.000123:56789:123","txid":"11111111-1111-1111-1111-111111111111:111"}},"data":"three","id":3}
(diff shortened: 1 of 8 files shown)


Alex-Bair (approved):
LGTM!

sqlcapture/capture.go
```diff
@@ -648,27 +648,30 @@ func (c *Capture) backfillStream(ctx context.Context, streamID string, discovery
...
+		Before    map[string]any `json:"before,omitempty"`
 	}{
 		Operation: event.Operation,
 		Source:    event.Source,
 		Before:    nil,
 	}
 	switch event.Operation {
 	case InsertOp:
-		record = event.After // Before is never used.
+		for k, v := range event.After {
```
jgraettinger, line 663: what motivated the extra allocation / copy here?

willdonnelly: Previously I had the key-change logic emitting a deletion event using the 'before' row state, and then an update with the 'before' state in the metadata, and the in-place modification of this map resulted in an extra `_meta` field getting embedded into that before-state metadata.

After changing the logic to omit any before-state metadata and just treat the creation as a pure insert-from-nothing I think it would be safe to revert this fix if we think the additional allocation is a significant bottleneck.

willdonnelly: I've reverted the sqlcapture memory-reuse change since it's no longer necessary for correctness after all, so now the changes in this PR (other than the new tests for Postgres and SQL Server) are isolated to just MySQL.

PR #2186, opened 2024-12-05
source-impact-native: fix `Actions` backfill cursor and other misc. fixes

Alex-Bair:
**Description:**

- Updating snapshot file names.
- Increasing the interval between sweeps to 5 minutes. The connector frequently hits Impact's 1000 requests/hour rate limit, and increasing the interval from 0s to 5m should help with that.
- Updated documentation URL to be connector specific.
- Renaming `stop_date` to `start_date` in the config.
- Changing the cursor used when backfilling `Actions`.
  - Previously, the "LockingDate" field was used during `Actions` backfills. All actions have a "LockingDate" that's a fixed day the month after the action is created. For example, all actions created in October have a "LockingDate" of November 28th. This effectively batches actions into month-sized chunks, and using "LockingDate" as a cursor field forced us to only get completed month-sized chunks. This caused the connector to miss actions whose "LockingDate" was after the cutoff date (i.e. within in-progress month-sized chunks) during backfills. 
  - Instead, `Actions` now uses the `EventDate` field during backfills (via the "ActionDateStart" and "ActionDateEnd" [query params](https://integrations.impact.com/impact-publisher/reference/list-actions-1)). This field has much finer granularity than "LockingDate", and backfills now capture the data that was previously missed.

**Workflow steps:**

Documentation does not exist for the `source-impact-native` connector, so it should be created.

**Notes for reviewers:**

Tested on a local stack. Confirmed that:
- `Actions` backfills now get all data between the config's start date & the present - data is no longer missed during a backfill if it's "LockingDate" is after the cutoff date.

After this change is merged:
- All existing tasks will need their config's `stop_date` to be renamed to `start_date`.
- The documentation URL in Supabase for this connector should be updated to "https://go.estuary.dev/source-impact-native".

Snapshot changes are expected due to the `stop_date` to `start_date` rename, updating the documentation URL, and increasing the interval to `PT5M`. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2186)


jonwihl (approved):
LGTM

PR #2187, opened 2024-12-06
source-redshift-batch: Add the 'use_schema_inference' feature flag

willdonnelly:
**Description:**

This PR introduces the `/advanced/feature_flags` config setting to the Redshift batch capture, adds a helper function in `source-boilerplate` for parsing those flags with default values and support for `no_` prefixes, and then uses that feature flags infrastructure to add a new one named `use_schema_inference` which when set causes discovered collection schemas to have `x-infer-schema: true` set so that they use the inferred schema in conjunction with the discovered schema.

This is useful because Redshift is a data warehouse where users often use very loose declared types on the tables, and then want to be able to rely on more precise as-used-in-practice type guarantees when materializing their dataset. For example, it's tolerably common to have no declared primary keys and declare all columns with a potentially-nullable type, then manually specify a collection key made of one or more of those properties, and then want to materialize that collection to a SQL database. And currently that doesn't work because the columns designated as the key are potentially nullable in the source DB, even if there are no actual nulls in practice. Turning on schema inference is the escape hatch which allows this to work right up until the moment they actually stick a null into their source data.

**Workflow steps:**

Put `use_schema_inference` in the "Feature Flags" section of the advanced endpoint config. In the future if more feature flags than just this one exist, they should be comma-separated.

Note that once enabled on a particular capture, simply removing the feature flag will not turn schema inference back on -- as I understand things the use of schema inference is "sticky" and disabling it would require manually fiddling with the collection schemas.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2187)

### DIFF
--- a/source-boilerplate/feature_flags.go
+++ b/source-boilerplate/feature_flags.go
@@ -1,0 +1,25 @@
+package boilerplate
+
+import "strings"
+
+// ParseFeatureFlags parses a comma-separated list of flag names and combines that with a
+// map describing default flag settings in the absence of any flags. A flag name can be
+// prefixed with 'no_' to explicitly set it to a false value, in case the default is (or
+// might soon become) true.
+func ParseFeatureFlags(flags string, defaults map[string]bool) map[string]bool {
+	var settings = make(map[string]bool)
+	for k, v := range defaults {
+		settings[k] = v
+	}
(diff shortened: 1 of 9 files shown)


Alex-Bair (approved):
LGTM!

Issue #2001, opened 2024-10-01
source-postgres: Drop and recreate the replication slot when appropriate


willdonnelly:
There are two situations where it makes sense and would be useful if the connector could drop and recreate the replication slot automatically:

- If our resume cursor is empty (either because this is a new capture or because the user has elected to backfill all bindings) and the replication slot's `restart_lsn` is significantly (multiple GiB) behind the current WAL LSN, dropping and recreating the replication slot automatically would allow us to immediately begin processing the latest events, whereas the current behavior requires the database to read through all WAL data from `restart_lsn` to our new start point which can take many minutes and provides zero actual benefit.
- Similarly, if the replication slot has been invalidated, it should be possible to fix this solely by hitting the "backfill everything" button and having the connector automatically drop the old slot and create a new one.

If you squint these are kind of the same situation: if the current slot appears unusable or merely sub-optimal, and our resume cursor is empty/reset meaning we want to jump ahead to the current end of the WAL anyway, then we should drop and create a new replication slot at that point.

I believe that the `pg_drop_replication_slot` and `pg_create_logical_replication_slot` operations are always permitted for a user with the `REPLICATION` permission, which our capture user must already have to do its job, so there should be no added potential for things to fail when we do this.

PR #2189, opened 2024-12-06
source-postgres: Drop/recreate slot on backfills

willdonnelly:
**Description:**

This commit adds logic to automatically drop and then recreate the replication slot when there is no prior replication cursor to resume from.

This has two effects:
- It avoids situations where the replication slot could be created by an initial validation operation but then the capture itself doesn't start for hours and has to read through the entire WAL between those points before actually reaching events it cares about.
- It allows users to recover from replication slot invalidation by simply hitting the "Backfill Everything" button, whereas before they had to go and drop the slot themselves as well.

This fixes https://github.com/estuary/connectors/issues/2001

**Workflow steps:**

Users should no longer have to manually drop replication slots pretty much ever. Simply hitting the backfill button should be sufficient.

**Notes for reviewers:**

This change is conceptually pretty simple, but it involves _deleting an irreplaceable database entity_ and tricky to test automatically (beyond what our normal CI tests exercise). Consequently I have tried to make the logic as fail-safe as possible:
- Any errors in this drop/recreate process are logged but do not cause an early exit, so we should always issue the create query no matter what.
- The code only runs when the cursor value is empty in the first place, so it should only apply to new captures and ones where the "Backfill Everything" button is hit. Theoretically (fingers crossed) there's no way this can break existing captures without explicit user action even if the drop+recreate logic were incorrect somehow.

However this still feels risky enough that I intend to hold off on merging it until next week and make sure I'm good and caffeinated when I do, just in case.


jgraettinger (approved):
LGTM

I didn't spot anything that could go wrong here.

PR #2190, opened 2024-12-06
Network Tunneling in Batch SQL Connectors

willdonnelly:
**Description:**

Integrates network/SSH tunneling into the batch SQL connectors, with the exception of `source-bigquery-batch` since it's not self-evident how to apply it there and I'm just picking some extremely low-hanging fruit here.

**Notes for reviewers:**

In order to avoid copy-pasting a bunch of unnecessary bits of boilerplate, I have factored out some of the config structs and startup logic that other connectors used into the network tunnel Go package and am using that here. I have not gone back and modified the other dozen connectors with SSH tunneling support to use the more concise integration in this PR, but it's something we might want to do at some point just for code health reasons.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2190)

### DIFF
--- a/go/network-tunnel/interface.go
+++ b/go/network-tunnel/interface.go
@@ -1,0 +1,41 @@
+package networkTunnel
+
+import (
+	"context"
+	"fmt"
+	"net"
+)
+
+type SSHForwardingConfig struct {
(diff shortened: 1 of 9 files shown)


jgraettinger (approved):
LGTM

PR #2195, opened 2024-12-11
source-braintree-native: run synchronous API calls in separate threads

Alex-Bair:
**Description:**

The Braintree SDK makes synchronous HTTP requests, which was blocking the main thread and preventing separate streams from sending concurrent API calls. Any HTTP requests made by the Braintree SDK are now wrapped in [`asyncio.to_thread`](https://docs.python.org/3/library/asyncio-task.html#asyncio.to_thread), which runs the function in a separate thread. This unblocks the main thread and lets separate streams make concurrent API calls.

The `maximum_size` property that indicates how many results are found in a specific date window is now used to throw search limit errors sooner instead of after we iterate through & count up all results.

Note: The `disputes` stream cannot use the `maximum_size` property because the Braintree SDK returns a `PaginatedCollection` instead of a `ResourceCollection` for `disputes`, and `PaginatedCollection`s don't have a `maximum_size` property.

**Workflow steps:**

**Notes for reviewers:**

- HTTP requests made by the Braintree SDK are done in separate threads, and separate streams are able to make concurrent API calls.
- Search limit errors are still raised when a search limit is exceeded.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2195)

### DIFF
--- a/source-braintree-native/source_braintree_native/api.py
+++ b/source-braintree-native/source_braintree_native/api.py
@@ -1,3 +1,4 @@
+import asyncio
 from braintree import (
     BraintreeGateway,
     AddOnGateway,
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2196, opened 2024-12-11
materialize-kafka: new connector

### LINKED ISSUE #2138: materialize-kafka: new materialization with support for Avro and schema registry

williamhbaker:
**Description:**

New connector for materializing data to Kafka topics. It can encode messages as JSON, or Avro when there is a schema registry configured.

There were initial attempts made at an exactly-once mode of operation, where Kafka producer transactions were used and checkpoints were written to a checkpoints topic that could then be recovered. This was taken out as a simplification, and because of complications that would likely arise from the existence of Kafka transaction timeouts, which limit how long transactions can be open. An at-least-once non-transactional mode of operation will be needed no matter what because of this timeout consideration, and so the connector is being initially released with only that capability and we can add transaction support later if there is an actual need for it.

Field selection is a little different for this connector than for example SQL materializations: Here, all top-level fields are recommended by default, so that the default path is for users to end up with their entire document materialized nicely without having to do anything different. It is also possible to de-select top-level fields, or include additional nested fields if that is needed for some reason.

Closes #2138

**Workflow steps:**

I will write new documentation for the connector along with merging this PR.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2196)


materialize-kafka/src/transactor.rs
```diff
@@ -0,0 +1,211 @@
...
+};
+use tuple::Element;
+
+use crate::{
+    binding_info::{get_binding_info, BindingInfo},
+    configuration::{EndpointConfig, MessageFormat},
+    Input, Output,
+};
+
+pub async fn run_transactions(mut input: Input, mut output: Output, open: Open) -> Result<()> {
```
mdibaiee, line 23: I imagine we want to extract this into a library at some point for writing Rust connectors? this seems like `materialize-boilerplate`, but for Rust

materialize-kafka/src/transactor.rs
```diff
@@ -0,0 +1,211 @@
...
+fn convert_tuple(packed: &Bytes) -> Vec<serde_json::Value> {
+    tuple::unpack::<Vec<_>>(packed)
+        .expect("tuple must unpack")
+        .iter()
+        .map(|element| match element {
+            Element::Bool(v) => serde_json::json!(v),
+            Element::Int(v) => serde_json::json!(v),
+            Element::Float(v) => serde_json::json!(v),
+            Element::Double(v) => serde_json::json!(v),
+            Element::String(v) => serde_json::json!(v),
```
mdibaiee, line 205: nit: this could be rewritten like so

```suggestion
            Element::Bool(v) |
            Element::Int(v) |
            Element::Float(v) |
            Element::Double(v) |
            Element::String(v) => serde_json::json!(v),
```

materialize-kafka/src/transactor.rs
```diff
@@ -0,0 +1,211 @@
...
+            loop {
+                match producer.send(msg) {
+                    Ok(_) => {
+                        break;
+                    }
+                    Err((err, original_message)) => match err {
+                        KafkaError::MessageProduction(RDKafkaErrorCode::QueueFull) => {
+                            tracing::debug!("polling producer since queue is full");
+                            producer.poll(Timeout::Never);
+                            msg = original_message;
```
mdibaiee, line 78: is this line necessary? 🤔

mdibaiee (commented):
LGTM in general, just a few questions

materialize-kafka/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
```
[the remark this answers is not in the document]
williamhbaker: Definitely! I had experimented a little with doing that in this PR but decided to hold off for the time being. There's some re-org/cleanup we could do with the common packages for the Go connectors as well. I'm thinking it'll be a bit of general work for adapting the repo to more of a multi-language format than we have previously assumed.

materialize-kafka/src/transactor.rs
williamhbaker: Yep for sure. Probably with the next Rust materialization we write, whenever that is, especially if it needs to do standard updates.

materialize-kafka/src/transactor.rs
williamhbaker: Hmm when I tried this it didn't work - I think the `v` would all have to be the same type, but they are different according to their enum variant, for example `bool`, `i64`, `f32`, etc.

materialize-kafka/src/transactor.rs
williamhbaker: `producer.send` takes ownership of its `msg` argument, so without this line the borrow checker complains that on subsequent iterations of the retry loop `msg` has been moved. The `producer.send` error response gives you back the message it took ownership of and this reassignment of that to the original `msg` lets the next round of the loop use that one (which is the same one, really) rather than the moved variable.

materialize-kafka/src/binding_info.rs
```diff
@@ -0,0 +1,331 @@
...
+        .collect();
+
+    let key_ptr = field_selection
+        .keys
+        .iter()
+        .map(|k| {
+            // The selected key fields are treated as unescaped strings. For
+            // example, a key field like `some/nested~1key` is taken directly as
+            // a property named "some/nested~1key" with that pointer rather than
+            // unescaping the ~1 into a nested pointer of ("some", "nested/key").
```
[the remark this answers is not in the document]
williamhbaker: If you have a document with a structure like this:

```json
{
  "some": {
    "nested/key": "value"
  } 
}
```

You can select the "nested/key" field, which is nested in the containing "some" object, and Flow provides a flattened alias for that field that is the string "some/nested~1key" - the "nesting" is represented by slashes, and actual slashes in the field names are escaped as `~1`.

SQL materializations etc. take that alias directly and create a column for it. We want to do the same thing with Kafka, creating a top-level property if a user explicitly selects a nested field (nested fields are not recommended by default, since the full structure of all top-level fields is materialized by default, so there's not really any reason to select nested fields, but you could if you want to).

The use of `Token::from_str` is in contrast to using `Pointer::from_str` more directly since `Pointer::from_str` would interpret such a string as representing a nested structure rather than just treating it as the desired named of the field.

It's all pretty obscure and not likely to come up very often, but it was an edge case I uncovered in my testing.

mdibaiee (approved):

PR #2198, opened 2024-12-12
source-http-ingest: support path parameters

psFried:
**Description:**

Introduces support for capturing path parameters from incoming requests.  The paths in the endpoint config are now allowed to specify parameters using the OpenAPI syntax.  For example, the path
`/vendors/{vendorId}/products/{productId}` can now be used to extract the `vendorId` and `productId`, which will be added to a `/_meta/pathParams` JSON object. The OpenAPI syntax was chosen because it's well-known and documented, and because it's simple and limited. This makes it easy to parse and less likely to result in surprising behavior. The syntax is documented at: https://swagger.io/docs/specification/v3_0/paths-and-operations/#path-templating

The `/_meta/reqPath` property was also added in order to record the configured path at which the request was handled.

Any path parameters are added as required properties in the discovered collection writeSchema. This allows them to be used as collection keys just by editing the collection in the UI during capture creation.

**Workflow steps:**

- Enter a path in the endpoint config like `/{paramA}/{paramB}`
- Send a request to `/foo/bar`
- The captured document will include:

```json
{
  "_meta": {
    "pathParams": {"paramA": "foo", "paramB": "bar"}, 
    "reqPath": "/{paramA}/{paramB}"
  }
}
```

I'll put up a separate PR to update: https://docs.estuary.dev/reference/Connectors/capture-connectors/http-ingest/

**Notes for reviewers:**

This feature was suggested by a user [in Slack](https://estuary-dev.slack.com/archives/C06R0CVD0KZ/p1733795655689899?thread_ts=1733790312.013659&cid=C06R0CVD0KZ), who would otherwise need to configure many different endpoint paths.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2198)


williamhbaker (approved):
LGTM

PR #2201, opened 2024-12-12
source-mysql: Fix primary-key update logic when applied to only-changes bindings

willdonnelly:
**Description:**

We recently (in https://github.com/estuary/connectors/pull/2185) added some logic to `source-mysql` which detects row update events which modify the collection key of the row. Or more pedantically, it detects row updates which modify the value of "whatever columns we consider the row key to be made of", which is almost always the collection key.

_(We play a little fast-and-loose with the concepts at times and it works out because in practice nobody ever sets the collection key to something other than the source DB primary key or tries to tells us to backfill using some other key distinct from the collection key. In fact we should probably deprecate the ability to have a backfill key distinct from the collection key since it's an undocumented backwards-compatibility feature from years ago and I bet nobody is actually using it)._

This distinction matters because in the past bindings with the `"Only Changes"` backfill mode haven't needed to care what the backfill key / scan key / collection key / whatever-you-want-to-call-it is set to, and so we haven't even initialized that state property for such bindings. They go straight from pending to active and don't set `KeyColumns` along the way.

But now, we need to detect update events which modify the row's key, even if they're on an `"Only Changes"` binding. This means that either we need to add entirely new plumbing into the replication logic for "what properties correspond to the collection key", or we need to just accept that this is already what `KeyColumns` holds and therefore it's a bug that we're not setting it on only-changes bindings.

I have opted to go with the latter solution because it's simpler and feels more elegant anyway.

**Workflow steps:**

No user action is required, and the new primary-key update logic in MySQL should now work reliably for only-changes bindings going forward.

**Notes for reviewers:**

This PR is pretty small but if you find it at all confusing it might be clearer when viewed commit-by-commit. It has three parts:

1. A test case demonstrating the issue and eventual fixed behavior.
2. A small refactoring of the `activatePendingStreams` so that any future only-changes bindings will have a correctly initialized `KeyColumns` property.
3. A bit of temporary migration logic in `reconcileStateWithBindings` which identifies preexisting only-changes bindings which are already active with a nil `KeyColumns` and fills in the appropriate value. This can be removed in the near future after it's had a chance to run everywhere that matters.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2201)

### DIFF
--- a/source-mysql/.snapshots/TestPrimaryKeyUpdateOfOnlyChangesBinding
+++ b/source-mysql/.snapshots/TestPrimaryKeyUpdateOfOnlyChangesBinding
@@ -1,13 +1,15 @@
 # ================================
-# Collection "acmeCo/test/test_primarykeyupdateofonlychangesbinding_51329336": 5 Documents
+# Collection "acmeCo/test/test_primarykeyupdateofonlychangesbinding_51329336": 7 Documents
 # ================================
 {"_meta":{"op":"c","source":{"ts_ms":1111111111111,"schema":"test","table":"PrimaryKeyUpdateOfOnlyChangesBinding_51329336","cursor":"binlog.000123:56789:123","txid":"11111111-1111-1111-1111-111111111111:111"}},"data":"three","id":3}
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

PR #2202, opened 2024-12-12
source-postgres: Support CDC without watermarks

willdonnelly:
**Description:**

This PR replaces the PostgreSQL CDC fence logic which was previously based on detection of our own watermark writes with some new (actually heavily copied from MySQL) fence logic based entirely on WAL LSN positions.

However by default we still write a watermark as part of establishing a fence position, in order to guarantee that there will be at least one WAL event whose LSN is greater than the minimum fence LSN. So the other half of this PR was adding a new advanced setting (`"Read-Only Capture"`) which disables that behavior.

Disabling watermark writes is not without its tradeoffs. Put simply, due to the way Postgres reorders and cleans up the underlying WAL events when doing logical replication, we can't always guarantee that the latest LSN value obtained from `pg_current_wal_flush_lsn()` (or any of its sibling functions) actually corresponds to an event which we'll be told about. This can most trivially be proven when you consider the situation of an idle database on a server which hosts other active databases, but it's actually possible to run into similar trouble even when there's just the one database on the server. So when enabling read-only mode one must be sure that the server will always have a modest rate of changes on the database which is being captured from.

But the upshot is that we can now run captures without any "writes" to the database (so long as you don't consider replication slot management a write, because that's unavoidable) and it's even possible to capture from read-only standby replicas (with a big big big asterisk that says "except sometimes those setups can be very prone to invalidating replication slots if there's even a tiny bit of downtime"). Yay.

**Workflow steps:**

To continue using the connector as before, no action is required. The change from watermark value detection to LSN position fencing should be invisible to users, and we still by default write watermarks at the same cadence as before.

To set up a new capture which doesn't write watermarks, enable the "Read-Only Capture" toggle in the advanced endpoint config.

We will need to update PostgreSQL documentation to discuss the new option and the user's responsibility to ensure that the database won't be fully idle when using it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2202)

### DIFF
--- a/source-postgres/.snapshots/TestGeneric-SpecResponse
+++ b/source-postgres/.snapshots/TestGeneric-SpecResponse
@@ -93,6 +93,11 @@
             "title": "Minimum Backfill XID",
             "description": "Only backfill rows with XMIN values greater (in a 32-bit modular comparison) than the specified XID. Helpful for reducing re-backfill data volume in certain edge cases.",
             "pattern": "^[0-9]+$"
+          },
+          "read_only_capture": {
+            "type": "boolean",
+            "title": "Read-Only Capture",
+            "description": "When set the capture will operate in read-only mode and avoid operations such as watermark writes. This comes with some tradeoffs; consult the connector documentation for more information."
           }
         },
         "additionalProperties": false,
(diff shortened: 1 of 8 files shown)


willdonnelly:
> Seems like the tests are broken in CI

Whoops, fixed now.

PR #2208, opened 2024-12-16
source-braintree-native: update `transactions` to fetch updates

Alex-Bair:
**Description:**

Previously, the `transactions` stream only incrementally captured creates since Braintree does not expose the `updated_at` field for API searches.

The `transactions` stream has been updated to capture updates, assuming Braintree's `updated_at` field is just the latest of the various searchable `???ed_at` fields. The strategy is:
1. Get the unique ids of all transactions that have been created, voided, settled, etc. (i.e. figure out what transactions have been
updated) in the date window.
2. Request the full transaction objects associated with the unique ids fetched in step 1. A scatter/gather strategy is used to fetch batches of transactions, yielding them via an `asyncio.Queue` as they arrive.

The `transactions` stream also has a distinct backfill task now. It ignores any documents with an `updated_at` field that's after the cutoff date, meaning that the incremental task will pick up the updated document.

**Workflow steps:**

Docs should be updated to reflect that the `transactions` stream captures updates and does not require regular backfills.

**Notes for reviewers:**

- Snapshot tests passed.
- The number of unique `transactions` document IDs in a date window fetched by `_fetch_unique_updated_transaction_ids` equals the number of documents yielded with the scatter/gather strategy within `fetch_transactions`.
- Confirmed `transaction` backfills complete.

### DIFF
--- a/source-braintree-native/source_braintree_native/api.py
+++ b/source-braintree-native/source_braintree_native/api.py
@@ -12,6 +12,7 @@ from braintree import (
     )
 from braintree.attribute_getter import AttributeGetter
 from braintree.paginated_collection import PaginatedCollection
+from braintree.search import Search
 from datetime import datetime, timedelta, UTC
 from logging import Logger
 from typing import AsyncGenerator
@@ -27,6 +28,18 @@ from .models import (
 # the connector could have missed data and we'll need to use smaller date windows.
 SEARCH_LIMIT = 10_000
 TRANSACTION_SEARCH_LIMIT = 50_000
+
+TRANSACTION_SEARCH_FIELDS = [
+    'authorization_expired_at',
+    'authorized_at',
+    'created_at',
+    'failed_at',
+    'gateway_rejected_at',
+    'processor_declined_at',
+    'settled_at',
+    'submitted_for_settlement_at',
+    'voided_at',
+]
 
 CONVENIENCE_OBJECTS = [
     'gateway'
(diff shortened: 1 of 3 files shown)


source-braintree-native/source_braintree_native/api.py
```diff
@@ -108,36 +132,134 @@ async def snapshot_resources(
...
+    return collection.ids
+
+
+async def _fetch_unique_updated_transaction_ids(
+    braintree_gateway: BraintreeGateway,
+    start: datetime,
+    end: datetime,
+    log: Logger,
+) -> list[str]:
+    async with asyncio.TaskGroup() as tg:
```
williamhbaker, line 162: I think this could be simplified using `asyncio.gather`

source-braintree-native/source_braintree_native/api.py
```diff
@@ -108,36 +132,134 @@ async def snapshot_resources(
...
+    ids = await _fetch_unique_updated_transaction_ids(
+        braintree_gateway,
+        log_cursor,
+        end,
+        log,
+    )
+
+    queue = asyncio.Queue(maxsize=10)
+
+    async with asyncio.TaskGroup() as tg:
```
williamhbaker, line 214: You could probably use `asyncio.gather` here as well.

A possible consideration is memory usage if there are a lot of batches which all get results concurrently before their documents can be emitted. I doubt this will be a problem in practice since the speed at emitting documents should be quite quick in comparison to fetching batches.

source-braintree-native/source_braintree_native/api.py
Alex-Bair: So something more like this?
```python
tasks: list[Task[list[str]]] = []
for field in TRANSACTION_SEARCH_FIELDS:
    task = asyncio.create_task(
        _fetch_transaction_ids_for_single_field(
            field,
            braintree_gateway,
            start,
            end,
            log,
        )
    )

    tasks.append(task)

await asyncio.gather(*tasks)
```

Is `asyncio.gather` preferred here because we don't need all the extra bells & whistles around task cancellation that come with `asyncio.TaskGroup`?

williamhbaker: `asyncio.gather` is just a more succinct way to express what is happening. It's basically a helper for this common pattern. Can you do something like [this](https://github.com/estuary/connectors/blob/027f511b7deedfdbb19fd641244f6827f8a985d2/source-hubspot-native/source_hubspot_native/api.py#L220-L225)?

Alex-Bair: Got it, I see now. Yep, I can do that.

source-braintree-native/source_braintree_native/api.py
Alex-Bair: I'll refactor to the `asyncio.gather` pattern you linked in your other comment.

I agree about memory usage. Theoretically with the current batch size of `TRANSACTION_SEARCH_LIMIT`, there could be at max 9 batches (the same as the the number of transaction search fields) if all of them return unique transaction IDs within the time window. In practice, there's a lot of overlap between some of these (like `created_at` and `submitted_for_settlement_at`), and searches for the non-`created_at` fields usually return fewer IDs than `created_at` searches. I could add some handling to only allow X batches to fetch results at a time, but I suggest we could wait & see if that's actually needed.

Alex-Bair: I was able to refactor this one, but I had to omit the `await`. Otherwise, the connector would get stuck putting docs onto the queue but never taking them off.

williamhbaker: The change in `_fetch_unique_updated_transaction_ids` looks good 👍 

For `fetch_transactions` I see how the same approach won't work, since we specifically _don't_ want to keep every single full record in memory.

I'm not 100% sure what you've got in `fetch_transactions` right now will work. I think the gather has to be await'd at some point for it to be assured to complete. 

As an alternative to using a queue, take a look at `asyncio.as_completed` - I think that would do what we want, which is to eagerly process the results of each task as it becomes available.

Alex-Bair: Based off our discussion, I made `fetch_transactions` fetch batches in sequence instead of in parallel. It'll probably be uncommon to have more than 1-2 batches while we use a batch size of 50,000. We can look into reducing batch size & processing the batches in parallel at a later point.

PR #2212, opened 2024-12-16
source-oracle-flashback: float types with precision < 18 are not string

mdibaiee:
**Description:**

- When emitting documents, [we emit numbers](https://github.com/estuary/connectors/blob/main/source-oracle-flashback/source_oracle_flashback/api.py#L326-L328) with a precision higher than 18 as python's `decimal.Decimal`, which is then encoded as a JSON string. This is applied on all numeric types. When doing discovery, we were correctly discovering `NUMBER` columns with a precision of less than 18 as `type: number`, and those with a higher precision as `type: string, format: number`, but we were not doing the same thing for `FLOAT` types (`DOUBLE PRECISION` is an alias for `FLOAT(126)`). Updated that code so that discovery's behavior matches the emitted values.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2212)


Alex-Bair (approved):
LGTM

PR #2217, opened 2024-12-17
Bair/braintree misc improvements

Alex-Bair:
**Description:**

This PR's scope includes:
- dynamically reducing window sizes for most incremental streams if they hit the limit for a single Braintree search.
  - Window sizes that are too large are cut in half until the number of returned results is below the search limit.
  - This is implemented for all incremental streams except `disputes`. `disputes` does not indicate how many results are in a search, meaning we have to iterate over & count results to know if the search limit is hit. Our CDK does not have a "rollback" mechanism to discard/ignore previously yielded but not checkpointed documents. So `disputes` still crashes if the configured window size is too large.
- a couple bug fixes around `disputes`
  - The start date does not need pushed back a day for each sweep. I initially thought `DisputeSearch.received_date.between(start,end)` was exclusive of the `start` date, but that was incorrect.
  - `most_recent_created_at` is now updated whenever a document's `created_at` is more recent than `most_recent_created_at`. Previously, this was updated when `created_at` was more recent than the previous `log_cursor`, and since results are returned in descending order, we were yielding duplicate documents and not updating the cursor to the most recent `created_at` value.

**Workflow steps:**

**Notes for reviewers:**

- Window sizes for non-`disputes` incremental streams are reduced until there are fewer than `search_limit` results in the date window.
- If the window size is less than 1 second, the connector errors.
- `disputes` now updates its cursor to the most recent document's `created_at` whenever any documents are found in a date window.

### DIFF
--- a/source-braintree-native/source_braintree_native/api.py
+++ b/source-braintree-native/source_braintree_native/api.py
@@ -384,15 +384,15 @@ async def fetch_disputes(
 ) -> AsyncGenerator[IncrementalResource | LogCursor, None]:
     assert isinstance(log_cursor, datetime)
     most_recent_created_at = log_cursor
-    # The start date must be shifted back 1 day since we have to query Braintree using the received_date field,
-    # which is less granular than the created_at cursor field (date vs. datetime).
-    start = log_cursor - timedelta(days=1)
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2218, opened 2024-12-18
source-genesys: fix `conversations` default param bug

Alex-Bair:
**Description:**

In Python, default parameters are evaluated once when the function definition is executed ([source](https://docs.python.org/3/reference/compound_stmts.html#function-definitions)). This means that a default param of `datetime.now(tz=UTC)` is only evaluated when the program starts up, and that same value is used for each subsequent function call. Meaning that within the connector, `_perform_conversation_job` was using a fixed end date and not incrementing it over time.

Now, a default value of `None` is used in `_perform_conversation_job` to indicate that the current time should be used as the end date.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that not providing an `end_date` argument for `_perform_conversation_job` will cause `end_date` to be the current datetime instead of when the connector last restarted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2218)

### DIFF
--- a/source-genesys/source_genesys/api.py
+++ b/source-genesys/source_genesys/api.py
@@ -87,13 +87,17 @@ async def _perform_conversation_job(
         domain: str,
         log: Logger,
         start_date: datetime,
-        end_date: datetime = datetime.now(tz=UTC),
+        end_date: datetime | None = None,
 ) -> AsyncGenerator[Conversation, None]:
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2220, opened 2024-12-19
source-oracle: support CDB databases

mdibaiee:
**Description:**

- Support CDB databases. It is simpler than I thought, basically either the user connects directly to a root container database (unlikely), or they connect to a pluggable database (PDB). What we need to do is essentially switching the session to the CDB when running logminer, since logminer cannot run on PDBs.
- I tested this by creating a container database in Oracle Cloud and running one of our tests against it. I have updated the tests of the connector so that they are run both against a non-container database and a container one. I have not run the full test suite yet... some connectivity issues from my machine are making it difficult... I'll see if I can do it
- Docs pull-request to go with this: https://github.com/estuary/flow/pull/1843

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2220)


source-oracle/replication.go
```diff
@@ -53,6 +53,13 @@ func (db *oracleDatabase) ReplicationStream(ctx context.Context, startCursor str
 		return nil, fmt.Errorf("set NLS_TIMESTAMP_TZ_FORMAT: %w", err)
 	}
 
+	// Logminer cannot run on PDB instances, so we switch to the CDB
```
williamhbaker, line 56: Can `switchToCDB` be used here?

source-oracle/replication.go
```diff
@@ -204,9 +211,10 @@ func (s *replicationStream) addLogFiles(ctx context.Context, startSCN, endSCN in
 			return fmt.Errorf("adding logfile %q (%s, %d, %s, %s) to logminer: %w", f.Name, f.Status, f.FirstChange, f.DictStart, f.DictEnd, err)
 		}
 	}
+
 	s.redoSequence = redoSequence
 
-	return nil
+	return err
```
williamhbaker, line 217: Can `err` ever not be `nil` here? I'm more used to seeing the `return nil` at the end of a function, and returning `err` makes me think there is some way for it to be set to non-nil somewhere in the function, but I'm not seeing anywhere for that to be the case.

williamhbaker (approved):
LGTM

source-oracle/replication.go
mdibaiee: we use a separate connection for replication and that method runs on `db.conn`, whereas this one is run on a new and separate connection

source-oracle/replication.go
mdibaiee: don't know why I have `return err` here! it is super confusing, changed it to `return nil` 👍🏽

PR #2222, opened 2024-12-19
source-hubspot-native: conditionally discover `email_events` binding

Alex-Bair:
**Description:**

Not all HubSpot accounts have enough permissions to access the endpoint for the `email_events` binding. This binding is now discovered only if we can successfully hit the associated endpoint without a permissions error.

The permissions checking code isn't very reusable, but it can be refactored later if/when the connector needs to conditionally discover other bindings.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that `email_events` is discovered if a user's HubSpot account has the appropriate permissions, and `email_events` is not discovered if the user's HubSpot account does not have the appropriate permissions.

Existing captures that have `email_events` enabled but whose tokens do not have the appropriate permissions should either disable the `email_events` binding or perform discoveries to remove `email_events` as a possible binding. The latter will likely be handled by auto-discovers.

The discover snapshot change is expected since the order of the bindings returned by `all_resources` has changed (`email_events` is now at the end of the list).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2222)

### DIFF
--- a/source-hubspot-native/source_hubspot_native/resources.py
+++ b/source-hubspot-native/source_hubspot_native/resources.py
@@ -2,12 +2,13 @@ import functools
 import itertools
 from datetime import UTC, datetime, timedelta
 from logging import Logger
+import re
 from typing import AsyncGenerator, Iterable
 
 from estuary_cdk.capture import Task
 from estuary_cdk.capture.common import LogCursor, PageCursor, Resource, open_binding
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #2225, opened 2024-12-20
source-twilio: capture `messages` updates based on `lookback_window`

Alex-Bair:
**Description:**

Twilio's API only allows us to query messages based on their creation date `date_sent`, not their updated date `date_updated`. Previously, the connector was only getting creates, with the option to use the `lookback_window` config setting to push back the
cursor value, causing all messages created in the past `lookback_window` minutes to get re-captured. The `messages` stream now uses the lookback window to query Twilio for messages created `lookback_window` minutes in the past, but it only captures messages that have been created or updated after the current cursor value.

**Workflow steps:**

**Notes for reviewers:**

- `messages` captures updates to records made in the past `lookback_window` minutes.
- `messages` still captures creates.
- `message_media` still functions incrementally.


williamhbaker (approved):
LGTM

PR #2229, opened 2024-12-23
source-oracle-flashback: order first, then limit row count

mdibaiee:
**Description:**

- We were using `ROWNUM` to limit the number of rows, but that filter is applied before ordering, and that would lead to inconsistencies if ROWIDs were not in the order of insertion (sometimes ROWIDs can be reused if they have been freed up due to deletion / resize of data in data segments), a common pitfall of `ROWNUM` filtering which I fell for
- This was not  covered by our test cases because reproducing this behaviour is hard, we need ot somehow force the database to allocate an "older" ROWID to a new row
- Instead of using `ROWNUM` filtering we can use `FETCH NEXT %d ROWS ONLY` which has been introduced in Oracle 12c (released 2013). An alternative implementation is using a subquery like `SELECT * FROM (SELECT ...) WHERE ROWNUM < 50000` for older versions, but I'm not sure about its performance implications, and after talking with Dave we haven't had customers ask for Oracle versions below 12c yet. I put a note in the code for that so we can revisit if necessary.
- I tried to write a test to emulate this scenario by deleting rows and inserting rows to see if the database would reuse a ROWID smaller than the largest ROWID, but I was not successful. I'm not sure how the database makes the decision yet (probably need to understand how large are the data segments and how much space needs to be freed before it will be re-allocated to newer rows)

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2229)

### DIFF
--- a/source-oracle-flashback/source_oracle_flashback/api.py
+++ b/source-oracle-flashback/source_oracle_flashback/api.py
@@ -200,9 +200,9 @@ async def fetch_page(
                     c.arraysize = backfill_chunk_size
                     c.prefetchrows = backfill_chunk_size + 1
                     c.outputtypehandler = number_to_decimal
-                    query = template_env.get_template("backfill").render(table=table, rowid=last_rowid or page, max_rowid=cutoff[0])
+                    query = template_env.get_template("backfill").render(table=table, rowid=last_rowid or page, max_rowid=cutoff[0], chunk_size=backfill_chunk_size)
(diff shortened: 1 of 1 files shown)


psFried (approved):
Looks good to me. I'm assuming that none of our existing customers are using oracle < 12c

PR #2232, opened 2025-01-02
source-oracle-flashback: filter out already-processed SCNs

mdibaiee:
**Description:**

`VERSIONS BETWEEN SCN` results in any valid rows for the specified SCN range, and rows which have `VERSIONS_STARTSCN < rangeBegin && VERSIONS_ENDSCN >= rangeBegin` are also considered. In our logic however we only consider `VERSIONS_STARTSCN` as the transaction identifying checkpoint that's why we need the extra `VERSIONS_STARTSCN >= {{cursor}}` to avoid results where `VERSIONS_STARTSCN` is smaller than the latest checkpoint we have emitted.

We had seen this with some tasks where the connector would try to emit a checkpoint that was already processed again, and Estuary CDK complains in those cases. In essence we were re-emitting previously emitted documents.

Tested by running our test suite. I'm not sure how to trivially reproduce the scenario where we have two queries with overlapping SCNs, it requires some rows to have `ENDSCN > STARTSCN` and for us to checkpoint somewhere in between `[STARTSCN, ENDSCN]`

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2232)

### DIFF
--- a/source-oracle-flashback/source_oracle_flashback/api.py
+++ b/source-oracle-flashback/source_oracle_flashback/api.py
@@ -377,6 +377,11 @@ SELECT ROWID, {% for c in table.columns -%}
     ORDER BY ROWID ASC
     FETCH FIRST {{ chunk_size }} ROWS ONLY
 """,
+    # VERSIONS BETWEEN SCN results in any valid rows for the specified SCN range, and rows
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #2234, opened 2025-01-02
source-stripe-native: fix `Accounts` related bugs

Alex-Bair:
**Description:**

- Handling results that don't have a `created` field. 
  - When backfilling, the connector was relying on all results having a `created` field to determine whether the result was within the backfill timespan. However, `Accounts` results do not always have a `created` field (I assume a legacy Stripe system or Stripe support created these accounts & circumvented whatever system adds the `created` field). Since the connector can't tell otherwise without `created`, it assumes the result is within the backfill timespan and yields it.
- Improving handling for when `Persons` cannot be read for an external account.
- Handling insufficient permission errors for `ExternalBankAccount` and `ExternalAccountCards` streams if the error is due to insufficient Stripe Connect permissions.
  - From my brief research, it seems like connecting an external account to a Stripe account requires using the "Stripe Connect" feature, and it's possible to limit what data Stripe can read from this external account. If the user restricts Stripe's permissions in this way, then we receive a `403` response when trying to read the `ExternalBankAccount` and `ExternalAccountCards` streams for the account. I assume these types of restricted permissions are preferred for some users, so the connector logs a warning and continues instead of failing.
- Updating the capture snapshot to include fields that Stripe added.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack with a Stripe account that has a non-trivial number of external accounts. Confirmed:
- `Account`s without `created` fields are yielded during backfills.
- The connector does not crash when trying to read `ExternalBankAccount` and `ExternalAccountCards` and receiving a `403` response with the specific "You cannot perform this request as you do not have Platform Controls for the Stripe Dashboard on this account." message.
- The connector does not request `Persons` if we know the account is an external account & the request would fail with a `403` error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2234)

### DIFF
--- a/source-stripe-native/source_stripe_native/api.py
+++ b/source-stripe-native/source_stripe_native/api.py
@@ -340,7 +340,19 @@ async def _capture_substreams(
 
     # Fetch child records
     while True:
-        if cls_child.NAME == "Persons" and parent_data.controller["requirement_collection"] == "stripe" :
+        # Per the Stripe docs, Persons can't be accessed for accounts where `account.controller.requirement_collection` is "stripe". It also
+        # appears that when `account.controller.type` is "account", then Persons can't be accessed. This handling may need refined further
+        # if we encounter further access issues and our understanding of the Stripe API deepens.
+        # Docs reference: https://docs.stripe.com/api/persons
+        if (
+                cls_child.NAME == "Persons" and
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #2238, opened 2025-01-03
materialize-{bigquery,snowflake}: fix key comparison for composite keys

williamhbaker:
**Description:**

The assumption that the first Store request is the minimum value for all keys of a composite key was incorrect, and could cause invalid merge query conditions for collections with composite keys.

An explicit comparison needs to be done to determine the minimum and maximum observed keys for a transaction, and that is implemented here.

This will have a bit more overhead, but I'm assuming it will be negligible in the grand schema of things. An alternate strategy would be to only base the merge condition on the first part of a composite key, but using all parts of the key is desirable so that users can include additional fields in their collection key if they know these are valid keys and will provide a benefit in merge conditions, rather than being limited to only the first key component.

I also have some suspicion that we may want to eventually include these hints in load queries, and load requests are not ordered at all so they would need a similar comparison, but this is more speculative.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2238)

### DIFF
--- a/materialize-sql/templating.go
+++ b/materialize-sql/templating.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/estuary/flow/go/protocols/fdb/tuple"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


materialize-sql/templating.go
```diff
@@ -64,56 +63,85 @@ type MergeBound struct {
...
-		panic(fmt.Sprintf("application error: %d key fields vs. %d key columns for merge query bounds", len(key), len(b.table.Keys)))
+// NextKey updates the observed minimum and maximum key for a transaction.
+func (b *MergeBoundsBuilder) NextKey(key []any) {
+	if len(key) != len(b.keyColumns) {
+		panic(fmt.Sprintf("application error: %d key fields vs. %d key columns for merge query bounds", len(key), len(b.keyColumns)))
 	}
 
 	if b.lower == nil {
-		b.lower = key
+		b.lower = append([]any(nil), key...)
```
mdibaiee, line 87: is this `append` just copying the data? I don't really understand it

mdibaiee (commented):
Can we have a test case for this scenario?

materialize-sql/templating.go
williamhbaker: I added a comment here to explain it a little better. Let me know if that makes sense!

I had initially tried this with `b.lower = key` and `b.upper = key`, but then they both point to the same slice and that doesn't work for tracking the lower and upper bound separately. So there's this initial copy operation to initialize the values for each transaction, which then will be updated as needed by the code just below here.

The existing tests cover this since `b.lower` is `nil` when initializing the `MergeBoundsBuilder` struct.

mdibaiee (approved):

materialize-sql/templating.go
mdibaiee: I thought we would have updated snapshots for this scenario that's why I asked about test case

PR #2242, opened 2025-01-04
source-hubspot-native: filter search results that are earlier than the requested time range

williamhbaker:
**Description:**

The search API will return records with a modification time before the requested "since" (the start of the window) if their updatedAt timestamp is within the same millisecond, effectively ignoring the microseconds part of the range criteria. These spurious results can be safely ignored in the rare case that there is a record with a modification time within the same millisecond as requested at the start of the time window, but some smaller fraction of a second earlier.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2242)


dyaffe (approved):
LGTM

PR #2254, opened 2025-01-08
source-intercom-native: handle error when hitting /companies/list limit

Alex-Bair:
**Description:**

The `/companies/list` limitation mentioned in https://github.com/estuary/connectors/pull/2204 has been reached by a user, and the connector is crashing as a result. It turns out the API returns a specific `400` error when we try to page past that 10k limit. Since the connector does not currently have an option to use the `/companies/scroll` endpoint & avoid that limit, we should handle this specific error to avoid crashing the connector.

If/when support for the `/companies/scroll` endpoint is added, we should re-evaluate whether or not the connector should crash when hitting the `/companies/list` limit (i.e. force users with more than 10k companies to use the `/companies/scroll` endpoint).

**Workflow steps:**

**Notes for reviewers:**

- A warning is logged if there are more than 10k companies in the user's Intercom account.
- The connector does not crash when trying to paginate past `/companies/list`'s 10k results limit, and the documents that were found within those first 10k results are checkpointed.
- Tests passed locally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2254)

### DIFF
--- a/source-intercom-native/source_intercom_native/api.py
+++ b/source-intercom-native/source_intercom_native/api.py
@@ -1,10 +1,11 @@
 from datetime import datetime, UTC, timedelta
 import json
 from logging import Logger
+import re
 from typing import AsyncGenerator
 
 from estuary_cdk.capture.common import LogCursor
-from estuary_cdk.http import HTTPSession
+from estuary_cdk.http import HTTPSession, HTTPError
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2255, opened 2025-01-08
source-hubspot-native: parallel fetching of batches with associations & optional property history

williamhbaker:
**Description:**

Adds parallel fetching of batches of records with their associations when the full list of record IDs is known, but the records must be enriched further with additional relatively slow API calls - see the first commit message.

In the second commit, adds the groundwork for making the capture of `propertiesWithHistory` optional, by plumbing through the configuration option for that and associated handling. A follow-up change will be necessary to complete this, once all pre-existing tasks have had their configuration set to `true` so that their current behavior is maintained.

Closes [#2248](https://github.com/estuary/connectors/issues/2248)

**Workflow steps:**

Need to update docs for the new configuration option

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2255)


source-hubspot-native/source_hubspot_native/models.py
```diff
@@ -139,7 +145,7 @@ class Owner(BaseDocument, extra="allow"):
 class BaseCRMObject(BaseDocument, extra="forbid"):
     ASSOCIATED_ENTITIES: ClassVar[list[str]]
 
-    class History(BaseDocument, extra="forbid"):
+    class History(BaseModel, extra="forbid"):
```
williamhbaker, line 148: This should have always been `BaseModel` instead of `BaseDocument`, since `BaseDocument` includes top-level document fields like `_meta/op`. More significantly to the discovery snapshot is that the `propertiesWithHistory` won't be output in the "minimal" schema at all, since in the majority of cases users won't capture property history so it should be supressed from the schema.

Alex-Bair (approved):
LGTM! The `buffered_ordered` module seems like it'll be pretty useful.

PR #2256, opened 2025-01-08
source-intercom-native: add config option to use `/companies/scroll` endpoint

Alex-Bair:
**Description:**

Intercom has two separate endpoints for retrieving companies: `/companies/list` (limited to 10k companies) and `/companies/scroll` (limited to only one "scroll" happening at a time). For users that have more than 10k companies in their Intercom account, they will need to use the `/companies/scroll` endpoint to ensure no data is missed.

The connector previously always used the `/companies/list` endpoint, and this PR adds the ability to use the `/companies/scroll` endpoint instead. The new `use_companies_list_endpoint` config option controls which of the two company-related endpoints are used. Within the connector, an `asyncio.Lock` is used to avoid attempting concurrent "scrolls" between the two streams that want to use `/companies/scroll` - `companies` and `company_segments`.

The connector defaults to using `/companies/scroll`, preferring to capture all data instead of potentially missing data if the user has >10k companies. 

It may be worthwhile to remove the option to use `/companies/list` at a later date. The main reasons I anticipate users would check the `use_companies_list_endpoint` option are:
- they have another application using `/companies/scroll` that they can't disable (not sure how common that is)
- they want the slight speed boost in `companies` and `company_segments` (which seems relatively negligible with the limited testing I've done)

**Workflow steps:**

`source-intercom-native` connector docs should be updated to reflect the new `use_companies_list_endpoint` config option.

**Notes for reviewers:**

- Companies can be successfully fetched via the `_scroll_companies` function.
- The `companies` and `company_segments` streams do not concurrently try to hit `/companies/scroll`.
- The `use_companies_list_endpoint` config option controls whether `/companies/list` or `/companies/scroll` is used.
- The connector crashes & logs an error message if it tries to access `/companies/scroll` but receives a `400` response indicating some other application is using the endpoint.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2256)

### DIFF
--- a/source-intercom-native/source_intercom_native/api.py
+++ b/source-intercom-native/source_intercom_native/api.py
@@ -1,3 +1,4 @@
+import asyncio
 from datetime import datetime, UTC, timedelta
 import json
 from logging import Logger
@@ -16,6 +17,7 @@ from .models import (
     ConversationResponse,
     SegmentsResponse,
     CompanyListResponse,
+    CompanyScrollResponse,
     CompanySegmentsResponse,
 )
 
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM, nice!

PR #2259, opened 2025-01-09
materialize-databricks: debugging logs for duplicates

mdibaiee:
**Description:**

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2259)

### DIFF
--- a/materialize-databricks/driver.go
+++ b/materialize-databricks/driver.go
@@ -261,11 +261,25 @@ func (d *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage) 
 				}
 			} else {
 				switch v := converted[0].(type) {
-				case int:
-					if v < bmx.Min.(int) {
+				case int64:
+					if v < bmx.Min.(int64) {
 						bmx.Min = v
 					}
-					if v > bmx.Max.(int) {
+					if v > bmx.Max.(int64) {
+						bmx.Max = v
+					}
+				case uint64:
+					if v < bmx.Min.(uint64) {
+						bmx.Min = v
+					}
+					if v > bmx.Max.(uint64) {
+						bmx.Max = v
+					}
+				case float64:
+					if v < bmx.Min.(float64) {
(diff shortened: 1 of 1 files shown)


materialize-databricks/driver.go
```diff
@@ -246,6 +252,31 @@ func (d *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage)
...
 			return fmt.Errorf("encoding row for load: %w", err)
+		} else {
+			bindingName := b.target.Identifier
+			if bmx, ok := bindingMinMaxKeys[bindingName]; !ok {
+				bindingMinMaxKeys[bindingName] = minMax{
+					Min: converted[0],
+					Max: converted[0],
+				}
+			} else {
+				switch v := converted[0].(type) {
```
williamhbaker, line 263: See [here](https://github.com/estuary/connectors/blob/96714ea85e5e4ed5405fc6c7cef843d645c4ffde/materialize-sql/templating.go#L103-L119) for what I believe are the full set of types that keys can take. They might be encoded `uint64`, or possibly `float64` (not 100% sure on that) if the otherwise-integer values are really big.

materialize-databricks/driver.go
```diff
@@ -477,11 +514,43 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
 				}
 				return nil, fmt.Errorf("query %q failed: %w", query, err)
 			}
+
+		}
+
+		var binding = d.bindingForStateKey(stateKey)
```
williamhbaker, line 520: You might run this both before and after the Store query is run, that way we can tell if there were already duplicates in the table before our query ran.

williamhbaker (approved):
LGTM, in terms of a debug-only build that we won't actually be merging. I don't see any reason to think that this would cause any kind of data corruption. It should only be a performance hit from running the duplicate-finder queries.

Issue #1173, opened 2024-01-16
Update Databricks to retry if it hits a timeout


dyaffe:

dyaffe:
It would be ideal if our Databricks materialization retried after hitting a timeout that was caused by needing to turn the warehouse on.  

Serverless warehouses can be turned on in 30-40 seconds so its reasonable for us to wait until they're on.  Warehouses with a server take 5 minutes, so it's probably not ideal to wait for them to come back up.

psFried:
Noting that we need to have the connector log when it's waiting for a databricks server to come up, so that users can see the log during a publish and know why it's taking so long (during the validate RPC).

williamhbaker:
Part of this was implemented in https://github.com/estuary/connectors/commit/1cf4936bfa37cc8852a1f87df3b4c68ca7dbcbae, where we will no longer need to query the warehouse for existing tables and their column metadata.

We still need a way to not have to query the prior spec out of the warehouse though to fully realize the intent here.

PR #2262, opened 2025-01-09
materializations: don't persist a spec in the destination system

williamhbaker:
**Description:**

Implements using the prior materialization specification from Validate and Apply requests, rather than loading it out of and persisting it into the destination system somehow. This is good since it eliminates a metadata table we have to create, and some systems have limitations on how large a spec can be. There's quite a bit of simplification overall.

The first several commits are mostly maintenance on our Validate & Apply test suite, getting them to run reliably for me on my local machine, which was important to verify that everything still works as expected with the tests after not loading the applied spec from the metadata table anymore.

See individual commit messages for some more details.

Closes https://github.com/estuary/connectors/issues/2076
Closes https://github.com/estuary/connectors/issues/1173

**Workflow steps:**

N/A

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2262)


materialize-sql/driver.go
```diff
@@ -200,6 +189,22 @@ func (d *Driver) Apply(ctx context.Context, req *pm.Request_Apply) (*pm.Response
 		return nil, err
 	}
 
+	if endpoint.MetaCheckpoints != nil && !is.HasResource(endpoint.MetaCheckpoints.Path) {
```
williamhbaker, line 192: I moved the concept of a "metadata table" entirely out of the materialize-boilerplate and into materialize-sql. The checkpoints table is now created here if it is needed.

mdibaiee (approved):
LGTM, feels good to see a lot of code gone!

PR #2267, opened 2025-01-11
source-*-batch: Timeouts and partial-progress logging

willdonnelly:
**Description:**

This PR adds no-data watchdog timeouts to all of the batch SQL connectors, such that if a long period of time passes without any rows being received (when we're awaiting polling query results) it will kill the capture. The time period is pretty generous: 30 minutes when we're waiting for the first result row and 5 minutes for subsequent rows. Pretty much the only times this watchdog should fire are:

 - When the database connection has silently dropped on us (which is a thing that happens sometimes, especially when using SSH tunneling)
 - When the polling query is performing a full-table sort on some massive dataset. And generally speaking if that hasn't finished after 30 minutes it's not likely that we'll actually end up with a successful capture and a pleasant user experience anyway.

I'm still a _little bit_ concerned that even a 30 minute timeout on the initial result rows might cause us to error out in situations where we otherwise could have just kept waiting and eventually gotten results. For example if it takes an hour to sort a particular table but we're only polling it once a day and the user is okay with the DB load of that full-table sort, then in that one particular scenario the user would be better-served by a connector without any such watchdog timeout.

But I think that in practice that scenario basically never happens. Note that when we're doing a cursorless full-refresh of a table we just issue a `SELECT * FROM foobar` query without any ordering, so the ordering clause only comes into play when the user has set up a cursor and clearly _wants_ more efficient incremental sync behavior. So 30 minutes is probably fine. Even ten minutes would probably be fine.

This PR also adds a couple much smaller improvements:

- Partial-progress logging so we can tell when looking at the task logs whether a capture was sitting there idly waiting for results, or whether it was furiously processing results but there were just a lot of them.
- MySQL connection timeouts, the same as we use on the MySQL CDC connector. This is just for consistency with the CDC connector, and because I was already working on "batch MySQL timeouts" and so it seemed related-enough to throw in at the same time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2267)

### DIFF
--- a/source-bigquery-batch/driver.go
+++ b/source-bigquery-batch/driver.go
@@ -701,12 +701,19 @@ func (c *capture) poll(ctx context.Context, binding *bindingInfo, tmpl *template
 				return err
 			}
 		}
+		if count%100000 == 1 {
+			log.WithFields(log.Fields{
+				"name":  res.Name,
+				"count": count,
+			}).Info("processing query results")
+		}
 	}
 
 	log.WithFields(log.Fields{
+		"name":  res.Name,
 		"query": query,
 		"count": count,
-	}).Info("query complete")
+	}).Info("polling complete")
 	state.LastPolled = pollTime
 	if err := c.streamStateCheckpoint(stateKey, state); err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

PR #2272, opened 2025-01-14
materialize-elasticsearch: handle fields with multiple types, and array fields

williamhbaker:
**Description:**

Adds handling for fields that have multiple types & fields that are arrays so that they can be materialized to Elasticsearch.

Arrays where the array items are known to have a single type (not including `null`) can use an index mapping of that type. For example, a mapping of type `text` will work with values that are `"some value"`, `["some value", "another value"]`, `["yet another value", null]`, etc. The only exception is an array field where the items are singly-typed as an array: Although you can actually send a value like `[["first", "second"], ["third"]]` to a `text` mapping like this and have it work, and the connector protocol does not provide "recursive" inference like this. All that can be known about such a field is that it is an array containing arrays, and so there is no "leaf" type to use for the mapping type.

Support for a few other things - fields with multiple types, arrays with unknown or non single-typed items, and the aforementioned singly-typed arrays where the type is itself `array` - is added here by wrapping the values in an object. An `object` type itself is already able to be materialized to a `flattened` mapping type, which basically stores arbitrary JSON data (as long as it is an object) in the mapping. We will construct a synthetic object to wrap these values in, in the form of:

```json
{
  "json": "whatever the JSON representation of the value is"
}
```

...where the value of the "json" field is obviously the JSON representation of the value, which may itself be an object, and array, or anything else.

Closes https://github.com/estuary/connectors/issues/2271

**Workflow steps:**

N/A since we don't mention anything about fields that may or may not be allowed to be materialized in the docs. This may be an area for future documentation improvement.

**Notes for reviewers:**

- [x] Needs https://github.com/estuary/flow/pull/1827 to be merged for the tests to pass. I ran them locally on `flowctl` built on that branch. It adds the array inference information to builds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2272)


materialize-elasticsearch/type_mapping.go
```diff
@@ -148,13 +147,16 @@ func (constrainter) NewConstraints(p *pf.Projection, deltaUpdates bool) *pm.Resp
 	switch {
 	case p.IsPrimaryKey:
 		constraint.Type = pm.Response_Validated_Constraint_LOCATION_REQUIRED
-		constraint.Reason = "Primary key locations are required"
```
williamhbaker, line 151: This `NewConstraints` function is now exactly the same as the one used in `materialize-sql`: https://github.com/estuary/connectors/blob/7f596b091b0614b01a4bf786b8e43abd2e4ddc69/materialize-sql/type_mapping.go#L414-L453

Alex-Bair (approved):
LGTM

PR #2274, opened 2025-01-15
source-facebook-marketing: remove wish_bid from `ads_insights.json`

Alex-Bair:
**Description:**

Multiple production tasks have encountered errors where any async job requesting the `wish_bid` field have started consistently failing on Facebook's side. Async jobs that do not request `wish_bid` still succeed.

Removing `wish_bid` from the `ads_insights.json` schema file also removes it from the list of fields requested from async jobs, and tasks that were previously failing should now work fine. It's unclear to me what changed that causes async jobs requesting the `wish_bid` field to fail, and maybe it can be added back later if there's motivation to investigate further & determine root cause.

Discover snapshot changes are expected since `wish_bid` has been removed from the `ads_insights` stream and its children.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed tasks that were previously failing due to failed async jobs during the `ads_insights` stream (and children) now completed & work fine.

Existing collections that contain documents with `wish_bid` values do not have `additionalProperties` explicitly set to `false`, so the now extra `wish_bid` field should not cause validation failures when reading from those collections.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2274)

### DIFF
--- a/source-facebook-marketing/source_facebook_marketing/schemas/ads_insights.json
+++ b/source-facebook-marketing/source_facebook_marketing/schemas/ads_insights.json
@@ -309,9 +309,6 @@
     },
     "website_purchase_roas": {
       "$ref": "ads_action_stats.json"
-    },
-    "wish_bid": {
-      "type": ["null", "number"]
     }
   },
   "type": ["object"],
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #2276, opened 2025-01-15
materialize-bigquery: use the storage read API if available for load queries

williamhbaker:
**Description:**

The storage read API is dramatically faster at retrieving large query results than the standard job read or table read API. On a test with 10 million rows of synthetic data returned by a load query, I have observed a reduction in processing time from ~21 minutes using the job read API down to ~3 minutes using the storage read API with a `materialize-bigquery` task running on our public data plane.

Historically we have not enabled the storage read API due to its extreme negative effect on memory usage in the connector. Based on testing with the more recent version of the BigQuery client we are now using, this memory situation seems to no longer be an issue at all, with query results read by the storage read API using comparable amounts of memory as the slower standard APIs.

This commit will enable the storage read API if possible. If it isn't possible, probably due to missing permissions, the standard APIs will be used automatically by the BigQuery client and we will log a message indicating that the faster API is not being used with a link to our documentation to help with enabling it.

Closes https://github.com/estuary/connectors/issues/968

**Workflow steps:**

Update `materialize-bigquery` docs with the recommendation to add the `BigQuery Read Session User` role to the service account used for the connector for optimal load query performance.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2276)

### DIFF
--- a/materialize-bigquery/bigquery.go
+++ b/materialize-bigquery/bigquery.go
@@ -84,6 +84,14 @@ func (c *config) client(ctx context.Context) (*client, error) {
 	if err != nil {
 		return nil, fmt.Errorf("creating bigquery client: %w", err)
 	}
+	// Use the much faster storage read API for reading query results if the
+	// authorization provides sufficient access, typically via the "BigQuery
+	// Read Session User" role. Result iterators will automatically fall back to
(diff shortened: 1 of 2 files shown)


Alex-Bair (approved):
LGTM

PR #2277, opened 2025-01-15
source-shopify-native: new connector

Alex-Bair:
**Description:**

This is a native capture connector for Shopify leveraging their GraphQL API.

Using the GraphQL API is motivated by Shopify stating that their REST API (specifically the `/products` and `/variants` endpoints in the near-term) will be inaccessible soon. There are multiple differences between the REST and GraphQL schemas for equivalent resources, including:
- snake_case vs. camelCase field names
- renamed fields
- different nesting of subresources

Due to these differences, compatibility with our existing `source-shopify` import was not a goal during development; we don't want our native connector to serve as a transformation layer to try and maintain compatibility with Shopify's REST API schemas.

This initial version of the connector only contains the `product` stream since that's the first one to supposedly stop working on 01FEB2025.

Important design decisions include:
- Retrieving data via bulk query operations. This avoids the rate/complexity limits enforced on non-bulk queries & allows us to fetch significantly more fields and results than using non-bulk queries.
- Introspection was not used to programmatically build queries since it would have been possible to get in an infinite loop due to circular GraphQL connections.
- Abstracting bulk query execution into a `graphql` module within the connector. `BulkJobManager` handles executing bulk queries, and stream specific queries and result processing are located within individual files. This should make it a little easier to add more streams later; we'll just need to add the query and result processing logic for the specific stream, and we can use `products.py` as a template.

I've tested out what edge cases I can to ensure `BulkJobManager` works well, but I've left some info level logging to make monitoring easier as the connector begins being used. The logging can be tamped down to debug later as we tune & improve `BulkJobManager`.

**Workflow steps:**

Documentation should be created for `source-shopify-native`. It should specify what permissions are needed for the connector to work, especially since it needs the `read_publications` permission that our imported `source-shopify` connector does not need.

**Notes for reviewers:**

Tested on a local stack with a few Shopify accounts. Confirmed bulk query jobs are submitted and `products` results are processed appropriately.


source-shopify-native/source_shopify_native/api.py
```diff
@@ -0,0 +1,58 @@
...
+        return
+
+    last_seen_dt = log_cursor
+
+    lines = http.request_lines(log, url)
+    async for record in products.process_result(log, lines):
+        product = TimestampedResource.model_validate(record)
+
+        if product.updatedAt > last_seen_dt:
+            # If we've yielded any documents, checkpoint them.
```
williamhbaker, line 46: Do we need to be checkpointing after every document? I would think that the final document after emitting all the records would be sufficient. My assumption is that waiting for the bulk job to complete will be slow, but reading through all the results once it is done will be relative fast.

source-shopify-native/source_shopify_native/graphql/products.py
```diff
@@ -0,0 +1,300 @@
+from datetime import datetime
+from logging import Logger
+import json
+from typing import Any, AsyncGenerator
+
+from .common import dt_to_str, round_to_latest_midnight
+
+def build_query(start: datetime, end: datetime) -> str:
+    lower_bound = dt_to_str(start)
+    upper_bound = dt_to_str(round_to_latest_midnight(end))
```
williamhbaker, line 10: This seems to be at odds with logic in `fetch_products` where the end time is calculated with a 1 minute horizon from the present time, since `round_to_latest_midnight` will move the end time forward to midnight of the next day, which may be in the future.

source-shopify-native/source_shopify_native/resources.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+AUTHORIZATION_HEADER = "X-Shopify-Access-Token"
+
+
+async def validate_credentials(
+        log: Logger, http: HTTPMixin, config: EndpointConfig
+):
+    """
+    Validates that the provided access token is a valid Front access token.
```
williamhbaker, line 30: Docstring needs updated. TBH I'd just remove it entirely since the function is self-explanatory.

williamhbaker (commented):
A couple of comments

source-shopify-native/source_shopify_native/api.py
Alex-Bair: With just the single stream right now, we don't need to be checkpointing after every document. My thought was that in the future when there's more than a single stream, one stream could be in the middle of reading results when another stream fails, and checkpointing while reading results could help that first stream make a little incremental progress. But thinking about it more, that's a scenario we can address when/if it pops up. I'll remove the mid-read checkpoints.

williamhbaker: There's actually handling in the CDK for the scenario where one stream fails, but wanting to make as much progress as possible on the other streams. I can't remember exactly where it is off hand, but the gist of it is that all bindings will be allowed to at least return from their "fetch" calls even if one of them fails, and their checkpoints will be persisted before failing the connector.

source-shopify-native/source_shopify_native/graphql/products.py
Alex-Bair: That's right, `rount_to_latest_midnight` will almost always move the end time used in the query to midnight of the next day. I did that because Shopify's API filters results based on midnight of the specified datetime. For example, for a start of `2025-01-10T12:00:00Z` and an end of `2025-01-16T12:00:00Z`, Shopify returns results updated between `2025-01-10T00:00:00Z` and `2025-01-16T00:00:00Z`. Meaning, if the end date used in the query isn't moved into the future at/past midnight of the next day, Shopify's API won't return results updated within the current day.

source-shopify-native/source_shopify_native/api.py
Alex-Bair: Ahh, gotcha, thank you. I see now that's handled by [these lines](https://github.com/estuary/connectors/blob/f9ad0c49922a42b714edbc5c9c35717350a1ffe0/estuary-cdk/estuary_cdk/__init__.py#L126-L162) using `asyncio.TaskGroup` and `ExceptionGroup`. Nice!

source-shopify-native/source_shopify_native/graphql/products.py
Alex-Bair: And regarding the 1 minute horizon, I haven't observed anything to suggest that there are distributed clock/data availability delays. I was just a little skeptical that a bulk query job would return results updated in real-time, so I added `WINDOW_END_LAG` as a hedge if those delays do exist/occur later. I can remove it though if that's something we should tackle later if it's an actual issue.

williamhbaker: Am I understanding correctly that the horizon won't really do anything anyway due to the rounding of end times to the next midnight time? The only other thing we could do is client-side filtering of results that are too recent, but I also don't think that is a great idea since we'd end up re-running jobs over data again and again.

Overall I'd say remove it, and if we see problems with eventual consistency we'd have to add a second trailing stream that maybe lags an entire day behind, similar to what we have done elsewhere.

Alex-Bair: Yeah, the horizon won't do anything to affect what results are returned by the API. The purpose of the horizon was more around cursor management. If there are new results since the previous `log_cursor`, then the horizon won't do anything. But if there aren't any new results since the last `log_cursor`, the horizon limits the max cursor update to the present time minus the horizon.

I'll remove the horizon, and we can tackle those types of issues later (if they ever happen🤞).

PR #2281, opened 2025-01-16
source-intercom-native: add `tickets` stream

Alex-Bair:
**Description:**

The `tickets` stream is extremely similar to the `conversations` stream. Their `/{resource}/search` [endpoints](https://developers.intercom.com/docs/references/rest-api/api.intercom.io/tickets/searchtickets) are queried and behave in the same way, so a good chunk of this PR was copy + paste of `conversations`, then renaming & refactoring.

Snapshot changes are expected due to the addition of the `tickets` stream.

Our production Intercom OAuth app will need updated to request the "Read and List tickets" permission for the connector to successfully make calls to `/tickets/search`. After the app is updated, any existing tasks will need to be reauthenticated if they want to use the `tickets` stream.

**Workflow steps:**

Documentation should be updated to reflect the added stream. We may also want to list what scopes are required if users authenticate with access tokens.

**Notes for reviewers:**

Tested on a local stack. Confirmed that the `tickets` stream captures data & behaves like the `conversations` stream.


source-intercom-native/source_intercom_native/api.py
```diff
@@ -232,6 +248,56 @@ async def fetch_contacts(
...
+        if _is_page_number_to_log(page_num, total_pages):
+            log.info(f"Processing page {page_num} of {total_pages}.", {
+                'start': _s_to_dt(start),
+            })
+
+        for ticket in response.tickets:
+            if ticket.updated_at > last_seen_ts:
+                # Checkpoint any yielded documents if the next result was updated later than
+                # the prior results and it's not the first result we've seen.
+                if last_seen_ts != start:
```
williamhbaker, line 286: Not a big deal since this appears to be a copy/paste from `fetch_conversations`, but as I've commented on a separate PR we don't really need to emit checkpoints so often, and could wait until all the results are processed before doing so. It wouldn't be a terrible ideal to update both this and `fetch_conversations` in the way they emit checkpoints.

williamhbaker (approved):
LGTM

source-intercom-native/source_intercom_native/api.py
Alex-Bair: Since both `fetch_tickets` and `fetch_conversations` don't have a max date window size & could be trying to process months/years of results when backfilling, what do you think about emitting checkpoints after yielding N documents like in [source-front](https://github.com/estuary/connectors/blob/ff8c7a22e1c1233480c2831078bfd16846a5eba3/source-front/source_front/api.py#L171-L176)? Checkpoints wouldn't be emitted as often & progress could still be made during backfills without needing to process all results in a single shot.

williamhbaker: That seems like it would work well!

Alex-Bair:
I'm holding off on merging until our production Intercom OAuth app has been updated to request the permissions needed for the `tickets` stream.

Alex-Bair:
The "Read and List Tickets" permission has been requested for our OAuth app, so I'm going to merge this PR now. If the permission isn't granted until after Intercom's review is complete, we can disable the `tickets` stream for captures until then.

PR #2282, opened 2025-01-17
source-postgres: Add 'Maximum Backfill XID' property

willdonnelly:
**Description:**

This is the obvious counterpart to the 'Minimum Backfill XID' option, allowing users to specify an upper-bound for the XID range which ought to be backfilled.

The specific use-case this addresses is that in databases which have received enough writes that the XID has wrapped around at least once, there may be old rows from prior XID epochs with an XID greater than a particular min-XID, and we can't easily distinguish between these and legitimate new rows which ought to be backfilled.

Specifying an upper-bound doesn't entirely fix the issue as we will still capture rows from the specified XID range in prior wraparound epochs, but it reduces the worst-case impact from "a range of 2^31 rows starting at min_backfill_xid" to "a range of N rows from min_backfill_xid to max_backfill_xid" where N will generally be much smaller than 2 billion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2282)


source-postgres/backfill.go
```diff
@@ -203,6 +203,9 @@ func (db *postgresDatabase) keylessScanQuery(_ *sqlcapture.DiscoveryInfo, schema
 	if db.config.Advanced.MinimumBackfillXID != "" {
 		fmt.Fprintf(query, ` AND (((xmin::text::bigint - %s::bigint)<<32)>>32) > 0 AND xmin::text::bigint >= 3`, db.config.Advanced.MinimumBackfillXID)
 	}
+	if db.config.Advanced.MaximumBackfillXID != "" {
+		fmt.Fprintf(query, ` AND (((%s::bigint - xmin::text::bigint)<<32)>>32) > 0 AND xmin::text::bigint >= 3`, db.config.Advanced.MaximumBackfillXID)
```
willdonnelly, line 207: There were three obvious ways to negate the "minimum XID" filter expression when adding the maximum, and I arbitrarily went with "swap the operands of the subtraction". I don't see any possible way it could matter, but this could instead have been done by changing `> 0` to `< 0` or by slapping a `NOT` on it.

Let me know if you think one of those alternatives would be clearer.

williamhbaker (approved):
LGTM

PR #2284, opened 2025-01-17
source-braintree-native: fix missing `disputes` documents

Alex-Bair:
**Description:**

We've received reports of missing `disputes` documents. The provided missing disputes both had a `received_date` that was one day before their `created_at`, meaning we can't assume that `received_date` is a close approximation of `created_at` when searching Braintree. Examples of missing documents are below.
```json
{"id":"dispute_1","received_date":"2025-01-10","created_at":"2025-01-11T00:50:12Z"}
{"id":"dispute_2","received_date":"2025-01-16","created_at":"2025-01-17T00:09:34Z"}
```
I think we missed these disputes when the following conditions were true:
1. The capture was caught up to the present and replicating incrementally.
2. It was early-ish in the next day. 
3. The window size was small (ex: 1 hour).

For example, if the capture is using a 1 hour date window, it's possible that a start/end of `2025-01-10T23:45:00Z`/`2025-01-11T00:45:00Z` are used in a sweep. We wouldn't capture `dispute_1` above because it isn't created yet (based on it's `created_at` of `2025-01-11T00:50:12Z`). On the next sweep, we use a start of `2025-01-11T00:45:00Z`. Since we're searching Braintree based on the `received_date`, we _still_ wouldn't capture `dispute_1` since its `received_date` is the previous day `2025-01-10`.

Unfortunately, since Braintree doesn't expose `created_at` for searching disputes, we're stuck using `received_date`. To avoid missing these type of results, `disputes` now searches at least a two day wide date window for new results. Assuming `received_date` and `created_at` are never more than one day apart, this should ensure the connector doesn't miss data for the same reason again.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack with a Braintree account with a non-trivial amount of data. Confirmed that:
- The previously missing documents were captured
- When caught up to the present, `disputes` searches the previous and current day for new results.
- The 10k search limit was not exceeded by `disputes` due to increasing the search date window. I observed <1.5k results in a given two day window. If a Braintree user that has >10k disputes created in a two day window starts using the connector, then we'll need to re-evaluate this strategy.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2284)

### DIFF
--- a/source-braintree-native/source_braintree_native/api.py
+++ b/source-braintree-native/source_braintree_native/api.py
@@ -382,6 +382,10 @@ async def fetch_subscriptions(
         yield end
 
 
+def _are_same_day(start: datetime, end: datetime) -> bool:
+    return start.date() == end.date()
+
+
 async def fetch_disputes(
         braintree_gateway: BraintreeGateway,
         window_size: int,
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2285, opened 2025-01-20
source-postgres: Make stream-to-fence watchdog 12 hours

willdonnelly:
**Description:**

It's causing problem for at least one production task and the failure mode this guards against is much less important in the immediate term than allowing production to catch back up.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2285)


jgraettinger (approved):
LGTM

PR #2287, opened 2025-01-22
materialize-mongodb: use multiple concurrent load workers

williamhbaker:
**Description:**

Use multiple concurrent load workers to maximize throughput when there are a lot of load requests to evaluate. This can be a limiting factor particularly when round-trip latency is high.

The original code was structured to allow for easily adding concurrency to these requests, but I just hadn't seen a need for it until now, when we have an example of a task that seems to be getting limited at this point.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2287)

### DIFF
--- a/materialize-mongodb/transactor.go
+++ b/materialize-mongodb/transactor.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"sync"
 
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM!

PR #2291, opened 2025-01-23
source-hubspot-native: use v3 of the owners API

williamhbaker:
**Description:**

The v2 owners API has been deprecated and actually doesn't even work on newer accounts. We need to use the v3 owners API instead of the v2 owners API to prevent issues with newer HubSpot accounts.

The API output is slightly different between the v3 and the v2 API, but since this is a snapshot stream anyway and very low volume for our current use cases, I'm not anticipating that being a huge issue.

We could build in some kind of backward compatibility to make the v2 and v3 output more the same, or have captures that are able to use the v2 API or something like that, but I don't think that is worth the long-term maintenance cost we would take on.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api.py
+++ b/source-hubspot-native/source_hubspot_native/api.py
@@ -79,11 +79,30 @@ async def fetch_deal_pipelines(
 
 
 async def fetch_owners(
-    log: Logger, http: HTTPSession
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM! Thanks!

PR #2297, opened 2025-01-27
materialize-snowflake: check pipe reports without delay unless 0 results

mdibaiee:
**Description:**

GitHub diff is a mess because of indentation, so I've commented on actual changes

- Rework the pipe report checking logic. We used to set a number of iterations per pipe to approximate taking 5 minutes, but in cases where users have > 100 pipes we would end up always falling back to `COPY_HISTORY`, and failing because we had tried to get all results from `COPY_HISTORY` on the first iteration (whereas `COPY_HISTORY` is supposed to be a last resort option). It was this way due to some vestigial reasons that no longer apply, we used to paginate through results of reports, etc., which we don't do anymore and we are generally more confident that we will get results from the API in less than a minute.
- The new logic has a time limit across all pipes and instead of iteration numbers, we actually check the time spent so far querying insertion reports. I also reduced the delay between iterations of the loop, since with 3 seconds a task with 200 pipes would take a minimum of 10 minutes to acknowledge each transaction.
- I'm a bit worried about hitting Snowflake ratelimits with 500ms delay between report requests, I have to test this with a large number of pipes and data somehow, although it would a bit egregious if 2 requests per second was too much for Snowflake

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2297)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -717,113 +717,105 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
 	// Keep asking for a report on the files that have been submitted for processing
 	// until they have all been successful, or an error has been thrown
 
-	// If we see no results from the REST API for `maxTries` iterations, then we
-	// fallback to asking the `COPY_HISTORY` table. We allow up to 5 minutes for results to show up in the REST API.
-	if len(pipes) > 0 {
-		// 5 minutes divided by three seconds for each pipe
-		var maxTries = (5 * 60) / (3 * len(pipes))
-		var retryDelay = 3 * time.Second
-		for tries := 0; tries < maxTries; tries++ {
-			for pipeName, pipe := range pipes {
-				// if the pipe has no files to begin with, just skip it
-				// we might have processed all files of this pipe previously
-				if len(pipe.files) == 0 {
-					continue
-				}
+	// If we see no results from the REST API for 5 minutes, then we fallback to asking the `COPY_HISTORY` table
+	var retryDelay = 500 * time.Millisecond
+	var maxTryTime = time.Now().Add(5 * time.Minute)
 
-				// We first try to check the status of pipes using the REST API's insertReport
-				// The REST API does not wake up the warehouse, hence our preference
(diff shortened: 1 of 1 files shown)


materialize-snowflake/snowflake.go
```diff
@@ -717,113 +717,105 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
...
+	// If we see no results from the REST API for 5 minutes, then we fallback to asking the `COPY_HISTORY` table
+	var retryDelay = 500 * time.Millisecond
+	var maxTryTime = time.Now().Add(5 * time.Minute)
+
+	for len(pipes) > 0 {
+		for pipeName, pipe := range pipes {
+			// if the pipe has no files to begin with, just skip it
+			// we might have processed all files of this pipe previously
+			if len(pipe.files) == 0 {
+				delete(pipes, pipeName)
```
mdibaiee, line 729: Once a pipe has no more files to process, delete it from the map so the outer loop stops eventually

materialize-snowflake/snowflake.go
```diff
@@ -717,113 +717,105 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
...
+			// load those files again, our request will not show up in reports.
+			// Moreover, insertReport only retains events for 10 minutes.
+			// One way to find out whether files were successfully loaded is to check
+			// the COPY_HISTORY table to make sure they are there. The COPY_HISTORY is much more
+			// reliable. If they are not there, then something is wrong.
+			if len(report.Files) == 0 {
+				if time.Now().Before(maxTryTime) {
+					time.Sleep(retryDelay)
+					continue
+				}
```
mdibaiee, line 751: Check for time elapsed to decide whether it's time to check `COPY_HISTORY` table in case of no results from API

williamhbaker (approved):
LGTM

mdibaiee:
I ran some tests on >200 pipes [here](https://github.com/estuary/connectors/actions/runs/13007819648/job/36278584635?pr=2305) and there are no rate-limits, so I think the 500ms delay is fine 👍🏽

PR #2298, opened 2025-01-27
source-sqlserver: Always lowercase sys.* discovery

willdonnelly:
**Description:**

So apparently only the `INFORMATION_SCHEMA` tables are officially named in uppercase in SQL Server, and the `sys.foobar` information tables are officially lowercase. Luckily we consult either one or the other data source in specific queries so it's easy enough to just go back to unconditionally using the old lowercased queries for secondary index and computed column listings.

**Notes for reviewers:**

This hasn't been tested against a database running a case-sensitive (or case-insensitive but Turkish) collation, but as it's just stripping out some feature-flagged logic it shouldn't change anything by default so I think it's safe to just YOLO it assuming CI passes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2298)


williamhbaker (approved):
LGTM

PR #2304, opened 2025-01-27
source-postgres: Always stringify float keys

willdonnelly:
**Description:**

Flow won't / can't accept a `type: [string, number], format: number` as a collection key. We have to support strings as serialized float values because they need to be able to hold special values such as `"NaN"` and `"Infinity"`, so in order to make float keys work right we have to always stringify them.

**Workflow steps:**

Capturing from tables with `REAL` and `DOUBLE PRECISION` primary keys should now Just Work, where previously it would inevitably fail. Any captures which were previously working should be unaffected, because this only changes the behavior when the floating-point value occurs in the primary key of the table.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2304)

### DIFF
--- a/source-mysql/discovery.go
+++ b/source-mysql/discovery.go
@@ -187,7 +187,7 @@ func columnsNonNullable(columnsInfo map[string]sqlcapture.ColumnInfo, columnName
 	return true
 }
 
-func (db *mysqlDatabase) TranslateDBToJSONType(column sqlcapture.ColumnInfo) (*jsonschema.Schema, error) {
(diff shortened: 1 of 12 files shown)


jgraettinger (approved):
LGTM

I presume a DECIMAL PRIMARY KEY will already Just Work too, because it's already `type: string, format: number` ?

willdonnelly:
> I presume a DECIMAL PRIMARY KEY will already Just Work too, because it's already `type: string, format: number` ?

Yep, decimal/numeric values should always be stringified.

PR #2306, opened 2025-01-28
materialize-sql: change column migration misleading log message

mdibaiee:
**Description:**

- The previous log line made it seem like it was actually doing the migration, but in reality it is only rendering queries for migration. In cases of multiple column migration the previous message made it seem like all but the last column had already been migrated in milliseconds while the last one was taking its time. This threw me off just now while debugging a task.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2306)


Alex-Bair (approved):
LGTM

PR #2308, opened 2025-01-29
materialize-s3-iceberg: fix checkpoint accounting for disabled bindings

williamhbaker:
**Description:**

This fixes a bug where the previous/current tracked checkpoint hash for disabled bindings would be erroneously advanced for disabled bindings that have a pending staged commit that will be applied if their binding is ever enabled again.

It also includes a temporary disabling of the sanity check that ensures checkpoint hashes advanced as we expect, since some tasks are in a state where this is not true.

**Workflow steps:**

**Notes for reviewers:**

As a point of clarification, these checkpoint hashes are not actual recovered checkpoints as you might see from a "remote store is authoritative" type of materialization, and are only really used for giving us to best chance of not re-applying a prior commit to the same table again, which itself would not even be a catastrophic problem since this is an at-least-once materialization, although with the checkpoint hash tracking it is very unlikely for commits to be re-applied.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2308)

### DIFF
--- a/materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
+++ b/materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
@@ -355,14 +355,16 @@ def append_files(
     if cp == next_checkpoint:
         print(f"checkpoint is already '{next_checkpoint}'")
         return  # already appended these files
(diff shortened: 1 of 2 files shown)


jgraettinger (approved):
LGTM

PR #2309, opened 2025-01-29
source-zendesk-support-native: new connector

Alex-Bair:
**Description:**

source-zendesk-support-native: initial connector

This is an initial version of a native Zendesk Support capture connector. It currently only has a subset of the streams that exist in the imported `source-zendesk-support` connector, and the remaining streams will be added at a later date to achieve parity with the imported connector.

This initial version ended up having more streams than I had originally anticipated. When planning out the right abstractions for the most important streams, I ended up making a lot of the simpler streams in the process.

To support string cursors, our CDK has also been updated to allow `LogCursor`s of type `tuple[str]`. Since the applicable Zendesk cursors are a base 64 encoded string of the resource timestamp & id, I was able to keep the strictly increasing nature of the `tuple[str]` log cursors.

**Workflow steps:**

Documentation will need to be created for `source-zendesk-support-native`.

**Notes for reviewers:**

Tested on a local stack with a few different Zendesk accounts. Confirmed:
- OAuth worked.
- All streams captured data & didn't crash.
- For one Zendesk account, the document counts for the native connector were the same/more than the imported connector.


Alex-Bair:
I did some additional testing & noticed it takes a while for `satisfaction_ratings` to backfill. Results for this endpoint are returned in descending order, and we can't change that with query params. We can use `start_time` and `end_time` params to bound the results, so I'm going to update the stream to use date windows.

source-zendesk-support-native/tests/snapshots/snapshots__spec__capture.stdout.json
```diff
@@ -0,0 +1,163 @@
...
+      "additionalProperties": false,
+      "description": "ResourceConfig is a common resource configuration shape.",
+      "properties": {
+        "name": {
+          "description": "Name of this resource",
+          "title": "Name",
+          "type": "string"
+        },
+        "interval": {
+          "default": "PT0S",
```
williamhbaker, line 133: Sanity checking - any reason to think we should use a less aggressive default? API limits or anything like that?

source-zendesk-support-native/source_zendesk_support_native/api.py
```diff
@@ -0,0 +1,683 @@
...
+
+    while True:
+        response = response_model.model_validate_json(
+            await http.request(log, url, params=params)
+        )
+
+        for resource in response.resources:
+            resource_dt = _str_to_dt(getattr(resource, cursor_field))
+            if resource_dt > last_seen_dt:
+                if count > MIN_CHECKPOINT_COUNT:
```
williamhbaker, line 170: Some food for thought: What happens if there are long runs of records with the exact same cursor value? For example, thousands of records updated at the same time. Typically there is a potential to miss data if we checkpoint in the the middle of reading those sequences, and then have to resume later without having fully read them and effectively skip ahead. 

This is probably a problem that can be generalized and solved somehow maybe down the road. Perhaps a simple mitigation here would be to not emit a checkpoint unless the cursor value from the current record is different than the cursor value from the prior record. Or, maybe this isn't even worth worrying about now or is impossible for other reasons.

source-zendesk-support-native/source_zendesk_support_native/api.py
```diff
@@ -0,0 +1,683 @@
...
+async def _make_incremental_cursor_export_request(
+    http: HTTPSession,
+    url: str,
+    params: dict[str, str | int],
+    response_model: type[IncrementalCursorExportResponse],
+    log: Logger,
+) -> IncrementalCursorExportResponse:
+    # Instead of using Pydantic's model_validate_json that uses json.loads internally,
+    # use json.JSONDecoder().raw_decode to reduce memory overhead when processing the response.
+    raw_response_bytes = await http.request(log, url, params=params)
```
williamhbaker, line 324: What are the sensitivities here? Is the raw response bytes size unbounded? Avoiding some of Pydantic's overhead is a good idea, but if there is no practical limit on how large the response size can be we'll inevitably hit memory issues.

I've got some WIP that is almost finished on a kind of streaming decoder strategy which only pulls bytes off the wire as they can be parsed and yield, so that might be something we can look into adding here later.

source-zendesk-support-native/source_zendesk_support_native/api.py
```diff
@@ -0,0 +1,683 @@
...
+    ClientSideIncrementalCursorPaginatedResponse,
+    IncrementalCursorPaginatedResponse,
+    SatisfactionRatingsResponse,
+    AuditLog,
+    AuditLogsResponse,
+    INCREMENTAL_CURSOR_EXPORT_TYPES,
+)
+
+CURSOR_PAGINATION_PAGE_SIZE = 100
+MIN_CHECKPOINT_COUNT = 1500
```
williamhbaker, line 27: Looks like this is used in a couple of places, where we are looping over pages of results. First off, I'll recognize that I am probably starting to sound like a lunatic with the different suggestions I have regarding when to checkpoint, so take this suggestion as you wish. But perhaps we could employ a strategy of just checkpointing after we read each "page" (see my other comment below regarding sequences of identical timestamps though), rather than tracking a separate count and variable here?

I think I'll summarize my thoughts on how often to checkpoint:
- Checkpoints really _can_ be an arbitrarily large size; the system is designed to handle this. Practically speaking though we usually don't want to risk losing a lot of work the connector has done when it gets nuked for whatever reason and has to restart. There are also some implications on materialization performance if there is a huge amount of data between checkpoints.
- Going to the opposite extreme, it can sometimes work to checkpoint after every single document, and it may even make sense to do that sometimes. For example, when doing something like continuously reading a change stream, each record may have an offset associated with it, and checkpoint after each document is reasonable since we are reading them as they come, and the sooner we checkpoint a document the sooner it can be processed downstream, which reduces end to end latency. Another thing to be aware of is that if that capture emits huge volumes of checkpoints, the runtime will actually internally combine them when committing data to the target collections, so that's an additional optimization.
- But there's a cost to checkpointing: Depending on the document size, serializing the checkpoint may be a significant amount of CPU work for the connector if checkpoints are emitted at an extreme rate, like when processing lots of records from a backfill.

In this particular case I think the code would be just a little simpler without the MIN_CHECKPOINT_COUNT and related accounting. Page sizes of 100 are kind of small, but they provide a nice logical breaking point that the code has to handle anyway so we might as well take advantage of that.

source-zendesk-support-native/tests/snapshots/snapshots__spec__capture.stdout.json
Alex-Bair: Are you asking about the CDK defined default [here](https://github.com/estuary/connectors/blob/c11b95bc209b088dd4ca94eebe963813a338a9bc/estuary-cdk/estuary_cdk/capture/common.py#L98-L100)? It probably wouldn't be a bad idea to increase it. Frequently whenever I've left it as `PT0S` for a stream, the connector gets rate limited frequently & inundates the logs.

Alex-Bair: Yep, I have the `interval` for all resources' `initial_config`s to be 5 minutes, like [here](https://github.com/estuary/connectors/blob/7a01923150781513a5195ec8ff28d36d03eb148b/source-zendesk-support-native/source_zendesk_support_native/resources.py#L361-L363), and that's reflected in the discover snapshot.

source-zendesk-support-native/source_zendesk_support_native/api.py
Alex-Bair: Yes, I've learned that a lot of custom fields & data can be added to Zendesk tickets, increasing the response size. So it theoretically seems unbounded, or at least the limit is large enough that it can OOM the connector.

I'll try out using the streaming decoder strategy you merged before I merge this PR.

PR #2310, opened 2025-01-29
source-google-sheets-native: stream spreadsheet rows

### LINKED ISSUE #2312: source-google-sheets: fix ooms

williamhbaker:
**Description:**

The structure of the response for the data of a google sheet lends itself to incrementally parsing each row, rather than loading it all into memory and parsing it at once. Sheets generally aren't that large, but if they are even modestly large the amount of overhead involved in entire response has the potential to cause OOMs. Parsing the response row-by-row is much more memory efficient.

Closes #2312

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2310)


estuary-cdk/estuary_cdk/http.py
```diff
@@ -21,6 +22,30 @@
 
 DEFAULT_AUTHORIZATION_HEADER = "Authorization"
 
+StreamedObject = TypeVar("StreamedObject", bound=BaseModel)
+
+class _AsyncStreamWrapper:
```
williamhbaker, line 27: This feels like it was a lot more difficult than it needed to be. `ijson.items` needs something "file-like" with a `read` method. `aiohttp` doesn't give us anything quite like that (not even its raw `content` seemed to work). So this adapter is the only way I could get things working with the byte stream coming from `aiohttp`.

source-google-sheets-native/source_google_sheets_native/api.py
```diff
@@ -105,7 +102,7 @@ def convert_row(
             else:
                 d[headers[ind]] = nv
 
-    return Row(_meta=Row.Meta(op="u", row_id=id), **d)
```
williamhbaker, line 108: Setting the `op` and `row_id` here never did anything since they are set by the CDK.

estuary-cdk/estuary_cdk/http.py
```diff
@@ -99,6 +124,32 @@ async def request_lines(
             yield buffer
 
         return
+    
+    async def request_object_stream(
```
Alex-Bair, line 128: To make sure I understand the intended use of this method, it's only to parse JSON objects located at a single prefix in the response, right? For example, in this response body, this method could be used to iterate over the objects under `tickets`, and a separate request will need to be made to get any of the top level fields (`after_cursor`, etc.)?
```json
{
    "after_cursor": "cursor_to_next_page",
    "total_count": 2,
    "tickets": [
        {"id": 1},
        {"id": 2}
    ]
  }
```

Alex-Bair (approved):
LGTM!

estuary-cdk/estuary_cdk/http.py
williamhbaker: Yep that's right. `ijson` does have a lower-level interface where you could do it all in a single request, something like this:

```python
    parser = ijson.parse(file)

    after_cursor = None
    total_count = None

    for prefix, event, value in parser:
        if prefix == "after_cursor" and event == "string":
            after_cursor = value
        elif prefix == "total_count" and event == "number":
            total_count = value
        elif prefix == "tickets.item": 
            yield value
```

I'm not totally sure what the best generalization of this pattern is. I felt like it was probably worth having this higher level `request_object_stream` capability, although it won't work for everything.

Alex-Bair: Yep, I agree this will be useful even if it won't work for everything. Thanks for putting this together; it's definitely going to be useful when we're processing a huge response.

PR #2318, opened 2025-01-31
estuary-cdk: expand capabilities for incremental json processing

williamhbaker:
**Description:**

A prior implementation of incremental JSON processing was able to yield records from a JSON stream by prefix, discarding everything else from the stream. That suits cases where the entire dataset is represented in a single large response, but there are cases where the dataset is represented by multiple potentially very large responses, and each of those responses contains some extra information we need to proceed, such as a paging token. Parsing and processing those responses while keeping memory usage efficient within the connector can be a challenge.

This adds the ability to read a JSON stream incrementally and yield records by prefix, while also returning anything that's left at the end. Effectively this "remainder" is built up while processing the document as an object in-memory. As long as this remainder is relatively small, memory usage should be negligible.

I had assumed CPU performance for this mechanism would be worse than the standard processing of a response in its entirety through a Pydantic model, but it turns out it's actually about 5-10% faster. I believe this is because the JSON parser used by ijson is faster than the one used by Pydantic. We're still building Pydantic models from the python objects yielded by the ijson parser, but doing so using `model_validate` instead of `model_validate_json`.

I suspect CPU performance could be improved even more by not using Pydantic models at all and instead using msgspec structs, but this does not seem necessary right now.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2318)

### DIFF
--- a/estuary-cdk/estuary_cdk/http.py
+++ b/estuary-cdk/estuary_cdk/http.py
@@ -1,8 +1,8 @@
 from dataclasses import dataclass
 from logging import Logger
-import ijson
+from estuary_cdk.incremental_json_processor import Remainder
 from pydantic import BaseModel
-from typing import AsyncGenerator, Any, AsyncIterator, TypeVar
+from typing import AsyncGenerator, Any, TypeVar
 import abc
 import aiohttp
 import asyncio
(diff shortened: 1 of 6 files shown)


estuary-cdk/estuary_cdk/http.py
```diff
@@ -125,31 +103,20 @@ async def request_lines(
 
         return
     
-    async def request_object_stream(
+    async def request_stream(
```
williamhbaker, line 106: Having this functionality tied to `http` felt a little cumbersome, and I guess you could end up with a byte stream in some other way, so I separated the JSON processing from the byte stream requesting.

source-zendesk-support-native/source_zendesk_support_native/api.py
```diff
@@ -313,23 +313,6 @@ async def backfill_satisfaction_ratings(
...
-async def _fetch_top_level_fields(
-    http: HTTPSession,
-    url: str,
-    params: dict[str, str | int],
-    response_model: type[IncrementalCursorExportResponse],
-    log: Logger,
-) -> tuple[str | None, bool]:
-    # Instead of using Pydantic's model_validate_json that uses json.loads internally,
-    # use json.JSONDecoder().raw_decode to reduce memory overhead when processing the response.
-    raw_response_bytes = await http.request(log, url, params=params)
```
williamhbaker, line 325: This would previously require buffering the entire response's bytes and plucking out a couple values from it, then making the same request again to incrementally process. Now we will only need to make the request a single time while still being able to incrementally process it.

williamhbaker:
@Alex-Bair I think you have a test for zendesk-native memory usage so I'd be interested to know what that looks like with this change.

Alex-Bair (approved):
LGTM! Really nice job!

PR #2321, opened 2025-02-02
source-mysql: Ignore ROLLBACK query events

willdonnelly:
**Description:**

Generally speaking we shouldn't see BEGIN/COMMIT/ROLLBACK events in the binlog, and we can't actually support a rollback on a table that we're actively processing since we've already emitted the changes. But since we do see them in certain edge cases we ought to ignore them properly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2321)


dyaffe (approved):
Lgtm

PR #2323, opened 2025-02-03
materialize-bigquery: objects_and_arrays_as_strings feature flag

williamhbaker:
**Description:**

Adds a feature flag for materializing objects and arrays as strings.

This is the current behavior, and the feature flag is being added so that all pre-existing tasks can have it be set to `true`. A subsequent change will enable the flag's behavior, where if it is not set (which will be the default going forward), objects and arrays will instead be materialized into JSON columns.

**Workflow steps:**

**Notes for reviewers:**


materialize-bigquery/bigquery.go
```diff
@@ -28,6 +32,12 @@ type config struct {
 	HardDelete       bool                       `json:"hardDelete,omitempty" jsonschema:"title=Hard Delete,description=If this option is enabled items deleted in the source will also be deleted from the destination. By default is disabled and _meta/op in the destination will signify whether rows have been deleted (soft-delete).,default=false" jsonschema_extras:"order=7"`
 	Schedule         boilerplate.ScheduleConfig `json:"syncSchedule,omitempty" jsonschema:"title=Sync Schedule,description=Configure schedule of transactions for the materialization."`
 	DBTJobTrigger    dbt.JobConfig              `json:"dbt_job_trigger,omitempty" jsonschema:"title=dbt Cloud Job Trigger,description=Trigger a dbt job when new data is available"`
+
+	Advanced advancedConfig `json:"advanced,omitempty" jsonschema:"title=Advanced Options,description=Options for advanced users. You should not typically need to modify these." jsonschema_extra:"advanced=true"`
+}
+
+type advancedConfig struct {
+	FeatureFlags string `json:"feature_flags,omitempty" jsonschema:"title=Feature Flags,description=This property is intended for Estuary internal use. You should only modify this field as directed by Estuary support."`
```
mdibaiee, line 40: Why not an advanced property specifically for materializing objects and arrays as strings without an exposed JSONSchema? 🤔

williamhbaker:
I changed the feature flag from `objects_and_arrays_as_strings` and `objects_and_arrays_as_json` based on my better understanding of how these feature flags should be used. I suppose I could/should also include the behavior change it will toggle in this same PR instead of a follow-up, but since I've already got this one ready and then next one (https://github.com/estuary/connectors/pull/2324) is fully drafted I'm going to keep them as they are, except for the change of the flag to `objects_and_arrays_as_json`.

materialize-bigquery/bigquery.go
```diff
@@ -16,7 +17,10 @@ import (
 	"google.golang.org/api/option"
 )
 
-// Config represents the endpoint configuration for BigQuery.
+var featureFlagDefaults = map[string]bool{
+	"objects_and_arrays_as_json": false,
```
willdonnelly, line 21: The name is fairly self-explanatory in this case, but I've been trying to maintain a convention of describing the impact of a flag here in the defaults map.

willdonnelly (approved):
LGTM

mdibaiee (approved):

materialize-bigquery/bigquery.go
williamhbaker: That's a good idea, I added a comment here to explain what the flag does.

PR #2324, opened 2025-02-03
materialize-bigquery: JSON column type for array and objects

williamhbaker:
**Description:**

When the BigQuery materialization connector was initially built, BigQuery did not have a native JSON column type, so arrays and objects were materialized as "stringified" JSON.

There is now a native JSON column type, so we can use that for these kinds of fields.

The main complication here is preserving the behavior of pre-existing materializations, while allowing new ones to use the JSON column type. The `objects_and_arrays_as_json` feature flag was added previously, and that flag will toggle using the old stringified columns instead of the JSON columns.

Most of the diff here is mechanical changes needed to thread through the value of that `objects_and_arrays_as_json` flag to where it needs to be.

The rollout of this change will be:
1. Set all existing bigquery materializations to have the flag `no_objects_and_arrays_as_json`, which was added in https://github.com/estuary/connectors/pull/2323.
2. Shortly after that, this pull request should be merged so that new tasks use the default behavior going forward of using JSON columns.

Closes https://github.com/estuary/connectors/issues/2319
Closes https://github.com/estuary/connectors/issues/2333

**Workflow steps:**

**Notes for reviewers:**

Manually tested that an existing materialization can transition successfully from having no feature flag set to having the `no_objects_and_arrays_as_json` set. Trying to set `objects_and_arrays_as_json` on an existing materialization fails because of the newly-added compatibility check for existing root document columns, and the change to not allow them to be migrated. A brand new task with `objects_and_arrays_as_json` set works fine with JSON columns from the start.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2324)


materialize-bigquery/transactor.go
```diff
@@ -32,77 +35,86 @@ type transactor struct {
 	loggedStorageApiMessage bool
 }
 
-func newTransactor(
```
williamhbaker, line 35: Most of this diff is just from wrapping the `newTransactor` behavior in a closure so that it has access to the previously computed dialect, templates, and flag value.

mdibaiee:
@williamhbaker regarding release process, why do we not set the flag on tasks now, and then release this with `objects_and_arrays_as_json` as the default?

williamhbaker:
> @williamhbaker regarding release process, why do we not set the flag on tasks now, and then release this with `objects_and_arrays_as_json` as the default?

That's a good point, since I did this a little strangely and already added the flag in a prior PR we can do it like you suggest, and not have to have another follow-up.

I'll update the PR and release process description for that!

materialize-bigquery/.snapshots/TestValidateAndApply
```diff
@@ -87,7 +87,7 @@ Big Schema Changed Types Constraints:
 {"Field":"multipleField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
 {"Field":"nullField","Type":4,"TypeString":"FIELD_OPTIONAL","Reason":"Object fields may be materialized"}
 {"Field":"numField","Type":6,"TypeString":"UNSATISFIABLE","Reason":"Field 'numField' is already being materialized as endpoint type 'FLOAT' but endpoint type 'BOOLEAN' is required by its schema '{ type: [boolean] }'"}
-{"Field":"objField","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
+{"Field":"objField","Type":6,"TypeString":"UNSATISFIABLE","Reason":"Field 'objField' is already being materialized as endpoint type 'JSON' but endpoint type 'STRING' is required by its schema '{ type: [string] }'"}
```
mdibaiee, line 90: is this expected?

williamhbaker: Yeah, these snapshots are kind of confusing and I should probably change them. But what this is testing is that a field `objField` that was originally a `type: object` can't have it's type changed to `type: string` now. It actually used to be able to since `type: object` was materialized as a string, so changing it to `type: string` was fine.

mdibaiee (approved):

PR #2329, opened 2025-02-03
source-zendesk-support-native: misc fixes

Alex-Bair:
**Description:**

This PR includes a few miscellaneous fixes for bugs in `source-zendesk-support-native`. These fixes include:
1. Validating the `start_date` is at least the epoch so converting to a timestamp always results in a non-negative number.
2. Avoiding `aiohttp` `TimeoutError`s for ticket child streams by not trying to maintain a long-lived streaming response when fetching parent tickets.
3. Using a smaller page size for `ticket_metrics` to reduce how much processing Zendesk must do for a response & avoid intermediate gateways from sending us 504 timeout responses.
  i. We might want to make this configurable later or surface 5xx responses through our CDK at some point, but that level of effort isn't necessary right now.

**Workflow steps:**

**Notes for reviewers:**

- Entering a start date of before the epoch is not accepted as a valid value & prevents task creation.
- Ticket child streams, like `ticket_comments` and `ticket_audits`, no longer experience `TimeoutErrors` from `aiohttp`.
- `ticket_metrics` requests complete successfully & avoid 504 responses when using a smaller page size.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2329)

### DIFF
--- a/source-zendesk-support-native/source_zendesk_support_native/api.py
+++ b/source-zendesk-support-native/source_zendesk_support_native/api.py
@@ -425,7 +425,7 @@ async def backfill_incremental_cursor_export_resources(
             return
 
 
-async def _fetch_ticket_child_resource(
+async def _fetch_ticket_child_resources(
     http: HTTPSession,
     subdomain: str,
     path: str,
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

Issue #2215, opened 2024-12-17
source-sqlserver: Support for CDC without watermark writes


willdonnelly:
Our MySQL and PostgreSQL CDC connectors both support watermark-less operation now (as the only mode of operation in MySQL and as an advanced option in Postgres), so we'd like to complete our core SQL database trifecta by adding watermark-less CDC for SQL Server as well.

After investigating the problem in some detail, it turns out to be trickier than it sounds for two reasons. Firstly, there is no direct SQL Server equivalent to `pg_current_wal_lsn()` (or using MySQL's `SHOW MASTER STATUS`) which provides an LSN corresponding to the current end of the WAL. And in fact most ways of getting that information at all are pretty awful -- for instance using `fn_dblog()` would "work" but requires sysadmin permissions and also is an undocumented function that's not officially supported.

After mulling over the problem for a while, I've found a solution which I think works:

- Barring certain edge cases around the maximum transactions limit, the CDC agent process will usually get itself caught up to the end of the WAL after each invocation.
- We can observe metadata about CDC agent invocations in the `sys.dm_cdc_log_scan_sessions` database management view. By polling this periodically, we can see when a new agent invocation has finished. It's also possible to detect whether the agent hit the maximum transaction limit, so we can account for that edge case.
- If, after the point where we decide we need to establish a fence, we wait for a new CDC agent run to begin and end, and then query the CDC tables themselves to read out all changes through that point, then we can know we're fully caught up to a position in the WAL which is not-before the point where we decided to establish a fence.

This implies a bit of a restructuring of how the CDC polling process is run. Currently it's still a separate worker thread executing `pollChanges()` periodically, but the easiest way to integrate this logic into the `StreamToFence()` operation is to directly call `pollChanges()` at the appropriate times. And that in turn has some follow-on effects on how we currently schedule the other periodic operations like cleaning up acknowledged change table data and handling source table alterations. But that's just a matter of programming and not a fundamental obstacle.

The only _actual_ issue with this approach is that querying `dm_cdc_log_scan_sessions` requires the `VIEW DATABASE STATE` (or possibly `VIEW DATABASE PERFORMANCE STATE` depending on DB version) permission which preexisting users may not have. So to make this work while not breaking existing captures we'll need a migration strategy.

[...]

Issue #1994, opened 2024-09-30
captures: strategy for account for historical schema decisions of an active capture


willdonnelly:
Currently we serialize Postgres `date` columns as an RFC3339 date-time in the capture output. In cases where people care at all, they would probably prefer for us to preserve them as purely dates rather than adding a fictitious time component.

In one sense this is a trivial change: modify the discovery type mapping code to emit `type: string, format: date` instead of `format: date-time` and the value translation code to serialize these column values as dates instead of RFC3339 timestamps. It'd take about 10 minutes and most of that is double-checking how exactly we receive these values from the Postgres client library.

Unfortunately, there's an unknown number of preexisting captures which currently capture dates as timestamps, and this would be a breaking change to all of them. Thus our options are:

1. Do nothing and stick with `format: date-time` for the moment.
2. Try to establish how many capture bindings actually contain `date` columns and hope really hard that they're few/small enough that we'd be willing to just re-backfill all of them end to end.
3. Implement some sort of advanced config toggle for "do the old thing" and then do a whole migration where we set that flag on all preexisting captures (or all of them which have `date` columns) and then implement the new behavior so that future captures will always have them be `format: date`.
4. Generalize the problem and try to figure out if there's any way we could make breaking datatype changes more painless in the future, possibly by building some sort of generic compatibility-flag mechanism or even just a script to reliably bulk-edit a bunch of capture specs in production to set a new backwards-compatibility flag?

willdonnelly:
@jgraettinger I'm intrigued by the idea you mentioned during sprint planning, of using the capture creation timestamp as a sort of implicit "feature version" for should-be-trivial-but-we-worry-about-backwards-compatibility changes like this. Is that something we currently have access to, or something that would have to be added to the protocol+runtime?

jgraettinger:
It's available in a very ham-fisted way. Current CaptureSpec's `shard_template.id`'s encode the publication ID which first created the task as the last path component, and that publication ID embeds a timestamp and can be ordered lexicographically. But, we also still have captures which pre-date the control-plane change which introduced this convention, and those shard_template ID's don't have it.

Having it available as a structured field would require new control plane and protocol plumbing.

williamhbaker:
[...]

PR #2330, opened 2025-02-04
source-sqlserver: Support CDC without watermarks

willdonnelly:
**Description:**

This PR adds support to the SQL Server CDC connector for running without watermark writes. This new support exists side-by-side with the existing watermark-based mechanism, and can be enabled on a specific capture by adding the `read_only` feature flag to the advanced endpoint config.

Watermark-free CDC requires an additional database permission (`VIEW DATABASE STATE`, or in SQL Server 2022+ `VIEW DATABASE PERFORMANCE STATE` also suffices) which isn't granted by default, so we can't currently turn it on for preexisting captures. However other than the setup obstacle, there should be no downside to the new implementation [1].

[1] This is contrasted with the PostgreSQL situation, where read-only CDC has legitimate operational downsides and so we expect to retain both modes of operation forever. In SQL Server it's theoretically possible that we will get to a point where every SQL Server capture is using the new implementation and we can remove watermark writes entirely.

The plan is that after https://github.com/estuary/connectors/issues/1994 is complete we will add the `no_read_only` feature flag to all currently-existing SQL Server captures and then toggle the default behavior so any new captures will be created in watermark-free mode (we might also do some additional testing in between those points). We can then work with users to get preexisting captures moved over to watermark-less operation, or we can simply not do that if we don't think it's worth the effort.

This fixes https://github.com/estuary/connectors/issues/2215

**Workflow steps:**

Most users aren't expected to do anything. If someone creates a new capture tomorrow, it will still use watermark writes, and at some point in the not too distant future the default behavior and documentation will both change to instead ask for the `VIEW DATABASE STATE` permission and not say anything about watermarks at all.

If a user needs to opt into watermark-less mode today, they can do so by putting `read_only` in the feature flags, and after the default-flip a user could in theory opt back into watermark-writing mode by putting `no_read_only` in the feature flags.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2330)

### DIFF
--- a/source-sqlserver/backfill.go
+++ b/source-sqlserver/backfill.go
@@ -141,7 +141,7 @@ func (db *sqlserverDatabase) ScanTableChunk(ctx context.Context, info *sqlcaptur
 					Snapshot: true,
 					Table:    table,
 				},
-				LSN:    []byte{},
+				LSN:    LSN{},
 				SeqVal: seqval,
 			},
 			Before: nil,
--- a/source-sqlserver/docker-initdb.sh
+++ b/source-sqlserver/docker-initdb.sh
@@ -18,6 +18,8 @@ CREATE LOGIN flow_capture WITH PASSWORD = 'we2rie1E';
 GO
 CREATE USER flow_capture FOR LOGIN flow_capture;
 GO
+GRANT VIEW DATABASE STATE TO flow_capture;
+GO
 GRANT SELECT ON SCHEMA :: dbo TO flow_capture;
 GO
 GRANT SELECT ON SCHEMA :: cdc TO flow_capture;
--- a/source-sqlserver/main.go
+++ b/source-sqlserver/main.go
@@ -41,6 +41,12 @@ var featureFlagDefaults = map[string]bool{
 	// to work in the Turkish_CI_AS locale), but didn't want to release it unconditionally
 	// on a Friday without much testing so it's gated behind a flag for now.
 	"uppercase_discovery_queries": false,
+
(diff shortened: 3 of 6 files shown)


willdonnelly:
One issue which I just realized exists with the new implementation: the new implementation queries `sys.dm_cdc_log_scan_sessions` to determine that a new CDC scan operation has completed (because this implies that, subject to certain conditions, the current maximum CDC LSN is a valid fence position). But unlike all of the actual CDC tables, `sys.dm_cdc_log_scan_sessions` is a "management view" which is not replicated.

So after all that effort, this still doesn't actually work to allow captures from a read replica. Dangit.

(Restating the core issue here: the replicated information in the CDC tables lets us know whether or not we're caught up to the CDC worker process on the server, but if that worker manages to itself get bogged down in its WAL processing (or if it isn't running at all) it is not necessarily the case that being caught up with the CDC data tables implies the required causal relationship to backfilled table states. And as far as I can tell every source of information which would give us that last piece of the puzzle _isn't replicated_).

willdonnelly:
To be clear, the new implementation still works if the goal is "capture from the primary database without needing a watermarks table or watermark writes or anything else that would constitute a write permission on that database". The issue is just that (similar to Postgres watermark-less CDC) it turns out that doesn't actually make captures from read replicas viable.

williamhbaker (approved):
LGTM with respect to the code changes. The caveat that it won't work with read replicas does seem like a bummer, but I don't have anything too insightful to add with regards to that unfortunately.

PR #2331, opened 2025-02-04
source-zendesk-support-native: add configurable page size for incremental export streams

Alex-Bair:
**Description:**

If a user's Zendesk account has extremely large tickets/users and we request a sufficiently large number of results per page, it's possible to receive 504 responses from the intermediate gateways if Zendesk takes too long to send back results. Instead of tinkering with this in the connector code whenever this comes up, I'm opting to make this a configurable setting that's easily adjusted on a per capture basis.

A future improvement would be to automatically detect these types of 504 responses & dynamically reduce the page size. That's a bit more involved though since our CDK automatically retries any requests that receive 5xx responses, meaning the connector code never sees these 5xx errors. Determining if/how the CDK should expose 5xx responses is left as a later exercise.

Spec snapshot changes are expected due to the added config setting.

**Workflow steps:**

Docs should be updated to reflect the new config setting.

**Notes for reviewers:**

- All streams complete.
- The page size used for incremental export streams does change based on the configured page size.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2331)

### DIFF
--- a/source-zendesk-support-native/source_zendesk_support_native/api.py
+++ b/source-zendesk-support-native/source_zendesk_support_native/api.py
@@ -25,7 +25,6 @@ from .models import (
 )
 
 CURSOR_PAGINATION_PAGE_SIZE = 100
-MAX_INCREMENTAL_EXPORT_PAGE_SIZE = 1000
 MAX_SATISFACTION_RATINGS_WINDOW_SIZE = timedelta(days=30)
 # Zendesk errors out if a start or end time parameter is more recent than 60 seconds in the past.
 TIME_PARAMETER_DELAY = timedelta(seconds=60)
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

To address this with the cdk's http caller, I was thinking it would make sense to allow it to be configured with specific error codes that won't be automatically retried, maybe even on a per-call basis. Then these problematic ones could be set to not retry 504 errors and we could do the dynamic sizing and retrying in the connector code. But probably not something we need to do right now.

Alex-Bair:
I like that idea, and the per-call aspect would be helpful in `source-zendesk-support-native` since we'd typically want to always retry 504s except for these subset of requests. But I agree, not something we need to do now. Hopefully as we build more connectors & interact with more APIs, we'll get a clearer/reinforced view of what approach would be best to incorporate into the CDK.

PR #2337, opened 2025-02-04
source-postgres: Feature flag `use_schema_inference`

willdonnelly:
**Description:**

This PR adds a feature flag `use_schema_inference` to `source-postgres` which when set causes `x-infer-schema: true` to be added to discovered schemas.

Actually figuring out the right spot to add the `x-infer-schema` annotation was surprisingly tricky, what I ended up doing was adding a third "partial schema" to the top-level `allOf`, where the new schema is literally just `{"x-infer-schema": true}`. I think that works?

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2337)

### DIFF
--- a/source-postgres/.snapshots/TestGeneric-SpecResponse
+++ b/source-postgres/.snapshots/TestGeneric-SpecResponse
@@ -104,6 +104,11 @@
             "type": "boolean",
             "title": "Read-Only Capture",
             "description": "When set the capture will operate in read-only mode and avoid operations such as watermark writes. This comes with some tradeoffs; consult the connector documentation for more information."
+          },
+          "feature_flags": {
+            "type": "string",
+            "title": "Feature Flags",
+            "description": "This property is intended for Estuary internal use. You should only modify this field as directed by Estuary support."
           }
         },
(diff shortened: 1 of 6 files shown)


jgraettinger:
This actually doesn't work: our handling of `x-infer-schema` is simplistic and we only inspect it if it's a property of the schema root. Would it be easier to instead unconditionally include `x-infer-schema`, but vary whether it's value is `true` or `false` ?

willdonnelly:
> This actually doesn't work: our handling of x-infer-schema is simplistic and we only inspect it if it's a property of the schema root

Ah, I see. Okay, I've uploaded a new implementation then, which I think puts it in the right spot. The generated schema will now look like:

```json
{
    "$defs": {"...definition of the actual table properties..."},
    "allOf": [
        {"...the usual stuff..."},
        {"...and the reference to the main table properties definition..."}
    ],
    "x-infer-schema": true
}
```

Does that seem correct to you? I opted to still only add it conditionally because it was about as easy to do it either way and I prefer knowing that the default schema generation is 100% unaltered by this PR.

PR #2338, opened 2025-02-04
source-http-ingest: improve error message on invlalid json schemas

psFried:
Improves the error message when one of the collection JSON schemas cannot be parsed as an `openapi::Schema`. The openapi library we use only supports a subset of JSON schema, so things like multiple types or conditionals don't work. We might in the future want to try a different library, but for now this just improves the error message in the (hopefully rare) case that someone tries using any fancy json schema features.


williamhbaker (approved):
LGTM

PR #2353, opened 2025-02-06
Docs and scripts for bulk editing feature flags

willdonnelly:
**Description:**

This PR adds the following:
* Documentation in `docs/feature_flags.md` which describes the intended purpose of connector feature flags, how to introduce a new feature flag when changing the default behavior of a connector, and how to set a flag on all production tasks.
* A tool `scripts/list-tasks.sh` which can query the Supabase `live_specs` table to list all tasks running (any variant of) a particular connector and, optionally, add them all to the active `flowctl` draft.
* Another tool `scripts/bulk-config-editor.sh` which can edit all endpoint config YAML files under a particular directory to add a feature flag setting. It handles encrypted-vs-plaintext configs, extends or introduces the `/advanced/feature_flags` property as needed, and skips configs which already have a setting for the same "base name" of the flag (the `foobar` vs `no_foobar` distinction).

After thinking about the problem for a while, those two scripts are what I settled on as the cleanest way to break down the process. I was originally going to have a more monolithic tool, but it started feeling like I was writing a bunch of wrappers around `flowctl` commands and just exposing the underlying "create and modify a flowctl draft" process would be both cleaner and more trustworthy (since the new tools are just manipulating the draft and specs, the results can be cross-checked or refined in any way one desires before publishing).

This partially addresses https://github.com/estuary/connectors/issues/1994, now I just need to make some connector default behavior changes using the tool to close it out.

**Workflow steps:**

As described in the new docs, the overall process of bulk setting a feature flag looks like:

```bash
$ mkdir -p draft && cd draft && echo '*' > .gitignore          # Create a directory somewhere to hold the Flow specs being edited
$ flowctl draft create                                         # Create a new draft
$ ../scripts/list-tasks.sh --connector=source-mysql --draft    # Add all source-mysql (and variants) captures to the current draft
$ flowctl draft develop                                        # Pull down draft specs into a local directory
$ ../scripts/bulk-config-editor.sh --set_flag=do_some_thing    # Add the 'do_some_thing' feature flag to the local endpoint configs, with interactive diffs and prompting
$ flowctl draft author --source flow.yaml                      # Upload modified specs to the draft
$ flowctl draft publish                                        # Publish the uploaded specs
```

**Notes for reviewers:**

The scripts are written in Go, but I've added little shell helper scripts for running them because those can be symlinked into a local binaries directory (or the whole `scripts/` directory added into one's `$PATH`).

It's tempting to try and refine the scripts further, but I think that in their current form they'll work just fine for the initial rounds of feature-flag assisted behavior changes, so spending more time on them right now risks going into speculative over-engineering territory and I'd rather get them out for review in case there are real issues I've overlooked.

I was originally thinking of adding these in the `scripts` directory of the `flow` repository, but then I got to thinking about adding some connector documentation about feature flags as well and the fact that these tools are straddling the line between "flowctl helper" and "operational thing for connectors" and decided to put them here instead.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2353)

### DIFF
--- a/scripts/bulk-config-editor/main.go
+++ b/scripts/bulk-config-editor/main.go
@@ -10,7 +10,6 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -52,6 +51,10 @@ func main() {
 	} else {
 		log.SetLevel(lvl)
 	}
+	log.SetFormatter(&log.TextFormatter{
+		FullTimestamp: true,
+		PadLevelText:  true,
+	})
 
 	if err := performEdits(context.Background()); err != nil {
 		log.WithField("err", err).Fatal("error")
@@ -59,18 +62,43 @@ func main() {
 }
 
 func performEdits(ctx context.Context) error {
-	// As a sanity check, verify that there's a flow.yaml file in the current directory
-	if _, err := os.Stat("./flow.yaml"); errors.Is(err, os.ErrNotExist) {
-		return fmt.Errorf("no flow.yaml in the current directory")
+	// List all task directories (dirs containing leaf flow.yaml files) and endpoint configs
+	// (*.config.yaml files) under the current directory. This allows us to do a couple of
+	// basic sanity checks to ensure that we'll be sucessful.
+	//
+	// The way you could end up with a task directory but no corresponding config file is if
+	// that task's config is plaintext and also short enough to fall below the 512 byte threshold
+	// for 'flowctl catalog pull-specs' to break it out into a separate file. An encrypted config
+	// can never fall below this threshold because the SOPS stanza is around 700 bytes alone, and
+	// in practice this almost never happens. If it does happen, you will need to manually edit
+	// that task spec to break out the config into a separate file if you want to use this script
+	// to modify it.
+	var taskDirs, err = listTaskSpecDirs(*configsDir)
+	if err != nil {
+		return fmt.Errorf("error listing task specs: %w", err)
 	}
-
-	// List all *.config.yaml files under the current directory
-	var configFiles, err = listConfigFiles(*configsDir)
+	configFiles, err := listConfigFiles(*configsDir)
 	if err != nil {
 		return fmt.Errorf("error listing config files: %w", err)
 	}
 	if len(configFiles) == 0 {
 		return fmt.Errorf("no files named '*.config.yaml' under the current directory")
+	}
+	if len(taskDirs) > len(configFiles) {
+		// Find the directory which doesn't have a corresponding config file
+		for _, taskDir := range taskDirs {
+			var found bool
+			for _, configFile := range configFiles {
+				if strings.HasPrefix(configFile, taskDir) {
+					found = true
+					break
+				}
+			}
+			if !found {
+				fmt.Printf("task dir %q has no corresponding config file\n", taskDir)
+			}
+		}
+		return fmt.Errorf("found %d task directories but only %d endpoint configs", len(taskDirs), len(configFiles))
 	}
 
 	// Compute edits for each config file
@@ -87,6 +115,8 @@ func performEdits(ctx context.Context) error {
 			fmt.Printf("%s\n", err.Error())
 		}
 	}
+	var successCount = len(configFiles) - len(errs)
+	fmt.Printf("edited %d/%d configs\n", successCount, len(configFiles))
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


scripts/list-tasks/main.go
```diff
@@ -0,0 +1,157 @@
+// The list-tasks script enumerates all tasks using any variant of a particular
+// connector, and add them to the active flowctl draft.
```
williamhbaker, line 2: In my experiences of editing a whole bunch of task configs in a very manual way, I have found that it generally doesn't work very well to do a lot of them in the same publication, because one of them will inevitably fail validation for some reason (often because of some unrelated issue with a task that the validation expands to) and that will cause the entire publication to fail. Doing them in smaller chunks is usually what I've had to resort to.

I could see a functionality like "give me N configs where this feature flag isn't already set" being useful for doing smaller batches at a time. I see there's a name prefix included already that would help to accomplish the small batches part as it is.

scripts/list-tasks/main.go
```diff
@@ -0,0 +1,157 @@
...
+draft.
+
+Refer to 'docs/feature_flags.md' for more information on the intended
+use-case and workflow.
+`
+
+var (
+	logLevel = flag.String("log_level", "info", "The log level to print at.")
+
+	taskType   = flag.String("type", "capture", "The type of catalog spec to list (typically 'capture' or 'materialization').")
```
williamhbaker, line 44: Do we need this flag? It  seems redundant with the `imageName`, since a given image will exclusively be a capture or a materialization.

scripts/bulk-config-editor/main.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+	}
+
+	// If the preexisting feature flags already include a setting for this base flag name,
+	// we don't want to add a redundant or contradictory one. In theory we might want to
+	// be able to flip flags here, but it seems safer to just refuse to do anything when
+	// this happens.
+	var flagBase = strings.TrimPrefix(flagName, "no_")
+	if slices.Contains(featureFlags, flagBase) || slices.Contains(featureFlags, "no_"+flagBase) {
+		return nil, fmt.Errorf("task config already has a setting for feature flag %q, doing nothing", flagBase)
```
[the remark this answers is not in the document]
willdonnelly: We actually do that, up at the top-level "edit every config" loop we just accumulate any errors into a list so they can be printed out after all the diffs from successful edits.

scripts/bulk-config-editor/main.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+}
+
+func performEdits(ctx context.Context) error {
+	// As a sanity check, verify that there's a flow.yaml file in the current directory
+	if _, err := os.Stat("./flow.yaml"); errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("no flow.yaml in the current directory")
+	}
+
+	// List all *.config.yaml files under the current directory
```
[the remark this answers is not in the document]
willdonnelly: Yeah, I was assuming that the configs we're editing always come from doing a `flowctl draft develop` to pull things down and I _think_ that always consistently breaks out configs into a separate file. I'll think about whether there's a good way to add a safeguard just in case a config occurs inline though.

scripts/list-tasks/main.go
willdonnelly: This is a very good point.

I can see the appeal of adding a "where the feature flag is unset" filter option to the task listing script, but using that to try and break things down into small batches feels like it would still make the whole process a lot more hands-on than I'd prefer.

I'm currently considering implementing a third helper script which can take the directory full of specs on disk and iteratively add them to drafts and publish the drafts, either individually or in small batches.

scripts/list-tasks/main.go
willdonnelly: Yeah, there's probably not much utility in the task type filter option for bulk config edits since practically speaking the connector name needs to be used.

This script grew out of a helper script I wrote a while back for other reasons, where it actually was useful to be able to say like "give me a list of all capture tasks for tenant `acmeCo`" even when I wasn't initially sure what type of database they were capturing from.

I've changed the default value for this setting so that it won't filter on task type at all, but I've left the actual filter option in because I could still see it being useful in some circumstances and it's only like 4 lines of code.

willdonnelly:
I believe these bulk-editing scripts are in a fully usable state now and have done some test edits to a batch of test captures I created in production. Everything seems good.

willdonnelly:
Going to go ahead and get this merged now just because it feels like about time to close out this branch. All the code is new so it can't exactly break anything to do that, and if anyone has feedback on the changes or the new bulk-publishing script I can make improvements in a followup PR.

PR #2360, opened 2025-02-07
source-google-analytics-data-api-native: new connector

Alex-Bair:
**Description:**

This is an initial version of a connector to capture [basic reports](https://developers.google.com/analytics/devguides/reporting/data/v1/basics) from Google Analytics' Data API. It has parity with the current ATF connector, plus the following features:
- Support for `dimensionFilter`s and `metricFilter`s in custom reports. Stricter validation of the possible [variations](https://developers.google.com/analytics/devguides/reporting/data/v1/rest/v1beta/FilterExpression) of these fields did not seem worth the additional time & effort right now. If there's a significant number of issues with users misconfiguring filters for custom reports, we can improve the validation around these optional filters later.

A few quirks about the API/connector that are worth stating:
- The API uses the Google Analytics property's timezone, which is not always UTC. We determine the appropriate timezone by hitting the API instead of having the user configure it. This assumes the timezones returned by Google are IANA compliant, which they should be according to the [docs](https://developers.google.com/analytics/devguides/reporting/data/v1/rest/v1beta/ResponseMetaData).
- All metric values are returned as strings in the API response, but their actual type is specified in the associated metric header within the response. The ATF connector attempts to convert metrics to the appropriate type (usually an integer or float), ~~so the `_convert_to_python_type method` serves that purpose within this connector~~ but we can just keep metric values as strings and rely on schema inference to interpret whether or not the value follows a specific format.
- Similar to the ATF connector, I added the property id and the date of the report to each document. These are necessary as primary keys to differentiate the same report across different days & to pre-position for the likely feature request to support multiple property ids in a single connector.
- There is no capture snapshot since the test account I created doesn't have any analytics data in it.

**Workflow steps:**

Documentation will need created for this new connector. It should be clear about the format required for custom reports; the input should be a JSON array, _not_ a string containing a valid JSON array. For example, a valid custom reports input for the native connector looks like:
```json
[{"name": "my_custom_report", "dimensions": ["date"], "metrics": ["totalUsers"]}]
```
While an invalid custom reports input looks like:
```json
"[{\"name\": \"my_custom_report\", \"dimensions\": [\"date\"], \"metrics\": [\"totalUsers\"]}]"
```

**Notes for reviewers:**

- Default report streams create valid streams & capture data.
- Custom report streams are discovered if any are specified in the custom reports input.
- `dimensionFilter`s and `metricFilter`s can be specified in custom reports.
- Any metrics of `TYPE_INTEGER` are converted to integers if possible, and then floats if the integer conversion fails.
- OAuth doesn't fully complete, but does reach the expected failure point when testing on a local stack (i.e. `localhost:3000` is an invalid redirect URI). The generated OAuth2 spec in Supabase is extremely similar to the one for the ATF connector, and the connector's `OAUTH2_SPEC` is almost identical to the one for `source-google-sheets-native`. I expect OAuth to work fine in production.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2360)


source-google-analytics-data-api-native/source_google_analytics_data_api_native/api.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+def _convert_to_python_type(
+    name: str,
+    value: str,
+    metric_type: MetricTypes,
+) -> str | float | int:
+    # It seems like the metric type for custom metrics is not always 100% accurate.
+    # The only case I've seen so far is that TYPE_INTEGER custom metrics can actually be floats.
+    match metric_type:
```
williamhbaker, line 37: Question about this type conversion: I understand that the values are all coming through as strings. Do we need to do this conversion here in the connector though, or can we let schema inference figure it out? Schema inference should "do the right thing" in interpreting string-encoded integers and numbers as `type: string, format: integer` or `type: string, format: number`. I think in the discovered collection schemas there's a list of types like `string, integer, number`, and that's going to not be very useful since it will cause the fields to be materialized as JSON columns. 

If scheme inference marks them as `type: string, format: integer` though, for example, they will get materialized as integer columns. Schema inference will also handle the widening case of going from `format: integer` to `format: number`.

Alex-Bair: Good point. Yep, we could let schema inference handle it instead of the connector. It does interpret them as `type: string, format: integer`, etc. That'll also let me update the discovered collection schemas to only have `string` as a possible metric type. I'll make those updates.

williamhbaker (approved):
LGTM % comment about fields in the schemas

source-google-analytics-data-api-native/acmeCo/locations.schema.yaml
```diff
@@ -0,0 +1,88 @@
...
+    title: Bouncerate
+    type: string
+  property_id:
+    title: Property Id
+    type: string
+  report_date:
+    title: Report Date
+    type: string
+required:
+  - region
```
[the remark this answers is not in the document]
Alex-Bair: Another good point. All of the dimensions are needed as part of the collection key, but the metrics aren't needed. I'll remove the metric fields from the write schema, and that should cut down the number of required fields a fair bit.

PR #2374, opened 2025-02-11
source-mysql: Feature-flagged date fix

willdonnelly:
**Description:**

Adds feature flag `date_schema_format` which when set will cause date columns to have `format: date` in their discovery output.

**Workflow steps:**

If absolutely necessary you could add that manually to a task's advanced config, but if you can wait a bit I'm going to change the default in the near future and then no action will be required for new tasks and preexisting tasks will be able to opt in by removing the `no_date_schema_format` flag setting.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2374)


williamhbaker (approved):
LGTM

PR #2380, opened 2025-02-11
estuary-cdk: additional SIGQUIT logging & patching `requests.Session.send` for Airbyte imports

Alex-Bair:
**Description:**

There have been tricky, intermittent cases of imported Airbyte connectors hanging with essentially no network traffic. I've added some additional logging when a `SIGQUIT` is injected & in the `CaptureShim`'s methods to help debug when these imports hang.

I also have a theory that since Airbyte imports use the `requests` package to make synchronous HTTP requests, the `requests` package doesn't have any default timeouts, and I can't find anywhere in the `airbyte_cdk` where a timeout is specified for `requests`, that setting a timeout for `requests` _might_ prevent Airbyte imports from hanging until the connector is restarted after 24 hours. I've made `requests_session_send_patch.py` to patch in a default timeout, and imported it in all Airbyte imports. I'm not certain the missing timeouts are the problem, but it's currently the best, actionable theory I have.

We've only observed `source-zendesk-support` and `source-jira-native` hanging like this so far, but I chose to patch `requests.Session.send` for all currently imported connectors to avoid playing whack-a-mole with imports that we haven't seen have this issue yet. If any imports still hang for `DEFAULT_TIMEOUT` seconds, then we'll know that the issue isn't due to a mising `requests` timeout.

**Notes for reviewers:**

Confirmed Airbyte imports can still make HTTP requests & receive responses. Confirmed that using `requests` to send an HTTP request to a server that accepts the request but doesn't send data back will timeout after `DEFAULT_TIMEOUT` seconds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2380)

### DIFF
--- a/estuary-cdk/estuary_cdk/__init__.py
+++ b/estuary-cdk/estuary_cdk/__init__.py
@@ -95,6 +95,11 @@ class BaseConnector(Generic[Request], abc.ABC):
                 if task is this_task:
                     continue
 
+                log.info("Attempting to inject SIGQUIT exception into task.", {
+                    "task.get_name()": task.get_name(),
+                    "task.get_coro().__name__": task.get_coro().__name__,
+                })
+
                 # Reach inside the task coroutine to inject an exception, which
                 # will unwind the task stack and lets us print a precise stack trace.
                 try:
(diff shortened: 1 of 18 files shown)


williamhbaker (approved):
LGTM

PR #2381, opened 2025-02-12
source-postgres: Flagged DATE and TIME fixes

willdonnelly:
**Description:**

This PR adds two new feature flags:

 - The `date_as_date` flag causes DATE columns to be captured as YYYY-MM-DD dates which satisfy `{type: string, format: date}` where previously they were serialized as full RFC3339 timestamps.
 - The `time_as_time` flag causes TIME (aka TIME WITHOUT TIME ZONE) columns to be captured as HH:MM:SS.PPPZ time values which satisfy `{type: string, format: time}` where previously they were captured as Unix microsecond timestamps.

Both of these changes are long-awaited fixes to user complaints that would have been trivial to address except that it would conflict with an unknown amount of historical collection data.

We will be setting `no_date_as_date,no_time_as_time` on all production captures in the near future and then flipping the default so that new captures will get the new behavior automatically.

**Workflow steps:**

Remove these flags in the near future to opt into the new behavior for a preexisting capture.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2381)

### DIFF
--- a/source-postgres/backfill.go
+++ b/source-postgres/backfill.go
@@ -130,7 +130,7 @@ func (db *postgresDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture
 				return false, fmt.Errorf("error encoding row key for %q: %w", streamID, err)
 			}
 		}
-		if err := translateRecordFields(info, fields); err != nil {
+		if err := db.translateRecordFields(info, fields); err != nil {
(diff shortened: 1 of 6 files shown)


williamhbaker (approved):
LGTM

PR #2387, opened 2025-02-12
source-oracle-batch: new connector

mdibaiee:
**Description:**

- This is a copy-pasta from source-postgres-batch, with changes to queries, interfaces, etc. to make it an Oracle capture
- The main deviation is the second commit, which is actually a problem postgres also has and needs to be fixed, I will file a separate PR for that

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2387)

### DIFF
--- a/source-oracle-batch/Dockerfile
+++ b/source-oracle-batch/Dockerfile
@@ -16,7 +16,6 @@ COPY source-boilerplate ./source-boilerplate
 
 RUN go install -v ./go/...
 RUN go install -v ./source-boilerplate/...
-RUN go install -v github.com/jackc/pgx/v5
 
 # Run tests and build the connector
 COPY source-oracle-batch ./source-oracle-batch
--- a/source-oracle-batch/driver.go
+++ b/source-oracle-batch/driver.go
@@ -29,6 +29,9 @@ var (
 	// throughput, and isn't really needed anyway. So instead we emit one for every
 	// N rows, plus another when the query results are fully processed.
 	documentsPerCheckpoint = 1000
+
+	// A flag used during tests to shut down the connector gracefully after one `poll` run
+	TestShutdownAfterQuery = false
 )
 
 const (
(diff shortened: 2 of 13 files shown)


source-oracle-batch/main_test.go
```diff
@@ -73,13 +73,81 @@ func TestBasicCapture(t *testing.T) {
...
 		}
 		cupaloy.SnapshotT(t, cs.Summary())
 	})
 }
 
+type doc struct {
+	Id string `json:"ID"`
+}
+
+func TestCaptureCheckpointSCN(t *testing.T) {
```
mdibaiee, line 87: I tested this behavior by setting `scnCursor=false` and running the test and seeing it fail, while with `scnCursor=true` it works 👍🏽

source-oracle-batch/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# caching opportunities when no dependencies are updated.
+COPY go.* ./
+RUN go mod download
+
+COPY go                 ./go
+COPY source-boilerplate ./source-boilerplate
+
+RUN go install -v ./go/...
+RUN go install -v ./source-boilerplate/...
+RUN go install -v github.com/jackc/pgx/v5
```
willdonnelly, line 19: Probably don't need PGX here?

source-oracle-batch/driver.go
```diff
@@ -0,0 +1,839 @@
...
+// BatchSQLDriver represents a generic "batch SQL" capture behavior, parameterized
+// by a config schema, connect function, and value translation logic.
+type BatchSQLDriver struct {
+	DocumentationURL string
+	ConfigSchema     json.RawMessage
+
+	Connect              func(ctx context.Context, cfg *Config) (*sql.DB, error)
+	TranslateValue       func(val any, databaseTypeName string) (any, error)
+	GenerateResource     func(resourceName, schemaName, tableName, tableType string) (*Resource, error)
+	DefaultQueryTemplate string
```
willdonnelly, line 60: I've actually got a change out for review today (https://github.com/estuary/connectors/pull/2406/commits/e931edec6474265e5c172595d091ef46aea32501) which replaces this with a `SelectQueryTemplate` function.

The rationale here is that we've been talking about adding (probably optional) view discovery to the batch SQL captures, but of course views don't have any sort of XMIN / ROWSCN property for us to use so we'd probably want to make full-refresh and cursor-incremental capture behaviors work properly by default (that is, without forcing the user to input a query template).

willdonnelly (approved):
LGTM

The batch SQL captures are kind of a moving target at the moment, I'm in the process of doing some code health overhauls (see https://github.com/estuary/connectors/pull/2400 and https://github.com/estuary/connectors/pull/2406) and once those are complete I intend to add some new features like (optional) view discovery and the CDK-style full-refresh behavior with consistent keys and inferred deletions.

I think this implementation is good to merge as-is, just noting that there's a lot of stuff in the works that we'll need to apply to this connector as well.

PR #2389, opened 2025-02-12
source-mixpanel-native: increase `annotations` date window size

Alex-Bair:
**Description:**

The `annotations` stream has signficantly fewer records than other streams that use the `DateSlicesMixin`, and `annotations` isn't incremental like many of those other streams. By using small date window sizes, `annotations` makes multiple HTTP requests checking each date window for records. Since there are typically only KB of annotations in users' Mixpanel accounts, the connector _could_ make a single request for a really large date window & get all records faster.

This PR makes `annotations` always use a large date window to try & get all records in a single request. An equivalent, cleaner, more time-intensive change would be rewriting `annotations` to not use the `DateSlicesMixin` and just specifying `from_date` and `to_date` query params in the `Annotations` class. But that refactor can be done later after confirming this approach works.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that all annotations currently captured by existing tasks with `annotations` enabled are now captured in a single request.

### DIFF
--- a/source-mixpanel-native/source_mixpanel_native/streams/annotations.py
+++ b/source-mixpanel-native/source_mixpanel_native/streams/annotations.py
@@ -27,6 +27,21 @@ class Annotations(DateSlicesMixin, MixpanelStream):
 
     primary_key: str = "id"
 
+    def __init__(
+        self,
+        date_window_size: int = 30,
+        **kwargs,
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2390, opened 2025-02-12
materialize-snowflake: remove password authentication as an option

mdibaiee:
**Description:**

see https://www.snowflake.com/en/blog/blocking-single-factor-password-authentification/

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2390)


willdonnelly:
Three questions:
1. Do we actually validate endpoint configs against the schema anywhere that might be an issue here?
2. If the user tries to make a unrelated change via the UI, like adding/disabling a binding or triggering a backfill, will they be forced to also change their authentication mode or will the existing endpoint spec just be passed through unmodified?
3. I'm pretty sure that an unrelated change to the endpoint spec will definitely force them to go reconfigure their authentication mode. Is that desirable now given that there are still 9 months to go before username/password auth is disabled?

PR #2394, opened 2025-02-12
source-jira-native: misc fixes

Alex-Bair:
**Description:**

This PR's scope includes:
- Removing the capture snapshot test since the associated Jira Cloud account has been deactivated for inactivity.
- Avoid 404 responses by skipping `project_emails` and `project_versions` for deleted projects.
- Prevent the connector from paginating non-paginated endpoints for `screen_tabs` and `screen_tab_fields` when there are `maxResults` (i.e. 50) or more results in a single response.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed tests pass & streams lexicographically between `project_emails` and `screen_tabs` complete.


source-jira-native/source_jira_native/streams.py
```diff
@@ -1073,6 +1073,10 @@ def path(self, stream_slice: Mapping[str, Any], **kwargs) -> str:
 
     def read_records(self, stream_slice: Optional[Mapping[str, Any]] = None, **kwargs) -> Iterable[Mapping[str, Any]]:
         for project in read_full_refresh(self.projects_stream):
+            # Skip fetching emails for deleted projects since attempting to do so will return a 404 error.
+            if project.get('deleted', False):
```
williamhbaker, line 1077: This is blowing my mind a little so I have to ask...should this be checking for `True` instead of `False`? I would expect a deleted project have `deleted: true` instead of `deleted: false`.

Alex-Bair: That's right. Deleted projects have `deleted: true`, so we skip them. Non-deleted projects do not have a `deleted` field at all. So if `deleted` is not present `project.get('deleted', False)` will return `False` and we don't skip. I can change the default to be `None` instead since that's a bit clearer.

williamhbaker: Oh I see. I didn't understand that the second argument to `.get` was the default if the field isn't present. Just my poor understanding of python at a glance...I think this is fine as it is then!

williamhbaker (approved):

PR #2395, opened 2025-02-13
source-postgres: Flag for new DATE and TIME format

willdonnelly:
**Description:**

This reverts commit 8fd9b99543b28e2fd47891956ae4af66b25ff55c, which was itself a revert of the original implementation, and adds fixes/tests for:

  - The incorrect null-value handling that forced the revert.
  - Also arrays-of-dates being translated incorrectly when the new flag is set.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2395)

### DIFF
--- a/source-postgres/datatypes.go
+++ b/source-postgres/datatypes.go
@@ -201,7 +201,7 @@ func decodeRawJSONB(m *pgtype.Map, oid uint32, format int16, src []byte) (any, e
 	return json.RawMessage(src), nil
 }
 
-func translateRecordFields(table *sqlcapture.DiscoveryInfo, f map[string]interface{}) error {
+func (db *postgresDatabase) translateRecordFields(table *sqlcapture.DiscoveryInfo, f map[string]interface{}) error {
 	if f == nil {
 		return nil
 	}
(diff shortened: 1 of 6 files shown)


source-postgres/datatypes.go
```diff
@@ -403,13 +428,19 @@ func formatRFC3339(t time.Time) (any, error) {
 	return t.Format(time.RFC3339Nano), nil
 }
 
-func translateArray(_ *sqlcapture.ColumnInfo, isPrimaryKey bool, x pgtype.Array[any]) (any, error) {
+func (db *postgresDatabase) translateArray(column *sqlcapture.ColumnInfo, isPrimaryKey bool, x pgtype.Array[any]) (any, error) {
+	// Construct a ColumnInfo representing a theoretical scalar version of the array column
```
williamhbaker, line 432: If I'm understanding this new bit of code and the PR description correctly, it seems that arrays of dates are unique in how their elements' data types are described? I'm thinking we want to hit the correct switch case in `translateRecordField`, and so the data type needs to be `_date` instead of `date`, but I'm not 100% sure how this does/did not also apply to other types like `timetz` and `time`.

williamhbaker (approved):
LGTM % one question

source-postgres/datatypes.go
williamhbaker: Got it, thanks!

PR #2397, opened 2025-02-13
source-postgres: Omit generated columns from schemas

willdonnelly:
**Description:**

Generated columns cannot be captured via logical replication, so we should't mention them in our discovered schemas or capture their values via backfill either.

This bring PostgreSQL up to parity with how SQL Server handles computed columns.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2397)


jgraettinger (approved):
LGTM

PR #2399, opened 2025-02-13
materialize-boilerplate: CSV encoder optimizations

williamhbaker:
**Description:**

Eliminate allocations and allow for specifying an alternate quote character to improve the performance of CSV encoding.

The benefits from these changes are pretty significant, about an order of magnitude faster encoding for streams of representative JSON documents. For optimum performance it is important to not have to escape many quote characters, so picking something than a double quote is best since our JSON documents will typically have a lot of those ex. the `flow_document` field.

The CSV encoder is about equivalent with our optimized JSON encoder now, and the CPU time for both is dominated by the compression activity rather than the serialization.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2399)


Alex-Bair (approved):
LGTM

PR #2400, opened 2025-02-14
source-mysql-batch: Paying down some technical debt

willdonnelly:
**Description:**

This PR takes the `source-mysql-batch` test suite from "hilariously poorly tested" to "probably a bit over-tested" with about 700 lines of new test code covering all the important MySQL column types as well as various edge cases of key/cursor handling.

This PR also adds `table` and `schema` properties to the resource config struct and implements the same logic that `source-postgres-batch` has had for some time where if a query template is specified in the resource config we use it, but otherwise the connector knows what default query to use and plugs in the table/schema name from the resource into that hard-coded default, and when discovering new tables we always leave the template override blank.

I intend to do both of these things for the other batch SQL capture connectors in the near future, because I would really like to be able to make changes to them all with more confidence.

**Workflow steps:**

Nothing should change for existing users. New users will no longer have a big ugly template to look at if they inspect the discovered resource bindings, but they still have the option of writing their own if they really want to.

**Notes:**

If you look closely at the datatype test snapshots you might notice that our serialization of some column types is...less than optimal. We might want to consider improving that soon, but right now my goal was just to document via snapshots the current connector behavior so we can at least tell if a future change impacts anything.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2400)


williamhbaker (approved):
LGTM

PR #2403, opened 2025-02-14
source-postgres: Don't fail on pre-12 PostgreSQL

willdonnelly:
**Description:**

The new generated columns logic assumed that the PostgreSQL version is recent enough to support generated columns, but they were only added in v12 and we support earlier versions, so we ought to handle failures of that query non-fatally.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2403)

### DIFF
--- a/source-postgres/discovery.go
+++ b/source-postgres/discovery.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2420, opened 2025-02-18
scripts/list-tasks: Fix VARIANTS file 404

willdonnelly:
**Description:**

This just makes the operation of the list-tasks script a bit less aesthetically unpleasant. Since there is no connector named `ghcr.io/estuary/404: Not Found` and never will be the old code technically "worked" but it bothered me anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2420)


williamhbaker (approved):
LGTM

Issue #2421, opened 2025-02-18
source-mysql: Handle DDL alterations when column name capitalization differs from prior metadata


willdonnelly:
There is a bug in the source-mysql connector's DDL alteration handling. Alterations that require looking up old column indices may fail if the capitalization of the column name in the query doesn't match the capitalization in our metadata.

(This should only impact column name lookups at the moment, because our handling of stream identifiers internally is always lowercased. When we eventually change that we will need to account for the possibility of case-insensitive table names, but that is a completely different issue from what we've got here)

According to https://dev.mysql.com/doc/refman/8.4/en/identifier-case-sensitivity.html

> Partition, subpartition, column, index, stored routine, event, and resource group names are not case-sensitive on any platform, nor are column aliases.

That seems pretty clear: it's always correct to specify column names in DDL alterations with arbitrary capitalization, and we ought to be using case-insensitive comparisons everywhere we look up column metadata by name.

PR #2422, opened 2025-02-18
source-mysql: Handle inconsistent column name case

willdonnelly:
**Description:**

Apparently column names in MySQL are always case-insensitive, which means that it's possible to issue a DDL query naming a column with different capitalization from what we've got in our metadata, and that ought to work.

This is solved by using a case-sensitive lookup from the column name in the DDL alteration query to an index in our metadata, and then "canonicalizing" the name to match what's in our metadata so that the rest of the metadata update logic works reliably.

This fixes https://github.com/estuary/connectors/issues/2421

**Workflow steps:**

No user action required.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2422)


jgraettinger (approved):
LGTM

PR #2425, opened 2025-02-19
materialize-databricks: reset `needsMerge` flag after each transaction

williamhbaker:
**Description:**

The next transaction might not have any updates, so reset the `needsMerge` flag to prevent merge queries from being run indefinitely after one merge query has been run.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2425)


mdibaiee (commented):
Can you add a test case to our integration test to show this scenario please?

tests/materialize/fixture.json
```diff
@@ -1,4 +1,5 @@
 [
+  [],
```
williamhbaker, line 2: Added this initial empty transaction for unrelated reasons since I was updating all the test snapshots anyway (see commit message) - note that this causes the materialized timestamps to shift up for the subsequent transactions, which accounts for a lot of the diff in the snapshots.

williamhbaker:
> Can you add a test case to our integration test to show this scenario please?

👍 I extended our existing "duplicate keys" test to illustrate this scenario. We need to transition through transactions where there are merges to one where there aren't, so I had to add on an additional transaction to the fixture. The outcome is apparent for `materialize-databricks` since it shows the queries it will run in the connector checkpoint.

I made some other opportunistic changes while I was updating all the test snapshots which are described in the commit message. It took some fiddling to get all the tests passing actually but in the end I think we're better off for it!

mdibaiee (approved):
Thanks!

PR #2428, opened 2025-02-19
scripts/list-tasks: Flag to list tasks without setting

willdonnelly:
**Description:**

I'm trying to modify about 700 tasks in production and it would be really convenient when trying to make sure every last one has been updated to just throw everything away and start over with just he tasks that didn't get successfully published.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2428)


williamhbaker (approved):
LGTM! This seems like it will be really helpful.

PR #2430, opened 2025-02-19
source-postgres: Don't use IDENTIFY_SYSTEM

willdonnelly:
**Description:**

Apparently Google's AlloyDB added a gratuitous fifth field to the IDENTIFY_SYSTEM command results, which currently breaks the helper function pglogrepl.IdentifySystem which expects there to be exactly the four fields one gets back from PostgreSQL.

The only thing we use IdentifySystem for is to establish the WAL flush location on the server, and we already have our own function (`queryLatestServerLSN`) which does precisely that and doesn't rely on parsing the IDENTIFY_SYSTEM result, so let's just use that one instead.

**Workflow steps:**

There shouldn't be any user-visible impact except that AlloyDB captures work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2430)

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -62,15 +62,12 @@ func (db *postgresDatabase) ReplicationStream(ctx context.Context, startCursor s
 			logrus.WithField("err", err).Debug("error recreating replication slot")
 		}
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2432, opened 2025-02-20
source-redshift-batch: Code health improvements

willdonnelly:
**Description:**

As with source-postgres-batch and source-mysql-batch, this PR is mostly a bunch of test suite improvements along with the change to omit query templates from discovered bindings by default in favor of letting the connector pick the appropriate default query when unspecified.

I've also added CI tests for source-redshift-batch which simply launch a local Postgres instance to test against. While Redshift is different from Postgres in some important ways, it's still close enough that our test suite passes when run against Postgres and if it's a choice between "no tests in CI builds" and "somewhat unrepresentative tests in CI builds" the latter is definitely preferable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2432)


williamhbaker (approved):
LGTM

We do use an actual redshift (serverless) application for CI tests wtih `materialize-redshift`, see this [config](https://github.com/estuary/connectors/blob/main/tests/materialize/materialize-redshift/config.yaml). Not sure if that's worth trying to work in for the redshift batch capture tests now or in the future.

willdonnelly:
> We do use an actual redshift (serverless) application for CI tests wtih materialize-redshift
Oh thanks, somehow I missed that. I'll see if I can get that working for source-redshift-batch as well, testing against the actual system is definitely preferable if it's not terribly complicated.

willdonnelly:
PTAL. I've added another commit which gets the full connector test suite running against that same Redshift instance as part of CI builds. The setup for that felt a bit complicated (adding encrypted config files and rewriting the test helper logic to sops-decrypt those encrypted configs was straightforward enough, but I had to add an extra CI step for running Go tests outside of a Docker build (because plumbing GCP KMS credentials into the Docker build was complicated and I didn't feel 100% confident it wouldn't manage to leak them somehow), and I'm not sure how exactly SOPS gets those credentials during the integration tests but I had to add a bit of logic to write them into a tempfile to get it working here.

williamhbaker:
I'm also not precisely sure how the [materialization connector integration tests](https://github.com/estuary/connectors/blob/b38ecc1a3b34ecd8d18631938b050c72638ff0c9/.github/workflows/ci.yaml#L268-L292) get the GCP credentials where they need to be, which is [here](https://github.com/estuary/connectors/blob/b38ecc1a3b34ecd8d18631938b050c72638ff0c9/tests/materialize/run.sh#L54-L56) where the configs are decrypted using the `sops` command directly, unless its via the specific name of the environment variable being what `sops` expects it to be (`GCP_SERVICE_ACCOUNT_KEY`).

PR #2438, opened 2025-02-21
source-postgres-batch: XID wraparound fix and view discovery option

willdonnelly:
**Description:**

This PR bundles together three closely-related fixes to the `source-postgres-batch` connector:

- Support for non-XMIN cursor settings. Previously there was a user-modifiable cursor property in the resource spec, but in practice if it ever got set to something other than `["xmin"]` without the user also supplying their own query template this would just break. We omit query templates by default, so for that reason it makes sense that we'd want this to just work automatically (applying the same full-refresh or cursor-incremental polling behavior that other connectors use).
- Support for view discovery. Implements a new advanced option `Discover Views` which when set causes views to be returned by discovery. In a truly stunning coincidence, we need the non-XMIN cursor support to make this work correctly since views don't have an XMIN column.
- A fix for the XID wraparound failure mode. Previously there was a theoretical (never actually observed in production AFAIK) failure state for this connector where after XID wraparound occurs on a database it could go into an infinite loop reading out the full contents of the table over and over. This is remedied by adding an explicit upper bound to the XMIN-incremental polling queries so that a given polling iteration will never return rows with a higher XID than the bottom 32-bit portion of `txid_current()`.
  - Expanding slightly, we know that any row whose XID is greater (in the same mod-2^32 circular comparison sense that we use for other XID comparisons) than the bottom 32 bits of `txid_current()` must be from an earlier "epoch" (that is, the upper not-actually-stored 32 bits of the full 64-bit XID).
  - This fixes https://github.com/estuary/connectors/issues/2383

**Workflow steps:**

Nothing should change by default, the connector should just be a bit more flexible and failure-proof in general, plus there's now an advanced option for if a user wants to capture their views automatically.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2438)

### DIFF
--- a/source-postgres-batch/.snapshots/TestCaptureFromView-Capture
+++ b/source-postgres-batch/.snapshots/TestCaptureFromView-Capture
@@ -1,0 +1,45 @@
+# ================================
+# Collection "acmeCo/test/test_capturefromview_194890": 23 Documents
+# ================================
+{"_meta":{"polled":"<TIMESTAMP>","index":999},"id":0,"name":"Row 0","txid":999999,"updated_at":"2025-02-13T12:00:00Z","visible":true}
+{"_meta":{"polled":"<TIMESTAMP>","index":999},"id":1,"name":"Row 1","txid":999999,"updated_at":"2025-02-13T12:01:00Z","visible":false}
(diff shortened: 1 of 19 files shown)


mdibaiee (approved):
LGTM

PR #2439, opened 2025-02-24
source-intercom-native: dynamically reduce `conversation_parts` window size

Alex-Bair:
**Description:**

In `conversation_parts`, we attempt to only yield parts that have been updated since the previous sweep. Because of this, we can only checkpoint after checking a full date window. If a date window contains a large number of updated conversations, it can take `conversation_parts` more than one day to check all those conversations for updated parts. This has been observed for at least one task; about 4,000 pages of conversations were updated within 1 day (the smallest configurable date window).

This commit adds some dynamic reduction logic to size of the date window `conversation_parts` checks in a single `fetch_conversation_parts` invocation. If there are more than 25 pages of conversations, then the date window is split in half until there are <= 25 pages or the date window is the smallest possible size (1 second).

An alternative approach would be to yield all parts of an updated conversation, then `conversation_parts` can checkpoint after each page of conversations. I chose to do dynamic window reduction since that's keeps the connector's current behavior of only yielding updated parts, but this alternative approach could be worth while later if the date window approach is too slow & we aren't extremely concerned about yielding duplicate, non-updated parts.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed the date window for a single invocation of `fetch_conversation_parts` is reduced until there are <= 25 pages of conversations or the date window is the smallest possible size of 1 second.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2439)

### DIFF
--- a/source-intercom-native/source_intercom_native/api.py
+++ b/source-intercom-native/source_intercom_native/api.py
@@ -263,8 +263,6 @@ async def fetch_tickets(
     url = f"{API}/tickets/search"
     body = _generate_conversations_or_tickets_search_body(start, end)
 
-    count = 0
-
     while True:
         response = TicketsSearchResponse.model_validate_json(
                 await http.request(log, url, "POST", json=body)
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2440, opened 2025-02-24
source-mixpanel-native: increase `funnels` date window size

Alex-Bair:
**Description:**

This implements the fix made to `annotations` in https://github.com/estuary/connectors/pull/2389 to the `funnels` stream. The connector has to send at least one API request for each funnel, and for users with hundreds of funnels, this will at minimum take multiple hours. If the connector uses a small date window size & makes more than a single request per funnel, then it's possible for `funnels` to never complete in 24 hours. 

Since each funnel typically only has KB of data, we can force `funnels` to make a single request per funnel by using a very large date window size.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that a single request is made per funnel even when the configured date window size is small (1 day).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2440)

### DIFF
--- a/source-mixpanel-native/source_mixpanel_native/streams/funnels.py
+++ b/source-mixpanel-native/source_mixpanel_native/streams/funnels.py
@@ -43,6 +43,23 @@ class Funnels(DateSlicesMixin, IncrementalMixpanelStream):
         # we'll have to emit state every 15 records
         return 1000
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2441, opened 2025-02-24
source-zendesk-support-native: handle cursor paginated responses with no results

Alex-Bair:
**Description:**

When there are no results at all in a cursor paginated response, the after and before cursor fields are not present. Although that shouldn't affect the code flow (the `has_more` field would stop any attempted pagination), a default value is needed to avoid validation errors when validating the response.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed `posts` (the stream I observed this issue on) does not fail when the response contains no `after_cursor` or `before_cursor` field.


williamhbaker (approved):
LGTM

PR #2447, opened 2025-02-25
source-postgres-batch: Extend watchdog timeouts

willdonnelly:
**Description:**

Increases the no-data watchdog from 30 minutes to 6 hours when waiting for initial results, and from 5 to 15 minutes while results should be streaming in.


jgraettinger (approved):
LGTM

PR #2449, opened 2025-02-25
source-postgres: Set application_name param

willdonnelly:
**Description:**

This means we show up like this in pg_stat_replication:

    postgres=> SELECT * FROM pg_stat_replication;
    -[ RECORD 1 ]----+------------------------------
    pid              | 36917
    usesysid         | 16540
    usename          | flow_capture
    application_name | estuary_flow
    client_addr      | 99.27.226.225
    client_hostname  |
    client_port      | 40560
    backend_start    | 2025-02-25 16:40:03.981342+00
    [...]

Currently the `application_name` field is entirely blank, so it seems like having anything at all there is a clear win. We might consider making it customizable or filing in the capture task name in the future, but for now this is better than nothing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2449)

### DIFF
--- a/source-postgres/.snapshots/TestConfigURI-Basic
+++ b/source-postgres/.snapshots/TestConfigURI-Basic
@@ -1,2 +1,2 @@
-postgres://will:secret1234@example.com:5432/somedb
+postgres://will:secret1234@example.com:5432/somedb?application_name=estuary_flow
 config valid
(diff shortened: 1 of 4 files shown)


Alex-Bair (approved):
LGTM

willdonnelly:
> looks like there is a small test updated needed

Huh, I completely forgot we had a test covering that.

Issue #2382, opened 2025-02-12
Make Batch SQL Captures Keys work like the CDK


willdonnelly:
### Summary

We need to make the following changes to all four batch SQL capture connectors (source-postgres-batch, source-mysql-batch, source-redshift-batch, and source-bigquery-batch):

1. Add a new metadata property `/_meta/row_id`. This property will be present in all captured documents going forward, and holds a sequential integer which counts up from 0 on the first document after the start of a full refresh of a given table (and resetting back to 0 the next time a full refresh is started).
2. Add a new `streamState` property `DocumentCount` which holds the number of documents emitted since the start of the last full refresh of a table.
3. Add a `/_meta/op` property, the same as many other captures have.
    a. For bindings whose output collection key is `[/_meta/row_id]` we can fill this out intelligently, setting the operation to "c" for create/insert whenever emitting a document with `row_id >= DocumentCount` and "u" when `row_id < DocumentCount`.
    b. For any other bindings we can't make any assumptions and should omit the property entirely.
4. Upon the successful completion of a full-refresh query and before updating the persisted `DocumentCount`, and _only_ if the output collection's key is `[/_meta/row_id]`, a sequence of "inferred" deletion documents should be emitted for `row_id` values between the upper bound of the current refresh and the persisted `DocumentCount`, with an `op` value of "d".
5. Add a new feature flag `keyless_row_id` with default value false. When false, the recommended collection key for a keyless table will continue to be whatever it already is today. When true, the recommended collection key for a keyless table will instead be `[/_meta/row_id]`.

All of these changes are fully backwards-compatible, with (1-4) being safe to implement unconditionally and (5) being made safe by a feature flag.

### Backstory

Currently our batch SQL connectors discover keyed tables with their primary key as the collection key, and keyless tables with a fallback key of `["/_meta/polled", "/_meta/index"]`. Completely independently of the collection keys, sometimes tables are configured with a cursor for incremental polling and other times they have no cursor configured (and sometimes in PostgreSQL the incremental polling cursor is the `xmin` system column, but that's irrelevant here).

So we have a matrix of four combinations:
* **Keyed tables with a cursor**: these are handled fine today. Incremental changes are captured efficiently and reduced into the destination according to the table's primary key.
* **Keyed tables without a cursor**: these are also handled basically fine. Each time the table is polled we perform a full refresh, but all of the resulting rows are reduced into the destination by primary key. The only suboptimal bit here is that it might be nice if we could somehow infer row deletions.
[...]

PR #2450, opened 2025-02-25
source-postgres-batch: Add `/_meta/row_id` and `/_meta/op`

willdonnelly:
**Description:**

This PR implements https://github.com/estuary/connectors/issues/2382 for `source-postgres-batch`:

- Adds `/_meta/row_id` and `/_meta/op` properties to all documents, plus logic to fill them out as appropriate.
- Adds a feature flag `keyless_row_id` which when set causes tables without a PK in the source DB to be discovered with collection key `[/_meta/row_id]` instead of the old `[/_meta/polled, /_meta/index]` behavior.
- Adds special case logic for inferring deletions when a collection's key is `row_id` and it's a full-refresh binding.
- Adds tests exercising the new logic and both states of the `keyless_row_id` feature flag.

**Notes:**

- In the course of making this change I opted to change the document sanitization logic so that `/_meta/index` is usually untouched, and only tests which run a capture in parallel with DB writes now perform that sanitization (since the breakdown of changes into polling intervals isn't consistent for those tests but is for any others). This causes a bunch of test snapshot churn, but it makes it slightly easier to verify correct operation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2450)


williamhbaker (approved):
LGTM

PR #2452, opened 2025-02-26
Feature/materialize cratedb

surister:
**Description:**

First draft of CrateDB connector, it still has uknown issues, if merged please set it to BETA.


materialize-postgres/Dockerfile
```diff
@@ -12,7 +12,7 @@ COPY go.* ./
 RUN go mod download
 
 # Copy in the latest flowctl-go for usage by tests.
-COPY flow-bin/flowctl-go /usr/local/bin/flowctl-go
+#COPY flow-bin/flowctl-go /usr/local/bin/flowctl-go
```
williamhbaker, line 15: This change needs to be backed out before we can merge.

materialize-cratedb/CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
+# materialize-postgres
```
williamhbaker, line 1: This changelog file is not relevant to the new connector - it can just be deleted for now for cratedb.

materialize-cratedb/Dockerfile
```diff
@@ -0,0 +1,43 @@
+ARG BASE_IMAGE=ghcr.io/estuary/base-image:v1
```
williamhbaker, line 1: We need to add the connector to the CI process so it will get built. See https://github.com/estuary/connectors/commit/8bce00f175db9824b658d2644435f6366c6608ff for an example of doing this. 

For this case, it will need to be added to the list in the `build_connectors` matrix, but probably not to the `Materialization connector ${{ matrix.connector }} integration tests` matrix, since it doesn't look like those are fully working yet.

williamhbaker (changes requested):
A couple of small changes needed before we can merge this beta version.

williamhbaker (approved):
LGTM for a beta version

PR #2453, opened 2025-02-26
source-oracle: use rs_id and ssn for keyless tables

mdibaiee:
**Description:**

- majority of changes are test updates. While at it, reworked our test suite a bit to make it faster (use online dictionary mode), less brittle (close logminer sessions between test runs to avoid resource exhaustion) and shorter (only run tests on a PDB instance, running every test twice for PDB and non-PDB isn't necessary anymore IMO, if we work in PDB, non-PDB is just a simpler version of the same operation)

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2453)


source-oracle/decode.go
```diff
@@ -366,6 +366,8 @@ func (s *replicationStream) decodeMessage(msg logminerMessage) (sqlcapture.Datab
 		},
 		SCN:   msg.SCN,
 		RowID: rowid,
+		RSID:  msg.RSID,
+		SSN:   msg.SSN,
```
mdibaiee, line 370: RSID + SSN related change

source-oracle/replication.go
```diff
@@ -553,6 +557,9 @@ func (s *replicationStream) receiveMessages(ctx context.Context, startSCN, endSC
 			return err
 		}
 
+		// For some reason RSID comes with a space before and after it
+		msg.RSID = strings.TrimSpace(msg.RSID)
+
```
mdibaiee, line 562: RSID + SSN related change

source-oracle/replication.go
```diff
@@ -245,6 +245,10 @@ type oracleSource struct {
 	SCN int `json:"scn,omitempty" jsonschema:"description=SCN of this event, only present for incremental changes"`
 
 	RowID string `json:"row_id" jsonschema:"description=ROWID of the document"`
+
+	RSID string `json:"rs_id" jsonschema:"description=Record Set ID of the logical change"`
+
+	SSN int `json:"ssn" jsonschema:"description=SQL sequence number of the logical change"`
```
mdibaiee, line 251: RSID + SSN related change

source-oracle/backfill.go
```diff
@@ -117,6 +118,10 @@ func (db *oracleDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.D
 					Table:    table,
 				},
 				RowID: rowid,
+
+				// Artificial RSID and SSN during backfill
+				RSID: "0x000000.00000000.0000",
+				SSN:  rowOffset,
```
mdibaiee, line 124: RSID + SSN related change

source-oracle/.snapshots/TestAllTypes-discover
```diff
@@ -269,13 +269,23 @@ Binding 0:
...
+                      "description": "SQL sequence number of the logical change"
                     }
                   },
                   "type": "object",
                   "required": [
                     "schema",
                     "table",
-                    "row_id"
+                    "row_id",
+                    "rs_id",
```
willdonnelly, line 287: Are there any concerns about breakage to preexisting users when adding these as required? I think the answer is no (very few users at present and at least one seems to have automatic discovery disabled anyway), but just wanted to make sure this has been considered and we're confident no compatibility strategies are required.

willdonnelly (approved):
LGTM overall, only question is whether there's any backwards-compatibility concerns from this change (I think no?)

source-oracle/.snapshots/TestAllTypes-discover
mdibaiee: Yeah no real customers are using this seriously yet. Launchmetrics has been the only one who tried it but ended up not fitting their requirements and they are waiting for the batch connector

PR #2455, opened 2025-02-26
materialize-boilerplate: fix staged files cleanup again

williamhbaker:
**Description:**

In f4a82c4 a partial fix was implemented to prevent attempts to cleanup the same files again when a binding had data for one but not a subsequent transaction.

That change fixed that issue, but it also made it so that staged files just never got cleaned up, because `f.started` is set to `false` by `f.flush()`, which is always called before the cleanup.

A complete fix is here, which clears out the list of uploaded files after sending them off for deletion.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2455)


Alex-Bair (approved):
LGTM

PR #2457, opened 2025-02-26
source-postgres: Detect tuple/relation mismatch

willdonnelly:
**Description:**

This should never happen but we observed it once. I still have no idea how it could possibly happen so we need more information in case it happens again someday.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2457)


williamhbaker (approved):
LGTM

PR #2459, opened 2025-02-26
source-bigquery-batch: Row IDs and View Discovery

willdonnelly:
**Description:**

This PR adds two distinct features which I'm copying over from source-postgres-batch to other SQL batch captures:

- Row IDs and their use as the fallback collection key for tables without a primary key.
- An advanced option to return views from discovery.

I'm partly just being lazy by combining the two into a single PR, but it also partly makes sense to do so because views don't have a primary key and also don't always have a useful cursor column, and so we need the keyless-table full-refresh row ID behavior to really make them work nicely.

This PR implements https://github.com/estuary/connectors/issues/2382 for source-bigquery-batch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2459)

### DIFF
--- a/source-bigquery-batch/.snapshots/TestBinaryTypes-Capture
+++ b/source-bigquery-batch/.snapshots/TestBinaryTypes-Capture
@@ -1,14 +1,14 @@
 # ================================
 # Collection "acmeCo/test/binarytypes_537491": 6 Documents
 # ================================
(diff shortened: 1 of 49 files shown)


williamhbaker (approved):
LGTM

Issue #2463, opened 2025-02-28
source-intercom-native: Support querying all contact tags


keatmin:
We rely on contact tags for customer journey heavily and we normally have more than 10 tags per contact. 

**Current behaviour**:
When there's more than 10 tags assigned to a contact, `tags/data` will truncate it and only show 10 tags as per the [documentation](https://developers.intercom.com/docs/references/2.8/rest-api/api.intercom.io/models/contact_tags). 
> An object containing tags meta data about the tags that a contact has. Up to 10 will be displayed here. Use the url to get more.

In this scenario, `tags/has_more` will be true and in order to get the full 10+ tags there will be a url in `tags/url` to get all the tags. I believe the connector currently doesn't query that url to get all the data. What we are currently seeing is that we are missing tags for a contact that has more than 10+ tags.

**Expected behaviour**:
when a contact data has `tags/has_more = true` and `tags/count > 10` Estuary connector should query for all the tags using the url provided in tags/url

Alex-Bair:
Hey, @keatmin! I've been monitoring this behavior after merging https://github.com/estuary/connectors/pull/2465, and we're capturing all tags for contacts even when there are more than 10 tags attached to the contact. I'm closing this issue, but let us know if you see this new behavior not happening & we'll look into it.

PR #2465, opened 2025-02-28
source-intercom-native: hydrate `contacts` with all attached tags

Alex-Bair:
**Description:**

A contact contain multiple sub-resources, like tags. There can be a large number of sub-resources attached to one contact, but only a subset are included when fetching a contact from Intercom's API. For example, only the [first 10 tags](https://developers.intercom.com/docs/references/2.8/rest-api/api.intercom.io/models/contact_tags) are included in a contact. To get all sub-resources, the connector would need to make an additional request to fetch them then hydrate the contact with the full set of sub-resources.

This commit adds this hydration functionality for tags within a contact. There are other possible sub-resources that we could hydrate (notes, companies, opted_out_subscription_types, etc.), but users are only asking about tags right now. So I'm only adding functionality for tags for now. Hydrating other sub-resources should be fairly straightforward if we need to do so later.

Capture snapshot changes are expected. I added >10 tags to both contacts in our developer Intercom account to test for this new tag hydration functionality.

Addresses https://github.com/estuary/connectors/issues/2463.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed contacts that previously had only 10 tags included & have more than 10 attached tags now contain the full set of >10 tags.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2465)

### DIFF
--- a/source-intercom-native/source_intercom_native/api.py
+++ b/source-intercom-native/source_intercom_native/api.py
@@ -12,6 +12,9 @@ from pydantic import TypeAdapter
 from .models import (
     IntercomResource,
     TimestampedResource,
+    Contact,
+    NestedTag,
+    ContactTagsResponse,
     ContactsSearchResponse,
     TicketsSearchResponse,
     ConversationsSearchResponse,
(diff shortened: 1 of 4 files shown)


source-intercom-native/source_intercom_native/api.py
```diff
@@ -184,6 +187,33 @@ def _is_large_date_window(start: int, end: int) -> bool:
...
 
+async def _hydrate_contact(
+    http: HTTPSession,
+    contact: Contact,
+    log: Logger,
+) -> Contact:
+    if contact.tags.has_more:
+        url = f"{API}/contacts/{contact.id}/tags"
+        response = ContactTagsResponse.model_validate_json(
+            await http.request(log, url)
```
williamhbaker, line 198: Do we always get the all the tags in a single request? For example, would this ever need to be paginated?

williamhbaker (approved):
LGTM

source-intercom-native/source_intercom_native/api.py
Alex-Bair: Yep, we always get all tags in a single request. Unless Intercom updates the endpoint to use pagination in a future API version, we won't need to paginate to get tags.

PR #2466, opened 2025-02-28
Revert "source-postgres-batch: Fix XID wraparound behavior"

willdonnelly:
**Description:**

This reverts commit e57acbcb75aed8f74b15ff6cfeeb9ed3e200f981 and un-fixes https://github.com/estuary/connectors/issues/2383

Turns out the particular function we're using here doesn't work on a standby replica, which is...bad. Since the XID wraparound failure mode this guards against is still mostly theoretical (has never been seen in production yet) and the failures on read-only standby replicas are very much real, we should revert this and then revisit how to make it work on a standby as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2466)


jgraettinger (approved):
LGTM

PR #1969, opened 2024-09-19
source-stripe-native: config update, typing, and refactors

Alex-Bair:
**Description:**

This PR covers a handful of quick updates:
- Remove a redundant line in the GH python workflow.
- Rename the config's `stop_date` to `start_date`.
  - Update the default `start_date` to be 30 days in the past rather than a fixed date in 2010.
- Improve typing.
  - Create classes that contain shared attributes among models.
  - Add typing to `fetch_*` functions in `api.py`.
- Minor refactors related to variable naming & reuse that improve readability.

Spec snapshot changes are expected due to the `start_date` updates. 

Discover snapshot changes are expected due to the new classes containing shared attributes. The `created` field exists on all non-child streams' documents, and we were already relying on `created` to know when to stop backfilling. Making `created` explicit in the model improves Python's type checking.

**Workflow steps:**

(How does one use this feature, and how has it changed)

**Documentation links affected:**

(list any [documentation links](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing) that you created, or existing ones that you've identified as needing updates, along with a brief description)

**Notes for reviewers:**

Tested locally & confirmed start date field renders correctly in the UI. Also confirmed tasks are created without issue, and streams still backfill & incrementally replicate for a short time span.

After merging, existing tasks will need updated to rename their `stop_date` to `start_date`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/1969)

PR #2467, opened 2025-02-28
source-stripe-native: Support Stripe Connected Accounts

JustinASmith:
**Description:**

- Enhanced `open_binding` function to support dictionary-based fetch functions for `fetch_changes`, `fetch_page`, and `fetch_snapshot`. This allows each function to run as a separate subtask with its own independent state.
- Updated `source-stripe-native` to use the CDK feature to capture data for connected accounts when enabled in the EndpointConfig.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2467)


estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -510,55 +525,140 @@ def open_binding(
...
-    fetch_changes: FetchChangesFn[_BaseDocument] | None = None,
-    fetch_page: FetchPageFn[_BaseDocument] | None = None,
-    fetch_snapshot: FetchSnapshotFn[_BaseDocument] | None = None,
+    fetch_changes: FetchChangesFn[_BaseDocument]
+    | dict[str, FetchChangesFn[_BaseDocument]]
+    | None = None,
+    fetch_page: FetchPageFn[_BaseDocument]
+    | dict[str, FetchPageFn[_BaseDocument]]
+    | None = None,
+    fetch_snapshot: FetchSnapshotFn[_BaseDocument]
```
williamhbaker, line 534: I don't think we should handle multiple cursors for a `fetch_snapshot`, at least not right now. These "snapshot" bindings use a `_meta/row_id` key and I'm not sure how we would make that work with multiple snapshot processes writing documents into the same collection. I'm not sure if it really could be made to work.

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -510,55 +525,140 @@ def open_binding(
...
+    the state for each subtask in state.inc, state.backfill, or state.snapshot.
     """
 
     prefix = ".".join(binding.resourceConfig.path())
 
     if fetch_changes:
+        if isinstance(fetch_changes, dict):
+            for subtask_id, subtask_fetch_changes in fetch_changes.items():
+
+                async def incremental_subtask(
```
williamhbaker, line 556: Does this work if the closure here just only takes `subtask_task` as an input and doesn't set the `fetch_fn` and `subtask_id` as defaults in the function signature? It took me a while to figure out that it was basically doing the same thing as `closure` in the non-dictionary case, so it might be nice to make the two more symmetrical, unless there's some reason the values need to be initialized like that.

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -510,55 +525,140 @@ def open_binding(
...
+                )
 
-        task.spawn_child(f"{prefix}.incremental", closure)
+            task.spawn_child(f"{prefix}.incremental", closure)
 
     if fetch_page and state.backfill:
+        if isinstance(fetch_page, dict):
+            for subtask_id, subtask_fetch_page in fetch_page.items():
+
+                async def backfill_subtask(
```
williamhbaker, line 594: Same question as above on the inputs to `incremental_subtask`.

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -510,55 +525,140 @@ def open_binding(
...
+                    fetch_page,
+                    state.backfill,
+                    task,
+                )
 
-        task.spawn_child(f"{prefix}.backfill", closure)
+            task.spawn_child(f"{prefix}.backfill", closure)
 
     if fetch_snapshot:
+        if isinstance(fetch_snapshot, dict):
```
williamhbaker, line 627: As noted above, I don't think we should attempt to support multiple cursors for snapshot tasks right now.

source-stripe-native/source_stripe_native/resources.py
```diff
@@ -88,221 +143,396 @@ async def all_resources(
...
 
     def open(
         binding: CaptureBinding[ResourceConfig],
         binding_index: int,
         state: ResourceState,
         task: Task,
-        all_bindings
+        all_bindings,
     ):
+        fetch_chnages_fns = {
```
williamhbaker, line 238: Typo in this variable name, and all similar declarations in below it in other resource builder functions.

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -179,18 +179,22 @@ class Snapshot(BaseModel, extra="forbid"):
...
     )
 
-    backfill: Backfill | None = Field(
+    backfill: Backfill | dict[str, Backfill | None] | None = Field(
         default=None,
         description="Backfill progress, or None if no backfill is occurring",
     )
 
-    snapshot: Snapshot | None = Field(default=None, description="Snapshot progress")
+    snapshot: Snapshot | dict[str, Snapshot | None] | None = Field(
```
williamhbaker, line 191: See note below about not having this change apply to snapshot bindings

source-stripe-native/source_stripe_native/models.py
```diff
@@ -74,20 +79,24 @@ class BaseStripeObject(BaseDocument, extra="allow"):
     """
     Shared properties between all Stripe objects.
     """
+
     NAME: ClassVar[str]
     SEARCH_NAME: ClassVar[str]
 
     id: str
     object: str
+    account_id: str | None = None
```
williamhbaker, line 88: This schema generated from this ends up being a little weird, where it has a default of `null`. I'm not exactly sure the best way to deal with this, but what we have done elsewhere is a this bit: `json_schema_extra=lambda x: x.pop("default")` to not schematize the default value. The end goal is to allow the field to be absent. A schema with a default `null` value will mean that the field is always present, but a `null` is substituted if it is absent. 

To be very precise, I think we want the schema for this to be `type: string`, and have it not show up in the list of required fields. Type `[string, null]` isn't really correct since if it is present, it won't ever be `null`, and will always be a string.

source-stripe-native/acmeCo/Accounts.schema.yaml
```diff
@@ -21,17 +21,29 @@ $defs:
...
-      - $ref: "#/$defs/Meta"
+    $ref: "#/$defs/Meta"
     default:
       op: u
       row_id: -1
     description: Document metadata
   id:
     title: Id
     type: string
+  object:
```
williamhbaker, line 32: Do you know why all of these schemas are now getting the required field `object` added to them, and also `created`? I think I can see in code why they have them, but don't really understand why they weren't there before.

It might be alright to have them in the schema really, but if we are adding a new required field `object` we definitely should make sure it is always present in existing collections.

williamhbaker (commented):
Looking really good; made a few comments for consideration, nothing major.

source-stripe-native/acmeCo/Accounts.schema.yaml
Alex-Bair: I made `created` required in this [PR](https://github.com/estuary/connectors/pull/1969) and `object` required in this [PR](https://github.com/estuary/connectors/pull/1976). IIRC, these fields are always present, and the connector was already relying/needed to rely on them being present for one reason or another. I missed updating all the yaml files under `acmeCo` whenever I made those changes though, so we're seeing those previous changes reflected in `acmeCo/foobar.schema.yaml` now.

source-stripe-native/source_stripe_native/api.py
```diff
@@ -348,72 +410,75 @@ async def _capture_substreams(
...
+                    log, child_url, method="GET", params=parameters, headers=headers
+                )
             )
         except HTTPError as err:
             # It's possible for us to process events for deleted parent resources, making
             # the requests for the associated child resources fail. Stripe returns a 404
             # error & a message containing "resource_missing" and "No such" when this happens.
-            if err.code == 404 and any(
-                re.search(regex, err.message, re.DOTALL)
-                for regex in MISSING_RESOURCE_REGEXES
```
Alex-Bair, line 371: We should still check for all of the possible missing messages in `MISSING_RESOURCE_REGEXES` in this if condition.

JustinASmith: Thanks for catching that. I believe merge conflict resolution removed that. I'll make sure to reverse that change and go through things once more.

estuary-cdk/estuary_cdk/capture/common.py
JustinASmith: Ah okay. I'll remove this changes related to snapshots. I need to dig more into the quirks so I understand this a bit more.

JustinASmith: Done.

estuary-cdk/estuary_cdk/capture/common.py
JustinASmith: Updated. See above comment.

estuary-cdk/estuary_cdk/capture/common.py
JustinASmith: Updates to remove this.

source-stripe-native/source_stripe_native/resources.py
JustinASmith: Good catch. Fixed typo. Any recommended extensions/tools for quick checks for spelling errors in code?

source-stripe-native/source_stripe_native/api.py
JustinASmith: Reverted this to prior logic.

estuary-cdk/estuary_cdk/capture/common.py
williamhbaker: Looks good!

source-stripe-native/source_stripe_native/resources.py
williamhbaker: :shrug: Sometimes fresh eyes are the only thing that catches it.

williamhbaker (approved):
LGTM

source-stripe-native/source_stripe_native/__init__.py
```diff
@@ -47,6 +48,35 @@ async def validate(
...
+            if (
+                resource_name in prev_bindings_by_name
+                and current_binding.backfill <= prev_bindings_by_name[resource_name]
+            ):
+                raise ValidationError(
+                    [
+                        "Cannot change the `capture_connected_accounts` property without backfilling all collections again."
+                    ]
+                )
+
```
JustinASmith, line 79: Validation logic to require a backfill of all collections when the `capture_connected_accounts` endpoint config property changes

estuary-cdk/estuary_cdk/capture/request.py
```diff
@@ -28,15 +28,28 @@ class Discover(GenericModel, Generic[EndpointConfig]):
...
+
+class ValidateLastCapture(ValidateBase, Generic[EndpointConfig, ResourceConfig]):
+    config: ValidateLastCaptureEndpointConfig
+
+
+class Validate(ValidateBase, Generic[EndpointConfig, ResourceConfig]):
+    config: EndpointConfig
+    lastCapture: ValidateLastCapture[EndpointConfig, ResourceConfig] | None = None
+
+
```
JustinASmith, line 52: Updated Validate models to include `lastCapture` spec. Currently uses a work around due to issues serializing `lastCapture` into the `EndpointConfig` model.

williamhbaker (approved):
Updates LGTM! I think this is fine as-is, with the workaround for the `lastCapture` issue. That's probably worth looking into more, but for now this seems good.

PR #2469, opened 2025-02-28
source-sftp: Additional key-exchange algorithms

willdonnelly:
**Description:**

Extends the SSH client config with two more key-exchange algorithms which golang.org/x/crypto/ssh supports but doesn't include in its default list. This means we support all currently-supported ones, but it's been written like this so that if the package itself adds new ones in the future we don't have to explicitly update our code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2469)

### DIFF
--- a/source-sftp/main.go
+++ b/source-sftp/main.go
@@ -138,6 +138,13 @@ func (c config) PathRegex() string {
 	return c.MatchFiles
 }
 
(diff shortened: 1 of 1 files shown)


dyaffe (approved):
LGTM

PR #2470, opened 2025-02-28
common/feature_flags: Trim whitespace

willdonnelly:
**Description:**

Makes it so a setting like `do_foo, do_bar` parses the same as `do_foo,do_bar`.

Without this fix it instead parses the second one as a flag setting `" do_bar"` which is clearly very distinct from the intended `"do_bar"` flag.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2470)


Alex-Bair (approved):
LGTM!

Issue #2460, opened 2025-02-27
source-postgres: use flat arrays rather than array objects w/ dimensions


willdonnelly:
**Backstory:**

Currently `source-postgres` captures array columns as a JSON object with `dimensions` and `elements` properties, each of which is an array. This is because PostgreSQL arrays are inherently multidimensional, and we're trying to preserve them as faithfully as possible.

(The JSON schema for a recursive "array whose elements can each be type X or another nested array" type is so incredibly terrible that it was never a realistic option, so "a flat array of values plus a list of array dimensions" was the best we could do to represent the original value faithfully)

But it turns out that nobody ever wants that, they just want simple, boring one-dimensional arrays of values to translate into a JSON array of equivalent values. So we should do that.

**Plan:**

1. Add a feature flag `flattened_arrays`.
  a. When `flattened_arrays` is set, value translation will return just the elements array instead of the array-with-dimensions object.
  b. When `flattened_arrays` is set, discovered schemas will of course reflect that change in behavior.
2. Set `no_flattened_arrays` on all `source-postgres` captures in production.
3. Merge a followup PR which toggles the default for new captures.

willdonnelly:
This work is complete. All preexisting captures set `no_flattened_arrays` and the default will be `flattened_arrays` going forward.

PR #2471, opened 2025-02-28
source-postgres: Add 'flatten_arrays' flag

willdonnelly:
**Description:**

Currently source-postgres captures array columns as a JSON object with dimensions and elements properties, each of which is an array. This is because PostgreSQL arrays are inherently multidimensional, and we're trying to preserve them as faithfully as possible.

But it turns out that nobody ever wants that, they just want simple, boring one-dimensional arrays of values to translate into a JSON array of equivalent values. So we should do that by default going forward.

This partially addresses https://github.com/estuary/connectors/issues/2460

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2471)


williamhbaker (approved):
LGTM

PR #2474, opened 2025-03-03
source-{postgres,mysql,redshift,bigquery}-batch: Enable 'keyless_row_id' by default

willdonnelly:
**Description:**

This PR makes 'keyless_row_id' the default behavior for the Postgres, MySQL, Redshift, and BigQuery batch SQL captures going forward. It is broken up into individual commits just in case we have to revert, but it didn't seem necessary to make it four distinct PRs since I'm going to be fiddling with all the production tasks to set `no_keyless_row_id` anyway, so I might as well just do all four connectors and then merge the one PR.

This is part of https://github.com/estuary/connectors/issues/2382

**Workflow steps:**

New captures will get the new Row ID behavior by default. Preexisting captures will be able to opt in by removing the `no_keyless_row_id` flag from their advanced settings.

This PR will only be merged after I have set `no_keyless_row_id` on all production tasks so their current behavior doesn't change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2474)


williamhbaker (approved):
LGTM

PR #2475, opened 2025-03-03
source-postgres: Set 'flatten_arrays' by default

willdonnelly:
**Description:**

Another feature-flagged default change. This one will make new captures going forward discover array columns as just the JSON array of flattened element values.

As with other changes of this kind, this will only be merged after I have set the corresponding `no_<flag>` on all tasks in production to preserve current behavior.

This is part of https://github.com/estuary/connectors/issues/2460

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2475)


williamhbaker (approved):
LGTM

PR #2476, opened 2025-03-03
scripts/list-tasks: Rework missing-flags logic

willdonnelly:
**Description:**

The main purpose of thise flag is so that when doing a bulk flag flip in production, one can go and re-pull just tasks that still needh to be modified.

And my reasoning is that if I'm going and setting a flag like `no_foobar` in prod, there are three cases:
 - A task has no feature flag setting for `foobar`. These are the ones we want to modify.
 - A task has setting `no_foobar`. These don't need to be touched since they already have the value we want.
 - A task has setting `foobar`. These _really_ shouldn't be touched since somebody has explicitly indicated that they want the opposite of what we're doing.

Previously the 'missing flag' logic was written such that one might go `--missing=no_foobar` to get all tasks without the specific flag setting `no_foobar` in their feature flags. And since that's still the typical desire and it's weird to have to write `--missing=foobar` and then `--set_flag=no_foobar` for different commands and one could easily screw that up, the logic also normalizes so `--missing=no_foobar` and `--missing=foobar` do the same thing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2476)

### DIFF
--- a/scripts/list-tasks/main.go
+++ b/scripts/list-tasks/main.go
@@ -26,7 +26,6 @@ import (
 	"os"
 	"os/exec"
 	"path"
-	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2477, opened 2025-03-04
Remove delete tables

surister:
**Description:**
* Make integration ~tests pass by tweaking a fixture value~ and modifying two cratedb types.
* Add remove temp tables

**Notes for reviewers:**
I'm not sure if this is the right place to delete tables, I tried to do at the end of `Load` but on the last load iteration it tries to insert a new value, and of course the temp table does not exist.

Next best thing I thought was to delete when the session was over, similarly to how postgres temp table works, I worked out that `Destroy` was the right place to do that, on top of that I'm also trying to delete the binding `before` adding a new Binding, just in case.

In the `Destroy` thingy, I'm just creating a blank `context.Background()` for `removeBinding`, does that affect in any way?, in https://github.com/estuary/connectors/commit/c50cb09a3ec6d3458ae2fe5a41eeda7f076847ee I tried to pass the context to the `transactor` to be able to re-utilize it later and the effect was the same in my tests

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2477)

### DIFF
--- a/materialize-cratedb/sqlgen.go
+++ b/materialize-cratedb/sqlgen.go
@@ -170,6 +170,11 @@ CREATE TABLE {{ template "temp_name" . }} (
 );
 {{ end }}
 
+-- Templated deletion of a temporary load table:
+{{ define "dropLoadTable" }}
+DROP TABLE IF EXISTS {{ template "temp_name" . }};
+{{ end }}
+
 -- Templated insertion into the temporary load table:
 
 {{ define "loadInsert" }}
@@ -296,6 +301,7 @@ UPDATE {{ Identifier $.TablePath }}
 	AND   fence     = {{ $.Fence }};
 {{ end }}
 `)
+	tplDropLoadTable     = tplAll.Lookup("dropLoadTable")
 	tplCreateLoadTable   = tplAll.Lookup("createLoadTable")
 	tplCreateTargetTable = tplAll.Lookup("createTargetTable")
 	tplAlterTableColumns = tplAll.Lookup("alterTableColumns")
--- a/materialize-cratedb/driver.go
+++ b/materialize-cratedb/driver.go
@@ -277,6 +277,7 @@ func newTransactor(
 	if d.load.conn, err = pgx.Connect(ctx, cfg.ToURI()); err != nil {
 		return nil, nil, fmt.Errorf("load pgx.Connect: %w", err)
 	}
+
 	if d.store.conn, err = pgx.Connect(ctx, cfg.ToURI()); err != nil {
 		return nil, nil, fmt.Errorf("store pgx.Connect: %w", err)
 	}
@@ -290,6 +291,11 @@ func newTransactor(
 	}
 
 	for _, binding := range bindings {
+		// Make sure that the binding does not exist before creating it.
+		if err = d.removeBinding(ctx, binding); err != nil {
+			return nil, nil, fmt.Errorf("remove binding: %w", err)
+		}
+
 		if err = d.addBinding(ctx, binding, is); err != nil {
 			return nil, nil, fmt.Errorf("addBinding of %s: %w", binding.Path, err)
 		}
@@ -316,6 +322,18 @@ type binding struct {
 	storeInsertSQL string
 	deleteQuerySQL string
 	loadQuerySQL   string
+}
+
+func (t *transactor) removeBinding(ctx context.Context, target sql.Table) error {
+	var b = &binding{target: target}
+	var w strings.Builder
+
+	if err := tplDropLoadTable.Execute(&w, &b.target); err != nil {
+		return fmt.Errorf("executing dropLoadTable template: %w", err)
+	} else if _, err := t.load.conn.Exec(ctx, w.String()); err != nil {
+		return fmt.Errorf("Exec(%s): %w", w.String(), err)
+	}
+	return nil
 }
 
 func (t *transactor) addBinding(ctx context.Context, target sql.Table, is *boilerplate.InfoSchema) error {
@@ -519,6 +537,13 @@ func (d *transactor) Store(it *m.StoreIterator) (_ m.StartCommitFunc, err error)
 }
 
 func (d *transactor) Destroy() {
+	for _, b := range d.bindings {
+		err := d.removeBinding(context.Background(), b.target)
+		if err != nil {
+			return
+		}
+	}
+
 	d.load.conn.Close(context.Background())
 	d.store.conn.Close(context.Background())
 }


materialize-cratedb/sqlgen.go
```diff
@@ -33,8 +33,8 @@ var crateDialect = func() sql.Dialect {
 			sql.ARRAY:          sql.MapStatic("TEXT", sql.UsingConverter(jsonConverter)),
 			sql.BINARY:         sql.MapStatic("TEXT", sql.AlsoCompatibleWith("character varying")),
 			sql.MULTIPLE:       sql.MapStatic("OBJECT", sql.UsingConverter(sql.ToJsonBytes)),
-			sql.STRING_INTEGER: sql.MapStatic("NUMERIC(22, 0)", sql.AlsoCompatibleWith("numeric"), sql.AlsoCompatibleWith("integer")),
-			sql.STRING_NUMBER:  sql.MapStatic("NUMERIC(22, 0)", sql.AlsoCompatibleWith("numeric"), sql.AlsoCompatibleWith("integer")),
+			sql.STRING_INTEGER: sql.MapStatic("INTEGER", sql.AlsoCompatibleWith("numeric"), sql.AlsoCompatibleWith("integer")),
```
williamhbaker, line 36: Often these `type: string, format: integer` fields are for very large integers that do not fit into a 64 bit signed int, hence the string encoding. `NUMERIC(22, 0)` would typically make sense if that's the largest exact precision integer representation that's available.

Looking at the docs [here](https://cratedb.com/docs/crate/reference/en/latest/general/ddl/data-types.html#integer)  it seems that `INTEGER` is a 32-bit integer, which probably isn't going to work well for these larger integers.

I'm inclined to think that the type for the `sql.INTEGER` case should actually be `BIGINT`, and this `sql.STRING_INTEGER` should remain a `NUMERIC(22, 0)`.

williamhbaker (commented):
This seems like it will work, since the `Destroy` is a best-effort for the transactor to clean up after itself, and you are also explicitly calling the `removeBinding` at startup in case it wasn't able to clean up successfully (which can happen for a variety of reasons).

We don't do anything exactly like this with any of our existing materializations since they all either support actual session-scoped temporary tables, or they read from objects in cloud storage directly, or in some cases they support transactional DML where we can create & drop table in the same transaction the evaluates the query against it. Doesn't seem like any of those are an option here.

Aside from that, see my comment about data types for consideration.

materialize-cratedb/sqlgen.go
surister: > Often these type: string, format: integer fields are for very large integers that do not fit into a 64 bit signed int, hence the string encoding. NUMERIC(22, 0) would typically make sense if that's the largest exact precision integer representation that's available.

You are right, I was so narrowly focused on making the tests pass to be able to test other issues like temp tables  that now I realize that I modified fixture.json which  is a global test file, not a connector specific one.

> I'm inclined to think that the type for the sql.INTEGER case should actually be BIGINT
If I set `sql.Integer` to `BIGINT` I get:

```
Error: commit failed: store at index 0: error building query 
INSERT INTO doc.unsigned_bigint (
                id,
                flow_published_at,
                unsigned_bigint,
                flow_document
) VALUES ($1, $2, $3, $4);
: failed to encode args[2]: unable to encode 0xffffffffffffffff into binary format for int8 (OID 20): 18446744073709551615 is greater than maximum value for int64
```
Setting `sql.STRING_INTEGER should remain a NUMERIC(22, 0).` makes sense. ✅

If I set back `sql.Integer` to numeric, I still get a similar error:

```
2025-03-04T15:41:01.686877Z  INFO ops: opening database fields={"address":"materialize-cratedb-cratedb-1.flow-test:5432","database":"crate","user":"crate"}
2025-03-04T15:41:05.032153Z  INFO ops: finished waiting for materialization Open fields={"took":"3.345166565s"}
2025-03-04T15:41:06.130638Z ERROR ops: commit failed: store at index 33: ERROR: Cannot detect the type of the value: 18446744073709551615 (SQLSTATE XX000) fields={}
2025-03-04T15:41:06.131138Z ERROR ops: connector failed fields={"module":"connector_init::rpc","status":"exit status: 1"}
2025-03-04T15:41:06.131176Z  WARN ops: connector exited with unconsumed input stream remainder fields={"module":"connector_init::rpc"}
Error: commit failed: store at index 33: ERROR: Cannot detect the type of the value: 18446744073709551615 (SQLSTATE XX000)
connector invocation failed
```

I set everything to numeric(22,0) for testing
```
sql.INTEGER:        sql.MapStatic("NUMERIC(22, 0)", sql.AlsoCompatibleWith("numeric"), sql.AlsoCompatibleWith("integer")),
			sql.NUMBER:         sql.MapStatic("NUMERIC(22, 0)"),
			sql.BOOLEAN:        sql.MapStatic("BOOLEAN"),
			sql.OBJECT:         sql.MapStatic("OBJECT"),
			sql.ARRAY:          sql.MapStatic("TEXT", sql.UsingConverter(jsonConverter)),
			sql.BINARY:         sql.MapStatic("TEXT", sql.AlsoCompatibleWith("character varying")),
			sql.MULTIPLE:       sql.MapStatic("OBJECT", sql.UsingConverter(sql.ToJsonBytes)),
			sql.STRING_INTEGER: sql.MapStatic("NUMERIC(22, 0)", sql.AlsoCompatibleWith("numeric"), sql.AlsoCompatibleWith("integer")),
			sql.STRING_NUMBER:  sql.MapStatic("NUMERIC(22, 0)", sql.AlsoCompatibleWith("numeric"), sql.AlsoCompatibleWith("integer")),
			sql.STRING: sql.MapString(sql.StringMappings{
				Fallback: sql.MapStatic(
					"TEXT",
					sql.AlsoCompatibleWith("character varying"),
					sql.UsingConverter(sql.StringCastConverter(func(in string) (any, error) {
						// Postgres doesn't allow fields with null bytes, so they must be stripped out if
						// present.
						return strings.ReplaceAll(in, "\u0000", ""), nil
					})),
				),
				WithFormat: map[string]sql.MapProjectionFn{
					"date":      sql.MapStatic("TIMESTAMP WITH TIME ZONE", sql.UsingConverter(sql.ClampDate)),
					"date-time": sql.MapStatic("TIMESTAMP WITH TIME ZONE", sql.AlsoCompatibleWith("timestamp with time zone"), sql.UsingConverter(sql.ClampDatetime)),
					"duration":  sql.MapStatic("INTEGER"),
					"ipv4":      sql.MapStatic("IP"),
					"ipv6":      sql.MapStatic("IP"),
					"macaddr":   sql.MapStatic("TEXT"),
					"macaddr8":  sql.MapStatic("TEXT"),
					"time":      sql.MapStatic("TEXT"),
					// UUID format was added on 30-Sept-2024, and pre-existing
					// text type of columns are allowed to validate for
					// compatibility with pre-existing columns.
					"uuid": sql.MapStatic("TEXT"),
				},
			}),
```
and produces the same error.

CrateDB shouldn't have an issue since that number is numeric(20,0)

##### Query ran at 2025-03-04T15:44:15.051Z on CrateDB 5.10.1
```
select
  18446744073709551615 :: numeric(22, 0),
  '18446744073709551615' :: numeric(22, 0)
```
##### 1 record(s) returned in 0.0014s, showing 1 row(s).

|18446744073709551615|18446744073709551615|
 |-|-| 
|18446744073709552000|18446744073709552000|

Any pointers?

williamhbaker: The `unsigned_bigint` test tries to put an integer value that is larger than int64, and is actually the largest uint64 value possible, into a field for `type: integer`. So it does make sense that `sql.INTEGER` being `BIGINT` would fail in that case. We have a way to handle this, check out the `materialize-postgres` code [here](https://github.com/estuary/connectors/blob/8a473972de943a9afc39f918c049d25576f7d3fd/materialize-postgres/sqlgen.go#L15-L18). You can define an alternate column type that is used when the `type: integer` values can be larger than an int64, so it would probably be `NUMERIC(22,0)` in this case.

If you're getting errors with `sql.STRING_NUMBER` being `NUMERIC(22,0)`, you might need to explicitly convert string values to integers. These files can also be like `type: [string, integer], format: integer`, meaning the values can be strings that are encoded integers like `"123"`, or actual integers like `123`. Sometimes we need to convert the string to an integer for the client SDK to work, for example [here](https://github.com/estuary/connectors/blob/8a473972de943a9afc39f918c049d25576f7d3fd/materialize-snowflake/sqlgen.go#L60C3-L61C127) with the use of `sql.UsingConverter`.

surister: I'm still facing issues with the integers, I will drop 1a85056c17f540b62b1b50c44e6bcca006430841 and make the PR only about temp tables, will address this types issues in later PR.

williamhbaker (approved):
LGTM

PR #2480, opened 2025-03-04
source-google-analytics-data-api-native: validate dimensions and metrics for custom reports

Alex-Bair:
**Description:**

Users have quickly run into errors when creating custom reports. We weren't validating the dimensions or metrics specified in custom reports, so the connector was failing after it was published if an invalid dimension/metric was specified. This PR adds validation of all dimensions and metrics specified in any custom reports. The Google API has a handy `/metadata` endpoint that returns all valid dimensions & metrics, so that list is used when validating the user's custom report input.

This validation does _not_ extend to any filters included in custom reports. We need to have a better model for the various types of valid filters before we even think about validating their contents, and developing that model would take a lot more time than I think it's worth investing right now.

**Workflow steps:**

It's probably worthwhile to link Google's [docs](https://developers.google.com/analytics/devguides/reporting/data/v1/api-schema) on the various valid dimensions & metrics in the connector documentation, if only so RunLLM can answer questions about it a little better.

**Notes for reviewers:**

Tested on a local stack. Confirmed invalid dimensions/metrics in custom reports throw a `ValidationError` before publication & provide a clearer message of what caused the failure.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2480)

### DIFF
--- a/source-google-analytics-data-api-native/source_google_analytics_data_api_native/resources.py
+++ b/source-google-analytics-data-api-native/source_google_analytics_data_api_native/resources.py
@@ -15,6 +15,7 @@ from .api import (
     fetch_property_timezone,
     fetch_report,
     backfill_report,
+    fetch_metadata,
 )
 from .models import (
     EndpointConfig,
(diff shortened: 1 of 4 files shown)


source-google-analytics-data-api-native/source_google_analytics_data_api_native/resources.py
```diff
@@ -70,12 +73,33 @@ def validate_custom_reports_json(
...
     for custom_report_details in custom_reports:
         try:
             assert isinstance(custom_report_details, dict)
             model = Report.model_validate(custom_report_details)
             if model.name in default_report_names:
                 errors.append(f'Custom report name "{model.name}" already exists as a default report. Please rename the custom report.')
+
+            for dimension in model.dimensions:
+                if dimension not in valid_dimensions:
+                    errors.append(f'"{dimension}" in report "{model.name}" is not a valid dimension. Consult Google\'s documentation for a list of valid dimensions.')
```
jgraettinger, line 97: nit: can we include a URL here, to make this error even more actionable?

jgraettinger (approved):
LGTM

source-google-analytics-data-api-native/source_google_analytics_data_api_native/resources.py
Alex-Bair: Yep, I'll add it in before merging.

PR #2482, opened 2025-03-04
Phil/ops catalog

psFried:
**Description:**

Rolls up a few changes related to the ops catalog work in estuary/flow#1962:

**Update materialize-sql to make key locations recommended instead of required**

This is to allow taking a collection that's keyed on `[/shard/name, /shard/keyBegin, /shard/rClockBegin]` and materialize it just using `[/shard/name]` as the key. This has been supported by the runtime for a while, but requires the validation constraints to allow the location to be omitted.

**Start emitting some connectorStatus events**

See #2483
This is pretty much just committing the connectorStatus logging that I was using to test during development. The statuses are all super basic, but that seems like a fine enough place to start.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2482)


materialize-sql/type_mapping.go
```diff
@@ -417,8 +417,8 @@ func (constrainter) NewConstraints(p *pf.Projection, deltaUpdates bool) *pm.Resp
 	var constraint = pm.Response_Validated_Constraint{}
 	switch {
 	case p.IsPrimaryKey:
-		constraint.Type = pm.Response_Validated_Constraint_LOCATION_REQUIRED
-		constraint.Reason = "All Locations that are part of the collections key are required"
+		constraint.Type = pm.Response_Validated_Constraint_LOCATION_RECOMMENDED
```
williamhbaker, line 420: I'm concerned that we're going to need to do more work here to make this capability robust. I _think_ we need to make sure at least 1 key is selected if it is standard updates, and we also need to make sure the selected key fields can't be changed...you shouldn't be able to add additional key fields to the materialization after the fact.

The scenario that worries me is somebody not selecting all the keys for a materialization initially, then later adding one of the key fields which breaks their materialization. One issue with that is that all the existing rows will have `null` values for those fields, and generally materializations aren't set up to do comparisons with `null`'s for load queries.

I'm not sure we should release this without doing that additional work up front, since I am quite certain it will be needed as a follow-up anyway.

materialize-sql/driver.go
```diff
@@ -318,6 +318,8 @@ func (d *Driver) NewTransactor(ctx context.Context, open pm.Request_Open, be *bo
 		}
 	}
 
+	log.WithField("eventType", "connectorStatus").Info("Started")
```
williamhbaker, line 321: I think we're going to want this in the [boilerplate file](https://github.com/estuary/connectors/blob/55b604c4d95f7b6baa6cd02ded414a131ae497ab/materialize-boilerplate/boilerplate.go#L165) so that it applies for all materializations, not just the ones using `materialize-sql`.

williamhbaker (commented):
A couple of comments, particularly the one about needing additional Validation smarts to handle key field selection.

materialize-sql/type_mapping.go
williamhbaker: There is some prior art for doing more complex validation checks in this [general area](https://github.com/estuary/connectors/blob/55b604c4d95f7b6baa6cd02ded414a131ae497ab/materialize-boilerplate/validate.go#L122-L308). There we have checks to prevent the original "root document" projection for a standard updates materialization from being changed to some other user-defined root document projection, and also handle cases of "ambiguous fields" (ex: fields that have different capitalization which is fine for Flow, but sometimes a problem for different systems) where you are only allowed to select 1 of the ambiguous fields initially, and after you have done so the others are forbidden. There are some test cases [here](https://github.com/estuary/connectors/blob/main/materialize-boilerplate/validate_test.go) that illustrate these scenarios and others.

Cases like ambiguous fields kind of need to be in the connector code since the definition of ambiguous depends on the specific system being materialized to. Along those lines, I'm not sure it would always be the rule that a standard updates materialization _shouldn't_ be allowed to have its key field selection changed, although that is the case for what we have now. So my inclination would be to keep in the connector code.

psFried: Thanks, I think that gets me on the right track. I'll update and see how it goes.

materialize-boilerplate/validate_test.go
```diff
@@ -194,6 +197,33 @@ func TestValidate(t *testing.T) {
 			fieldNameTransform: simpleTestTransform,
 			maxFieldLength:     20,
 		},
+		{
```
williamhbaker, line 200: For these tests I think you only need the `key-subset-base.flow.proto` file and can get rid of the `-add` and `-remove` ones. The "can materialize a subset of key fields" would stay the same, and then the second test would be more like "can't change selected keys of an existing materialization" and be this:

```
		{
			name:               "cannot add to selected key fields for standard updates",
			deltaUpdates:       false,
			specForInfoSchema:  loadValidateSpec(t, "key-subset-base.flow.proto"),
			existingSpec:       loadValidateSpec(t, "key-subset-base.flow.proto"),
			proposedSpec:       loadValidateSpec(t, "key-subset-base.flow.proto"),
			fieldNameTransform: simpleTestTransform,
			maxFieldLength:     0,
		},
```

It would be most similar to the "update an existing materialization with ambiguous fields" test.

williamhbaker (approved):
LGTM

PR #2486, opened 2025-03-05
source-braintree-native: cover potential gaps after scheduled backfills with the incremental tasks

Alex-Bair:
**Description:**

When a scheduled backfill occurs, it's possible for a "gap" to exist where the old incremental task left off & the new incremental task picks up. For instance, if the incremental task is caught up at 19:58 UTC & is sleeping until it's `interval` has elapsed, and a scheduled backfill re-initializes the capture's state at 20:01 UTC, there's now a gap between 19:58 - 20:01 UTC where the incremental task did not check for newly created data. The backfill will eventually check this gap, but it could take days for the backfill to progress to that point.

This commit shifts the cutoff between backfill & incremental tasks so these gaps are covered by the new incremental tasks instead of the backfill tasks. This should ensure any gaps resulting from scheduled backfills when an incremental task is sleeping are checked soon after re-initialization.

In the future, we could consider preventing the creation of these gaps in the CDK somehow. For example, if there's both a `fetch_pages` and `fetch_changes` function for a binding, only re-initializing the state for `fetch_pages` while maintaining the state for `fetch_changes` would avoid creation of these gaps. I'd like to think about this a bit more before making a CDK change though, so in the interim this connector level change should address the issue. 

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2486)

### DIFF
--- a/source-braintree-native/source_braintree_native/resources.py
+++ b/source-braintree-native/source_braintree_native/resources.py
@@ -52,6 +52,14 @@ INCREMENTAL_RESOURCES: list[tuple[str, IncrementalResourceFetchChangesFn, Increm
     ("subscriptions", fetch_subscriptions, backfill_subscriptions),
 ]
 
+# When a scheduled backfill occurs, it's possible for an incremental task to be sleeping until its interval has elapsed.
+# This can create a gap from where the old incremental task left off & where the new incremental task picks up. While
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

PR #2494, opened 2025-03-06
source-oracle: fix integer key encoding

mdibaiee:
**Description:**

- A customer hit an issue when using integer columns with precision < 18 as primary keys, this is a case that we had not covered with our tests, so I added tests, reproduced the issue and here is the implemented fix. Left comments to explain the various changes.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2494)

### DIFF
--- a/source-oracle/backfill.go
+++ b/source-oracle/backfill.go
@@ -40,7 +40,7 @@ func (db *oracleDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.D
 		query = db.keylessScanQuery(info, schema, table)
 		args = []any{afterRowID}
 
-	case sqlcapture.TableStatePreciseBackfill:
+	case sqlcapture.TableStatePreciseBackfill, sqlcapture.TableStateUnfilteredBackfill:
 		if resumeAfter != nil {
 			var resumeKey, err = sqlcapture.UnpackTuple(resumeAfter, decodeKeyFDB)
 			if err != nil {
@@ -195,7 +195,7 @@ var columnBinaryKeyComparison = map[string]bool{
 // render a "cast" expression for a column so that we can cast it to the
 // type and format we expect for that column (e.g. for timestamps we expect a certain format with UTC timezone, etc.)
 func castColumn(col sqlcapture.ColumnInfo) string {
-	var dataType = col.DataType.(oracleColumnType).original
+	var dataType = col.DataType.(oracleColumnType).Original
 	var isDateTime = dataType == "DATE" || strings.HasPrefix(dataType, "TIMESTAMP")
 	var isInterval = dataType == "INTERVAL"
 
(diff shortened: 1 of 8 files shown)


source-oracle/replication.go
```diff
@@ -551,15 +551,20 @@ func (s *replicationStream) receiveMessages(ctx context.Context, startSCN, endSC
 
 		var msg logminerMessage
 		var ts time.Time
+		var redoSql sql.NullString
```
mdibaiee, line 554: Hit a case where this was null, hence the update

source-oracle/discovery.go
```diff
@@ -268,18 +267,18 @@ SELECT t.owner, t.table_name, t.column_id, t.column_name, t.nullable, t.data_typ
...
-	jsonType  string
-	nullable  bool
+	Original  string
+	Length    int
+	Scale     int16
+	Precision int16
+	T         reflect.Type
+	Format    string
+	JsonType  string
+	Nullable  bool
```
mdibaiee, line 277: Made these public so debug logging will be able to log these values, they are very useful for debugging

source-oracle/discovery.go
```diff
@@ -109,37 +109,36 @@ func translateRecordField(column *sqlcapture.ColumnInfo, val interface{}) (inter
...
-		if dataType.jsonType != "string" {
-			var out = reflect.New(dataType.t).Interface()
-			if err := json.Unmarshal([]byte(v), out); err != nil {
-				return nil, err
-			}
-			return out, nil
+		if dataType.JsonType == "integer" {
+			return strconv.Atoi(v)
+		} else if dataType.JsonType == "number" {
+			return strconv.ParseFloat(v, 64)
```
mdibaiee, line 115: Do a more direct conversion from string to integer or float. The previous approach was more error-prone

source-oracle/backfill.go
```diff
@@ -40,7 +40,7 @@ func (db *oracleDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.D
 		query = db.keylessScanQuery(info, schema, table)
 		args = []any{afterRowID}
 
-	case sqlcapture.TableStatePreciseBackfill:
+	case sqlcapture.TableStatePreciseBackfill, sqlcapture.TableStateUnfilteredBackfill:
```
mdibaiee, line 43: Following source-mysql to share the implementation between these two modes. This is necessary since sometimes `sqlcapture.TableStateUnfilteredBackfill` becomes the default mode for tables and users are confused when this happens and we complain about an invalid mode

source-oracle/main.go
```diff
@@ -316,9 +310,15 @@ func encodeKeyFDB(key any, colType oracleColumnType) (tuple.TupleElement, error)
 	case time.Time:
 		return key.Format(sortableRFC3339Nano), nil
 	case string:
-		if colType.format == "integer" {
+		if colType.JsonType == "integer" {
+			return strconv.Atoi(key)
+		} else if colType.JsonType == "number" {
+			return strconv.ParseFloat(key, 64)
+		}
```
mdibaiee, line 317: Follow the new implementation in `translateRecordField` to be consistent

williamhbaker (approved):
LGTM

PR #2499, opened 2025-03-07
materialize-iceberg: serialize pyspark command inputs to a file

williamhbaker:
**Description:**

The maximum argument length for an EMR job is actually quite limited, around 10k characters, so if the command input for a job gets very long it will fail. This will happen if any number of significant bindings is associated with a transaction or even a single binding with a large number of fields, since all of the fields and their types need to be provided to the script in a serialized form, in addition to the query to execute.

The fix here is to write out the input to a temporary cloud storage file and read that in the PySpark script. Rather than providing the input as an argument, the input is now a URI to the input file.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2499)

### DIFF
--- a/materialize-iceberg/emr.go
+++ b/materialize-iceberg/emr.go
@@ -84,10 +84,10 @@ func ensureEmrSecret(ctx context.Context, client *ssm.Client, parameterName, wan
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


Alex-Bair (approved):
LGTM

PR #2506, opened 2025-03-10
Column Migration: process migrations in batches

mdibaiee:
**Description:**

I have been unable to run tests on all connectors, I think the credentials / env for motherduck and redshift seem to have changed and I don't have credentials for cratedb

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2506)


williamhbaker (commented):
Generally LGTM. The test credentials for Redshift and Motherduck were updated in https://github.com/estuary/connectors/pull/2388, and the working ones are in their respective config files in the integration test folders, [here](https://github.com/estuary/connectors/blob/main/tests/materialize/materialize-redshift/config.yaml) and [here](https://github.com/estuary/connectors/blob/main/tests/materialize/materialize-motherduck/config.yaml). I think these updates will work with them, but it would be good to run the tests and confirm.

williamhbaker (approved):
LGTM for the number migrations.

It would be nice to cover `materialize-azure-fabric-warehouse` for the number migration case as well. It does migrations a little differently since it doesn't support dropping columns, but it might be fairly trivial to add these cases to it if they can use the default CAST syntax.

PR #2514, opened 2025-03-11
materialize-sql: don't use merge bounds for datetimes

williamhbaker:
**Description:**

At least one destination (BigQuery) has known issues with keys that are date-times with sub-microsecond precision when used as a merge bound. For simplicity such formatted strings will never appear in merge bounds, although we should figure out how to do this more selectively in the future.

I suspect that at least Redshift would also fail with a nanosecond timestamp if included in a merge bound query, but haven't tested that directly. These types of keys are pretty rare.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2514)


willdonnelly (approved):
LGTM

PR #2516, opened 2025-03-11
source-google-analytics-data-api-native: add lookback window functionality

Alex-Bair:
**Description:**

Per a user, GA4 data underlying some reports can still change N days after the report was captured. As the connector is right now, the incremental task fetches the most recent day's report & never looks back to recapture previous days reports. Meaning, the older reports can be outdated if the underlying data changed. The value of N seems to vary depending on the type of data, report, ingestion method, etc. based on these [Google docs](https://support.google.com/analytics/answer/11198161).

To fix this, this PR adds a configurable lookback window that forces the incremental task to always fetch reports for the past `lookback_window_size` days whenever it runs.

**Workflow steps:**

The `source-google-analytics-data-api-native` docs should be updated to reflect the new lookback window setting.

**Notes for reviewers:**

Tested on a local stack. Confirmed that the incremental task fetches the report for the past `lookback_window_size` days before the current `log_cursor` datetime on each sweep.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2516)

### DIFF
--- a/source-google-analytics-data-api-native/source_google_analytics_data_api_native/api.py
+++ b/source-google-analytics-data-api-native/source_google_analytics_data_api_native/api.py
@@ -157,13 +157,25 @@ async def fetch_report(
     timezone: ZoneInfo,
     report_doc_model: type[ReportDocument],
     report: Report,
+    start_date: datetime,
+    lookback_window_size: int,
     log: Logger,
     log_cursor: LogCursor,
 ) -> AsyncGenerator[ReportDocument | LogCursor, None]:
     assert isinstance(log_cursor, datetime)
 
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

Alex-Bair:
With the addition of a default lookback window of 30 days, I think it makes sense to increase the resource interval to reduce how much of the API limit/quota is used up without the user overriding any defaults. The user requesting the lookback window functionality mentioned they will likely use a large interval, like `24H`, for this same reason. That might be a little too large for the default case, so I'm going to increase it to `30M` instead.

PR #2519, opened 2025-03-12
source-salesforce-native: new connector

Alex-Bair:
**Description:**

This is an initial version of a native Salesforce capture connector. There are a _lot_ of details I'd like to mention, so this will be a fairly long PR description.

### Overall Strategy

Salesforce's [REST API](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_rest.htm) and [Bulk API 2.0](https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/asynch_api_intro.htm) are used for capturing data. Interacting with each API is abstracted into separate classes, either `RestQueryManager` or `BulkJobManager`.

The [`queryAll`](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_queryall.htm) query option is used to capture soft deletes (i.e. when a record is in Salesforce's Recycle Bin for [~15 days](https://developer.salesforce.com/docs/atlas.en-us.242.0.apexcode.meta/apexcode/langCon_apex_dml_examples_delete.htm#:~:text=Deleted%20records%20aren%E2%80%99t%20deleted%20permanently%20from%20Salesforce%2C%20but%20they%20are%20placed%20in%20the%20Recycle%20Bin%20for%2015%20days%20from%20where%20they%20can%20be%20restored.) before being hard deleted). Salesforce sometimes nulls out fields when a record is soft deleted, and those are reflected in the records we get from a `queryAll`. 

Since the actual pagination of results is abstracted into `RestQueryManager`/`BulkJobManager`, it would be messy to keep checkpointing to only between pages. Instead, I used the "checkpoint after N results have been yielded" strategy, while keeping N close to the max page size for each API.

### Discovery
The complete list of possible bindings the connector could support can be determined by hitting the [`/sobjects` endpoint](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_describeGlobal.htm) and checking the list of objects & their abbreviated metadata. However, to determine if the object has a valid field we can use as a cursor & incrementally replicate it, we would have to hit the [`/sobjects/:sobjectName/describe` endpoint](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_sobject_describe.htm) for each possible object & check the list of available fields. There can be hundreds to thousands of possible objects, and sending a request for each one in order to determine if it has a valid cursor field quickly churns through the API limit. For reference, the ATF connector's discovery is like this, and it uses 800+ requests during discovery for my hardly customized dev Salesforce account. With auto-discovers happening multiple times per day, we need to use some other way to determine if we can incrementally replicate an object.

To avoid those excessive requests during discovery, I compiled a list of supported standard Salesforce objects in `supported_standard_objects.py` that list a cursor field if there should be one present for that object. This means we'll need to maintain this list manually & add new objects as needed, but that cost seems worth it to avoid burning API requests during each discovery. Also, Salesforce says that custom objects always have the `SystemModstamp` field, so we should always be able to incrementally replicate those.

We _will_ still need to hit the `/sobjects/:sobjectName/describe` endpoint to fetch the complete list of fields to use when querying the APIs. The connector does this during `validate` and `open`, and only for enabled bindings (aligning with the strategy implemented for `source-oracle-flashback` in https://github.com/estuary/connectors/pull/2013).

Similar to the ATF `source-salesforce` connector, only a small subset of the available bindings are enabled by default (approx. 17). This keeps the number of enabled bindings to a reasonable number in the garden path where users don't do much configuration.

#### Bulk API 2.0
Some objects and fields (ex: non-scalar fields) are not able to be queried via the Bulk API 2.0. When that is the case, we attempt to query via the REST API instead.

Results fetched via the Bulk API 2.0 are returned in a CSV file, and field values are always read as strings. The connector attempts to use Salesforce's reported `soapType` for the field to coerce it into the appropriate type before yielding the document. However, we have observed that custom fields don't always adhere to the `soapType` Salesforce says it should be, so if coercion fails, the field is left as a string & we'll rely on schema inference to do it's job. Along the same lines, if the CSV-read value is an empty string, the connector assumes it's actually null in Salesforce & converts it to `None`.

The `aiocsv` package is leveraged to asynchronously read CSV results files. While this capability may be useful for future connectors, handling all the variations of CSVs (different delimiters, line endings, quotes, etc.) beyond what Salesforce uses would be additional scope I didn't want to tackle right now. If this capability is needed later for different connectors, we could investigate improving the flexibility of the `_process_csv_lines` and move it into the CDK.

#### REST API
When querying via the REST API, the fields are explicitly specified in the query params. The alternative is to use the [`FIELDS(ALL)`](https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_select_fields.htm#:~:text=The%20API%20distinguishes,and%20unbounded%20queries%3A) syntax to avoid needing to explicitly specify fields, but that requires using a page size of 200 with `LIMIT 200`. By explicitly specifying fields, the maximum page size is 2,000.

As a result of explicitly specifying fields in the URL, we have to ensure the combined length of the URI and headers stays below Salesforce's limit of [16,384 bytes](https://developer.salesforce.com/docs/atlas.en-us.salesforce_app_limits_cheatsheet.meta/salesforce_app_limits_cheatsheet/salesforce_app_limits_platform_api.htm#:~:text=In%20each%20REST%20call%2C%20the%20maximum%20length%20for%20the%20combined%20URI%20and%20headers%20is%2016%2C384%20bytes.). To handle this, we chunk fields across multiple API requests & merge them together using a strategy very similar to the one in `source-hubspot-native`. When there are not many custom fields, we shouldn't need to chunk fields. Doing some rough math considering the [maximum number of custom fields](https://help.salesforce.com/s/articleView?id=platform.custom_field_allocations.htm&type=5) allowed per object, the field chunking strategy should still be more API request-efficient than using the `FIELDS(ALL)` syntax and a 10x smaller page size.

Records returned by the REST API are slightly different from those returned by the Bulk API 2.0. Datetimes from the REST API are returned in the format `2024-09-24T03:26:07.000+0000` where the Bulk API 2.0 returns them as `2024-09-24T03:26:07.000Z`. The connector attempts to convert datetimes to the format returned by the Bulk API (with the escape hatch if the conversion fails since custom fields aren't always the type Salesforce says they are). The REST API also automatically includes additional metadata for each record:
```json
      "attributes": {
        "type": "SetupAuditTrail",
        "url": "/services/data/v62.0/sobjects/SetupAuditTrail/{record_id}"
      }
```
This metadata can be derived from the binding name & record `Id`, so I opted to remove it to stay consistent with what's returned by the Bulk API 2.0.

### Formula Fields
Salesforce has a concept called [formula fields](https://help.salesforce.com/s/articleView?id=platform.customize_formuladef.htm&type=5), which are fields calculated based on some other values. They are calculated at query time, and changes to formula fields do not cause any of the possible cursor fields to update per Salesforce's [docs](https://help.salesforce.com/s/articleView?id=000382843&type=1#:~:text=This%20is%20because%20a%20formula%20field%20is%20a%20way%20of%20arriving%20at%20a%20value%2C%20but%20does%20not%20actually%20store%20the%20value%20and%20will%20not%20update%20the%20last%20modified%20timestamp%20on%20a%20record%20when%20it%20changes.). This makes it possible for us to capture a record, and then it's formula fields are updated without any change in the cursor field, causing the previously captured record to have outdated formula field values.

To address this issue, the connector initiates a backfill per some schedule & yield the most recent formula field values for previously captured records. A top level [`merge` reduction strategy](https://docs.estuary.dev/reference/reduction-strategies/merge/) is used to combined these more recent formula field values and the associated previously yielded record.

### Assumptions
- All full refresh bindings are able to be queried via the Bulk API 2.0. Since these objects do not have any valid cursor fields, fetching them via the REST API is tricky - we can't be sure that a record wasn't updated between queries for separate field chunks. We could decide that's ok (we'll eventually snapshot all records again on the next sweep), but I figured we could wait and see if this is an actual issue before adding this capability. There are relatively few Salesforce objects that don't have a valid cursor field, so I feel comfortable deferring this decision.

**Workflow steps:**

Documentation should be created for the connector. In particular, we've done a lot of troubleshooting around the following with the ATF connector & it would be beneficial to explain these behaviors in the docs:
- Salesforce null-ing some fields for soft deleted records
- [Field level security / object permissions](https://help.salesforce.com/s/articleView?id=platform.users_profiles_field_perms.htm&language=en_US&type=5) preventing the connector from discovering certain fields/objects

**Notes for reviewers:**

~~This PR is still in development. Most of the core connector is complete, but there's still scope that needs completed before the PR is ready for review & could be merged. Remaining scope is all around supporting periodically refreshing formula fields per some schedule (i.e. running a connector-initiated backfill, yielding records that only contain formula fields and using a top-level `merge` reduction strategy). The remaining scope I anticipate to support this is:~~
- [x] Update the CDK to allow setting a top level reduction strategy on resource schemas, similar to how we [set `x-infer-schema`](https://github.com/estuary/connectors/blob/ac3166a030628ad6a1a52a133d0711f436444f98/estuary-cdk/estuary_cdk/capture/common.py#L334-L358). https://github.com/estuary/connectors/pull/2526
- [x] Update the [scheduled backfill functionality](https://github.com/estuary/connectors/blob/ac3166a030628ad6a1a52a133d0711f436444f98/estuary-cdk/estuary_cdk/capture/common.py#L438-L499) in the CDK so ongoing backfills take precedence over scheduled backfills. https://github.com/estuary/connectors/pull/2526
- [x] Update the CDK to support a different backfill (i.e. `fetch_page` function) that executes different logic if the backfill was connector-initiated or not. https://github.com/estuary/connectors/pull/2526
- [x] Update `source-salesforce-native` to use the different `fetch_page` function to only fetch formula fields for connector-initiated backfills, and to do connector-initiated backfills on a schedule. Also, update the connector to use a top level `merge` reduction strategy.
- [x] Perform final manual tests with a Salesforce account containing a significant amount of data.

Tested on a local stack with a couple Salesforce accounts. Confirmed:
- The connector works with both sandbox and production Salesforce accounts.
- OAuth works for authentication with an OAuth app I created in my dev Salesforce account. Using our prod OAuth app credentials, the connector's OAuth process succeeds up to the expected "invalid redirect uri" error (i.e. `http:localhost:3000` isn't a valid redirect URI for our prod app).
- Documents yielded between the REST and Bulk APIs are identical in terms of field types & datetime precisions.
- Results from the REST and Bulk APIs are received in ascending order of the cursor field, and the connector checkpoints after approximately every page of results.
- The connector does not try to query the Bulk API if it knows the querying the object isn't supported by the Bulk API.
- Custom objects and custom fields are captured fine, like standard objects and fields.
- Reading from a collection & materializing into a dockerized PostgreSQL instance works.
- If an object has formula fields and there's a schedule set, the formula fields are backfilled per the schedule & documents containing just the primary key, cursor field, and formula fields are yielded. Materializing into PostgreSQL results in a single row per primary key with the most recently captured formula field values represented in the row.
- If an object does not have formula fields and there's schedule set, the `fetch_page` function returns early before sending any API requests.
- Cached access tokens are reused until it's almost expired. Effectively, we get a new access token every 45 minutes instead of for every request.
- Discoveries do not send API requests to fetch objects' fields. Validating and opening bindings use a scatter-gather technique to fetch objects' fields & only do so for enabled bindings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2519)


williamhbaker:
Only just started looking at this and will give it a high-level go through later, but one initial thing I'd suggest is pairing down the recovery snapshot somehow. Either omit it completely or post-process it so that it is quite a bit less than 45k lines. Maybe prune out all but the bindings that are enabled by default (if that's still a thing)?

Alex-Bair:
Yep, we still only enable a handful (~19) of the possible bindings by default. I'll update the PR description to mention that we disable most of the bindings by default, and I'll cut down how many bindings are included in the discover snapshot so it's not as large.

source-salesforce-native/source_salesforce_native/api.py
```diff
@@ -0,0 +1,173 @@
...
+
+        async for record_or_dt in _execution_wrapper(gen, cursor_field, start, end, BULK_CHECKPOINT_INTERVAL):
+            if isinstance(record_or_dt, datetime):
+                yield dt_to_str(record_or_dt)
+            else:
+                yield SalesforceResource.model_validate(record_or_dt)
+
+    except BulkJobError as err:
+        # If this object can't be queried via the Bulk API, we use the REST API.
+        if err.errors and (CANNOT_FETCH_COMPOUND_DATA in err.errors or NOT_SUPPORTED_BY_BULK_API in err.errors):
```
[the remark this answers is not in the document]
Alex-Bair: Yep. For the objects that plain can't be queried via the Bulk API, I can add another property to the associated dict in `SUPPORTED_STANDARD_OBJECTS` to indicate that. For everything else, I _should_ be able to figure it out from their fields' `soapType`s (ex: if there's a `urn:address` type field, we can't use the Bulk API). I'll try that out.

source-salesforce-native/source_salesforce_native/bulk_job_manager.py
```diff
@@ -0,0 +1,258 @@
...
+            await self.http.request(self.log, url)
+        )
+
+        return response
+
+
+    async def _process_csv_lines(
+        self,
+        line_generator: AsyncGenerator[bytes, None],
+    ) -> AsyncGenerator[dict[str, str], None]:
```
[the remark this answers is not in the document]
Alex-Bair: Hmmm, if I have `_process_csv_lines` parse models, that'd shove a lot of the transformation logic into it since I'd need to do the transformations before model validation. I agree with the sentiment of passing the intended class into the bulk job & rest query managers. What do you think about doing the validation within each manager right before yielding from `execute`? Then downstream handling will know what's yielded is always an validated instance of the passed in model & it doesn't matter as much when the model validation happens within the managers' logic.
https://github.com/estuary/connectors/blob/c9d1b1d51359be19d51036329d72c2e28287991f/source-salesforce-native/source_salesforce_native/bulk_job_manager.py#L258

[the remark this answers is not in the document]
williamhbaker: I see what you mean about needing to do the CSV transformation. I think you can just keep it like it is actually.

Alex-Bair:
I've added commits that start with "fixup" that address the comments so far. Before merging, I'll rebase all the "fixup" commits into the main "new connector" commit, but they're left alone now to make reviewing easier.

Alex-Bair:
After testing out those "fixup" changes a bit more & testing with multiple bindings enabled, I found room for improvement. They're mostly around speeding up the resource building process & re-using the cached access token we receive after exchanging the refresh token.

Alex-Bair:
I've added support for scheduled formula field backfills. I'll update the PR description to reflect this feature soon, but otherwise it's ready for review.

source-salesforce-native/source_salesforce_native/rest_query_manager.py
```diff
@@ -0,0 +1,154 @@
...
+        mandatory_fields_length = sum([len(field) for field in mandatory_fields])
+
+        field_names = [f for f in list(fields.keys()) if f not in mandatory_fields]
+        chunks: list[list[str]] = []
+
+        chunk_fields_length = mandatory_fields_length
+        chunk: list[str] = [*mandatory_fields]
+        for field in field_names:
+            if (
+                len(chunk) > 0 and
```
williamhbaker, line 84: The `len(chunk) > 0` seems unnecessary here, since `chunk` always has at least the two mandatory fields.

source-salesforce-native/source_salesforce_native/rest_query_manager.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+    async def execute(
+        self,
+        object_name: str,
+        fields: FieldDetailsDict,
+        cursor_field: CursorFields,
+        start: datetime,
+        end: datetime,
+    ) -> AsyncGenerator[dict[str, Any], None]:
```
williamhbaker, line 108: This doesn't really need to be a generator since it is collecting all the results into a list, but I think it's written this way for symmetry with the bulk job manager and the common handling of those two in the execution wrapper.

source-salesforce-native/source_salesforce_native/bulk_job_manager.py
```diff
@@ -0,0 +1,265 @@
...
+    async def _fetch_results(self, job_id: str) -> AsyncGenerator[dict[str, str], None]:
+        url = f"{self.base_url}/{job_id}/results"
+        request_headers = {"Accept-Encoding": "gzip"}
+        params: dict[str, str] = {}
+
+        while True:
+            headers, body = await self.http.request_stream(self.log, url, params=params, headers=request_headers)
+            count = headers.get('Sforce-NumberOfRecords')
+
+            if count is None or count == 0:
```
williamhbaker, line 130: Since we've got a count of records here, it would be nice to sanity check that the number of records we read matches up. A cute way to do it would be to decrement `count` for every yielded `record`, and then make sure it is 0 at the end.

source-salesforce-native/source_salesforce_native/bulk_job_manager.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def _transform_value(
+        self,
+        name: str,
+        soap_type: SoapTypes,
+        value: str,
+    ) -> Any:
+        try:
+            match soap_type:
+                case SoapTypes.ID | SoapTypes.STRING | SoapTypes.DATE | SoapTypes.DATETIME | SoapTypes.TIME | SoapTypes.BASE64:
```
williamhbaker, line 179: Do we ever get timestamps that are almost valid RFC3339 timestamps, but not quite?

We have some handling for these in our ATF hubspot connector actually, see [here](https://github.com/estuary/airbyte/blob/03bb0496d9345c1396b20f36e953eb3011f622e4/airbyte-to-flow/src/interceptors/normalize.rs#L250-L281) for the test cases that describe them. Particularly for a string like `2023-01-30 02:34:15` where we just need to add the `T` and `Z`, and for ISO8601 timestamps like `2020-03-25T21:03:18.000+0000` which we'd like to be `2020-03-25T21:03:18Z`.

There may also be something to be said about a datetime with 0 time part in a `SoapTypes.DATE` that we coudl normalize to be a valid `format: date`, and similarly normalize a date in `SoapTypes.DATETIME` as a timestamp with a 0 HH:MM:SS component.

source-salesforce-native/source_salesforce_native/bulk_job_manager.py
```diff
@@ -0,0 +1,265 @@
...
+from .models import (
+    SoapTypes,
+    FieldDetailsDict,
+    CursorFields,
+    BulkJobStates,
+    BulkJobSubmitResponse,
+    BulkJobCheckStatusResponse,
+)
+
+INITIAL_SLEEP = 5
```
williamhbaker, line 20: This seems a little conservative...unless you have experience suggesting otherwise. How often do jobs usually take to complete? I might suggest an initial value of more like 1 second here.

source-salesforce-native/source_salesforce_native/resources.py
```diff
@@ -0,0 +1,319 @@
...
+    rest_query_manager = RestQueryManager(http, log, config.credentials.instance_url)
+
+    # If we concurrently send multiple requests that exchange the same refresh token for an access token,
+    # some of those requests intermittently fail. 
+    # https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oauth_refresh_token_flow.htm&type=5#:~:text=Avoid%20sending%20simultaneous%20requests%20that%20contain%20the%20same%20refresh%20token.%20If%20your%20client%20sends%20identical%20requests%20at%20the%20same%20time%2C%20some%20of%20the%20requests%20fail%20intermittently%20and%20the%20Status%20column%20in%20the%20Login%20History%20displays%20Failed%3A%20Token%20request%20is%20already%20being%20processed.
+    # To avoid this, we make a noop request to set the token_source's access token before using the scatter-gather
+    # technique to make multiple requests concurrently. This prevents the first BUILD_RESOURCE_SEMAPHORE_LIMIT
+    # requests from all exchanging the same access token and encountering that intermittent error.
+    await _fetch_queryable_objects(log, http, config.credentials.instance_url)
```
williamhbaker, line 283: I remember looking at this a little bit in the CDK HTTP module. We should probably add some synchronization there. Doesn't have to be right now though, since what you have here should work.

williamhbaker (approved):
LGTM, with some non-blocking comments

source-salesforce-native/tests/snapshots/snapshots__spec__capture.stdout.json
```diff
@@ -0,0 +1,146 @@
...
+          ],
+          "title": "OAuth",
+          "type": "object",
+          "x-oauth2-provider": "salesforce"
+        }
+      },
+      "properties": {
+        "start_date": {
+          "default": "1700-01-01T00:00:00Z",
+          "description": "UTC data and time in the format YYYY-MM-DDTHH:MM:SSZ. Any data generated before this date will not be replicated. If left blank, all data will be replicated.",
```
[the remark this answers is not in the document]
Alex-Bair: :facepalm: Thanks for catching that, I'll get it fixed before merging.

source-salesforce-native/tests/snapshots/snapshots__spec__capture.stdout.json
```diff
@@ -0,0 +1,146 @@
...
+        },
+        "interval": {
+          "default": "PT0S",
+          "description": "Interval between updates for this resource",
+          "format": "duration",
+          "title": "Interval",
+          "type": "string"
+        },
+        "schedule": {
+          "default": "",
```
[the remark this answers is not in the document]
Alex-Bair: Yes, an empty string makes it never refresh. I tried using `None` when I introduced the feature originally, but it was rendered in the UI as something like "<object Object>", so I went with an empty string instead since the UI showed it as an empty input field.

source-salesforce-native/source_salesforce_native/rest_query_manager.py
Alex-Bair: Ah, another good catch. That condition was a relic before I had the mandatory fields included in each chunk, I'll remove it.

source-salesforce-native/source_salesforce_native/rest_query_manager.py
Alex-Bair: Yep, that's right. I didn't necessarily consider that `RestQueryManager.execute` could return a list of all results, but I did want to keep how we receive results from both the rest query manager and bulk job manager as alike as possible so it's easier to swap between them. For example, when we fallback to backfilling with the REST API if the Bulk API says it doesn't support querying an object. This could also make it easier for `fetch_incremental_resources` to swap to the Bulk API if it falls far enough behind (if we want to have that behavior in the future).

source-salesforce-native/source_salesforce_native/bulk_job_manager.py
Alex-Bair: Good idea, will do.

For my own selfish future reference, the `Sforce-NumberOfRecords` header is the number of records in the current set/page, _not_ the total number of records for the bulk job. There's also `numberRecordsProcessed` included in the response that returns the current job status, and I suspect that represents the total number of results we should expect for the overall bulk job; I'll look into that & consider doing a similar count check.

source-salesforce-native/source_salesforce_native/bulk_job_manager.py
Alex-Bair: You're right, it's too conservative. Testing fetching all non-complex fields for 91k contacts, it took ~7 seconds. But the connector would break those into individual date windows, and some could be empty if users don't provide a start date & choose to replicate everything. I'll trim it down to 1 second or maybe a little smaller.

williamhbaker: Even something like 200ms wouldn't be too aggressive I'd say!

source-salesforce-native/source_salesforce_native/bulk_job_manager.py
Alex-Bair: We do sometimes get timestamps that aren't quite RFC3339 compliant. 

The REST API returns datetimes as `2022-09-13T02:18:00.000+0000` (no colon in the offset). They're always UTC per Salesforce's docs, so I used `dt_to_str(str_to_dt(ts))` to normalize them to be like `2022-09-13T02:18:00.000Z`.

The Bulk API returns datetimes as `2022-09-13T02:18:00.000Z`, so they're compliant.

Both APIs return dates as `2022-09-13`.

> There may also be something to be said about a datetime with 0 time part in a SoapTypes.DATE that we coudl normalize to be a valid format: date, and similarly normalize a date in SoapTypes.DATETIME as a timestamp with a 0 HH:MM:SS component.

I think I'm following. Are you suggesting I add that normalization? I haven't seen evidence of datetimes in `SoapTypes.DATE` fields or dates in `SoapTypes.DATETIME` fields with the testing I've done so far, but I know Salesforce sometimes lies about the types of custom fields so I wouldn't be surprised if that occurs for standard fields too.

williamhbaker: It's probably not worth worrying about too much unless we see it in practice. Sounds like you have the common cases handled!

Alex-Bair:
Support for surfacing response headers to connectors was reverted by https://github.com/estuary/connectors/pull/2538. `source-salesforce-native` needs to be able to read response header values when paging through Bulk API results, so that functionality will need re-implemented before merging this PR.

source-salesforce-native/source_salesforce_native/bulk_job_manager.py
Alex-Bair: I was able to use the `Sforce-NumberOfRecords` header value to do a per-set/page sanity check & the `numberRecordsProcessed` response field to do a per job check. I wanted the count mismatch errors to include the expected & received number of records, so I ended up not doing the cute `count` decrement method since keeping track of separate `expected` and `received` was a little clearer & fed into the error messages nicely.

Alex-Bair:
I noticed that some of the current `source-salesforce` tasks have bindings enabled to capture custom object histories. Custom object histories always end with `__History` and have a `CreatedDate` field, so they can always be replicated incrementally. I added support for these.

There are also custom object feeds (end in `__Feed`) and custom object metadata (end in `__mdt`), but it's not clear to me if these always have a cursor field available. I opted to not try to support these now, and we can add support later if anyone asks for them.

PR #2526, opened 2025-03-13
estuary-cdk: improve scheduled re-initialization behavior & extend FetchPageFn type

Alex-Bair:
**Description:**

This PR improves the scheduled, connector-initiated backfill behavior introduced in https://github.com/estuary/connectors/pull/2437 and adds the capability to detect within a `fetch_page` function if an invocation was due to a connector-initiated backfill. This behavior is necessary for an intended feature for https://github.com/estuary/connectors/pull/2519 where the behavior of the `fetch_page` function needs differ between the initial backfill & recurring connector-initiated backfills.

PR scope includes:
- Support setting a top level reduction strategy for a collection's write schema.
- Make ongoing backfills take precedence over any scheduled, connector-initiated backfills.
  - If an ongoing backfill completes after a scheduled backfill, the connector is stopped. Otherwise, if an ongoing backfill completes before a scheduled backfill, a scheduled stop coroutine is set.
- Use the cursor value of the existing incremental state for the `fetch_changes` function as the cutoff of the re-initialized backfill state. This only occurs when there's a single `fetch_page` function, a single `fetch_changes` function, and when the both the cutoff and cursor are `datetime`s.
- Introduce an new type of `fetch_page` function - `RecurringFetchPageFn` - that has a boolean parameter `is_connector_initiated` to indicate whether or not the current `fetch_page` function invocation was connector initiated.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack with `source-braintree-native` (our only connector currently using the scheduled, connector-initiated backfill feature). Confirmed:
- Ongoing backfills take precedence over scheduled backfills. State for ongoing backfills are not re-initialized if the connector should backfill according to its schedule.
- Backfills that completed before the next scheduled backfill set a coroutine to stop the connector in time for the next scheduled backfill.
- Backfills that complete after the next scheduled backfill immediately set `stopping`. This is done to try and keep as close adherence to the schedule as possible.
- When there's a single `fetch_page` function, a single `fetch_changes` functiopn, and both the backfill state's cutoff and the incremental state's cursor are datetimes, the re-initialized backfill state's cutoff is set to the current incremental state's cursor.
- The top level reduction strategy for a collection's write schema can be set.
- Using a single `fetch_page` function that matches the `RecurringFetchPageFn` type will be called with the boolean `is_connector_initiated` argument.
  - `is_connector_initiated` correctly reflects whether or not the current backfill was initiated by the connector.

Also tested on a local stack with `source-zendesk-support-native` (a native connector that doesn't use a schedule) and `source-zendesk-support` (an imported connector). Confirmed that backfills for both progressed & completed like they did before these changes.


estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -292,6 +338,26 @@ class AssociatedDocument(Generic[_BaseDocument]):
 """
 
 
+def is_recurring_fetch_page_fn(fn: FetchPageFn | RecurringFetchPageFn, log: Logger, page: PageCursor, cutoff: LogCursor, is_connector_initiated: bool) -> bool:
+    """Check if the function signature accepts the arguments of a RecurringFetchPageFn."""
+    try:
+        inspect.signature(fn).bind(log, page, cutoff, is_connector_initiated)
+        return True
+    except TypeError:
+        return False
```
Alex-Bair, line 347: I'm not fully happy with this approach to differentiate between `FetchPageFn` and `RecurringFetchPageFn`, but I couldn't think of anything better right now. The approach is based on the resolution to this [thread](https://discuss.python.org/t/check-if-function-is-callable-with-given-args-and-kwargs/18820/7) for a similar issue.

The usual `isinstance` check won't work since `FetchPageFn` and `RecurringFetchPageFn` are type annotations, not concrete classes. There are lengthier checks (is `fn` callable, is it an async generator, does it accept 4 arguments, are those arguments the appropriate type, etc.), but those checks don't seem to add more value than using `bind` here.

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -521,25 +522,18 @@ async def _run(task: Task):
...
-                return None
-
-            sleep_duration = future_dt - datetime.now(tz=UTC)
-            await asyncio.sleep(sleep_duration.total_seconds())
-            task.stopping.event.set()
 
         # Gracefully exit to ensure relatively close adherence to any bindings'
         # re-initialization schedules.
-        asyncio.create_task(scheduled_stop(soonest_future_scheduled_initialization))
+        asyncio.create_task(scheduled_stop(task, soonest_future_scheduled_initialization))
```
williamhbaker, line 596: I'm not totally following how this ends up not scheduling the connector to restart if there is an on-going non-connector-initiated backfill. I think it has something to do with `soonest_future_scheduled_initialization` not being set under some circumstance, but I haven't been able to totally figure it out. Can you explain this a little more to help me understand?

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -472,35 +481,27 @@ async def _run(task: Task):
                         cron_schedule, state.last_initialized
                     )
 
-                    if (
-                        next_scheduled_initialization
-                        and next_scheduled_initialization < datetime.now(tz=UTC)
```
williamhbaker, line 462: I don't think it is an immediate cause for concern, but there are several places in this logic where `datetime.now` is invoked. It may be better to get the "now time" once at the top, and then use that same time everywhere that `datetime.now` is being called.

williamhbaker (commented):
A couple of comments

estuary-cdk/estuary_cdk/capture/common.py
Alex-Bair: Yep, I'll also restructure the conditionals to make it clearer too.

`soonest_future_scheduled_initialization` starts as `None`. As we iterate over each binding, if the binding has a schedule (its resource config is a `ResourceConfigWithSchedule` and if there's no ongoing backfill (`state.backfill` is falsey), then we determine when the next scheduled backfill should happen & set/update `soonest_future_scheduled_initialization`. If all bindings have an ongoing backfill, then `soonest_future_scheduled_initialization` is never set & `scheduled_stop` doesn't set a stop.

Is this rewrite clearer about what's happening?
```python
soonest_future_scheduled_initialization: datetime | None = None
now = datetime.now(tz=UTC)

for index, (binding, resource) in enumerate(resolved_bindings):
    state: _ResourceState | None = open.state.bindingStateV1.get(
        binding.stateKey
    )

    should_initialize = state is None or binding.stateKey in backfill_requests
    is_connector_initiated = False

    if state:
        if state.last_initialized is None:
            state.last_initialized = now
            task.checkpoint(
                ConnectorState(bindingStateV1={binding.stateKey: state})
            )

        if not state.backfill and isinstance(binding.resourceConfig, ResourceConfigWithSchedule):
            cron_schedule = binding.resourceConfig.schedule
            next_scheduled_initialization = next_fire(
                cron_schedule, state.last_initialized
            )

            if next_scheduled_initialization:
                if next_scheduled_initialization < now:
                    should_initialize = True
                    is_connector_initiated = True
                    next_scheduled_initialization = next_fire(cron_schedule, now)
                elif not soonest_future_scheduled_initialization:
                    soonest_future_scheduled_initialization = next_scheduled_initialization
                else:
                    soonest_future_scheduled_initialization = min(soonest_future_scheduled_initialization, next_scheduled_initialization)

# Skip ahead in open method to scheduled stop logic

if soonest_future_scheduled_initialization:
    # Gracefully exit to ensure relatively close adherence to any bindings'
    # re-initialization schedules.
    asyncio.create_task(scheduled_stop(task, soonest_future_scheduled_initialization))
```

estuary-cdk/estuary_cdk/capture/common.py
Alex-Bair: Good point, I'll do that.

estuary-cdk/estuary_cdk/capture/common.py
Alex-Bair: > I'm not totally following how this ends up not scheduling the connector to restart if there is an on-going non-connector-initiated backfill.

And also to clarify, a connector restart is scheduled if _any_ of the bindings do not have an on-going backfill (it doesn't matter whether it's connector initiated or not) & have a scheduled backfill in the next 24 hours. If one binding is backfilling, but another binding isn't backfilling & has a scheduled backfill in the next 24 hours, a scheduled restart is set.

Alex-Bair: I was reviewing this a little more, and I think this is actually clearer. It separates "a missed scheduled backfill" and "a future scheduled backfill in the next 24 hours" into separate variables `missed_scheduled_initialization` and `future_scheduled_initialization`.
```python
soonest_future_scheduled_initialization: datetime | None = None
NOW = datetime.now(tz=UTC)

for index, (binding, resource) in enumerate(resolved_bindings):
    state: _ResourceState | None = open.state.bindingStateV1.get(
        binding.stateKey
    )

    should_initialize = state is None or binding.stateKey in backfill_requests
    is_connector_initiated = False

    if state:
        if state.last_initialized is None:
            state.last_initialized = NOW
            task.checkpoint(
                ConnectorState(bindingStateV1={binding.stateKey: state})
            )

        if not state.backfill and isinstance(binding.resourceConfig, ResourceConfigWithSchedule):
            cron_schedule = binding.resourceConfig.schedule
            missed_scheduled_initialization = next_fire(
                cron_schedule, state.last_initialized, NOW
            )
            future_scheduled_initialization = next_fire(
                cron_schedule, NOW,
            )

            if missed_scheduled_initialization:
                should_initialize = True
                is_connector_initiated = True

            if future_scheduled_initialization:
                if not soonest_future_scheduled_initialization:
                    soonest_future_scheduled_initialization = future_scheduled_initialization
                else:
                    soonest_future_scheduled_initialization = min(soonest_future_scheduled_initialization, future_scheduled_initialization)
```

Sorry for the back & forth on the refactoring here.

williamhbaker: That seems pretty good!

Alex-Bair: Sweet! I also updated the similar schedule logic in `_binding_backfill_task` to align with the "missed" and "future" scheduled initialization distinctions

williamhbaker (approved):
LGTM

I don't have much in the way of other good ideas for the `RecurringFetchPageFn` either. It could make sense for it to be a separate input entirely to `open_binding` rather than a different kind of `fetch_page`, but I don't think one way is obviously better than another.

Alex-Bair:
For my own sanity & documentation: I confirmed that the scheduled backfill behavior after the refactor is still the same as before refactoring. I plan on merging tomorrow morning.

PR #2536, opened 2025-03-14
source-http-ingest: support okta verification challenge

### LINKED ISSUE #2433: source-http-ingest: support Okta "verification" challenge
Okta has this weird "verification" step that they do before they'll start sending webhooks. The process is described [here](https://developer.okta.com/docs/concepts/event-hooks/#one-time-verification-request), and there's some example code [here](https://developer.okta.com/docs/guides/event-hook-implementation/nodejs/main/#configure-initial-event-hook-verification). The TLDR is that they send a GET request with an `x-okta-verification-challenge` header, and we just need to echo back that header value as part of a JSON response. I can't imagine what's actually being verified with this, apart from the fact that the endpoint was specifically intending to support Okta webhooks. But it seems to be required, and is probably the easiest way for us to support capturing data from Okta.

I'm thinking that the easiest thing would be to just automatically create routes for all bindings that handle these GET requests. They'd be harmless in other usages of the connector, so I see no reason to conditionally enable them or add new endpoint configuration.

psFried:
**Description:**

Resolves #2433
Okta requires the server to respond to a GET request with a special header value before it will start sending webhooks. This implements the GET request handler.

**Workflow steps:**

This is enabled universally, since I saw no downside to doing so. So this should allow the connector to just start accepting Okta webhooks, without requiring any extra steps.

I tested this by deploying the PR build of the connector, and having the user configure Okta to start sending webhooks. They were able to successfully verify the endpoint and capture data.

I also manually tested using:

```
curl -k 'https://ca1f4fbcdaac1efc-8080.reactor.flow.localhost:9000/webhook-data' -H 'accept: application/json' -H 'x-okta-verification-challenge: canary'
{"verification":"canary"}
```

and also for a route using path parameters `/foo/{foo}/bar/{bar}`:

```
curl -k 'https://ca1f4fbcdaac1efc-8080.reactor.flow.localhost:9000/foo/fid/bar/bid' -H 'accept: application/json' -H 'x-okta-verification-challenge: canary'
{"verification":"canary"}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2536)

### DIFF
--- a/source-http-ingest/src/server.rs
+++ b/source-http-ingest/src/server.rs
@@ -71,7 +71,9 @@ pub async fn run_server(
         let openapi_path = binding.url_path();
         let axum_path = openapi_to_axum_path(&openapi_path);
         tracing::info!(path = %openapi_path, %axum_path, "configuring handler for route");
-        router = router.route(&axum_path, routing::post(handle_webhook));
+        router = router
+            .route(&axum_path, routing::post(handle_webhook))
+            .route(&axum_path, routing::get(handle_get));
     }
 
     // Set the body limit to be the same as the max document size allowed by Flow (64MiB)
@@ -125,6 +127,18 @@ pub async fn run_server(
         .await
         .context("running server")?;
     Ok(())
+}
+
+/// Handles GET requests, in order to satisfy Okta's requirements for webhooks.
+/// See issue: https://github.com/estuary/connectors/issues/2433
+/// The TLDR is that we need to take the value of the `x-okta-verification`
+/// header and return it in a JSON response. I have no idea why they do this.
+async fn handle_get(
+    OktaVerificationHeader(okta_header): OktaVerificationHeader,
+    path: axum::extract::MatchedPath,
+) -> axum::Json<Value> {
+    tracing::info!(path = %path.as_str(), %okta_header, "handling Okta verification challenge");
+    axum::Json(serde_json::json!({"verification": okta_header}))
 }
 
 async fn handle_webhook(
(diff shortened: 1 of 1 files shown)


source-http-ingest/src/server.rs
```diff
@@ -127,6 +129,18 @@ pub async fn run_server(
     Ok(())
 }
 
+/// Handles GET requests, in order to satisfy Okta's requirements for webhooks.
+/// See issue: https://github.com/estuary/connectors/issues/2433
+/// The TLDR is that we need to take the value of the `x-okta-verification`
+/// header and return it in a JSON response. I have no idea why they do this.
+async fn handle_get(
+    OktaVerificationHeader(okta_header): OktaVerificationHeader,
```
jshearer, line 137: This is kind of a nit-pick since the end result is the same and there's nothing else involved in serving a `GET` request... but:

I wasn't expecting  the `OktaVerificationHeader` extractor to also be responsible for returning the 404 if it's not provided, rather I would have expected an `Option<OktaVerificationHeader>` and for the route handler to decide what to do if it's `None`. Looks like this is [supported](https://docs.rs/axum/latest/axum/extract/index.html#optional-extractors) via `OptionalFromRequestParts`. Thoughts on this?

jshearer (approved):
I have one small, optional nit-pick that doesn't change the external behavior at all. Either way, LGTM 👍

source-http-ingest/src/server.rs
psFried: The way I see it, `handle_get` _is_ deciding what to do if the header is missing, by declaring the function parameter as just `OktaVerificationHeader`. I guess it doesn't decide what the response body should be, but 🤷. We can change the parameter type later to an `Option` or `Result` if we want to handle GET requests where that header is missing, but for now we don't have a need.

PR #2540, opened 2025-03-16
materialize-iceberg: column type migrations

williamhbaker:
**Description:**

Adds support for column type migrations to materialize-iceberg. See commit messages for more details.

**Workflow steps:**

**Notes for reviewers:**

I manually tested all of these migrations, and added as much unit test coverage as I could. Our SQL materializations have a test suite that covers migration scenarios, but this is the first materialization that doesn't use the `materialize-sql` framework, and so it can't use those same tests. We should make a more generalized set of tests for that and I'll probably get around to it at some point, but didn't do that right now since I'm not totally sure what it would look like (I want to get another materialization or two using the `Materializer` interface), and it's just a big pain to do the equivalent of "INSERT INTO" on an Iceberg table for putting in test value, so that will be a project unto itself - perhaps for if we pursue a compute-less delta updates mode.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2540)


mdibaiee (approved):
LGTM, there is a nit typo below

PR #2541, opened 2025-03-17
Minor materialization error message change

jwhartley:
Made materialization error more specific: key needs to be *required and* non-nullable, or have a default value

### DIFF
--- a/materialize-boilerplate/validate.go
+++ b/materialize-boilerplate/validate.go
@@ -89,7 +89,7 @@ func (v Validator) ValidateBinding(
 		hasDefault := p.Inference.DefaultJson != nil
 
 		if !mustExist && !hasDefault {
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

jwhartley:
@williamhbaker resolved but I don't have merge permission

williamhbaker:
I think you'll need to rebase on the latest from main. Also we are going to want to keep the collection name in the error message, which is something I added recently, see https://github.com/estuary/connectors/blob/7ce6635c785738d774417ad87be26b7528dfd5ae/materialize-boilerplate/validate.go#L93

It might be easiest to put in an access request for repo access to the connectors repo so that you can open a PR directly there and merge it after it is approved

PR #2543, opened 2025-03-17
source-facebook-marketing: use existing session to fetch thumbnail urls

Alex-Bair:
**Description:**

The `fetch_thumbnail_data_url` function was not re-using the existing `requests.Session` & was instead making one-off `requests.get` calls. This caused a DNS query for each `fetch_thumbnail_data_url` invocation, and overall causes a TON of excessive, identical DNS queries.

By using the existing `requests.Session` instance to make fetch thumbnail URLs, DNS queries are much less frequent since connections can be reused.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that the frequency of DNS queries after this change are significantly lower.


jgraettinger (approved):
LGTM

PR #2547, opened 2025-03-18
source-sqlserver: Read Replica CDC

willdonnelly:
**Description:**

This PR adds support for SQL Server CDC capturing from a read-only standby replica. It does this by:

- Implementing a new fence mechanism in `streamToReplicaFence` which only consults information which is available on a read-only replica.
- Detecting at startup whether the target database is a read replica.
- Automatically using `streamToReplicaFence` when the target database is a read replica (or when the `replica_fencing` feature flag is explicitly enabled, just in case we need to be able to manually turn it on in other situations).

This new fence mechanism has a minor downside, which is that it doesn't (and can't, at least in every possible scenario) guarantee that when a row is modified concurrently with a backfill, the backfilled row state and the concurrent changes from replication will be captured in the same Flow transaction.

Other than this caveat the capture output is still entirely correct, so any users with standard reduction behavior into a materialization wouldn't even be able to notice a difference, but it's just a _tiny_ bit weaker of a correctness guarantee compared to what our SQL CDC connectors usually offer.

But I have proven to my own satisfaction that it is literally impossible to make the more precise guarantee in all situations when looking solely at the available data on a read replica, so if we want to support read replica CDC for SQL Server at all this is the best we're going to be able to do. And we do want to support that.

This fixes https://github.com/estuary/connectors/issues/2548

**Workflow steps:**

Setting up a capture from a SQL Server read replica should just work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2547)

### DIFF
--- a/source-boilerplate/testing/testing.go
+++ b/source-boilerplate/testing/testing.go
@@ -46,6 +46,11 @@ type CaptureSpec struct {
 	Validator  CaptureValidator
 	Sanitizers map[string]*regexp.Regexp
 	Errors     []error
+
+	// CaptureDelay is an optional delay to introduce before each Capture() call. It is mostly
+	// useful in replicated setups, where we may need to wait a brief period after making changes
+	// before expecting to reliably capture them from the replica.
+	CaptureDelay time.Duration
 }
 
 // Validate performs validation against the target database.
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

PR #2550, opened 2025-03-18
source-mixpanel-native: skip fetching members for cohorts associated with deleted data groups

Alex-Bair:
**Description:**

Mixpanel cohorts can be associated with a data group. If a data group that has cohorts associated with it is deleted, the members of that cohort can't be queried via `/engage` and doing so results in a somewhat specific 404 response. I think this can be fixed by removing the association between the cohort & data group in Mixpanel, but the user has to do that in their own account. Until the user makes that fix, we log a warning & skip to fetching the next cohort's members.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that trying to fetch members of a cohort with a deleted data group & receiving a 404 response with "Missing Datagroup" in the body causes the connector to log out a warning & move on to the next cohort.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2550)


jonwihl (approved):
LGTM

PR #2552, opened 2025-03-19
source-postgres: Fix 'infinity' value in DATE column

willdonnelly:
**Description:**

Also adds a test of the various weird inputs that date/time columns accept, even though I think most of them translate to some specific date/time value and infinity is the only weird one here.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2552)


Alex-Bair (approved):
LGTM

PR #2553, opened 2025-03-19
source-sqlserver-batch: Default to 24h polling interval

willdonnelly:
**Description:**

I can't remember why I made it 1h for this new capture when the others like MySQL (barring Postgres which is a special case as it uses xmin) default to 24h.

This changes the default behavior to 24h, when it's not overridden by either the endpoint spec or a particular resource config.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2553)


dyaffe (approved):
LGTM

PR #2556, opened 2025-03-21
sqlcapture + source-boilerplate: The most trivial connector status implementation

willdonnelly:
**Description:**

Adds some very generic lifecycle status logging to `source-boilerplate` and more fine-grained but still pretty brief status messages throughout the SQL CDC capture logic. With these changes a generic Go capture with no other logging will produce:

    Initializing connector   # Protocol initialization. May not be worth logging?
    Starting capture         # When the '.Pull()' method of the driver is invoked
    Capture started          # When the '.Ready()' message indicates it's started

And a SQL CDC connector will produce something like:

    Initializing connector            # Very Briefly
    Starting capture                  # Very Briefly
    Connecting to database            # For however long this takes, usually not much
    Verifying capture requirements    # For however long this takes, usually not much
    Capture started                   # Very Briefly
    Discovering database tables       # For however long this takes, usually not much
    Catching up on CDC history        # For however long this takes, usually not much
    Backfilling 1 out of 5 tables
    Streaming live changes

Obviously we discussed doing more than this during standup today and we still want that stuff too, but this is the simplest most trivial implementation of status logging imaginable and it probably still delivers a good 80-90% of what users want to know, so I figured it was worth stopping here and uploading this much right away for discussion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2556)

### DIFF
--- a/source-boilerplate/boilerplate.go
+++ b/source-boilerplate/boilerplate.go
@@ -134,7 +134,7 @@ func (s *ConnectorServer) Capture(stream pc.Connector_CaptureServer) error {
 				return err
 			}
 		case request.Discover != nil:
-			log.WithField("eventType", "connectorStatus").Info("Discovering tables")
+			// No connectorStatus logging here, Discover RPC logging doesn't go into the ops log collection.
 			if response, err := s.Connector.Discover(ctx, request.Discover); err != nil {
 				return err
 			} else if err := stream.Send(&pc.Response{Discovered: response}); err != nil {
--- a/source-boilerplate/boilerplate.go
+++ b/source-boilerplate/boilerplate.go
@@ -68,6 +68,9 @@ func RunMain(connector Connector) {
 		log.SetLevel(lvl)
 	}
 
+	// Log an 'Initializing' message as soon as logging is configured.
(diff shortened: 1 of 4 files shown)


source-boilerplate/boilerplate.go
```diff
@@ -131,24 +134,28 @@ func (s *ConnectorServer) Capture(stream pc.Connector_CaptureServer) error {
 				return err
 			}
 		case request.Discover != nil:
+			log.WithField("eventType", "connectorStatus").Info("Discovering tables")
```
psFried, line 137: Just as an FYI, these logs won't get plumbed through as the status because they don't actually get added to the ops logs collections. It won't hurt anything to have them here, but it might be confusing to have them. The status updates during `Open` and `Apply` would get plumbed through though.

psFried (approved):
LGTM

source-boilerplate/boilerplate.go
willdonnelly: Yeah, that's a good point. I've removed this log message for discover RPCs.

PR #2557, opened 2025-03-21
source-salesforce-native: fix RestQueryManager to handle multiple large field chunks

Alex-Bair:
**Description:**

In the initial version of `source-salesforce-native`, I made an assumption that all REST API queries would use the same page size. It turns out I was wrong; Salesforce dynamically reduces the page size for a query depending on how much data it returns. This means that when the connector uses more than one `Query` to fetch fields, it's possible for those queries to have different page sizes, and fields for a specific record could be on different pages for each query (ex: a set of fields could be on page 5 of query 1 but the other set of fields could be on page 6 of query 2). This breaks assumptions the `RestQueryManager` relied on.

To fix this, the `RestQueryManager` now builds records up across pages of separate queries, keeping records in memory until they're completed & yielding them in the order they were originally received. Some effort has been made to reduce how many records are kept in memory (ex: try to keep each query's overall progress as close to each other as possible, delete a record after all fields are merged into it).

There's future room for improved memory usage if needed. Primarily, the connector could discard incomplete records that have a cursor value that's before the most recently yielded record's cursor value. These incomplete records were updated between kicking off the individual queries, so the current set of queries won't capture all fields for them & we'll fully capture them on a future incremental sweep. I don't anticipate this will be an issue since the window between kicking off queries should be fairly small, but it's something to keep in mind if we see memory issues later.

A side note, the `RestQueryManager` was built with the assumption that chunking up fields across separate would always use a page size of 2,000, making it a faster way to retrieve records than using the `FIELDS(ALL)` syntax that requires a `LIMIT 200` clause. However, since Salesforce dynamically reduces the page size, this assumption also may not hold water when there are a large number of fields. Another potential future improvement, it may be worth investigating using the `FIELDS(ALL)` syntax when we'd end up using more than one chunk & see if we'd yield completed records faster than explicitly specifying the fields.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack with a couple Salesforce accounts containing a non-trivial amount of data. Confirmed for the `RestQueryManager`:
- records are built up across pages of queries.
- records are yielded in ascending order of their cursor field.
- memory usage is not significantly different from before (hovered between 6-8% with a couple streams enabled).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2557)

### DIFF
--- a/source-salesforce-native/source_salesforce_native/rest_query_manager.py
+++ b/source-salesforce-native/source_salesforce_native/rest_query_manager.py
@@ -1,9 +1,9 @@
 from datetime import datetime
 from logging import Logger
-from typing import Any, AsyncGenerator
+from typing import Any, AsyncGenerator, TypedDict
 
 from estuary_cdk.http import HTTPSession
-from .shared import async_enumerate, build_query, dt_to_str, str_to_dt, VERSION
+from .shared import build_query, dt_to_str, str_to_dt, VERSION
 from .models import (
     CursorFields,
     SoapTypes,
@@ -27,6 +27,7 @@ class Query:
         self.query_locator = None
         self.done = False
         self.total_size = 0
+        self.records_yielded = 0
 
     def _set_query_locator(self, next_records_url: str | None) -> None:
         if next_records_url is None:
@@ -54,12 +55,26 @@ class Query:
             await self.http.request(self.log, url, params=params)
         )
 
+        count = 0
+
         for record in response.records:
             yield record
+            count += 1
 
+        self.records_yielded += count
         self.done = response.done
         self.total_size = response.totalSize
         self._set_query_locator(response.nextRecordsUrl)
+
+
+# RecordAndChunksCompleted contains an in-progress record that's being built-up from multiple Queries
+# that are fetching separate chunks of available fields. Fields are merged into the record
+# when they are received. chunks_completed_count is used to determine when a record has
+# been completed & all of its fields have been merged into it; once chunks_completed_count
+# is the same as the number of total field chunks, the record is completely built and can be yielded.
+class RecordAndChunksCompleted(TypedDict):
+    record: dict[str, Any]
+    chunks_completed_count: int
 
 
 class RestQueryManager:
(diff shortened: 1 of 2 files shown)


source-salesforce-native/source_salesforce_native/rest_query_manager.py
```diff
@@ -112,40 +149,56 @@ async def execute(
...
 
+        records: dict[str, RecordAndChunksCompleted] = {}
         while True:
-            records: list[dict[str, Any]] = []
+            # next_query is the query that's yielded the fewest records so far. Since Salesforce dynamically reduces the page size of queries
+            # depending on how much data is returned for each query, the page size for each query can be different.
+            # We cache a record in memory until all of its fields are fetched. Each query's records_yielded attribute is used to gauge the
+            # overall progress of each query. Executing the query that's made the least progress ensures the number of incomplete records we
+            # hold in memory is bounded.
+            next_query = queries[0]
```
mdibaiee, line 159: is it possible that `queries[0].done`, in which case setting this initial value is wrong I guess? I don't fully understand this though.

mdibaiee (approved):
From reading your description and reading the changes LGTM, but I don't fully understand the underlying logic. I just noted one potential logical bug in the code

source-salesforce-native/source_salesforce_native/rest_query_manager.py
Alex-Bair: This section is suppose to find the `query` that's made the least overall progress (measured via `query.records_yielded`), and fetch a page of records for that query so it doesn't lag behind the other queries. That's done to minimize how many incomplete records are kept in memory.

For example, say we have two queries. Query 1 has a page size of 200 and Query 2 has a page size of 150. With the previous logic where we alternated between queries & fetching a page for each, the number of incomplete records in memory grows as more pages are fetched:

| HTTP Request | Query 1 Records Fetched Count | Query 2 Records Fetched Count | Number of Records in Memory |
| ----- | ----- | ----- | ----- |
| 1 | 200 | 0 | 200 | 
| 2 | 200 | 150 | 50 |
| 3 | 400 | 150 | 250 | 
| 4 | 400 | 300 | 100 |
| 5 | 600 | 300 | 300 |
| 6 | 600 | 450 | 150 |
| 7 | 800 | 450 | 350 |
| 8 | 800 | 600 | 200 |

By fetching a page of records for the query that's made the least progress, the number of records in memory is bounded by the largest page size:

| HTTP Request | Query 1 Records Fetched Count | Query 2 Records Fetched Count | Number of Records in Memory |
| ----- | ----- | ----- | ----- |
| 1 | 200 | 0 | 200 | 
| 2 | 200 | 150 | 50 |
| 3 | 200| 300 |  100 | 
| 4 | 400 | 300 | 100 |
| 5 | 400 | 450 | 50 |
| 6 | 600 | 450 | 150 |
| 7 | 600 | 600 | 0 |
| 8 | 800 | 600 | 200 | 

Does that help clarify?

Alex-Bair: As for it being possible for `queries[0]` to be done, that's a good point. It is possible `queries[0].done` is `True`, but `next_query` should be reassigned to a not done query when we compare the number of records each query has yielded later in the `if` statement. However, that assumes the total number of records for each query is the same, which isn't always true (the totals can be different if records are updated between kicking off queries).

I'll update the logic to something more like what's below so `next_query` is always a not done query:
```
next_query: Query | None = None
for query in queries:
    if query.done:
        continue
    elif next_query is None:
        next_query = query
    elif query.records_yielded < next_query.records_yielded:
        next_query = query

```

PR #2567, opened 2025-03-24
materialize-webhook: Make relativePath required in resource config

psFried:
estuary/flow#2001 introduced a change to validate that the `x-collection-name` field is required by the resource spec JSON schema. This connector did not previously mark that as required, so here's a commit to fix that.


JustinASmith (approved):
LGTM

PR #2568, opened 2025-03-24
source-gainsight-nxt: new connector

JustinASmith:
**Description:**

Adds support for Gainsight NXT (previously Customer Success) and the following streams: `Company`, `gsuser` (Users), `cta_group` (Success Plan), `cs_task` (CS Task), `activity_timeline`, `call_to_action`, and `da_picklist`.

`da_picklist` stream is snapshot only. Every other stream is incremental.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2568)


source-gainsight-nxt/source_gainsight_nxt/models.py
```diff
@@ -0,0 +1,105 @@
...
+class User(GainsightResourceWithModifiedDate):
+    OBJECT_NAME: ClassVar[str] = "gsuser"
+
+
+class CsTask(GainsightResourceWithModifiedDate):
+    OBJECT_NAME: ClassVar[str] = "cs_task"
+
+
+class SuccessPlan(GainsightResourceWithModifiedDate):
+    OBJECT_NAME: ClassVar[str] = "cta_group"
```
williamhbaker, line 91: Is this object name right? Wondering since `cta_group` is quite different than the name of the class

source-gainsight-nxt/source_gainsight_nxt/resources.py
```diff
@@ -0,0 +1,196 @@
...
+                model=resource,
+                open=functools.partial(open, fields, resource),
+                initial_state=common.ResourceState(
+                    inc=common.ResourceState.Incremental(cursor=cutoff),
+                    backfill=common.ResourceState.Backfill(
+                        cutoff=cutoff, next_page=None
+                    ),
+                ),
+                initial_config=common.ResourceConfig(
+                    name=name, interval=timedelta(minutes=5)
```
williamhbaker, line 175: Does this need such a long interval? Usually for incremental bindings we would have this interval be much smaller (perhaps with no interval at all) unless there were rate limit considerations.

williamhbaker (approved):
LGTM, see comments

source-gainsight-nxt/source_gainsight_nxt/models.py
JustinASmith: I used [Gainsight metadata API](https://support.gainsight.com/gainsight_nxt/API_and_Developer_Docs/Data_Management_APIs/Data_Management_APIs#Method_3) to list all objects and found this:

```json
{
    "objectName": "cta_group",
    "label": "Success Plan",
    "objectType": "SYSTEM",
    "keyPrefix": "1S025QA4T6K6E278T2",
    "transactional": false,
    "multiCurrencySupported": false,
    "readable": true,
    "createable": false,
    "schemaUpdateable": false,
    "description": "Success Plan"
}
```

However, the documentation has no mention of this that I could find. In hindsight and further investigation, there is a [V2 API for Success Plans](https://support.gainsight.com/gainsight_nxt/API_and_Developer_Docs/Success_Plan_APIs/Success_Plan_APIs#Method), but is slightly different and I am not fully sure I can use the same fields obtained from the Metadata endpoint in the API request to capture data.

I need to dig a bit more to make sure using the V1 API with `cta_group` is okay to get Success Plan data or if there needs to be more work to account for V1 and V2 endpoints and their differences.

source-gainsight-nxt/source_gainsight_nxt/resources.py
JustinASmith: Updated to 1 minute instead per our conversation via Slack.

source-gainsight-nxt/source_gainsight_nxt/models.py
JustinASmith: Plan to use the V1 API which calls Success Plan objects `cta_group`. This is to reduce complexity of supporting V1 and V2 APIs right now. However, we may need to revisit this as the use case for this connector grows.

PR #2569, opened 2025-03-24
materialize-snowflake: hash pipe names with special characters

mdibaiee:
**Description:**

- Hash pipe names following Iceberg's example with some modification to fit the requirements here
- Add a test case for a table name with special characters to test it out

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2569)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -309,7 +309,7 @@ func (d *transactor) addBinding(target sql.Table) error {
 
 	if b.target.DeltaUpdates && d.cfg.Credentials.AuthType == JWT {
 		var keyBegin = fmt.Sprintf("%08x", d._range.KeyBegin)
-		var tableName = strings.ReplaceAll(b.target.Path[len(b.target.Path)-1], " ", "_")
+		var tableName = b.target.Path[len(b.target.Path)-1]
 		var pipeName = strings.ToUpper(sanitizeAndAppendHash("flow_pipe", b.target.Binding, keyBegin, d.version, tableName))
 		b.pipeName = fmt.Sprintf("%s.%s.%s", d.cfg.Database, d.cfg.Schema, pipeName)
 	}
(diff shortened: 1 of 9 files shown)


materialize-snowflake/snowflake.go
```diff
@@ -308,14 +308,10 @@ func (d *transactor) addBinding(target sql.Table) error {
...
-		var pipeName string
-		var err error
-		if pipeName, err = renderTableShardVersionTemplate(b.target, d._range.KeyBegin, d.version, d.templates.pipeName); err != nil {
-			return fmt.Errorf("pipeName template: %w", err)
-		} else {
-			pipeName = strings.ToUpper(strings.Trim(pipeName, "`"))
-			b.pipeName = fmt.Sprintf("%s.%s.%s", d.cfg.Database, d.cfg.Schema, pipeName)
-		}
+		var keyBegin = fmt.Sprintf("%08x", d._range.KeyBegin)
+		var tableName = strings.ReplaceAll(b.target.Path[len(b.target.Path)-1], " ", "_")
```
williamhbaker, line 312: I think this `tableName` part gets sanitized in `sanitizeAndAppendHash` already, so maybe the spaces don't need this replacement?

williamhbaker (commented):
Generally this looks good, but I am wanting to understand the transition scenario better where we have pre-existing materializations with pipes named in the old way that will restart on this connector which names pipes in a new way.

I see [this code](https://github.com/estuary/connectors/blob/3a45d7cf984402b0c51e583b823b9cb33b94e8e0/materialize-snowflake/snowflake.go#L507) which is splitting the pipe name on dots - is this change backward compatible?  Also if the connector restarts on the new code with previously staged files & a pipe named the old way, is that transition going to do the right thing?

mdibaiee:
@williamhbaker we store the pipe name as part of the checkpoint item, and the code that uses that pipe name (`Acknowledge`) has not changed, therefore existing checkpoint items will be processed as they always have been given their prior values.

New pipe names will be constructed differently when it comes to their last part (the part based on the table name), but the other two parts which are separated by a dot `.` are still the same: the database and the schema, so the code you mentioned that splits the pipe name based on dots continues to work as intended.

materialize-snowflake/snowflake.go
mdibaiee: You are right that is a leftover from before, removed 👍🏽

williamhbaker (approved):
LGTM

PR #2575, opened 2025-03-26
source-postgres: Query/log current XID periodically

willdonnelly:
**Description:**

Adds logic to source-postgres discovery to query and log the current server XID. This is a hacky place to put that behavior but discovery runs at connector startup and periodically throughout the connector lifecycle, so it should give us the information we need.

The query/log function shouldn't be able to cause task failures since errors are logged but not returned. It first tries `txid_current()` and then if that fails (which should only happen on a read-only standby replica) it tries `pg_last_committed_xact()` (which should give a similar result and work on a standby according to what I've read, but I haven't used that function before so I can't vouch for it firsthand).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2575)

### DIFF
--- a/source-postgres/discovery.go
+++ b/source-postgres/discovery.go
@@ -17,6 +17,15 @@ import (
 
 // DiscoverTables queries the database for information about tables available for capture.
 func (db *postgresDatabase) DiscoverTables(ctx context.Context) (map[sqlcapture.StreamID]*sqlcapture.DiscoveryInfo, error) {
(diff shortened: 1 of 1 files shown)


dyaffe (approved):
LGTM

willdonnelly:
Confirmed that the `txid_current()` logging path works on a normal database instance and the `txid_snapshot_xmax(txid_current_snapshot())` fallback works on a read-only standby replica. Merging now.

PR #2577, opened 2025-03-26
source-{mysql,postgres,sqlserver}: Initial backfill cursor

willdonnelly:
**Description:**

This commit adds two hidden features for internal use:
- For all three captures, setting the "Skip Backfills" advanced option to `*.*` will cause all tables to immediately transition to active without any backfilling. This lets us avoid needing to modify the capture mode for a whole bunch of bindings.
- For all three captures, it is possible to put a setting like `initial_backfill_cursor=12345` into the feature flags list. If this is done, the value `12345` will be used as the initial cursor value when there is no resume cursor, instead of the normal behavior of taking the current end of the WAL.

As a reminder, this is what cursor values look like for each DB:

    // MySQL
    binlog.12345:7890

    // PostgreSQL
    123/45678901

    // SQL Server
    ALUrCQADl2AAGA==

**Workflow steps:**

Set "Skip Backfills" to `"*.*"` and put `initial_backfill_cursor=XYZ` in the feature flags box when recreating a capture task which is intended to take over from a preexisting one at a specific WAL position.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2577)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -91,9 +91,20 @@ func connectMySQL(ctx context.Context, name string, cfg json.RawMessage) (sqlcap
 		logrus.WithField("flags", featureFlags).Info("parsed feature flags")
 	}
 
+	// This is a bit of an ugly hack to allow us to specify a single _value_ as part of a feature flag.
+	// The alternative would be that we'd have to add a visible advanced option for what is really just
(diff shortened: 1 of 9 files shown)


williamhbaker (approved):
LGTM

PR #2580, opened 2025-03-27
materializations: don't log apply action for non-causal apply action failures

williamhbaker:
**Description:**

The apply action that fails will cause a context cancellation, and all other actions will then fail due to that context being cancelled. This causes a bunch of log spam and makes it very difficult to see which action it was that failed for something other than a context cancellation.

This change will make it so only the erroring action will log out its action.
**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2580)

### DIFF
--- a/materialize-boilerplate/apply.go
+++ b/materialize-boilerplate/apply.go
@@ -2,6 +2,7 @@ package boilerplate
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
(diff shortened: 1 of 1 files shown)


JustinASmith (commented):
LGTM

JustinASmith (approved):

PR #2581, opened 2025-03-27
source-{mysql,postgres,sqlserver}: Add 'force_reset_cursor=XYZ'

willdonnelly:
**Description:**

This is like the 'initial_backfill_cursor=XYZ' flag hackery, but it doesn't even require that you perform a backfill to trigger it. This should generally be set, and then you watch it take effect, and then you immediately publish another change to unset it again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2581)


williamhbaker (approved):
LGTM

PR #2582, opened 2025-03-27
build: upgrade go connectors to go1.24

williamhbaker:
**Description:**

It could be said that the goal of this commit is to upgrade all of our Go connectors to 1.24, and that really is the objective. But unfortunately it isn't as simple as just bumping the Go version, as illustrated in prior (004ffff) attempts at doing that.

A primary knock-on effect is that in Go 1.23, using the `//go:linkname` directive is no longer allowed. We don't do that anywhere, but the version of the Vitess MySQL parser we were using does, so that needs to be updated to a more current version (we could build source-mysql with `-checklinkname=0` I suppose, but that seems like it would be delaying the inevitable). There are some changes to the source-mysql connector because of that.

Bringing in an updated Vitess also updates some AWS packages to the point where we run into the "ResolveEndpointV2" issue with the AWS Go SDK where if some packages are before v1.23.0 and some are after, stuff breaks. So all the AWS packages are being updated to their latest versions.

Finally I am updating the version of `arrow-go` that we are using to v18. I honestly can't remember if that was needed for the rest of this or not, but it probably was, and I'll want to do it at some point anyway.

**Workflow steps:**

**Notes for reviewers:**


willdonnelly (approved):
LGTM

It's unfortunate that a "simple" version upgrade requires this many changes, but I see no issues with any of the changes themselves.

PR #2583, opened 2025-03-27
source-chargebee-native: new connector

JustinASmith:
**Description:**

This PR introduces the Chargebee native connector, including the necessary resource schemas, API integration, and configuration files.

This foundational setup enables further development and testing of the Chargebee integration. There are multiple TODO items to revisit after this initial release.

Below are the incremental and full refresh streams.

> **NOTE: The Product Catalog 2.0 is disabled for now, and all streams associated with 2.0 are not fully tested, even if they are implemented currently.**

**Incremental Streams**

These streams support incremental sync using a cursor field:

- [x] **Addons**
  - Cursor: `updated_at`
- [x] **Attached Items** (Product Catalog 2.0 only)
  - Cursor: `updated_at`
  - Optional: include deletions
- [x] **Coupons**
  - Cursor: `updated_at`
  - Optional: include deletions
- [x] **Credit Notes**
  - Cursor: `updated_at`
- [x] **Customers**
  - Cursor: `updated_at`
- [x] **Events**
  - Cursor: `occurred_at`
- [x] **Hosted Pages**
  - Cursor: `updated_at`
- [x] **Invoices**
  - Cursor: `updated_at`
  - Optional: include deletions
- [x] **Items** (Product Catalog 2.0 only)
  - Cursor: `updated_at`
- [x] **Item Prices** (Product Catalog 2.0 only)
  - Cursor: `updated_at`
- [x] **Item Families** (Product Catalog 2.0 only)
  - Cursor: `updated_at`
- [x] **Orders**
  - Cursor: `updated_at`
  - Optional: include deletions
- [x] **Payment Sources**
  - Cursor: `updated_at`
- [x] **Plans**
  - Cursor: `updated_at`
  - Optional: include deletions
- [x] **Quotes**
  - Cursor: `updated_at`
  - Optional: include deletions
  - Requires a Performance or Enterprise Chargebee plan
- [x] **Subscriptions**
  - Cursor: `updated_at`
  - Optional: include deletions
- [x] **Transactions**
  - Cursor: `updated_at`
- [x] **Virtual Bank Accounts**
  - Cursor: `updated_at`

**Full Refresh Streams**

These streams only support full refresh:

- [x] **Site Migration Details**

## Not Yet Implemented

- [ ] **Gifts** - Possible client-side filtering
- [ ] **Unbilled Charges** - Possible client-side filtering
- [ ] **Comments** - Potentially incremental, needs revisiting
- [ ] **Contacts** - Child resource of customers, paginated by offset without incremental cursor
- [ ] **Differential Prices** - No updated_at cursor, pagination only
- [ ] **Promotional Credits** - Potentially incremental, needs revisiting
- [ ] **Quote Line Groups** - Need to handle 404 errors when quotes feature is not enabled
- [ ] **Subscriptions With Scheduled Changes** - Child resource of subscriptions, potentially incremental

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2583)


source-chargebee-native/source_chargebee_native/models.py
```diff
@@ -0,0 +1,302 @@
...
+    ConnectorState as GenericConnectorState,
+)
+
+
+def default_start_date():
+    dt = datetime.now(tz=UTC) - timedelta(days=30)
+    return dt
+
+
+class ApiKey(BasicAuth):
```
williamhbaker, line 23: This seems strange. In other connectors that only use an API key we do it like this: https://github.com/estuary/connectors/blob/b41fc920848ff69123516e68861c6e969ba9c6fa/source-gladly/source_gladly/models.py#L34-L38

williamhbaker (approved):
LGTM % issue with the authentication configuration - I think there's a way to make that cleaner

source-chargebee-native/source_chargebee_native/models.py
JustinASmith: Chargebee uses an API Key as the username in Basic Auth and an empty password..

I could probably approach it the was you've linked, but would have to encode like we do here: https://github.com/estuary/connectors/blob/main/estuary-cdk/estuary_cdk/http.py#L184-L189

williamhbaker: Oh ok. That's kind of odd but I suppose this will work!

williamhbaker (approved):

source-chargebee-native/source_chargebee_native/models.py
williamhbaker: For follow-up: I do think we should change the endpoint config for the API token. You should be able to have a input for the API token as source-gladly does, and then initialize the `http.token_source` in a similar way, where the `username` input to `BasicAuth` is the API token and the `password` is hard-coded to an empy string: https://github.com/estuary/connectors/blob/230abdbd0d3d02374839da1e8f50cbaab04d51dc/source-gladly/source_gladly/resources.py#L26-L29

PR #2584, opened 2025-03-27
materialize-webhook: don't use chunked transfer encoding

Alex-Bair:
**Description:**

The `http.DefaultClient` uses chunked transfer encoding when Content-Type header is not set or the request body doesn't have a set length (ex: an `*io.PipeWriter`). Some webhook destinations, like Azure Logic Apps, don't work well with chunked transfer encoding. This commit reverts the connector's behavior to not stream the request body as chunks & instead send the request all at once. Request bodies are maxed at ~1 MiB before they are sent.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed requests are ~1 MiB or less, chunked transfer encoding is not used, and that Azure Logic Apps now display the full request body.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2584)

### DIFF
--- a/materialize-webhook/driver.go
+++ b/materialize-webhook/driver.go
@@ -1,20 +1,20 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	m "github.com/estuary/connectors/go/protocols/materialize"
 	schemagen "github.com/estuary/connectors/go/schema-gen"
 	boilerplate "github.com/estuary/connectors/materialize-boilerplate"
 	pf "github.com/estuary/flow/go/protocols/flow"
 	pm "github.com/estuary/flow/go/protocols/materialize"
-	"golang.org/x/sync/errgroup"
 )
 
 // driver implements the pm.DriverServer interface.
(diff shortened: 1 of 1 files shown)


materialize-webhook/driver.go
```diff
@@ -191,124 +191,111 @@ func (d *transactor) Load(it *m.LoadIterator, _ func(int, json.RawMessage) error
...
-
-			if err != nil {
-				pipeReader.CloseWithError(err)
-				return fmt.Errorf("http.NewRequest(%s): %w", address, err)
+	const requestSizeCutoff = 1024 * 1024 * 1 // 1 MiB
+	var bodies = make([]body, len(d.addresses))
+	hasMore := true
+	ctx := it.Context()
+
+	for hasMore {
```
williamhbaker, line 206: I think we can simplify and clean this up quite a bit. I'd suggest looking [here](https://github.com/estuary/connectors/blob/899b4651be0c8532700ae556e65ac4c11a28a6f8/materialize-mysql/driver.go#L701-L800) and [here](https://github.com/estuary/connectors/blob/899b4651be0c8532700ae556e65ac4c11a28a6f8/materialize-pinecone/transactor.go#L60-L111) for inspiration.

Generally, use the characteristic of stores where all store requests for a binding are received in sequential groups. Build up a batch of data, flushing it every 1 MiB (or whatever), and flush it when the binding changes - and then one final flush after reading the iterator fully. We shouldn't need a loop around reading the iterator like this.

Also, previously it seems like we were writing an array of JSON objects, but this is going back to NDJSON?

williamhbaker: Another Q: Do we need to set ContentLength on req to make it not use chunked encoding?

williamhbaker: > Also, previously it seems like we were writing an array of JSON objects, but this is going back to NDJSON?

Oh nevermind - I see the open and closing brackets now.

Alex-Bair: Thanks for those links! Really helpful. I took some inspiration from the `materialize-pinecone` one & refactored. I tested it out some, and I still want to do just a smidge more testing, but I pushed what I have now since I'm going to look into a different issue for the next few hours.

materialize-webhook/driver.go
```diff
@@ -191,118 +191,88 @@ func (d *transactor) Load(it *m.LoadIterator, _ func(int, json.RawMessage) error
...
-		}
 
-		previousAddress = address
+			body.Reset()
 
-		if pipeWriter == nil {
-			startWebhook(address)
-			if _, err := pipeWriter.Write([]byte("[")); err != nil {
-				return nil, fmt.Errorf("pipeWriter.Write: %w", err)
+			if it.Binding != previousBinding {
```
williamhbaker, line 254: It should work to unconditionally set `previousBinding = it.Binding` right after the `if it.Binding != previousBinding || body.Len() >= requestSizeCutoff ...` block.

williamhbaker (approved):
LGTM % remaining comments

materialize-webhook/driver.go
Alex-Bair: That's true, updated.

PR #2586, opened 2025-03-27
source-mixpanel-native: handle split `export` records better

Alex-Bair:
**Description:**

Data from Mixpanel sometimes inserts `�` bytes in response data. This has happened before in the `insert_id` property name (75c33a20708a00fd4dcdc77a8cd4abc8df1c2347) , and evidently it also happens within records. The `response.iter_lines(decode_unicode=True)` interprets `�` as a line break, and the connector tries to yield separate parts of a record.

Airbyte must have encountered this & attempted to address this with the `parts`-related logic in `iter_dicts`. But they didn't anticipate a single part being valid JSON. For example, if the record has a property like:
```
{
    "description": "Sing�99�Luftballons."
}
```

Then the `99` part will be seen as valid JSON, the connector will try to transform & yield it, then fail with a `TypeError``:
```
TypeError: 'int' object is not subscriptable
```

I've improved the handling so `iter_dicts` only yields actual, dictionary records that contain the two properties that should always be present in every `export` record.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that the task that's failing with `TypeError: 'int' object is not subscriptable` no longer fails & gets past the problematic record with `�`s.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2586)

### DIFF
--- a/source-mixpanel-native/source_mixpanel_native/streams/export.py
+++ b/source-mixpanel-native/source_mixpanel_native/streams/export.py
@@ -139,6 +139,9 @@ class Export(DateSlicesMixin, IncrementalMixpanelStream):
             return True
         return super().should_retry(response)
 
+    def is_valid_record(self, potential_record: Any) -> bool:
+        return isinstance(potential_record, dict) and potential_record.get('event', None) and potential_record.get('properties', None)
+
     def iter_dicts(self, lines):
         """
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM

PR #2589, opened 2025-03-28
source-sqlserver: check for encryption of connection

mdibaiee:
**Description:**

We already enable encryption and we have the default tlsmin=1.2, but a customer is insisting on making sure the connection is encrypted, so here is a log message to make sure the connection is encrypted.

There is no failure mode for this check, we always pass, and only log a DEBUG if we can't check for the status, this is intentionally a debug log to avoid freaking users out about encryption. Checking the status requires `VIEW SERVER STATE` permission which is quite broad and we will likely not have in most scenarios.

Logging the session ID allows an administrator from the customer side to check for encryption status themselves.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2589)


Alex-Bair (approved):
LGTM

PR #2592, opened 2025-03-29
source-postgres: Filter backfill XMIN on client side

willdonnelly:
**Description:**

As currently implemented, XMIN backfills have a problem.

The fundamental issue with XMIN backfills as they exist today is that we literally just slap in a `WHERE {xmin is within the specified range} filter or filters into the backfill queries without changing anything else about our behavior. And the problem with this is that our backfill logic is built around the assumption that we are getting result rows back.

This turns out to not be the case when doing an XMIN backfill of a large table (the only kind worth using XMIN backfills on) whose primary keys are mostly ordered (that is, most writes will be at the very end of the table).

What happens in this case is we tell the DB "hey can you give me any rows newer than {min_xid} and stop after N" and the DB says "Sure, I can do that!" and then it goes off and scans over the entire table and if we're lucky it eventually finishes that, many hours later, and gives us back the first N new rows, after which point we can start issuing much faster queries to read out the rest of the new data after those ones.

But if anything interrupts that incremental scan, we have no recourse but to start all over again from the beginning.

And this is clearly dumb, because our backfill queries proceed according to primary key (or CTID) order, so if we _knew_ how far the scan had gotten we could tell the DB to start from that point. But the only way we learn about that progress is when we get back a result row, so that doesn't work right when the result set is heavily filtered.

This commit removes the DB-side filtering in favor of doing that filtering on the client side, and implements a bit of additional plumbing so that the backfill implementation can keep track of a "resume from here" key separate from any actual ChangeEvents it outputs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2592)

### DIFF
--- a/source-mysql/backfill.go
+++ b/source-mysql/backfill.go
@@ -10,7 +10,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func (db *mysqlDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.DiscoveryInfo, state *sqlcapture.TableState, callback func(event *sqlcapture.ChangeEvent) error) (bool, error) {
+func (db *mysqlDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.DiscoveryInfo, state *sqlcapture.TableState, callback func(event *sqlcapture.ChangeEvent) error) (bool, []byte, error) {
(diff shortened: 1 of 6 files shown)


williamhbaker (approved):
LGTM

PR #2600, opened 2025-03-31
source-intercom-native: fix `conversation_parts` bug that could cause missing data

Alex-Bair:
**Description:**

`conversation_parts` works by fetching parts of updated conversations, then only yielding the parts that have been updated since the last log cursor. This was done to avoid yielding conversation parts that were previously yielded. However, this strategy has potential to miss data if another part is added to a conversation that's beyond the date window the connector is currently checking & the conversation is updated again before the connector checks the next date window.

For example, say there's a conversation `123` that had a part added to it at `1742999957`. The connector is checking the window of `1742999856` - `1742999956`. Before the connector finishes that date window, conversation `123` is updated again and it's `updated_at` is pushed to a later time `1743000057`.

Next, the connector checks the date window `1742999956` - `1743000056`. Conversation `123` is not included in this date window, so it's not checked for parts updated after `1742999956`.

Next, the connector checks the date window `1743000056` - `1743000356`. Conversation `123` is included, but the part that was added at `1742999957` is not yielded because it was updated before the most recent log cursor.

This commit greatly simplifies this logic for `conversation_parts` and aligns its checkpointing strategy with `conversations`. If we detect an updated conversation, we yield all of its parts no matter when they were updated. This should ensure the connector captures all parts for updated conversations.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed `conversation_parts` now yields every part of an updated conversation & yields checkpoints after every page when it's safe to do so (like `conversations`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2600)

### DIFF
--- a/source-intercom-native/source_intercom_native/api.py
+++ b/source-intercom-native/source_intercom_native/api.py
@@ -384,106 +384,17 @@ async def fetch_conversations(
         body = _generate_conversations_or_tickets_search_body(start, end, response.pages.next.starting_after)
 
 
-async def _determine_window_end(
-        http: HTTPSession,
-        start: datetime,
-        initial_end: datetime,
-        log: Logger,
-) -> datetime:
-    MAX_PAGE_QUANTITY = 25
-    MIN_WINDOW_SIZE = timedelta(seconds=1)
-    url = f"{API}/conversations/search"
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

PR #2601, opened 2025-03-31
source-sqlserver: Assume primary on error

willdonnelly:
**Description:**

If the `isReplicaDatabase()` function encounters an error, it will now assume that this is not a replica instead of erroring out. The main reason we're likely to see this is on databases which don't even have a `sys.dm_hadr_database_replica_states` table, so erring on the side of assuming it's the primary DB should always be fine.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2601)


jgraettinger (approved):
LGTM

PR #2603, opened 2025-03-31
source-postgres: Sample 0.1% of XMIN-excluded rows

willdonnelly:
**Description:**

When doing an XMIN filtered backfill, instead of reading out all rows and filtering solely on the client side, we can instead read out all rows satisfying the XID range _plus_ 0.1% of rows that don't.

This provides enough data movement to permit incremental progress through a table, while still eliminating 99.9% of unnecessary network throughput and connector CPU usage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2603)


williamhbaker (approved):
LGTM

Issue #2524, opened 2025-03-13
new connector: source-chargebee-native


jonwihl:

JustinASmith:
- [#2583](https://github.com/estuary/connectors/pull/2583) - Initial connector PR
- [#2585](https://github.com/estuary/connectors/pull/2585) - Add to CI PR

PR #2605, opened 2025-04-01
source-chargebee-native: support more streams and both product catalogs

JustinASmith:
**Description:**

Addresses: #2524 
Follow up to the following PR: #2583 

This expands the connector's resource handling capabilities:

Resource Additions:
- Added new Pydantic models, bindings, etc. for:
  * Comments (with incremental sync support via created_at cursor)
  * Contacts
  * Gifts
  * Promotional Credits
  * Quote Line Groups (requires Performance/Enterprise plan)
  * Subscriptions with Scheduled Changes
  * Unbilled Charges

Configuration Improvements:
- Updated resource configurations to support both Product Catalog 1.0 and 2.0
- Enhanced validation logic for resource-specific configurations
- Added proper type hints and validation rules for all new schemas

API Integration Enhancements:
- Improved handling of associated resources with parent-child relationships
- Added support for filtering in associated resource queries
- Implemented proper cursor handling for incremental syncs

This enhancement provides broader coverage of Chargebee's API capabilities while maintaining proper error handling through the API's native responses.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2605)

### DIFF
--- a/source-chargebee-native/source_chargebee_native/__init__.py
+++ b/source-chargebee-native/source_chargebee_native/__init__.py
@@ -14,8 +14,9 @@ from estuary_cdk.capture import (
 )
 from estuary_cdk.capture.common import ResourceConfig
 from estuary_cdk.http import HTTPMixin
+from estuary_cdk.flow import ValidationError
 
-from source_chargebee_native.resources import all_resources, validate_credentials
+from source_chargebee_native.resources import all_resources, validate_credentials_and_configuration
 from source_chargebee_native.models import (
     ConnectorState,
     EndpointConfig,
@@ -48,7 +49,19 @@ class Connector(
         log: Logger,
         validate: request.Validate[EndpointConfig, ResourceConfig],
     ) -> response.Validated:
-        await validate_credentials(log, self, validate.config)
+        await validate_credentials_and_configuration(log, self, validate.config)
+
+        if (
+            validate.lastCapture
+            and validate.lastCapture.config.config.get("product_catalog")
+            != validate.config.product_catalog
+        ):
+            raise ValidationError(
+                [
+                    "Cannot change the `product_catalog` property after the connector has been created."
+                ]
+            )
+
         resources = await all_resources(log, self, validate.config)
         resolved = common.resolve_bindings(validate.bindings, resources)
         return common.validated(resolved)
--- a/source-chargebee-native/source_chargebee_native/api.py
+++ b/source-chargebee-native/source_chargebee_native/api.py
@@ -11,6 +11,7 @@ from source_chargebee_native.models import (
     APIResponse,
     ChargebeeResource,
     IncrementalChargebeeResource,
+    AssociationConfig,
 )
 
 
@@ -37,15 +38,17 @@ async def _fetch_resource_data(
     offset: str | None = None,
     include_deleted: bool = False,
     resource_type: type[ChargebeeResourceType] = ChargebeeResource,
+    filter_params: dict[str, str] | None = None,
+    returns_list: bool = True,
 ) -> tuple[list[ChargebeeResourceType], str | None]:
     url = f"https://{site}.chargebee.com/api/v2/{resource_name}"
 
-    params: dict[str, int | str | list[int]] = {
-        "limit": MAX_PAGE_LIMIT,
-    }
+    params: dict[str, int | str | list[int]] = {}
 
-    if offset is not None:
-        params["offset"] = offset
+    if returns_list:
+        params["limit"] = MAX_PAGE_LIMIT
+        if offset is not None:
+            params["offset"] = offset
 
     if issubclass(resource_type, IncrementalChargebeeResource):
         cursor_field = resource_type.CURSOR_FIELD
@@ -62,6 +65,9 @@ async def _fetch_resource_data(
     if include_deleted:
         params["include_deleted"] = "true"
 
+    if filter_params:
+        params.update(filter_params)
+
     response = APIResponse[resource_type].model_validate_json(
         await http.request(log, url, params=params)
     )
@@ -77,6 +83,7 @@ async def snapshot_resource(
     site: str,
     resource_name: str,
     log: Logger,
+    filter_params: dict[str, str] | None = None,
 ) -> AsyncGenerator[ChargebeeResource, None]:
     offset = None
 
@@ -87,6 +94,7 @@ async def snapshot_resource(
             site,
             resource_name,
             offset=offset,
+            filter_params=filter_params,
         )
 
         if not resource_data:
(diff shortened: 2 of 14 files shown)


source-chargebee-native/source_chargebee_native/resources.py
```diff
@@ -34,92 +39,136 @@
...
 
 
-# TODO(jsmith): revisit
-# PARENT_CHILD_RESOURCES: dict[str, tuple[str, str]] = {
-#     "contacts": ("customers", "id"),  # (parent_resource, parent_key_field)
-#     "subscriptions_with_scheduled_changes": ("subscriptions", "id"),
-# }
+COMMON_STANDALONE_FULL_REFRESH_RESOURCES = [
+    "site_migration_details",
+    "quote_line_groups",  # Requires Performance or Enterprise Chargebee subscription plan - will throw HTTP 404 if not enabled
```
JustinASmith, line 50: @williamhbaker Note that this will throw a 404 if a capture is set up to capture this binding and their account does not have this feature available (via the plan they are on) or enabled in their account.

Should this be handled in another way or should we allow the HTTP 404 to occur as long as we have it stated in the documentation?

source-chargebee-native/source_chargebee_native/resources.py
```diff
@@ -34,92 +39,136 @@
...
+    ("events", Event),
+    ("hosted_pages", HostedPage),
+    ("invoices", Invoice),
+    ("orders", Order),
+    ("payment_sources", PaymentSource),
+    ("promotional_credits", PromotionalCredits),
+    (
+        "quotes",
+        Quote,
+    ),  # Requires Performance or Enterprise Chargebee subscription plan - will throw HTTP 404 if not enabled
```
JustinASmith, line 81: @williamhbaker Same comment above regarding `quote_line_groups`

source-chargebee-native/source_chargebee_native/api.py
```diff
@@ -184,3 +192,175 @@ async def fetch_resource_changes(
...
+    end_date = min(max_updated_at + timedelta(days=30), datetime.now(tz=UTC))
+
+    parent_ids = await _get_parent_ids(
+        http,
+        site,
+        association_config,
+        log,
+    )
+
+    for parent_id in parent_ids:
```
williamhbaker, line 328: The general strategy of doing a separate request for each `parent_id` appears inevitable, but I do have some general concerns about it being slow to do this.

So, a couple of questions / things to think about here. None of this is necessarily a blocker, and I'll leave it to your judgement & knowledge of what may be typical for the binding types that use these routines:
- Are there any "bulk" capabilities for the APIs we are using for this, where instead of setting 1 `parent_id` we could send more than one at a time? I doubt it, but wanted to ask anyway
- Particulary for this fetch_changes style one, is there anything we can do to checkpoint for better allowing incremental progress? For example, say we have a date window that contains a huge number of updates. We'll have to get through _all_ the parent_ids and emit those documents before any checkpoints are emitted. This has frequently been a pain point, since connectors can restart frequently and it may be difficult to get through such a large region of documents since it has to start over at the beginning if it restarts in the middle. I also kind of doubt there's anything that can be done differently here, practically.
- If performance does turn out to be a problem & it seems like more concurrency would help, we could probably use a scatter/gather strategy on these since all the parent IDs are known up front.

source-chargebee-native/source_chargebee_native/resources.py
```diff
@@ -34,92 +39,136 @@
...
     "2.0": [
-        # "comments", # TODO(jsmith): revisit - potentially incremental
-        # "differential_prices", # TODO(jsmith): revisit
-        # "gifts", # TODO(jsmith): revisit - potential client-side filtering
-        # "promotional_credits", # TODO(jsmith): revisit - potentially incremental
-        "site_migration_details",
-        # "unbilled_charges", # TODO(jsmith): revisit - potential client-side filtering
-        # "quote_line_groups", # TODO(jsmith): add this back in after figuring out the best way to handle 404 (quotes feature not enabled)
+        *COMMON_STANDALONE_FULL_REFRESH_RESOURCES,
+        "differential_prices",
```
williamhbaker, line 61: The only difference I see between the 1.0 and 2.0 full refresh resources is that 2.0 has `differential_prices`, so maybe this could be simplified?

source-chargebee-native/source_chargebee_native/api.py
JustinASmith: Good points. At the very least I can easily update to use `asyncio.gather` and look at adding more frequent check pointing (yielding cursor) so we can make progress in smaller bits for these larger associated style steams.

When using the `asyncio.gather` I would want to be sure to balance memory usage, API rate limits, etc. - i am thinking this might be something configurable in the advanced section of the endpoint configuration. What do you think? Maybe a task limit field for associated resources or something comparable? Otherwise, I can select something that seems reasonable and it be in the code rather than configurable in the endpoint config - or just have a reasonable default and allow advanced configuration to change that with a warning description.

williamhbaker: I wouldn't have it in the endpoint config - nobody will ever know what to put there and it will just be us figuring it out for them. Probably just have a limit of 3-5 and call it a day.

Although for now I would suggest just leaving this as-is and we'll see if it becomes an issue and revisit later.

JustinASmith: Will revisit this soon if performance is an issue. Low hanging fruit would be to use the scatter/gather via `asyncio.gather` and keep the yielded log cursor the same. Then if we want to address yielding a cursor more frequent, I could look at chunk things up in a way that allows us to yield a cursor more frequently to allow picking up at a logical start/end point rather than needing to process _all_ parent ids again after restart.

source-chargebee-native/source_chargebee_native/resources.py
```diff
@@ -218,24 +278,129 @@ def open(
...
+        )
+
+    return [
+        common.Resource(
+            name=name,
+            key=["/_meta/row_id"],
+            model=ChargebeeResource,
+            open=functools.partial(open_child, association_config),
+            initial_state=common.ResourceState(),
+            initial_config=common.ResourceConfig(name=name, interval=timedelta(minutes=2)),
```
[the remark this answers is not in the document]
JustinASmith: I think updating the full refresh resources to snapshot every hour makes sense. There could potentially be a larger amount of data returned.

source-chargebee-native/source_chargebee_native/__init__.py
```diff
@@ -48,7 +49,24 @@ async def validate(
         log: Logger,
         validate: request.Validate[EndpointConfig, ResourceConfig],
     ) -> response.Validated:
-        await validate_credentials(log, self, validate.config)
+        await validate_credentials_and_configuration(log, self, validate.config)
+
+        if not validate.lastCapture or not validate.lastCapture.bindings:
```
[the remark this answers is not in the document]
williamhbaker: Yeah I think just remove 54-57. That way check on line 59 will work: If there is a `lastCapture` and that one's catalog isn't equal to this one's, error out. Otherwise continue as normal.

williamhbaker (approved):
LGTM

PR #2606, opened 2025-04-01
source-intercom-native: fetch parts of separate conversations concurrently

Alex-Bair:
**Description:**

The `conversation_parts` stream is very slow since it iterates through a page of conversations and makes an API call to fetch the parts of each conversation one after another.

This process can be sped up significantly by fetching parts for multiple conversations concurrently. The connector now leverages `asyncio.as_completed` to fetch parts of conversations concurrently and yield them in the order that they arrive. The connector caches a list of conversation ids to fetch parts for before the next cursor to checkpoint,
so it's alright for the ordering of `conversation_ids` to not be maintained when yielding conversation parts. The `parts_semaphore` ensures only 15 API requests for parts are in-flight at a time.

The checkpoint strategy for `conversations` was also updated to checkpoint within processing a page as well as between pages. This was done to better handle large stretches of conversations that were updated at the same time / right after each other (ex: 150 conversations updated at `1743530557`, 150 conversations updated at `1743530558`, etc.)

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack with an account containing a non-trivial number of conversations. Confirmed:
- `conversation_parts` progresses much faster.
- memory usage with just `conversation_parts` enabled remained below 7%.
- the connector did not hit rate limit errors.
- `conversations` and `conversation_parts` checkpoints at a safe spot after yielding `SEARCH_PAGE_SIZE` or more documents.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2606)

### DIFF
--- a/source-intercom-native/source_intercom_native/api.py
+++ b/source-intercom-native/source_intercom_native/api.py
@@ -349,6 +349,8 @@ async def fetch_conversations(
     url = f"{API}/conversations/search"
     body = _generate_conversations_or_tickets_search_body(start, end)
 
+    count = 0
+
     while True:
         response = ConversationsSearchResponse.model_validate_json(
                 await http.request(log, url, "POST", json=body)
@@ -365,19 +367,20 @@ async def fetch_conversations(
                 'start': _s_to_dt(start),
             })
 
-        if (
-            last_seen_ts > start
-            and response.conversations
-            and response.conversations[0].updated_at > last_seen_ts
-        ):
-            yield _s_to_dt(last_seen_ts)
-
         for conversation in response.conversations:
+            # It's possible to update multiple conversations at the same time in Intercom, making it difficult to find
+            # a safe spot to checkpoint between pages returned by the API. To checkpoint more frequently when processing these bulk updates,
+            # conversations looks for any safe spot to checkpoint after yielding SEARCH_PAGE_SIZE documents.
             if conversation.updated_at > last_seen_ts:
+                if count >= SEARCH_PAGE_SIZE:
+                    yield _s_to_dt(last_seen_ts)
+                    count = 0
+
                 last_seen_ts = conversation.updated_at
 
             if conversation.updated_at > start:
                 yield conversation
+                count += 1
 
         if response.pages.next is None:
             yield _s_to_dt(last_seen_ts)
--- a/source-intercom-native/source_intercom_native/api.py
+++ b/source-intercom-native/source_intercom_native/api.py
@@ -34,6 +34,8 @@ COMPANIES_LIST_LIMIT_REACHED_REGEX = r"page limit reached, please use scroll API
 COMPANIES_SCROLL_IN_USE_BY_OTHER_APPLICATION_REGEX = r"scroll already exists for this workspace"
 
 companies_scroll_lock = asyncio.Lock()
+parts_semaphore = asyncio.Semaphore(15)
+
 
 def _dt_to_s(dt: datetime) -> int:
     return int(dt.timestamp())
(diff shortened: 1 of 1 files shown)


source-intercom-native/source_intercom_native/api.py
```diff
@@ -389,29 +390,46 @@ async def fetch_conversations_parts(
...
+    CONCURRENCY_LIMIT = 15
+    conversation_ids: list[str] = []
+
     async for conversation_or_dt in fetch_conversations(http, log, log_cursor):
         if isinstance(conversation_or_dt, TimestampedResource):
-            async for part in _fetch_part(http, log, conversation_or_dt):
-                yield part
+            conversation_ids.append(conversation_or_dt.id)
         else:
+            for chunk in itertools.batched(conversation_ids, CONCURRENCY_LIMIT):
```
williamhbaker, line 400: Something to think about: This is certainly faster than doing each one sequentially, but it's still not as fast as it could be. What'll happen here is we'll fire off CONCURRENCY_LIMIT number of requests and then wait for them all to finish. What would be ideal is instead to always have CONCURRENCY_LIMIT number of requests going, and process the results as they are available.

You could probably slot in [buffer_ordered](https://github.com/estuary/connectors/blob/main/source-hubspot-native/source_hubspot_native/buffer_ordered.py) here to get more of that effect. This would yield results in same order as `conversation_ids`, which may or may not be important. If it isn't important, there could be even more efficiency gained by making a similar thing called `buffer_unordered` (does python already have that? maybe) that would not enforce the same ordering of inputs -> outputs.

Alex-Bair: Good point, I have been eyeing [`asyncio.as_completed`](https://docs.python.org/3/library/asyncio-task.html#asyncio.as_completed) to process results as they're available. I don't think keeping the same order as `conversation_ids` is really important since we already have the next cursor we're going to checkpoint. I'll give `asyncio.as_completed` a whirl, and if it doesn't work out, I'll look at adapting `buferred_ordered` to not maintain ordering.

williamhbaker: Ah yeah that seems good. The only thing is you would need to use a semaphore in the task itself to bound the concurrency. An implementation of `buffer_unordered` that just wraps `asyncio.as_completed` could make sense.

Alex-Bair: I pushed an update with `asyncio.as_completed`. It definitely boosts the speed. I didn't end up wrapping it in `buffer_unordered` since the implementation was pretty simple.

Alex-Bair: I figured out what caused the difference between backfill counts. After the backfill with the batched CONCURRENCY_LIMIT approach this morning, large swaths of conversations were updated in Intercom at the same time, and huge chunks of sequential conversations now had the same `updated_at` value. This made it extremely difficult for `conversations` to find a safe spot to checkpoint between pages & allow `conversation_parts` to fetch records when I ran a backfill with `asyncio.as_completed` later in the afternoon. I was evidently too impatient & didn't wait long enough during that backfill - I just re-ran an `asyncio.as_completed` backfill while waiting longer & I received significantly more records.

At times during the second `asyncio.as_completed` backfill, the connector had 10,000+ conversation ids in memory before it found a safe checkpoint between pages & fetched the parts for each conversation. Getting through this batch didn't take an extremely long time (~30 minutes or so), but it'd be nice to checkpoint sooner if a safe spot exists within a page instead of between pages. We were previously doing this, but I changed `conversations`'s checkpointing behavior in 7702e444366d42fd45898e84bd354737f7ea35f5  to align with other streams. Since `conversation_parts` now relies on `conversations` and it's clearly possible to update conversations in bulk so multiple have the same `updated_at`, I'm considering changing `conversations` back to the `CHECKPOINT_INTERVAL` strategy it was using a while ago.

Alex-Bair: I tested out processing that same swath of bulk updated conversations with the `MIN_CHECKPOINT_INTERVAL` strategy, and it's handled much better. Instead of waiting to find a safe spot to checkpoint between pages, the connector now checkpoints if there's a safe spot within a page & `conversation_parts` never exceeded 400 conversation ids in memory before fetching the associated parts.

williamhbaker (approved):
LGTM

PR #2609, opened 2025-04-01
source-salesforce-native: increase csv field size limit

Alex-Bair:
**Description:**

Python's `csv` module has a default field size limit of 131,072 bytes, and it will raise a `_csv.Error` exception if a field value is larger than that limit. Some users have fields larger than 131,072 bytes in their Salesforce accounts (ex: entire email bodies), so we increase the limit to 10 MiB, which should be sufficiently large.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack with a Salesforce account that has field values over 131,072 bytes. Confirmed raising the field size limit to 10 MiB allowed the connector to progress, processing the CSV's rows & finishing the backfill.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2609)

### DIFF
--- a/source-salesforce-native/source_salesforce_native/bulk_job_manager.py
+++ b/source-salesforce-native/source_salesforce_native/bulk_job_manager.py
@@ -2,9 +2,11 @@ import asyncio
 import aiocsv
 import aiocsv.protocols
 import codecs
+import csv
 from datetime import datetime
 from logging import Logger
 import re
+import sys
 from typing import Any, AsyncGenerator
 
 from estuary_cdk.http import HTTPSession, HTTPError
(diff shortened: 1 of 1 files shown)


source-salesforce-native/source_salesforce_native/bulk_job_manager.py
```diff
@@ -27,6 +28,12 @@
 NOT_SUPPORTED_BY_BULK_API = r"is not supported by the Bulk API"
 
 
+# Python's csv module has a default field size limit of 131,072 bytes, and it wil raise an _csv.Error exception if a field value
+# is larger than that limit. Some users have fields larger than 131,072 bytes, so we increase the limit to 10 MiB, which should be
+# sufficiently large.
+csv.field_size_limit(10_000_000)
```
williamhbaker, line 34: MiB actually means mebibyte, which is 1,048,576 (aka: 1024 * 1024) bytes. MB is a megabyte which is 1,000,000 bytes.

Anyway, I would suggest just using `sys.maxsize` (I think that is a thing? We just want the largest possible number) here, so that we never have to worry about this again. I could imagine somebody have a 10+ MiB field and if we ever hit that we would be back here raising the limit once more.

If users have individual rows so large that even one single row OOMs the connector, we'll need to figure out how to incrementally process a single row :smile:

Alex-Bair: 🤦 I'm used to writing out byte limits like `1024 * 1024 * 10`, where it is MiB, but it didn't register that typing out the limit directly with 0s is MB.

I'll update to use `sys.maxsize` regardless, that's the better solution here.

I went down a rabbit hole with this [SO post](https://stackoverflow.com/a/54517228) trying to figure out if we'll need to account for the system architecture when setting `csv.field_size_limit(sys.maxsize)`. It's still not clear to me if we will, but I'll keep in mind that if some `source-salesforce-native` connectors fail with an error like `OverflowError: Python int too large to convert to C long`, I'll have to look back at this more.

PR #2610, opened 2025-04-01
source-sqlserver: Skip `cdc_log_scan_sessions` check on replicas

willdonnelly:
**Description:**

We only need this prerequisite check when in read-only non-replica mode and it's incorrect to let that break a replica capture.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2610)


dyaffe (approved):
LGTM

PR #2613, opened 2025-04-02
materialize-snowflake: a more thorough pipe cleanup

mdibaiee:
**Description:**

- Instead of only cleaning up pipes after finishing a transaction for that specific pipe name as compared to a previous version, clean up all extra `FLOW_PIPE_%` pipes that are not pending use by checkpoint items and share the same table name and keyBegin
- While I was at it, improved our `pipeExists` logic to be more precise and likely more performant as well
- Tested this with our integration tests

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2613)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -893,7 +893,7 @@ func (d *transactor) deleteFiles(ctx context.Context, files []string) error {
 }
 
 func (d *transactor) dropPipe(ctx context.Context, pipeName string) error {
-	if _, err := d.db.ExecContext(ctx, fmt.Sprintf("DROP PIPE %s", pipeName)); err != nil {
+	if _, err := d.db.ExecContext(ctx, fmt.Sprintf("DROP PIPE IF EXISTS %s", pipeName)); err != nil {
 		return fmt.Errorf("dropping pipe %q: %w", pipeName, err)
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


materialize-snowflake/snowflake.go
```diff
@@ -875,6 +869,43 @@ func (d *transactor) deleteFiles(ctx context.Context, files []string) error {
 	return nil
 }
 
+func (d *transactor) cleanupPipes(ctx context.Context, exclude []string) error {
+	var query = "SELECT PIPE_CATALOG, PIPE_SCHEMA, PIPE_NAME FROM INFORMATION_SCHEMA.PIPES WHERE PIPE_NAME LIKE 'FLOW_PIPE_%';"
```
williamhbaker, line 873: Two related questions:
- Is this going to remove pipes used by other materializations going to the same Snowflake database?
- How will this interact with split task shards for the same materialization?

mdibaiee: @williamhbaker both valid points... this would delete pipes from other materializations and also from split shards. I'm going to change this to only look for pipes with the same binding, keyBegin and table name, but with a different version 🤔

williamhbaker (approved):
LGTM

PR #2620, opened 2025-04-02
source-postgres: Make XMIN backfills easier to use correctly

willdonnelly:
**Description:**

This changes the PostgreSQL XMIN backfill behavior so that users can no longer specify an upper-bound XID. The original motivation for specifying an upper-bound was so that, on long-lived servers with a high enough change volume to actually reach XID wraparound, we wouldn't go and capture the whole 50% of the table whose XIDs fell within the 2^31-XID range starting from the minimum, when we really only wanted a much smaller slice of that representing rows which could possibly have been modified during _the current_ epoch and not the prior ones.

The thing is, there's a lot of ways things can subtly go wrong if you specify a maximum. The temptation is to interpret the min/max settings as a sort of "I can precisely specify the time range of the interruption and recapture just that data" patch and specify a maximum _less than_ the current server XID. Heck, I did that my- self recently by mistake.

And unfortunately, it doesn't work that way and if you try to use the maximum value like that you just end up (theoretically) missing changes because the existence of a maximum XID less than the current server XID violates some important assumptions of the backfill process (basically there are certain cases where we can safely do things under the assumption that we'll get to a particular row or change later on in the backfill, and the presence of a maximum can violate that assumption and result in missing data).

So it's much safer to just implement the actual important part of the logic (excluding rows whose `xmin` value is within the range `[txid_current(), min_xid+2147483648)` which satisfies the min_xid filter but cannot possibly be newer than that XID _in the current wraparound epoch_) so that it happens automatically without room for the user to screw it up. Now XMIN backfills just specify a lower bound and capture everything newer than that XID.

**Notes for reviewers:**

This addresses both of the concerns I brought up during standup today:
1. Precise backfills will be correct again even when doing an XMIN backfill, because the assumption that if we see a row change via CDC in a not-yet-backfilled region of the table we will definitely output it when the backfill reaches that row is correct again.
2. We can no longer miss data during the initial "catchup streaming" phase before the backfill starts, because so long as the minimum XID is in the past we can again assume that any changes occurring between the point where the capture restarted and the point where the table is reactivated will be reflected in the backfill output.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2620)


williamhbaker (approved):
LGTM

PR #2621, opened 2025-04-03
source-postgres: Add TestRowLevelSecurity

willdonnelly:
**Description:**

This test case demonstrates that RLS policies don't necessarily apply to replication events, since there was some question about that earlier today. If we ever change that, this test will show the results.


Alex-Bair (approved):
LGTM

PR #2622, opened 2025-04-03
source-{various Python connectors}: fix failing tests

Alex-Bair:
**Description:**

Tests for multiple Python connectors have been failing in CI for various reasons. Typically, the failures have been due to to the SaaS provider adding/removing fields, our credentials becoming invalid, or other testing related reasons that don't actually indicate issues with the connector. I've updated the failing tests to pass. See individual commits for details on each one.

This should help combat alert fatigue when inspecting test results.

I didn't look at fixing `source-oracle-flashback` tests. The fix for those look a little more involved, and I think that connector might be getting deprecated soon(?) anyway.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2622)

### DIFF
--- a/source-criteo/pyproject.toml
+++ b/source-criteo/pyproject.toml
@@ -1,5 +1,5 @@
 [tool.poetry]
-name = "source-criteo-estuary"
+name = "source_criteo"
 version = "0.1.0"
 description = ""
 authors = []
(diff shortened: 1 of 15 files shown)


JustinASmith (approved):
LGTM % question(s) below:

Are the redacted properties necessary to prevent future failures due to these changing often? Typically this is just for redacting sensitive information right? I see we have updated at fields and others now redacted, so my assumption was this needed to be done to prevent failures now and possibly in the future.

Alex-Bair:
> Are the redacted properties necessary to prevent future failures due to these changing often?

Exactly, yep. I redact fields to make the snapshot tests more resilient to changes that we/SaaS systems make within our various dev testing account, as well as for any sensitive information that might be in the snapshots. If we go tinkering around in the test accounts or if the SaaS system automatically makes updates (like some kind of auto archiving or subscriptions), that activity usually causes the snapshot tests to fail since the values for `updated_at` type fields and occasionally other fields change.

PR #2624, opened 2025-04-04
source-braintree-native: incrementally capture past day(s) disbursed transactions

Alex-Bair:
**Description:**

This PR's scope includes:
- Ensuring searches for all transaction search fields will return fewer than `TRANSACTION_SEARCH_LIMIT` results
  - The connector was already checking & raising an exception if any of the searches exceeded `TRANSACTION_SEARCH_LIMIT`, but it wasn't reducing the date window's end if non-`created_at` fields hit the limit. This has never happened (it seems like transactions are created much more frequently than any of the other actions tracked by the other fields), but this should help the connector handle this situation better if it does occur.
- Fetching disbursed transactions in the incremental `fetch_transactions` function.

When a transaction's `disbursement_details/disbursed_date` is updated, none of the `TRANSACTION_SEARCH_FIELDS` or the non-searchable `updated_at` field are updated. Meaning, only incrementally replicating based off the `TRANSACTION_SEARCH_FIELDS` will cause the connector to miss updates to transactions' disbursement dates.

To capture these elusive updates, we now search the previous day(s) for disbursed transactions and yield them when the date window spans more than a single day. When the capture is caught up & capturing incrementally, it should only perform these lookback-type searches once a day the first time it invokes `fetch_transactions` after 00:00 UTC.

**Workflow steps:**

**Notes for reviewers:**

- When a date window's start & end are on different days, `transactions` fetches all transactions with a `disbursement_date` between start (inclusive) and end (exclusive).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2624)

### DIFF
--- a/source-braintree-native/source_braintree_native/api.py
+++ b/source-braintree-native/source_braintree_native/api.py
@@ -247,6 +247,45 @@ async def _determine_next_transaction_window_end(
     return end
 
 
+async def _fetch_transactions_disbursed_between(
+    braintree_gateway: BraintreeGateway,
+    start: datetime,
+    end: datetime,
+    log: Logger,
+) -> AsyncGenerator[IncrementalResource, None]:
+    # Both start_date and end_date are kept as datetimes to retain timezone information (e.g. UTC)
+    # when searching. If timezone information is not included (like with dates), Braintree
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM

PR #2627, opened 2025-04-05
source-postgres: Feature flag for truly multidimensional arrays

willdonnelly:
**Description:**

Implements a new feature flag `multidimensional_arrays`. When set, PostgreSQL array types will be discovered as merely `{type: array}` without any element type information at all, but the capture output will be unflattened back into nested arrays of the appropriate dimensionality.

So for instance the array `[[a,b],[c,d]]` can be represented as such instead of flattened out to `[a, b, c, d]`.

**Workflow steps:**

Set the `multidimensional_arrays` feature flag in the advanced settings for the capture. This flag is not going to become the default, but it's present for power users.

**Notes for reviewers:**

I'm not sure if the test coverage for array unflattening is sufficient or not. On the one hand it seems to work both on simple cases like a 1D 4-element array vs a 2x2 array and even on 2x3x4 vs 2x3x2x2 arrays. On the other hand the logic I wrote worked on the first try and for this sort of function that makes me _deeply uncomfortable_ so I keep looking for some sort of catch, but I don't really see any other edge cases that seem important to cover.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2627)

### DIFF
--- a/source-postgres/.snapshots/TestMultidimensionalArrays-Capture
+++ b/source-postgres/.snapshots/TestMultidimensionalArrays-Capture
@@ -1,0 +1,26 @@
+# ================================
+# Collection "acmeCo/test/test/multidimensionalarrays_82375147": 18 Documents
+# ================================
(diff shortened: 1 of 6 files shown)


jgraettinger (approved):
LGTM

PR #2631, opened 2025-04-07
source-braintree-native: change all bindings to hit API directly

Alex-Bair:
**Description:**

- sorting the bindings in the discover snapshot to make the test more resilient to the order bindings are discovered.
- updating the connector's endpoint config to use the CDK's basic authentication
- switching all bindings to hit the Braintree API directly instead of using their SDK & wrapping it in `asyncio.to_thread`. This was done to speed up the connector significantly; I suspect using `asyncio.to_thread` was slow due to the cost of context switching between each thread & other nuances of thread-based concurrency.

Some interesting notes:
- By hitting the API directly, I discovered that `disputes` isn't subject to the same `SEARCH_LIMIT` the other incremental streams are subject to. This isn't extremely surprising since `disputes` has always been a bit different from everything else (i.e. the Braintree SDK iterated through them differently).
- My goal was to created equivalent/similar functions to what the Braintree SDK provided. That made it easier to swap them into the existing `fetch_foobar` and `backfill_foobar` functions wherever we were using the Braintree SDK. 
- I made `api.py` into a directory to organize the various interactions with the Braintree API. I found this helpful since the Braintree API's endpoints aren't documented anywhere except within their SDK's code.

**Workflow steps:**

**Notes for reviewers:**

- Documents have the same schemas as the previous connector. I followed the same transformation process as the previous version of the connector (XML -> dictionary -> Braintree object -> dictionary -> validated instance of a Pydantic model). We may be able to simplify this transformation some, but I opted not to do that yet since this PR was already pretty large & I wanted to limit the scope to improvements around fetching data from the API.
- Backfills & incremental replication work as they did previously.
- Streams complete much faster. I didn't test backfilling multiple months of data, but in an hour I was able to get at least 4x more data than the existing connector.

All existing production captures will need their spec & config updated to use this new version of the connector. Their credentials need to use `username` and `password` instead of `public_key` and `private_key`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2631)


source-braintree-native/source_braintree_native/models.py
```diff
@@ -28,16 +31,16 @@ class EndpointConfig(BaseModel):
...
-        public_key: str = Field(
+        username: str = Field(
             title="Public Key",
             json_schema_extra={"secret": True},
         )
-        private_key: str = Field(
+        password: str = Field(
             title="Private Key",
             json_schema_extra={"secret": True},
         )
```
JustinASmith, line 46: Will changing authentication method disrupt existing captures and require an update by the user?

JustinASmith (approved):
LGTM % one question! Overall, there is a lot that changed, but looks to be in line with what was here before with some good improvements.

PR #2644, opened 2025-04-08
materialize-postgres: add timeout for queries that have been known to hang

williamhbaker:
**Description:**

On numerous occasions we have observed certain queries hanging forever when materializing to postgres, and as far as we know these have always been due to the database running out of disk space.

This adds a conservative timeout to queries that are prone to hang on this condition, so that the connector will fail with an error and restart, rather than hanging indefinitely.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2644)

### DIFF
--- a/materialize-postgres/driver.go
+++ b/materialize-postgres/driver.go
@@ -3,12 +3,14 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM

PR #2645, opened 2025-04-08
source-postgres-batch: Smarter CTID+XMIN incremental syncs

willdonnelly:
**Description:**

This PR implements a new mode of operation for source-postgres-batch which does a smarter XMIN-incremental sync using CTIDs for incremental progress across task restarts. This represents a departure from the typical batch SQL model because this new capture mode isn't based solely around the execution of a single polling query.

The motivation for this is that representing XMIN captures in that query-execution model required asking the database to scan the entire table and accumulate results and then `ORDER BY xmin` that entire result set. This was especially inefficient for the first query which essentially amounted to a full-table sort, but even after that the requirement that each incremental polling operation complete in a single shot was a pretty big drag.

So instead this new capture mode explicitly scans over the table in a resumable fashion using CTIDs plus a WHERE-clause filtering by `xmin`, and establishes XID values to compare against by directly querying `txid_current()` (or the equivalent standby mechanism) and keeping track of that state explicitly. This is actually pretty similar to how `source-postgres` implements an XMIN-filtered backfill of a keyless table, except that we establish a constant upper-bound for the duration of a polling cycle (because that allows us to guarantee exactly-once capture semantics).

**Workflow steps:**

The new capture mode will be automatically selected for bindings with a `["txid"]` cursor, no template override, and no preexisting cursor state from the old mode of operation. In effect that means that it will automatically kick in for new bindings or preexisting bindings that haven't ever actually succeeded, and you can trigger it for other bindings by backfilling them.

This should be fully backwards-compatible since it only impacts new or failing bindings, and all others will continue to use the old mode of operation indefinitely.


williamhbaker (approved):
LGTM

PR #2646, opened 2025-04-08
source-stripe-native: add `events` stream

Alex-Bair:
**Description:**

- Updating the discover and capture snapshots to be more resilient to changes in the order bindings are discovered.
- Adding the `events` stream and updating the associated `fetch_incremental` and `fetch_backfill` functions to support the new stream.

**Workflow steps:**

The connector docs should be updated to reflect the new stream.

**Notes for reviewers:**

Tested on a local stack. Confirmed events are replicated incrementally & the backfill completes once `events` finishes reading the final page of results.


JustinASmith (approved):
LGTM

PR #2649, opened 2025-04-09
source-postgres: Cleaner XMIN filtering

willdonnelly:
**Description:**

I've just spent the past couple of hours up to my eyeballs in PostgreSQL XIDs and the different `txid_snapshot_foo(...)` functions so I'm going to take this opportunity to clean up and simplify some of the PostgreSQL CDC logic around XMIN filtered backfills while it's fresh in my mind.

Strictly speaking only one of these changes is truly needed ASAP, using `compareXID32` to do the appropriate circular comparison for XMIN filtering in the result row processing.

The other stuff is just cleanups, because we don't need to be using a complicated case statement based on whether or not we are hitting a standby when the `txid_snapshot_xmin/xmax` bits work perfectly fine on the master DB as well.

**Workflow steps:**

Should be no functional changes except that the XMIN filtering will actually work correctly again on databases which have received 2^31 or more writes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2649)


williamhbaker (approved):
LGTM

PR #2650, opened 2025-04-09
source-oracle: optimise keyless backfill query

mdibaiee:
**Description:**

- Ordering the whole table based on `ROWID`, and then filtering for a specific ROWID range is slow. With some testing I found that if we order ROWIDs on their own (without all the other columns), and then filter based on them, it is much faster and actually workable.
- I also realised we were capturing all logminer events for all discovered tables and only filtering them later after fetching, which was quite inefficient. Changed the logic so we filter out disabled tables as part of the query we send to logminer and have noticed a performance improvement.
- Lastly found that it is possible for a log file to have both an ARCHIVED file and an ACTIVE file, similar to ARCHIVED and CURRENT, so updated the logic to allow for that as well
- Would like to try this with a customer who is suffering from this issue before merging

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2650)

### DIFF
--- a/source-oracle/replication.go
+++ b/source-oracle/replication.go
@@ -191,22 +191,23 @@ func (s *replicationStream) addLogFiles(ctx context.Context, startSCN, endSCN in
 		logrus.WithField("file", fmt.Sprintf("%+v", f)).Debug("adding log file")
 
 		// The current log file has the same sequence as the last archived log file
-		// in this case we prefer reading from the current log file
+		// in this case we prefer reading from the current log file. It is also possible for
+		// log files to have an ACTIVE and an ARCHIVED version at the same time, we prefer ACTIVE
 		if f.Sequence == redoSequence {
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

source-oracle/replication.go
```diff
@@ -614,7 +622,7 @@ func (s *replicationStream) receiveMessages(ctx context.Context, startSCN, endSC
 	}
 
 	var totalMessages = 0
-	var relevantMessages = 0
+	var fullMessages = 0
```
mdibaiee, line 625: We no longer have "irrelevant" messages from disabled tables, but rather we may have to piece together small messages to compose a larger message which includes all the data, so changing the naming here to reflect that

williamhbaker (approved):

PR #2651, opened 2025-04-09
source-postgres-batch: Backfill XID filtering

willdonnelly:
**Description:**

The initial backfill query from a CTID+XMIN table shouldn't ever be filtered on base XID because we don't have a meaningful lower XID to filter on.

The upper XID filter will generally be applied, except when this is an initial backfill _and_ the server XID has already wrapped around, because in that case we'd filter out rows with a higher XID-4 from the previous wraparound epoch(s).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2651)


williamhbaker (approved):
LGTM

PR #2654, opened 2025-04-10
source-postgres-batch: Fix query cancellation special case

willdonnelly:
**Description:**

Just a dumb oversight, the error is wrapped so we need to use `errors.As()` for the check.


Alex-Bair (approved):
LGTM

PR #2655, opened 2025-04-10
source-postgres-batch: Fix statement timeout handling

willdonnelly:
**Description:**

I've gotten this one wrong twice now for subtle reasons so I actually sat down and put together a reliable repro case and verified that it works, however making the actual server-side query execution reliably take long enough to trigger a statement timeout requires either some ugly hacks or a ton of special-case plumbing that's only useful for that one job, so I'm not adding that test here.

However I think there is some utility to being able to override or set the default statement timeout via the advanced config, so I kept that bit in.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2655)


williamhbaker (approved):
LGTM

PR #2661, opened 2025-04-11
materialize-snowflake: only cleanup leftovers if not already cleaned up

mdibaiee:
**Description:**

- I realised we were adding old version pipe names to `currentPipes` which would break the `cleanupPipes` process because:

Let's say we have binding X, and it has two checkpoints, one X.v1 and one X.v2
X.v1 has pipe name with version A
X.v2 has pipe name with version B

We clean up X.v1 as we realise this is an older version of the same binding (`item.PipeName != b.pipeName`), but we have added "A" to currentPipes. When we run `cleanupPipes`, we loop over pipes that exist in Snowflake, and we find B, and we compare it against A, the schema and the table name match but the version doesn't, so we clean up B. This is not destructive since we only run cleanup after having finished a commit, but it unnecessarily re-creates pipes constantly.

- Also added more logging to be able to troubleshoot this better

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2661)

### DIFF
--- a/materialize-snowflake/snowflake.go
+++ b/materialize-snowflake/snowflake.go
@@ -827,8 +827,8 @@ func (d *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
 	var currentPipeNames []string
 	for stateKey, item := range d.cp {
 		if item.PipeName != "" {
-			currentPipeNames = append(currentPipeNames, item.PipeName)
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2663, opened 2025-04-13
source-oracle: track transactions across SCN ranges

mdibaiee:
**Description:**

- Prior to this pull-request we are asking Logminer to only give us committed transactions. That works well most of the time, but I just realised there is an edge case where I ask it for updates in a range [0, N], and then I ask it for updates in the range [N, M]…
It is possible that a transaction started at some point before N, but didn’t finish until some point after N. In this case, the transaction is split between the two ranges, and Logminer will not give us this transaction in either of the ranges if we ask it for committed transactions only!
The only way to get those is to not ask it for committed transactions only, which means we need to track transaction commits ourselves, and if we detect an incomplete transaction during this, we should try the next range while including the beginning of that incomplete transaction until we find its termination

Tested this by adding a test case, and watching the logs, basically what I see is:

In one of the SCN ranges, we see the `BEGIN` and `UPDATE` operation from Logminer, but we don't see `COMMIT` yet, we try again, this time keeping starting SCN of the next logminer query to include the `BEGIN` and `UPDATE` operations until we also see `COMMIT`, we then emit the full message and clean it up.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-oracle/.snapshots/TestCrossSCNTransactions
+++ b/source-oracle/.snapshots/TestCrossSCNTransactions
@@ -1,0 +1,10 @@
+# ================================
+# Collection "acmeCo/test/c__flow_test_logminer/t23135019": 2 Documents
+# ================================
+{"FULLNAME":"No Such State","POPULATION":"10000","STATE":"BB","YEAR":"1930","_meta":{"op":"c","source":{"schema":"C##FLOW_TEST_LOGMINER","snapshot":true,"table":"T23135019","row_id":"AAAAAAAAAAAAAAAAAA","rs_id":"111111111111111111","ssn":111}}}
+{"FULLNAME":"No Such State","POPULATION":"10000","STATE":"BB","YEAR":"1950","_meta":{"op":"u","source":{"ts_ms":1111111111111,"schema":"C##FLOW_TEST_LOGMINER","table":"T23135019","scn":11111111,"row_id":"AAAAAAAAAAAAAAAAAA","rs_id":"111111111111111111","ssn":111},"before":{"FULLNAME":"No Such State","POPULATION":"10000","STATE":"BB","YEAR":"1930"}}}
+# ================================
+# Final State Checkpoint
+# ================================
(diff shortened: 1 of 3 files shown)


willdonnelly:
> It is possible that a transaction started at some point before N, but didn’t finish until some point after N. In this case, the transaction is split between the two ranges, and Logminer will not give us this transaction in either of the ranges if we ask it for committed transactions only!

🤮

This is absolutely 100% going to OOM a bunch in actual production use.

> Perhaps there is some relevant prior art in the mysql/postgres/sqlserver CDC capture strategies - although I can't remember anything specifically, and it's not fresh in my mind.

There isn't really any prior art. For PostgreSQL the transaction commit reordering is handled for us -- the server-side transaction reordering is the main thing that separates logical decoding from physical replication in Postgres. For MySQL the main binlog only contains committed transactions one at a time so we don't have to handle it. And for SQL Server the log scanning worker process handles this by only copying over changes from newly committed transactions at each scan interval.

> it would be nice if we could optimistically write out the documents for an uncommitted transaction to the runtime and only emit a checkpoint when we have observed the transaction commit, but that would seem to lend itself to scenarios where there were sequences of overlapping transaction boundaries where we'd possibly never be able to find a point to commit.

This might actually be our best bet since it pushes off the buffering into a part of the runtime that is purpose-built for doing exactly that with high performance. If memory serves the logical decoding process inside PostgreSQL deals with a similar issue in order to establish valid restart_lsn points, and empirically it seems to make progress as long as no specific transaction remains open too long.

mdibaiee:
Debezium seems to suffer from the same problem:

> By default, the LogMiner adapter maintains an in-memory buffer of all running transactions. Because all of the DML operations that are part of a transaction are buffered until a commit or rollback is detected, long-running transactions should be avoided in order to not overflow that buffer. Any transaction that exceeds this configured value is discarded entirely, and the connector does not emit any messages for the operations that were part of the transaction.

https://debezium.io/documentation/reference/stable/connectors/oracle.html#oracle-logminer-mining-strategies

They offer Infinispan caching which can be configured to avoid OOMs

PR #2673, opened 2025-04-14
source-salesforce-native: fallback to REST when Bulk volume has been exceeded

Alex-Bair:
**Description:**

We've started to see production tasks hit the 24-hour rolling limit for how much data can be queried via the Bulk API 2.0 (1 TB). We should investigate ways to improve the connector's usage of the Bulk API (ex: resume reading bulk jobs that were started in the connector's previous invocation), but a simpler improvement that should be made regardless is falling back to the REST API when we receive this error.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed the connector starts using the REST API if we receive an error saying that the Bulk API's 1 TB limit has been exceeded for the current rolling 24 hour window.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2673)

### DIFF
--- a/source-salesforce-native/source_salesforce_native/api.py
+++ b/source-salesforce-native/source_salesforce_native/api.py
@@ -5,7 +5,13 @@ from typing import Any, AsyncGenerator
 from estuary_cdk.capture.common import LogCursor, PageCursor
 from estuary_cdk.http import HTTPSession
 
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #2674, opened 2025-04-14
source-mysql: Fix mysterious I/O timeouts

willdonnelly:
**Description:**

Usually backfill queries are fast. But every so often we'll have a long-running backfill query, and in MySQL land those very often fail after exactly two minutes with error:

    io.ReadFull(header) failed. err read tcp <addr>: i/o timeout: connection was bad

The fact that it always seems to be two minutes is what made me start being suspicious. And it turns out I was right to suspect, because that's actually our connector's fault.

Previously, we set connection timeouts of one minute as part of the initial connection-establishment logic so that certain classes of "database is unreachable but there's an SSH tunnel involved and so we can't actually tell that the connection dropped" of error would be reliably detected.

Unfortunately that same timeout is also applied to every other query we run on that connection, and there's no way to adjust it after the fact (because for some terrible reason the MySQL packet connection has its own separate _private_ read/write timeout properties).

We also can't use `(net.Conn).SetDeadline()` to ensure that the connection process times out correctly because the connection is not fully set up and the deadline doesn't carry over.

So okay, fine. Instead we'll abandon subtlety and orderly control flow here, since all we were going to do in the event of a timeout was fail the task anyway, and instead we'll just set up a simple timeout and if it triggers we'll kill the task directly.

This change does have one side-effect, which is that now if some backfill query were to hang indefinitely we might just keep wait- ing. We should consider explicitly setting a deadline ourselves as part of the backfill query logic if we observe that happening in a way that isn't already caught by TCP keepalives.

But for now I think this is an improvement.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2674)

### DIFF
--- a/source-mysql/main.go
+++ b/source-mysql/main.go
@@ -248,17 +248,23 @@ func (db *mysqlDatabase) connect(_ context.Context) error {
 		address = "localhost:3306"
 	}
 
+	// The MySQL client library plumbing around net.Conn timeouts is hackish and terrible
+	// and must not be trusted. However we still want to protect against hangs during the
+	// initial connection process. Since all we're going to do if the connection hangs is
+	// error out anyway, we can do this by cutting out the middleman and just setting up
+	// a directly fatal-if-not-cancelled timer.
+	var connectionTimer = time.AfterFunc(60*time.Second, func() {
+		logrus.WithField("addr", address).Fatal("failed to connect before deadline")
+	})
+	defer connectionTimer.Stop()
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

dyaffe:
Is this safe to merge?  I have a task that is waiting on it

willdonnelly:
Yep, just merged it now

PR #2680, opened 2025-04-16
source-oracle: track pending transactions and capture them separately

mdibaiee:
**Description:**

- Transactions can span across multiple SCN ranges, when that happens, we would previously keep track of their messages in memory, but for long / large transactions that would lead to memory issues.
- In order to avoid memory consumption when working with large transactions, we now keep track of the starting SCN of and the XID of pending transactions, and we persist this information as part of the checkpoint
- Once we see terminal operation for pending transactions, if it is a rollback, we remove the transaction from checkpoint, and if it is a commit, we initiate an ad-hoc capture of that specific transaction, this means we process the SCN range from the beginning of that large transaction to the commit as usual (still paginated), but this time we will ask Logminer to filter only the specific XID
- Note that I am also tracking message count for transactions. Since transactions have no association with a table, we have to inspect all transactions across the whole database, and a lot of them will be irrelevant for our use case, in these cases we will not see any other messages for those transactions because we filter for table ID, and insert / update / delete operations do have table IDs. This allows us to only initiate an ad-hoc capture if we know this table has had relevant operations for us

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2680)

### DIFF
--- a/source-oracle/backfill.go
+++ b/source-oracle/backfill.go
@@ -13,7 +13,7 @@ import (
 
 // ScanTableChunk fetches a chunk of rows from the specified table, resuming from `resumeKey` if non-nil.
 func (db *oracleDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.DiscoveryInfo, state *sqlcapture.TableState, callback func(event *sqlcapture.ChangeEvent) error) (bool, []byte, error) {
-	logrus.WithField("state", state).Debug("ScanChunk")
+	logrus.WithField("state", state).Debug("backfill: ScanChunk")
 	var keyColumns = state.KeyColumns
 	var resumeAfter = state.Scanned
 	var schema, table = info.Schema, info.Name
@@ -35,7 +35,7 @@ func (db *oracleDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.D
 		if resumeAfter != nil {
 			afterRowID = string(resumeAfter)
 		}
-		logEntry.WithField("rowid", afterRowID).Debug("scanning keyless table chunk")
+		logEntry.WithField("rowid", afterRowID).Debug("backfill: scanning keyless table chunk")
 		query = db.keylessScanQuery(info, schema, table)
 		args = []any{afterRowID}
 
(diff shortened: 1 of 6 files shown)


source-oracle/backfill.go
willdonnelly, line 1: Should `logrus.WithField("state", state).Debug("ScanChunk")` on line 16 have "backfill: " added for consistency with the other debug messages here?

source-oracle/replication.go
```diff
@@ -484,6 +534,10 @@ func (s *replicationStream) poll(ctx context.Context) error {
 		// that is small enough to not exceed the limit of files we need
 		var endSCN = startSCN + int64(s.db.config.Advanced.IncrementalSCNRange)
 
+		if maxSCN < endSCN {
+			endSCN = maxSCN + 1
```
willdonnelly, line 538: Is the `+ 1` here because endSCN is exclusive but maxSCN is inclusive, or is there some other reason it's necessary?

source-oracle/replication.go
```diff
@@ -625,46 +729,86 @@ const (
...
     SEG_OWNER NOT IN ('SYS', 'SYSTEM', 'AUDSYS', 'CTXSYS', 'DVSYS', 'DBSFWUSER', 'DBSNMP', 'QSMADMIN_INTERNAL', 'LBACSYS', 'MDSYS', 'OJVMSYS', 'OLAPSYS', 'ORDDATA', 'ORDSYS', 'OUTLN', 'WMSYS', 'XDB', 'RMAN$CATALOG', 'MTSSYS', 'OML$METADATA', 'ODI_REPO_USER', 'RQSYS', 'PYQSYS'))
-    AND ((DATA_OBJ#=0 AND DATA_OBJD#=0) OR %s)`, opPredicate, strings.Join(conditions, " OR "))
+    AND ((DATA_OBJ#=0 AND DATA_OBJD#=0) OR %s) %s`, opPredicate, strings.Join(conditions, " OR "), txPredicate)
+
+	return s.conn.PrepareContext(ctx, query)
+}
+
+func (s *replicationStream) pendingAndRollbackedTransactions(ctx context.Context, startSCN, endSCN int64) (map[string]int64, map[string]struct{}, error) {
+	// we use maps for O(1) lookup
+	var excludeXIDs = make(map[string]struct{})
```
willdonnelly, line 772: You could make the types a bit more self-explanatory in various places by defining aliases `type XID = string` and `type SCN = int64`. That would make these ones `map[XID]struct{}` and `map[XID]SCN` for instance.

source-oracle/replication.go
```diff
@@ -625,46 +729,86 @@ const (
...
+	var txs = make(map[string]int64)
```
willdonnelly, line 773: Since this function also handles rolled-back transactions it might be clearer to name this something like `pendingTXs`

PR #2682, opened 2025-04-16
sqlcapture: Refactor cursor to json.RawMessage

willdonnelly:
**Description:**

The day we kept putting off has finally come, we finally need to represent replication stream state as something more complex than a simple LSN cursor into the WAL.

This commit should have no observable effect on its own. The representation of our current string cursors in state checkpoints is unchanged, all we've done is leave the bytes makine up that quoted string unparsed inside the generic machinery and made the database-specific logic do a bit more work to quote and unquote them.

But this will allow Oracle to migrate to a more complex object as its cursor value without the need for an ugly "JSON serialized and then quoted inside another JSON string" bit of nesting.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2682)

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -37,7 +37,7 @@ var (
 // PersistentState represents the part of a connector's state which can be serialized
 // and emitted in a state checkpoint, and resumed from after a restart.
 type PersistentState struct {
-	Cursor  string                               `json:"cursor"`                   // The replication cursor of the most recent 'Commit' event
+	Cursor  json.RawMessage                      `json:"cursor"`                   // The replication cursor of the most recent 'Commit' event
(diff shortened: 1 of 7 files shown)


sqlcapture/capture.go
```diff
@@ -338,7 +338,7 @@ func (c *Capture) reconcileStateWithBindings(_ context.Context) error {
 		} else {
 			log.Info("capture has no bindings, resetting replication cursor")
 		}
-		c.State.Cursor = ""
+		c.State.Cursor = nil
```
mdibaiee, line 341: Is this going to be explicitly included in the output json to clear the value? That was the previous behaviour

willdonnelly: I believe this should work correctly. The cursor field on the state struct is not marked `omitempty` so it should always be included as an explicit null value when the `json.RawMessage` field is set to nil (see https://go.dev/play/p/3HpAYjPnn6F for demonstration of how that plays out)

mdibaiee (approved):

PR #2685, opened 2025-04-17
source-google-analytics-data-api-native: support metricAggregations

Alex-Bair:
**Description:**

Users have requested support for [`metricAggregations`](https://developers.google.com/analytics/devguides/reporting/data/v1/rest/v1beta/MetricAggregation) for reports. These aggregations are returned under separate top-level fields (like "totals") intead of within the `rows` field of the response. For example, when using a "TOTAL" metric aggregation, the response can look like:
```json
{
  "dimensionHeaders": [...],
  "metricHeaders": [...],
  "rows": [...],
  "totals": [
    {
      "dimensionValues": [
        {
          "value": "RESERVED_TOTAL"
        }
      ],
      "metricValues": [
        {
          "value": "64"
        }
      ]
    }
  ],
  ...
}

```

We can treat aggregates like the other rows inside the `rows` field when they're present, and their dimension values will indicate the type of aggregation (ex: "RESERVED_TOTAL").

**Workflow steps:**

The connector docs should be updated to reflect that `metricAggregations` are supported and provide an example of a custom report with valid `metricAggregations`.

**Notes for reviewers:**

- Aggregates are emitted as separate documents with the appropriate dimension value for the aggregate (ex: "RESERVED_TOTAL").
- Providing an invalid value in `metricAggregations` raises a validation error stating which provided value is not supported.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2685)


JustinASmith (approved):
LGTM

PR #2687, opened 2025-04-17
source-salesforce-native: discover custom share, feed, and metadata objects

Alex-Bair:
**Description:**

There are a few special custom objects that don't end in `__c`, and users are asking us to support them. This PR adds a few we're currently being asked about.

[Feeds](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_customobject__feed.htm) for custom objects end in `__Feed` and always have a `SystemModstamp` field.

Custom [metadata](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_custommetadatatype__mdt.htm) end in `__mdt` and always have a `SystemModstamp` field.

Sharing rules for custom objects end in `__Share` and always have a `LastModifiedDate` field.

There are other special, custom objects we don't support (like [change events](https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_associated_objects_change_event.htm) ending in `__ChangeEvent`), but no one is asking about these and it seems like these would need some additional special handling beyond what's already in the connector (ex: there may be query restrictions or a parent-child relationship).

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed objects ending in `__Share`, `__Feed`, and `__mdt` are discovered as possible bindings & the connector captures data for them like other incremental streams.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2687)

### DIFF
--- a/source-salesforce-native/source_salesforce_native/resources.py
+++ b/source-salesforce-native/source_salesforce_native/resources.py
@@ -10,8 +10,9 @@ from estuary_cdk.capture.common import ReductionStrategy
 from estuary_cdk.http import HTTPMixin
 
 from .supported_standard_objects import (
-    COMMON_CUSTOM_OBJECT_DETAILS,
-    COMMON_CUSTOM_OBJECT_HISTORY_DETAILS,
+    CUSTOM_OBJECT_WITH_SYSTEM_MODSTAMP_DETAILS,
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

PR #2688, opened 2025-04-17
sns materialization 

akshaymomentco:
**Description:**
This PR introduces a new materialization connector for publishing messages to AWS SNS.

The connector supports flexible authentication methods:

Static credentials (accessKey, secretKey)
IAM role assumption via roleArn
Default AWS credential chain if neither is provided

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2688)


materialize-sns/driver.go
```diff
@@ -0,0 +1,126 @@
...
+	return nil
+}
+
+func (driver) Spec(ctx context.Context, req *pm.Request_Spec) (*pm.Response_Spec, error) {
+	endpointSchema, err := json.Marshal(config{})
+	if err != nil {
+		return nil, fmt.Errorf("generating endpoint schema: %w", err)
+	}
+	return &pm.Response_Spec{
+		ConfigSchemaJson: json.RawMessage(endpointSchema),
```
williamhbaker, line 49: This needs a `ResourceConfigSchemaJson` as well.

materialize-sns/driver.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	// Either RoleARN OR (AccessKey + SecretKey) must be provided
+	if c.RoleARN == "" && (c.AccessKey == "" || c.SecretKey == "") {
+		return fmt.Errorf("either roleArn or both accessKey & secretKey must be provided")
+	}
+	return nil
+}
+
+func (driver) Spec(ctx context.Context, req *pm.Request_Spec) (*pm.Response_Spec, error) {
+	endpointSchema, err := json.Marshal(config{})
```
williamhbaker, line 44: This is not a valid JSON schema - see here for an example of how this is done: https://github.com/estuary/connectors/blob/867cff323173e3023647fd4b2949b0946ee4ea66/materialize-google-pubsub/driver.go#L98-L120

materialize-sns/driver.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("generating endpoint schema: %w", err)
+	}
+	return &pm.Response_Spec{
+		ConfigSchemaJson: json.RawMessage(endpointSchema),
+	}, nil
+}
```
williamhbaker, line 51: The `Driver` is missing a `Validate` method, in which it will need to calculate constraints and resource paths, see here for an example: https://github.com/estuary/connectors/blob/867cff323173e3023647fd4b2949b0946ee4ea66/materialize-google-pubsub/driver.go#L122-L185

materialize-sns/driver.go
```diff
@@ -0,0 +1,126 @@
...
+	snsClient := sns.New(sess)
+
+	return &transactor{
+		snsClient: snsClient,
+		topicARN:  cfg.SNSTopicARN,
+	}, &pm.Response_Opened{}, nil, nil
+}
+
+func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
+	for it.Next() {
```
williamhbaker, line 106: Is there a way to send requests in bulk? It's very unlikely that sending them one by one is going to be performant.

materialize-sns/driver.go
```diff
@@ -0,0 +1,126 @@
...
+		msg, err := json.Marshal(it.RawJSON)
```
williamhbaker, line 107: `it.RawJSON` is already a `json.RawMessage`, so this `json.Marshal` is not necessary.

materialize-sns/driver.go
```diff
@@ -0,0 +1,126 @@
...
+func (driver) Spec(ctx context.Context, req *pm.Request_Spec) (*pm.Response_Spec, error) {
+	endpointSchema, err := json.Marshal(config{})
+	if err != nil {
+		return nil, fmt.Errorf("generating endpoint schema: %w", err)
+	}
+	return &pm.Response_Spec{
+		ConfigSchemaJson: json.RawMessage(endpointSchema),
+	}, nil
+}
+
```
williamhbaker, line 52: The driver also seems to be missing an `Apply` method.

materialize-sns/driver.go
```diff
@@ -0,0 +1,126 @@
...
+	m "github.com/estuary/connectors/go/protocols/materialize"
+	boilerplate "github.com/estuary/connectors/materialize-boilerplate"
+	pf "github.com/estuary/flow/go/protocols/flow"
+	pm "github.com/estuary/flow/go/protocols/materialize"
+)
+type driver struct{}
+
+type config struct {
+	SNSTopicARN string `json:"snsTopicArn" jsonschema:"title=SNS Topic ARN,description=AWS SNS Topic ARN to publish messages to."`
```
williamhbaker, line 22: Materializations must support having multiple bindings, and I think each one would have a separate topic. It probably wouldn't have a direct ARN input; rather the name of a topic, and then it would have to manage creating topics that were needed and then looking up their ARNs.

williamhbaker (changes requested):
Left a few comments. I'd recommend checking out our Google PubSub connector code as a point of reference, since this one seems like it should be quite similar to that: https://github.com/estuary/connectors/tree/main/materialize-google-pubsub

PR #2689, opened 2025-04-17
source-firestore: Don't interrupt ongoing backfills

willdonnelly:
**Description:**

When change streaming fails to catch up, we allow the change stream to "skip ahead" and flag the binding as inconsistent so we can run another backfill in the future.

But previously, if change streaming failed repeatedly during an ongoing backfill, that backfill would keep getting interrupted and rescheduled for 24 hours in the future. This meant that a really flaky stream in conjunction with frequent task restarts could prevent the backfill from ever getting anything done, because it would keep on getting through just a small bit of the data, then the task would restart, we'd see the 'Inconsistent: true' flag, and we'd interrupt the backfill and schedule it to start over from the top the next day.

The solution is that we should never interrupt an ongoing backfill, once started we should see it through and then run _another_ one afterwards (subject to our minimum delay between backfills) to resolve the inconsistent flag.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2689)

### DIFF
--- a/source-firestore/pull.go
+++ b/source-firestore/pull.go
@@ -81,7 +81,7 @@ func (driver) Pull(open *pc.Request_Open, stream *boilerplate.PullOutput) error 
 		}
 	}
 
-	updatedResourceStates, err := initResourceStates(prevState.Resources, open.Capture.Bindings)
+	updatedResourceStates, err := initResourceStates(prevState.Resources, open.Capture.Bindings, time.Now())
 	if err != nil {
 		return fmt.Errorf("error initializing resource states: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


source-firestore/pull.go
```diff
@@ -243,12 +216,57 @@ func initResourceStates(prevStates map[boilerplate.StateKey]*resourceState, bind
...
+		}
+
+		// Since we're already inconsistent we can't make it worse, and resetting the read time
+		// here makes it slightly more likely that we'll remain caught up going forward.
+		state.ReadTime = now
+
+		// If we're inconsistent but there's an ongoing backfill, we need to let it finish,
+		// otherwise we need to initialize a new backfill.
+		if prevState.Backfill != nil && !prevState.Backfill.Completed {
+			state.Backfill = prevState.Backfill
```
Alex-Bair, line 239: If we're inconsistent but there's an ongoing backfill, does `state.Inconsistent` need set to `true` here to ensure we schedule another backfill after the current one completes? Or maybe I missed something elsewhere that ensures that?

Alex-Bair (approved):
LGTM % one question about making sure another backfill is triggered if there's already an ongoing backfill when we're inconsistent.

source-firestore/pull.go
willdonnelly: Oh, that's a very good catch. Ironically the very test I added in this PR would have told me I got that wrong, if I had paid enough attention. Fixed.

PR #2690, opened 2025-04-18
materialize-dynamodb: no option for delta updates

williamhbaker:
**Description:**

The delta updates mode for DynamoDB never really made sense, and was basically assured to fail if & when multiple documents with the same key show up in the same transaction, which is a thing that regularly happens.

There's not any valid use case for delta updates with DynamoDB that can be presently conceived of, so this removes it from the resource configuration entirely.

**Workflow steps:**

See https://github.com/estuary/flow/pull/2086

**Notes for reviewers:**

Noting that this will require manually updating existing configurations that specify delta updates in their resource config, but this should be managable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2690)


Alex-Bair (approved):
LGTM

PR #2691, opened 2025-04-18
sqlcapture: Logging tweaks

willdonnelly:
**Description:**

This commit makes two logging changes:
1. The sqlcapture logic will now maintain separate counts of the important event types and log the counts separately rather than a single opaque "events" count.
2. Fixes a handful of spots I noticed while testing the previous change where serialized JSON values were being logged as byte arrays `[65 66 67]` which is ugly. Most of these are new from the cursors-as-JSON refactoring I did yesterday, so I figured I'd just address those real quick while I was at it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2691)


williamhbaker (approved):
LGTM

PR #2692, opened 2025-04-18
materialize-iceberg: always create identity columns as required

williamhbaker:
**Description:**

Materializations do not support nullable keys in general, and materialize-iceberg is no exception.

A non-required source collection field is allowed as a key if it has a default value set, and so if a key field has been selected we know it will never be null and should create the column for the key field accordingly. On a practical level, this is necessary since key fields are created as identity columns which cannot be not required.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2692)


jonwihl (approved):
LGTM

PR #2693, opened 2025-04-18
source-mysql: Add "tinyint1_as_bool" feature flag

willdonnelly:
**Description:**

MySQL doesn't technically have a boolean type. If you declare a column of type `BOOLEAN` what actually happens is it gets treated as an alias for `TINYINT(1)` and similarly `TRUE` and `FALSE` are just aliases for one and zero.

However, in practice nobody would ever bother adding a display width parameter of 1 on a tinyint _except_ as the idiomatic way of getting a MySQL boolean column, so we can reliably assume that if we see a tinyint column that is specifically `TINYINT(1)` the user intended for it to be treated as a boolean.

This PR adds a "tinyint1_as_bool" feature flag, currently defaulting to off. When this flag is enabled, columns of type `TINYINT(1)` and only `TINYINT(1)` will be treated as booleans in both discovery and value translation.

We might want to make this the default going forward, but that's a separate concern.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2693)

### DIFF
--- a/source-mysql/.snapshots/TestFeatureFlagTinyintAsBool-Default-Capture
+++ b/source-mysql/.snapshots/TestFeatureFlagTinyintAsBool-Default-Capture
@@ -1,0 +1,16 @@
+# ================================
+# Collection "acmeCo/test/test/featureflagtinyintasbool_21925183": 8 Documents
+# ================================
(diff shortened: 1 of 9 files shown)


williamhbaker (approved):
LGTM

PR #2696, opened 2025-04-21
source-postgres: Fix XMIN backfill oversight

willdonnelly:
**Description:**

Two issues, both of which I'm a bit confused how I failed to see them in the previous XMIN backfill tweaks PR but whatever:

- The lower-bound filter clause needs to be `>=` rather than `>` in order for our XMIN backfill test case to pass correctly.
- The query generation snapshots need to be updated

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2696)


Alex-Bair (approved):
LGTM

PR #2698, opened 2025-04-21
materialize-snowflake: support for Snowpipe streaming

williamhbaker:
**Description:**

Adds a new mode of operation for Snowflake bindings that use delta updates - [Snowpipe streaming](https://docs.snowflake.com/en/user-guide/data-load-snowpipe-streaming-overview). As implemented here, this will use the post-commit apply pattern to achieve exactly-once semantics.

This includes a fair amount of code ported over from the Java Snowflake Ingest SDK for the specifics related to writing the data files and interacting with the Streaming HTTP API to write data from those files, which is primary in the `bdec.go` and `stream_http.go` file. Our materialization-specific logic is in `stream.go`, as well as being sprinkled in the 2 previously mentioned ones. For example, we aren't reading the entirety of a transaction into memory like the Java SDK does, and instead use a byte-streaming version of the data writing & encryption that is in `bdec.go`.

The individual commit messages have more details, and I tried to lay them out in a reasonable way for reviewing.

An overall objective of this set of changes is to make the code as defensive as possible, since bugs on our end are more likely to result in corrupt tables that require backfilling when using streaming, rather than possibly getting a useful error message back from Snowflake like we (sometimes) do when running queries. To that end, streaming will only be enabled for now if the binding uses delta updates & JWT authentication, _and_ the materialization has the `snowflake_streaming` feature flag set. 

For now there are 3 possible modes of operation for a delta updates binding:
1. Using `COPY INTO` queries, if password auth is used. 
2. Using "regular" Snowpipe, if the `snowflake_streaming` feature flag is not set or the table doesn't support streaming
3. Using Snowpipe streaming if it is and does

Long-term, I'm not sure if we need all 3 of those. There are [limitations](https://docs.snowflake.com/en/user-guide/data-load-snowpipe-streaming-overview#limitations) on tables that support Snowpipe streaming, so we'll never be able to use it 100% of the time for delta updates tables, but it will probably be quite rare that we can't use it. A possible eventual strategy would be to use Snowpipe streaming if possible, and if not fall back to "something", where that "something" is either regular Snowpipe or `COPY INTO`. For now we'll keep them all though, since we still need them.

My thought is to keep Snowpipe streaming gated behind the feature flag for a while, until we get some opt-in usage of it to verify that it works as expected. Then it would be possible to enable it for all new materializations, and after a while even enable it retroactively on existing ones.

Closes https://github.com/estuary/connectors/issues/2676

**Workflow steps:**

The docs will eventually need some updates for Snowpipe streaming, although its usage should be pretty transparent to the user once the feature flag is enabled by default

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2698)

### DIFF
--- a/materialize-boilerplate/stream-encode/parquet.go
+++ b/materialize-boilerplate/stream-encode/parquet.go
@@ -8,14 +8,17 @@ import (
 	"math"
 	"math/big"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
 	"unsafe"
 
+	"github.com/apache/arrow-go/v18/arrow/decimal128"
 	"github.com/apache/arrow-go/v18/parquet"
 	"github.com/apache/arrow-go/v18/parquet/compress"
 	"github.com/apache/arrow-go/v18/parquet/file"
+	"github.com/apache/arrow-go/v18/parquet/metadata"
 	"github.com/apache/arrow-go/v18/parquet/schema"
 	"github.com/google/uuid"
 	"github.com/segmentio/encoding/json"
@@ -77,7 +80,7 @@ func newRowSizing(sch ParquetSchema) rowSize {
 			rs.fixed += 8
 		case PrimitiveTypeBoolean:
 			rs.fixed += 1
-		case PrimitiveTypeBinary, LogicalTypeJson:
+		case PrimitiveTypeBinary, LogicalTypeJson, LogicalTypeDecimal:
 			rs.fixed += 24 // slice header
 			rs.calcLen = append(rs.calcLen, idx)
 		case LogicalTypeString, LogicalTypeUuid, LogicalTypeDate, LogicalTypeTime, LogicalTypeTimestamp, LogicalTypeInterval:
@@ -129,6 +132,7 @@ type parquetConfig struct {
 	disableDictionaryEncoding bool
 	rowGroupRowLimit          int
 	rowGroupByteLimit         int
+	metadata                  map[string]string
 }
 
 type ParquetEncoder struct {
@@ -181,6 +185,12 @@ func WithParquetRowGroupByteLimit(n int) ParquetOption {
 	}
 }
 
+func WithParquetMetadata(meta map[string]string) ParquetOption {
+	return func(cfg *parquetConfig) {
+		cfg.metadata = meta
+	}
+}
+
 func NewParquetEncoder(w io.WriteCloser, sch ParquetSchema, opts ...ParquetOption) *ParquetEncoder {
 	cfg := parquetConfig{
 		compression:       Uncompressed,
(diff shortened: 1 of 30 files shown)


materialize-snowflake/snowflake.go
```diff
@@ -465,27 +505,61 @@ func (d *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
...
-	// Upload the staged files and build a list of merge and copy into queries that need to be run
-	// to effectively commit the files into destination tables. These queries are stored
-	// in the checkpoint so that if the connector is restarted in middle of a commit
-	// it can run the same queries on the next startup. This is the pattern for
-	// recovery log being authoritative and the connector idempotently applies a commit
-	// These are keyed on the binding table name so that in case of a recovery being necessary
-	// we don't run queries belonging to bindings that have been removed
-	for _, b := range d.bindings {
+	return func(ctx context.Context, runtimeCheckpoint *protocol.Checkpoint) (*pf.ConnectorState, m.OpFuture) {
+		var checkpointJSON, err = d.buildDriverCheckpoint(ctx, runtimeCheckpoint)
```
williamhbaker, line 530: I actually just realized that I was wrong in an earlier commit message about this being able to run in the async part of the transaction - this is the StartCommitFn, which is invoked shortly after `Store` returns. So that part of the justification for the split out `buildDriverCheckpoint` function isn't actually valid. But, the Snowpipe streaming checkpoints do need the runtimeCheckpoint, so I think the restructuring still makes sense.

materialize-snowflake/stream.go
```diff
@@ -0,0 +1,536 @@
...
+}
+
+// write registers a series of blobs to the table. All the blobs must be for the
+// same table.
+//
+// Exactly-once semantics are achieved using the offset tokens for the blobs:
+// The offset token consists of the "base" token which is a string, and for each
+// blob it is appended with a sequential counter, starting at 0. So if there are
+// multiple blobs to append, they are organized with offset tokens like
+// "basetoken:0", "basetoken:1", "basetoken2:" etc.
```
mdibaiee, line 194:
```suggestion
// "basetoken:0", "basetoken:1", "basetoken:2" etc.
```

materialize-snowflake/stream.go
```diff
@@ -0,0 +1,536 @@
...
+	}
+
+	var schema = blobs[0].Chunks[0].Schema
+	var table = blobs[0].Chunks[0].Table
+	var channelName = blobs[0].Chunks[0].Channels[0].Channel
+	var thisChannel *channel
+	for _, v := range sm.tableStreams {
+		matches := v.channel.Schema == schema && v.channel.Table == table && v.channel.Channel == channelName
+		if matches && thisChannel != nil {
+			return fmt.Errorf("internal error: found duplicate duplicate channel %s in tableStreams", channelName)
```
mdibaiee, line 215:
```suggestion
			return fmt.Errorf("internal error: found duplicate channel %s in tableStreams", channelName)
```

mdibaiee (approved):
LGTM % some nits

materialize-boilerplate/stream-encode/parquet.go
```diff
@@ -237,15 +258,24 @@ func writerOpts(cfg parquetConfig) file.WriteOption {
...
-		// scratch file.
-		scratchOpts := []parquet.WriterProperty{parquet.WithDictionaryDefault(false)}
+
+		scratchOpts := []parquet.WriterProperty{
+			// Don't use dictionary encoding for the scratch file, since it is
+			// much slower to write than plain encoding with the small row
+			// groups of the scratch file.
+			parquet.WithDictionaryDefault(false),
+			// Turning off stats calculations helps too, but just a tiny bit.
+			parquet.WithStats(false),
```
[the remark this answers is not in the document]
williamhbaker: Hmm since the scratch file is purely ephemeral with no way for us to ever access it, I don't think we could get at the stats even if we wanted to. The stats are calculating things like min/max values for columns etc. for meta queries, and the scratch file is only read from column-by-column by the writer when constructing the output. We do enable stats for the output file, so I think we could get what we wanted from that if it ever came up.

materialize-snowflake/stream.go
williamhbaker: Some kind of subliminal force at work here where I typed the word "duplicate" twice 😃

PR #2699, opened 2025-04-21
estuary-cdk: support refresh token rotation & easier event logging

Alex-Bair:
**Description:**

The individual commits have additional details, but the overall scope for this PR includes:
- Supporting connectors that need to rotate refresh tokens within their config.
- Making logging connector events easier within the CDK.

The features in this PR are to support an upcoming `source-outreach` connector. I split the CDK changes out into this PR to decouple them with the release of `source-outreach` & give some time to ensure nothing goes sideways when rolling out these changes.

### Rotating refresh tokens

Some SaaS connectors that use OAuth2 require refresh tokens to be rotated every so often. This could be based on some time duration or total usage. For example, Outreach [says](https://developers.outreach.io/api/getting-started/#token-availability) their refresh tokens expire 2 weeks after creation or after 100 access tokens are generated.

To handle these situations where the connector config's credentials need periodically updated by the connector itself, the CDK emits a `configUpdate` event (which is just an info level log with certain fields present) for the control plane to republish the connector with an updated config.

When a connector uses `RotatingOAuth2Credentials`, there are two refresh tokens. One is in the config, and another is in the connector's state. When the config token is no longer valid for whatever reason, the connector uses the token in its state to get new refresh tokens for its config and state. It also uses our config encryption endpoint to `sops` encrypt
an updated config with the new config token & emits the updated config in a `configUpdate` event.

**Workflow steps:**

**Notes for reviewers:**

- refresh tokens are rotated when the config's token is invalid. 
- events are logged out in the format expected by the control plane.

This change relies on the control plane updates in https://github.com/estuary/flow/pull/2091 to republish captures that emit `configUpdate` events.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2699)


estuary-cdk/estuary_cdk/capture/base_capture_connector.py
```diff
@@ -0,0 +1,227 @@
...
+                self.token_source.credentials.refresh_token = open.state.refresh_token
+
+                # Get a new token for the config.
+                new_config_token = (await self.token_source.initialize_oauth2_tokens(
+                    log,
+                    self,
+                )).refresh_token
+
+                # Get a new token for the state. This is done second to ensure the state 
+                # token will still be valid if the config token is expired .
```
williamhbaker, line 207: Extra space before the dot at the end of this comment

williamhbaker (approved):
LGTM

estuary-cdk/estuary_cdk/capture/base_capture_connector.py
```diff
@@ -0,0 +1,227 @@
...
+            )
+        )
+
+        self._emit(r)
+
+    async def _encrypt_config(
+        self,
+        log: FlowLogger,
+        config: EndpointConfig,
+    ) -> str:
```
jgraettinger, line 156: typing says the return value is a string, but you're using json.loads here -- am I missing something?

Context: i was trying to determine if you're string-wrapping the JSON-encoded endpoint config payload in configUpdate logs emitted by the CDK, or if those go through as regular JSON objects.

A call-out you may already be aware of: Flow does *not* preserve the arbitrary order of properties in objects that it processes -- instead, we always sort them lexicographic-ally. This can trip up `sops`, which attaches an HMAC that relies on encountered property order of endpoint config properties _other_ than the added `sops` stanza.

There are two options:
* (Preferred) Ensure endpoint configs submitted to the encryption service are already sorted. Log endpoint configs as nested regular JSON objects in the updateConfig message. Then, Flow's enforced lexicographic ordering already agrees with the `sops` HMAC and everybody's happy.
* Don't expect endpoint configs are sorted, and string-wrap their JSON when logged out as `updateConfig`.

Alex-Bair: I missed updating that function's return value type to `dict[str, Any]` when I was testing. I'll update that.

I wasn't aware of Flow lexicographically sorting object properties (or if I was, I didn't remember). Thanks for mentioning that. I'll update to sort the config's properties before sending it off to be encrypted, then log the encrypted endpoint config as a regular nested JSON object in the `configUpdate` event.

Alex-Bair:
I realized that the import path for `BaseCaptureConnector` in `source-netsuite` will also need updated due to me moving `BaseCaptureConnector` into a separate file. Instead of doing that, I'm going to refactor how the CDK's `capture` directory is structured so connectors' current import paths aren't broken by this change.

estuary-cdk/estuary_cdk/capture/base_capture_connector.py
Alex-Bair: I've made those updates. Thanks again for the feedback!

PR #2700, opened 2025-04-22
source-sqlserver: Truncate oversized fields to 8MiB

willdonnelly:
**Description:**

Same as we do for PostgreSQL, this PR adds a simple "if length > 8MiB truncate to 8MiB" check for SQL Server's "Large Object" types (text and image and varchar/varbinary(max)). This doesn't make it impossible for a document to exceed Flow's 64MiB limit, but it's a decent starting point that catches the common case where there's just one really huge column in a table.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2700)

### DIFF
--- a/source-sqlserver/datatypes.go
+++ b/source-sqlserver/datatypes.go
@@ -36,6 +36,10 @@ const (
 	// field is ignored and we just use the original input string, since that
 	// way the collation-sortable encoding doesn't have to be reversible.
 	collatedTextTag = "ctext"
+
+	// The maximum size of a single column's value. Values larger than this threshold
+	// will be truncated to fit. Arbitrarily set at 8MiB.
+	truncateColumnThreshold = 8 * 1024 * 1024
 )
 
 func encodeKeyFDB(key, ktype interface{}) (tuple.TupleElement, error) {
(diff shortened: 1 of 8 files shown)


source-sqlserver/datatypes.go
```diff
@@ -143,6 +147,15 @@ func (db *sqlserverDatabase) translateRecordField(columnType interface{}, val in
 			}
 			return u.String(), nil
 		}
+		if len(val) > truncateColumnThreshold {
+			val = val[:truncateColumnThreshold]
+		}
+		return val, nil
+	case string:
+		if len(val) > truncateColumnThreshold {
+			val = val[:truncateColumnThreshold]
```
williamhbaker, line 156: Is it possible / would it be a problem for this to split a multi-byte character?

williamhbaker (approved):
LGTM, see question

source-sqlserver/datatypes.go
williamhbaker: Ah yeah that makes sense. Since the JSON serialization won't produce invalid UTF-8 (I suppose it can't, by definition) then it doesn't seem like we need to worry about this anymore.

PR #2705, opened 2025-04-22
source-sqlserver: Fix rare map assignment panic

willdonnelly:
**Description:**

I have only ever seen this once in production, but since we run discovery at a regular cadence it really ought to be bulletproof and not able to randomly cause a panic if the computed-columns info and the main column metadata are out of sync.

### DIFF
--- a/source-sqlserver/discovery.go
+++ b/source-sqlserver/discovery.go
@@ -88,7 +88,10 @@ func (db *sqlserverDatabase) DiscoverTables(ctx context.Context) (map[sqlcapture
 			details.ComputedColumns = computedColumns[streamID]
 		}
 		for _, columnName := range computedColumns[streamID] {
-			var info = table.Columns[columnName]
+			var info, ok = table.Columns[columnName]
+			if !ok {
+				continue
+			}
 			info.OmitColumn = true
 			table.Columns[columnName] = info
 		}


source-sqlserver/discovery.go
```diff
@@ -88,7 +88,10 @@ func (db *sqlserverDatabase) DiscoverTables(ctx context.Context) (map[sqlcapture
 			details.ComputedColumns = computedColumns[streamID]
 		}
 		for _, columnName := range computedColumns[streamID] {
-			var info = table.Columns[columnName]
+			var info, ok = table.Columns[columnName]
+			if !ok {
+				continue
+			}
```
mdibaiee, line 94: Do we have any idea why they are out of sync?

willdonnelly: I don't have a bulletproof explanation, but I have some observations:
1. They're coming from different information tables. The main column listing is based on `information_schema.columns` while the computed-columns info comes from `sys.columns`. So it's theoretically possible those could differ due to permissions or something.
2. The discovery queries aren't all run in a single transaction, so it's possible for them to be mismatched if the user is modifying a table at the precise moment we're running discovery.
3. In the particular case I observed, the panic happened once and then didn't reoccur, so I suspect (2) was actually the case.

mdibaiee: Isn't it better to re-try in this case until we get in-sync values? otherwise we are giving them unstable discovery results which may change on a second run I guess? 🤔

willdonnelly: No, the ordering of operations here is fine, I think. The main column metadata listing is authoritative, and if a column wasn't present in that listing then we simply shouldn't be augmenting it with the computed-column information either. The results are only "unstable" inasmuch as the user has just changed something and the next discovery run might contain an extra column, but that's the same situation we'd be in if the current discovery run had simply executed a few seconds sooner.

This is also consistent with how we handle columns-on-tables when assembling discovery results: if a table wasn't returned by the main "list tables" query then we skip augmenting it with more detailed information from the other queries.

mdibaiee (approved):

PR #2706, opened 2025-04-22
source-klaviyo: address eventual consistency with `events`

Alex-Bair:
**Description:**

We've observed eventual consistency with the Klaviyo API's `/events` endpoint. To avoid asking for data while Klaviyo is inconsistent, we set an upper bound on how recent the data can be. This limits how "real-time" the connector is, but these Airbyte imports are often relatively slow anyway since streams don't run concurrently, and if we wanted a more real-time solution, we should build a native connector.

I ran a few tests to find how "eventual" Klaviyo's `/events` endpoint is. After testing some possible `LAG` durations, I ended up using 110 minutes of lag to ensure results from Klaviyo are consistent. It's possible even this isn't sufficient, but I couldn't' find any evidence of events appearing after 110 minutes of lag during my (albeit somewhat limited) testing.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed `events` still works & places an upper bound in the `filter` parameter for API requests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2706)

### DIFF
--- a/source-klaviyo/source_klaviyo/streams.py
+++ b/source-klaviyo/source_klaviyo/streams.py
@@ -4,6 +4,7 @@
 
 import urllib.parse
 from abc import ABC, abstractmethod
+from datetime import datetime, timedelta, UTC
 from typing import Any, Iterable, List, Mapping, MutableMapping, Optional, Union
 
 import pendulum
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM! Would the `greater-than` vs `greater-or-equal` cause any unwanted side affects? I assume not since you are subtracting 1 second when we use the `greater-than` filter to combat the mostly likely case where the updated at is at the same second as our log cursor. That's the only case I can think off, but wanted to ask this question just in case!

Alex-Bair:
The main side effect of using `greater-than` and `greater-or-equal` is that we'll emit duplicate documents when the connector restarts. We could avoid emitting duplicates if we had better control over when documents are checkpointed or checkpointed the page cursors instead. But those changes are more difficult to make in these imported connectors. I figure that some duplicate data is ok for now since we'll eventually make a native version with our CDK & we'll be able to avoid emitting duplicate documents then.

PR #2709, opened 2025-04-23
source-outreach: new connector

Alex-Bair:
**Description:**

This PR introduces an initial version of `source-outreach` to capture data from [Outreach](https://developers.outreach.io/guides/).

All endpoints are queried in extremely similar ways, and that made the actual connector logic fairly simple. 

The main challenges with this connector were supporting refresh token rotation & capturing `dataConnections` for various streams.

#### Refresh token rotation

Per Outreach's docs and questions I made to their support team, refresh tokens expire 14 days after creation or after 100 more access/refresh tokens are generated. This requires persisting new refresh tokens somehow in the connector's published endpoint config. The CDK handles this when a connector uses `RotatingOAuth2Credentials` introduced in https://github.com/estuary/connectors/pull/2699.

#### `dataConnections`

Some Outreach entities have some `dataConnections` metadata that indicates if a record has a connection to some non-Outreach system. Per Fivetran's [docs](https://fivetran.com/docs/connectors/applications/outreach#dataconnectionlimitationsforoutreachsalesforceintegration), adding a data connection to a record doesn't cause that record's `updatedAt` to change.

Currently, the connector doesn't try to find these newly added data connections. I haven't been able to replicate the behavior Fivetran describes in our dev Outreach account, so I want to wait until users start using the connector, confirm that new `dataConnections` don't update a record's `updatedAt`, then implement some strategy to capture them. Likely, we'll have to periodically backfill with `ResourceConfigWithSchedule` to reliably capture these added data connections.

In our Outreach dev account, our OAuth app will need released as a production app that others can install & use. There's a review step that Outreach has to perform, so I'll get that moving after merging this connector. Our production Supabase account will also need updated with our Outreach OAuth app's client ID & client secret once we start that review process.

**Workflow steps:**

Documentation needs to be created for `source-outreach`. 

**Notes for reviewers:**

- All streams do not crash (i.e. we're making valid API calls).
- The connector incrementally captures data for streams that include data in our dev Outreach account.
- When the refresh token in the endpoint config is invalid, the connector uses the refresh token in its state to get two new refresh tokens & emits a `configUpdate` event containing a config with a valid refresh token. 

`source-outreach` uses the new CDK features introduced in https://github.com/estuary/connectors/pull/2699. That PR needs to be merged into main before this PR is rebased & merged.


source-outreach/tests/test_snapshots.py
```diff
@@ -0,0 +1,44 @@
+import json
+import subprocess
+
+
```
Alex-Bair, line 4: There's no capture snapshot since credentials are not long-lived. It would be a hassle to constantly update `config.yaml` every 2 weeks with a new refresh token to get the capture snapshot to pass, so I opted to omit it instead.

JustinASmith:
@Alex-Bair  I will start looking at this today!

Alex-Bair:
Thanks! This one isn't a huge rush. I anticipate I'll need to address comments on the related control plane PR before I would release this connector anyway.

JustinASmith (approved):
LGTM! The connector seems pretty straight forward other than the authentication aspects, so I didn't have a ton to say about this one. I noticed we only have incremental resources, which is great.

PR #2710, opened 2025-04-23
estuary-cdk: support `SourcedSchema` messages

Alex-Bair:
**Description:**

The Flow runtime has recently been updated to support `SourcedSchema` messages - schemas emitted by capture connectors to inform the runtime of some source-defined schema. See https://github.com/estuary/flow/pull/1946 for more details.

This PR adds support in our CDK for emitting `SourcedSchema` messages. `SourcedSchema` messages are written to each task's buffer and emitted as part of the next checkpoint.

`SourcedSchema` messages should be emitted after `open` is called. They can be emitted when opening a resource, like below:

```python
def open(
    binding_index: int,
    state: ResourceState,
    task: Task,
    all_bindings,
):
    schema = {
        "additionalProperties": False,
        "type": "object",
        "properties": {
            "myPropertyName": {
                "type": "string",
                "format": "integer",
            }
        }
    }
    # Emit the schema.
    task.sourced_schema(binding_index, schema)
    # Checkpoint an empty connector state so the SourcedSchema response will be applied to the inferred schema.
    task.checkpoint(state=ConnectorState())
```

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that `SourcedSchema`s emitted when opening a resource causes the binding's inferred schema to widen & accommodate whatever fields were specified in the `SourcedSchema`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2710)

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -60,6 +60,17 @@ and "no pages remain" in a response context.
 class Triggers(Enum):
     BACKFILL = "BACKFILL"
 
+@dataclass
+class SourcedSchema:
+    """
+    SourcedSchema encapsulates a source-defined schema for a specific binding.
+    SourcedSchemas are used to inform the runtime about types & metadata that
+    aren't easily inferred by inspecting specific documents. The runtime
+    widens inferred schemas to accommodate the current inferred schema along
+    with any SourcedSchemas.
+    """
+    value: dict[str, Any]
+
 
 class BaseDocument(BaseModel):
     class Meta(BaseModel):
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM, as an isolated change. It will be interesting to see a capture actually use this, and I am wondering about the ergonomics of making the `dict[str, any]` of the schema. There may not be much that can be done for that though.

Alex-Bair:
I considered different ways to make it easier to specify the schema for a generic connector, but I couldn't come up with something I liked. The connector will always have to translate however the source specifies the schema to a JSON schema. I figured until this gets some use, requiring the connector to build that `dict[str, Any]` is ok for now, and we can refactor later once this gets some use.

PR #2713, opened 2025-04-24
materialize-iceberg: advanced config option to create all table columns as lowercase

williamhbaker:
**Description:**

Querying S3 table bucket tables with Athena doesn't work unless all the column names are lowercase. This seems pretty dumb and hopefully AWS fixes that eventually, but for now we have to accommodate it since that is a pretty obvious use case if you are using S3 tables.

We already had the general restriction that column names that differed only in caps weren't allowed since Glue etc. don't like that, so adding this option is too disruptive.

**Workflow steps:**

The docs will need updated for this new config option, and should mention something about it in the section about S3 table buckets.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2713)


materialize-iceberg/.snapshots/TestValidateAndApplyLowercaseColumnNames
```diff
@@ -0,0 +1,362 @@
+Big Schema Initial Constraints:
```
williamhbaker, line 1: This snapshot is exactly the same as the `TestValidateAndApply` one, except in the table schema outputs (note: not the constraint outputs, since those are based on Flow field names) the column names are lowercase - which is the expected outcome.

Alex-Bair (approved):
LGTM

PR #2717, opened 2025-04-24
source-zendesk-support-native: validate subdomain input

Alex-Bair:
**Description:**

If users use OAuth to authenticate, they need to input their subdomain before authenticating since the subdomain is used in some of the OAuth related URLs. If users enter an invalid subdomain, they'll get an error that may look like OAuth doesn't work when the actual problem is that they provided the wrong subdomain. There's not a ton we can do to ensure they enter the _correct_ subdomain, but we can add some input validation to at least show some indication if the subdomain they provided is definitely wrong (like providing the full URL instead of just the subdomain).

This PR adds basic validation to make sure subdomains:
- are 3-63 characters long.
- start with a lowercase letter.
- contain only lowercase letters, digits, or dashes.

These restrictions are stated in [this](https://support.zendesk.com/hc/en-us/articles/4408845973914-Renaming-your-subdomain?page=1&sort_by=created_at&sort_order=desc#topic_m5k_pzs_nvb) Zendesk article.

I also updated a handful of capture snapshots that were failing because the source system added fields. I had to do that for `source-zendesk-support-native` to get the test suite to pass, so I did a few more while I was at it.

**Workflow steps:**

**Notes for reviewers:**

Confirmed the `subdomain` input only accepts values that meet the criteria stated above, and red "must match $PATTERN" subtext is shown if an invalid looking subdomain is entered.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2717)


williamhbaker (approved):
LGTM

PR #2720, opened 2025-04-25
materialize-motherduck: add GCS as a staged file option

williamhbaker:
**Description:**

Adds GCS as an option for the staging bucket the `materialize-motherduck`. Previously only S3 was supported.

The commits tell the story of the changes that lead up to this. The summary is that this is the second case of us needing an abstraction over a blob storage provider - the first being Snowpipe streaming. It is used here in a more typical materialization way, where files are staged, read from, and then deleted. `StagedFiles` is a relatively recent abstraction for doing this, so I am updating the MotherDuck materialization to use `StagedFiles`, and updating `StagedFiles` to use `blob.Bucket` so that blob storage provider can be abstracted. A primary knock-on effect of this is that `materialize-iceberg` uses `StagedFiles` and it also needs to list objects, so that motivated adding a `List` method (in addition to `Delete`) to `blob.Bucket`.

There's a version bump of the connector because the previous config structure is not compatible with the new `oneOf` structure for the staging bucket. I'll do a manual migration of existing tasks, which should be managable.
**Workflow steps:**

Will need to update the motherduck docs for the config changes, and add a section about using GCS.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2720)


go/blob/gcs.go
```diff
@@ -30,21 +38,21 @@ func NewGCSBucket(ctx context.Context, bucket string, opts []option.ClientOption
...
 		return nil, err
 	}
 
 	return r, nil
 }
 
-func (s *GCSBucket) NewWriter(ctx context.Context, key string, opts ...WriterOption) io.WriteCloser {
-	return newBlobWriteCloser(ctx, s.Upload, key, opts...)
+func (b *GCSBucket) NewWriter(ctx context.Context, key string, opts ...WriterOption) io.WriteCloser {
+	return b.client.Bucket(b.bucket).Object(key).NewWriter(ctx)
```
williamhbaker, line 51: A small change here - `NewWriter` from the GCS SDK already gives us a writer and manages the goroutine lifecycle internally, so we don't need to do that again with `newBlobWriteCloser`.

go/blob/azure.go
```diff
@@ -48,9 +54,9 @@ type AzureBlobBucket struct {
 // NewAzureBlobBucket creates an Azure Blob object storage bucket. At least one
 // AzureConfigOption must be provided to specify authentication; either storage
 // account key or SAS token.
-func NewAzureBlobBucket(ctx context.Context, container string, accountName string, opts []AzureConfigOption) (*AzureBlobBucket, error) {
-	cfg := azureBlobStoreConfig{endpoint: defaultAzureEndpoint}
-	for _, opt := range opts {
+func NewAzureBlobBucket(ctx context.Context, container string, accountName string, authOpt AzureConfigOption, otherOpts ...AzureConfigOption) (*AzureBlobBucket, error) {
```
williamhbaker, line 57: I reworked the function signature here and for `NewGCSBucket` to explicitly require an option for authentication, since that is something that is needed anyway, and most callers will only provide that one option.

materialize-motherduck/config.go
```diff
@@ -50,21 +45,91 @@ func (c *config) Validate() error {
...
+}
+
+type stagingBucketGCSConfig struct {
+	BucketGCS           string `json:"bucketGCS" jsonschema:"title=GCS Staging Bucket,description=Name of the GCS bucket to use for staging data loads." jsonschema_extras:"order=0"`
+	CredentialsJSON     string `json:"credentialsJSON" jsonschema:"title=Service Account JSON,description=The JSON credentials of the service account to use for authorization to the staging bucket." jsonschema_extras:"secret=true,multiline=true,order=1"`
+	GCSHMACAccessID     string `json:"gcsHMACAccessID" jsonschema:"title=HMAC Access ID,description=HMAC access ID for the service account." jsonschema_extras:"order=2"`
+	GCSHMACSecret       string `json:"gcsHMACSecret" jsonschema:"title=HMAC Secret,description=HMAC secret for the service account." jsonschema_extras:"secret=true,order=3"`
+	BucketBucketPathGCS string `json:"bucketPathGCS,omitempty" jsonschema:"title=Bucket Path,description=An optional prefix that will be used to store objects in the staging bucket." jsonschema_extras:"order=4"`
```
Alex-Bair, line 130: Should this be `BucketPathGCS` (one `Bucket`, not two)?

Alex-Bair (approved):
LGTM % it looks like the `materialize-snowflake` tests failed? I'm not sure if it's related to the changes in this PR that affect that connector but wanted to mention it just in case.

materialize-motherduck/config.go
williamhbaker: lol yeah I'll fix that

williamhbaker:
Thanks for noticing the snowflake tests failing; fortunately that was just a result of some other tests running concurrently with them on another branch - they operate on the same tables, so another set of tests running at the same time will throw the other one off. Not exactly a great situation but a re-run usually clears it up!

Alex-Bair:
Ahh gotcha, sorry for the false alarm! I saw that they didn't fail for a couple other PRs merged earlier today, so I was suspicious. Glad it was due to a factor external to the PR!

PR #2727, opened 2025-04-26
sqlcapture: Add `emit_sourced_schemas` feature flag

willdonnelly:
**Description:**

Adds two feature flags to all SQL CDC connectors:

- `emit_sourced_schemas`: When true, the connector will periodically emit `SourcedSchema` messages to inform the runtime about changes to the source table schema. These updates are driven by the preexisting periodic-rediscovery behavior and are not automatically triggered by replicated changes not matching the schema, so we may want to improve on that in the future.
- `use_schema_inference`: When true, the connector will mark all discovered collections as `x-infer-schema: true` so that the user will get the usual schema-inference behavior.

Currently both of these default to false, this PR only makes them available for optional use. Toggling defaults will come later. When false, the current connector behavior should be entirely unchanged in every way so this should be a very low-risk PR.

**Notes for reviewers:**

Using the new `SourcedSchema` message required upgrading the Flow dependency in this repository. Since there hasn't been a tagged release since the `SourcedSchema` changes I opted to upgrade directly to the commit which added the protobuf definitions, but there are valid arguments that we ought to instead upgrade to today's HEAD version or that we ought to take a minute to tag an actual numbered version instead. Let me know if this seems worth addressing further.

Ironically, none of the test snapshot churn here is actually due to adding `SourcedSchema` messages, instead it's a side effect of some test code changes I made to more easily test the messages. Specifically, most of the churn here is because I changed the default capture output validator from `SortedCaptureValidator` to `OrderedCaptureValidator` which preserves the order in which the capture output the documents -- I've been meaning to make that change for a while because it's not really helpful to sort the outputs and it makes things confusing on occasion, and since I needed to fiddle with things anyway to add the ability to include SourcedSchema messages in the snapshot contents I figured this was as good a time as any to bite that bullet.


mdibaiee (approved):

jgraettinger:
Awesome to see this land!

>  These updates are driven by the preexisting periodic-rediscovery behavior and are not automatically triggered by replicated changes not matching the schema, so we may want to improve on that in the future.

I'm repeating myself, but just in case, for GA I would consider it a requirement that:

* If we emit a new column `foobar` in a Captured document, then
* We MUST emit SourcedSchema which minimally covers `foobar` as part of its same capture checkpoint
  * It's fine if the `SouredSchema` message is after a `Captured` message, so long as there isn't an interleaving `Checkpoint`.

Ideally we would do this whenever the apparent type of any column `foobar` changes (where "new" columns are just a particular class of "the type changed").

The reason for this is it prevents a bunch of gross materialization churn which would otherwise happen if we start emitting documents with a nested JSON `foobar`, and _then_ have a SourcedSchema update which schematizes it as JSON.

PR #2731, opened 2025-04-28
source-stripe-native: handle eventual consistency

JustinASmith:
**Description:**

We have evidence that Stripe's Events API is eventually consistent. Examples from [Sequin](https://blog.sequin.io/finding-and-fixing-eventual-consistency-with-stripe-events/) show how to combat this by yielding a cursor only if it is older than 5 seconds, otherwise do not update the cursor. This approach has been augmented for Estuary's connector to never fetch data that's more recent than 1 minute in the past by just returning and waiting for the connectors interval to pass. This means the most "real-time" the connector can be is 1 minutes (often a little more due to the default 5 minute interval), but it hopefully helps avoid eventual consistency issues causing the connector to miss updates.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2731)

### DIFF
--- a/source-stripe-native/source_stripe_native/api.py
+++ b/source-stripe-native/source_stripe_native/api.py
@@ -40,6 +40,16 @@ DO_NOT_HAVE_PLATFORM_CONTROLS_REGEX = r"You cannot perform this request as you d
 EVENT_NO_LONGER_RETAINED = r"is no longer available because it's aged out of our retention policy"
 
 
+# We have reason to believe the Stripe's Events API is eventually consistent. Examples of how to handle this
+# are written in the following article: https://blog.sequin.io/finding-and-fixing-eventual-consistency-with-stripe-events.
+# The approach in the article aims is to ensure the incremental tasks never fetch data more recent than 5 seconds in the past.
+# This is a good approach, however, we err on the side of caution and set the lag to 1 minute to ensure we don't miss any events.
+# This caps how "real-time" the streams can be but it's a simple fix that easily rolled back if we come up with a
+# better solution. The default interval is already 5 minutes, so the connector is not really "real-time" anyway.
+LAG = timedelta(minutes=1)
+MIN_INCREMENTAL_INTERVAL = timedelta(minutes=1)
+
+
 def add_event_types(
     params: dict[str, str | int], event_types: dict[str, Literal["c", "u", "d"]]
 ):
@@ -81,6 +91,15 @@ async def fetch_incremental(
     url = f"{API}/events"
     parameters: dict[str, str | int] = {"limit": MAX_PAGE_LIMIT}
     parameters = add_event_types(parameters, cls.EVENT_TYPES)
+    end = datetime.now(tz=UTC) - LAG
+
+    if end < log_cursor or (end - log_cursor < MIN_INCREMENTAL_INTERVAL):
+        # Return early and sleep if the end date is before the log cursor
+        # or the difference between the end date and log cursor is less than
+        # the minimum incremental interval.
+        return
+
+    parameters["created[lte]"] = int(end.timestamp())
     headers = {}
     account_id = (
         connected_account_id
@@ -240,6 +259,15 @@ async def fetch_incremental_substreams(
     url = f"{API}/events"
     parameters: dict[str, str | int] = {"limit": MAX_PAGE_LIMIT}
     parameters = add_event_types(parameters, cls_child.EVENT_TYPES)
+    end = datetime.now(tz=UTC) - LAG
+
+    if end < log_cursor or (end - log_cursor < MIN_INCREMENTAL_INTERVAL):
+        # Return early and sleep if the end date is before the log cursor
+        # or the difference between the end date and log cursor is less than
+        # the minimum incremental interval.
+        return
+
+    parameters["created[lte]"] = int(end.timestamp())
     max_ts = log_cursor
     headers = {}
     account_id = (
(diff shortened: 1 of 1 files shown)


source-stripe-native/source_stripe_native/api.py
```diff
@@ -40,6 +40,15 @@
 EVENT_NO_LONGER_RETAINED = r"is no longer available because it's aged out of our retention policy"
 
 
+# We have reason to believe the Stripe's Events API is eventually consistent. Examples of how to handle this
+# are written in the following article: https://blog.sequin.io/finding-and-fixing-eventual-consistency-with-stripe-events.
+# The approach in the article aims is to ensure the incremental tasks never fetch data more recent than 5 seconds in the past.
+# This caps how "real-time" the streams can be but it's a simple fix that easily rolled back if we come up with a
+# better solution. The default interval is already 5 minutes, so the connector is not really "real-time" anyway.
+LAG = timedelta(minutes=5)
```
Alex-Bair, line 48: Should the lag be 1 minute or 5 seconds? I remember both being mentioned, but the current lag is neither - it's 5 minutes, which seems a little high (unless you tested it out & saw events showing up after 5 minutes?).

source-stripe-native/source_stripe_native/api.py
```diff
@@ -81,6 +90,16 @@ async def fetch_incremental(
     url = f"{API}/events"
     parameters: dict[str, str | int] = {"limit": MAX_PAGE_LIMIT}
     parameters = add_event_types(parameters, cls.EVENT_TYPES)
+    max_end = datetime.now(tz=UTC) - LAG
+    end = min(max_end, log_cursor)
```
Alex-Bair, line 94: Typically, `log_cursor` represents a start time, so it seems weird that we're taking the minimum of the start time & some maximum end time. Wouldn't `min(max_end, log_cursor)` always evaluate to `log_cursor`, and we'd end up making the same request each sweep? 

It seems like not taking a minimum & using `max_end` would be the behavior we want; fetch events from the API created at or before `max_end`, then paginate through them until we reach events created before the `log_cursor`.

Alex-Bair (commented):
Looks like the right approach. It's nice there's a `created.lte` param we can use for this. I had a couple questions before approving.

source-stripe-native/source_stripe_native/api.py
JustinASmith: Dohh! I had this as 1-minute but updated to 5-minutes when looking at Salesforce native as an example of where we did this before. 1-minute should work just fine from what I can tell. I'll make sure to change that back.

source-stripe-native/source_stripe_native/api.py
JustinASmith: Good catch. You are right on both counts. We should just use `max_end` without taking`min` or `max`. That way we always use the desired max end time rather than ever choosing the `log_cursor` which could presumably be greater than the `max_end` (now - LAG) which would cause us to miss any data between `now - LAG` and `log_cursor`. But since we have a check to determine if the `max_end` is earlier than the `log_cursor` or the difference between them is less than the minimum incremental interval we want to consider we won't ever have that issue with just setting `end = datetime.now(tz=UTC) - LAG`.

source-stripe-native/source_stripe_native/api.py
JustinASmith: I think `LAG = timedelta(minutes=1)` and `MIN_INCREMENTAL_INTERVAL = timedelta(minutes=1)` make sense unless we would have any reason to make the minimum interval less than the `LAG`.

@williamhbaker  @Alex-Bair 
Would either of you think `MIN_INCREMENTAL_INTERVAL` and `LAG` set to the same value of 1-minute would cause a problem? I can't think of a reason why it would.

Alex-Bair: I think having them set to the same value would be fine. I don't foresee any issues with that.

source-stripe-native/source_stripe_native/api.py
```diff
@@ -40,6 +40,16 @@
 EVENT_NO_LONGER_RETAINED = r"is no longer available because it's aged out of our retention policy"
 
 
+# We have reason to believe the Stripe's Events API is eventually consistent. Examples of how to handle this
+# are written in the following article: https://blog.sequin.io/finding-and-fixing-eventual-consistency-with-stripe-events.
+# The approach in the article aims is to ensure the incremental tasks never fetch data more recent than 5 seconds in the past.
+# This is a good approach, however, we air on the side of caution and set the lag to 1 minute to ensure we don't miss any events.
```
Alex-Bair, line 46: Small typo. "air" should be "err".

Alex-Bair (approved):
LGTM % small typo

source-stripe-native/source_stripe_native/api.py
williamhbaker: TBH I'd get rid of the `MIN_INCREMENTAL_INTERVAL` entirely. If the default interval for all the bindings is 5 minutes I'm not sure it serves any real purpose. If a `fetch_changes` returns no results it won't be called again for another 5 minutes anyway. If we end up requesting some tiny time interval of changes like "1 minute ago minus 1 millisecond, up to 1 minute ago" I don't think that would be a problem either.

It's fine to keep it, too, since I don't think it hurts anything, but this might be a useful simplification.

source-stripe-native/source_stripe_native/api.py
JustinASmith: 😆  spelling is not my top skill sometimes. I just fixed another typo right before your comment

PR #2732, opened 2025-04-28
source-salesforce-native: emit `SourcedSchema`s & cast custom fields to strings

Alex-Bair:
**Description:**

Users frequently want their materialized tables to include columns for fields that haven't been populated yet. Normally, materializations won't add a column until we've seen it contain non-`null` values & that's reflected in the collection's schema. However, if the connector emits a `SourcedSchema` with the expected types for those fields, inferred schemas will have a non-`null` type for those fields, and materializations will include the corresponding columns.

This PR adds emitting `SourcedSchema`s to `source-salesforce-native` to address the above issue. Almost all fields are included in the `SourcedSchema` - only field types I haven't seen in the wild yet (like `tns:json`, `urn:JunctionIdListNames`, etc.) are omitted. Once we see a field with these types & understand what their values actually looks like, we can include them in the sourced schema. 

`SourcedSchema`s are also only emitted when the binding is opened. The connector currently doesn't re-check Salesforce for new fields while it's running; it only fetches that metadata when opening a binding. If we later decide to re-check Salesforce for new fields, we should also emit additional `SourcedSchemas` at the same time. Additional considerations when doing this type of periodic schema update are mentioned by Johnny here https://github.com/estuary/connectors/pull/2727#issuecomment-2836190973.

Additionally, Salesforce is frequently wrong about custom field types. So instead of trusting Salesforce's reported type, the connector will now always casts scalar type custom fields to strings. Since materializations are smart enough to materialize formatted string fields based on the format (i.e. materialize `type:string, format:number` as a number column), casting custom fields to strings should be mostly invisible to users. However, there is one change - boolean custom fields will now be strings, not booleans since `type:string, format:boolean` isn't a thing.

**Workflow steps:**

**Notes for reviewers:**

- `SourcedSchema` messages are emitted when a binding is opened.
- A collection's inferred schema is updated with the information in the emitted `SourcedSchema`.
- Fields that are not populated in Salesforce have associated columns in the tables created by `materialize-postgres`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2732)


source-salesforce-native/source_salesforce_native/models.py
```diff
@@ -286,3 +288,95 @@ class BulkJobCheckStatusResponse(BulkJobSubmitResponse):
...
+    schema = {
+        "additionalProperties": False,
+        "type": "object",
+        "properties": {}
+    }
+
+    for field, details in fields_dict.items():
+        field_schema: dict[str, Any] = {}
+        match details.soapType:
+            case SoapTypes.ID | SoapTypes.BASE64 | SoapTypes.STRING | SoapTypes.TIME:
```
williamhbaker, line 308: There is a `format: time` for times, see https://github.com/estuary/flow/blob/master/crates/json/src/schema/formats.rs

williamhbaker (approved):
LGTM, see comment about time fields

source-salesforce-native/source_salesforce_native/models.py
Alex-Bair: It doesn't look like the runtime is detecting `format: time`. I checked this earlier - see the `my_time_field__c` field for this [collection](https://dashboard.estuary.dev/collections/details/spec?catalogName=alexbair%2F19MAR2025_test%2FUser). The value from Salesforce is `12:43:00.000Z` but it's schematized as an unformatted string.

Do you know if there's a reason `format: time` isn't detected? I'll still add `format: time` to the connector, but I suspect the first time we see an actual value, the format will be dropped from the inferred schema.

Alex-Bair: I was wrong. I just tested out adding `format: time`, and it's retained in the inferred schema even after capturing documents.

williamhbaker: From what I can tell the `format` detection goes through [here](https://github.com/estuary/flow/blob/2cbdd78e3c1f8213b8031d049394899a6475eeb9/crates/doc/src/shape/widen.rs#L13) and then [here](https://github.com/estuary/flow/blob/2cbdd78e3c1f8213b8031d049394899a6475eeb9/crates/json/src/schema/formats.rs#L188-L198) - indeed it does not seem to detect a `time` format (maybe for performance reasons, only the most common formats are detected; I'm not really sure why it is like this) - but I also _think_, from looking through the code, that if there is an initial `format: time` annotation, then subsequent values will be  validated against that format, so it might actually end up remaining a `format: time` based on the `SourcedSchema` as long as all future values validate against that format.

PR #2736, opened 2025-04-29
source-google-analytics-data-api-native: validate start date

Alex-Bair:
**Description:**

Google enforces that start dates must be greater than 2015-08-13 and less than 3000-01-01. We weren't validating that previously & a user ran into an error because of that. This commit adds that missig validation.

There's a little extra handling since dates in the config are UTC but we convert to the user's timezone within the connector, so we have to make sure that converted datetime is still greater than `MIN_START_DATE` in the user's timezone.

**Workflow steps:**

**Notes for reviewers:**

- start dates on or before 2015-08-13 are not allowed in the config and raise a validation error. 
- start dates used within the connector are at a minimum on 2015-08-14, even if a user's timezone pushes the configured start date before then.


JustinASmith (approved):
LGTM

PR #2739, opened 2025-04-29
sqlcapture: Omit reductions from SourcedSchemas

willdonnelly:
**Description:**

They don't really make sense here, and the if/then/else deletion bits specifically aren't compatible with schema inference at all.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2739)


jgraettinger (approved):
LGTM but i see test failures that look related

willdonnelly:
Whoops, yeah, I'm just forgetful and didn't actually update the relevant test snapshot for those connectors.

PR #2740, opened 2025-04-29
source-chargebee-native: ensure API request interval is valid

JustinASmith:
**Description:**

This PR should fix an issue identified in a capture where the `start` was after the `end`  end time for `created_at[between]` and `updated_at[between]` for incremental `fetch_changes` due to updating the `max_updated_at` timestamp in the `while` loop without checking that it is before the `log_cursor` and simultaneously using the `max_updated_at` in the API request instead of using the same `start`/`end` times for the API requests.

This PR introduces a check that will skip yielding documents if the `updated_at` is beyond the `log_cursor` and keeps the `start` time for the API requests the same. Additionally, there is a `le=datetime.now(tz=UTC)` check on the `EndpointConfig` model which will prevent the user from specifying a `start_date` that is in the future.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2740)

### DIFF
--- a/source-chargebee-native/source_chargebee_native/api.py
+++ b/source-chargebee-native/source_chargebee_native/api.py
@@ -154,9 +154,10 @@ async def fetch_resource_changes(
 ) -> AsyncGenerator[ChargebeeResource | LogCursor, None]:
     assert isinstance(log_cursor, int)
 
-    max_updated_at = _ts_to_dt(log_cursor)
+    start_date = _ts_to_dt(log_cursor)
+    end_date = min(start_date + timedelta(days=30), datetime.now(tz=UTC))
+    max_updated_at = start_date
     has_results = False
-    end_date = min(max_updated_at + timedelta(days=30), datetime.now(tz=UTC))
     offset = None
 
     while True:
@@ -165,7 +166,7 @@ async def fetch_resource_changes(
             log,
             site,
             resource_name,
-            max_updated_at,
+            start_date,
             end_date,
             offset,
             True,
@@ -178,6 +179,9 @@ async def fetch_resource_changes(
         has_results = True
 
         for doc in resource_data:
+            if doc.cursor_value > log_cursor:
+                continue
+
             max_updated_at = max(max_updated_at, _ts_to_dt(doc.cursor_value))
 
             if doc.deleted:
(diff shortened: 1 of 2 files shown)


source-chargebee-native/source_chargebee_native/api.py
```diff
@@ -122,6 +123,11 @@ async def fetch_resource_page(
     assert isinstance(cutoff, datetime)
     assert isinstance(offset, str | None)
 
+    if cutoff < start_date or (cutoff - start_date) < MIN_INCREMENTAL_INTERVAL:
```
Alex-Bair, line 126: We shouldn't use the `(cutoff - start_date) < MIN_INCREMENTAL_INTERVAL` for the backfill task, right? Even if the backfill window is 1 second wide, we should check it for records.

And do you anticipate a situation where `cutoff < start_date` would be true? The only time I could see this happening is if a user specified a start date in the future.

source-chargebee-native/source_chargebee_native/api.py
```diff
@@ -154,18 +160,24 @@ async def fetch_resource_changes(
...
 
-    max_updated_at = _ts_to_dt(log_cursor)
+    start_date = _ts_to_dt(log_cursor)
+    end_date = min(start_date + timedelta(days=30), datetime.now(tz=UTC))
+    max_updated_at = start_date
     has_results = False
-    end_date = min(max_updated_at + timedelta(days=30), datetime.now(tz=UTC))
     offset = None
 
+    if end_date < start_date or (end_date - start_date) < MIN_INCREMENTAL_INTERVAL:
```
Alex-Bair, line 169: I'm not sure we need these checks in the incremental tasks. `end_date` should always be larger than `start_date`, and (unless I'm missing some Chargebee-specific behavior) there's not a reason we need to ensure the date window we check is a certain size, right?

Alex-Bair (approved):
LGTM % a couple questions. If I'm understanding the issue correctly, I think we just need to pin the `start_date` value used in requests & skip any records from Chargebee that are updated outside the date window we're checking.

source-chargebee-native/source_chargebee_native/api.py
JustinASmith: You are right. I did some more thinking and really the update in this PR that fixes the issue is not using the `max_updated_at` in the API request since it updates in the while loop and there is not a reason to update it anyway.

I will remove these checks for `cutoff < start_date or (cutoff - start_date) < MIN_INCREMENTAL_INTERVAL` since they are no longer needed. Plus, I will add a `le=datetime.now(tz=UTC)` to the `EndpointConfig.start_date` so that all start dates specified by the user need to be less than or equal to the current time.

JustinASmith: Additionally, the check in the `fetch_changes` functions where we were previously just yielding `doc` will now only yield the doc and update the `max_updated_at` if the `cursor_value` is less than or equal to the `log_cursor`. This should prevent any jumps in our `log_cursor` which could cause missed data. The only scenario I can think of where this may happen is when a resource in Chargebee is updated as we make our API request and when their backend is collecting the results it somehow includes the record that was updated outside of are `start`/`end` time window.

PR #2742, opened 2025-04-30
source-iterable: require primary keys in schemas

Alex-Bair:
**Description:**

Primary keys were previously not required in the various streams' schemas. Materializations can't materialize collection with nullable primary key fields when there's no default value annotation, so these collections couldn't be materialized. Making the primary keys required fixes this & allows collections to be materialized.

I updated `lists` to disallow `null` as a possible `id`. PKs for all other streams were already not allowed as explicit `null`s. 

Discover snapshot changes are expected due to requiring primary key fields. (jsondiff)

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed tables for collections could be created in a dockerized PostgreSQL instance.

I did notice that `events` does not have a primary key, but I decided to not add one to avoid re-versioning existing collections. We may want to revisit this later, or consider adding a primary key if we decide to make a native version of this connector.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2742)


williamhbaker (approved):
LGTM

PR #2746, opened 2025-05-01
source-shopfiy-native: add logging & retries for failed job cancelations

Alex-Bair:
**Description:**

There have been failures when captures attempt to cancel an ongoing bulk job. These failures suggest that the `data` field is absent from the `BulkJobCancelResponse`, but per Shopify's docs & my testing, that field should always be present. I haven't been able to replicate this failure with API calls, so I'm adding some logging & retries when this failure happens to let me troubleshoot better when it happens.

I also reduced the initial sleep time after submitting a bulk job - 5 seconds seemed a little too long.

**Workflow steps:**

**Notes for reviewers:**


source-shopify-native/source_shopify_native/graphql/bulk_job_manager.py
```diff
@@ -165,11 +167,28 @@ async def _cancel(self, job_id: str) -> bool:
...
+            try:
+                response = BulkJobCancelResponse.model_validate_json(response_bytes)
+                break
+            except ValidationError as err:
+                if attempts <= 3:
+                    self.log.info("Error validating BulkJobCancelResponse.", {
+                        "attempts": attempts,
+                        "job_id": job_id,
+                        "err": err,
+                        "response_bytes": response_bytes,
```
williamhbaker, line 186: I'm not sure how python handles logging, but if this is actual bytes it might get base64 encoded, so maybe just be aware of that and consider doing an equivalent of `string(response_bytes)` on it.

williamhbaker (approved):
LGTM

source-shopify-native/source_shopify_native/graphql/bulk_job_manager.py
Alex-Bair: Good point, I'll do `json.loads(response_bytes.decode('utf-8'))` on it - that should make it pretty readable in the log. Thanks for the suggestion!

PR #2747, opened 2025-05-01
materialize-iceberg: allow load result lines of any length

williamhbaker:
**Description:**

The default line length limit of bufio.Scanner is 64KiB, which is quite a bit shorter than documents we commonly see.

There's no real reason for us to have any limit on this line length, since the data we are loading back is data we put there in the first place.

This change sets an initial relatively small buffer, but allows the scanner to resize it as large as it needs to based on the data received.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2747)

### DIFF
--- a/materialize-iceberg/transactor.go
+++ b/materialize-iceberg/transactor.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"math"
 	"path"
 	"slices"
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #2748, opened 2025-05-01
materializations: use parallel gzip compression for writing JSON and CSV

williamhbaker:
**Description:**

Although connectors don't have a lot of cores available, they do have more than 1 - and to be precise, they have 2. The second core isn't fully utilized during the combiner draining phase of a transaction when the connector itself is CPU limiting, so we can get some extra throughput out of it by using the parallel gzip compressor.

It's not an enormous benefit, but it increases the throughput of the connector during combiner draining by about 30-40%, which makes a measurable difference in overall connector throughput, without any downside.

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM!

PR #2752, opened 2025-05-01
connectors: install curl for all Go connectors

williamhbaker:
**Description:**

`curl` is generally useful to have available in a connector image to access Go's pprof HTTP server, so install it by default rather than needing to rely on annoying workarounds.

**Workflow steps:**

**Notes for reviewers:**

The automated CI tests aren't really effective for a change like this, since all connector images use the base image tagged as `v1`, rather than the one that is built on this branch. But I did spot check a connector container built with the image pushed by this branch, and it worked as expected. There's a bit of cleanup to some of the materialization dockerfiles where they are installing some redundant things as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2752)

### DIFF
--- a/base-image/Dockerfile
+++ b/base-image/Dockerfile
@@ -1,6 +1,6 @@
 FROM --platform=linux/amd64 debian:stable-slim as debian
 
-RUN apt-get update && apt-get install -y openssh-client ca-certificates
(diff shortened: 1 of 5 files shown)


willdonnelly (approved):
LGTM

PR #2754, opened 2025-05-02
materialize-postgres: don't start a postgres transaction if there are no keys to load

williamhbaker:
**Description:**

Lazily start a Postgres transaction only when there is at least once key to load.

Previously, we would always start a transaction, even if no load requests were received. In cases such as delta updates bindings, this would cause a transaction to be held open and then be closed out without doing anything for the duration of the runtime's reading of documents to store.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2754)


Alex-Bair (approved):
LGTM

PR #2761, opened 2025-05-06
sqlcapture: Set additionalProperties: false in SourcedSchemas

willdonnelly:
**Description:**

I wasn't entirely sure whether we just wanted that on the top level or if it needed to be present in the metadata objects as well, so I've erred on the side of putting it everywhere.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2761)


willdonnelly:
This PR in combination with https://github.com/estuary/flow/pull/2126 _almost_ fixes everything in the basic test I've been running. The simple table with some data in it works correctly, and the empty table produces the following vaguely-correct inferred schema in Supabase:

```
{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "additionalProperties": false,
  "properties": {
    "_meta": false,
    "data": {
      "description": "(source type: text)",
      "type": "string"
    },
    "id": {
      "description": "(source type: non-nullable int4)",
      "type": "integer"
    }
  },
  "required": [
    "_meta",
    "id"
  ],
  "type": "object",
  "x-collection-generation-id": "11113881ac000211",
  "x-infer-schema": true
}
```

However the inferred schema publication fails with:

```
publication failed with status: BuildFailed { incompatible_collections: [], evolution_id: None } (will retry)
```

The obvious culprit would seem to be the weird-looking `"_meta": false` property definition (possibly in conjunction with `_meta` being required?). I'm not entirely sure what's causing that, but in case there's some obvious error here's the sourced schema the connector outputs:

```
{
  "$defs": {
    "PublicEmpty": {
      "$anchor": "PublicEmpty",
      "additionalProperties": false,
      "properties": {
        "data": {
          "description": "(source type: text)",
          "type": [
            "string"
          ]
        },
        "id": {
          "description": "(source type: non-nullable int4)",
          "type": "integer"
        }
      },
      "required": [
        "id"
      ],
      "type": "object"
    }
  },
  "allOf": [
    {
      "properties": {
        "_meta": {
          "additionalProperties": false,
          "properties": {
            "before": {
              "$ref": "#PublicEmpty",
              "description": "Record state immediately before this change was applied."
            },
            "op": {
              "description": "Change operation type: c Create/Insert, u Update, d Delete.",
              "enum": [
                "c",
                "d",
                "u"
              ]
            },
            "source": {
              "additionalProperties": false,
              "properties": {
                "loc": {
                  "description": "Location of this WAL event as [last Commit.EndLSN; event LSN; current Begin.FinalLSN]. See https://www.postgresql.org/docs/current/protocol-logicalrep-message-formats.html",
                  "items": {
                    "type": "integer"
                  },
                  "maxItems": 3,
                  "minItems": 3,
                  "type": "array"
                },
                "schema": {
                  "description": "Database schema (namespace) of the event.",
                  "type": "string"
                },
                "snapshot": {
                  "description": "Snapshot is true if the record was produced from an initial table backfill and unset if produced from the replication log.",
                  "type": "boolean"
                },
                "table": {
                  "description": "Database table of the event.",
                  "type": "string"
                },
                "ts_ms": {
                  "description": "Unix timestamp (in millis) at which this event was recorded by the database.",
                  "type": "integer"
                },
                "txid": {
                  "description": "The 32-bit transaction ID assigned by Postgres to the commit which produced this change.",
                  "type": "integer"
                }
              },
              "required": [
                "schema",
                "table",
                "loc"
              ],
              "type": "object"
            }
          },
          "required": [
            "op",
            "source"
          ],
          "type": "object"
        }
      },
      "required": [
        "_meta"
      ]
    },
    {
      "$ref": "#PublicEmpty"
    }
  ],
  "x-infer-schema": true
}
```

willdonnelly:
> * additionalProperties: false should be on both the top-level object, and also under `_meta/before`

Yeah, that's what I'm doing in the current version of this PR. Summarizing and eliding some unimportant bits, the schemas currently look like this:

```
{
  "$defs": {
    "<anchorName>": {
      "$anchor": "<anchorName>",
      "type": "object",
      "required": ["id" ],
      "properties": {
        "data": {
          "description": "(source type: varchar)",
          "type": ["string"]
        },
        "id": {
          "type": "integer",
          "description": "(source type: non-nullable int)"
        }
      }
    }
  },
  "allOf": [
    {
      "required": ["_meta"],
      "properties": {
        "_meta": {
          "type": "object",
          "required": ["op", "source"],
          "additionalProperties": false,
          "properties": {
            "before": {
              "$ref": "#<anchorName>",
              "additionalProperties": false
            },
            "op": {...}
            "source": {
              "type": "object",
              "properties": {...}
              "additionalProperties": false,
              "required": ["schema", "table", "cursor"]
            }
          }
        }
      }
    },
    {"$ref": "#<anchorName>"}
  ],
  "additionalProperties": false
}
```

That's an `additionalProperties: false` on the top-level object, on the `/_meta` object, and on `/_meta/before` and `/_meta/source`.

>     * the `_meta: false` result is a valid JSON schema construction which just means "there is no acceptable JSON construction that can exist at this location"

Yeah, I figured it probably meant something along those lines. I'm not sure what about the current schema makes the top-level `id` and `data` properties get treated like that though. I could believe that I've screwed something up with all the complexities around `/_meta` but the `id` and `data` properties are only mentioned in the shared definition part and I can't see anything wrong with how they're being defined that would make their typing unacceptable.

PR #2762, opened 2025-05-07
source-postgres: Add "create_replication_slot" feature flag

willdonnelly:
**Description:**

The `create_replication_slot` flag setting is the default behavior and amounts to doing what we were already doing, but by adding a feature flag we can allow users to specify `no_create_replication_slot` in order to disable any automatic replication slot management on our part.

In conjunction with this change, I've reworked the "does the slot exist" validation check so that it never attempts to create the slot. This part is a change I've been meaning to make for a while, as once we've verified that the capture user has the REPLICATION role and that the slot doesn't already exist in another database, there is no way we could fail to recreate it later on when the capture proper starts.

The reason this is preferable is that it significantly reduces the chances of several error conditions:
1. The user fiddles with the slot name setting after having already triggered validation. When this happens, they can end up with an unused slot which possibly eats up WAL retention without bound.
2. The user fiddles with the database name setting after having run validation once. When this happens, prerequisite validation will fail with a "slot exists in wrong database" error when really if we hadn't been so eager there would be no problems.
3. The user runs validation and the slot is created, but then they don't ever actually publish the capture. This is similar to (1) and results in the same basic failure pattern of an unused slot retaining a potentially unbounded amount of WAL.

Much nicer to avoid all that mess by not creating slots during the validation process in the first place.

The check itself still exists so that it can detect the "slot does exist but in the wrong logical database" condition and so that we can provide a useful error when `no_create_replication_slot` is set but the user actually didn't create it themselves, but we will only autocreate the slot as part of the actual capture.

**Workflow steps:**

Nothing should change by default.

Users can specify feature flag setting `no_create_replication_slot` to be guaranteed that we will never create the replication slot ourselves.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2762)

### DIFF
--- a/source-postgres/.snapshots/TestCaptureAfterSlotDropped-capture2
+++ b/source-postgres/.snapshots/TestCaptureAfterSlotDropped-capture2
@@ -2,4 +2,8 @@
 # Final State Checkpoint
 # ================================
 {"bindingStateV1":{"test%2Fcaptureafterslotdropped_46115540":{"backfilled":2,"key_columns":["id"],"mode":"Active"}},"cursor":"0/1111111"}
+# ================================
+# Captures Terminated With Errors
+# ================================
+error creating replication stream: unable to start replication: ERROR: replication slot "flow_slot" does not exist (SQLSTATE 42704)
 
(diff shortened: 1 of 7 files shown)


willdonnelly:
One additional note: it might not be immediately obvious why it's okay to rely on the empty-cursor slot recreation logic to always trigger when the replication slot needs to be created initially, but the rationale is basically that we can never arrive at a non-empty resume cursor value until we've gone through that code path.

Alex-Bair (approved):
LGTM

PR #2763, opened 2025-05-07
source-firestore: Configurable backfill intervals

willdonnelly:
**Description:**

Adds a resource spec and advanced endpoint config setting for the minimum (re)backfill interval. If neither is set we will continue to apply the same "5m with a restart cursor and 24h otherwise" policy we always have, but if either is set we will use the specified interval instead. The resource level setting overrides the endpoint config, while the endpoint config allows users to set a default which will definitely be applied to any new bindings, so I think it's useful to have both.

I've tried to keep this PR limited to just adding that new setting, so it's worth noting that the current connector-restart behavior is absolutely terrible so I haven't even bothered updating it to account for the new customizable intervals. As just one concrete example, if change streaming happens to fail and then the connector restarts on its own before the forced-restart timeout fires, there will no longer be any sort of restart timeout and we just have to hope it restarts again after the backfill interval is reached.

I'm not bothering to fix that here because there is no simple fix for that issue, if/when we decide to tackle this we really just need to implement the ability to trigger new backfills without restarting the connector. This should be easier now than it used to be since we've removed some of the complications like needing to interrupt an ongoing backfill and managing a many-to-one binding to backfill state mapping, but it's still beyond the scope of this quick fix.

**Workflow steps:**

Nothing should change by default. Users may now customize the minimum re-backfill interval if they want it to be longer or shorter than our 24h default.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2763)

### DIFF
--- a/source-firestore/.snapshots/TestSpec
+++ b/source-firestore/.snapshots/TestSpec
@@ -29,6 +29,11 @@
             "type": "boolean",
             "title": "Skip Automatic Discovery",
             "description": "When set the connector will skip automatic collection discovery. This generally only makes sense when the \"Extra Collections\" setting is used."
+          },
+          "min_backfill_interval": {
+            "type": "string",
+            "title": "Minimum Backfill Interval",
+            "description": "Controls how often a collection may be re-backfilled in the event of unrecoverable change stream failure. May be overridden by the per-resource setting."
           }
         },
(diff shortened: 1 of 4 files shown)


williamhbaker (approved):
LGTM

PR #2766, opened 2025-05-08
Fix/normalize column names

surister:
**Description:**
Some column names CrateDB does not allow, for example _id, since it's a system reserved column, because of this, any column that starts with underscore is not valid.

Fixes https://github.com/crate/cratedb-estuary/issues/13

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2766)


williamhbaker (approved):
LGTM

PR #2771, opened 2025-05-09
materialize-elasticsearch: cap request retry delay

williamhbaker:
**Description:**

Previously there was no upper limit on the retry delay for requests, and it would get very long if there was a request that was never going to succeed. This caps the retry delay to something reasonable, so that intermittent failures can be retried, but if they are doomed to fail they'll do so in a timely manner.

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM

PR #2772, opened 2025-05-09
source-iterable: fix checkpointing behavior for `IterableExportStream` and family

Alex-Bair:
**Description:**

Previously, the `IterableExportStream` streams would checkpoint state while processing records in a single response, inspecting the cursor field of each record & yielding it if it's more recent than the state. This works fine if records returned by the API are sorted based on their cursor fields, but the API does no such sorting. Because of that, the connector's state can jump ahead if it's restarted in the middle of processing a response. We've seen this cause lots of missing data.

Since we can't sort the API's results, the connector's been updated to checkpoint the end of the date window after processing all records in the date window. This ensures all records between the start & end are captured before checkpointing, preventing the connector's state from jumping ahead due to a restart.

I also fixed a couple schema violations for `users` as well.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed `users` captures significantly more records (300k vs. 39k) and transitioning from the old checkpointing behavior to the new behavior does not cause errors. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2772)

### DIFF
--- a/source-iterable/source_iterable/streams.py
+++ b/source-iterable/source_iterable/streams.py
@@ -786,10 +786,14 @@ class Users(IterableExportStreamRanged):
 
 
     def read_records(self, **kwargs) -> Iterable[Mapping[str, Any]]:
-        for record in super().read_records(**kwargs):
+        # These datetime fields are not ISO 8601 compliant, so we re-format them so they comply
+        # with their schematized format: datetime annotation.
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2775, opened 2025-05-10
source-monday: fix handling of empty item responses items backfill

JustinASmith:
**Description:**

There was a small bug introduced in a minor refactor that was apparently not tested as expected. We were backfilling up to the point a board did not have any items. This ensures that we backfill until there are no more boards. That is a key distinction since we want to check each board for items to backfill rather than stopping early when a board may not have items to backfill.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2775)

### DIFF
--- a/source-monday/source_monday/graphql.py
+++ b/source-monday/source_monday/graphql.py
@@ -358,9 +358,8 @@ async def fetch_items_by_boards(
     page: int,
     itemsLimit: int,
 ) -> AsyncGenerator[Item | str, None]:
-    """
-    Note: If `board_ids` is not provided, items from all boards will be fetched.
-    """
+    board_cursors: dict[str, str] = {}
+
     response = await execute_query(
         ItemsByBoardResponse,
         http,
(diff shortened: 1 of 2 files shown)


source-monday/source_monday/graphql.py
```diff
@@ -361,6 +361,9 @@ async def fetch_items_by_boards(
     """
     Note: If `board_ids` is not provided, items from all boards will be fetched.
     """
```
Alex-Bair, line 363: nit: it looks like `board_ids` is no longer a parameter & this doc string isn't applicable any more.

source-monday/source_monday/graphql.py
```diff
@@ -437,7 +439,7 @@ async def fetch_items_by_boards(
     log.debug(
         f"Finished fetching items for board {board_id}. Remaining cursors: {board_cursors.keys()}"
     )
-    if board_id is not None:
+    if has_board:
```
Alex-Bair, line 442: `has_board` is always `True` at this point, right? I'm not sure `has_board` is needed. It seems like if we get to this point in the code, we know there's always a `board_id` to yield and should unconditionally yield it.

Alex-Bair (approved):
LGTM % a couple comments

source-monday/source_monday/graphql.py
JustinASmith: Doh.. you're right. Will update that 😆

PR #2776, opened 2025-05-12
source-sage-intacct: new connector

williamhbaker:
**Description:**

New capture connector for Sage Intacct (yes, with two C's) that uses the XML web services API.

The XML format of the API isn't too bad since query responses can be requested in JSON format, which gives us a nice stream of JSON objects that we can process incrementally. Only the metadata and other general "setup" responses need to be handled directly as XML.

There is robust handling implemented here for querying records more recent than a given timestamp, where the results are returned as a fixed-sized page, and there is no token-based pagination of a static result set. It deals with things like runs of records with the same timestamp spanning page boundaries, races of deleted records with on-going queries, and timestamp "cycles" where an entire page of results has the same timestamps. 

Also of note is that it is known that the API is eventually consistent, so we'll use a "lookback" `fetch_changes` function to re-fetch change events after some period of time, right now hard-coded at 2 hours. This might need tweaking in the future to make it longer, or perhaps be user-configurable.

We only currently have access to a fairly limited test account, so not everything could be thoroughly tested, and there are some TODO's commented in the code where we will need to add stuff in once we see examples of what it looks like, and figure out how we should be handling it. The most significant TBD item is processing deletions from the `AUDITHISTORY` object, although I think that should be a relatively simple matter of having it be its own binding, and emitting associated documents for other bindings when it gets deletion events.

I'll add snapshot tests to this PR if I'm able to do, once I confirm the test data we are using is alright for that.

**Workflow steps:**

Will need new docs created

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2776)


source-sage-intacct/source_sage_intacct/resources.py
```diff
@@ -0,0 +1,147 @@
...
+            state,
+            task,
+            fetch_changes={
+                "realtime": functools.partial(
+                    fetch_changes, obj, sage, None, PAGE_SIZE
+                ),
+                "lookback": functools.partial(
+                    fetch_changes, obj, sage, timedelta(hours=2), PAGE_SIZE
+                ),
+            },
```
JustinASmith, line 79: Glad to see this feature used and be useful here!

JustinASmith (approved):
LGTM - pretty different connector considering the XML, various quirks of the API, etc. Overall I understand the high points without getting too into the details of every little thing. Appreciate the testing you did, that really proves out the core logic you've detailed!

Just so I'm clear, we are expecting this to likely fail since we had limited ability to test due to the read only API test account? Then we will address the issues and finish out the TODOs thereafter?

williamhbaker:
> Just so I'm clear, we are expecting this to likely fail since we had limited ability to test due to the read only API test account? Then we will address the issues and finish out the TODOs thereafter?

Yeah, generally. I mean it might not _fail_, but as these things go we typically learn a lot more when a connector is run with data at scale. And this is a fairly extreme example, where we have very limited ability to simulate something even approaching realistic conditions to capture from.

PR #2778, opened 2025-05-13
sqlcapture: Emit checkpoint after SourcedSchemas

willdonnelly:
**Description:**

This avoids an extra one-minute delay when the DB is idle.


williamhbaker (approved):
LGTM

PR #2779, opened 2025-05-13
source-postgres: Avoid nil dereference in edge case

willdonnelly:
**Description:**

The translateArray function constructs a "hypothetical scalar equivalent" column metadata which is used for the recursive translation of element values, however in some edge cases we may not actually have column metadata for the value in the first place.

If that happens, we should roll with it and produce a nil value for the scalar equivalent column as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2779)


williamhbaker (approved):
LGTM

PR #2785, opened 2025-05-13
source-stripe-native: listen to more event types during incremental replication

Alex-Bair:
**Description:**

Users have mentioned that `source-stripe-native` is missing updates. The example update we missed was before https://github.com/estuary/connectors/pull/2731 was merged, but it's also possible that the connector isn't listening to enough event types during incremental replication. This PR adds a handful of event types in case the missed update(s) were because the connector wasn't listening for those events.

Additionally, it turns out that Stripe _does_ generate `payout.*` events, so the `Payouts` has been switched to a `BaseStripeObjectWithEvents` stream.

**Workflow steps:**

**Notes for reviewers:**

I confirmed most of the new event types are actually generated by Stripe's API. There are a few I could not confirm are generated, but I'd like to err on the side of caution here. Including these event types doesn't cause the Stripe API to throw errors, so I'm including them in case they are rarely generated & I just haven't observed them yet.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2785)


JustinASmith (approved):
LGTM

PR #2788, opened 2025-05-14
materialize-motherduck: add merge bounds to load and store queries

williamhbaker:
**Description:**

Adds merge bounds to load and store queries in MotherDuck, as we have added for the other warehouse materializations.

In the past I have mentioned a desire to have a more principled way to generally incorporate the "bounds" concept in materializations, since most of them will want access to those. My best idea for this so far is to incorporate them into the transactor, but that would also require more deeply incorporating the concept of "value converters" that materializations use to convert values to their specific needs. Probably a good idea, but a lot more work.

But there is an immediate use case for performance optimization of MotherDuck load queries, and following this pattern that we have used before is relatively simple and realistically I don't know if I'm ever going to get around to doing it in a more elegant way.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2788)

### DIFF
--- a/materialize-motherduck/.snapshots/TestSQLGeneration
+++ b/materialize-motherduck/.snapshots/TestSQLGeneration
@@ -55,9 +55,9 @@ JOIN read_json(
 		"key!binary": 'VARCHAR NOT NULL'
 	}
 ) AS r
-	 ON  l.key1 = r.key1
+	 ON  l.key1 = r.key1 AND l.key1 >= 10 AND l.key1 <= 100
 	 AND l.key2 = r.key2
-	 AND l."key!binary" = r."key!binary"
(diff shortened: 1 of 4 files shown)


materialize-motherduck/.snapshots/TestSQLGeneration
```diff
@@ -125,25 +125,6 @@ SELECT * FROM read_json(
 ) WHERE flow_document != '"delete"';
 --- End ""."a-schema".key_value storeQuery ---
 
---- Begin ""."a-schema".delta_updates loadQuery ---
```
williamhbaker, line 128: These two snapshots were bogus, since delta updates bindings never run load queries or merges. Note: A merge query in MotherDuck is accomplish by first deleting all the pre-existing rows that match the set of keys we are going to store, followed by inserting said rows to store.

Alex-Bair (approved):
LGTM

PR #2789, opened 2025-05-14
source-chargebee-native: add incremental json processing and configurable page limit

JustinASmith:
**Description:**

We've been seeing timeouts in some captures, however, I have not had success reproducing this locally. Since I cannot reproduce this reliably and some other captures do not have this issue, I am making an educated guess that the underlying issue is with Chargebee taking longer than the default 5-minute timeout the `aiohttp.ClientSession` has set.

This PR updates the `aiohttp.ClientSession` timeout to be 30-minutes for the `total` and 5-minutes for the `sock_connect`.

References:
- `ClientTimeout`: https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientTimeout.
- `aiohttp.client.py` defaults: https://github.com/aio-libs/aiohttp/blob/82df143c70f9bbff6f46a69b6c8cd0e637cc54c1/aiohttp/client.py#L221.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2789)


source-chargebee-native/source_chargebee_native/api.py
```diff
@@ -62,38 +62,47 @@ async def _fetch_resource_data(
...
         params["include_deleted"] = "true"
 
     if filter_params:
         params.update(filter_params)
 
-    response = APIResponse[resource_type].model_validate_json(
-        await http.request(log, url, params=params)
-    )
+    _, body = await http.request_stream(log, url, params=params)
+    async for record in IncrementalJsonProcessor(
```
williamhbaker, line 75: I don't believe this is going to accomplish anything - it's going to still read the entire `list` into memory all at once, and then callers of this function iterate through that list and yield documents from it.

Incremental processing isn't a bad idea in general, but you would need to incrementally yield items while reading the actual list to get any benefit from it.

I don't think that's really necessary here though, or for this particular timeout problem. Even the max page size of 100 records is pretty small. The current implementation will read that response body as quickly as the connector can parse it into memory, so that's very unlikely to cause timeouts from the http connection. The benefit of incremental processing is to reduce memory usage; it will probably make fully reading the http response body slower though, actually.

I'd suggest starting with just extending the http timeout and seeing if the timeouts continue. The configuration for the page size is ok, but again I really doubt requesting 100 items is what is causing the issue, as that is still a very small number.

JustinASmith: Ah you're right. I overlooked a few things there. I agree, this would (if set up properly) likely slow things down anyways unless the data is very large (unlikely for 100 items).

For now I have the timeout in the CDK set to 30 minutes and the limit parameter configuration update. Since this was not something I could reproduce, I believe the best bet is for me to do some final smoke tests to ensure all is good before getting the ✅ and merging to verify this does or does not fix the TimeoutError for some captures.

JustinASmith: @williamhbaker updated the description of the PR and pushed a new update. LMK if this aligns with what you have in mind. I kept the page limit configuration in case a user does need to change this later on.

estuary-cdk/estuary_cdk/http.py
```diff
@@ -332,7 +332,12 @@ class HTTPMixin(Mixin, HTTPSession):
     token_source: TokenSource | None = None
 
     async def _mixin_enter(self, _: Logger):
-        self.inner = aiohttp.ClientSession()
+        self.inner = aiohttp.ClientSession(
+            timeout=aiohttp.ClientTimeout(
+                total=30 * 60,  # 30-minutes - default is 5-minutes
+                sock_connect=5 * 30, # 5-minutes - default is 30-seconds
```
williamhbaker, line 338: My understanding of this from looking at the docs & code briefly is that it if no `timeout` is set at all, the default timeout of `(total=5*60, sock_connect=30)` is used. But if you set any timeout, that default is not used, and an explicit timeout setting just `total=30*60` would end up with a `None` for `sock_connect`, essentially disabling it.

That seems like a simpler way to set this, unless there's some specific reason for having a 30 seconds `sock_connect` - if unset, I believe it won't have any particular limit and the entire request will just be governed by the 30 minute timeout.

williamhbaker (approved):
LGTM % comment about the timeouts

estuary-cdk/estuary_cdk/http.py
JustinASmith: Ah good point. I'll unset the `sock_connect` and let it default to `None`!

Issue #2738, opened 2025-04-29
estuary-cdk: connector status covering back-filling bindings


jgraettinger:
Model on SQL CDC status

PR #2790, opened 2025-05-14
estuary-cdk: connectorStatus logging for captures that use common.open_binding

williamhbaker:
**Description:**

Adds some general logging of connectorStatus events for captures that use common.open_binding, which is currently ~all of them.

It is modeled after the SQL captures, which indicate how many bindings are still backfilling, or if they are all backfilled.

The CDK is highly async, with tasks that can spawn subtasks, and the structure of the CDK status logging reflects that. There's no primary loop that drives backfills & incremental captures in turn (they all kind of run at the same time), so the status logging itself runs on a polling loop that checks for any changes in the metrics we are logging about.

Closes [#2738](https://github.com/estuary/connectors/issues/2738)

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2790)

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
+++ b/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
@@ -1,3 +1,4 @@
+from estuary_cdk.capture.connector_status import ConnectorStatus
 from pydantic import BaseModel
 from typing import Generic, Awaitable, Any, BinaryIO, Callable
 from logging import Logger
@@ -109,11 +110,13 @@ class BaseCaptureConnector(
 
                 task = Task(
                     log.getChild("capture"),
+                    ConnectorStatus(log, stopping, tg),
                     "capture",
                     self.output,
                     stopping,
                     tg,
                 )
+                log.event.status("Capture started")
                 await capture(task)
 
             # When capture() completes, the connector exits.
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -632,6 +632,7 @@ def open_binding(
     the state for each subtask in state.inc, state.backfill, or state.snapshot.
     """
 
+    task.connector_status.inc_binding_count()
     prefix = ".".join(binding.resourceConfig.path())
 
     if fetch_changes:
@@ -687,6 +688,7 @@ def open_binding(
             subtask_id: str | None = None,
         ):
             assert state and not isinstance(state, dict)
+            task.connector_status.inc_backfilling(binding_index)
             await _binding_backfill_task(
                 binding,
                 binding_index,
@@ -697,6 +699,7 @@ def open_binding(
                 task,
                 subtask_id,
             )
+            task.connector_status.dec_backfilling(binding_index)
 
         if isinstance(fetch_page, dict):
             assert resource_state.backfill and isinstance(resource_state.backfill, dict)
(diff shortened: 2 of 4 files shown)


estuary-cdk/estuary_cdk/capture/base_capture_connector.py
```diff
@@ -109,11 +110,13 @@ async def periodic_stop() -> None:
 
                 task = Task(
                     log.getChild("capture"),
+                    ConnectorStatus(log, stopping, tg),
                     "capture",
                     self.output,
                     stopping,
                     tg,
                 )
+                log.info("Capture started", {"eventType": "connectorStatus"})
```
Alex-Bair, line 119: Is this log recognized by the runtime as a connector event? This would log something like:
```
{
  "level": "INFO",
  "msg": "Capture started",
  "fields": {
    "args": {
      "eventType": "connectorStatus"
    }
  }
}
```
where the `eventType` is tucked into `fields.args`. I'm pretty sure the [runtime](https://github.com/estuary/flow/blob/81f5877a64ff2cd5caa34af866fe888f816acb5d/crates/runtime/src/container.rs#L285-L288) doesn't look deeper than `fields` for the `eventType` key.

When I was working on emitting `configUpdate` events, I added the capability to emit `connectorStatus` events as well in this [commit](https://github.com/estuary/connectors/commit/f53b8f6bd8283ed79901d9a042fc13d553130f1d) like `log.event.status("Something happened")`. It feels like that'd be helpful here.

JustinASmith (approved):
LGTM % a few comments/questions:

- @Alex-Bair comment on the connector event structure.
- Looks like it will achieve what we want, however, would we want the ability (as Alex mentioned) to do something like `log.event.status("Some important connector event")`? As you mentioned, the CDK is highly asynchronous and allowing this would likely not be desirable since the periodic connector status event emitted here would show up most of the time in the UI from my understanding.

estuary-cdk/estuary_cdk/capture/base_capture_connector.py
williamhbaker: Oh, yeah, that definitely wasn't going to work the way I had it. I switched to using the `log.event.status` and confirmed that it does get picked up correctly by the event handling in the runtime - thanks!

williamhbaker:
> * Looks like it will achieve what we want, however, would we want the ability (as Alex mentioned) to do something like `log.event.status("Some important connector event")`? As you mentioned, the CDK is highly asynchronous and allowing this would likely not be desirable since the periodic connector status event emitted here would show up most of the time in the UI from my understanding.

The latch/unlatch mechanism on `self.should_log` of the connector status periodic logger will keep it from logging anything unless bindings have finished backfilling. That said, there's definitely a possibility that a binding would finish backfilling around the same time as some other call of `log.event.status` and would end up clobbering it as the latest status that was logged. I don't think there's anything we could/should do about that though, since it's a completely valid sequence of events.

It seems like there is a potential source of confusion between regular logs vs. connector status events. We should keep using regular logs for important connector events that are specific to the connector. Connector status events are for these high-level things, IMO it's a little ambiguous when they should be used, but we seem to have settled on this pattern of using it to say if the capture is done backfilling or not for now. The little text box in the UI inherently cannot show any history (that's just in the regular logs!) so I think it's use will need to be limited & deliberate.

PR #2795, opened 2025-05-15
materialize-snowflake: use VARCHAR instead of TEXT for string columns

williamhbaker:
**Description:**

Snowflake will soon be changing the DATA_TYPE reported from INFORMATION_SCHEMA.COLUMNS for string columns from TEXT to VARCHAR. We always could have created columns as either TEXT or VARCHAR, but used TEXT for
convenience since that is what the DATA_TYPE said. Since the DATA_TYPE will be VARCHAR going forward, we'll use VARCHAR instead.

There needs to be a compatibility DDL check for columns still reported as TEXT though, since this Snowflake behavior is configurable as part of the 2025_03 behavior change bundle. It will be the default behavior soon, but there will probably still be some instances where the bundle is disabled by users for the foreseeable future.

This change is fairly simple, but also delicate, since it impacts how materialized field constraints are generated, and when the system would think it needs to backfill a table on new publications.

I've updated all the test snapshots against our test account with the behavior change enabled, and also manually ran through a couple of scenarios on a local stack.

To simulate what it will be like when merging this PR, prior to the behavior change being the default across Snowflake accounts:
1. Create a materialization of a collection with string fields, with the existing connector and behavior change off
2. Update the connector to using the image built from this PR. Everything continues to work as expected - data flowing, Validate RPCs return normally without `UNSATISFIABLE`
3. Enable the behavior change, and everything continues to work as expected

An alternate simulation where we hadn't rolled this out yet:
1. Create a materialization of a collection with string fields, with the existing connector and behavior change off
2. Enable the behavior change. Observe that the materialization does keep moving data, but if any attempts to publish it are made, `UNSATISFIABLE` constraints are returned since columns that the connector thinks should be `TEXT` are now `VARCHAR`
3. Switch the materialization to this new version. It publishes fine and runs as expected after that

**Workflow steps:**

**Notes for reviewers:**

I'm not sure what was going on with the `materialize-snowflake/.snapshots/TestValidateAndApplyMigrations` snapshot file being interpreted as binary format; it had some non-UTF8 bytes in it somehow and doesn't anymore since I've updated the snapshots for it, so I think it will be fine going forward.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2795)


materialize-snowflake/sqlgen.go
```diff
@@ -46,25 +46,27 @@ func (m timestampTypeMapping) valid() bool {
 	return m == timestampNTZ || m == timestampLTZ || m == timestampTZ
 }
 
-// snowflakeDialect returns a representation of the Snowflake SQL dialect.
+// VARCHAR also being compatible with "text" is a compatibility for Snowflake's
+// behavior change 2025_03/bcr-1960, which changes the DATA_TYPE of
+// INFORMATION_SCHEMA.COLUMNS from TEXT to VARCHAR for all string columns.
```
mdibaiee, line 51: Can we put a TODO here to remove this once 2025_03 is Generally Available?

mdibaiee (approved):
LGTM % note regarding TODO for removing this compatibility layer

PR #2796, opened 2025-05-15
source-sqlserver: Never emit nulls on NOT NULL column deletions

willdonnelly:
**Description:**

We have now observed multiple unrelated SQL Server captures in production where a `BIT NOT NULL` column is/was captured as a null value in a deletion event. This should never happen [1] and is a pretty clear-cut violation of the documented CDC behavior around deletions, but nonetheless it's happened and continues happening so we need to deal with it somehow.

This PR fixes the entire class of issues by extending the existing "omit null values of image/text/ntext column in deletes" logic to also apply to null values of any column which should be non-nullable according to the source table schema.

[1] I suspect that there's an actual database bug here, possibly related to the decoding of multiple `BIT NOT NULL` columns packed into whatever sort of bitfield the SQL Server WAL encoding uses. In all cases where we've observed this happening, the table had multiple `BIT NOT NULL` columns, of which only one was impacted in this way, and in at least one case it was also the last column of the table.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2796)

### DIFF
--- a/source-sqlserver/.snapshots/TestBitNotNullDeletion
+++ b/source-sqlserver/.snapshots/TestBitNotNullDeletion
@@ -1,0 +1,24 @@
+# ================================
+# Collection "acmeCo/test/dbo/test_bitnotnulldeletion_32831599": 16 Documents
+# ================================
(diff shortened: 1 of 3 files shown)


jgraettinger (approved):
LGTM

PR #2797, opened 2025-05-15
source-hubspot-native: hide access token auth in UI

Alex-Bair:
**Description:**

In order to publish this connector in the HubSpot marketplace, HubSpot requires that "setup documentation [and] in platform experience should not include an app that requires API keys in order to be installed". Essentially, users should only be presented the option to authenticate with OAuth in the UI.

There are already existing captures authenticating with access tokens, so we can't simply strip out support for access token authentication. Patching model_json_schema to remove AccessToken makes OAuth2 the only authentication option in the UI, and effectively "hides" the capability to use access tokens. This should satisfy HubSpot's requirement while retaining support for access token authentication for already created captures.

Editing an existing capture that authenticates with an access token will either have to be done via `flowctl` to maintain the access token auth option or will have to be re-authenticated with OAuth2 in the UI.

Addresses a portion of this [draft issue](https://github.com/orgs/estuary/projects/13?pane=issue&itemId=90884122).

**Workflow steps:**

This connector's docs need updated to reflect that OAuth2 is the only way to authenticate the connector (this is _also_ required before listing the connector in HubSpot's marketplace).

**Notes for reviewers:**

- access token auth is no longer shown as an authentication option in the UI.
image
- an existing capture using access token authentication still works.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2797)

### DIFF
--- a/source-hubspot-native/source_hubspot_native/models.py
+++ b/source-hubspot-native/source_hubspot_native/models.py
@@ -88,6 +88,37 @@ class EndpointConfig(BaseModel):
         default=False,
     )
 
+    # In order to publish this connector in the HubSpot marketplace, HubSpot requires
+    # that "setup documentation [and] in platform experience should not include an app that
+    # requires API keys in order to be installed". Essentially, users should only be presented
+    # the option to authenticate with OAuth in the UI.
+    # 
+    # There are already existing captures authenticating with access tokens, so we can't
+    # simply strip out support for access token authentication. Patching model_json_schema to remove
+    # AccessToken makes OAuth2 the only authentication option in the UI, and effectively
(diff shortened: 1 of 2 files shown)


williamhbaker (approved):
LGTM

I'm not sure if its helpful here, but maybe this new UI feature could be used? https://github.com/estuary/ui/pull/1614

Alex-Bair:
I had hoped it would, but unfortunately it didn't (or at least I couldn't get it to work after shoving `x-hidden-field` in a ton of spots). The closest I could get was hiding the contents of the access token auth option, but not the actual tab in the Authentication drop down. :/

image

PR #2799, opened 2025-05-16
source-pendo: add `TrackType` & `Visitor` streams, capture from all applications

Alex-Bair:
**Description:**

This PR's scope includes:
- Capturing data for all applications within a single Pendo subscription.
- Adding the [`TrackType`](https://engageapi.pendo.io/#9f83f648-4fe7-45db-b30d-963679af6304) and [`Visitor`](https://engageapi.pendo.io/#dd943e68-5bff-4a1a-9891-b55638ae2c3d) streams that capture the same named resources from Pendo's aggregation API.

Some existing streams (`Page`, `Guide`, etc.) could also be captured incrementally via the aggregation API, and that would be better than constantly taking snapshots of all pages, guides, etc. like we currently do. But that would also require changing the primary key for those streams, and that'd impact existing Pendo captures. Changing those streams from full refresh to incremental can be done later when we have more time to generalize the connector's code & potentially bump the connector version.

**Workflow steps:**

Documentation should be updated to reflect the added streams.

**Notes for reviewers:**

Tested on a local stack. Confirmed data is captured for `Track` & `Visitor` streams. Confirmed data for more than just the default application is captured from the user's Pendo subscription.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2799)

### DIFF
--- a/source-pendo/source_pendo/api.py
+++ b/source-pendo/source_pendo/api.py
@@ -56,12 +56,15 @@ def generate_events_body(
             "pipeline": [
                 {
                     "source": {
-                        entity: None,
+                        entity: {
+                            # Capture events for all applications within this Pendo subscription.
+                            "appId": "expandAppIds(\"*\")",
+                        },
                         "timeSeries": {
                             "period": "hourRange",
                             "first": f"{lower_bound}",
                             # If an upper bound isn't specified, retrieve all events up to the present.
-                            "last": f"{upper_bound or "now()"}"
+                            "last": f"{upper_bound or 'now()'}"
                         }
                     }
                 },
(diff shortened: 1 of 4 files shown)


source-pendo/source_pendo/api.py
```diff
@@ -476,3 +552,171 @@ async def backfill_aggregated_events(
...
+
+        yield resource
+
+    if doc_count == 0:
+        # If there were no documents, don't update the cursor.
+        return
+    elif last_dt > log_cursor:
+        # If there was at least one document and the last document's timestamp is
+        # later than the cursor, update the cursor.
+        yield last_dt
```
williamhbaker, line 618: If I'm understanding this correctly with the inclusive lower bound on the time range query, every time this runs it will get at least 1 duplicated document. If that is indeed the case, you may be able to use the same "emitted changes cache" strategy used [here](https://github.com/estuary/connectors/blob/df778e76fe2d0eab10a1f91a04d7d756c5741200/source-sage-intacct/source_sage_intacct/api.py#L103-L104) and elsewhere to filter out these duplicates. That example is for Sage Intacct, which is doing something quite similar as to here I think, although I made it more complicated by only checkpointing a cursor value if we were certain we've already seen all the records at that timestamp, but I think at least the cache approach could apply here.

williamhbaker (approved):
LGTM

source-pendo/source_pendo/api.py
Alex-Bair: You're right, the cache approach is applicable here. We're using the emitted changes cache for incremental resources now, thanks for mentioning it.

PR #2802, opened 2025-05-19
Phil/update flow

psFried:
Updates the flow dependencies so that these will work with the new specs, which now include inactive bindings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2802)


dyaffe (approved):
LGTM

PR #2805, opened 2025-05-19
source-mysql-batch: Add SourcedSchema feature flag

willdonnelly:
**Description:**

This PR makes two changes to `source-mysql-batch`:
1. Adds full column type discovery. All columns will now be included in discovered schemas with (hopefully) appropriate JSON schema type constraints that match the preexisting capture output.
2. Adds the `emit_sourced_schemas` feature flag, defaulting to off. When enabled, the capture will run discovery and output SourcedSchema updates immediately prior to polling a binding (subject to a minimum interval between rediscoveries, which is currently set to 5 minutes).

This is part of https://github.com/estuary/flow/issues/1919

**Workflow steps:**

All capture tasks will now autodiscover a more detailed write schema. In theory this could lead to failures of preexisting captures if the discovered schema generation doesn't actually match what we output for a specific column type, but such errors should be easy enough to fix as a fast-follow so long as I merge this at a time when I'm prepared to keep a close eye on things for the next few hours.

To start getting SourcedSchema updates, the user will need to set the `emit_sourced_schemas` feature flag. This will probably become the default at some point in the near future, either for new captures or across the board.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2805)


source-mysql-batch/driver.go
```diff
@@ -575,28 +309,105 @@ func (c *capture) Run(ctx context.Context) error {
...
+func (c *capture) emitSourcedSchemas(ctx context.Context) error {
+	if !c.Config.Advanced.parsedFeatureFlags["emit_sourced_schemas"] {
+		return nil
+	}
+
+	c.shared.Lock()
+	defer c.shared.Unlock()
+
+	// Skip re-running discovery and emitting schemas if we just did it.
+	if time.Since(c.shared.lastDiscoveryTime) < minDiscoveryInterval {
```
williamhbaker, line 321: Is this `lastDiscoveryTime` ever getting set? I can't seem to find where it is getting set to "now", but I would expect it to get set a little further down in this function.

williamhbaker (approved):
LGTM, with one question about setting the `lastDiscoveryTime` time

source-mysql-batch/driver.go
willdonnelly: :man_facepalming: Good catch, it's never getting set

PR #2807, opened 2025-05-19
source-chargebee-native: timeouts

JustinASmith:
**Description:**

Adds retry logic to `asyncio.TimeoutErrors` that happen `source-chargebee-native` and are not easily reproduced.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2807)

### DIFF
--- a/source-chargebee-native/source_chargebee_native/api.py
+++ b/source-chargebee-native/source_chargebee_native/api.py
@@ -2,6 +2,8 @@ from datetime import datetime, timedelta, UTC
 from logging import Logger
 from typing import AsyncGenerator, TypeVar
 
+import asyncio
+
 from estuary_cdk.http import HTTPSession
 from estuary_cdk.capture.common import (
     PageCursor,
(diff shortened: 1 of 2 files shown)


source-chargebee-native/source_chargebee_native/api.py
```diff
@@ -62,22 +64,65 @@ async def _fetch_resource_data(
...
+
+        if not response.list:
+            return [], response.next_offset
+
+        return response.list, response.next_offset
+
+    max_retries = 10
+    retry_count = 0
+
+    while retry_count < max_retries:
```
williamhbaker, line 97: I would change this condition to `while True`, and then remove the `return [], None` at the end of this function. I was a little confused at first how we could ever get to the final `return [], None` - and indeed it appears we can't, so this will make it a little more obvious about what is happening with the retry logic.

williamhbaker (approved):
LGTM % comment

PR #2811, opened 2025-05-20
source-firestore: Validate minBackfillInterval settings

willdonnelly:
**Description:**

Renames `min_backfill_interval` to `minBackfillInterval` since we only have one capture using it at present and that's more consistent with other resource config settings (though sadly it's inconsistent with other advanced config properties -- there is no solution here that's consistent with everything because the state of these names is already an incoherent mess).

Adds validation logic so that non-empty settings on the capture or resource specs for the `minBackfillInterval` property are validated in the Validate() method as is appropriate.

Adds a validation regexp to the JSON schema so that users will be prevented from adding invalid settings in the UI even before it reaches the capture's validation logic.

**Workflow steps:**

It should be harder for users to end up with incorrect settings for this property. I'll go and fix the one capture that was previously using it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2811)

### DIFF
--- a/source-firestore/pull.go
+++ b/source-firestore/pull.go
@@ -260,10 +260,10 @@ func computeBackfillStartTime(cfg *config, res *resource, prevBackfill *backfill
 	// - Advanced Endpoint Config
 	// - Default to 24h if no cursor or 5m if there's a cursor
 	var delay time.Duration
-	if d, err := time.ParseDuration(res.MinBackfillInterval); err == nil && d > 0 {
-		delay = d
-	} else if d, err := time.ParseDuration(cfg.Advanced.MinBackfillInterval); err == nil && d > 0 {
-		delay = d
+	if res.MinBackfillInterval != "" {
+		delay, _ = time.ParseDuration(res.MinBackfillInterval) // Cannot fail, already validated
+	} else if cfg.Advanced.MinBackfillInterval != "" {
+		delay, _ = time.ParseDuration(cfg.Advanced.MinBackfillInterval) // Cannot fail, already validated
 	} else if hasRestartCursor {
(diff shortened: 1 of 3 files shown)


source-firestore/pull.go
```diff
@@ -260,10 +260,10 @@ func computeBackfillStartTime(cfg *config, res *resource, prevBackfill *backfill
 	// - Advanced Endpoint Config
 	// - Default to 24h if no cursor or 5m if there's a cursor
 	var delay time.Duration
-	if d, err := time.ParseDuration(res.MinBackfillInterval); err == nil && d > 0 {
-		delay = d
-	} else if d, err := time.ParseDuration(cfg.Advanced.MinBackfillInterval); err == nil && d > 0 {
-		delay = d
+	if res.MinBackfillInterval != "" {
+		delay, _ = time.ParseDuration(res.MinBackfillInterval) // Cannot fail, already validated
```
jgraettinger, line 264: this is somewhat unsafe, as it doesn't account for changes made to validation logic and the propagation delay as tasks are re-published, and it would fail "open" in that it backfills over and over as quickly as possible.

Probably better to panic here, if we want to avoid threading through an error.

willdonnelly: I'm not sure I understand, either there's a correct value for the appropriate property name (in which case we're fine) or there's a non-empty and incorrect value, in which case `(resource).Validate()` will fail when it's called via `UnmarshalStrict()` during the task startup?

But sure, I'm happy to be more defensive here just in case.

jgraettinger: Ah, I missed that Validate is called on every startup and not just when driven by a control plane Validate RPC.
I retract this comment.

PR #2812, opened 2025-05-21
source-stripe-native: dynamic subtask state initialization

JustinASmith:
**Description:**

This pull request introduces changes to the `source-stripe-native/source_stripe_native/resources.py` file to improve the initialization logic for resource state handling. The most notable updates include adding type annotations for `all_bindings`, introducing additional validation for state properties, and restructuring the logic for initializing `fetch_changes_fns` and `fetch_page_fns`.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2812)

### DIFF
--- a/source-stripe-native/source_stripe_native/resources.py
+++ b/source-stripe-native/source_stripe_native/resources.py
@@ -233,6 +233,9 @@ def base_object(
     It requires a single, parent stream with a valid Event API Type
     """
     all_account_ids = [platform_account_id, *connected_account_ids]
+    initial_state = _create_initial_state(
+        all_account_ids if connected_account_ids else platform_account_id
+    )
 
     def open(
         binding: CaptureBinding[ResourceConfig],
@@ -258,18 +261,31 @@ def base_object(
                 http,
             )
         else:
-            fetch_changes_fns = {
-                account_id: functools.partial(
+            fetch_changes_fns = {}
+            fetch_page_fns = {}
+
+            for account_id in all_account_ids:
+                # If there's persisted state and the connector found a new account id that's
+                # not present in that state, dynamically add that account id and that account id's
+                # initial state to it.
+                if (
+                    isinstance(state.inc, dict)
+                    and isinstance(state.backfill, dict)
+                    and isinstance(initial_state.inc, dict)
+                    and isinstance(initial_state.backfill, dict)
+                ):
+                    if account_id not in state.inc and account_id not in state.backfill:
+                        state.inc[account_id] = initial_state.inc[account_id]
+                        state.backfill[account_id] = initial_state.backfill[account_id]
+
+                fetch_changes_fns[account_id] = functools.partial(
                     fetch_incremental,
                     cls,
                     platform_account_id,
                     account_id,
                     http,
                 )
-                for account_id in all_account_ids
-            }
-            fetch_page_fns = {
-                account_id: functools.partial(
+                fetch_page_fns[account_id] = functools.partial(
                     fetch_backfill,
                     cls,
                     start_date,
@@ -277,8 +293,6 @@ def base_object(
                     account_id,
                     http,
                 )
-                for account_id in all_account_ids
-            }
 
         open_binding(
             binding,
@@ -288,10 +302,6 @@ def base_object(
             fetch_changes=fetch_changes_fns,
             fetch_page=fetch_page_fns,
         )
-
-    initial_state = _create_initial_state(
-        all_account_ids if connected_account_ids else platform_account_id
-    )
 
     return Resource(
         name=cls.NAME,
@@ -318,6 +328,9 @@ def child_object(
     """
 
     all_account_ids = [platform_account_id, *connected_account_ids]
+    initial_state = _create_initial_state(
+        all_account_ids if connected_account_ids else platform_account_id
+    )
 
     def open(
         binding: CaptureBinding[ResourceConfig],
@@ -345,8 +358,24 @@ def child_object(
                 http,
             )
         else:
-            fetch_changes_fns = {
-                account_id: functools.partial(
+            fetch_changes_fns = {}
+            fetch_page_fns = {}
+
+            for account_id in all_account_ids:
+                # If there's persisted state and the connector found a new account id that's
+                # not present in that state, dynamically add that account id and that account id's
+                # initial state to it.
+                if (
+                    isinstance(state.inc, dict)
+                    and isinstance(state.backfill, dict)
+                    and isinstance(initial_state.inc, dict)
+                    and isinstance(initial_state.backfill, dict)
+                ):
+                    if account_id not in state.inc and account_id not in state.backfill:
+                        state.inc[account_id] = initial_state.inc[account_id]
+                        state.backfill[account_id] = initial_state.backfill[account_id]
+
+                fetch_changes_fns[account_id] = functools.partial(
                     fetch_incremental_substreams,
                     cls,
                     child_cls,
@@ -354,10 +383,7 @@ def child_object(
                     account_id,
                     http,
                 )
-                for account_id in all_account_ids
-            }
-            fetch_page_fns = {
-                account_id: functools.partial(
+                fetch_page_fns[account_id] = functools.partial(
                     fetch_backfill_substreams,
                     cls,
                     child_cls,
@@ -366,8 +392,6 @@ def child_object(
                     account_id,
                     http,
                 )
-                for account_id in all_account_ids
-            }
 
         open_binding(
             binding,
@@ -377,10 +401,6 @@ def child_object(
             fetch_changes=fetch_changes_fns,
             fetch_page=fetch_page_fns,
         )
-
-    initial_state = _create_initial_state(
-        all_account_ids if connected_account_ids else platform_account_id
-    )
 
     return Resource(
         name=child_cls.NAME,
@@ -410,6 +430,9 @@ def split_child_object(
     """
 
     all_account_ids = [platform_account_id, *connected_account_ids]
+    initial_state = _create_initial_state(
+        all_account_ids if connected_account_ids else platform_account_id
+    )
 
     def open(
         binding: CaptureBinding[ResourceConfig],
@@ -436,18 +459,31 @@ def split_child_object(
                 http,
             )
         else:
-            fetch_changes_fns = {
-                account_id: functools.partial(
+            fetch_changes_fns = {}
+            fetch_page_fns = {}
+
+            for account_id in all_account_ids:
+                # If there's persisted state and the connector found a new account id that's
+                # not present in that state, dynamically add that account id and that account id's
+                # initial state to it.
+                if (
+                    isinstance(state.inc, dict)
+                    and isinstance(state.backfill, dict)
+                    and isinstance(initial_state.inc, dict)
+                    and isinstance(initial_state.backfill, dict)
+                ):
+                    if account_id not in state.inc and account_id not in state.backfill:
+                        state.inc[account_id] = initial_state.inc[account_id]
+                        state.backfill[account_id] = initial_state.backfill[account_id]
+
+                fetch_changes_fns[account_id] = functools.partial(
                     fetch_incremental,
                     child_cls,
                     platform_account_id,
                     account_id,
                     http,
                 )
-                for account_id in all_account_ids
-            }
-            fetch_page_fns = {
-                account_id: functools.partial(
+                fetch_page_fns[account_id] = functools.partial(
                     fetch_backfill_substreams,
                     cls,
                     child_cls,
@@ -456,8 +492,6 @@ def split_child_object(
                     account_id,
                     http,
                 )
-                for account_id in all_account_ids
-            }
 
         open_binding(
             binding,
@@ -467,10 +501,6 @@ def split_child_object(
             fetch_changes=fetch_changes_fns,
             fetch_page=fetch_page_fns,
         )
-
-    initial_state = _create_initial_state(
-        all_account_ids if connected_account_ids else platform_account_id
-    )
 
     return Resource(
         name=child_cls.NAME,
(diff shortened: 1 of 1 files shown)


JustinASmith:
@williamhbaker @Alex-Bair This PR is still in progress, however, I need to step away very soon for personal reasons and wanted to get any feedback while I am away. The PR is getting close, but I expect there to be holes poked in the approach and/or test cases I have or do not have.

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -602,10 +618,59 @@ async def _run(task: Task):
...
+) -> ResourceState:
+    subtask_ids = []
+    if fetch_changes and isinstance(fetch_changes, dict):
+        subtask_ids = list(fetch_changes.keys())
+    if fetch_page and isinstance(fetch_page, dict):
+        subtask_ids.extend(list(fetch_page.keys()))
+
+    if subtask_ids:
+        assert resource_state.inc and isinstance(resource_state.inc, dict)
+        assert resource_state.backfill and isinstance(resource_state.backfill, dict)
```
Alex-Bair, line 641: These assertions will cause the connector to crash if either the incremental or backfill states are not `dict`s, right? I'm thinking of a case where the incremental state is a `dict` but the backfill state isn't, like in `source-sage-intacct`.

It feels like instead of only dynamically adding subtask states when both `state.inc` and `state.backfill` are `dict`s, we should also handle the case where only `state.inc` is a `dict`. And _maybe_ where only `state.backfill` is a `dict` (although we don't currently have any connectors that follow that pattern & I can't think of a situation that would call for it).

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -349,8 +349,24 @@ class AssociatedDocument(Generic[_BaseDocument]):
 (use `ResourceConfig.interval`).
 """
 
+InitializeSubtaskStatesFn = Callable[[Logger, ResourceState, str], ResourceState]
```
Alex-Bair, line 352: Thinking out loud, do we need an `InitializeSubtaskStatesFn`? When a `Resource` is initialized, it should already have an `initial_state` that contains the initial state for all subtasks, including any new ones. My initial thought on how this could work was that when the connector starts up & sees there's a `fetch_changes`/`fetch_pages` function without an associated state, we'd reach inside that `Resource.initial_state`, pluck out the missing initial state, then place it into the current resource state.

I'm likely missing something though, and I'm mainly trying to understand the benefits of the `InitializeSubtaskStatesFn` approach over leveraging the existing `Resource.initial_state`.

Alex-Bair (commented):
Had a couple comments/questions. I'm sure you and Will have already talked about the approach, so feel free to update me if my comments go against something you guys have already discussed.

estuary-cdk/estuary_cdk/capture/common.py
williamhbaker: :+1: The case of backfills not using subtasks and incremental using subtasks is probably the most common that we'll have going forward. And since we're doing this, we might as well make it work for all scenarios, including:
- Both `fetch_page` and `fetch_changes` are using subtasks and a new `fetch_page` subtask appears
- Both `fetch_page` and `fetch_changes` are using subtasks and a new `fetch_changes` subtask appears
- `fetch_page` is not using subtasks, but `fetch_changes` is and a new `fetch_changes` subtask appears
- `fetch_changes` is not using subtasks, but `fetch_page` is and a new `fetch_page` subtask appears

estuary-cdk/estuary_cdk/capture/common.py
Alex-Bair: > But that original initial state could have been from a long time ago [...]

Isn't each resource's `initial_state` re-evaluated each time the connector starts up, and we wouldn't carry over old `initial_state`s from previous invocations? And if a new subtask is identified by the capture when opening a binding, that subtask's state would be present in the `initial_state` (but not in the resource's persisted state)?

My understanding of what's happening is:
1. The capture starts up for the first time.
2. When opening a binding, the resource's `initial_state` is evaluated. This is not the resource's actual state, but rather what the resource's state should be if it's initialized, like after triggering a backfill.
3. Since the resource's persisted state is `None` initially, we reassign its persisted state to be `initial_state`.
4. The capture runs for a while, then restarts. Assume the capture determines there should be a new subtask `subtask4` after restarting.
5. We open each binding again, and each resource's `initial_state` is re-evaluated. The `initial_state` now looks like:
```json
{
  "subtask1": "new_initial_subtask1_state",
  "subtask2": "new_initial_subtask2_state",
  "subtask3": "new_initial_subtask3_state",
  "subtask4": "new_initial_subtask4_state"
}
```
6. The capture sees that a new subtask, `subtask4`, exists in the `fetch_changes`/`fetch_pages` dictionaries. But in the resource's persisted state, `subtask4` doesn't exist. The connector ends up crashing because of this.

We've been addressing this by backfilling bindings (i.e. replacing the resource's actual state with `initial_state` that contains all of the current subtasks). My thought was that when this happens, we'd instead reach inside `initial_state`, pull out the values for the new subtasks, then place them in the connector's persisted state. Conceptually, it'd be like "backfilling" newly added subtasks.

williamhbaker: It is yeah, that's a good observation.

Previously I wasn't sure how that initial state would get to where it needs to be in `common.open_binding`, but looking at it again now, we've already got `resource.open` set up to receive a list of all of the resolved bindings, which should have their initial states included: https://github.com/estuary/connectors/blob/587ad545ff62c2d8639c891ceafc2f292a85da97/estuary-cdk/estuary_cdk/capture/common.py#L597

That comes back around to the specific capture implementation's creation of a `Resource`, for example here for `source-strip-native`: https://github.com/estuary/connectors/blob/587ad545ff62c2d8639c891ceafc2f292a85da97/source-stripe-native/source_stripe_native/resources.py#L225-L290

We _could_ thread that `all_bindings` parameter back into `common.open` here and I think it would enable what you are suggesting.

Reflecting on this though, I am actually included to think that these dynamic subtask initializations would be better off in the connector implementation itself, in this case `source-strip-native`. Dynamic subtasks still feels like a one-off thing to me, since only Stripe is using them, and it would be great to keep the extra complexity out of the CDK until we see more use for this (unless I am forgetting something).

So I would propose we move the new state-filling logic out of the CDK and directly into `source-stripe-native`, specifically into its `base_object` function. It could perhaps be modified to have an `open` closure something like this:

```python
    def open(
        binding_index: int,
        state: ResourceState,
        task: Task,
        all_bindings: list[tuple[CaptureBinding, Resource]]
    ):
        if not connected_account_ids:
            fetch_changes_fns = functools.partial(
                fetch_incremental,
                cls,
                platform_account_id,
                None,
                http,
            )
            fetch_page_fns = functools.partial(
                fetch_backfill,
                cls,
                start_date,
                platform_account_id,
                None,
                http,
            )
        else:
            for (idx, b) in enumerate(all_bindings):
                if idx == binding_index:
                    initial_state = b[1].initial_state
                    break
                
            assert isinstance(initial_state, ResourceState)
            assert isinstance(state.inc, dict)
            assert isinstance(state.backfill, dict)
            assert isinstance(initial_state.inc, dict)
            assert isinstance(initial_state.backfill, dict)

            for account_id in all_account_ids:
                if account_id not in state.inc:
                    state.inc[account_id] = initial_state.inc[account_id]
                if account_id not in state.backfill:
                    state.backfill[account_id] = initial_state.backfill[account_id]

            fetch_changes_fns = {
                account_id: functools.partial(
                    fetch_incremental,
                    cls,
                    platform_account_id,
                    account_id,
                    http,
                )
                for account_id in all_account_ids
            }
            fetch_page_fns = {
                account_id: functools.partial(
                    fetch_backfill,
                    cls,
                    start_date,
                    platform_account_id,
                    account_id,
                    http,
                )
                for account_id in all_account_ids
            }

        open_binding(
            binding,
            binding_index,
            state,
            task,
            fetch_changes=fetch_changes_fns,
            fetch_page=fetch_page_fns,
        )
```

Overall I think this would simplify the implementation for the immediate use case...we wouldn't have to worry about all the permutations for dynamic backfill / incremental parts, since it would be specific for Stripe's connected accounts.

Alex-Bair: Yep, I'm on board with that if we're treating this dynamic subtask initialization behavior as a one-off behavior for `source-stripe-native` (which I agree with. I'd rather see if/how more connectors use this dynamic behavior before building it into the CDK).

JustinASmith: Great observations guys! Thanks for great feedback on this while I was away. I agree that for now we can simplify this and do what @williamhbaker has shown based on @Alex-Bair suggestions. If we find later that this makes more sense to move into the CDK we at least know it should be possible, but may require more logic to handle the various use cases that may come up. I'll update this PR with the simplified approach where `source-stripe-native` contains the `open` closure that handles this via the `initial_state`.

JustinASmith: @Alex-Bair @williamhbaker let me know if the update I just pushed aligns with what you guys are saying. All CDK changes were removed and the `source-stripe-native` connector `open` closure in `base_object` was updated to set states to the `initial_state` for accounts that are not in current state.

JustinASmith: Testing wise, this is a bit more nuanced to test. I can certainly aim to do more testing, but thus far I have not went through setting up a personal Stripe test account with connected accounts added after connector is set up and running.

source-stripe-native/source_stripe_native/resources.py
```diff
@@ -258,27 +264,42 @@ def open(
...
+            assert isinstance(initial_state.backfill, dict)
+
+            fetch_changes_fns = {}
+            fetch_page_fns = {}
+
+            for account_id in all_account_ids:
+                if account_id not in state.inc:
+                    state.inc[account_id] = initial_state.inc[account_id]
+                if account_id not in state.backfill:
+                    state.backfill[account_id] = initial_state.backfill[account_id]
```
Alex-Bair, line 286: I'd suggest adjusting this to take advantage of the `initial_state` we're already calculating further down in the `base_object` function by moving it up to the beginning of the function. Then `base_object` could look like:
```python
def base_object(
    cls,
    http: HTTPSession,
    start_date: datetime,
    platform_account_id: str,
    connected_account_ids: list[str],
) -> Resource:
    """Base Object handles the default case from source-stripe-native
    It requires a single, parent stream with a valid Event API Type
    """
    all_account_ids = [platform_account_id, *connected_account_ids]
    initial_state = _create_initial_state(
        all_account_ids if connected_account_ids else platform_account_id
    )

    def open(
        binding_index: int,
        state: ResourceState,
        task: Task,
        all_bindings,
    ):
        if not connected_account_ids:
            fetch_changes_fns = functools.partial(
                fetch_incremental,
                cls,
                platform_account_id,
                None,
                http,
            )
            fetch_page_fns = functools.partial(
                fetch_backfill,
                cls,
                start_date,
                platform_account_id,
                None,
                http,
            )
        else:
            fetch_changes_fns = {}
            fetch_page_fns = {}

            for account_id in all_account_ids:
                # If there's persisted state and the connector found a new account id that's
                # not present in that state, dynamically add that account id and that account id's
                # initial state to it.
                if (
                    isinstance(state.inc, dict)
                    and isinstance(state.backfill, dict)
                    and isinstance(initial_state.inc, dict)
                    and isinstance(initial_state.backfill, dict)
                ):
                    if account_id not in state.inc and account_id not in state.backfill:
                        state.inc[account_id] = initial_state.inc[account_id]
                        state.backfill[account_id] = initial_state.backfill[account_id]

                fetch_changes_fns[account_id] = functools.partial(
                    fetch_incremental,
                    cls,
                    platform_account_id,
                    account_id,
                    http,
                )
                fetch_page_fns[account_id] = functools.partial(
                    fetch_backfill,
                    cls,
                    start_date,
                    platform_account_id,
                    account_id,
                    http,
                )

        open_binding(
            binding,
            binding_index,
            state,
            task,
            fetch_changes=fetch_changes_fns,
            fetch_page=fetch_page_fns,
        )

    return Resource(
        name=cls.NAME,
        key=["/id"],
        model=cls,
        open=open,
        initial_state=initial_state,
        initial_config=ResourceConfig(name=cls.NAME, interval=timedelta(minutes=5)),
        schema_inference=True,
    )
```

estuary-cdk/estuary_cdk/capture/common.py
Alex-Bair: Those changes line up with what I was thinking. We'll also need to do the same thing in the `child_object`, `split_child_object`, `usage_records`, and `no_events_objects` functions.

I also have a couple comments on the changes, but nothing major.

source-stripe-native/source_stripe_native/resources.py
```diff
@@ -239,7 +240,12 @@ def open(
         binding_index: int,
         state: ResourceState,
         task: Task,
-        all_bindings,
+        all_bindings: list[
+                tuple[
+                    CaptureBinding[ResourceConfig],
+                    "Resource[BaseDocument, ResourceConfig, ResourceState]",
+                ]
+            ],
```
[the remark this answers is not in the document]
JustinASmith: The only problem with that is type hints/linting will complain. I can add a `# type: ignore` to likely get around this. Namely the line `initial_state = b[1].initial_state` will complain.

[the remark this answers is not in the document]
JustinASmith: Disregard. The next suggestion would remove this as a concern.

source-stripe-native/source_stripe_native/resources.py
JustinASmith: Ready for one final review! 🤞 @Alex-Bair

PR #2822, opened 2025-05-21
source-bigquery-batch: Add SourcedSchema feature flag

willdonnelly:
**Description:**

As with https://github.com/estuary/connectors/pull/2798 and https://github.com/estuary/connectors/pull/2805, this PR modifies another batch SQL connector to:

1. Implement full column and type discovery. All columns of discovered tables will now be included in discovered (write) schemas with appropriate JSON schema types, rather than just primary-key columns with somewhat half-assed type translation.
2. Add the `emit_sourced_schema` feature flag, default off. When enabled, the capture will run discovery and output SourcedSchema updates immediately prior to polling a binding (subject to a minimum interval between rediscoveries, which is currently set to 5 minutes).

This is part of https://github.com/estuary/flow/issues/1919

**Workflow steps:**

All capture tasks will now autodiscover a more detailed write schema. In theory this could lead to failures of preexisting captures if the discovered schema generation doesn't actually match what we output for a specific column type, but such errors should be easy enough to fix as a fast-follow so long as I merge this at a time when I'm prepared to keep a close eye on things for the next few hours.

To start getting SourcedSchema updates, the user will need to set the emit_sourced_schemas feature flag. This will probably become the default at some point in the near future, either for new captures or across the board.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2822)


Alex-Bair (approved):
LGTM

PR #2824, opened 2025-05-21
source-jira-native: rename to source-jira-import

Alex-Bair:
**Description:**

The current `source-jira-native` isn't actually a native connector built with our CDK; it's an imported Airbyte connector that's wrapped with our CDK. This misnaming has been constantly confusing, and would be even more so when we introduce an _actually_ native Jira capture soon.

To address this, I'm renaming the current `source-jira-native` to be `source-jira-import` to accurately reflect what type of connector it is. This will open up the `source-jira-native` name for the actually native connector.

The transition plan is to:
- Rename the connector (done in this commit)
- Build and publish new `source-jira-import` packages to GitHub (handled by our CI).
- Switch all existing `source-jira-native` captures over to using the newly published `source-jira-import` package (I'll handle this manually).

**Workflow steps:**

The connector's documentation should be updated to reflect the package name is now changed to `source-jira-import`. And probably the documentation's URL/short link too.

**Notes for reviewers:**

Tested switching over from the current `source-jira-native:v2` image to `source-jira-import:alex-test` (which is just a renamed `source-jira-native:v2` image) on a production task I created. Confirmed the connector still worked afterwards. 

Rows in the `connectors` and `connector_tags` tables should be updated/added to reflect this renaming prior to switching the current `source-jira-native` captures over to the renamed image.


williamhbaker (approved):
LGTM

PR #2833, opened 2025-05-22
source-boilerplate: Ensure stdout buffer flush on exit

willdonnelly:
**Description:**

We have a 1MiB buffer on writes to stdout, which is normally flushed after each non-document message. (The reason for that being that it improves throughput if we allow multiple docs in a transaction to be batched up into a single stdout write).

The problem arises if we output >1MiB worth of data in a single Flow transaction, and then the connector errors out before emitting a checkpoint. One way this might happen is if a fatal error occurs partway through a backfill chunk, where the total amount of data is >1MiB.

Since the buffer is 1MiB and we've already written >1MiB of data to it, it will have flushed to stdout at least once. But the buffer doesn't care about our message framing, so it will likely have split the final message in half. Which is fine in normal operation since we'll eventually write a checkpoint and flush the remaining transaction data, but when a fatal error occurs we never reach the checkpoint and flush, which means we just exit with a partial final message output.

This commit fixes that by ensuring we always flush the buffer on shutdown regardless of error status.

I don't believe there's an equivalent fix to be made in the materialization boilerplate because the output writer there is unbuffered.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2833)

### DIFF
--- a/source-boilerplate/boilerplate.go
+++ b/source-boilerplate/boilerplate.go
@@ -73,7 +73,7 @@ func RunMain(connector Connector) {
 	log.WithField("eventType", "connectorStatus").Info("Initializing connector")
 
 	var ctx, _ = signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
-	var stream pc.Connector_CaptureServer
+	var stream streamCodec
 
 	switch codec := getEnvDefault("FLOW_RUNTIME_CODEC", "proto"); codec {
 	case "proto":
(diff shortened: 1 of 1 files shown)


source-boilerplate/boilerplate.go
```diff
@@ -315,7 +321,15 @@ func (out *PullOutput) SourcedSchema(binding int, schema json.RawMessage) error
 	return nil
 }
 
-func newProtoCodec(ctx context.Context) pc.Connector_CaptureServer {
+type streamCodec interface {
+	pc.Connector_CaptureServer
+
+	// Since we buffer our writes internally, we need to be able to close the codec
+	// during shutdown and guarantee that the buffers will be flushed properly.
+	Close() error
```
jgraettinger, line 329: nit: `Flush() error` instead of `Close() error` ? No actual closing happening.

jgraettinger (approved):
LGTM

PR #2834, opened 2025-05-22
source-postgres: Debug logging for empty deletion

willdonnelly:
**Description:**

As explained in the comments, it should never be possible for a PostgreSQL CDC deletion event to be empty, because the deletion has to contain the table's REPLICA IDENTITY columns and you can't delete from a published table with an empty replica identity.

And yet, somehow this appears to be happening on a production task, so we need more information about what's going on. This adds a check for the should-be-impossible situation, and if it occurs it logs everything I can think of that might be useful before erroring out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2834)

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -689,6 +689,46 @@ func (s *replicationStream) decodeChangeEvent(
 		return nil, fmt.Errorf("'after' tuple: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #2837, opened 2025-05-23
source-http-ingest: set resource paths in discover response

psFried:
Populates the resource path in discover responses, which is once again required.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2837)


jgraettinger (approved):
LGTM

PR #2842, opened 2025-05-23
materialize-azure-blob-parquet: new connector

williamhbaker:
**Description:**

Creates a new materialization connector for writing Parquet files to Azure Blob storage, based on the `filesink` framework and our general blob storage package.

Future work may consider refactoring the S3 and GCS implementations in `filesink` to also use `bucket` from our blob storage package, and additional simplifications to `filesink` that would result.

**Workflow steps:**

Will need new docs for this connector

**Notes for reviewers:**


Alex-Bair (approved):
LGTM

PR #2845, opened 2025-05-23
source-jira-native: initial native connector

Alex-Bair:
**Description:**

This PR introduces a minimal native Jira connector. To deliver the connector sooner, only a subset of the possible streams (4 out of 50+) have been implemented. I anticipate a significant amount of development remains to add the other streams we care to add. There are multiple other types of streams, like child streams, child streams of child streams (grandchild streams?), streams that use the Jira Software API instead of the Jira Platform API, etc.

`source-jira-native` used to be an imported Airbyte connector. That imported connector was renamed to `source-jira-legacy` in https://github.com/estuary/connectors/pull/2824 to free up the `-native` suffix for the actually native connector. The imported `source-jira-native` was v2, and the actually native connector increments the `source-jira-native` version to v3.

**Workflow steps:**

Documentation will need created for the new `source-jira-native`. The docs for `source-jira-legacy` should also be renamed & updated for that new name. Short links should also be updated. I'll do all of that when I tackle making the docs.

**Notes for reviewers:**

Tested on a local stack. Confirmed all streams capture the expected number of documents based off of API responses.


source-jira-native/source_jira_native/api.py
```diff
@@ -0,0 +1,332 @@
...
+    lower_bound_jql = f"updated >= '{lower_bound}'"
+    upper_bound_jql = f"updated <= '{upper_bound}'"
+
+    # If users provided specific projects, only request issues from those projects.
+    # Otherwise, omitting the `project in (p1,p2,...)` is the same as requesting
+    # issues for all projects.
+    if projects:
+        projects_jql = f"project in ({projects})"
+
+        jql = " AND ".join([projects_jql, lower_bound_jql, upper_bound_jql])
```
JustinASmith, line 177: User-supplied projects are passed directly into JQL. Should there be validation that project IDs are well-formed? (Could result in JQL errors if not.)

source-jira-native/source_jira_native/api.py
```diff
@@ -0,0 +1,332 @@
...
+
+def _build_jql(
+    start: datetime,
+    end: datetime,
+    projects: str | None,
+) -> str:
+    # There are a couple quirks with filtering issues by their updated field.
+    # - Top level timestamps are in the system default user timezone, not UTC.
+    # - Jira filters issues with minute-level granularity.
+    JQL_DATETIME_FORMAT = "%Y-%m-%d %H:%M"
```
JustinASmith, line 163: If updates occur within a minute, could a record be missed or duplicated between fetches?

source-jira-native/source_jira_native/api.py
```diff
@@ -0,0 +1,332 @@
...
+    ):
+        # Checkpoint previously yielded documents if the
+        # current document moves forward in time.
+        if (
+            last_seen > log_cursor
+            and issue.fields.updated > last_seen
+            and count >= MIN_CHECKPOINT_INTERVAL
+        ):
+            yield last_seen
+            count = 0
```
JustinASmith, line 273: If capturing large volumes, could checkpointing be too infrequent? Should it be time as well as count-based? I am thinking "it depends", but most likely Jira data volume is likely not that significant even for bigger tenants?

source-jira-native/source_jira_native/models.py
```diff
@@ -0,0 +1,145 @@
...
+    return dt
+
+
+class EndpointConfig(BaseModel):
+    domain: str = Field(
+        description="This is the domain for your Jira account. Ex: estuary.atlassian.net, estuary.jira.com, jira.your-domain.com.",
+        title="Domain",
+        pattern=DOMAIN_REGEX,
+        json_schema_extra={"order": 0}
+    )
```
JustinASmith, line 52: Nit pick, but I am wondering if this REGEX could be too restrictive. I've seen Jira hosted in such a way that required specifying a Port in the domain.. I doubt we encounter that case and if we do maybe this REGEX handles it or we just require the user has their DNS set up in a way that specifying a port is not needed to reach Jira.

source-jira-native/source_jira_native/models.py
```diff
@@ -0,0 +1,145 @@
...
+        title="Authentication",
+        json_schema_extra={"order": 2}
+    )
+
+    class Advanced(BaseModel):
+        projects: str = Field(
+            description="Comma-separated list of project IDs from which to replicate issues. If left blank, issues from all projects wil be replicated.",
+            title="Projects",
+            default="",
+        )
```
JustinASmith, line 71: projects field is a comma-separated string, but passed as-is to JQL. What if users include spaces, or invalid project keys? Would it be worth adding validation?

source-jira-native/source_jira_native/resources.py
```diff
@@ -0,0 +1,171 @@
...
+        ),
+        schema_inference=True,
+    )
+
+
+async def all_resources(
+    log: Logger, http: HTTPMixin, config: EndpointConfig, should_fetch_timezone: bool = True
+) -> list[common.Resource]:
+    http.token_source = TokenSource(oauth_spec=None, credentials=config.credentials)
+    timezone = await fetch_timezone(http, config.domain, log) if should_fetch_timezone else ZoneInfo("UTC")
```
JustinASmith, line 164: If fetch_timezone fails, falls back to UTC. Is this safe? Could it cause incorrect JQL filtering if user's data is not in UTC? What if a user changes timezones they right after we get this information at connector start?

JustinASmith (commented):
Looks good! I just have a few questions/comments.

source-jira-native/source_jira_native/resources.py
Alex-Bair: If `fetch_timezone` fails, the connector crashes. The connector doesn't try to fetch the timezone from Jira during discovery, but it does as part of `validate` & `open`.  I mainly did this so we'd be able to have a discover snapshot even though we don't have credentials for a dev Jira account.

source-jira-native/source_jira_native/models.py
Alex-Bair: Yep, good point. I'll add some basic validation to confirm it's either an empty string or strings separated by commas.

source-jira-native/source_jira_native/models.py
Alex-Bair: Hmm, interesting. I was pattern matching against the domains for existing captures, but none of them have a port specified. I wouldn't hurt to allow a port since it's possible. The `DOMAIN_REGEX` doesn't handle an optional port, so I'll update it.

JustinASmith (approved):
LGTM % questions/comments - disregard anything irrelevant or silly. Overall I think everything checks out and is solid!

source-jira-native/source_jira_native/api.py
Alex-Bair: Yeah, I think there should be some kind of validation on the values provided in the `projects` input. I'm planning to validate by fetching all of the projects in the account, then raising an error if the user provided a project id/key that's not in the list of projects we receive from Jira's API. I'll add that in real quick.

source-jira-native/source_jira_native/api.py
Alex-Bair: What do you mean by time based? I'm not following.

As it is now, the current checkpointing strategy is:
- If we've yielded at least `MIN_CHECKPOINT_INTERVAL` documents since the last checkpoint & the next document was updated at a later time than those previous documents, checkpoint the previous documents.
- Or if we've paginated through all pages & there was at least one result, yield a final checkpoint.

We'll do that on every sweep, which is determined by the resource `interval` (default is 5 minutes).

source-jira-native/source_jira_native/api.py
Alex-Bair: Added validation that checks if the provided projects are actually projects that exist in the user's Jira account.

source-jira-native/source_jira_native/models.py
Alex-Bair: I think it's worthwhile to be a little more flexible here, especially since it's pretty simple to expand the regex to allow an optional port. That regex has been updated.

PR #2847, opened 2025-05-27
materialize-snowflake: create tables with no text column collation

williamhbaker:
**Description:**

We need to explicitly not use any collation for text columns on tables we create, even if there is a default collation set for the database or account.

Flow itself uses a binary comparison for string fields used as collection keys, so it is necessary that text columns not use collation for comparisons that are done during load and merge queries. Otherwise, rare issues can occur - one recent example is where values of a string key field sometimes had a trailing zero-width space (U+200B) character but were otherwise identical. Flow considers these values to be distinct, but Snowflake with certain collations evaluates them to be equal.

A further reason for not using collation is that Snowpipe streaming will not work with any table containing any column with a collation set.

It's not very well documented, but it is possible to override a default collation set on the database or account by creating the table with a `DEFAULT_DDL_COLLATION = ''` clause. Although using an explicit 'BINARY' would correctly evaluate equality for string keys, it still breaks Snowpipe streaming. Using the empty string for the `DEFAULT_DDL_COLLATION` has the desired effect of ending up with "none" for the column collations.

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM

PR #2859, opened 2025-05-28
materialize-snowflake: quote schema names when listing tables

williamhbaker:
**Description:**

Schema names need to be quoted if they contain special characters in SHOW TABLES IN <schema> queries.

In general there is no reason to not simply quote them all the time, which is what this commit does for these SHOW queries.

Tested manually by running Validate with schema names containing hyphens. Previously these would not work and result in a syntax error, but now they will work.

**Workflow steps:**

**Notes for reviewers:**


jgraettinger (approved):
LGTM

PR #2860, opened 2025-05-28
source-oracle: calculate rowid ranges for keyless backfill

mdibaiee:
**Description:**

1. Snapshot churn is due to two factors:
    1. Pending Transactions are not deterministic, since it checks for transactions across the whole database, and at times transactions will be run by the database itself as part of routine processes. I added sanitization to sanitize Pending Transactions away to make the files more deterministic
    2. the LOCAL timestamp types are not deterministic and change based on the local time zone. I actually don't know what is the best way to handle these, what's the prior art for local time zones when capturing databases? IIRC we have a configuration knob which allows customers to specify the session timezone used by the connectors? I can implement that as a next step
2. Updated the keyless backfill path to first create ranges of ROWIDs to be captured and then capture them based on these ranges in chunks of 256mb.
3. Improve performance by not decoding numbers from strings and always output them as strings
4. Fix edge cases where `UNDO_SQL` was not available for some queries
5. Fix edge case where we were setting certain values to explicit `null` in `_meta/before` due to them being absent in Oracle logs
6. Update go_ora library to latest which resolves an issue we had with an internal timeout in the library being 2 minutes and causing timeouts when database was under heavy load

Tested this on a 4GB table to ensure the 256mb bucketing is working correctly and it worked as expected. Tested with a smaller bucketing value (16mb) to cover edge cases such as empty ranges, etc.

I will test this first on a customer who has very large indexless / keyless tables before merging.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2860)


source-oracle/backfill.go
```diff
@@ -183,6 +208,122 @@ func (db *oracleDatabase) WatermarksTable() sqlcapture.StreamID {
...
+        and decode(e.partition_name, o.subobject_name, 0, 1) = 0
+      where e.segment_type in ('TABLE', 'TABLE PARTITION', 'TABLE SUBPARTITION')
+        and e.owner = '%s'
+        and e.segment_name = '%s'
+    )
+    , extents_with_sums as (
+      select sum(blocks) over() total_blocks,
+        sum(blocks) over(order by data_object_id, file_id, block_id) - blocks cumul_prev_blocks,
+        sum(blocks) over(order by data_object_id, file_id, block_id) cumul_current_blocks,
+        greatest(1, FLOOR((sum(blocks) over())/(SELECT (1024*1024*1024) / TO_NUMBER(value) FROM v$parameter WHERE name='db_block_size'))) num_buckets,
```
mdibaiee, line 237: addition here from me to calculate number of buckets based on how many 1GB buckets would be required to cover data of the table

source-oracle/backfill.go
```diff
@@ -240,21 +381,34 @@ func castColumn(col sqlcapture.ColumnInfo) string {
...
+func (db *oracleDatabase) keylessScanQuery(info *sqlcapture.DiscoveryInfo, schemaName, tableName, afterRowID string) (string, string) {
 	var query = new(strings.Builder)
 	var columnSelect []string
 	for _, col := range info.Columns {
 		columnSelect = append(columnSelect, castColumn(col))
 	}
 
+	var streamID = sqlcapture.JoinStreamID(schemaName, tableName)
+
+	var idx = slices.Index(db.backfillRowIDRanges[streamID], afterRowID)
```
willdonnelly, line 395: Can the endpoints of a range change across task restarts? If I'm reading this correctly the `afterRowID` value is coming from the persisted resume cursor when the task restarts, but the ranges are fetched anew, so it seems like it might not always be guaranteed that the persisted `afterRowID` value will be present in the new ranges list.

source-oracle/backfill.go
```diff
@@ -183,6 +208,122 @@ func (db *oracleDatabase) WatermarksTable() sqlcapture.StreamID {
 	return sqlcapture.JoinStreamID(bits[0], bits[1])
 }
 
+func (db *oracleDatabase) fetchBackfillRowIDRange(ctx context.Context, schemaName string, tableName string) error {
```
willdonnelly, line 213: Instead of passing in the schema/table name strings separately, this could just be the stream ID. 
(I recently got around to refactoring things so that it's a `struct{Schema, Table}` so you can use `streamID.Schema` and `streamID.Table` as needed)

It's also fine if you want to leave it like this though, there's plenty of other places we're still threading the two values around separately.

source-oracle/backfill.go
```diff
@@ -132,13 +142,28 @@ func (db *oracleDatabase) ScanTableChunk(ctx context.Context, info *sqlcapture.D
 		}
 		resultRows++
 		rowOffset++
+
+		// Keyless backfills are not ordered, so we need to check every row to see if it includes the last ROWID known
+		// at the time of starting the backfill which is the last rowid in the backfill range slice
+		if state.Mode == sqlcapture.TableStateKeylessBackfill && !backfillComplete {
+			backfillComplete = rowid >= lastTableRowID
```
willdonnelly: What happens here if the user starts a backfill and then drops the last row in the table?

source-oracle/backfill.go
willdonnelly, line 395: nit: I don't think there's any actual problem in practice, but it might be cleaner to pass in `db.backfillRowIDRanges[streamID]` as an argument to the `keylessScanQuery` function, just to make the data/control flow clearer.

willdonnelly (approved):
LGTM for the most part. A couple of minor nits and two more significant questions about edge cases:

- What happens if the user starts a backfill (so we establish a particular `lastRowID`) and then deletes the last row of the table?
- Is it possible for substantial changes to a table to change the range boundaries, such that after a task restart the persisted cursor is no longer exactly present in the list?

willdonnelly:
One other note: does the DBA_EXTENTS table require additional privileges? If so it might be good to add a prerequisite check for that, although ideally we'd want to only run it when there's at least one keyless table enabled.

source-oracle/backfill.go
```diff
@@ -4,13 +4,16 @@ import (
...
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 
 	"github.com/estuary/connectors/sqlcapture"
 	"github.com/sirupsen/logrus"
 )
 
+const StartingRowID = "AAAAAAAAAAAAAAAAAA"
```
willdonnelly, line 15: Is this used anywhere?

source-oracle/backfill.go
mdibaiee: I already changed this logic while you were reviewing this, I think the new logic is fine since we now just check whether we were capturing the last range, and if so, we mark `backfillComplete = true` by comparing `rowidRangeEnd` with `lastTableRowID`

source-oracle/backfill.go
mdibaiee: I tried this but ran into lowercasing, I tried to prevent that by setting `sqlcapture.LowercasingStreamID` to false, it broke tests so I just gave up for now, we can revisit later

source-oracle/base64.go
```diff
@@ -0,0 +1,37 @@
...
+func base64SortKey(s string) []int {
+	s = strings.TrimRight(s, "=") // Remove padding
+	key := make([]int, len(s))
+	for i, ch := range s {
+		key[i] = base64Index[ch]
+	}
+	return key
+}
+
+func base64LessThan(a, b string) bool {
```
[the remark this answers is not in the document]
mdibaiee: I did some search on this before implementing and apparently base64 does not preserve ordering of the unencoded value when encoded, and since we are relying on base64 string representation and its ordering (that's what happens in the bucketing logic), I implemented a base64 ordering which is based on the base64 string itself and not its underlying data. https://stackoverflow.com/questions/53301280/does-base64-encoding-preserve-alphabetical-ordering

mdibaiee:
> One other note: does the DBA_EXTENTS table require additional privileges? If so it might be good to add a prerequisite check for that, although ideally we'd want to only run it when there's at least one keyless table enabled.

I didn't need to grant additional permissions to the C##FLOW_LOGMINER_TEST user which runs the tests, and that user was set up using our documentation so I suppose no

source-oracle/base64.go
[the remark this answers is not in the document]
willdonnelly: Oh, I see. I didn't read closely enough and assumed this logic was meant to correct for the fact that base64 doesn't preserve lexicographic ordering of the input bytes. And I assume that Oracle uses some batshit insane collation on the ROWID strings which doesn't actually match up with the ASCII/UTF-8 bytewise ordering?

In that case I'm still concerned but for different reasons: given that these ROWID values represent a physical row locator in the database, shouldn't we be sorting them according to the bytes that they represent rather than whatever insane ordering Oracle uses? Sorting the encoded strings makes very little sense in this context and seems like it's probably subtly incorrect in some way.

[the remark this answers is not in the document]
mdibaiee: @willdonnelly base64 generally does not have the same ordering as ascii/utf-8 bytewise ordering, see https://en.wikipedia.org/wiki/Base64#Base64_table_from_RFC_4648 and Oracle uses the ordering as specified in the table, which is the same as I have implemented. I initially tried sorting the strings bytewise but that turned out to be out of order with Oracle

That's a good point, if we order based on their actual physical location, we are likely to have better throughput. With the current implementation of using ROWIDs ordered based on their base64 ordering, we are basically scanning a larger section of the table to read a small part of the data... 🤔 

I will look into this to update both the query which chunks the ROWID space and this comparison to use the underlying bytes which represent the physical location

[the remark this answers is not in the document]
willdonnelly: > base64 generally does not have the same ordering as ascii/utf-8 bytewise ordering

Yeah, I'm aware that the base64 alphabet is not in the same order as ASCII, I'm just surprised that if you ask Oracle to `ORDER BY` a base64 rowid string it orders them according to the base64 alphabet and not the ASCII character values.

mdibaiee:
Cleaned up the pull-request commits and did a review myself, up for a final review before merging now

source-oracle/discovery.go
```diff
@@ -317,61 +258,39 @@ func getColumns(ctx context.Context, conn *sql.DB, tables []*sqlcapture.Discover
...
-			t = reflect.TypeFor[string]()
 			jsonType = "string"
 			format = "date-time"
 		} else if dataType == "DATE" || strings.Contains(dataType, "TIMESTAMP") {
-			t = reflect.TypeFor[string]()
 			jsonType = "string"
 		} else if strings.Contains(dataType, "INTERVAL") {
-			t = reflect.TypeFor[string]()
-			jsonType = "string"
-		} else if slices.Contains([]string{"CLOB", "RAW"}, dataType) {
```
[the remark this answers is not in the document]
mdibaiee: It is on purpose. Basically I removed the `translateRecord` layer to improve performance but also because it was not necessary for the majority of cases. In case of `CLOB`, first of all, we cannot capture it easily using Logminer, it requires us to do some work to capture `CLOB`s. Regarding `RAW`, with the removal of `translateRecord` we are not casting it to `[]byte` and then perhaps base64 it.

`RAW` can be supported, but I haven't seen use cases of it yet, and I would like to add it back when we have a use case and we know how we want to handle it.

PR #2868, opened 2025-05-29
source-jira-native: use incremental json processor for `issues`

Alex-Bair:
**Description:**

We've already observed a single issues response causing the connector to run out of memory and crash. Switching to the `IncrementalJSONProcessor` avoids these OOMs.

I originally didn't use the `IncrementalJSONProcessor` because I anticipated adding streams that are children of issues — i.e., we fetch an issue, then fetch some subresource using that issue's ID. In other connectors with similar child streams where the parent used the `IncrementalJSONProcessor`, reading the HTTP response for the parent stream would time out due to how long it takes to fetch child resources for each parent ID.

I was cautiously optimistic that a single `issues` response wouldn't consume too much memory and that we wouldn't need to worry about parent streams' HTTP responses timing out when adding `issues` child streams, but I was clearly too optimistic. I'll tackle that complexity later once those `issues` child streams are added.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed the connector no longer OOMs when processing a large `issues` response.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2868)

### DIFF
--- a/source-jira-native/source_jira_native/api.py
+++ b/source-jira-native/source_jira_native/api.py
@@ -5,6 +5,7 @@ from zoneinfo import ZoneInfo
 
 from estuary_cdk.capture.common import LogCursor, PageCursor
 from estuary_cdk.http import HTTPSession
+from estuary_cdk.incremental_json_processor import IncrementalJsonProcessor
 from pydantic import TypeAdapter
 
 from .models import (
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM!

PR #2869, opened 2025-05-29
source-jira-native: remove `editmeta` from the `expand` param for `issues`

Alex-Bair:
**Description:**

Including `editmeta` when searching for issues can cause repeat 500 internal server errors on Jira's side with a message like `Cannot invoke \"com.atlassian.jira.issue.fields.config.FieldConfig.getScope()\" because \"config\" is null`.

Omitting `editmeta` avoids these 500 responses. The imported `source-jira-legacy` doesn't include `editmeta`, so any existing users likely won't expect us to fetch it. If they do request this, we can add some config options to let users control what's included in the `expand` param.

**Workflow steps:**

**Notes for reviewers:**

Checked by hitting the API directly. A request that previously returned the 500 error mentioned above with `editmeta` included in the `expand` param no longer caused an error after removing `editmeta`.

### DIFF
--- a/source-jira-native/source_jira_native/api.py
+++ b/source-jira-native/source_jira_native/api.py
@@ -215,7 +215,7 @@ async def _fetch_issues_between(
         params["fields"] = "id,updated"
     else:
         params["fields"] = "*all"
-        params["expand"] = "renderedFields,transitions,operations,editmeta,changelog"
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM!

PR #2875, opened 2025-05-30
source-oracle-batch: always discover numerics as string with format

mdibaiee:
**Description:**

- The logic to actually discover the specific cases which can fit into integer is quite tricky, and subsequently we need to translate the values from string to these numeric types, which given the current structure of BatchSQLDriver is not straightforward.
- Now that oracle batch uses schema inference, I think it is fine for us to capture all numeric types as string (as they come from the Oracle driver) with format, and let schema inference do its thing when it comes to materializing the values
- I think it makes sense to do the same thing for source-oracle (cdc) as currently a good bit of computing time is spent on converting strings to numeric types there, which might not be worthwhile after all
- We have a single customer using this connector right now and they will not be affected negatively by this change, but actually will be unblocked, so no need to worry about backward compatibility with existing tasks yet

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2875)


willdonnelly (approved):
LGTM. If I'm reading this correctly, the previous discovered types were simply incorrect as a description of what the capture would output for these columns and this brings them into alignment, right?

mdibaiee:
Yep as you both said, basically go-ora used to return integers and floats for some numeric fields, but they ended up giving up on doing so and instead return string for all numeric types now, see: https://github.com/sijms/go-ora/issues/544

and I think given schema inference, this is okay for us and makes our lives easier

PR #2876, opened 2025-05-30
materialize-google-sheets-native: allow frozen header values that are numbers

williamhbaker:
**Description:**

A frozen header cell may contain a number. Previously we weren't using this cell's value as the field name, but it could be used after casting it to a string.

Booleans are still not used as header values. I don't know for sure how you end up with a typed boolean in a Google sheet, but if you somehow did it doesn't seem like that would be very useful as a field name and you'd be better off with the A,B,C... style header for the column.

Admittedly numbers as headers & field names don't make the most sense either, but it at least seems somewhat plausible that a sheet would have something like this and a user would want to use the number as the field name.

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM % super minor nit - the commit subject has `materialize-` instead of `source-`. Not uber important but could help with searchability later.

williamhbaker:
yikes, thanks - fixed

PR #2882, opened 2025-06-02
source-*-batch: Detect and error out if a cursor column is missing in query results

willdonnelly:
**Description:**

Detects and errors out if one of the configured cursor properties is not present in the results of a polling query. However, as demonstrated in the new tests/snapshots, this can only be detected after the query successfully executes and starts returning results, so _usually_ it should never happen.

In fact I'm still trying to figure out whether/why it's happening in BigQuery at all here. But evidently it can in at least some edge case, so it seems prudent to add the appropriate sanity checking to all the batch SQL connectors, since checking properly is a one-time cost paid only during the setup at the start of a polling operation anyway.

### DIFF
--- a/source-bigquery-batch/.snapshots/TestNonexistentCursor-Capture
+++ b/source-bigquery-batch/.snapshots/TestNonexistentCursor-Capture
@@ -1,0 +1,9 @@
+# ================================
+# Final State Checkpoint
+# ================================
+
+# ================================
+# Captures Terminated With Errors
+# ================================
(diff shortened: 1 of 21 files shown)


willdonnelly:
Followup: this was happening because the binding in question used a custom query template. The _query_ was currect but the _cursor setting_ was subtly wrong (trailing whitespace, to be precise). The fix contained in this PR would have caused us to catch that sooner, and more importantly we'd have errored out _before_ emitting a state checkpoint with an unusable cursor value.

mdibaiee (approved):

PR #2887, opened 2025-06-03
source-postgres: Remove CTID ordering sanity check

willdonnelly:
**Description:**

In https://github.com/estuary/connectors/pull/2878 the CTID backfill logic was modified so that instead of a LIMIT clause plus ascending-TID-order assumption we'd use an explicit upper bound on each backfill query.

That means this ordering check is no longer required, and removing it was actually the entire point of the previous PR, I just forgot to actually delete the check. Whoops.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2887)


dyaffe (approved):
LGTM

PR #2894, opened 2025-06-05
source-google-ads: add logging to debug hanging captures & add a request timeout

Alex-Bair:
**Description:**

We've observed several hanging `source-google-ads` captures in production. I've tried recreating a hanging capture locally, but I haven't been successful yet. I suspect it's an intermittent issue that would require running the capture for hours to replicate. To speed up debugging, I'm adding a ton of logging to identify where the connectors are getting stuck when it happens in production again. We can revert the logging commit later once the problem is addressed.

My leading suspicion is that the Google API is silently dropping connections, and the connector's stuck waiting on a response that will never arrive. So, I also wired in a rough timeout to catch this if/when it happens. It's set at 5 minutes right now (which seems like a pretty generous duration).

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack & confirmed the connector runs like it did before. I did observe some `TimeoutErrors` get raised, so _hopefully_ this mitigates the "hanging connector" issue🤞 .

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2894)

### DIFF
--- a/source-google-ads/source_google_ads/google_ads.py
+++ b/source-google-ads/source_google_ads/google_ads.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
-
+from concurrent.futures import ThreadPoolExecutor, TimeoutError as FutureTimeout
 import logging
 from enum import Enum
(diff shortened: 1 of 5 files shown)


williamhbaker (approved):
LGTM

PR #2896, opened 2025-06-05
estuary-cdk: prevent `RateLimiter.delay` from growing larger than 5 minutes

Alex-Bair:
**Description:**

We've seen infrequent cases where a connector is rate limited for an extended period of time (often when another ingestion tool is contributing to API rate limits), causing the `RateLimiter.delay` to grow to an absurd value. In one case, it grew to 43 minutes!

This causes multiple problems:
- With the low delay decay rate, it would take hours/days to reduce the delay to a reasonable amount (ex: if the delay grew to 43 minutes, the delay would only decay to 39 minutes after 1,000 successful requests. And keep in mind that we'd be waiting 39+ minutes between successful requests in most streams too).
- For streams that make multiple HTTP requests in a `fetch_foobar` invocation, those streams could spend multiple hours/days before yielding control back to the CDK & checking if the `stopping` event is set. This prevents the 24 hour graceful restart interval we want to maintain.

This PR takes a simple approach to solving these problems: prevent the delay from growing larger than 5 minutes. There are more sophisticated approaches that might work a little better (use a more aggressive decay, allow the delay to grow unbounded but reduce it immediately to 5 minutes upon a successful request), but I'd like to see if this simpler approach is sufficient before spending more time on this.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/estuary-cdk/estuary_cdk/http.py
+++ b/estuary-cdk/estuary_cdk/http.py
@@ -308,9 +308,12 @@ class RateLimiter:
     It initially uses quadratic decrease of `delay` until a first failure is
     encountered. Additional failures result in quadratic increase, while
     successes apply a linear decay.
+
+    To avoid excessively long delays, `delay` cannot grow larger than `MAX_DELAY`.
     """
 
     delay: float = 1.0
+    MAX_DELAY: float = 300.0 # 5 minutes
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #2897, opened 2025-06-06
source-jira-native: add remaining streams & disable most by default

Alex-Bair:
**Description:**

This PR adds multiple more streams to `source-jira-native`. There are a lot of additions, and I'd recommend reviewing the individual commits in order to help break it up into more readable chunks.

Some interesting notes:
- Jira has multiple different APIs (Platform, Software, Service Desk). Authentication is handled the same way for each. If you squint, it may look like the endpoints among them behave similarly, but there are subtle differences (using `start` vs `startAt` for pagination, returning results in a generic `values` field vs. a more uniquely named field like `dashboards`, etc.).
- There are multiple child streams (where we fetch the ids of some resource, then get other resources for each of those ids). These can rack up quite a few API requests. To mitigate hitting rate limits:
  - Most streams are disabled by default. We can change these defaults later if users seem to always want some stream enabled, but hopefully this reduces hitting rate limits on the "garden path".
  - The incremental `IssueChildStream`s attempt to sideload the child resources alongside each issue when fetching the parent issue ids. If all child resources for an issue are sideloaded, the connector yields them without making additional HTTP requests. But if there are more child resources for an issue than can be sideloaded, the connector makes additional HTTP requests to fetch all child resources for that issue.

**Workflow steps:**

The connector documentation should be updated to include the new streams.

**Notes for reviewers:**

- Each stream works & doesn't encounter errors.
- Most streams are disabled by default when setting up a new capture.

The imported `source-jira-legacy` has some streams I chose not to add to `source-jira-native` yet, like `issue_watchers`, `issue_properties`, `board_issues`, etc. I chose not to add these because they either didn't capture any useful data for existing captures or the data is already captured by other streams in some other way. If users truly want full stream parity with `source-jira-legacy`, we can add those streams in later.

To hopefully help during the review process, I kept the full commit history showing the smaller changes I made to add these streams. But I think it's a bit messy to have all those commits to the `main` branch, so I'll squash those commits into a single commit before merging this PR.


source-jira-native/source_jira_native/resources.py
```diff
@@ -185,9 +301,74 @@ def open(
...
+                config.domain,
+                timezone,
+                config.advanced.projects,
+                stream,
+            )
+        )
+
+    # Shift the cutoff back ISSUE_JQL_SEARCH_LAG duration to ensure backfills
+    # always cover ranges where Jira's API returns consistent results.
+    cutoff = datetime.now(tz=timezone) - ISSUE_JQL_SEARCH_LAG
```
JustinASmith, line 345: I  may be misunderstanding, but is this the same thing as `eventual consistency`? If so, would it be worth considering a `lookback` and `realtime` resource binding similar to `source-sage-intacct`?

JustinASmith: There are many streams already, so maybe this isn't as feasible for this connector anyway.

JustinASmith (approved):
LGTM % a couple of comments - this is somewhat a quick pass TBH. I am not that familiar with Jira's APIs as you are at this point. The decisions you have made here seem sound and consistent!

source-jira-native/source_jira_native/resources.py
JustinASmith: Thinking about this again... The API rate limit being a concern would make this `lookback`/`realtime` approach not that great since it would cause more requests to be fired off.

JustinASmith: Gotcha! Yep, that sounds good. I think that's a good plan

PR #2899, opened 2025-06-06
source-snowflake: Support captures from Snowflake Data Sharing

willdonnelly:
**Description:**

Some connector tweaks and additional advanced configuration options to enable capturing from shared tables via Snowflake Data Sharing.

**Workflow steps:**

1. Ensure that the source table has `CHANGE_TRACKING` enabled with an appropriate retention period as described [here](https://docs.snowflake.com/en/user-guide/data-sharing-provider#streams-on-shared-objects). This may require reaching out to the publisher of the shared table(s) if necessary.
2. When creating the capture task, in the capture configuration:
    a. Set `/database` to the shared DB that you want to capture from
    b. Set `/advanced/flowDatabase` to a DB you own which will hold change streams and staging tables
    c. Set `/advanced/fullCopySnapshots` to true

We should update https://docs.estuary.dev/reference/Connectors/capture-connectors/snowflake/ with a new section describing how to make CDC from data shares work.


willdonnelly:
> we're not changing the resource path, and it continues to be only [schema, table]. This means we're unable to represent a capture across multiple DBs of the same [schema, table]

That's correct, the resource path is unchanged by this work and continues to operate as it always has, so for now we still can only capture from a single Snowflake DB at a time.

In theory this PR gets us about halfway to supporting captures across multiple Snowflake DBs, the remaining work would basically just be:
1. Figuring out how we want to express the endpoint configuration for multiple DBs.
2. Extending the discovery logic to be able to discover across DBs.
3. Extending the resource configs with a DB property, probably making it optional IFF there is exactly one capture DB in the endpoint config so it's backwards compatible.

PR #2901, opened 2025-06-07
materialize-kafka: implement explicit delivery confirmation with fail-fast semantics

jgraettinger:
Replace producer.flush() with custom AckTrackingContext that provides:
- Explicit delivery confirmation via ProducerContext callbacks
- Fail-fast on first delivery failure with proper error propagation
- Enhanced producer config (acks=-1, enable.idempotence=true)

This eliminates silent failures during broker instability by ensuring
StartCommit only proceeds after confirmed delivery of ALL messages.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2901)

### DIFF
--- a/materialize-kafka/src/configuration.rs
+++ b/materialize-kafka/src/configuration.rs
@@ -1,12 +1,17 @@
 use anyhow::Result;
 use rdkafka::{
     admin::AdminClient,
-    client::DefaultClientContext,
-    producer::{DefaultProducerContext, ThreadedProducer},
+    client::{ClientContext, DefaultClientContext},
+    error::KafkaError,
+    producer::{DeliveryResult, ProducerContext, ThreadedProducer},
     ClientConfig,
 };
 use schemars::{schema::RootSchema, JsonSchema};
 use serde::{Deserialize, Serialize};
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Condvar, Mutex,
+};
 
 #[derive(Serialize, Deserialize)]
 pub struct EndpointConfig {
@@ -14,6 +19,8 @@ pub struct EndpointConfig {
     pub credentials: Option<Credentials>,
     pub tls: Option<TlsSettings>,
     pub message_format: MessageFormat,
+    #[serde(default)]
+    pub compression: Compression,
     pub schema_registry: Option<SchemaRegistryConfig>,
     pub topic_partitions: i32,
     pub topic_replication_factor: i32,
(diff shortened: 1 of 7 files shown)


materialize-kafka/src/configuration.rs
```diff
@@ -220,10 +262,12 @@ impl EndpointConfig {
         Ok(())
     }
 
-    pub fn to_producer(&self) -> Result<ThreadedProducer<DefaultProducerContext>> {
+    pub fn to_producer(&self) -> Result<ThreadedProducer<AckTrackingContext>> {
         let mut config = self.common_config()?;
-        config.set("compression.type", "lz4");
-        Ok(config.create()?)
+        config.set("compression.type", self.compression.to_string());
+        config.set("request.required.acks", "-1"); // Wait for all in-sync replicas
```
williamhbaker, line 268: The `request.required.acks` configuration is `-1` (aka `"all"`) by default AFAIK, but I guess `enable.idempotence` is _not_ default for `librdkafka` per its [docs](https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md), so setting `enable.idempotence` makes sense here.

The `// Ensure exactly-once semantics` comment is misleading though IMO since overall the connector is still at-least-once, and is won't maintain that idempotency state across connector restarts. I think we should modify / remove the comment about exactly-once (maybe: `Ensure message ordering and minimize duplication` ?). The `request.required.acks` explicit configuration doesn't seem necessary to me since it is the default but I guess it's not a huge issue to have it here, although there are many many other configs we aren't setting and just accepting the default.

materialize-kafka/src/configuration.rs
```diff
@@ -294,6 +338,85 @@ impl JsonSchema for Resource {
...
+    state: Mutex<Option<KafkaError>>,
+    notify: Condvar,
+}
+
+impl AckTrackingContext {
+    pub fn new() -> Self {
+        Self {
+            pending_messages: AtomicUsize::new(0),
+            completed_messages: AtomicUsize::new(0),
+            has_failure: AtomicBool::new(false),
```
williamhbaker, line 354: Using this `has_failure` seems like overkill to me. The `wait_for_all_successful_acks` loop should only get one notification as I am reading the code, so it seems like it would be enough to just do the `if let Some(ref error) = *guard` check at the top of the loop? That would also alleviate a secondary question I have, which is what happens if `has_failure` is true but `*guard` is `None`. Of course this is not a possible condition but it seems more complicated than it needs to be.

williamhbaker (approved):
LGTM % a couple of comments/questions

williamhbaker:
Oh you also might want to run `cargo fmt` since there is a bunch of extra whitespace in the new code. We should probably make this a CI check, but don't have that yet.

materialize-kafka/src/configuration.rs
jgraettinger: Yep, this is left over from an earlier factoring where we'd wake the thread every message. Removed.

jgraettinger:
Will merge in the AM when I can monitor

williamhbaker:
> Will merge in the AM when I can monitor

FYI we will need to update all existing tasks to have a configuration value for the compression setting, since they are all current missing it, and it is being added as a required config. There's not that many so I figured it would be quite manageable to do this.

jgraettinger:
Tweaked this to make `compression` non-required, with LZ4 as the default.

PR #2907, opened 2025-06-09
source-bigquery-batch: Fix numeric serialization

willdonnelly:
**Description:**

Previously BigQuery NUMERIC, DECIMAL, BIGNUMERIC, and BIGDECIMAL column types were being serialized as strings like "12345/100" for "123.45" (because the values arrived to us as big.Rat values with a power-of-ten denominator, and that's the default JSON serialization for big.Rat values).

The fix is to add some logic to stringify them ourselves with appropriate decimal precision.

This is a case of our capture output not matching our own discovered write schema, so it should be fine to just fix the bug by changing the capture output. Note that we did change the write schema recently as part of the full column type discovery / SourcedSchema work so over the longer term this is an actual behavioral change to previously non-failing captures, but they'd have to have already been failing for a couple of weeks now so it's probably fine to just move on with things.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2907)


jgraettinger (approved):
LGTM

PR #2910, opened 2025-06-10
source-mixpanel-native: misc. small improvements to avoid OOMs & timeouts

Alex-Bair:
**Description:**

This add various, small improvements to `source-mixpanel-native` (which is an imported Airbyte connector, not a native connector).
- Reduce OOMs caused by Mixpanel responses returning multiple GBs of data.
  - `engage` now streams the response (like `export` already did) so the entire response is not held in memory
  - the `date_window_size` default is reduced to `1` to _hopefully_ avoid hitting the 2 hour request-response cycle timeout we have for imported Airbyte connectors. If a streaming & processing a single response with a `date_window_size` of 1 day takes longer than 2 hours, then we'll need to re-evaluate this strategy (likely bump up the 2 hour limit...)
- Remove the `revenue` stream. It's hitting an undocumented, deprecated endpoint, and the data it returns is always the same. It's pretty useless & no one has asked for it; it's a relic leftover from when the connector was originally imported. I've been recommending it be disabled whenever a capture was set up, and removing the stream entirely will avoid having to manually disable the stream each time.

**Workflow steps:**

The connector's documentation should be updated to not include `revenue` as a stream.

**Notes for reviewers:**

- the `date_window_size` setting defaults to `1`
- the `engage` stream works as before (but streams the response like the `export` stream)
- the `revenue` stream is not discovered

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2910)


JustinASmith (approved):
LGTM!

PR #2913, opened 2025-06-10
source-postgres: Disable parallel workers again

willdonnelly:
**Description:**

We used to disable parallel workers when doing a keyless (CTID) backfill because it was necessary for correctness that the DB choose to use a TID Range Scan. This is no longer necessary for correctness now that we specify an upper-bound CTID in the query and don't rely on ascending TID order within the result set, but it turns out we should still probably do it because the query optimizer makes bad choices if it decides to use parallel workers.


jgraettinger (approved):
still LGTM

PR #2921, opened 2025-06-11
materialize-motherduck: add sync schedule

williamhbaker:
**Description:**

Adds sync schedule, in the same way is it is for most of the other SQL warehouse materializations.

**Workflow steps:**

https://github.com/estuary/flow/pull/2216

**Notes for reviewers:**

Shortly after merging this, I will update existing tasks to have configured `syncFrequency` of `0s`, to maintain their current behavior of not delaying at all. Going forward, new materializations that don't have anything set will use a `30m` frequency by default if not otherwise set.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2921)


Alex-Bair (approved):
LGTM

PR #2922, opened 2025-06-11
source-google-ads: support custom GAQL streams

Alex-Bair:
**Description:**

Custom GAQL streams didn't work with Flow because no primary keys were specified for them. This PR adds a `primary_key` field for every custom GAQL stream so users can specify what the primary keys are & successfully set up custom GAQL streams.

**Workflow steps:**

Connector docs should be updated with the new custom GAQL stream config.

**Notes for reviewers:**

Tested on a local stack. Confirmed a custom GAQL stream works & runs incrementally.


williamhbaker (approved):
LGTM

PR #2936, opened 2025-06-13
source-sqlserver: Add 'Discover Only Enabled' setting

willdonnelly:
**Description:**

This is the SQL Server analogue to the PostgreSQL "Discover Only Published Tables" setting, which causes discovery to only return bindings for tables which are already CDC-enabled.

This isn't the default behavior because there is still some value to having the default user setup flow list every table in the DB and let them select the ones they want and _then_ complain that they haven't enabled CDC on them, but it's a fairly common thing to want so we should have an option for it.

We'll want to add a short section on this to the SQL Server connector docs so people realize it's an option they can select.


Alex-Bair (approved):
LGTM % the build looks to have failed?

PR #2937, opened 2025-06-13
source-monday: enhance error handling and boards Pydantic model

JustinASmith:
**Description:**

The `source-monday` connector has shown to miss boards during backfill. In testing, I found that the query to get `boards` may return nothing if there are no boards to get. This would fail before due to the Pydantic model type being wrong. I've updated this and added a few more updates for error handling.

**Workflow steps:**

**Notes for reviewers:**

- _Tested locally - I was able to see all the boards captured in a backfill without any issues. I am not confident this fixes everything, but it seems this does fix the boards backfilll issue._

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2937)

### DIFF
--- a/source-monday/source_monday/graphql.py
+++ b/source-monday/source_monday/graphql.py
@@ -1,5 +1,6 @@
 import json
 import asyncio
+import time
 from logging import Logger
 from typing import Any, AsyncGenerator, Dict, Literal, Callable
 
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM. Just a couple general comments.

I'm not familiar with Monday.com so take this with a grain of salt, but given how many boards are missing (5k+), I'd wouldn't expect were all missed during incremental updates but rather from during the backfill. It doesn't look like any of these changes affect the `fetch_boards_page` function (unless I'm missing something), so there's probably something else going on causing the connector to miss data.

PR #2938, opened 2025-06-13
source-stripe-native: reconcile connector state

JustinASmith:
**Description:**

This update adds a check to each resource binding to ensure that all connected account IDs are present in both the incremental and backfill state dictionaries. If either state is missing for an account, it is initialized from the initial state, and a checkpoint is triggered to persist the update.

This PR also addresses a legacy issue where, after a connector restart, the incremental state for a subtask might not be initialized even though the backfill state has progressed due to not emitting checkpoints in the prior state initialization logic for connected accounts. In such cases, the subtask is reset to backfill and to ensure state/data consistency. This check is a temporary measure to reconcile state misalignments and prevent data loss or missed updates for affected subtasks. However, it does not hurt to keep this second condition in the code in place either.

**Workflow steps:**

**Notes for reviewers:**

This is not something I can easily test with our dev/test Stripe account. However, I verified this works as it usually has with `flowctl preview` and in a local stack.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2938)

### DIFF
--- a/source-stripe-native/source_stripe_native/resources.py
+++ b/source-stripe-native/source_stripe_native/resources.py
@@ -25,6 +25,7 @@ from .api import (
 
 from .models import (
     Accounts,
+    ConnectorState,
     EndpointConfig,
     ListResult,
     STREAMS,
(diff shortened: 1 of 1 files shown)


source-stripe-native/source_stripe_native/resources.py
```diff
@@ -86,6 +87,56 @@ async def _fetch_platform_account_id(
...
+            task.log.info(
+                f"Backfilling {binding.stateKey} due to missaligned backfill and incremental states."
+            )
+            task.stopping.event.set()
+            task.checkpoint(ConnectorState(backfillRequests={binding.stateKey: True}))
+        elif should_checkpoint:
+            task.log.info(
+                f"Checkpointing state to ensure any new state is persisted for {binding.stateKey}."
+            )
+            task.checkpoint(state)
```
Alex-Bair, line 137: `task.checkpoint` takes a `ConnectorState` as an argument, not a `ResourceState` (which is what `state` is here). You can checkpoint the connector state like we do [here](https://github.com/estuary/connectors/blob/ba5cb5df2f80d3bfc060981d33e398459a589aad/estuary-cdk/estuary_cdk/capture/common.py#L588-L592) in the CDK.
https://github.com/estuary/connectors/blob/ba5cb5df2f80d3bfc060981d33e398459a589aad/estuary-cdk/estuary_cdk/capture/common.py#L588-L592

source-stripe-native/source_stripe_native/resources.py
```diff
@@ -86,6 +87,56 @@ async def _fetch_platform_account_id(
...
+                    f"Misalignment in backfill and incremental states for account id {account_id}. This is unexpected behavior and the connector will automatically backfill."
+                )
+                should_backfill = True
+
+        if should_backfill:
+            task.log.info(
+                f"Backfilling {binding.stateKey} due to missaligned backfill and incremental states."
+            )
+            task.stopping.event.set()
+            task.checkpoint(ConnectorState(backfillRequests={binding.stateKey: True}))
```
Alex-Bair, line 132: This would have almost the same effect as clicking the "Backfill" button in the UI, without incrementing the backfill counter in the spec. The entire binding's & all of its subtasks' states would be backfilled. Can we only backfill (i.e. re-initialize the state of) only the subtasks that have been affected by the bug? That'd reduce how much data needs backfilled since we'd only kick off backfills for those few account ids instead of every account id.

That'd also simplify the code path a little. We wouldn't need use the `backfillRequests` feature at all. We could handle "the account id is in the backfill state but not incremental state" the same way as if a new account id showed up.

Alex-Bair: Additionally, I wouldn't anticipate we'd need the `if should backfill` check after the existing couple tasks are fixed. We shouldn't run into this situation again we'll start checkpointing the state after adding any account ids. It might be helpful to add a comment indicating the backfill check is a short term thing & could be removed in the future.

Alex-Bair (commented):
Looking good! Had a couple comments before approving.

Alex-Bair (approved):
LGTM!

PR #2939, opened 2025-06-13
source-mysql: Fix MySQL enum discovery

willdonnelly:
**Description:**

The previous MySQL enum discovery logic was pretty hacky and had several edge cases, one of which a production capture recently ran into. This commit rips that decoding logic out entirely and swaps it out for reusing the same Vitess SQL parser we already use for replication DDL query handling.

There might be a more elegant way to ask it to parse an enum type, but I briefly checked and didn't see one so simply splatting the type definition into a dummy `CREATE TABLE` statement seemed like a decent enough workaround.


jgraettinger (approved):
LGTM

PR #2942, opened 2025-06-16
network-tunnel: move async wait of tunnel until after READY is read

williamhbaker:
**Description:**

If the network tunnel fails ~instantly, the current asynchronously Wait handling may end up closing the stdout of the tunnel process before the READY message is read. This could end up causing the output of the real error message to race the output of a less useful "file is already closed" output.

Moving the asynchronously Wait handler to after the READY message is read should remedy this. If the network tunnel does fail instantly, that error will be bubbled up as intended. If the network tunnel writes READY, the connector will crash if at some point in the future the tunnel process exits, as before.

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM

PR #2946, opened 2025-06-17
materialize-sql: adapt to use `Materializer`

williamhbaker:
**Description:**

Note: Strongly recommended to review commit-by-commit: The first commit is exclusively copying over the `materialize-sql` code into a new folder called `materialize-sql-v2`. The commits that follow are the real changes. You won't want to spend a lot of time looking through the full diff since a good percentage of it is just copying the code.

For some background, as part of building the `materialize-iceberg` connector a while back, I created a complete representation of everything a fully-featured materialization needs to be able to do as an interface named `Materializer` (see 8a46b48), and some tools to abstract away most of the common behavior that such a materialization is expected to do for managing materialized resources (AKA: everything but storing the data).

The intention was to eventually add this to all the other materializations, so that there was a centralized place to make changes to higher-level logic, which may include things like:
- Having a consistent set of common feature flag, for example the emergent `castToString` one that is somewhat haphazardly implemented in materializations right now
-  Updates to how the materialization protocol works, such as the planned change to make materializations always do an Acknowledge before Apply
-  Having a consistent strategy for handling on-line migrations, merge bounds optimization, etc.
 
The hardest part about making all materializations use `Materializer` is certainly getting the `materialize-sql` framework adapted to use it, so I decided to start there, and this is that.

As noted, I have structured these changes so that individual SQL materialization connectors can be ported over individually, rather than doing them all at the same time. Although I don't really anticipate any issues, this is a risk mitigation strategy so that I can start with some of the lesser used ones first.

Once all SQL materializations are adapted, I'll get rid of the old `materialize-sql` folder and rename `materialize-sql-v2` back to `materialize-sql` to complete the change.

See individual commit messages for more specifics.

Part of #2911

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2946)


Alex-Bair (approved):
LGTM based on my understanding of materializations. It looks like `materialize-iceberg`'s spec snapshot needs updated for the addition of feature flags.

PR #2948, opened 2025-06-18
source-shopify-native: conditionally discover streams that capture PII

Alex-Bair:
**Description:**

Only specific Shopify plans can access PII data that's fetched via the `customers`, `orders`, and `fulfillment_orders` streams. The connector now discovers that stream if the Shopify plan is a Plus or higher account, is a partner development account, or is not one of the known non-PII plans (Basic & Starter) AND the connector is using an access token from the user's custom Shopify app (i.e. not our public OAuth app).

Additionally, the name of each resource was moved into the `ShopifyGraphQLResource` class to reduce a little duplication & make it easier to filter out specific resources after they're instantiated.

**Workflow steps:**

Connector docs could be updated to reflect why `customers` & friends may not be discovered for some users & we could link to Shopify's [documentation](https://help.shopify.com/en/manual/apps/app-types/custom-apps) for the finer details.

**Notes for reviewers:**

Confirmed that when authenticating with an `AccessToken` and a "Basic" Shopify plan, `customers` & friends are not discovered. When authenticating with an `AccessToken` and a non-"Basic" or "Starter" plan, `customers` & friends are discovered.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2948)

### DIFF
--- a/source-shopify-native/source_shopify_native/graphql/abandoned_checkouts.py
+++ b/source-shopify-native/source_shopify_native/graphql/abandoned_checkouts.py
@@ -8,6 +8,7 @@ from ..models import ShopifyGraphQLResource
 
 
 class AbandonedCheckouts(ShopifyGraphQLResource):
+    NAME = "abandoned_checkouts"
     QUERY = """
     id
     abandonedCheckoutUrl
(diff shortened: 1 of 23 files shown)


JustinASmith (approved):
LGTM % changes mentioned in separate slack thread.

Issue #2955, opened 2025-06-22
Postgres batch date-time out of bounds


dyaffe:
We're seeing a failure for a capture for due to the source having a date set to this: 52455-04-19T00:00:00Z.  I thought we handled large date values, but it's failing and we should handle it.

I can provide the capture name upon request.

PR #2957, opened 2025-06-23
source-postgres-batch: Normalize excessive dates/timestamps

willdonnelly:
**Description:**

We represent timestamp values as RFC3339 strings, which only allow the year to range from 0 to 9999, so we need to normalize dates outside of that range to a sentinel value.

This fixes https://github.com/estuary/connectors/issues/2955

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2957)

### DIFF
--- a/source-postgres-batch/main.go
+++ b/source-postgres-batch/main.go
@@ -249,6 +249,15 @@ func generatePostgresResource(cfg *Config, resourceName, schemaName, tableName, 
 }
 
 func translatePostgresValue(val any, databaseTypeName string) (any, error) {
+	if val, ok := val.(time.Time); ok {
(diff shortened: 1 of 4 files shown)


source-postgres-batch/main.go
```diff
@@ -249,6 +249,16 @@ func generatePostgresResource(cfg *Config, resourceName, schemaName, tableName,
 }
 
 func translatePostgresValue(val any, databaseTypeName string) (any, error) {
+	if val, ok := val.(time.Time); ok {
+		log.WithField("val", fmt.Sprintf("%#v", val)).Debug("translating Time value")
```
mdibaiee, line 253: Do you want to keep this? seems excessive even for debug level

PR #2960, opened 2025-06-23
Update NLTK to 3.9.1 to address security vulnerability

skord:
Resolves https://github.com/nltk/nltk/issues/3266 - remote code execution vulnerability in NLTK versions < 3.9.1. This addresses a high-severity dependabot alert by explicitly pinning NLTK to ^3.9.1 in both estuary-cdk and source-qualtrics dependencies.

🤖 Generated with [Claude Code](https://claude.ai/code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2960)

PR #2961, opened 2025-06-24
source-brevo: bump `nltk` to 3.9.1

Alex-Bair:
**Description:**

Bumps `nltk` for `source-brevo`, like I mentioned in https://github.com/estuary/connectors/pull/2960. 

This was more complicated than just running `poetry lock --no-update` since the `airbyte-cdk` version `source-brevo` was using didn't support `nltk` version 3.9.1. I had to:
- Update `estuary-cdk` to allow more recent `airbyte-cdk` versions & manage different import paths between versions for Airbyte CDK models.
- Update `source-brevo` to use `airbyte-cdk` version 6.6.3, which allows `nltk` version 3.9.1.
- Update `source-brevo`'s config to have dummy data so tests would pass.

It'll be really nice once we replace all these imported connectors with native ones. That will reduce how many dependencies we have to manage & resolve whenever dependabot tells us we have to update a dependency.

**Workflow steps:**

**Notes for reviewers:**

Ensured `estuary-cdk` and `source-brevo` tests passed. We don't have any active `source-brevo` captures in production, so I didn't take the time to test the connector on a local stack - if it breaks when someone makes a new capture, we can fix it then / build a native version to replace it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2961)


JustinASmith (approved):
LGTM % two major version changes for airbyte-cdk. Did you find any major breaking changes to look out for?

Alex-Bair:
The only breaking change I found was the import paths for Airbyte CDK models (handled by the `try/except` when importing the models in `shim_airbyte_cdk.py`). I didn't encounter any other changes that affect `estuary-cdk` other than that.

There may be breaking changes to `source-brevo` due to bumping `airbyte-cdk` to a newer version, but the image builds & passes the snapshot tests, so I didn't dig too much into it since the connector isn't currently being used at all. If someone sets up a `source-brevo` capture & it doesn't work, we can dig into it then. I'd probably prefer to rewrite it as a native connector instead of continuing to maintain an imported connector if we hit problems with the imported one.

PR #2966, opened 2025-06-25
source-shopify: add setuptools dependency to address path traversal v…

skord:
…ulnerability

Add setuptools>=78.1.1 dependency to address a path traversal vulnerability in PackageIndex.download that could allow arbitrary file writes.

Fixes dependabot issue 401


Alex-Bair (approved):
LGTM - interesting dependabot flagged this for an update since `setuptools` is already version `78.1.1` in `poetry.lock` 🤷.

PR #2968, opened 2025-06-25
source-klaviyo: add setuptools dependency to address path traversal v…

skord:
Add setuptools>=78.1.1 dependency to address a path traversal vulnerability in PackageIndex.download that could allow arbitrary file writes.

Fixes dependabot alert 413


Alex-Bair (commented):
`poetry.lock` needs updated with a `poetry lock` command.

PR #2971, opened 2025-06-25
estuary-cdk: introduce `IncrementalCSVProcessor`

Alex-Bair:
**Description:**

Incremental CSV processing will be required for the `source-google-play` connector I'm working on, so it makes sense to generalize the existing CSV processing code in `source-salesforce-native` and have it live in our CDK. This PR adds an `IncrementalCSVProcessor` to our CDK and updates `source-salesforce-native` to use it.

The implementation uses `aiocsv.AsyncDictReader` to incrementally read rows of a CSV from a byte stream. Different CSV dialects (delimiters, quote chars,  encodings, line terminators, etc.) are supported; connectors can specify these options in a `CSVConfig`, pass it in when instantiating an `IncrementalCSVProcessor`, and the options feed into the underlying `aiocsv`/`csv` packages for proper handling.

There are quite a few tests in `test_incremental_csv_processor.py` to demonstrate expected behavior. Claude wrote most of the tests, and I did some minor tweaking as I tidied up the `IncrementalCSVProcessor` implementation.

**Workflow steps:**

**Notes for reviewers:**

Tested the `EmailMessage` binding of `source-salesforce-native` (a notoriously large object that required CSV processing optimizations earlier this year) on a local stack after switching the connector over to the CDK's `IncrementalCSVProcessor`. The CSV processing behavior is the same as it was previously & no memory issues or crashes were observed during testing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2971)


williamhbaker:
The JSON processor is generic on a `StreamedItem`, and yields models from that. This implementation appears to be yielding dicts that would then have to be validated into a model. The asymmetry is not very satisfying - is there a way we could shore that up? Ideally IMO the CSV processor would similarly validate into models and yield them, and I'd imagine that would take some refactoring to the salesforce capture.

Alex-Bair:
That's correct, the CSV processor is yielding dicts that would have to be validated into a model. I'd like to keep that capability because it's easier to perform transformations on a dict before it's validated into a model (like we do in `source-salesforce-native` [here](https://github.com/estuary/connectors/blob/ffd556510891613253e9a09467674448c6a4e060/source-salesforce-native/source_salesforce_native/bulk_job_manager.py#L309)). If the CSV processor yielded models instead of dicts, I _could_ dump the model into a dict, transform it, then re-validate back into a model, but that feels pretty gross.

I'll try to figure out a way for the CSV processor to take an optional Pydantic model, then yield validated models if that's provided & yield dicts otherwise.

JustinASmith:
@Alex-Bair  You can always use a Pydantic `field_validators` to do that even with this always (or optionally) returning a validated/serialized Pydantic model. https://docs.pydantic.dev/latest/concepts/validators/#field-validators

JustinASmith:
`source-qualtrics`, `source-sage-intacct`, and `source-chargebee-native` use `field_validator` in some way or fashion with [source-sage-intacct](https://github.com/estuary/connectors/blob/ffd556510891613253e9a09467674448c6a4e060/source-sage-intacct/source_sage_intacct/models.py#L185) likely being the closest to what you would need to do.

williamhbaker:
FWIW I was thinking about this some more and if doing things the pydantic way is too cumbersome with needing to dynamically construct models and validators, I don't think it would be unreasonable to have the CSV processor take in callback function as a parameter that would do the value processing, before validating that into the pydantic model class to be output.

Alex-Bair:
I should be able to get it working with Pydantic. I'm close to having something worked out to switch `source-salesforce-native` over to use the `IncrementalCSVProcessor` once the processor starts yielding validated models. I need to put this on hold though as I work on a small `source-snowflake` improvement, so I'm moving this PR back to a draft in the meantime.

williamhbaker (approved):
LGTM

PR #2983, opened 2025-06-26
source-stripe-native: improve connected account processing

Alex-Bair:
**Description:**

When a user has hundreds/thousands of connected accounts, the connector can't process all of them at once without running out of memory. I suspect the memory overhead of 500,000+ subtasks (~6,500 connected accounts × 41 bindings × 2 subtasks per binding) pushes the connector over the 1 GB memory limit.

This PR implements account processing limits and automatic restarts to handle large numbers of connected Stripe accounts while avoiding memory issues.

- Process max 100 accounts per run using intelligent prioritization:
  1. Accounts with incomplete backfills (ensures data completeness)
  2. Accounts with oldest incremental cursors (fair rotation)
- Automatic connector restarts every 30 minutes to cover all accounts
- Move incremental cursors forward when there are no updates for a connected account, ensuring the prioritization algorithm works correctly and doesn't constantly prioritize inactive accounts.

This approach assumes:
- 30 minutes is a reasonable amount of time to allow subtasks to process a good chunk / all of their backfills.
- The rate of accounts being connected to a user's primary Stripe account is less than 100 accounts / 30 minutes. If the rate is faster than that, it's possible the connector can't process accounts faster than they're being connected. Accounts that haven't completed their backfill would constantly get prioritized, accounts that have completed their backfill would never get prioritized (meaning we wouldn't fetch their incremental updates), and it'd look like we're missing data. I'm hopeful that a 100 accounts / 30 minutes rate isn't realistic & that accounts would be added slower than that, but it's a new, possible "missing data" scenario to keep in mind.

This approach isn't truly ideal; it'd be best if we could process all connected accounts simultaneously. A way to do that is to shard the capture and assign a subset of connected accounts to each shard. That'd spread the overall memory overhead across shard & avoid OOMs. However, the CDK doesn't currently support sharding captures, and that's a pretty significant effort that will need prioritized if we want to do that. In the meantime, the alternative strategy implemented in this PR should work ok.

**Workflow steps:**

**Notes for reviewers:**

- only 100 connected accounts are processed in a single connector invocation.
- connected accounts are prioritized first based on their backfill status (is it complete?) and second based on how old their incremental cursors are.
- the connector doesn't OOM, although it can still use up a lot of memory. Captures I've tested have hovered around ~61% memory usage.
- when capturing from connected accounts, the `stopping` event is set every 30 minutes & a new set of connected accounts are prioritized on the next connector invocation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2983)

### DIFF
--- a/source-stripe-native/source_stripe_native/__init__.py
+++ b/source-stripe-native/source_stripe_native/__init__.py
@@ -1,3 +1,4 @@
+import asyncio
 from logging import Logger
 from typing import Callable, Awaitable
 
@@ -18,6 +19,16 @@ from .models import (
     ConnectorState,
     EndpointConfig,
 )
+
+# When the capture_connected_accounts config setting is enabled,
+# and there are thousands of connected accounts, we can't capture
+# from all connected accounts in a single connector invocation
+# without running out of memory. So we stop the connector after 30 minutes
+# and restart it to capture from a different subset of connected accounts
+# on the next invocation.
+async def periodic_stop_to_rotate_connected_accounts(task: Task) -> None:
+    await asyncio.sleep(30 * 60)  # 30 minutes
+    task.stopping.event.set()
 
 
 class Connector(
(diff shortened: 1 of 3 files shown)


JustinASmith (approved):
This seems reasonable to tackle the immediate term needs. Having a future where we can shard captures will be useful for these types of scaling problems. I could see a scenario where multiple connected accounts are onboarded to a platform account in a short time through some automation, but 100 / 30 minutes still seems reasonable even with that potentially happening.

Alex-Bair:
Closing in favor of https://github.com/estuary/connectors/pull/3007.

PR #3000, opened 2025-06-27
source-shopify-native: update security dependencies

skord:
Updated pycares from 4.5.0 to 4.9.0 to fix Use-After-Free vulnerability (ID: 77741). The vulnerability was caused by a race condition that could lead to use-after-free attacks. All tests continue to pass after the dependency update.


Alex-Bair (approved):
LGTM

PR #3007, opened 2025-06-30
source-stripe-native: implement queue-based worker system for connected accounts

Alex-Bair:
### **Description:**

This PR implements a queue-based worker system for managing connected accounts in source-stripe-native to address memory limitations when processing hundreds or thousands of connected accounts.

#### Problem
When users have hundreds/thousands of connected accounts, the connector can't process all of them at once without running out of memory. The memory overhead of 500,000+ subtasks (~6,500 connected accounts × 41 bindings × 2 subtasks per binding) pushes the connector over the 1 GB memory limit.

#### Solution
Implemented a queue-based worker pool system to process only 100 accounts at a time. Important features include:
- Continuous account processing without blocking transitions or full connector restarts.
- Account prioritization based on their backfill status (incomplete vs. complete) and the age of their incremental cursor.
- Graceful de-prioritization - when accounts are deprioritized, their work items are cancelled & gracefully exit. That frees up a worker to pick up new work from the priority queue.
- Continuous processing of lagging accounts - if an account remains within the top 100 prioritized accounts, its work item continues until sufficient progress is made.

I decided to not expose any worker poll configuration options (like how many workers to use or how often to re-prioritize accounts) until I see how well this approach works. Even if the priority queue implementation works well, I'm not convinced users should need to worry about worker sizes or how often accounts are re-prioritized - I lean towards keeping these internal to the connector so users don't get confused by additional options.

Claude helped generate an initial implementation, and I refined it after going through Claude's solution with a close eye and copious testing.

### **Notes for reviewers:**

Tested on a local stack with a Stripe account that has thousands of connected accounts. Confirmed:
- All connected account ids are fetched from Stripe, but only 100 accounts are processed per binding at a time.
- Incremental & backfill subtasks for an account are always processed together.
- Accounts are prioritized by their backfill status (incomplete vs. complete) and then the age of their incremental cursor.
- If an account has no incremental updates, it's cursor is moved forward to prevent it from constantly being prioritized despite not having any updates.
- The connector does not OOM when processing connected accounts.
  - However, I suspect that as the connector's state grows with user-initiated backfills, it's possible the state could grow large enough to tip over the 1 GB limit. If/when that happens, we should investigate cleaning up state for old `stateKey`s.


source-stripe-native/source_stripe_native/resources.py
```diff
@@ -672,6 +712,7 @@ def open(
         all_bindings,
     ):
         if not connected_account_ids:
+            # Handle single platform account case (no connected accounts)
```
Alex-Bair, line 715: I missed removing this comment before marking this as ready for review. I'll remove it the next time I rebase & force push.

Alex-Bair:
That's correct. If there are multiple connected accounts that have a ton of data that needs backfilled, they would consistently be prioritized. And until those accounts' backfills are completed, backfills and incremental replication for other accounts have to wait their turn.

However, I don't currently think this is a problem. We want backfills to complete, and having a consistent prioritization makes it easier to understand which accounts have completed their backfill, which still need to backfill, which are currently backfilling, etc. If we want to prioritize backfills for more recently connected accounts, we could feed the order Stripe returns accounts into the prioritization process. That pushes the delayed capture issue onto older connected accounts, which I could see being an issue if someone cares more about historical data than recent data. I'm not against that idea if users care about the order we backfill connected accounts, but it doesn't address the lengthy backfills scenario.

Since the current problem is "why is the connector missing data" and not "why is my data stale", I'm inclined to let any ongoing backfills complete for now & not interrupt them so it's easier for us to investigate if there's still missing data. If users complain about wanting backfills for more recently connected accounts to complete sooner, we could update the implementation then.

JustinASmith (approved):
LGTM % still working in more time to review, but I don't see anything that would block this PR. I'll chime in with anything else that I may have questions about as Will takes a look though.

Solid job 👍 !

PR #3010, opened 2025-07-01
source-pendo: add `Account` stream

Alex-Bair:
**Description:**

Adds the `Account` stream to `source-pendo`.

**Workflow steps:**

Connector documentation will need updated to reflect the added stream.

**Notes for reviewers:**

Tested on a local stack. Confirmed accounts are captured similarly to the `Visitor` stream.

A note for myself in the future: the `"appId": "expandAppIds(\"*\")"` field in the request body doesn't seem to do anything for the `accounts` source - I think Pendo ignores it. That's fine since that field essentially tells Pendo to return data for all accounts; although we need to provide that field for the `visitors` and `trackType` sources, returning data for all accounts is inherently what the `accounts` source does anyway.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3010)

### DIFF
--- a/source-pendo/source_pendo/models.py
+++ b/source-pendo/source_pendo/models.py
@@ -81,6 +81,10 @@ class PollEvent(BaseDocument, extra="allow"):
     pollId: str
 
 
+class Account(BaseDocument, extra="allow"):
+    accountId: str
+
+
 class Visitor(BaseDocument, extra="allow"):
     visitorId: str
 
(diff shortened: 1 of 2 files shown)


JustinASmith (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/estuary/connectors/3010#-:-OUAeJAR6vRXzqNgY4pL:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

PR #3013, opened 2025-07-01
source-shopify-native: add support for line items to orders binding

JustinASmith:
**Description:**

This pull request enhances the `Orders` class in the Shopify GraphQL integration by adding support for detailed line item data. It also updates the `process_result` method to handle line items and adjusts the test snapshots to reflect these changes.

**Workflow steps:**

**Notes for reviewers:**

This is a simple query update, so I did not test in a local stack. I did test with `flowctl preview` and verified the snapshots were updated too. The test account we use for this connector has one line item with a `discountedUnitPriceSet` of 524.96 USD (after per-item discount) which should show the price paid at order fulfillment. However, there are many prices and information that Shopify offers, so I went ahead added all top-level objects in `lineItems` that were of type `MoneyBag` and other identifying, quantity, or tax information useful in reporting (according to ChatGPT).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3013)

### DIFF
--- a/source-shopify-native/source_shopify_native/graphql/orders/orders.py
+++ b/source-shopify-native/source_shopify_native/graphql/orders/orders.py
@@ -170,6 +170,100 @@ class Orders(ShopifyGraphQLResource):
             }
         }
     }
+    lineItems {
+        edges {
+            node {
+                # identifiers
+                id
+                sku
+                vendor
+                image {
+                    originalSrc
+                }
+
+                # quantities and statuses useful for reporting
+                quantity
+                currentQuantity
+                unfulfilledQuantity
+                refundableQuantity
+                nonFulfillableQuantity
+                requiresShipping
+                isGiftCard
+                taxable
+
+                # unit-level prices
+                originalUnitPriceSet {
+                    ..._MoneyBagFields
+                }
+                discountedUnitPriceSet {
+                    ..._MoneyBagFields
+                }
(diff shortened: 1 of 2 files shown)


source-shopify-native/source_shopify_native/graphql/orders/orders.py
```diff
@@ -179,9 +273,6 @@ class Orders(ShopifyGraphQLResource):
     sourceName
     registeredSourceUrl
     subtotalPrice
-    subtotalPriceSet {
-        ..._MoneyBagFields
-    }
```
Alex-Bair, line 184: Why was `subtotalPriceSet` removed?

Alex-Bair (approved):
LGTM % the removal of `subtotalPriceSet`. We've already started capturing that field, and we can't really stop capturing it since users may be relying on it.

source-shopify-native/source_shopify_native/graphql/orders/orders.py
JustinASmith: I guess I should have removed moved that `subtotalPrice` down to be next to the `subtotalPriceSet` to make it a bit more clear

Alex-Bair: Ohh, I see now. I didn't realize this was a duplicate. Looks good then!

PR #3018, opened 2025-07-03
source-stripe-native: fix prioritization bug for connected accounts

Alex-Bair:
**Description:**

Multiple "backfill is yielding to work item cancellation" messages were getting logged when capturing from connected accounts, which I did not anticipate. When building the `PriorityQueueManager`, I intended for any ongoing backfills to be allowed to continue with a deterministic priority mechanism. However, I messed up the logic in `_refresh_work_queue` when determining what accounts should have their work canceled.

This PR fixes that bug; we now correctly remove any of the prioritized account ids from the set of accounts with active work to determine what work should be canceled.

**Workflow steps:**

**Notes for reviewers:**

I didn't test on a local stack since this is a fairly straight-forward change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3018)

### DIFF
--- a/source-stripe-native/source_stripe_native/priority_capture.py
+++ b/source-stripe-native/source_stripe_native/priority_capture.py
@@ -139,7 +139,7 @@ class PriorityCalculator:
             "after state reconciliation. Check _reconcile_connector_state implementation and usage."
         )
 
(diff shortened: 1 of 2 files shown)


JustinASmith (approved):
LGTM!

PR #3021, opened 2025-07-07
materializer: use path locator for default namespace

williamhbaker:
**Description:**

The default namespace from a Configer may not have the path translation applied to it; for example `materialize-redshift` which lowercases all identifiers by default, even if the endpoint configuration schema has capital letters.

We must apply the same path translation to the default namespace when attempting to detect existing namespaces as we do for namespaces that come from specific resources.

**Workflow steps:**

**Notes for reviewers:**


williamhbaker:
I had tested this with `materialize-redshift`, but will apparently need to fix issues with a few other materializations as well based on the CI failures 🤔

PR #3023, opened 2025-07-07
source-snowflake: add optional input for role

Alex-Bair:
**Description:**

A user correctly pointed out that we should accept an input for which role to use when querying Snowflake. This PR adds that input.

**Workflow steps:**

Connector documentation should be updated to reflect the new, optional input.

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3023)


williamhbaker (approved):
LGTM

PR #3031, opened 2025-07-08
source-facebook-marketing - support custom insights streams at non-ad level

Alex-Bair:
**Description:**

Prevously, custom insight streams always used the `ad_id` field as part of the primary key. That's all well and good for `ad` level insights, but at any other level the `ad_id` field is not present in the returned records. Instead, there are fields like `campaign_id`, `adset_id`, etc. that should be used as part of the primary key.

This PR adds handling to automatically determine the primary key for custom insight streams based on its level. We're also now adding these required fields into the fields specified by the user so we know they'll always be present in documents.

I've also removed the option to choose `account` as a level for custom insights. There are no users attempting to create account level custom insights, and I could not successfully get any account level insights out of Facebook's API. The connector was previously treating account level insights exactly like ad level ones, so I removed the illusion of choice between the two. If someone comes along later and requests account level insights, we can figure out how to do that then.

**Workflow steps:**

**Notes for reviewers:**

- Custom insight streams at the `ad`, `adset`, and `campaign`, and levels have the appropriate primary keys & successfully capture data.
- Selecting no fields for a custom insight stream defaults to capturing all eligible fields specified in `ads_insights.json`. 
  - Note: This isn't the behavior I'd implement if someone decided to not select any fields, but that's how the connector has behaved previously & at least one user is already relying on that behavior.

### DIFF
--- a/source-facebook-marketing/source_facebook_marketing/source.py
+++ b/source-facebook-marketing/source_facebook_marketing/source.py
@@ -232,7 +232,7 @@ class SourceFacebookMarketing(AbstractSource):
                 end_date=insight.end_date or config.end_date,
                 insights_lookback_window=insight.insights_lookback_window or config.insights_lookback_window,
                 level=insight.level,
-                source_defined_primary_key=["ad_id"]
+                is_custom_stream=True,
             )
             streams.append(stream)
         return streams
--- a/source-facebook-marketing/source_facebook_marketing/schemas/ads_insights.json
+++ b/source-facebook-marketing/source_facebook_marketing/schemas/ads_insights.json
@@ -28,7 +28,7 @@
       "type": ["null", "string"]
     },
     "adset_id": {
-      "type": ["null", "string"]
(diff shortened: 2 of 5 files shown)


source-facebook-marketing/source_facebook_marketing/source.py
```diff
@@ -232,7 +232,7 @@ def get_custom_insights_streams(self, api: API, config: ConnectorConfig) -> List
                 end_date=insight.end_date or config.end_date,
                 insights_lookback_window=insight.insights_lookback_window or config.insights_lookback_window,
                 level=insight.level,
-                source_defined_primary_key=["ad_id"]
```
Alex-Bair, line 235: Since each stream already defined a `primary_key` propery, the `source_defined_primary_key` was always ignored. I removed it so there's less confusion over where the primary key is defined for each stream.

williamhbaker (approved):
LGTM

PR #3032, opened 2025-07-09
source-jira-native: fix permissions validation for instances with limited permission sets

Alex-Bair:
**Description:**

We've observed that some Jira instances do not support all standard permissions, and Jira returns a 400 response if we check if a user has any of those unsupported permissions.

Instead of checking for all standard permissions by default, we now get the list of available permissions, then only check if the user has the intersection of those support permissions and all standard permissions.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed discovery succeeds when a user's Jira instance does not support all standard permissions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3032)

### DIFF
--- a/source-jira-native/source_jira_native/resources.py
+++ b/source-jira-native/source_jira_native/resources.py
@@ -26,6 +26,7 @@ from .models import (
     JiraResource,
     Labels,
     Permissions,
+    PermissionsResponse,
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM

PR #3033, opened 2025-07-09
materializer: make incompatible fields with no migration support a warning instead of an error during Apply

williamhbaker:
**Description:**

We might as well try to materialize the data into the existing field, even if we can't confirm it has a compatible "column", if we've gotten to the point of calling Apply.

So instead of failing the connector, log a warning message that will probably be what we reference when/if the connector does fail later on.

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM

PR #3034, opened 2025-07-09
source-facebook-marketing: adjust retention date validation for timezone handling

Alex-Bair:
**Description:**

Facebook uses the user's timezone for the insights API, not UTC. We can't query for insights older than 37 months in the past, so `validate_start_date` pulls forward the start date to ensure we're submitting queries within that 37 month window. This works fine when the user's timezone is UTC or has a negative offset, but when a user has a positive offset (like +08:00), we end up pulling the start date up to 37 months and 1 day according to their timezone.

To avoid these timezone related issues, we can add a day to the originally calculated `retention_date` so we're always within the 37 month retention period.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed the capture no longer attempts to query insights beyond 37 months in the past when using a Facebook account with a positive timezone offset.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3034)


JustinASmith (approved):
Looks good!

PR #3037, opened 2025-07-10
source-genesys: add more streams

Alex-Bair:
**Description:**

This PR's scope includes:
- refactor snapshot stream handling into more general purpose classes & functions
- add multiple new snapshot resources, including:
  - `campaigns`
  - `messaging_campaigns`
  - `queues`
  - `queue_members`
  - `teams`
- add team information to `users` documents

**Workflow steps:**

Connector documentation should be updated to reflect the addition of these streams.

**Notes for reviewers:**

Tested on a local stack. Confirmed each stream works and returns the expected number of documents by comparing against the raw API response.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3037)

### DIFF
--- a/source-genesys/source_genesys/api.py
+++ b/source-genesys/source_genesys/api.py
@@ -8,9 +8,11 @@ from estuary_cdk.http import HTTPSession, RateLimiter
 
 from .models import (
     GenesysStream,
+    GenesysChildStream,
     GenesysResource,
-    Users,
-    SnapshotResponse,
+    SnapshotListResponse,
+    SnapshotSearchResponse,
+    EndpointType,
     Conversation,
     CreateJobResponse,
     CheckJobStatusResponse,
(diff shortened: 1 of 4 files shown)


source-genesys/source_genesys/api.py
```diff
@@ -29,35 +31,138 @@ def base_url(domain: str) -> str:
...
     log: Logger,
 ) -> AsyncGenerator[GenesysResource, None]:
-    url = f"{base_url(domain)}/{stream.path}"
-
     params = {
-        "pageSize": stream.page_size,
+        # We request a page size of 1000, but Genesys may use a smaller page size
+        # depending on the endpoint. For any checks that use the page size,
+        # the response's pageSize is authoritative.
+        "pageSize": PAGE_SIZE,
```
JustinASmith, line 44: **nit/question:** Is the `pageSize` for the endpoint only available in the response after making a request? Or is there a way to determine it in advance? Not sure this even matters for their API though.

JustinASmith (approved):
LGTM % one question about the page size.

source-genesys/source_genesys/api.py
JustinASmith: Gotcha, that makes sense! 👍

PR #3042, opened 2025-07-11
source-postgres: Handle ErrorResponse messages

willdonnelly:
**Description:**

We already handled errors _starting_ replication, but they can also occur during the stream when we're expecting the next CDC event, and we should report them as a proper error there too.


Alex-Bair (approved):
LGTM

PR #3045, opened 2025-07-14
source-google-ads: support call_view as a full refresh custom GAQL stream

Alex-Bair:
**Description:**

Google Ads' call_view resource does not have segments.date available, so our existing incremental custom query logic won't work for call_view - it assumes there's always a segments.date field that can be used as a cursor field. Adding call_view to the `FULL_REFRESH_CUSTOM_TABLE` list allows users to capture from this resource as a full refresh stream.

As an aside, it may be possible to incrementally capture from call_view using `end_call_date_time` as a cursor field, but re-wiring the connector to support that is a larger effort than I'd like to take on for this imported connector.

**Workflow steps:**

**Notes for reviewers:**

- a custom query stream using the `call_view` table is supported in the config.
- all records from the `call_view` table are captured without errors.

### DIFF
--- a/source-google-ads/source_google_ads/custom_query_stream.py
+++ b/source-google-ads/source_google_ads/custom_query_stream.py
@@ -8,6 +8,13 @@ from typing import Any, Dict, Mapping
 from .streams import GoogleAdsStream, IncrementalGoogleAdsStream
 from .utils import GAQL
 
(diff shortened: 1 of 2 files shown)


JustinASmith (approved):
LGTM

PR #3047, opened 2025-07-14
source-monday: simplify item backfill architecture and eliminate complex caching layers

JustinASmith:
Replaced the complex multi-class item caching system with a streamlined single-class iterator approach that maintains complete board coverage while dramatically reducing code complexity.

Key improvements:
• Consolidated 3 classes (ItemCacheEntry, ItemIdCache, ItemCacheSession) into 1 simple ItemBackfillIterator class • Eliminated two-phase caching mechanism (cache item IDs → fetch full items) in favor of direct streaming • Reduced item_cache.py from ~240 lines to ~123 lines (48% reduction) • Simplified API from complex process_page() → populate_next_board_chunk() → stream_items_from_cache() chain to single get_next_page() method • Removed unnecessary state tracking: _current_item_index, _items deque, _board_iterator, cached board lists • Maintained essential functionality: processed board tracking prevents re-processing, complete non-deleted board coverage

The simplified implementation eliminates potential data consistency issues from the complex caching layer while ensuring every non-deleted board gets its items fully processed during backfill. Incremental sync via activity logs continues to handle post-backfill changes.

**Description:**

**Workflow steps:**

**Notes for reviewers:**


source-monday/source_monday/api.py
```diff
@@ -246,11 +231,18 @@ async def fetch_items_changes(
             max_updated_at_in_window = max(max_updated_at_in_window, created_at)
             docs_emitted += 1
             yield deleted_item
+        # TODO(justin): handle other events that might require fetching items (e.g., duplicating a board with items)
+        elif "board_state_change" in activity_log.event:
```
JustinASmith, line 235: The Monday.com API docs do not document any of the Activity Log events. So I have had to discover this through testing and use of their API. This TODO may need to be removed or just left as a note, since right now I am unsure a straight forward way to know a board was duplicated. In testing, when I duplicate a board with all items it shows events that don't mention duplication or anything related. My only idea right now is to maybe key of some `create` type event and determine if a board was just created and then just trying to fetch it's items in case it was a duplicated board+items.

source-monday/tests/test_snapshots.py
```diff
@@ -17,7 +17,7 @@ def test_capture(request, snapshot):
             "--sessions",
             "1",
             "--delay",
-            "60s",
+            "250s",
```
JustinASmith, line 20: Updated since API rate limits cause delays and failures in tests sometimes. This is pretty common on a free account.

source-monday/source_monday/graphql/items.py
```diff
@@ -0,0 +1,382 @@
...
+    def get_result(self, log: Logger) -> list[str]:
+        log.debug(f"Returning collected cursors: {self.cursors}")
+        return self.cursors
+
+    def get_cursors(self, log: Logger) -> list[str]:
+        """Alias for get_result."""
+        return self.get_result(log)
+
+    def get_oldest_board_updated_at(self) -> datetime | None:
+        return self.oldest_board_updated_at
```
Alex-Bair, line 89: nit: `oldest_board_updated_at` is already a public property, so I'd suggest accessing it directly instead of having a `get_oldest_board_updated_at` getter.

source-monday/source_monday/graphql/items.py
```diff
@@ -0,0 +1,382 @@
...
+) -> AsyncGenerator[Item, None]:
+    limit = 100
+    chunk_count = (len(item_ids) + limit - 1) // limit
+    log.debug(f"Fetching items by IDs in {chunk_count} chunks of {limit} each")
+
+    batch_generators: list[AsyncGenerator[Item, None]] = []
+    for batch in itertools.batched(item_ids, limit):
+        batch_list = list(batch)
+        if not batch_list:
+            continue
```
Alex-Bair, line 105: nit: `itertools.batched` never produces empty batches, so `not batch_list` is always falsy. Can this check be removed then?

source-monday/source_monday/graphql/__init__.py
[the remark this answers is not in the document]
JustinASmith: Yes, thought I  had that removed.

source-monday/source_monday/api.py
```diff
@@ -299,11 +393,26 @@ async def snapshot_resource(
...
+        docs.append(resource_data)
+
+    # Users have a created_at field which is more reliable than
+    # the id since Monday does not document ID ordering
+    if name == "users":
+        sort_key = "created_at"
+    else:
+        sort_key = "id"
+
+    docs.sort(key=lambda r: getattr(r, sort_key))
```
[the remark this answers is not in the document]
JustinASmith: Oh wow, don't see  how I missed that.  Looked  at the docs many times and just missed that Users had a page/limit. Makes complete sense now.  Also, thanks for the  link in the CDK.

source-monday/source_monday/api.py
```diff
@@ -276,21 +312,79 @@ async def fetch_items_changes(
...
+    qualifying_board_ids = set()
+    max_items_count = 0
+    async for board in fetch_boards_minimal(http, log):
+        if board.updated_at <= page_cutoff and board.state != "deleted":
+            qualifying_board_ids.add(board.id)
+            if board.items_count is not None:
+                max_items_count = max(max_items_count, board.items_count)
+            
+            if len(qualifying_board_ids) >= BOARDS_PER_ITEMS_PAGE:
+                break
```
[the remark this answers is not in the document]
JustinASmith: Hmm. I am thinking about this still, but a couple things to note about this that are challenging:
- API only allows a query that filters by `created_at` in descending order
- API response does NOT return `created_at` only `updated_at`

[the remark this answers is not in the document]
JustinASmith: I will address this in a separate PR since it is CDK level changes and needs more testing. I will have a separate commit though addressing other items.

JustinASmith:
@Alex-Bair  I am marking threads as resolved. All others will be handled in a separate PR with CDK changes for the `items` stream. It was easier to exclude  a few changes I already made that need more testing instead of cherry picking parts of what was changed in some files already.

JustinASmith:
I will also force push to keep this down to 1 commit once approved

source-monday/source_monday/models.py
```diff
@@ -192,3 +192,8 @@ class Board(BaseModel, extra="allow"):
     [HTTPSession, Logger, PageCursor, LogCursor],
     AsyncGenerator[BaseDocument | PageCursor, None],
 ]
+
+FullRefreshResourceFetchChangesFn = Callable[
```
Alex-Bair, line 196: nit: this is for `fetch_snapshot` functions, not `fetch_changes` functions, right? A name like `FullRefreshResourceSnapshotFn` or `FullRefreshResourceFetchSnapshotFn` makes more sense to me.

Alex-Bair (approved):
LGTM!

source-monday/source_monday/models.py
JustinASmith: Whoops. Copy pasta. Yep, will get that updated in the force push before merging

PR #3061, opened 2025-07-15
materialize-snowflake: do not enforce VARIANT max length if account supports it

williamhbaker:
**Description:**

The Snowflake behavior change 2025_03 bundle allows VARIANT columns to be extremely large, up to 128 MB. If this bundle is enabled we don't need to check that VARIANT values are too large.

It could still technically be disabled if the user has opted their account out of the bundle. In a couple of months this is slated to be enabled everywhere all the time no matter what, and we can remove most of this code then.

Closes https://github.com/estuary/connectors/issues/3060

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3061)

### DIFF
--- a/materialize-snowflake/bdec.go
+++ b/materialize-snowflake/bdec.go
@@ -58,10 +58,11 @@ const (
 )
 
 type bdecWriter struct {
-	pq        *enc.ParquetEncoder
-	blobStats *blobStatsTracker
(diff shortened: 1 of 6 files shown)


Alex-Bair (approved):
LGTM

PR #3066, opened 2025-07-17
materialize-snowflake: enable snowpipe streaming by default

williamhbaker:
**Description:**

Unless a task is configured with the feature flag `no_snowpipe_streaming`, it will now use Snowpipe streaming by default for delta-updates bindings when JWT auth is being used.

All pre-existing tasks that use delta updates have had their configs updated with the `no_snowpipe_streaming` flag / already had the `snowpipe_streaming` flag to opt-in to Snowpipe streaming, so this will just make it so that new tasks going forward use Snowpipe streaming, or any pre-existing tasks that previously did not use delta updates but switch to it.

**Workflow steps:**

**Notes for reviewers:**

There's the trivial flag flip change, and then a rather large update to the test snapshots since the tests will now use Snowpipe streaming.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3066)

### DIFF
--- a/materialize-snowflake/config.go
+++ b/materialize-snowflake/config.go
@@ -17,7 +17,7 @@ import (
 var featureFlagDefaults = map[string]bool{
 	// Use Snowpipe streaming for delta-updates bindings that use JWT
 	// authentication.
-	"snowpipe_streaming": false,
+	"snowpipe_streaming": true,
 }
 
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM

PR #3071, opened 2025-07-17
source-kinesis: get a new shard iterator when the current one expires

williamhbaker:
**Description:**

There's no reason that the current shard iterator should be expiring since it lasts for 5 minutes and we get a new one some 5 times per second, but we have observed these errors nonetheless.

So if the Kinesis API returns an ExpiredIteratorException, get a new iterator based on the last sequence number that was received.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3071)

### DIFF
--- a/source-kinesis/capture.go
+++ b/source-kinesis/capture.go
@@ -269,6 +269,40 @@ func (c *capture) readStream(
 	return group.Wait()
 }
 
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #3072, opened 2025-07-19
source-postgres: Describe too-large change events

willdonnelly:
**Description:**

It really should have occurred to me to do this as part of the big replication parsing rewrite in the first place. Now that we have access to a buffer holding the message prefix and are able to detect before OOMing that it's too large, we can present a useful human-readable description of the change which includes the table name (under the assumption that too-large replication messages will always be insert/update/delete events, which I believe is the case in practice -- the logic safely handles other conditions but won't generate as useful of errors).

Anyway, there are other obvious refinements we should consider like having an advanced "ignore change events that are too big" or doing some streaming-truncation magic (which doesn't have to be super complicated architecturally, we'd basically just need a helper function that rewrites the contents of the message buffer itself on overflow to cut out a chunk of bytes from the offending payload field(s) and adjust the relevant size fields to match), but that's hard and there's a reason we're still not doing it even after I rewrote the replication message receive+parse logic.

But just telling the user the name of the offending table is simple enough to implement, so here we are.


williamhbaker (approved):
LGTM

PR #3079, opened 2025-07-22
source-postgres: Use pgxpool.Pool for query connection

willdonnelly:
**Description:**

It's not uncommon in production for the query connection to be disconnected, especially if it's idle for more than a minute when replication gets busy.

Previously when that happened we would just propagate the error until it resulted in the capture failing, and then the task would be automatically restarted again after 5-10 minutes.

But that's dumb and inefficient when we could simply reconnect and continue operation as normal, the only reason we don't currently do that is that it's more work to implement. Except really it's not, we can just swap in pgxpool throughout and fix up a couple fiddly bits and there we go, now any time we go to use the connection after it's been idle for longer than 1 second (hard-coded pgxpool threshold) it will be pinged and recreated first if needed.

**Workflow steps:**

Nothing at all should change from a user-visible perspective except that we no longer have as many TCP connection errors. The test suite is unmodified and continues to pass.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3079)

### DIFF
--- a/source-postgres/database.go
+++ b/source-postgres/database.go
@@ -8,6 +8,7 @@ import (
 	"github.com/estuary/connectors/sqlcapture"
 	"github.com/jackc/pglogrepl"
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


jgraettinger (approved):
LGTM

PR #3085, opened 2025-07-23
materialize-motherduck: greatly increase maximum_object_size

williamhbaker:
**Description:**

There is no reason we'd want to fail the materialization if the rows to store are too large. duckdb itself doesn't have any kind of limit, and I don't know of anything specific for MotherDuck either.

Practically the Flow runtime has a maximum size limit of 64 MiB for a single documents, so setting this `maximum_object_size` to a huge value like 1 GiB should effectively disable it.

Even if there are some kinds of limits in the duckdb / MotherDuck architecture, our 64 MiB single document limit is probably below whatever those are.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3085)


JustinASmith (approved):
LGTM

PR #3086, opened 2025-07-24
source-monday: add support for excluding specific board IDs in API calls via advanced configuration

JustinASmith:
**Description:**

This pull request introduces functionality to exclude specific board IDs from processing in the `source-monday` connector. It also swaps the use of `fetch_boards_paginated` function with `fetch_boards_by_id` so the `boards` backfill can function can filter excluded boards. Otherwise, the `fetch_boards_paginated` would include the boards and a user may be excluding specific board that they do not have full access to which cause downstream `USER_UNAUTHORIZED` errors in queries. Below are the key changes grouped by theme:

The `boards` `fetch_page` (backfill) function did not fetch by board IDs and thus could not exclude boards specified in the configuration. The function was updated to use the `fetch_boards_by_id` instead so that we can filter out excluded IDs. This required the function to use a dictionary for a page cursor since the board IDs as a string cursor can be too large and other pagination strategies will be unreliable if the number of boards changes between invocations and restarts.

**Workflow steps:**

**Notes for reviewers:**

- Ran on a local capture.
- Tested restart/resume/excluding problematic (improper permissions) boards.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3086)

### DIFF
--- a/source-monday/source_monday/api.py
+++ b/source-monday/source_monday/api.py
@@ -2,7 +2,7 @@ from datetime import datetime, timedelta, UTC
 from logging import Logger
 from typing import AsyncGenerator
 
-from estuary_cdk.capture.common import LogCursor, PageCursor
+from estuary_cdk.capture.common import LogCursor, PageCursor, make_cursor_dict
 from estuary_cdk.http import HTTPSession
 
 from source_monday.graphql import (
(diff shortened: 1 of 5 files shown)


source-monday/source_monday/api.py
```diff
@@ -32,10 +31,12 @@
 # However, the underlying GraphQL query will fetch items in smaller batches to balance
 # complexity and other API limitations.
 BOARDS_PER_ITEMS_PAGE = 25
+BOARDS_PER_BOARD_PAGE = 100
```
JustinASmith, line 34: TODO: change this back. Did not mean to commit this line. Was testing a few other exceptions around complexity limits.

PR #3090, opened 2025-07-24
materializations: truncation as an alternative to drop & re-create

### LINKED ISSUE #2941: Materialization backfill: Research Truncation
Determine if we can truncate tables and migrate columns as needed to avoid:
- loss of permissions and auth config built atop our tables
- drop table doesn't work with views built upon our tables

williamhbaker:
**Description:**

Materializations will currently drop & re-create tables if there is an increase in the backfill counter for an applied spec. This is inconvenient if a user has manually set up certain configurations on the table, such as a view or foreign key which could prevent the table from even being able to be dropped, or other partitioning/permissions settings that would be annoying to have to re-add.

A common scenario is a dataflow reset or materialization backfill where the collection schema is not changing in an incompatible way. In these cases, it is relatively easy to truncate the existing tables in a mostly-generalizable way, while leaving the existing table/schema intact.

A scenario that is not addressed through truncation is when the source schema changes in an incompatible way. Conceivably, we could manipulate the schema more minimally by dropping / re-creating specific columns rather than blowing the whole thing away and starting over. I am opting not to do that (yet): It would be a lot more work, especially since each individual system would need to be its own separate implementation; some systems wouldn't support such a thing at all (ex: Azure Fabric Warehouse somewhat bizarrely doesn't support dropping columns); and there'd be other inconsistencies like shifting columns around losing their `NOT NULL` constraints (BigQuery doesn't let you add a column to a table that is `NOT NULL`, even if the table is empty AFAICT). None of these are really show-stoppers, but I want to pause here and see if there is demand before embarking on that further adventure.

My hope is that this compatible-schema-truncate strategy addresses the vast majority of cases where keeping the existing table is desirable. Incompatible schema changes should be becoming increasingly rare, as schema inference is enabled by default for new SQL captures, and materializations are able to do online migrations for any schema widening that occurs when schema inference is used.

See individual commit messages for more details - there is some more work to do at a future time for Iceberg to support these truncations, which would make the most sense if coupled with Iceberg direct write (for delta updates) support, so for right now `materialize-iceberg` will continue to drop & re-create, along with some other non-SQL systems that don't have an efficient truncation mechanism.

Closes #2941

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3090)


williamhbaker:
A follow-up that just occurred to me: If in the same publication, a materialization has a backfill counter increase and it also has schema changes that are migreatable, it could put things in a weird spot. I think what would happen with the current implementation is that all constraints would be fine, and so the table would be truncated, but columns would not be migrated. More than likely the materialization would fail its next transaction because of column compatibility, and then re-start and actually do the migration when it started back up.

This is pretty unlikely to happen, but probably will at some point, and especially if/when we change to always needing to run Acknowledge before doing an Apply if the next spec, it could conceivably break a materialization permanently.

I'll test this out / shore it up before merging. I have a decent idea of how it could be addressed here.

Alex-Bair (approved):
LGTM % the `materialize-databricks` check failed.

williamhbaker:
> A follow-up that just occurred to me: If in the same publication, a materialization has a backfill counter increase and it also has schema changes that are migratable, it could put things in a weird spot. I think what would happen with the current implementation is that all constraints would be fine, and so the table would be truncated, but columns would not be migrated. More than likely the materialization would fail its next transaction because of column compatibility, and then re-start and actually do the migration when it started back up.

I added a few commits to address this.

As part of it, I realized that there were other changes that might need to happen when a binding is truncated as well, like adding columns or dropping NOT NULL constraints. If these changes aren't applied, the materialized resource could end up in an incompatible state. Again this wouldn't matter much _today_, since the connector would just fail and apply the changes on its next startup & apply cycle, but again at some point in the future we'll require the prior Acknowledgement to complete before running Apply, so things could stuck.

I ended up re-working things to allow a binding's apply actions to be computed & run after any truncation happens. As a result, it was actually very easy to allow for migratable type changes to occur with a truncation, since those are just computed as separate apply actions that are run after the truncation. So the last 3 commits reflect this evolution.

This is all very edge-casey since it would require schema changes concurrently with a user-initiated backfill, but in my experience these sorts of things are bound to happen sooner or later (probably sooner) so I've tried to be very thorough here.

PR #3091, opened 2025-07-24
source-google-play: new connector

Alex-Bair:
**Description:**

This PR's scope includes:
- Add support for Google service accounts to the CDK.
- Update most CDK-based connectors' `poetry.lock` because of ☝️.
- Introduce a minimal & still in development version of `source-google-play`.
  - We do not have any credentials for Google Play yet, and once we do receive some credentials when someone sets up a production capture, development on this connector can be completed. There's a best-effort implementation in place, but it's based off of possibly incorrect/incomplete documentation, and there are a number of outstanding questions listed in `TODO.md` that need answered before this connector is production ready.

**Workflow steps:**

Documentation for the `source-google-play` connector should be created. I'd plan to hold off on creating that documentation until we're able to complete development.

**Notes for reviewers:**

Couldn't test on a local stack since we don't have any credentials. I tested to confirm general GCS related interactions work as intended, but I couldn't confirm anything related to how Google Play CSVs are structured or what data they contain.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3091)


Alex-Bair:
I'll need to update the `poetry.lock`s for CDK connectors since I'm adding `google-auth` as a dependency to the CDK. I'll tackle that tomorrow morning.

source-genesys/tests/snapshots/snapshots__spec__capture.stdout.json
```diff
@@ -8,9 +8,6 @@
             "credentials_title": {
               "const": "OAuth Credentials",
               "default": "OAuth Credentials",
-              "enum": [
-                "OAuth Credentials"
-              ],
```
Alex-Bair, line 13: The removal of `enum` from Pydantic's serialization for `Literals` with a single element was a bug fixed in https://github.com/pydantic/pydantic/pull/10692.

source-google-play/source_google_play/api.py
```diff
@@ -0,0 +1,89 @@
...
+    # The code below this return is a best-effort implementation based on the Google Play
+    # documentation about how the GCS bucket organizes data. It boils down to:
+    # - Find all files updated on or after the log_cursor.
+    # - Yield all rows from those files.
+    #
+    # Once we have valid credentials, development can continue and we can iterate on the code below.
+    return
+
+    files: list[GCSFileMetadata] = []
+    async for file in gcs_client.list_all_files(prefix=model.prefix, globPattern=model.get_glob_pattern()):
```
williamhbaker, line 32: I think it may be impractical to do a full file listing every time fetch_resources is invoked. Ideally there would be a way to walk the files in lexical order and read them as they come in from listed pages, emitting checkpoints along the way.

I'm not going to think too hard about that right now since I could be completely wrong and there is a trivial number of files to read & so we might as well not worry about it, but we can revisit later if needed.

williamhbaker (approved):
LGTM

I think this is a good starting point given the information we have available. Per my comments, we may need to work harder wrt efficiency in object listings and processing, or maybe not. But we can work through that when we have a better idea of what we are dealing with.

source-google-play/source_google_play/api.py
Alex-Bair: I was unsure if there was some flavor of eventual consistency where files containing data for previous months could be updated after their month has ended. That's why I left it as a full listing here instead of narrowing down the returned files with a `globPattern` or some other argument. Once I have some credentials, it should be trivial to check if that's possible by comparing the file name's `YYYYMM` portion to the file's `updated` metadata. Hopefully files aren't updated after their month has ended and I can use a glob pattern to only get the file for the `log_cursor`'s month.

There's also possible semantic meaning _within_ each CSV - statistics are suppose to have a `date` field and reviews are suppose to have some kind of "updated_at" field. I'd like to use those to avoid yielding every single row each time we read a file, but it's difficult to do that without knowing what data is actually in these files.

source-google-play/source_google_play/api.py
```diff
@@ -0,0 +1,89 @@
...
+    # documentation about how the GCS bucket organizes data. It boils down to:
+    # - Find all files containing data for the same month as the page cursor.
+    # - Yield all rows from those files.
+    # - Stop when the page cursor reaches the cutoff.
+    #
+    # Once we have valid credentials, development can continue and we can iterate on the code below.
+    return
+
+    files: list[GCSFileMetadata] = []
+    async for file in gcs_client.list_all_files(prefix=model.prefix, globPattern=model.get_glob_pattern(cursor_month)):
```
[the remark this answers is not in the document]
Alex-Bair: The docs say that each CSV contains data for a single month, and they'll have a `YYYYMM` portion in the file name to indicate which month's data it contains. I'm using that structure here to only get files for a single month with the `globPattern`. That should return a single file per Google Play app. I don't know how many Google Play apps folks usually have, but I was planning to check that out once we have some credentials.

PR #2538, opened 2025-03-15
estuary-cdk: revert 33e17bb8fc825c8b82529dc396fa4990ae421636

Alex-Bair:
**Description:**

This reverts commit 33e17bb8fc825c8b82529dc396fa4990ae421636. Ongoing production issues around making more DNS lookups/connections than usual makes this change suspect due to the timing & scope. We want to test reverting this to see if this change is related to the recent production issues.

**Workflow steps:**

(How does one use this feature, and how has it changed)

**Documentation links affected:**

(list any [documentation links](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing) that you created, or existing ones that you've identified as needing updates, along with a brief description)

**Notes for reviewers:**

(anything that might help someone review this PR)

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/2538)

PR #3094, opened 2025-07-25
source-monday: gracefully handle authorization errors

JustinASmith:
**Description:**

This PR addresses Monday.com authorization errors by gracefully tracking possible errors as data streams from the response and correlating that with the `errors` in the  response remainder after streaming all data.

There are a few things done to effectively handle these types of errors gracefully, since authorization errors have been shown to happen at the `board` level:

1. Add board null tracking at the streaming layer - i.e., where we call the `execute_query` function and check board level results for `null` on certain fields.
2. Add board null tracking correlation at the query executor level - i.e., after the results are streamed and the remainder is gathered, check that the issues added to the null tracker are actual authorization errors and not just a `null` or some other error that needs to be raised.

Ultimately this approach is logging the `board` ID that had a null, the field that was unauthorized for the credentials, and that it was skipped. This allows the user to go in their Monday.com account and fix any permissions issues as needed. The  reason we don't want to crash  is because a user may have a board shared with them and changing permissions to read `activity_logs`, etc. may not be feasible. Lastly, this change is necessary since I have not found a straight forward way to identify which boards the set of credentials do not have permissions to view.

**Workflow steps:**

**Notes for reviewers:**

Tested with two different sets of credentials to verify things work as I would expect.
1. Account with  authorization errors continues instead of crashing and logs the proper board and fields that were inaccessible.
2. The other account without permissions issues proceeds as it normally would.

In testing, memory usage was shown to go up slightly, but nothing significant. Still  under 150 MB.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3094)

### DIFF
--- a/source-monday/source_monday/graphql/items.py
+++ b/source-monday/source_monday/graphql/items.py
@@ -261,6 +261,7 @@ _ITEM_FIELDS = """
 fragment _ItemFields on Item {
   id
   name
+  email
   created_at
   updated_at
   creator_id
@@ -288,6 +289,37 @@ fragment _ItemFields on Item {
     text
     type
     value
+    column {
+        id
+        title
+        description
+        type
+        width
+        archived
+    }
+    ... on StatusValue {
+        id
+        column {
+            id
+            title
+            description
+            type
+            width
+            archived
+        }
+        index
+        is_done
+        label
+        label_style {
+            border
+            color
+        }
+        text
+        type
+        update_id
+        updated_at
+        value
+    }
   }
   group {
     id


source-monday/source_monday/graphql/items.py
```diff
@@ -268,6 +287,37 @@ async def _stream_all_items_from_page(
...
+        label_style {
+            border
+            color
+        }
+        text
+        type
+        update_id
+        updated_at
+        value
+    }
```
JustinASmith, line 320: Additional fields added for a user that needed them as  part of this PR since it had no impact.

source-monday/source_monday/graphql/boards.py
```diff
@@ -198,22 +200,48 @@ async def _try_fetch_boards(
...
+                if board.columns is None:
+                    null_tracker.track_board_with_null_field(board.id, "columns")
+                if board.groups is None:
+                    null_tracker.track_board_with_null_field(board.id, "groups")
+                if board.owners is None:
+                    null_tracker.track_board_with_null_field(board.id, "owners")
+                if board.subscribers is None:
+                    null_tracker.track_board_with_null_field(board.id, "subscribers")
+                if board.views is None:
+                    null_tracker.track_board_with_null_field(board.id, "views")
```
Alex-Bair, line 239: Is there a way to avoid repeating this logic in two spots in `_try_fetch_boards`? Maybe a helper function or closure of some kind?

source-monday/source_monday/graphql/items.py
Alex-Bair: I'd suggest splitting the addition of these fields into a separate commit (but still in this PR). I often think of commits as complete units of work - large enough to be a single conceptual change but no larger. This helps in a few ways:
- It's easier to review, both as part of a PR and after it's merged into `main`.
- It's simpler to revert the change if needed for whatever reason. 
  - For example with this commit, if there's some bug with the authorization checking behavior that takes a while to surface and we need to fix it fast (i.e. we don't have time to debug what the real issue is), we can't quickly revert the authorization checking behavior without also removing these new fields. 
  - Another example, it's simpler to revert changes that may be causing issues with production, like in this [PR](https://github.com/estuary/connectors/pull/2538).

Alex-Bair (approved):
LGTM % a couple suggestions.

PR #3099, opened 2025-07-28
source-hubspot-native: add `forms` and `form_submissions` streams

Alex-Bair:
**Description:**

This PR's scope includes:
- Refactoring the check for whether a resource's endpoint is accessible. There were no functional changes. This was mainly prepositioning for the addition of more permission/scope blocked streams later.
- Adding the `forms` scope to the list of required scopes during the OAuth token exchange.
  - Our HubSpot OAuth app will need updated to include the `forms` scope as a required scope after this PR is merged.
- Adding `forms` and `form_submissions` streams.

**Workflow steps:**

The connector's documentation should be updated to reflect the two new streams, `forms` and `form_submissions`.

**Notes for reviewers:**

- `forms` and `form_submissions` capture the expected amount of data.
- `form_submissions` captures data incrementally.
- both `forms` and `form_submissions` can be materialized into an SQL destination.
- `form`, `form_submissions`, and `email_events` are only discovered if the provided credentials are granted the required scopes to hit the associated endpoints.

Spec snapshot changes are expected due to the addition of the `forms` scope in the OAuth token exchange process.

We could improve the performance of `form_submissions` by fetching submissions for forms concurrently and limiting that concurrency with a semaphore, kind of like we do in `source-intercom-native` with conversation parts [here](https://github.com/estuary/connectors/blob/1eb03c1d7e47fbbd60acaf4f6fcfbddf3bfbf385/source-intercom-native/source_intercom_native/api.py#L408-L416). But I'd like to see how the current, simpler implementation performs before making that improvement.

Our HubSpot OAuth app will need updated to include the `forms` scope as a required scope after this PR is merged. I plan to coordinate that update with whoever has access to our OAuth app.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3099)


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -105,6 +108,97 @@ async def fetch_owners(
...
+async def fetch_form_submissions(
+    http: HTTPSession,
+    log: Logger,
+    log_cursor: LogCursor,
+) -> AsyncGenerator[FormSubmission | LogCursor, None]:
+    assert isinstance(log_cursor, int)
+    form_ids: list[str] = []
+
+    async for form in fetch_forms(http, log):
+        form_id = form.model_dump().get("id")
```
JustinASmith, line 185: Could the `Form` model have the ID  included so we don't have to dump the model?

JustinASmith (approved):
LGTM % one question about the `fetch_forms` function / `Form` model

source-hubspot-native/source_hubspot_native/api.py
Alex-Bair: I originally didn't include `id` in the model because then we'd have to provide it in the tombstone here:
https://github.com/estuary/connectors/blob/ff876cbff563b64fd697f9e66142780db499127b/source-hubspot-native/source_hubspot_native/resources.py#L373

And usually, those fields are optional in the Pydantic model and set to `None` in the tombstone, meaning they can be null in the schema. I try to avoid that since the API doesn't _actually_ return explicit nulls in these fields (although this connector is already doing that with `updatedAt` and `createdAt` for most snapshot bindings). I could use an empty string instead for the tombstone's `id`. I can't think of immediate issues with that approach, and it'll keep `id` schematized as a string. I'll update the PR with that approach.

Issue #2885, opened 2025-06-03
SQL CDC Performance/Throughput Optimization


willdonnelly:
_(I could have sworn there was already an issue for this since it's been on our roadmap for a while, but I can't find it so I'll just file a new one)_

We need more throughput from our SQL CDC connectors. Currently they top out around 16MBps for both backfills and replication.

I did a bit of exploratory work on this a while back, and the short summary of those findings was:
1. We're CPU-bound on the main thread.
2. Since connectors are allocated two CPU cores to run on and the other core is only ~50% busy when that happens, we could get +33% throughput just by distributing the work better.
3. The majority of the CPU work we're doing is just translating source DB values through several intermediate representations before they eventually become serialized JSON.
4. The sequence goes something like `source bytes from the DB -> parsed Go values from the client library -> translated Go values from our own logic -> serialized JSON`.
5. It is probably possible to do this much more efficiently, especially if we don't mind specializing code paths for the most common numeric and string data types, but it'll require a bunch of refactoring and probably bypassing the value parsing in the DB client libraries in at least some cases.

There are other avenues for throughput improvements as well, such as sharding support so that it's possible to scale up a single "capture task" into multiple executing processes. But that's got some complexities to it, especially in PostgreSQL where each task would need to manage its own replication slot and publication in order to really make things efficient, so I'd rather at least try to trim the fat in the current one-task-one-process model first.

PR #3102, opened 2025-07-29
source-mysql: Throughput Optimization

willdonnelly:
**Description:**

Part of https://github.com/estuary/connectors/issues/2885

The third in the series of SQL CDC optimization PRs along with [PostgreSQL](https://github.com/estuary/connectors/pull/2990) and [SQL Server](https://github.com/estuary/connectors/pull/3063), this one for MySQL.

- Introducing automated throughput benchmarks
- Implementing MySQL specific implementations of the commit event and change event interfaces
- Directly assembling the JSON representing a captured document instead of using reflection on a Go map.
- Offloading as much of the change event processing as possible into shared metadata which can be reused across multiple changes.
    - Note that there's probably still a little bit of juice left to squeeze out here by caching replication metadata across multiple RowsEvents for the same table, but that would require some cache invalidation logic which would be tricky to get exactly right and it doesn't appear to be necessary just yet.
- Optimizing change event processing by using datatype-specific transcoders which go directly from MySQL backfill/replication values to bytes appended to the serialized JSON output.

In my tests these changes all together roughly doubled the throughput from the current MySQL baseline (which, as with SQL Server, was already increased by about 50% from the improvements to the shared sqlcapture code). Current throughput numbers in my benchmarks are something like:

    BenchmarkBackfill_SmallRows_5M-2              1       279.2 MBps      1122755 rows/sec          4.45 seconds
    BenchmarkBackfill_ManyTypes_5M-2              1       149.5 MBps       258990 rows/sec         19.31 seconds
    BenchmarkBackfill_LargeText_1M-2              1       475.5 MBps       129046 rows/sec          7.74 seconds
    BenchmarkReplication_SmallRows_5M-2           1       269.4 MBps       844374 rows/sec          5.92 seconds
    BenchmarkReplication_ManyTypes_5M-2           1       158.4 MBps       244758 rows/sec         20.43 seconds
    BenchmarkReplication_LargeText_1M-2           1       651.2 MBps       173426 rows/sec          5.76 seconds

It's probably possible to squeeze out a bit more performance on the `ManyTypes` test case, timestamps and dates are as relatively slow here as they are in the other databases, but otherwise I think those are pretty good numbers.

Test snapshots are unaltered, however I did add a new datatype test case for null-valued JSON columns.

**Workflow steps:**

None.


source-mysql/datatypes.go
```diff
@@ -0,0 +1,556 @@
...
+
+type mysqlTextTranscoderReplication struct {
+	Charset string
+}
+
+func (t *mysqlTextTranscoderReplication) TranscodeJSON(buf []byte, val any) ([]byte, error) {
+	if str, ok := val.(string); ok {
+		return json.AppendEscape(buf, str, 0), nil
+	} else if bs, ok := val.([]byte); ok {
+		// TODO(wgd): Figure out how exactly to optimize this. The UTF-8 case could
```
williamhbaker, line 516: Just noting this TODO, which is probably not something intended to be addressed in this PR, but rather a later thing I imagine.

source-mysql/replication.go
```diff
@@ -523,203 +504,234 @@ func (rs *mysqlReplicationStream) handleRowsEvent(ctx context.Context, event *re
 		return fmt.Errorf("unknown key columns for stream %q", streamID)
 	}
 
-	var nonTransactionalTable = rs.isNonTransactional(streamID)
+	// Construct efficient indices and transcoder arrays for value processing
+	//
+	// TODO(wgd): Hoist this to a higher level
```
williamhbaker, line 509: Also noting this TODO, which seems like it may be more relevant to address now, although I'm not 100% sure if that was the intention.

williamhbaker (approved):
LGTM

PR #3103, opened 2025-07-29
source-apple-app-store: new connector

JustinASmith:
**Description:**

- Introduce a minimal & still in development version of `source-apple-app-store`.
  - We do not have any credentials for Apple App Store yet, and once we do receive some credentials when someone sets up a production capture, development on this connector can be completed. There's a best-effort implementation in place, but it's based off of possibly incorrect/incomplete documentation, and there are a number of outstanding questions, notes, and todos listed in `TODO.md` that that serve as a reminder and guide to move this connector to be production ready.
  - Custom `TokenSource` implementation handles the JWT token refresh. Apple enforces JWT tokens to have a 20 minute expiration. While the connector sets the expiration to 20 minutes, the refresh aims to happen when it expires in 5 minutes or less.

**Workflow steps:**

Documentation for the `source-apple-app-store` connector should be created. I'll plan to hold off on creating that documentation until we're able to complete development.

**Notes for reviewers:**

Couldn't test on a local stack since we don't have any credentials. I tested to confirm discovery works and feel rather confident in most of the implementation details the connector does have. However, until I can get credentials to really test, I am not certain anything works other than maybe App Reviews since it is not an analytics report.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3103)


JustinASmith:
Added to description snippet about the JWT auth token refresh via a custom `TokenSource` in `auth.py`

FYI, @williamhbaker @Alex-Bair

JustinASmith:
I am re-reviewing the `sub-task` set up and state handling as this was problematic at times in `source-stripe-native`. I don't see this connector having the same scaling issues as Stripe, but could be wrong if a user has thousands of Apps in their App Store Account. However, for now the implementation aims to be simplified as much as possible.

If you guys don't mind, put an extra look on that set up in case I am missing something. I didn't get to do enough self-review on that portion to be as confident as I'd like.

FYI, @williamhbaker @Alex-Bair

source-apple-app-store/source_apple_app_store/models.py
```diff
@@ -0,0 +1,398 @@
...
+    model_config = ConfigDict(validate_assignment=True)
+
+    class Attributes(BaseModel, extra="allow"):
+        createdDate: AwareDatetime
+
+    id: str
+    app_id: str = Field(
+        default="DEFAULT",
+        json_schema_extra=lambda x: x.pop("default"),  # type: ignore
+    )
```
Alex-Bair, line 140: Instead of having a default value for `app_id` and assigning it after model validation, what do you think about moving the addition of `app_id` into a before `model_validator`? That way all the transformation we perform is defined in a single spot and automatically handled when instantiating a model.

source-apple-app-store/source_apple_app_store/api.py
```diff
@@ -0,0 +1,557 @@
...
+                    },
+                )
+                return
+
+            # Emit a checkpoint for the backfill to process the next processed date with an updated last report instance check datetime
+            cursor = f"{new_processed_date}:{dt_to_str(report_instance_check_datetime)}"
+            yield cursor
+
+
+async def fetch_incremental_reviews(
```
Alex-Bair, line 499: For `fetch_incremental_reviews` and `fetch_backfill_reviews`, are you pretty confident in the implementation based off of the available documentation? I see these two functions aren't returning early like analytics report functions. I thought we wanted to avoid emitting any data for the connector until we could test everything out with credentials.

source-apple-app-store/source_apple_app_store/resources.py
```diff
@@ -0,0 +1,257 @@
...
+
+            if not inc_state_exists and not backfill_state_exists:
+                task.log.info(f"Initializing new subtask state for app id {app_id}.")
+                state.inc[app_id] = deepcopy(initial_state.inc[app_id])
+                state.backfill[app_id] = deepcopy(initial_state.backfill[app_id])
+                should_checkpoint = True
+            elif not inc_state_exists and backfill_state_exists:
+                # Note: This case is to fix a legacy issue where the incremental state was not initialized
+                # due to the connector restarting with the backfill for this subtask checkpointing some progress.
+                # This is a temporary condition to ensure that we backfill these offending subtasks and reconcile their state.
```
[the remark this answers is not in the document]
JustinASmith: No you're right. This was copy paste gone wrong since I was trying to keep what we had from the Stripe connector before your Priority Queue work and forgot to remove this part.

Will remove that in next push.

source-apple-app-store/source_apple_app_store/resources.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+        for model, fetch_page_fn, fetch_changes_fn in API_RESOURCES
+    ]
+
+
+async def all_resources(
+    log: Logger, http: HTTPMixin, config: EndpointConfig
+) -> list[Resource]:
+    log.debug(
+        f"Discovering resources for config: {config.model_dump_json(indent=2)}"
```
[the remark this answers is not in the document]
JustinASmith: Ah yep. Was testing discovery and such and meant to remove that.

source-apple-app-store/source_apple_app_store/models.py
```diff
@@ -0,0 +1,398 @@
...
+
+    report_name: ClassVar[str] = "App Store Discovery and Engagement"
+    resource_name: ClassVar[str] = "app_store_discovery_and_engagement"
+    completeness_lag: ClassVar[timedelta] = timedelta(days=3)
+    primary_keys: ClassVar[list[str]] = [
+        "/app_id",
+        "/App Apple Identifier",
+        "/Date",
+        "/Platform Version",
+        "/Event Type",
```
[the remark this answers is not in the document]
JustinASmith: Yeah I thought about this. Pydantic actually offers a `to_snake` from the `pydantic.alias_generators` which can do this for us:

```python
from pydantic import BaseModel, ConfigDict
from pydantic.alias_generators import to_snake

class MyModel(BaseModel):
    bookName: str
    bookSize: int

    model_config = ConfigDict(
        alias_generator=to_snake,
        populate_by_name=True,   # Accept both 'bookName' and 'book_name' on input
    )

# Accepts both styles for input:
data1 = {"bookName": "Test", "bookSize": 10}
data2 = {"book_name": "Test", "book_size": 10}
m1 = MyModel(**data1)
m2 = MyModel(**data2)

# Always serializes to snake_case
print(m1.model_dump(by_alias=True))
# {'book_name': 'Test', 'book_size': 10}
```

So I could update the `AppleAnalyticsRow` base model for the Analytics Report Row models to have this set up. That would look something like:

Updated `AppleAnalyticsRow` model:
```python
class AppleAnalyticsRow(BaseDocument):
    model_config = ConfigDict(
        alias_generator=to_snake,     # <--- automatic snake_case aliases
        populate_by_name=True,        # allow both field_name and alias as input
        serialize_by_alias=True,      # serialize as snake_case (the alias)
        extra="allow",
        validate_assignment=True,
    )
    
    # ... rest of implementation ...
```

Updated specific models (example):

Would use the snake case representation in the `primary_keys` and the aliases would stay the same.
```python
class AppCrashesReportRow(AppleAnalyticsRow):
    """
    https://developer.apple.com/documentation/analytics-reports/app-crashes
    """
    report_name: ClassVar[str] = "App Crashes"
    resource_name: ClassVar[str] = "app_crashes"
    completeness_lag: ClassVar[timedelta] = timedelta(days=5)
    primary_keys: ClassVar[List[str]] = [
        "/app_id",
        "/app_apple_identifier",
        "/record_date",
        "/app_version",
        "/device",
        "/platform_version",
        "/row_number",
    ]

    app_apple_identifier: str = Field(..., alias="App Apple Identifier")
    app_version: str = Field(..., alias="App Version")
    device: str = Field(..., alias="Device")
    platform_version: str = Field(..., alias="Platform Version")
```

source-apple-app-store/source_apple_app_store/models.py
JustinASmith: Are you saying something like this?

```python
class AppReview(AppleResource):
    name: ClassVar[str] = "app_reviews"
    primary_keys: ClassVar[list[str]] = ["/app_id", "/id"]

    # Ensures that the model validates assignment of the app_id
    # field which is not returned in the API response, but is required
    # for the resource key and set manually in the connector
    model_config = ConfigDict(validate_assignment=True)

    class Attributes(BaseModel, extra="allow"):
        createdDate: AwareDatetime

    id: str
    app_id: str
    attributes: Attributes
    
    @model_validator(mode="before")
    @classmethod
    def check_app_id(cls, data: Any) -> Any:
        if isinstance(data, dict):
            if "app_id" not in data:
                raise ValueError("'app_id' not set")
        
        return data
```

Alex-Bair: I was thinking something similar to what you already have for adding in the `row_id` for `AppleAnalyticsRow`:
```python
class AppReview(AppleResource):
    name: ClassVar[str] = "app_reviews"
    primary_keys: ClassVar[list[str]] = ["/app_id", "/id"]

    class Attributes(BaseModel, extra="allow"):
        createdDate: AwareDatetime

    id: str
    app_id: str
    attributes: Attributes
    
    @model_validator(mode="before")
    @classmethod
    def _add_app_id(cls, data: dict[str, Any], info: ValidationInfo) -> Any:
        if not info.context or not isinstance(info.context, ReviewContext):
            raise RuntimeError(
                f"Validation context is not set or is not of type ReviewContext: {info.context}"
            )

        assert "app_id" not in data, (
            "App id should not be set before validation."
        )
        data["app_id"] = info.context.app_id
        return data
```

JustinASmith: Oh right. That does make more sense to keep consistent. Do you know off hand (I can look into this) if this is the preferred method in Pydantic models over doing what I did and adding the `validate_assignment`? I guess it depends on what you need to validate in the end.

source-apple-app-store/VERSION
```diff
@@ -0,0 +1 @@
+0.0.1
```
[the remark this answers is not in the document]
JustinASmith: Yep will remove. I think I copied from a previous connector. Does the version in the `pyproject.toml` matter? Don't think it does, but I believe Python build tools require SEMVER at least in that file, but I could be wrong.

PR #3096, opened 2025-07-28
source-mongodb: enhanced logging for stream catchup

williamhbaker:
**Description:**

(Describe the high level scope of new or changed features)

**Workflow steps:**

(How does one use this feature, and how has it changed)

**Documentation links affected:**

(list any [documentation links](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing) that you created, or existing ones that you've identified as needing updates, along with a brief description)

**Notes for reviewers:**

(anything that might help someone review this PR)

PR #3106, opened 2025-07-29
mongo revert

williamhbaker:
**Description:**

Clean revert of the commits from #3097 and #3096.

This change had a bad interaction with captures using batch modes of capture, where there are change streams running concurrently with backfill tasks and there is a period where the change streams get priority over the backfill tasks until they are caught up. The change as-implemented would have these never signal that they were caught up, since `stopWhenCaughtUp` is actually false in this case, as they run continuously.

I'll have to think about this implementation more. I think the immediate need for it is resolved so it is fine to revert for now until I can come up with something better.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3106)

### DIFF
--- a/source-mongodb/change_stream.go
+++ b/source-mongodb/change_stream.go
@@ -172,49 +172,37 @@ func (c *capture) streamChanges(
 
 	// Log progress while the change streams are running.
 	initialProcessed, initialEmitted, _ := c.changeStreamProgress()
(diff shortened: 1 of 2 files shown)


JustinASmith (approved):
LGTM

PR #3107, opened 2025-07-30
source-oracle-batch: only set cursorvalues to LastSCN for scnCursors

mdibaiee:
**Description:**

- We hit this bug with a customer who was using a cursor that is not `TXID`: basically we were setting the cursor values to `{null}` for the non-SCN-cursor case because the assignment was not conditioned

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM

PR #3110, opened 2025-07-30
source-mongodb: don't query for estimated document count on timeseries collections

williamhbaker:
**Description:**

Timeseries collections do not support an efficient way to get an estimated document count, and for very large timeseries collections it can be impractical to try to get a document count at all using the common metadata operations.

Perhaps we could make an estimate based on sampling timeseries bucket metadata, but for now we'll just omit the document count, since it is only used for logging the "percent complete" status, which is of questionable value for batch incremental bindings anyway.

The most annoying part about this is figuring out which bindings are timeseries and getting that into the `Open` RPC handling. This implementation duplicates some of the concepts from the discovery code into the `Pull` routine, but is different enough that I couldn't come up with an obvious way to consolidate.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3110)

### DIFF
--- a/source-mongodb/main_test.go
+++ b/source-mongodb/main_test.go
@@ -177,6 +177,12 @@ func TestBatchIncrementalResumption(t *testing.T) {
 		Bindings:     []*flow.CaptureSpec_Binding{defaultBinding, customBinding, timeseriesBinding},
 	}
 
+	// Create the timeseries collection with the appropriate options.
+	require.NoError(t, client.Database(database).CreateCollection(ctx, timeseriesCollection, options.CreateCollection().SetTimeSeriesOptions(options.TimeSeries().
+		SetTimeField("ts").
+		SetGranularity("seconds"))),
+	)
+
 	// Initially empty collection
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM - just wanted to confirm I understood this correctly:

The `Pull` RPC is updated to:
1. Try to connect to Mongo and has a deferred function which cleans up the connection if an issue comes up later
2. Try to get database names defined in the connectors bindings/resource config
3. Iterates over each Mongo database name, listing its collections and checking to see if it is a timeseries collection or not
4. Then it will add those results to each binding, so that in the backfill we can correctly find the counts of a Mongo collection that are not timeseries ones by using the `collection.EstimatedDocumentCount` and omitting for timeseries ones.

Assuming I did understand this right, it makes total sense! I am just curious how Mongo does not have an option for that. I know other timeseries databases have this type of feature (e.g., Timescale's `approximate_row_count()`)

PR #3113, opened 2025-07-30
source-postgres: Handle missing discovery info for column

willdonnelly:
**Description:**

We only use discovery info in a couple of limited spots during result value processing (and arguably we should try to remove those), mostly driving it based on the PostgreSQL datatype OID and the Go values that result from using those.

Previously we were perfectly happy for that data to be missing, just carrying on as best we could without the discovery-type special cases. This was broken as part of the big optimization refactoring, so here I'm streamlining the relevant code to just the values we actually use and then making it not an error if the discovered type is unknown.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3113)


Alex-Bair (approved):
LGTM

PR #3117, opened 2025-07-30
source-mongodb: retries on empty change stream results during catch-up take 2

williamhbaker:
**Description:**

Commit 45884eb was an initial, but flawed attempt to accomplish this, which was reverted while I thought harder about how to do it correctly.

The general idea is that we have observed MongoDB change streams actually returning no results from `TryNext` even if there are more results, so we can't reliably treat the lack of results as a positive indication that the stream is caught up. Retrying a few times over the course of ~seconds seems to reliably get records if there are any though, so when a stream is being "caught up", it needs to be retried a few times if it gets no results, but has not yet observed an event timestamp later than the cluster op time watermark. This is almost exclusively relevant to databases with large existing tables that are backfilling, but no change events being generated from them.

The flaw from the initial attempt was about batch bindings running concurrently with change stream bindings that had completed their backfill. For these, it is important that `coordinator.gotCaughtUp` is reached, so that the backfill / stream coordinator apparatus can know when streams are caught up, and backfills can resume.

This change addresses the need for retries while also making sure `coordinator.gotCaughtUp` is hit.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3117)

### DIFF
--- a/source-mongodb/change_stream.go
+++ b/source-mongodb/change_stream.go
@@ -17,6 +17,10 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+// disableRetriesForTesting can be set `true` when running tests that require
+// timely operation.
+var disableRetriesForTesting bool
+
 // changeStream associates a *mongo.ChangeStream with which database it is for.
 type changeStream struct {
 	ms *mongo.ChangeStream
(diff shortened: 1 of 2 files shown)


williamhbaker:
I'm including a re-apply (revert of a revert) of the additional logging when streaming gets caught up between backfill rounds (original: 092d6b3) in 2501622, since it was actually useful, and not problematic in any way.

JustinASmith (approved):
LGTM

PR #3119, opened 2025-07-31
go/blob/s3: don't use bulk DeleteObjects

williamhbaker:
**Description:**

A bulk DeleteObjects response will close the client connection, since this is some kind of streaming XML API.

We typically don't have an enormous number of objects to delete, so using the single object deletion request with concurrency should be more than sufficient. This is how it has to work for GCS and Azure, since they don't have a bulk deletion API at all.

I have tested and confirmed that when using the single object deletion API, we are able to continue to re-use connections. This should be a big improvement for systems that do transactions at high frequencies, where each transaction requires deleting a single small object. Previously we would need to re-initialize a connection on each transaction, but now we'll re-use it.

**Workflow steps:**

**Notes for reviewers:**

I am also removing the additional logging that was added in cd0ae8d, since that is no longer useful, and never really was.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3119)

### DIFF
--- a/go/blob/s3.go
+++ b/go/blob/s3.go
@@ -8,19 +8,15 @@ import (
 	"iter"
 	"net/http"
 	"path"
-	"slices"
 	"strings"
-	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/aws/retry"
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #3121, opened 2025-07-31
sqlcapture: Prerequisite error reporting tweaks

willdonnelly:
**Description:**

This commit combines a couple of tweaks to the logging and error reporting around database prerequisite checks:

We now log (at DEBUG level) a message immediately after each prerequisite check fails, and immediately before failing the task with a prerequisites error. This should make it clearer exactly what's happening.

The number of distinct prerequisite failures reported to the user is now capped at 20. I have a vague suspicion that it's possible for there to be such a thing as "too many errors", such that trying to report them all at once causes mysterious task failure and "h2 protocol error" messages. And anyway if a user's setup is broken enough to cause >20 errors, they're probably not going to gain much from seeing the rest until they tackle the obvious issues.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3121)

### DIFF
--- a/sqlcapture/main.go
+++ b/sqlcapture/main.go
@@ -103,9 +103,22 @@ type prerequisitesError struct {
 func (e *prerequisitesError) Error() string {
 	var b = new(strings.Builder)
 	fmt.Fprintf(b, "the capture cannot run due to the following error(s):")
-	for _, err := range e.errs {
+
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM

PR #3125, opened 2025-07-31
Revert "source-mysql: Throughput Optimization"

willdonnelly:
Reverts estuary/connectors#3102

There's some "unhandled datatype" bugs related to legacy column type metadata, we'll need to address those types and/or reimplement the analogous fallback handling before trying this again.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3125)


dyaffe (approved):
LGTM

PR #3131, opened 2025-08-01
source-mongodb: experimental timestamp extraction from resume tokens

williamhbaker:
**Description:**

Adds an experimental routine for extracting timestamps from MongoDB resume tokens. I'm fairly confident it will work across all MongoDB versions, but right now I am adding it along with logging that can be enabled by `LOG_LEVEL=debug`. I would like to ensure it works across a sampling of tasks in production before using it for anything else.

The intended uses for this are to eventually:
* Provide a positive indication that a change stream is "caught up", even if it yields no events that have a cluster timestamp with them. We can extract a timestamp even from a post-bast resume token, which represents a time relative to the position of the change stream. This would replace the rather janky retry-based mechanism implement in 13c89ec.
* Similarly, give a better indication of how far behind a change stream might be if the change stream is not regularly yielding events.
* Log out a timestamp when a change stream fails, which could be fed back into a new feature to allow change streams to resume from an op time (specifically, `startAtOperationTime`) for failure recovery. An example of when this would be useful is if the shard configuration for a deployment is changed, which invalidates pre-existing change streams, but they can be resumed based on an operation time. This could further be automated, but is useful as a stand-alone manual feature.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3131)

### DIFF
--- a/source-mongodb/change_stream.go
+++ b/source-mongodb/change_stream.go
@@ -3,6 +3,8 @@ package main
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
+	"encoding/hex"
 	"fmt"
 	"slices"
 	"time"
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM

PR #3134, opened 2025-08-05
source-salesforce-native: always use the Bulk API for formula field refreshes

Alex-Bair:
**Description:**

Previously, the connector did not use the Bulk API if _any_ of an object's fields are complex & can't be queried via the Bulk API. That makes sense when we're trying to query all fields. But when we're only querying formula fields - which should never be any of these complex, non-Bulk API queryable types ([ref](https://help.salesforce.com/s/articleView?id=platform.choosing_a_formula_data_type.htm&type=5)) - that restriction is unecessary. This PR makes it so we always try to use the Bulk API for formula field refreshes.

We've observed formula field refreshes take a long time for `Account` which always has a standard field that's not able to be queried via the Bulk API. This change should help speed up formula field refreshes for `Account` and other bindings in a similar situation.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3134)

### DIFF
--- a/source-salesforce-native/config.yaml
+++ b/source-salesforce-native/config.yaml
@@ -2,7 +2,7 @@ credentials:
     credentials_title: OAuth Credentials
     client_id_sops: ENC[AES256_GCM,data:fMXpKq4TrDG9vbkRFa7EYhOsdk6FtTLLT+jU5hokrHEB2FF8h39CquHiOp0+El0Mg/4zQMTdJ5wDp90WKSPa9yxzaMAi97ExkqfNWPM2FNAK2KmAYg==,iv:8lNVaJITypPnTuUZM4kMXUMdJglGFdkDYGNGhyPsxVU=,tag:yT1ENGiQzGZKM8DCT8gS0g==,type:str]
     client_secret_sops: ENC[AES256_GCM,data:jJgwo0E280bY9JApzBDCXuFaEbnnMMhtT0zYrwybJcOtZc+6RXNkxtQSbXvw9DpAAC6jbGEXMMc6yONx4y21ww==,iv:KqFuPgJzjVgDdOdKGRrnvEda/ZUUBvvlBjUcEdiALeg=,tag:M2RoEtnWixP86GOv/uwMIw==,type:str]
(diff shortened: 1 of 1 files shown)


source-salesforce-native/config.yaml
```diff
@@ -2,7 +2,7 @@ credentials:
     credentials_title: OAuth Credentials
     client_id_sops: ENC[AES256_GCM,data:fMXpKq4TrDG9vbkRFa7EYhOsdk6FtTLLT+jU5hokrHEB2FF8h39CquHiOp0+El0Mg/4zQMTdJ5wDp90WKSPa9yxzaMAi97ExkqfNWPM2FNAK2KmAYg==,iv:8lNVaJITypPnTuUZM4kMXUMdJglGFdkDYGNGhyPsxVU=,tag:yT1ENGiQzGZKM8DCT8gS0g==,type:str]
     client_secret_sops: ENC[AES256_GCM,data:jJgwo0E280bY9JApzBDCXuFaEbnnMMhtT0zYrwybJcOtZc+6RXNkxtQSbXvw9DpAAC6jbGEXMMc6yONx4y21ww==,iv:KqFuPgJzjVgDdOdKGRrnvEda/ZUUBvvlBjUcEdiALeg=,tag:M2RoEtnWixP86GOv/uwMIw==,type:str]
-    refresh_token_sops: ENC[AES256_GCM,data:vqvATUauWsPBTFv1wTO7DljL4zOADONNj7+euRIQkXlB6nQgFIX11uH6tFrhhThAT99x5ipqRSxzzRgbGb/bd++AybniGSIy/Cw6AvPUDvDYAJ8yRX94,iv:4gQ10k3Z3oZ2/pKrP530yWpaha16KznbwXJilLmJGO0=,tag:VvwHjDwFNRkPeBGauLtajw==,type:str]
+    refresh_token_sops: ENC[AES256_GCM,data:quOxeAYT/e7YzQzsfdyUP09MpGgj1BD2ejUI5v4QQU4ULFvqryLBGipdknAC/KqF43bny1D9aY+mg72E86/sDm65WPVhFO6Uj6gdTIQoYQihO+VvwWjE,iv:XK5WiTA1rhRfuzFfdxuruZHjVvuu7xXq8Me/Myx4ueI=,tag:YbCxgijpPMvh6f40ND5NMA==,type:str]
```
JustinASmith, line 5: Curious if this was an expected change? If so, please ignore. Just pointing it out just in case.

JustinASmith (approved):
LGTM

source-salesforce-native/config.yaml
Alex-Bair: Yep! This was expected. I revoked the previous refresh token when I was trying to replicate an issue with our test Salesforce account, and that ended up breaking the tests. I had to update `config.yaml` to contain a valid refresh token for our tests to pass again.

PR #3141, opened 2025-08-07
source-sqlserver: Fix shutdown deadlock

willdonnelly:
**Description:**

In two places we were sending commit events directly to the events channel instead of using the emitEvent helper function which does the necessary context-cancellation logic, so it was possible for a task to deadlock with the replication thread blocked sending an event and the main thread blocked waiting for it to shut down.


Alex-Bair (approved):
LGTM

PR #3144, opened 2025-08-07
source-google-play: fix `reviews` incremental strategy

Alex-Bair:
**Description:**

An assumption underpinning the incremental replication strategy for `reviews` was that the CSVs were only ever updated when individual `reviews` were updated in the file. This assumption doesn't hold - for one reason or another, these CSVs' `updated` metadata can be moved forward in time without any reviews being modified.

This PR switches from using a file's `updated` timestamp to an individual review's `updated_at` as the cursor value to prevent missing reviews when CSV files are updated without actual review changes.

The decoupling of the files' `updated` and the individual reviews' `updated_at` is difficult to manage with a single cursor. The solution implemented in this PR will end up checking files that haven't been updated since the previous sweep for updated reviews since we aren't storing the latest file's `updated` metadata any longer. We won't yield any duplicate data, but we'll spend some extra time each sweep checking the same CSVs for updated reviews.

A more elegant solution would use a complex cursor to track _both_ the latest file's `updated` and review's `updated_at` and use both of them in concert to only check files updated since the last sweep _and_ only yield reviews that were updated since the last review we yielded. That'd require some changes to the CDK to support more complex `LogCursor`s.

I'm choosing to not make those CDK changes right now and see how this simpler strategy fairs. Based on my knowledge of these Google Play reports so far, I don't anticipate the additional CDK work is worth the effort to optimize this connector. There won't be many CSVs to process each sweep, the CSVs don't contain tons of rows, and the bindings have a default 1 hour interval; the connector won't be checking for updates very frequently, and when it does, there still won't be a ton of work to do.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that `reviews`' incremental cursor is based off individual reviews' `updated_at` rather than the containing CSV's `updated` metadata.

### DIFF
--- a/source-google-play/source_google_play/api.py
+++ b/source-google-play/source_google_play/api.py
@@ -7,18 +7,19 @@ from estuary_cdk.capture.common import LogCursor, PageCursor
 from .models import (
     GooglePlayRow,
     Reviews,
+    Statistics,
 )
 
 from .gcs import GCSClient, GCSFileMetadata
 from .shared import next_month, dt_to_str, str_to_dt
 
 
-async def fetch_resources(
+async def fetch_statistics(
     gcs_client: GCSClient,
-    model: type[GooglePlayRow],
+    model: type[Statistics],
     log: Logger,
     log_cursor: LogCursor,
-) -> AsyncGenerator[GooglePlayRow | LogCursor, None]:
+) -> AsyncGenerator[Statistics | LogCursor, None]:
     assert isinstance(log_cursor, datetime)
 
     files: list[GCSFileMetadata] = []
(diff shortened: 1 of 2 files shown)


source-google-play/source_google_play/api.py
```diff
@@ -32,21 +33,42 @@ async def fetch_resources(
...
+    gcs_client: GCSClient,
+    log: Logger,
+    log_cursor: LogCursor,
+) -> AsyncGenerator[Reviews | LogCursor, None]:
+    assert isinstance(log_cursor, datetime)
+
+    files: list[GCSFileMetadata] = []
+    async for file in gcs_client.list_files(prefix=Reviews.prefix, globPattern=Reviews.get_glob_pattern()):
+        if file.updated >= log_cursor:
+            files.append(file)
```
JustinASmith, line 53: Does the file `updated` field changing (moving forward) even when `reviews` do not change cause concern with this check? Can the `updated` field be unreliable and cause us to not consider a file if it was for some reason not changed at all even when `reviews` change? I am assuming that wouldn't happen, but wasn't sure based on your PR description.

Alex-Bair: I plan to continue checking the CSVs to better understand how & when they get updated over the next week. I may change the incremental strategy for `reviews` again based on what I observe, but I wanted to get a fix up now to avoid missing data in the interim.

JustinASmith (approved):
LGTM

PR #3156, opened 2025-08-12
source-monday: add newly observed activity log events

JustinASmith:
**Description:**

Observed new activity log events in production captures. This adds those to the model for processing.

Events and how they will  be handled:

- `DELETE_GROUP_PULSE` -  occurs when a group is deleted. Each item in the group will have this as an individual activity log entry to indicate that the item was deleted as part of the group deletion.
- `DELETE_GROUP` - occurs when a group is deleted. Will be a single activity log event that will be handled separately from the above event to refresh the `board` that was updated.
- `RESTORE_GROUP` - occurs if the group is restored. Will require a `board` fetch and all the board's `items` fetched since there is not an event that gives us the `item` IDs that were restored. Behaves the same as `RESTORE_COLUMN` and others.
- `RESTORE_COLUMN` - occurs when a column is deleted and later restored. Will only be a single activity log event. Similar to `RESTORE_GROUP`; we will need to fetch the `board` for the board's incremental task and all `items` for that board in the `items` incremental task.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3156)

### DIFF
--- a/source-monday/source_monday/models.py
+++ b/source-monday/source_monday/models.py
@@ -191,6 +191,7 @@ class ActivityLogEvents(StrEnum):
     CREATE_PULSE = "create_pulse"
     DELETE_COLUMN = "delete_column"
     DELETE_GROUP = "delete_group"
+    DELETE_GROUP_PULSE = "delete_group_pulse"
     DELETE_PULSE = "delete_pulse"
     MOVE_PULSE_FROM_BOARD = "move_pulse_from_board"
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM! Thanks for the describing what each event means & the action we'll take when we see them, that's helpful.

PR #3164, opened 2025-08-13
materialize-snowflake: load documents as compact JSON

williamhbaker:
**Description:**

Querying the VARIANT document column directly was actually returning results through the Go driver as pretty-printed JSON, which has a lot of extra whitespace. In extreme cases, this could cause the documents to exceed gRPC message size limits. In more common cases it is just wasteful.

Using the TO_JSON function causes these documents to be returned in their compact JSON representation instead.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3164)

### DIFF
--- a/materialize-snowflake/.snapshots/TestSQLGeneration
+++ b/materialize-snowflake/.snapshots/TestSQLGeneration
@@ -95,7 +95,7 @@ ALTER TABLE "a-schema".key_value ALTER COLUMN
 --- End alter table drop not nulls ---
 
 --- Begin "a-schema".key_value loadQuery ---
-SELECT 0, "a-schema".key_value.flow_document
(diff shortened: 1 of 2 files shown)


williamhbaker:
> LGTM
> 
> Would using TO_JSON on other VARIANT columns that we know should be JSON help reduce extra whitespace further? I'm not sure if the materialization is currently able to determine which VARIANT columns contain JSON though, and using TO_JSON on only `flow_document` likely removes most of the extra whitespace.

The only columns that materializations currently query is the document column, as part of their load queries. We've got plans to make it so materializations can do standard updates without needing a separate document column by selecting & querying all top-level fields, and when that happens using TO_JSON on all of the VARIANT columns will be a good idea.

PR #3175, opened 2025-08-14
materialize-snowflake: log failed blob metadata if a streaming blob fails to be written

williamhbaker:
**Description:**

Log failed blob metadata if a streaming blob fails to be written.

We try to rename a blob if it errors on a "malformed request" because that is usually just that the blob is too old and needs a new file name generated, but if it still fails after that we need to know more about the metadata for debugging.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3175)


JustinASmith (approved):
LGTM

PR #3178, opened 2025-08-15
source-pendo: delay event & event aggregates incremental tasks

Alex-Bair:
**Description:**

I noted in 6d90c5f that there are delays between when events occur & when they are available via the Pendo API. I previously assumed events appear sequentially & we don't need to be worried about the API being eventually consistent. Well, we've found evidence that the Pendo API is eventually consistent with its events related endpoints and need to handle it.

This PR limits how recent the data the events & event aggregates incremental tasks capture; if data is beyond the eventual consistency horizon defined by `API_EVENTS_LAG`, then it doesn't get returned by the Pendo API & the connector doesn't see or capture it.

I reduced the `API_EVENTS_LAG` from 12 hours to 8 hours as well. 12 hours was too conservative, but it didn't have an ongoing impact since it was only used to shift the cutoff between backfill & incremental tasks. I previously observed events showing up between 4-6 hours late and my most recent investigation aligns with that, so a lag of 8 hours seems sufficient.

**Workflow steps:**

**Notes for reviewers:**

Tested via API calls & on a local stack. Confirmed:
- The Pendo API does not return events / event aggregates beyond `upper_bound` when it's specified.
- The events & event aggregate streams capture the missing data when retreading earlier date ranges.
  - This tells me that the incremental & backfill strategies are fine - the missing data is due to eventual consistency challenges, not that the capturing strategy is flawed.
- The event & event aggregate streams do not capture data more recent than the `horizon`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3178)

### DIFF
--- a/source-pendo/source_pendo/api.py
+++ b/source-pendo/source_pendo/api.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta, UTC
 from logging import Logger
-from typing import AsyncGenerator
+from typing import Any, AsyncGenerator
 import pendulum
 from pydantic import TypeAdapter
 from estuary_cdk.capture.common import BaseDocument, LogCursor, PageCursor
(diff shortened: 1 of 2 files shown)


source-pendo/source_pendo/api.py
```diff
@@ -52,6 +57,8 @@ def generate_events_body(
         filter_condition = f"guideTimestamp == {lower_bound} && {identifying_field} >= \"{last_seen_id}\""
     else:
         filter_condition = f"guideTimestamp >= {lower_bound}"
+        if upper_bound:
+            filter_condition += f" && guideTimestamp <= {upper_bound}"
```
Alex-Bair, line 61: Although we do specify the `upper_bound` `timeSeries.last` field, that alone does not ensure the response does not contain data later than `upper_bound`. A relevant example in Pendo's [docs](https://engageapi.pendo.io/#3f910cb9-e159-4ddf-a40f-75956b647d06) say:
>With a period of hourRange and epoch timestamps of 1693540800000 (September 1st, 2023 12:00:00 AM ET) and 1693932275771 (September 5th, 2023 12:44:35 PM ET), this will returns all events from midnight of September 1st to 12:59:59 PM ET of September 5th (i.e. the end of the 12PM hour).

This extra filter condition is required to ensure Pendo doesn't send back data generated after `upper_bound`.

JustinASmith (approved):
LGTM

PR #3186, opened 2025-08-18
source-pendo: refactor & use IncrementalJSONProcessor

Alex-Bair:
**Description:**

`source-pendo` was the first native connector I built, and it doesn't follow a lot of the more standard conventions more recently built connectors follow. It also occasionally has OOM failures when backfilling the various event & event aggregates bindings.

This PR's scope includes:
- Refactoring `source-pendo` to bring it more in line with the conventions followed by other native connectors & to make maintaining the connector easier.
  - The refactors are all in the first three commits. No functional changes were made in these commits.
- Using the `IncrementalJSONProcessor` to reduce memory usage & avoid OOMs.

I'd recommend reviewing commit-by-commit. It looks like there's a lot of changes, but most of it is re-organization of the existing code. The only functional change is the switch to use the `IncrementalJSONProcessor`.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed all streams still work as they did before, and that the connector no longer OOMs when backfilling multiple streams (memory usage hovered around ~12% during the backfills).

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3186)

### DIFF
--- a/source-pendo/source_pendo/api.py
+++ b/source-pendo/source_pendo/api.py
@@ -280,30 +280,21 @@ async def snapshot_metadata(
     yield metadata
 
 
-async def fetch_events(
-        http: HTTPSession,
-        entity: str,
-        model: type[Event],
-        identifying_field: str,
-        log: Logger,
-        log_cursor: LogCursor,
-) -> AsyncGenerator[Event | LogCursor, None]:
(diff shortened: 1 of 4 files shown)


JustinASmith (approved):
LGTM!

PR #3190, opened 2025-08-18
source-mysql: Throughput optimization, take two

willdonnelly:
**Description:**

This is attempt number two at merging the MySQL throughput optimization fixes. It's identical to the original PR (https://github.com/estuary/connectors/pull/3102) which had to be reverted, except for the final commit which adds a default transcoder to be used when none of the explicit cases are matched. This should precisely reproduce the original fallthrough behavior when none of the explicit cases in the field value translation function matched, and I've added logging so that we'll be able to grep prod task logs after a week or so and see what cases are missing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3190)


williamhbaker (approved):
LGTM

PR #3198, opened 2025-08-20
source-sqlserver: Fix watermark fencing

willdonnelly:
**Description:**

Apparently the throughput optimization PR broke watermark fences about a month ago, but it took a while to notice because:
1. Very few people are still using watermark fences
2. If we fail to identify the watermark fence, the initial catchup streaming phase just continues forever, and that still works fine as long as you just need to stay caught up on replication. Things only actually break if you need to backfill or add a table.

The root of this bug is that I accidentally used a list of column names which is in _serialized document order_ as though it were in _database column order_ leading us to consult the wrong column value for the actual check. This should fix it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3198)


Alex-Bair (approved):
LGTM

PR #3205, opened 2025-08-20
materialize-google-pubsub: key fields should be OPTIONAL

jgraettinger:
**Description:**

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3205)

### DIFF
--- a/materialize-google-pubsub/driver.go
+++ b/materialize-google-pubsub/driver.go
@@ -151,6 +151,8 @@ func (d driver) Validate(ctx context.Context, req *pm.Request_Validate) (*pm.Res
 			case projection.IsRootDocumentProjection():
 				constraint.Type = pm.Response_Validated_Constraint_LOCATION_REQUIRED
 				constraint.Reason = "The root document must be materialized"
+			case projection.IsPrimaryKey:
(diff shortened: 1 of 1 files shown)


williamhbaker (commented):
This seems fine, although I'm actually not entirely sure how it will work / ever did work.

The materialization actually uses the packed key for publishing messages: https://github.com/estuary/connectors/blob/4f733655e7c596d815f6a50bf2127d6e5cccbf1f/materialize-google-pubsub/transactor.go#L58

If the collection keys were previously forbidden, I guess the packed key should have been empty? Or perhaps that packed key part disregards field selection...or maybe it used to and won't going forward?

I'm just a little confused, since I think the connector _did_ actually work before, but now I don't know how it did.

Another angle - from what I understand, the selection of key fields for materialization determines how documents are reduced. So selecting no key fields should have had the effect of reducing every transaction into a single document(?). If that is true, then I'd expect key fields should be recommended instead of optional. I am somewhat fuzzy on the semantics of constraints with the new field selection logic though, and vaguely recall that there may not be a different between the two going forward in actuality.

jgraettinger:
That's right, recommended is not a thing for constraints anymore

williamhbaker (approved):

PR #3208, opened 2025-08-21
source-hubspot-native: allow extra fields in responses

Alex-Bair:
**Description:**

This PR is a follow up to https://github.com/estuary/connectors/pull/3207. Instead of reactively changing the response models to allow extra fields whenever Hubspot decides to add a few fields, I'd rather proactively prevent connectors from crashing due to extra fields.

There are still a couple `extra="forbid"`s left in some models, but most are for documents that get emitted, and it seems like we would actually care if Hubspot adds extra fields to those.

**Workflow steps:**

**Notes for reviewers:**


williamhbaker (approved):
LGTM

PR #3210, opened 2025-08-21
source-mysql: Truncate logged RowsQueryEvent

willdonnelly:
**Description:**

Apparently there is no limit on the worst-case size of these queries. They can reach hundreds of megabytes. We can't easily stop the client library from reading them into memory, but we can avoid compounding it with a bytes->string cast of the full value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3210)


Alex-Bair (approved):
LGTM

PR #3211, opened 2025-08-21
estuary-cdk: remove default `_meta` value from document schemas

Alex-Bair:
**Description:**

The CDK's `BaseDocument` was schematizing a default value for the `_meta` field, but Pydantic doesn't add default values when the `exclude_unset` argument is set to `True` during serialization. This caused issues with Dekaf, where the schema said a field would exist, but documents didn't actually contain that field.

Should the CDK be injecting the default `_meta` value? I think that was the original intention, but I'm holding off on changing that behavior before confirming with the broader group.

**Workflow steps:**

**Notes for reviewers:**

I also updated recently failing snapshots to ensure every connector gets built with the new "don't schematize a default `_meta` value` behavior.


jgraettinger (approved):
LGTM

PR #3221, opened 2025-08-26
source-klaviyo-native: add worker health check to prevent deadlocks

Alex-Bair:
**Description:**

I noticed a production capture was deadlocked in the `events` backfill with no progress for ~8 hours. Injecting a SIGQUIT showed that the subdivision worker, shutdown monitor, and 4 out of 5 chunk worker tasks were still running. One chunk worker task was not included in the logs after the SIGQUIT, meaning it was no longer running. I suspect the deadlock occurred from this chunk worker exiting without decrementing the active worker count, and this prevented the shutdown monitor from detecting that all work was done.

I'm still investigating _how_ that chunk worker could have exited without decrementing the active worker count or raising an exception that propagates through the `TaskGroup`. In the meantime, implementing a health check that fails if any workers exit before the shutdown event is set should prevent these types of deadlocks.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack by forcing a chunk worker to exit after 15 seconds. Confirmed that the health check fails & crashes the connector when a worker exits before the shutdown event is set.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3221)

### DIFF
--- a/source-klaviyo-native/source_klaviyo_native/api/events.py
+++ b/source-klaviyo-native/source_klaviyo_native/api/events.py
@@ -153,6 +153,33 @@ class WorkManager:
     def are_active_workers(self) -> bool:
         return self._active_worker_count > 0
 
+    def worker_health_check(self) -> None:
+        alive_workers = [task for task in self.worker_tasks if not task.done()]
+        expected_count = len(self.worker_tasks)
+        actual_count = len(alive_workers)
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM as a temporary mitigation

PR #3223, opened 2025-08-26
source-postgres: Sanitize unpaired surrogates in JSON

willdonnelly:
**Description:**

PostgreSQL allows invalid `\uXXXX` escapes inside `JSON` column values and Go doesn't normalize those out when serializing a `json.RawMessage` (even though it does a bunch of other normalization), and sadly Rust's serde_json parser _will_ reject such documents. So we have to add a whole thing here just to transform those into U+FFFD replacements.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3223)


williamhbaker (approved):
LGTM

PR #3230, opened 2025-08-27
materialize-snowflake: start new streaming blob based on time as well as size

williamhbaker:
**Description:**

I have theorized that certain scenarios could arise where we'd be streaming into the same blob file for a very long time, without exceeding its size limit. Such a case may arise from data that is able to be highly compressed by parquet run-length encoding, for example.

In a case like this the AWS/etc. credentials could expire in the midst of a streaming upload.

I think we might actually be seeing this happen in practice, so this adds an additional condition to starting a new blob based on credential expiration time, in addition to blob size.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3230)

### DIFF
--- a/materialize-snowflake/stream.go
+++ b/materialize-snowflake/stream.go
@@ -114,7 +114,9 @@ func (sm *streamManager) writeRow(ctx context.Context, binding int, row []any) e
 		return fmt.Errorf("bdecWriter writeRow: %w", err)
 	}
 
-	if sm.bdecWriter.done {
+	// Start a new blob based on the size of the current blob, or expiration of
+	// the blob storage bucket credentials.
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

Does the connector get a fresh set of valid credentials after finishing a blob?

williamhbaker:
> LGTM
> 
> Does the connector get a fresh set of valid credentials after finishing a blob?

Yep, here: https://github.com/estuary/connectors/blob/843e1f0d479f6cc2dad1e1425447d9eb8da409f8/materialize-snowflake/stream.go#L141

PR #3231, opened 2025-08-27
source-salesforce-native: add more supported standard objects

Alex-Bair:
**Description:**

This PR contains a lot of additions to and two removals from `supported_standard_objects.py`.

The additions were generated by checking which standard objects with cursor fields we've seen returned in `/describe` and `/describe/:sobject/sobject` API calls now. I scripted the generation of these, but I also spot checked a handful of them to confirm the cursor fields are correct & we can query them via the REST API.

The removals were driven by users trying to capture these objects but getting an error stating that we have to add a `WHERE ParentId = 'some_parent_id'` clause. The connector doesn't support capturing child objects yet, so removing these from the list of supported standard objects will help avoid these errors. We can look into adding these objects back once the connector supports child objects.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed the new objects are discovered & that the connector works like before.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3231)

### DIFF
--- a/source-salesforce-native/source_salesforce_native/supported_standard_objects.py
+++ b/source-salesforce-native/source_salesforce_native/supported_standard_objects.py
@@ -54,6 +54,12 @@ SUPPORTED_STANDARD_OBJECTS: dict[str, ObjectDetails] = {
     "AIRecordInsight": {
         "cursor_field": CursorFields.SYSTEM_MODSTAMP
     },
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM!

PR #3236, opened 2025-08-28
source-postgres: Fix a bug in the unpaired-surrogates logic

willdonnelly:
**Description:**

When I wrote this logic yesterday I forgot to account for escapes which were not \uXXXX and so it was possible to write something like `"\\a\\b\\ufoobar"` and we'd accidentally try to decode the `\ufoob` portion as a Unicode escape.

All escapes other than `\uXXXX` are two bytes long so it's fairly simple to fix the logic so it handles them correctly, and I've also gone ahead and tweaked the hex decode function so that the presence of a non-hex digit will immediately result in a 0 return value which doesn't match the codepoint ranges we're checking for.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3236)


jgraettinger (approved):
lgtm

PR #3238, opened 2025-08-28
source-klaviyo-native: fix TaskGroup deadlock

Alex-Bair:
**Description:**

`source-klaviyo-native` captures have still deadlocked after merging https://github.com/estuary/connectors/pull/3221. After injected a `SIGQUIT` for these deadlocked captures, it was a very similar story. One or more chunk worker tasks were no longer running. However, the shutdown monitor task was also not running. The _only_ way that could exit without the shutdown event being set is if it raised an exception. This made me _heavily_ suspicious that something was preventing the `TaskGroup` from handling and propagating exceptions.

The issue was that I had synchronous code (streaming documents from a queue and processing them) within the `TaskGroup`'s execution context. When a `TaskGroup`'s task failed, the `TaskGroup` tried to cancel all tasks, but it couldn't complete it's cleanup because the document processing was blocking.

To fix this, the document processing work has been moved out of the `TaskGroup`'s execution context. A new task within the `TaskGroup` now puts documents and exceptions into a queue, and `backfill_events` consumes from that queue to yield documents or raise exceptions.

Additionally, I reverted the health check I added earlier in 4e3662f8939acbc1f43fbeda23170f985b1d1e06 since it didn't work & it's no longer needed.

**Workflow steps:**

**Notes for reviewers:**

- The happy path works like it did before - `events` records are fetched for a datetime range, yielded, then checkpointed in a single `backfill_events` invocation.
- An exception that occurs in a TaskGroup's task is propagated up into `backfill_events` and crashes the connector.


williamhbaker (approved):
LGTM

PR #3251, opened 2025-09-02
materialize-snowflake: fix bdec writing when column chunks are numerous

williamhbaker:
**Description:**

A bdec blob can only have a single row group. The previous logic was flawed because it assumed that row group creation was directed only by the bytes contained in the row group, but there is a secondary control where if a row group has a huge number of column chunks (e.g. because there are many selected fields), row groups are written more frequently to manage connector memory.

This implements a more direct accounting, where the number of row groups written is made available via the parquet writer, and the bdec blob can be closed out immediately after the first row group is written.

I also fixed a minor problem where the uncompressed length of the bdec blob was being set incorrectly. It is now taken directly from the reported metadata of the parquet file, once the file is closed.

**Workflow steps:**

**Notes for reviewers:**

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3251)

### DIFF
--- a/go/writer/parquet.go
+++ b/go/writer/parquet.go
@@ -138,11 +138,12 @@ type parquetConfig struct {
 type ParquetWriter struct {
 	cfg parquetConfig
 
-	schema     ParquetSchema
-	schemaRoot *schema.GroupNode
-	sinkWriter *file.Writer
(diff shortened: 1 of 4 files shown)


Alex-Bair (approved):
LGTM

PR #3252, opened 2025-09-02
source-shopify-native: move simple streams to use non-bulk queries

Alex-Bair:
**Description:**

- Introducing an async client to interact with Shopify's GraphQL API.
- Refactoring the bulk job manager to use the async client.
- Adding support for using non-bulk queries for simple streams (i.e. streams whose queries do not have nested connections).
  - Capturing _all_ data in nested connections is tricky & I anticipate doing so with non-bulk queries consumes a significant amount of the complexity limit. Bulk queries simplify this handling by fetching all nested connections for a query, and I anticipate they're faster than paginating through the nested connections for each record with separate API requests.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed the documents output by the connector are the same as before, and that streams with zero nested connections use non-bulk queries.


JustinASmith (approved):
I think this looks good to me! Curious how much faster you think the connector will be able to fetch data with this approach? Seems the async client, first/after params, and non-bulk queries should help a little bit.

Alex-Bair:
Without running some performance tests using a Shopify account with a lot of data, it's tough to quantify how much faster we'll be able to fetch data by moving streams without nested connections to non-bulk queries. I didn't run those performance tests since Shopify only allows a single bulk query to be processed at a time, and those tests would fight with other applications over who gets to run that bulk query.

I'll still write out my thoughts around _why_ this should speed up the connector though:
- 7 fewer streams will be taking turns making bulk queries. That's ~29% fewer streams making bulk queries, so a naive estimate is that the remaining streams making bulk queries will be ~29% faster. But that's not quite accurate since a bulk query for a small stream (ex: `locations`) take Shopify less time to process than a bulk query for a large stream (ex: `customers`). We _are_ moving some of those large streams over to non-bulk queries, so I'm hopeful we'll see a noticeable improvement.
- The streams using non-bulk queries run concurrently with whichever stream is making a bulk query. Testing locally, I saw that even these 7 streams used enough cost points to get throttled by the Shopify API every so often. So they won't be blazingly fast, but they still capture data much faster than when they were waiting to submit bulk queries.

After seeing how much these changes speed things up, it may be worthwhile to investigate how we would move streams with a single nested connection over to non-bulk queries. I anticipate that wouldn't be trivial (how do we paginate through each nested connection? how could we generalize that?), but that _may_ speed the connector up a bit more.

PR #3255, opened 2025-09-04
source-postgres-batch: Make /_meta/source not required

willdonnelly:
**Description:**

The /_meta/source property is newly added and while it's accurate to describe it as required in the write schema, that required-ness seems to be making its way into the read schema and causing trouble and the simple fix is to just not say it's required anywhere for now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/estuary/connectors/3255)


willdonnelly:
This has been addressed by updating all batch SQL output collections which were still using the old `flow://write-schema` in their read schema definition to instead use `flow://relaxed-write-schema`

PR #3265, opened 2025-09-05
source-klaviyo-native: handle worker cancellation to avoid deadlocks

Alex-Bair:
**Description:**

#### Summary / tl;dr

This PR prevents the `events` backfill from deadlocking when a worker task is cancelled. Workers should never be cancelled during normal operation (the shutdown monitor and shutdown event coordinate when tasks should exit naturally), and a non-`TaskGroup` initiated cancellation is an exceptional condition that should stop processing and crash the connector. This PR does that by implementing `CancelledError` handling for workers.

#### Investigation

I observed `source-klaviyo-native` captures getting deadlocked in the `events` backfill. Previous PRs to troubleshoot/address the deadlock were:
- https://github.com/estuary/connectors/pull/3221
- https://github.com/estuary/connectors/pull/3238
- https://github.com/estuary/connectors/pull/3253

After the last PR, I saw a log like "Task chunk_worker_1 cancelled before the shutdown event was set." surrounded by "Connection timeout error (will retry)" logs when a connector deadlocked. After some digging, I discovered that when making a request with `aiohttp` it [keeps a reference](https://github.com/aio-libs/aiohttp/blob/33de07852848eb51e82ad4132bb7038b27c348ad/aiohttp/helpers.py#L673) to the currently running task and [cancels it](https://github.com/aio-libs/aiohttp/blob/33de07852848eb51e82ad4132bb7038b27c348ad/aiohttp/helpers.py#L713-L718) when the `aiohttp.ClientSession`'s total timeout is exceeded. `aiohttp` converts the resulting `CancelledError` into a `TimeoutError` when exiting the timer's context manager, but it's possible for that conversion to _not_ occur (like when the connector is outside of that context manager and streaming the response body). When that conversion doesn't happen, a `CancelledError` is raised in the worker task. The `TaskGroup` does not cancel other tasks if a task fails with a `CancelledError`, and the backfill ends up deadlocked since the connector is waiting for the cancelled worker to finish its work.

To prevent the deadlock from happening, the first cancelled worker converts the `CancelledError` into an `Exception` that _will_ cause the `TaskGroup` to cancel the other child tasks and bubble up the `Exception` through the connector. Instead of a deadlock, the connector crashes then gets restarted.

**Workflow steps:**

**Notes for reviewers:**

I was able to replicate the original deadlock by reducing the `aiohttp.ClientSession`'s total timeout in `http.py` from 5 minutes (the default) to ~0.75 seconds. I also monkey patched in some logging to `aiohttp.helpers.TimerContext.timeout` to confirm when certain tasks were getting cancelled. That patch looked like:
```python
# Log when aiohttp cancels tasks
# Add this to the top of http.py after imports
import aiohttp.helpers
from logging import getLogger

# Store original method
_original_timeout = aiohttp.helpers.TimerContext.timeout

def logged_timeout(self):
    """Wrapped timeout to log task cancellations"""
    if not self._cancelled and self._tasks:
        log = getLogger(__name__)
        log.warning("aiohttp TimerContext.timeout() about to cancel tasks", {
            "task_count": len(self._tasks),
            "unique_tasks": len(set(self._tasks)),
            "tasks": [
                {
                    "task_name": task.get_name() if hasattr(task, 'get_name') else str(task),
                    "task_done": task.done(),
                    "task_cancelled": task.cancelled() if hasattr(task, 'cancelled') else False,
                }
                for task in self._tasks
            ]
        })

    # Always call original timeout function (no manual cancellation)
    return _original_timeout(self)

# Apply the wrapper
aiohttp.helpers.TimerContext.timeout = logged_timeout
```

I tested these changes on a local stack and confirmed that instead of deadlocking, the connector crashes when an `events` backfill worker task is cancelled.

I'm also leaving in a lot of the logging I added during troubleshooting just in case some other deadlock mechanism exists. I'm confident this PR will address the current deadlock mechanism, but I was also somewhat confident with my earlier attempts and I'd like to keep the logging in for now until I can confirm that production tasks are no longer deadlocking.

### DIFF
--- a/source-klaviyo-native/source_klaviyo_native/api/events.py
+++ b/source-klaviyo-native/source_klaviyo_native/api/events.py
@@ -1,7 +1,7 @@
 import asyncio
 import time
 import traceback
-from asyncio import Queue
+from asyncio import CancelledError, Queue
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from logging import Logger
@@ -9,6 +9,7 @@ from typing import AsyncGenerator, Optional
 
 from estuary_cdk.capture.common import LogCursor, PageCursor
 from estuary_cdk.http import HTTPSession
+from estuary_cdk.utils import format_error_message
 
 from ..models import Events, IncrementalStream, LowerBoundOperator, UpperBoundOperator
 from ..shared import dt_to_str, split_date_window, str_to_dt
@@ -129,6 +130,7 @@ class WorkManager:
         self.shutdown_monitor_task: Optional[asyncio.Task] = None
 
         self._active_worker_count = 0
+        self.first_worker_error: str | None = None
 
     def mark_worker_active(self) -> None:
         if self._active_worker_count >= DEFAULT_BACKFILL_CONFIG.num_event_workers:
@@ -160,7 +162,7 @@ class WorkManager:
             if task.cancelled():
                 status = "cancelled"
             elif exc:= task.exception():
-                status = f"failed with exception {exc}"
+                status = f"failed with exception {format_error_message(exc)}"
                 stack_frames = task.get_stack()
                 if stack_frames:
                     stack_trace = "\nStack trace:\n" + "".join(traceback.format_list(traceback.extract_tb(exc.__traceback__)))
@@ -171,6 +173,7 @@ class WorkManager:
                 self.log.debug(f"Task {task_name} {status} after the shutdown event was set.")
             else:
                 self.log.error(f"Task {task_name} {status} before the shutdown event was set.", {
+                    "first_worker_error": self.first_worker_error,
                     "active_worker_count": self._active_worker_count,
                     "stack trace": stack_trace,
                 })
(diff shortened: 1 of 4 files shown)


source-klaviyo-native/source_klaviyo_native/api/events.py
```diff
@@ -245,31 +248,31 @@ async def chunk_worker(
...
-    is_looking_for_work = False
-    while not shutdown_event.is_set():
-        chunk = await queue_manager.get_work_chunk(timeout=DEFAULT_BACKFILL_CONFIG.worker_timeout)
-        if chunk is None:
-            if not is_looking_for_work:
-                log.debug(f"Event worker {worker_id} is idle and checking for new work.")
-                is_looking_for_work = True
-            # No work is available. Check the shutdown_event and then check for more work.
-            continue
+    try:
```
Alex-Bair, line 251: This diff is a bit messy. The change here was wrapping the entire function body in a `try/except` and the new handling in the `except` blocks.

williamhbaker (approved):
LGTM

Issue #3240, opened 2025-08-28
source-*: Source tag metadata


willdonnelly:
Just writing up the takeaways from today's post-standup discussion before I forget.

The basic idea is that users with complex multi-database setups often want some way to combine data while still keeping track of where each document came from. For SQL Server and PostgreSQL this sometimes takes the form of wanting the database name to be included in the document, while MySQL (which doesn't have logical databases) it gets a bit more complicated.

The design we settled on is to have an optional **Source Tag** property in the endpoint advanced config. This property will live at `/advanced/source_tag` for any connector which supports it, and when set to a non-empty value that value will be included in each output document as `/_meta/source/tag`. When unset the `/_meta/source/tag` value will be omitted, so this is inherently backwards compatible. This will be our recommended way to get this sort of tagging functionality on PostgreSQL, MySQL, or SQL Server.

We do eventually want to better support "cookie-cutter" setups where users have multiple databases with similar contents and want us to capture from all of them. If/when we get around to that we may want to _also_ add a `/_meta/source/database` property representing the particular database name within that cookie-cutter capture setup, but for now we're going to only have `/_meta/source/tag`.

TL;DR: If you have one task per entity then you should use `/_meta/source/tag` to uniquely identify them with whatever string makes sense for your use-case. If/when we support automatic multi-database ("cookie-cutter") setups we may add an automatic `/_meta/source/database` property to facilitate that.

PR #3269, opened 2025-09-09
source-redshift-batch: Add /_meta/source and 'Source Tag' config

willdonnelly:
**Description:**

A copy-paste of https://github.com/estuary/connectors/pull/3245 for batch Redshift.

This PR adds the "Source Tag" advanced configuration property (https://github.com/estuary/connectors/issues/3240) to `source-redshift-batch`. This is a bit more complicated than it was for the SQL CDC connectors because we didn't have a `/_meta/source` object to add it to, so this PR has to add that as well and gives it a few other possibly-useful values while we're at it.

**Workflow steps:**

No action is required unless the user needs a "Source Tag" property set. If such a property is needed, the user should set it as their `/advanced/source_tag` endpoint config property and it will be added to all documents as `/_meta/source/tag`.

**Notes for reviewers:**

This is really a very straightforward copy-paste of the `source-postgres-batch` implementation of this stuff (https://github.com/estuary/connectors/pull/3245). I've divided it into two commits again, one which has the actual important bits and another which just updates all preexisting test snapshots to have the new `/_meta/source` object.


willdonnelly:
> I see the build hung for this connector. This is probably a known issue?

Good catch. I wouldn't say it's a known issue exactly, but the CI build runs tests against an actual Redshift instance so it's not super unexpected for it to randomly fail sometimes. Probably related to me uploading half a dozen PRs simultaneously.

Issue #3224, opened 2025-08-27
Fix collection naming for batch connectors


dyaffe:
**Background**
When we updated our collection naming for target schema naming, it appears we at least missed doing so for source-mysql-batch which is causing that feature not to work.

It's not postfixing /schema/table_name on the end of the name but rather /schema_table_name

PR #3272, opened 2025-09-09
source-*-batch: Recommend `schema/table` collection names

willdonnelly:
**Description:**

Previously we've recommended collection names like `schema_table` plus a special case which omitted `public` and `dbo` schema names, but this should be changed to `schema/table` to match current CDC capture behavior.

There is one rather annoying thing here, which is that we have to keep the _resource_ name unchanged or it will break discovery merging for preexisting capture bindings, so now we have two distinct name-recommendation functions, one for the resource name which follows the old pattern, and one for the collection name which follows the new pattern. But this is unavoidable, I think.

This fixes https://github.com/estuary/connectors/issues/3224


williamhbaker (approved):
LGTM

PR #3292, opened 2025-09-15
source-mysql: Switch to estuary/go-mysql fork

willdonnelly:
**Description:**

This fork currently has no differences from the upstream go-mysql client library but we intend to fix at least one bug so we ought to have our own forked version.

The only immediate impact of this change is that we're updating to the latest HEAD revision of the client library rather than the older commit-version we were previously on.


williamhbaker (approved):
LGTM

PR #3296, opened 2025-09-16
source-bigquery-batch: Improved STRUCT and ARRAY handling

willdonnelly:
**Description:**

Modifies the capture value translation code for STRUCT columns so that the captured output will be an object with properties for each struct value rather than just the values tuple as a JSON array.

Also tweaks discovery logic so that STRUCT and ARRAY composite types are discovered as `type: object` and `type: array` respectively, but discovery goes no further than that (no attempt is made to describe element types recursively) since schema inference ought to be doing the work on that end anyway.


williamhbaker (approved):
LGTM

PR #3302, opened 2025-09-18
source-jira-native: implement real-time and lookback stream approach for eventual consistency

Alex-Bair:
Jira's Issue Search API is eventually consistent, meaning recently updated issues may not immediately appear in search results. This can cause data loss when capturing issue updates in real-time. The connector previously avoided capturing any data generated in the most recent 5 minutes to combat eventual consistency, but 5 minutes is not long enough; Jira's own documentation/ changelogs suggest some changes could be delayed by up to 25 minutes.

This PR implements a dual-stream approach with two subtasks:
- REALTIME: Captures issues immediately but may miss some due to consistency
- LOOKBACK: Trails behind by 2 hours to capture any missed records

Key changes:
- Split single incremental cursor into two subtasks (REALTIME/LOOKBACK)
- Added emitted_changes_cache to deduplicate records between streams
- Extended consistency horizon from 5 minutes to 2 hours
- Added state migration for existing captures

**Description:**

**Workflow steps:**

**Notes for reviewers:**

- The real-time subtasks do not use a delay when capturing changes from the Issue Search API.
- The lookback subtasks do use a 2 hour delay when capturing changes from the Issue Search API, and they maintain that 2 hour delay.
- An existing capture that previously did not have subtasks will successfully migrate over to use subtasks.

### DIFF
--- a/source-jira-native/source_jira_native/resources.py
+++ b/source-jira-native/source_jira_native/resources.py
@@ -10,6 +10,7 @@ from estuary_cdk.http import HTTPMixin, TokenSource, HTTPError
 
 from .models import (
     BoardChildStream,
+    ConnectorState,
     EndpointConfig,
     Epics,
     FilterSharing,
@@ -64,11 +65,13 @@ from .api import (
     snapshot_statuses,
     snapshot_system_avatars,
     url_base,
-    ISSUE_JQL_SEARCH_LAG,
+    ISSUE_SEARCH_EVENTUAL_CONSISTENCY_HORIZON,
 )
 
 
 TEST_CONFIG_DOMAIN = "dontchangethisdomain.notareal.jira.domain"
+REALTIME = "realtime"
+LOOKBACK = "lookback"
 
 
 async def validate_projects(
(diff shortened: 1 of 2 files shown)


JustinASmith (approved):
LGTM

source-jira-native/source_jira_native/resources.py
```diff
@@ -324,18 +327,44 @@ def open(
...
+            task.checkpoint(
+                ConnectorState(
+                    bindingStateV1={binding.stateKey: state}
+                ),
+                # Merging in the new incremental state leaves the non-subtask state in the connector's state. Pydantic
+                # raises a validation error since there should never be _both_ non-subtask and subtask state for a binding's
+                # incremental state. To ensure the non-subtask state is cleared after the migration, merge_patch is set to
+                # False when checkpointing the migrated state.
+                merge_patch=False,
+            )
```
Alex-Bair, line 349: For future reference, this strategy did not work like I expected. When using `merge_patch=False`, then _entire_ connector 's state should be included in the `ConnectorState` checkpoint, not just the state for an individual resource. When this was rolled out, only a single resource's state was maintained and all other resource's states were wiped out in the flurry of `merge_patch=False` updates.

Instead of using `merge_patch=False`, I should have done something like:
```python
# Clear out old state.
task.checkpoint(
    ConnectorState(
        bindingStateV1={binding.stateKey: None}
    ),
)

# Set new state.
task.checkpoint(
    ConnectorState(
        bindingStateV1={binding.stateKey: state}
    ),
)
```
This would have removed the old state first, then checkpointed the new state without clearing out other resource's states.

PR #3303, opened 2025-09-18
materialize-elasticsearch: support "routing" field config

williamhbaker:
**Description:**

Adds support for a "routing" field configuration option, which designates a field in the collection key to be used as the routing parameter for Elasticsearch documents.

Updates the integration test configuration to use multiple shards for a couple of bindings, and add the routing config to one of the key fields.

**Workflow steps:**

See https://github.com/estuary/flow/pull/2406

**Notes for reviewers:**

Manually tested to verify that:
- A non-key routing field is not allowed
- More than one routing field is not allowed
- Documents are routed to indexes correctly. The integration tests exercise this too, but I verified that documents are distributed as expected across shards and can be retrieved by queries with the routing key set, but not without it, as expected

### DIFF
--- a/materialize-elasticsearch/transactor.go
+++ b/materialize-elasticsearch/transactor.go
@@ -47,6 +47,10 @@ type binding struct {
 	// Present if the binding includes the root document, empty if not. This is usually the default
 	// "flow_document" but may have an alternate user-defined projection name.
 	docField string
+
+	// Index within the collection key to use for the routing parameter. nil if no routing field
+	// is configured.
+	routingField *int
 }
 
 type transactor struct {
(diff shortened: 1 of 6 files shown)


materialize-elasticsearch/transactor.go
```diff
@@ -187,10 +199,25 @@ func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
...
+			}
+		}
+
 		if it.Delete && t.cfg.HardDelete {
 			// Ignore items which do not exist and are already deleted.
 			if it.Exists {
-				batch = append(batch, []byte(`{"delete":{"_id":"`+id+`"}}`)...)
+				batch = append(batch, []byte(`{"delete":{"_id":"`+id+`"`)...)
+				if routingVal != nil {
+					batch = append(batch, []byte(`,"routing":`+string(routingVal))...)
```
jgraettinger, line 218: nit: I'm pretty sure the compiler isn't smart enough to avoid the extra allocation / copies here, as compared to:

```
batch = append(append(batch, `,"routing":"`...), routingVal...)
```

jgraettinger (approved):
LGTM

PR #3304, opened 2025-09-18
source-*-batch: Fix missing /_meta/source on deletions

willdonnelly:
**Description:**

The logic for inferred deletions is entirely separate from the main "turn database rows into documents" loop for obvious reasons, and was missed when adding /_meta/source to the batch captures.

This only impacts a relatively small number of bindings, since this logic is only used when the key is /_meta/row_id and the binding has no cursor configured. Impacted bindings will have failed since the original change went live but should not have committed incorrect data.


source-bigquery-batch/.snapshots/TestFeatureFlagKeylessRowID-Enabled-Capture
```diff
@@ -1,24 +1,22 @@
 # ================================
-# Collection "acmeCo/test/featureflagkeylessrowid_enabled_905157": 23 Documents
```
williamhbaker, line 2: I'm not sure why there'd be fewer documents in this snapshot after this change, especially since all the other snapshots are changing only in the expected way.

williamhbaker (approved):
LGTM, a little puzzled by the bigquery test output change though

source-bigquery-batch/.snapshots/TestFeatureFlagKeylessRowID-Enabled-Capture
willdonnelly: The BigQuery test snapshots are just kind of unstable (because BigQuery is super slow and really goes hard on the "eventually your changes will show up" side of things, IIRC). They're still good enough to serve as a sanity check when run locally, but that's one of the reasons we don't run the BQ database tests in CI.

PR #3309, opened 2025-09-22
materialize-snowflake: retry harder on transient errors

williamhbaker:
**Description:**

Brings back the change from https://github.com/estuary/connectors/commit/61e23d4965e4b21d65cf2fe71b56f5af3b0d4aa2 that I had reverted. I'm still not sure if it is really necessary, but we've seen that these network failures are indeed quite transient, and sometimes don't happen for days on end before starting to happen again quite frequently. So it may have been too early to know if we can get by without retrying in our application code.

Also bumps up the `maxRetryCount` config quite a bit so that errors are retried for longer, up to about 10 minutes by my estimation. Since we do continue to see failures at 10 retries, it seems that the i/o issues might be more persistent and require more patience to resolve.

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM

PR #3317, opened 2025-09-24
source-mongodb: fix nested field lookup for cursor fields

JustinASmith:
**Description:**

The connector was treating dotted cursor field names like `"_id.ts"` as literal field names instead of nested document paths. This caused backfills to fail with "element not found" errors when trying to access nested fields.

Split cursor field once per batch in `pullCursor()` and call `LookupErr()` with individual path components via variadic expansion, enabling proper access to nested document fields like `document["_id"]["ts"]`.

_NOTE:_ This assumes that users do not actually want to use a cursor field that is literally named with a ".". The only thought I had on handling that would be trying to lookup the cursor field via the literal "_id.ts" and fallback to the nested dot-notation lookup. But I thought that may be a bold assumption to try one way and fallback to the other in case there is some situation where a user somehow added both representations to the document and we would always take the literal representation if we went about it this way.

**Workflow steps:**

**Notes for reviewers:**

Tested locally via new unit + integration tests for deeply nested cursor fields (up to 4 via unit tests) and 1-level deep via integration tests.

### DIFF
--- a/source-mongodb/main_test.go
+++ b/source-mongodb/main_test.go
@@ -806,3 +806,65 @@ func (v *correctnessInvariantsCaptureValidator) Summarize(w io.Writer) error {
 func (v *correctnessInvariantsCaptureValidator) bindingIdentifier(database, collection string) string {
 	return database + "." + collection
 }
+
+// TestNestedCursorFields tests nested cursor field lookups like "a.b".
+func TestNestedCursorFields(t *testing.T) {
+	const (
+		database   = "testDb"
+		collection = "nestedCursorCollection"
+	)
+	ctx := context.Background()
+	client, cfg := testClient(t)
+
+	cleanup := func() {
+		require.NoError(t, client.Database(database).Drop(ctx))
+	}
+	cleanup()
+	t.Cleanup(cleanup)
+
+	binding := makeBinding(t, database, collection, captureModeIncremental, "a.b")
+	cs := &st.CaptureSpec{
+		Driver:       &driver{},
+		EndpointSpec: &cfg,
+		Checkpoint:   []byte("{}"),
+		Validator:    &st.OrderedCaptureValidator{},
+		Sanitizers:   commonSanitizers(),
+		Bindings:     []*flow.CaptureSpec_Binding{binding},
+	}
+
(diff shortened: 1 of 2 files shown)


source-mongodb/main_test.go
```diff
@@ -806,3 +806,62 @@ func (v *correctnessInvariantsCaptureValidator) Summarize(w io.Writer) error {
...
+		Driver:       &driver{},
+		EndpointSpec: &cfg,
+		Checkpoint:   []byte("{}"),
+		Validator:    &st.OrderedCaptureValidator{},
+		Sanitizers:   commonSanitizers(),
+		Bindings:     []*flow.CaptureSpec_Binding{binding},
+	}
+
+	col := client.Database(database).Collection(collection)
+	refTime := time.Date(2023, 11, 7, 16, 57, 0, 0, time.UTC)
```
williamhbaker, line 836: This `refTime` and related handling appears to be a superfluous holdover of copypasta from the timeseries tests.

source-mongodb/main_test.go
```diff
@@ -806,3 +806,62 @@ func (v *correctnessInvariantsCaptureValidator) Summarize(w io.Writer) error {
...
+				"b": refTime.Add(time.Duration(idx) * time.Second),
+			},
+			"data": fmt.Sprintf("test data %d", idx),
+		}
+		_, err := col.InsertOne(ctx, doc)
+		require.NoError(t, err)
+	}
+
+	backdateState(t, cs, boilerplate.StateKey(binding.StateKey))
+	advanceCapture(ctx, t, cs)
```
williamhbaker, line 866: Need to actually do a snapshot, and that should be committed as well. See https://github.com/estuary/connectors/blob/32bd7d5f32802da101278046a3d015d00de91738/source-mongodb/main_test.go#L267

williamhbaker (commented):
Left a few comments to address

williamhbaker:
Latest update seems good but you need to run the new test locally and check in the snapshot

JustinASmith:
> Latest update seems good but you need to run the new test locally and check in the snapshot

@williamhbaker sorry about that! Thought I had that in my commit for the tests. Should be good to go now, right?

PR #3326, opened 2025-09-25
source-hubspot-native: add `marketing_emails` stream

Alex-Bair:
**Description:**

Hubspot is deprecating the `/marketing-emails/v1` API at the beginning of next month. That API is used in the imported `source-hubpot`'s `marketing-emails` stream, which will break once the API stops accepting requests. There are extremely significant differences between the data returned by `/marketing-emails/v1/emails/with-statistics` endpoint and the new `/marketing/v3/emails` endpoint that's suppose to be the replacement. There's no way the imported `source-hubspot` connector could be updated to _not_ make this a big breaking change.

Instead, we're adding the `marketing_emails` stream to the native `source-hubspot-native` connector. It will capture data from the `/marketing/v3/emails` endpoint without worrying about maintaining compatibility with the soon-to-be-sunset v1 endpoint.

Discover snapshot changes are expected due to the addition of the new stream.

**Workflow steps:**

`source-hubspot-native`'s documentation should be updated to reflect the new `marketing_emails` stream.

**Notes for reviewers:**

Tested on a local stack. Confirmed all expected documents are captured and the `marketing_emails` stream is only discovered if the provided credentials can successfully access the associated endpoint.

Note: we'll need to remove the `marketing_emails` stream from `source-hubspot` once the v1 endpoint stops working. If Hubspot does indeed sunset it on 01OCT2025, we'll be doing that pretty soon.

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api.py
+++ b/source-hubspot-native/source_hubspot_native/api.py
@@ -3,6 +3,7 @@ import functools
 import itertools
 from datetime import UTC, datetime, timedelta
 from logging import Logger
+import time
 from typing import (
     Any,
     AsyncGenerator,
@@ -38,6 +39,7 @@ from .models import (
     FormSubmission,
     FormSubmissionContext,
     LineItem,
+    MarketingEmail,
     Names,
     OldRecentCompanies,
     OldRecentContacts,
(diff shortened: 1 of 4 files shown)


Alex-Bair:
I added an `"includeStats": True` query param after marking this for review yesterday, but I didn't test it out until this morning. I constantly forget that `aiohttp`/`yarl` doesn't allow boolean query params since it's not clear what they should be serialized to (`"true"`? `"True"`? `"1"`? etc.). The latest push was just changing the `True` query param to be `"true"` to fix that bug.

PR #3327, opened 2025-09-25
source-sqlserver-batch: Discover lengths of CHAR/BINARY columns

willdonnelly:
**Description:**

This commit adds `maxLength` and (depending on the column type) possibly `minLength` properties to JSON schemas for CHAR/VARCHAR and BINARY/VARBINARY columns. I believe these are the only columns that need it.

This _should_ be backwards-compatible in all cases. For write schemas with automatic updates it will of course be fine (provided the actual data we capture matches the discovered constraints) and if automatic updates are disabled it will be fine because the new constraints won't exist. For inferred schemas on existing collections this should be a no-op since they've already merged in a sourced schema with no such constraints (and inference will never make the length constraints tighter) and new collections will just have them from the get-go (which is the entire point).

### DIFF
--- a/source-sqlserver-batch/discovery.go
+++ b/source-sqlserver-batch/discovery.go
@@ -280,6 +280,8 @@ type basicColumnType struct {
 	format          string
 	nullable        bool
 	description     string
+	minLength       *uint64
+	maxLength       *uint64
 }
 
 func (ct *basicColumnType) JSONSchema() *jsonschema.Schema {
@@ -292,6 +294,10 @@ func (ct *basicColumnType) JSONSchema() *jsonschema.Schema {
 	if ct.contentEncoding != "" {
 		sch.Extras["contentEncoding"] = ct.contentEncoding // New in 2019-09.
 	}
+
+	// Copy the min/max lengths if present
+	sch.MinLength = ct.minLength
+	sch.MaxLength = ct.maxLength
 
 	if ct.jsonTypes != nil {
 		var types = append([]string(nil), ct.jsonTypes...)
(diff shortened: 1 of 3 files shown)


source-sqlserver-batch/discovery.go
```diff
@@ -356,6 +363,21 @@ func discoverColumns(ctx context.Context, db *sql.DB, discoverSchemas []string)
 		}
 		dataType.nullable = isNullable
 
+		// Add length constraints for char/binary columns with defined lengths. A
+		// value of -1 is used for certain large-object types such as XML.
+		if charMaxLength.Valid && charMaxLength.Int64 > 0 {
```
jgraettinger, line 368: A wrinkle: since we're transiting this through JSON Schema, we need to honor how it sees binary data (as base64-encoded strings) and validates maxLength (as the number of UTF-8 codepoints).

That means that `maxLength` of a binary column is actually 4/3 times it's raw binary length.
We should also be setting `contentEncoding: base64` for binary columns.

Within materializations, we would convert going the other way when mapping back to binary.

willdonnelly: Oh, that's a very good point. I spent all my time thinking about how length constraints actually work for textual data and completely overlooked the base64 encoding on byte data.

willdonnelly: Okay, getting back to this now. I've updated the discovery logic so the JSON length constraints are `ceil(n/3)*4` characters for the binary(n) column types and verified that we are setting `contentEncoding: base64` correctly.

PR #3329, opened 2025-09-26
estuary-cdk: retry more HTTP connection errors

Alex-Bair:
**Description:**

We've observed more errors beyond `asyncio.TimeoutError` that crash connectors when trying to establish an HTTP connection. As long as these errors occur before any data is consumed by the connector, these connection-related errors are safe to retry.

What's tricky is some error types could be raised when establishing the connection _or_ reading the response body. To ensure we only retry errors when establishing the connection, I refactored the retry logic to live in `HTTPMixin._request_stream` and only apply when establishing the connection instead of for the entirety of `HTTPMixin._request_stream`.

Note: the GH workflow that builds CDK connectors doesn't run if there's _only_ CDK changes, so I updated `source-hubspot`'s failing capture snapshot to trigger a rebuild of all CDK connectors.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed HTTP requests are still made and succeed.

### DIFF
--- a/estuary-cdk/estuary_cdk/http.py
+++ b/estuary-cdk/estuary_cdk/http.py
@@ -88,23 +88,20 @@ class HTTPSession(abc.ABC):
     ) -> bytes:
         """Request a url and return its body as bytes"""
 
-        async def _do_request() -> bytes:
-            chunks: list[bytes] = []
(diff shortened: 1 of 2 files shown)


JustinASmith (approved):
LGTM

PR #3331, opened 2025-09-26
source-hubspot-native: capture missing and archived marketing emails

Alex-Bair:
**Description:**

Capturing all data from Hubspot's `/marketing/v3/emails` API requires multiple passes with different query parameter values due to how the API behaves: 
- includeStats=true returns incomplete data, omitting some records entirely
- archived and non-archived records must be queried separately

This commit implements a 4-pass workaround using nested loops over both `archived=[false,true]` and `includeStats=[true,false]` parameters. The first pass with stats captures statistics where available, subsequent passes capture the remaining data and deduplicates already-seen records by ID.

**Workflow steps:**

Connector docs should be updated to reflect the new stream (I forgot to do that after the previous PR https://github.com/estuary/connectors/pull/3326)

**Notes for reviewers:**

Tested on a local stack. Confirmed multiple more records are captured than before, and all previously missing, not deleted ids reported by the user are now captured.

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api.py
+++ b/source-hubspot-native/source_hubspot_native/api.py
@@ -1234,26 +1234,15 @@ def fetch_delayed_email_events(
     return _fetch_email_events(log, http, since, until)
 
 
-async def _fetch_marketing_emails_updated_between(
+async def _paginate_through_marketing_emails(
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM!

Issue #3334, opened 2025-09-29
SQL CDC: Mark unique secondary indices as fallback keys when used as collection keys


willdonnelly:
In our SQL CDC connectors, we normally just use the primary key of a table as the collection key and everything is happy. But when there is no primary key, we have a heuristic which discovers a unique secondary index on the table and uses _that_ as the collection key (arbitrarily picking one if there are multiple valid indices).

This feature predates the concept of "fallback keys" on a collection, but really these unique secondary indices are a best-effort fallback which the user might want to override with their own choice, so it feels correct to describe them as fallback keys in our discovery results.

PR #3336, opened 2025-09-29
sqlcapture: Mark secondary indices as fallback keys

willdonnelly:
**Description:**

This fixes https://github.com/estuary/connectors/issues/3334

Also adds the `IsFallbackKey: true` for truly keyless tables' fallback keys, because apparently we never actually used the new flag before.


mdibaiee (approved):
LGTM, does this in some way affect existing tasks?

Issue #3338, opened 2025-09-30
SQL Connectors: Discover text/binary column sizes


willdonnelly:
Because of how Sourced Schema messages interact with schema inference, it's important that we include size constraints in the JSON schema for `[VAR]CHAR(n) / [VAR]BINARY(n)` columns, because schema inference will loosen but never tighten those constraints, and so if we emit sourced schemas without any such constraints they will never be added. And sometimes materializations need to know the possible range of sizes so they don't make overly conservative assumptions.

Note that I'm pretty sure this means all currently existing collections for captures which use sourced schemas plus schema inference (which is essentially all SQL captures, both CDC and batch) will never have constraints inferred, since even if we start including them going forward the current inferred schema has no constraints and so they can never become tighter than that.

The plan is to just add them unconditionally to discovered schemas and sourced schema updates. I believe this should be entirely backwards compatible since:
- For write schemas without automatic updates this will make no difference since they're not updated, of course.
- For write schemas with automatic updates the new size constraints will of course be fine so long as they're accurate.
  - **This part** is the most likely source of potential bugs, in my opinion. If the size constraints are somehow _not_ accurate it will result in captures failing with schema validation issues. But we can just roll back the offending commit and fix the bug in that case, with no real harm done.
- For inferred schemas on existing collections the new constraints won't do anything since the current inferred schema is already looser and won't be tightened up.
- For inferred schemas on new collections the new constraints will provide a more strict starting point, but if they're wrong it will just be loosened up automatically. And this change in behavior is the whole point of the exercise.

PR #3341, opened 2025-09-30
source-{postgres,mysql,redshift,bigquery}-batch: Discover column lengths

willdonnelly:
**Description:**

Part of https://github.com/estuary/connectors/issues/3338

This adds column-size discovery for text/binary column types to the batch PostgreSQL, MySQL, Redshift, and BigQuery captures.

To be merged after https://github.com/estuary/connectors/pull/3327


source-redshift-batch/.snapshots/TestFieldLengthDiscovery
```diff
@@ -0,0 +1,172 @@
...
+            ]
+          },
+          "varbyte_16": {
+            "description": "(source type: varbyte)",
+            "contentEncoding": "base64",
+            "type": [
+              "string",
+              "null"
+            ]
+          },
```
Alex-Bair, line 137: Should `varbyte_16` have a `maxLength` in this snapshot? I think so based off the logic in `discovery.go` and the various comments in the code.

Claude's spitballing that `information_schema._pg_char_max_length()` may not work for Redshift's custom `VARBYTE(n)` type. That sounds feasible, but I'm entirely unfamiliar with Redshift so take that with a grain of salt.

Alex-Bair (approved):
LGTM % I think `varbyte_16` is missing a `maxLength` in Redshift's `TestFieldLengthDiscovery` snapshot?

Issue #3335, opened 2025-09-29
source-postgres: Change default to 'Discover Only Published'


willdonnelly:
In PostgreSQL, issuing an `ALTER PUBLICATION ___ ADD TABLE ___` query requires being the owner of both the publication and the table. Consequently there are basically three scenarios:

1. The user created the Flow publication `FOR ALL TABLES` and so every table in the database is published automatically.
2. The user set up the Flow capture user as a superuser and so we're able to automatically add tables to the publication manually.
3. The user did not create the publication `FOR ALL TABLES` or give us superuser access, and so they will have to explicitly add tables to the publication before we can successfully capture them.

Scenario (2) is rare. Users typically do not give us superuser access, our docs don't guide them to do so, and in any case a user who is inclined to give us that level of access can just as easily create the publication `FOR ALL TABLES`. So in practice, either the publication is `FOR ALL TABLES` or the user is explicitly adding tables to it, but either way we have to assume that tables will be provided to us via the publication.

We have an advanced option "Discover Only Published Tables". When set, source-postgres will only return bindings from discovery for tables which are present in the configured Flow publication. But it defaults to off, which means that typically we _will_ discover any newly created tables regardless of whether they've been added to the publication. Since by default we will create bindings for any newly discovered tables and they will then fail to actually work until the table is published, this is sub-optimal.

Things can probably be improved by simply removing that option and making the feature be enabled unconditionally. In the scenarios where things would have kept working correctly before (1 and 3 above) this will behave exactly the same as it always has, the only significant difference is that if a table is created and not added to the publication we will not discover a binding for it. But practically speaking that's the same thing that we've always done, it's just that in the past there's been a couple intermediate steps where autodiscovery publications fail and cause side effects like blocking inferred schema updates.

One additional consideration is our "create publication" prerequisite check. Currently this check first looks to see if the publication exists, and if it doesn't exist it will be autocreated. I think this is probably not actually what we want (see again the publication ownership constraints on manually adding tables) and we should just require the user to create the publication in all cases.

PR #3342, opened 2025-09-30
source-postgres: Make 'Discover Only Published' unconditional

willdonnelly:
**Description:**

This fixes https://github.com/estuary/connectors/issues/3335

This PR is somewhere between a serious PR to be merged as-is and a declaration of intent. Dave and I have discussed this (see https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1758826828286479) and come to the conclusion that the current behavior of the PostgreSQL capture around discovering non-published tables is a bad default and it's not clear that there's any real value to allowing users to discover non-published tables in any scenario.

Restating the issue:
* To add a table to a PostgreSQL publication one has to be the owner of both the table and the publication in question (or a superuser, I think).
* Since the `flow_capture` user is generally not the owner of a customer's tables, our ability to meaningfully help here is limited.
* So in practice there are 2-3 outcomes:
  * Either the user creates a publication `FOR ALL TABLES`, in which case all tables are published and the check doesn't do anything.
  * Or else the user has to manually add tables to the publication, in which case creating a binding for a new table before they publish it just causes the capture to start failing until they do so, _if_ they do so.
  * Or else the user gets tired of us screwing things up by adding new tables they didn't want us to, and turns off automatically adding new tables, at which point the whole "Discover Only Published" setting doesn't really make any difference.

Thus we should accept the reality of the situation and change the default behavior. But since this is a boolean toggle we can't really tell the difference between existing captures where the user really wants it to be off vs one where they just didn't change any defaults. I'm betting it's almost entirely the latter, and so we should just change it across the board.

The bits I'm not entirely sure about here:
1. Does this reasoning make sense, or am I mistaken about how common various use-cases are?
2. Will things work correctly if I just remove an endpoint config property or do we need to do a three-step thing where first we make it not do anything, then remove it from prod configurations, then delete the property outright?
3. Is it correct to also remove the logic which tries to auto- create publications? I think it does, it was added in order to be helpful but I think in practice the aforementioned issues around publication ownership make the connector-created publication useless in almost all cases.

**Workflow steps:**

Users will no longer have access to the "Discover Only Published Tables" advanced option, it will just always behave like that going forward.

All PostgreSQL CDC connector docs should probably be adjusted to reflect this change and maybe discuss publication management a bit more.


willdonnelly:
Answering my own question: we can't entirely delete the option because JSON unmarshalling will fail on an unexpected property name. Explicitly including the struct field but making it clear that it's unused and omitting it from the JSON schema seems to be the way to go.

PR #3346, opened 2025-10-01
materialize-boilerplate: skip nil actions during delete/create

danielnelson:
**Description:**

We allow a resource action to be nil and skip their execution, but when we run delete/create together we didn't handle the case where one or both are nil.  These must be ran together in a single action so provide ordering in case concurrency is enabled.

**Workflow steps:**

NA

NA

**Notes for reviewers:**


Alex-Bair (approved):
LGTM % a small nit.

Before merging, can the two commits please be squashed together into a single commit that starts with `materialize-boilerplate` instead of `materialize-mongodb`? That'd make it clearer that this commit affects all materializations using `materialize-boilerplate`, not just `materialize-mongodb`.

PR #3351, opened 2025-10-02
source-azure-blob-storage: fix prefix filtering for multiple bindings

Alex-Bair:
**Description:**

The Azure Blob Storage connector was ignoring the `query.Prefix` parameter in its `List()` method, causing all bindings to capture from the container root instead of their specified subpaths. This meant that bindings configured with different paths (e.g., "container/dir1", "container/dir2") would all see the same files from the entire container.

This brings the Azure connector in line with other `filesource` connectors (S3, GCS) which correctly handle prefix filtering. Each binding now only captures files under its configured subpath, enabling proper multi-binding usage patterns.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack by adding additional bindings with `flowctl`. Confirmed each binding only captures blobs within its specified path.

### DIFF
--- a/source-azure-blob-storage/main.go
+++ b/source-azure-blob-storage/main.go
@@ -157,9 +157,17 @@ func (az *azureBlobStore) check(ctx context.Context) error {
 }
 
 func (az *azureBlobStore) List(ctx context.Context, query filesource.Query) (filesource.Listing, error) {
-	pager := az.client.NewListBlobsFlatPager(az.cfg.ContainerName, &azblob.ListBlobsFlatOptions{
+	var container, prefix = filesource.PathToParts(query.Prefix)
+
+	listOptions := &azblob.ListBlobsFlatOptions{
 		Include: azblob.ListBlobsInclude{Snapshots: true, Versions: true},
-	})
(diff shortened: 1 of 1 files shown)


source-azure-blob-storage/main.go
```diff
@@ -228,14 +241,20 @@ func (l *azureBlobListing) Next() (filesource.ObjectInfo, error) {
...
 		obj.ModTime = *blob.Properties.LastModified
 		obj.ContentType = *blob.Properties.ContentType
 		obj.Size = *blob.Properties.ContentLength
 
 		if blob.Properties.ContentEncoding != nil {
 			obj.ContentEncoding = *blob.Properties.ContentEncoding
 		}
+
+		if l.query.StartAt != "" && obj.Path < l.query.StartAt {
+			log.WithField("path", obj.Path).Debug("skipping blob before StartAt")
```
danielnelson, line 254: How many logs is too many?  Should we count the number of skipped paths and log only the count instead of each file?

danielnelson (approved):
LGTM, just one question.

PR #3352, opened 2025-10-02
source-postgres: Discover CHAR/VARCHAR column lengths

willdonnelly:
**Description:**

Part of https://github.com/estuary/connectors/issues/3338

This adds column-size discovery for text column types to the PostgreSQL CDC connector. This ended up being a fairly hefty chunk of work, for two reasons:
1. The way that column types are represented and decoded inside the source-postgres discovery logic just didn't accomodate adding size metadata, and doing it properly required moving some logic around.
2. It then turned out that there's an even trickier edge case: `VARCHAR(123) ARRAY` columns require being able to express the concept of "an array, whose elements are varchars of maximum length 123" as a value, and so I ended up refactoring a decent chunk of the discovery type representation to more consistently use structs and interfaces instead of switching on bare string names.

I'm reasonably sure that this refactoring is correct and didn't break anything.


Alex-Bair (approved):
LGTM

PR #3356, opened 2025-10-03
source-datadog: support filtering logs with a user-provided query

Alex-Bair:
**Description:**

Often, users have a significant amount of logs in Datadog and they don't want to capture all of them. The `advanced.logs_query` config option lets users filter out which logs the connector captures.

**Workflow steps:**

The connector's documentation should be updated to reflect the new config option.

**Notes for reviewers:**

Tested on a local stack. Confirmed the `advanced.logs_query` field is used when fetching `logs`.


nicolaslazo (approved):
LGTM!

PR #3372, opened 2025-10-08
source-hubspot-native: add contact_lists, contact_list_memberships an…

### LINKED ISSUE #3339: source-hubspot-native: add `contact_lists`, `contact_list_memberships`, and `feedback_submissions` streams
There have been requests to add the following streams to `source-hubspot-native`:
- `contact_lists`
- `contact_list_memberships`
- [`feedback_submissions`](https://developers.hubspot.com/docs/api-reference/crm-feedback-submissions-v3/guide)

Both [`contact_lists` and `contact_list_memberships` exist](https://github.com/estuary/connectors/blob/25bb7850554034fe095d21c6fb2d1c036174e92a/source-hubspot/source_hubspot/streams.py#L1293-L1351) in the imported `source-hubspot` connector and hit the v1 API. When adding `contact_lists` and `contact_list_memberships` to `source-hubspot-native`, we should avoid using the v1 API since it will be sunset on 30APR2026. Instead, we should use the v3 API. HubSpot's [migration docs](https://developers.hubspot.com/docs/api-reference/crm-lists-v3/v1-migration-guide) should help identify which endpoints should be used for fetching records in each stream.

nicolaslazo:
…d feedback_submissions streams

**Description:**

This PR adds three new streams: `contact_lists`, `contact_list_streams`, and `feedback_submissions`. The first two are particular in that they don't follow the CRM objects API and have limited filtering and sorting options.

Closes #3339 

**Workflow steps:**

**Notes for reviewers:**

See the first commit for git hunks related to the PR itself. The second commit only pertains to linter corrections from `black`

### DIFF
--- a/source-hubspot-native/acmeCo/contact_list_memberships.schema.yaml
+++ b/source-hubspot-native/acmeCo/contact_list_memberships.schema.yaml
@@ -1,0 +1,42 @@
+---
+$defs:
+  Meta:
+    properties:
+      op:
+        default: u
+        description: "Operation type (c: Create, u: Update, d: Delete)"
+        enum:
+          - c
+          - u
+          - d
+        title: Op
+        type: string
+      row_id:
+        default: -1
+        description: "Row ID of the Document, counting up from zero, or -1 if not known"
+        title: Row Id
+        type: integer
+    title: Meta
+    type: object
+additionalProperties: true
+properties:
+  _meta:
+    $ref: "#/$defs/Meta"
+    description: Document metadata
+  listId:
+    title: Listid
+    type: integer
+  recordId:
+    title: Recordid
+    type: string
+  membershipTimestamp:
+    format: date-time
+    title: Membershiptimestamp
+    type: string
+required:
+  - listId
+  - recordId
+  - membershipTimestamp
+title: ContactListMembership
+type: object
+x-infer-schema: true
--- a/source-hubspot-native/acmeCo/contact_lists.schema.yaml
+++ b/source-hubspot-native/acmeCo/contact_lists.schema.yaml
@@ -1,0 +1,43 @@
+---
+$defs:
+  Meta:
+    properties:
+      op:
+        default: u
+        description: "Operation type (c: Create, u: Update, d: Delete)"
+        enum:
+          - c
+          - u
+          - d
+        title: Op
+        type: string
+      row_id:
+        default: -1
+        description: "Row ID of the Document, counting up from zero, or -1 if not known"
+        title: Row Id
+        type: integer
+    title: Meta
+    type: object
+additionalProperties: true
+properties:
+  _meta:
+    $ref: "#/$defs/Meta"
+    description: Document metadata
+  listId:
+    title: Listid
+    type: integer
+  updatedAt:
+    format: date-time
+    title: Updatedat
+    type: string
+  additionalProperties:
+    additionalProperties: true
+    title: Additionalproperties
+    type: object
+required:
+  - listId
+  - updatedAt
+  - additionalProperties
+title: ContactList
+type: object
+x-infer-schema: true
--- a/source-hubspot-native/acmeCo/flow.yaml
+++ b/source-hubspot-native/acmeCo/flow.yaml
@@ -4,6 +4,15 @@ collections:
     schema: companies.schema.yaml
     key:
       - /id
+  acmeCo/contact_list_memberships:
+    schema: contact_list_memberships.schema.yaml
+    key:
+      - /listId
+      - /recordId
+  acmeCo/contact_lists:
+    schema: contact_lists.schema.yaml
+    key:
+      - /listId
   acmeCo/contacts:
     schema: contacts.schema.yaml
     key:
--- a/source-hubspot-native/test.flow.yaml
+++ b/source-hubspot-native/test.flow.yaml
@@ -59,6 +59,14 @@ captures:
       - resource:
           name: marketing_emails
         target: acmeCo/marketing_emails
+      - resource:
+          name: contact_lists
+          interval: PT3H
+        target: acmeCo/contact_lists
+      - resource:
+          name: contact_list_memberships
+          interval: PT3H
+        target: acmeCo/contact_list_memberships
     interval: 3m
     shards:
       logLevel: debug
--- a/source-hubspot-native/tests/snapshots/snapshots__discover__stdout.json
+++ b/source-hubspot-native/tests/snapshots/snapshots__discover__stdout.json
@@ -1911,5 +1911,137 @@
     "key": [
       "/id"
     ]
+  },
+  {
+    "recommendedName": "contact_lists",
+    "resourceConfig": {
+      "name": "contact_lists",
+      "interval": "PT3H"
+    },
+    "documentSchema": {
+      "$defs": {
+        "Meta": {
+          "properties": {
+            "op": {
+              "default": "u",
+              "description": "Operation type (c: Create, u: Update, d: Delete)",
+              "enum": [
+                "c",
+                "u",
+                "d"
+              ],
+              "title": "Op",
+              "type": "string"
+            },
+            "row_id": {
+              "default": -1,
+              "description": "Row ID of the Document, counting up from zero, or -1 if not known",
+              "title": "Row Id",
+              "type": "integer"
+            }
+          },
+          "title": "Meta",
+          "type": "object"
+        }
+      },
+      "additionalProperties": true,
+      "properties": {
+        "_meta": {
+          "$ref": "#/$defs/Meta",
+          "description": "Document metadata"
+        },
+        "listId": {
+          "title": "Listid",
+          "type": "integer"
+        },
+        "updatedAt": {
+          "format": "date-time",
+          "title": "Updatedat",
+          "type": "string"
+        },
+        "additionalProperties": {
+          "additionalProperties": true,
+          "title": "Additionalproperties",
+          "type": "object"
+        }
+      },
+      "required": [
+        "listId",
+        "updatedAt",
+        "additionalProperties"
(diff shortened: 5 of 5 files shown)


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -1502,82 +1513,142 @@ async def fetch_contact_lists(
...
-            json.loads(await http.request(log, url, params=params)),
-            context=contact_list,
-        )
+    try:
+        while True:
+            async with concurrency_context.requests_semaphore:
+                response = ContactListMembershipResponse.model_validate(
+                    json.loads(await http.request(log, url, params=params)),
+                    context=contact_list,
+                )
```
Alex-Bair: nit: Instead of `model_validate(json.loads(...))`, you should be able to use `model_validate_json(...)`.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -1557,32 +1561,64 @@ async def _request_contact_list_memberships(
...
+            "list_count": len(all_lists),
+            "start": start.isoformat(),
+            "end": end.isoformat() if end else None,
+        },
+    )
+
+    if not all_lists:
+        return
+
+    queue: asyncio.Queue[ContactListMembership | None] = asyncio.Queue()
```
Alex-Bair: We'll want to bound the max size of this `asyncio.Queue()` to avoid allowing an infinite number of documents in the queue (this can cause the connector to exceed its 1 GB memory limit and get OOM killed). In practice, queues are almost always empty or full, not in between. Unless there's a reason to use a larger buffer size, I'd use the smallest allowed size of 1 here.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -1525,17 +1526,20 @@ async def _request_contact_list_memberships(
...
     while True:
-        response = ContactListMembershipResponse.model_validate(
-            json.loads(await http.request(log, url, params=params)),
-            context=contact_list,
-        )
+        async with requests_semaphore or nullcontext():
+            response = ContactListMembershipResponse.model_validate(
+                json.loads(await http.request(log, url, params=params)),
+                context=contact_list,
+            )
```
Alex-Bair: nit: Instead of `model_validate(json.loads(...))`, you should be able to do `model_validate_json(...)`. That's consistent with how we validate a `bytes` response from `http.request` into a Pydantic model instance.

source-hubspot-native/source_hubspot_native/models.py
```diff
@@ -594,3 +620,74 @@ def set_lastmodifieddate(cls, values):
...
+class ContactListMembershipResponse(BaseModel, extra="allow"):
+    class Cursor(BaseModel, extra="allow"):
+        after: str
+
+    class Paging(BaseModel, extra="allow"):
+        next: ContactListMembershipResponse.Cursor | None
+
+    results: list[ContactListMembership]
+    paging: Paging | None = None
+    total: int
```
Alex-Bair: During local testing, the connector hit this error:
```
Exception Group Traceback (most recent call last):
  |   File \"/opt/estuary-cdk/estuary_cdk/capture/task.py\", line 201, in run_task
  |     await child(t)
  |   File \"/opt/estuary-cdk/estuary_cdk/capture/common.py\", line 752, in backfill_closure
  |     await _binding_backfill_task(
  |   File \"/opt/estuary-cdk/estuary_cdk/capture/common.py\", line 944, in _binding_backfill_task
  |     async for item in pages:
  |   File \"/opt/source-hubspot-native/source_hubspot_native/api.py\", line 1754, in fetch_contact_list_memberships_page
  |     async for membership in _request_all_contact_list_memberships(
  |   File \"/opt/source-hubspot-native/source_hubspot_native/api.py\", line 1722, in _request_all_contact_list_memberships
  |     async with asyncio.TaskGroup() as tg:
  |                ^^^^^^^^^^^^^^^^^^^
  |   File \"/usr/local/lib/python3.12/asyncio/taskgroups.py\", line 71, in __aexit__
  |     return await self._aexit(et, exc)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File \"/usr/local/lib/python3.12/asyncio/taskgroups.py\", line 164, in _aexit
  |     raise BaseExceptionGroup(
  | ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File \"/opt/source-hubspot-native/source_hubspot_native/api.py\", line 1656, in _request_contact_list_memberships
    |     response = ContactListMembershipResponse.model_validate(
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File \"/opt/venv/lib/python3.12/site-packages/pydantic/main.py\", line 705, in model_validate
    |     return cls.__pydantic_validator__.validate_python(
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | pydantic_core._pydantic_core.ValidationError: 1 validation error for ContactListMembershipResponse
    | total
    |   Field required [type=missing, input_value={'results': []}, input_type=dict]
    |     For further information visit https://errors.pydantic.dev/2.11/v/missing
    +------------------------------------
```

That tells me that `total` can be absent from the API response, so this model will need updated to allow that. We've done this before by allowing the field to be `None` with a default value of `None`. 

If you don't use `total` at all for processing in the connector, removing the field from the model works too.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -1313,49 +1465,346 @@ async def fetch_marketing_emails_page(
...
+        yield item
+
+
+async def fetch_contact_lists_page(
+    http: HTTPSession,
+    log: Logger,
+    page: PageCursor,
+    cutoff: LogCursor,
+) -> AsyncGenerator[ContactList | PageCursor, None]:
+    assert isinstance(page, str | None)
```
Alex-Bair: nit: If `fetch_contact_lists_page` fetched a subset of results & yielded a new page cursor for the next `fetch_contact_list_page` invocation to use, it'd be possible for `page` to be a `str`. However since `fetch_contact_lists_page` fetches _all_ results in a single invocation then returns, `page` will always be the initial cursor of `None`. You could narrow this assertion to confirm `page` is `None` and just assign `page_ts` to `EPOCH_PLUS_ONE_SECOND` later in the function body.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -1521,68 +1523,138 @@ async def fetch_contact_lists(
...
+                await context.results.put(item)
+
+            if (next_cursor := response.get_next_cursor()) is None:
+                break
+            params["after"] = next_cursor
 
         log.info(
-            "Fetched page",
+            "Finished fetching memberships for list",
+            {"list": contact_list.model_dump()},
```
Alex-Bair: Instead of including the entire document with every field in the log with `model_dump()`, lets only include the `listId`. That makes it easier to see at a glance which list finished processing.

nicolaslazo:
Alex and I agreed that we could yield memberships one list at a time during backfills, in ascending list id order -- that should stop the connector from running for too long. We can consider concurrently working on batches of list ids in the future if throughput becomes a concern.

source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -1336,6 +1376,242 @@ def fetch_delayed_marketing_emails(
...
+    next_list = await anext(
+        (
+            item
+            async for item in _request_contact_lists(http, log, False)
+            if page is None or item.listId > page
+        ),
+        None,
+    )
+    if not next_list:
+        return
```
Alex-Bair: nit: Is it possible for a contact list to have an id of `0`? If so, `not next_list` would be truthy and the backfill would finish before capturing all data. It may not be possible for a contact list to have an id of `0`, but just in case it'd be safer to check `if next_list is None` instead.

source-hubspot-native/tests/snapshots/snapshots__spec__stdout.json
```diff
@@ -125,7 +125,7 @@
...
+      "authUrlTemplate": "https://app.hubspot.com/oauth/authorize?client_id={{#urlencode}}{{{ client_id }}}{{/urlencode}}&scope=crm.lists.read%20crm.objects.companies.read%20crm.objects.contacts.read%20crm.objects.deals.read%20crm.objects.owners.read%20crm.schemas.companies.read%20crm.schemas.contacts.read%20crm.schemas.deals.read%20e-commerce%20forms%20tickets&optional_scope=automation%20content%20crm.lists.read%20crm.objects.custom.read%20crm.objects.feedback_submissions.read%20crm.objects.goals.read%20crm.schemas.custom.read&redirect_uri={{#urlencode}}{{{ redirect_uri }}}{{/urlencode}}&response_type=code&state={{#urlencode}}{{{ state }}}{{/urlencode}}",
```
Alex-Bair: Just noting that if we do remove `crm.lists.read` from the optional scopes, the spec snapshot will need regenerated/updated as well.

Alex-Bair (approved):
LGTM! 🥳 The `contact_list_memberships` backfill makes incremental progress now since only a single list's memberships are fetched in a `fetch_contact_list_memberships_page` invocation, and it doesn't block for nearly as long as it did before.

I just had a couple small comments to address before merging. Let's also wait to merge this tomorrow once there's some more time in the day to monitor the release.

source-hubspot-native/source_hubspot_native/api.py
nicolaslazo: Yeah, good point

Alex-Bair (approved):

PR #3389, opened 2025-10-14
source-braintree-native: request resources individually after 500 error

Alex-Bair:
**Description:**

Braintree's API sometimes returns 500 errors when requesting multiple large resources in a single page, causing `source-braintree-native` to get stuck in retry loops. The CDK retries 5xx errors and does not surface 5xx errors to connectors, making it impossible for `source-braintree-native` to know when these 500 error retry loops are happening.

To address that, this PR's scope includes:
- Adding a `should_retry` parameter for HTTP requests made through the CDK. `should_retry` is a handler that tells the CDK which 5xx errors should _not_ be retried & should be bubbled up to the connector.
  - The current behavior of unconditionally retrying requests that receive 5xx responses is maintained for existing connectors since errors are only bubbled up if a `should_retry` handler is provided.
- Falling back to requesting resources individually in `source-braintree-native` when attempting to fetch a page of multiple resources results in a 500 response.

See individual commits for more detail. 

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed when the connector receives a 500 response after requesting a page of resources, it requests resources individually & those requests then succeed.

### DIFF
--- a/source-braintree-native/source_braintree_native/api/searchable_incremental_resource.py
+++ b/source-braintree-native/source_braintree_native/api/searchable_incremental_resource.py
@@ -2,11 +2,11 @@ import asyncio
 import itertools
 from datetime import datetime
 from logging import Logger
-from typing import Any, AsyncGenerator, TypeVar
+from typing import Any, AsyncGenerator, Awaitable, TypeVar
 
 from braintree import BraintreeGateway
 
-from estuary_cdk.http import HTTPSession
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM! I like the use of a handler function and the typing with a Protocol. Clean and useful.

JustinASmith:
Whoops - didn't see @dyaffe review on my end until I submitted my review.

PR #3394, opened 2025-10-16
materialize-s3-iceberg: Additional icebergctl debug logging

willdonnelly:
**Description:**

I'm still not entirely sure why the program is dying without any sort of exception being logged in the middle of committing a transaction, but seeing exactly what the exit code was should help narrow things down a bit further.


Alex-Bair (approved):
LGTM

PR #3400, opened 2025-10-16
materialize-s3-iceberg: Even more exit status info

willdonnelly:
**Description:**

Adds even more exit status information when iceberg-ctl fails.


Alex-Bair (approved):
LGTM

PR #3403, opened 2025-10-16
materialize-s3-iceberg/iceberg-ctl: Memory improvements

willdonnelly:
**Description:**

This PR adds two tweaks to hopefully reduce the likelihood of iceberg-ctl getting OOM killed:
1. We no longer reload the table metadata after committing a transaction. The reloaded metadata wasn't really used for anything except for a human-readable log message, and we have reason to suspect that the metadata was the driver of excessive memory usage here, so reloading it ~doubled that.
2. We now tell Python to try and stay under 900MiB if it can.


willdonnelly:
The print is actually the right operation here, stdout from this helper command is logged by the connector when successful.

PR #3406, opened 2025-10-17
Revert "materialize-databricks: support OAuth2 credentials"

mdibaiee:
Databrick OAuth only works in the same account and not across accounts... so we can't provide OAuth to customers. Reverts estuary/connectors#3384


Alex-Bair (approved):
LGTM

PR #3409, opened 2025-10-17
source-facebook-marketing-native: new connector

JustinASmith:
**Description:**

This is an initial implementation of a native Facebook Marketing capture connector. Long-term, this will replace the current imported Airbyte Facebook Marketing connector.

The connector includes 17 resource types organized into multiple categories:
  - Full refresh resources: ad_account, ad_creatives, custom_conversions
  - Standard incremental resources: campaigns, ad_sets, ads, activities
  - Reverse-chronological incremental resources: images, videos
  - Insights resources with various breakdowns: ads_insights (base), ads_insights_age_and_gender, ads_insights_country, ads_insights_region, ads_insights_dma, ads_insights_platform_and_device, ads_insights_action_type
  - Custom insights: user-configurable insights with custom fields, breakdowns, and action breakdowns

Key technical features include:
  - Async and generalized Client wrapper around Facebook's API that handles pagination, filtering, and other nuances of the API
  - Async job manager for Facebook's Insights API that handles job submission, polling, and result pagination with configurable retry logic
  - Comprehensive field enumerations with 600+ marketing API fields and extensive breakdown/action breakdown options for insights customization
  - Permission validation during discovery to ensure proper access to ad accounts and API endpoints
  - Support for include_deleted configuration to capture soft-deleted resources
  - Lookback window support for insights to handle eventual consistency

The connector uses Facebook Marketing API v23.0 with proper OAuth2 authentication and leverages. The connector should support future Facebook Marketing API upgrades with clearly defined models, abstractions, and separation of concern.

**Workflow steps:**

Once this PR is ready and merged there will be some follow up work to migrate existing Production capture tasks that are using the Airbyte `source-facebook-marketing` connector to use this new `source-facebook-marketing-native` connector. This will be a fully "Online-Migration" where downstream collections and their schemas, Materializations, and destination systems should not be impacted. I will work with the support team to go through this process and ensure a smooth transition for all users.

This migration process will be as follows:

1. Update existing production captures by widening their collection's inferred schemas via sourced schemas ✅ 
2. Update existing production captures to use schema inference ✅ 
3. Merge this PR
4. Create an automated script for translating existing production captures' `EndpointConfig` and `ResourceConfig` specifications from the Airbyte format to match our native format with the necessary removal of removed config options, addition of intervals, etc.
5. Test migration process locally once the "capture reset" feature is released and ready in the runtime/`flowctl`
6. Finally, work with the support team to perform the migration on a couple production tasks and ensure no issues before rolling out completely.

This PR will require documentation updates in separate follow up PRs once this PR is ready for merge:
- Deprecate Airbyte `source-facebook-marketing` connector in docs.
- Add new docs for `source-facebook-marketing-native` along with other release steps to make this connector available in the Production UI

**Notes for reviewers:**

TBD (will provide more info on testing, etc. next week)

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -52,6 +52,7 @@ on:
       - "source-dynamics-365-finance-and-operations/**"
       - "source-incident-io/**"
       - "source-navan/**"
+      - "source-facebook-marketing-native/**"
 
   pull_request:
     branches: [main]
@@ -105,6 +106,7 @@ on:
       - "source-dynamics-365-finance-and-operations/**"
       - "source-incident-io/**"
       - "source-navan/**"
+      - "source-facebook-marketing-native/**"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -305,6 +307,10 @@ jobs:
             version: v1
             usage_rate: "1.0"
           - name: source-navan
+            type: capture
+            version: v1
+            usage_rate: "1.0"
+          - name: source-facebook-marketing-native
             type: capture
             version: v1
             usage_rate: "1.0"


source-facebook-marketing-native/source_facebook_marketing_native/models.py
```diff
@@ -0,0 +1,1335 @@
...
+    include_deleted: bool = Field(
+        title="Include Deleted Records",
+        description="Set to active if you want to include data from deleted Activities",
+        default=False,
+    )
+    fetch_thumbnail_images: bool = Field(
+        title="Fetch Thumbnail Images from Ad Creative",
+        description="Set to active if you want to fetch the thumbnail_url and store the result in thumbnail_data_url for each Ad Creative.",
+        default=False,
+    )
```
Alex-Bair: Can these be tucked into an `advanced` field in the config? I usually try to put anything optional & unlikely to be set by the average user into an "Advanced" dropdown to reduce visual clutter in the UI. As long as we have sensible defaults for any advanced settings, most folks won't have to touch them.

source-facebook-marketing-native/source_facebook_marketing_native/models.py
```diff
@@ -0,0 +1,1335 @@
...
+    action_breakdowns: list[str] = Field(
+        title="Action Breakdowns",
+        description="A list of chosen action_breakdowns for action_breakdowns",
+        default=[],
+    )
+    time_increment: int = Field(
+        title="Time Increment",
+        description=(
+            "Time window in days by which to aggregate statistics. The sync will be chunked into N day intervals, where N is the number of days you specified. "
+            "For example, if you set this value to 7, then all statistics will be reported as 7-day aggregates by starting from the start_date. If the start and end dates are October 1st and October 30th, then the connector will output 5 records: 01 - 06, 07 - 13, 14 - 20, 21 - 27, and 28 - 30 (3 days only)."
```
Alex-Bair: This may be an outdated description, and if it is please ignore. But I think when we spoke yesterday, we discussed that the behavior for a 7 day `time_increment` would be 01 - 06, 02 - 07, 03 - 08, etc. That'd cover all possible `time_increment` windows users could want.

source-facebook-marketing-native/source_facebook_marketing_native/models.py
```diff
@@ -0,0 +1,1335 @@
...
+
+
+class FacebookError(BaseModel):
+    """Facebook API error response model."""
+
+    message: str
+    type: str
+    code: int
+    error_subcode: int | None = None
+    fbtrace_id: str | None = None
```
Alex-Bair: It looks to me like `FacebookError` isn't used in the standard connector files (`api.py`, `resources.py`, etc.) but rather is used by the client and job manager. What do you think about locating this model in `client.py` instead of having it live here? There are already other models related to client specific concerns in `client.py`, and `FacebookError` feels like another client specific concern.

source-facebook-marketing-native/source_facebook_marketing_native/models.py
```diff
@@ -0,0 +1,1335 @@
...
+    Used by the job manager to track job progress and completion.
+    """
+
+    id: str = Field(description="Job ID (report_run_id)")
+    async_status: str = Field(
+        description="Job status: 'Job Completed', 'Job Failed', 'Job Running', 'Job Skipped'"
+    )
+    async_percent_completion: int = Field(
+        description="Progress percentage (0-100)", ge=0, le=100
+    )
```
Alex-Bair: Similar song & dance as with `FacebookError`. What do you think about moving `FacebookInsightsJobStatus` into `job_manager.py`?

source-facebook-marketing-native/source_facebook_marketing_native/models.py
```diff
@@ -0,0 +1,1335 @@
...
+
+class AdsInsights(FacebookInsightsResource):
+    name: ClassVar[str] = "ads_insights"
+    endpoint: ClassVar[str] = "insights"
+    interval: ClassVar[timedelta] = timedelta(hours=1)
+    requires_account_id: ClassVar[bool] = True
+
+    fields: ClassVar[list[str]] = [
+        # Required fields
+        "account_id",
```
Alex-Bair: A lot of these field names are used multiple times throughout the connector. Whenever I'm working with groups of strings & I want to make sure I don't mistype one of them, I like to throw them in a `StrEnum` and reference them like `Fields.ACCOUNT_ID`. It may be more verbose, but it centralizes the spelling of the field to one spot & helps avoid runtime errors from mistyping one of these strings somewhere. You don't have to do that if you prefer not using a `StrEnum`, but it's something to consider.

source-facebook-marketing-native/source_facebook_marketing_native/resources.py
```diff
@@ -0,0 +1,476 @@
...
+            errors.append(
+                f'Custom insight "{insight.name}" has invalid level "{insight.level}". '
+                f"Valid levels are: ad, adset, campaign"
+            )
+
+    if errors:
+        raise ValidationError(errors)
+
+
+async def full_refresh_resource(
```
Alex-Bair: Do the various resource creation related functions need to be async? I'm not seeing any actual asynchronous calls in `full_refresh_resource`, `incremental_resource`, `create_insights_resource`, etc.

source-facebook-marketing-native/source_facebook_marketing_native/api.py
```diff
@@ -0,0 +1,650 @@
...
+            resource_model,
+            params,
+            account_id=account_id,
+        ):
+            thumbnail_url = getattr(creative, "thumbnail_url", None)
+
+            if thumbnail_url:
+                thumbnail_data_url = await client.fetch_thumbnail_data_url(
+                    thumbnail_url
+                )
```
Alex-Bair: Should the thumbnail URL fetching only happen if the associated setting is selected in the endpoint config?

source-facebook-marketing-native/source_facebook_marketing_native/client.py
```diff
@@ -0,0 +1,581 @@
...
+        self, log: Logger, job_id: str
+    ) -> AsyncGenerator[dict[str, Any], None]:
+        url = f"{self.base_url}/{job_id}/insights"
+        params = {"limit": DEFAULT_PAGE_SIZE}
+
+        log.debug(f"Fetching results for job: {job_id}")
+
+        while True:
+            response_bytes = await self.http.request(log, url, params=params)
+            response_data = json.loads(response_bytes)
```
Alex-Bair: Is there a Pydantic model for the response we could use here? This pattern of getting the response bytes then passing them in to `json.loads` is in a number of spots (I'm guessing Claude is partial to it), and it'd be nicer if we could use a Pydantic model & the `model_validate_json` method instead to align with how we do this elsewhere.

source-facebook-marketing-native/source_facebook_marketing_native/api.py
JustinASmith: Ah yep, thought I fed that config through and had this conditional. Thanks for the catch!

JustinASmith:
@Alex-Bair I've updated the PR with my latest work, more detailed commit and PR description. I will update the  PR description with any other notes/information along with another commit for adding to CI. Please let me know if you think this PR would be easier to review if I broke the commit down into more granular commits or anything else. I'll try to help in any way I can since this is such a large PR and first time doing an "online" migration of imported connectors.

JustinASmith:
@Alex-Bair I've addressed all of your feedback too. It may be worth chatting through that next week if it isn't immediately clear what I did. Otherwise, I will try to post follow up comments next week on what was done to address feedback!

Have a great weekend 🎉

source-facebook-marketing-native/source_facebook_marketing_native/client.py
```diff
@@ -0,0 +1,483 @@
...
+    "unique_clicks",
+    "age_targeting",
+    "gender_targeting",
+    "labels",
+    "location",
+    "estimated_ad_recall_rate_lower_bound",
+    "estimated_ad_recall_rate_upper_bound",
+    "estimated_ad_recallers_lower_bound",
+    "estimated_ad_recallers_upper_bound",
+}
```
Alex-Bair: Can `UNSUPPORTED_FIELDS` reference field names via the `Fields` enum?

source-facebook-marketing-native/source_facebook_marketing_native/client.py
```diff
@@ -0,0 +1,483 @@
...
+
+        return ",".join(fields) if fields else None
+
+    @field_serializer("after_cursor", mode="plain")
+    def serialize_after_cursor(self, after_cursor: str | None) -> str | None:
+        return after_cursor
+
+    @field_serializer("since", mode="plain")
+    def serialize_since(self, since: int | None) -> int | None:
+        return since
```
Alex-Bair: Could you please help me understand the purpose of the field serializers for `after_cursor` and `since`? I haven't used them much before, but it looks like these two don't have any special serialization behavior?

source-facebook-marketing-native/source_facebook_marketing_native/client.py
```diff
@@ -0,0 +1,483 @@
...
+                )
+                raise ValueError(
+                    f"Unexpected response type: {type(response)}; content: {raw_json}"
+                )
+
+    async def fetch_thumbnail_data_url(self, url: str) -> str | None:
+        try:
+            headers, body = await self.http.request_stream(self.log, url)
+            fallback_content_type = "image/jpeg"
+            content_type = headers.get("Content-Type", fallback_content_type)
```
Alex-Bair: Additionally, I was wondering if this lookup for a `Content-Type` header is case insensitive. I don't think it is since the CDK is converting the `CIMultiDictProxy` returned from `aiohttp` (which supports case-insensitive lookups for exactly this use case) into a basic `dict`.
https://github.com/estuary/connectors/blob/24f5ea0939c018b14142f32777f45581eee5cb5f/estuary-cdk/estuary_cdk/http.py#L586

We'll want to fix that behavior in the CDK so this lookup succeeds whether it's checking for a `Content-Type` or `content-type` header.

source-facebook-marketing-native/.vscode/launch.json
Alex-Bair: Did you mean to commit `.vscode`? We have one at the top level of the `connectors` repo already. If we want to make any changes, I'd prefer to keep everything in the single top level `.vscode/launch.json` instead of having separate files per connector (unless there's a reason to do so that I'm missing).

source-facebook-marketing-native/test.flow.yaml
```diff
@@ -0,0 +1,82 @@
...
+captures:
+  acmeCo/source-facebook-marketing-native:
+    endpoint:
+      local:
+        command:
+        - python
+        - -m
+        - source_facebook_marketing_native
+        config: config.yaml
+    bindings:
```
Alex-Bair: Since we don't/can't have a capture snapshot, it's fine to leave `bindings: []` instead of populating it with the various resources. These are only used during the capture snapshot. The same goes for everything under the `acmeCo/` directory too - they're only used for the capture snapshot, and we don't need them if there's no capture snapshot.

source-facebook-marketing-native/source_facebook_marketing_native/models.py
```diff
@@ -0,0 +1,1326 @@
...
+                        f"Coercing non-dict value {type(value).__name__} to empty dict"
+                    )
+                    return {}
+                return value
+
+            case _:
+                logger.debug(
+                    f"Unknown target type {target_type}, returning value as-is"
+                )
+                return value
```
Alex-Bair: A note here about why we're doing all this transformation would be helpful so we remember later that it's to maintain compatibility with the documents output by the imported `source-facebook-marketing` connector.

source-facebook-marketing-native/source_facebook_marketing_native/models.py
```diff
@@ -0,0 +1,1326 @@
...
+        Logger,
+        LogCursor,
+    ],
+    AsyncGenerator[FacebookResource | LogCursor, None],
+]
+
+
+class AdAccount(FacebookResource):
+    name: ClassVar[str] = ResourceName.AD_ACCOUNT
+    endpoint: ClassVar[str] = "/"
```
Alex-Bair: It seems a little off that `AdAccount`'s `endpoint` contains a leading `/` but none of the other models' endpoints do. Should this be an empty string instead to avoid having a double `//` constructed in the URL for fetching ad accounts?

source-facebook-marketing-native/source_facebook_marketing_native/models.py
```diff
@@ -0,0 +1,1326 @@
...
+        F.QUALITY_RANKING,
+        F.ENGAGEMENT_RATE_RANKING,
+        F.CONVERSION_RATE_RANKING,
+    ]
+
+    primary_keys: ClassVar[list[str]] = [
+        "/account_id",
+        "/date_start",
+        "/ad_id",
+    ]
```
Alex-Bair: For all ad insights streams (including custom ones), we should make sure the ordering of `primary_keys` elements are consistent & the same as the imported connector. We want to do that to avoid triggering dataflow/collection resets; if a collection key's elements are the same but in a different order, the runtime sees that as change to the collection key & triggers a dataflow/collection reset.

source-facebook-marketing-native/tests/__init__.py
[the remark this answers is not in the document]
Alex-Bair: Coming back to this after mentioning the primary key ordering needs to remain the same, it'd be beneficial to have a discover snapshot to easily review and catch stuff like that (primary key changes, etc.). You can see how I built an escape hatch in `source-jira-native` for some inspiration:
https://github.com/estuary/connectors/blob/24f5ea0939c018b14142f32777f45581eee5cb5f/source-jira-native/source_jira_native/resources.py#L518-L522

source-facebook-marketing-native/source_facebook_marketing_native/job_manager.py
```diff
@@ -0,0 +1,311 @@
...
+    error: FacebookError | None = None
+
+    model_config = ConfigDict(extra="allow")
+
+
+class AsyncJobStatusResponse(BaseModel):
+    """Response from checking async insights job status."""
+
+    id: str
+    async_status: str  # Job|Failed|Completed
```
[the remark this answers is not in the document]
JustinASmith: Updated this to be a bit more clear. The values  can be `Job Skipped`, `Job Failed`, etc.

source-facebook-marketing-native/source_facebook_marketing_native/models.py
```diff
@@ -0,0 +1,1326 @@
...
+    cursor_field: ClassVar[str] = "updated_time"
+    enable_deleted_filter: ClassVar[bool] = False
+    entity_prefix: ClassVar[str] = ""  # Used for API filtering (e.g., "campaign", "ad")
+    fields: ClassVar[list[str]] = []
+
+    id: str | None = Field(
+        default=None,
+        # Don't schematize the default value.
+        json_schema_extra=lambda x: x.pop("default"),  # type: ignore
+    )
```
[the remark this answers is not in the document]
JustinASmith: This was because some resources did not use  `id` as part of their PK, however, I  have moved this out to the models inheriting from this instead.  Just as  the other models were defining `primary_keys` and the fields when they were different, I opted to do the same for the other models that just have `id` as their  primary key. This is more clear.

source-facebook-marketing-native/source_facebook_marketing_native/models.py
JustinASmith: yes it should be, however, it didn't cause any issues surprisingly

source-facebook-marketing-native/source_facebook_marketing_native/job_manager.py
```diff
@@ -0,0 +1,311 @@
...
+        max_wait_time: timedelta | None = None,
+    ) -> str:
+        attempt = 0
+
+        while attempt <= self._max_retries:
+            try:
+                if attempt > 0:
+                    log.info(
+                        f"Retrying job for account {account_id} (attempt {attempt}/{self._max_retries})"
+                    )
```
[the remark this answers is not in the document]
JustinASmith: I've included `job_id` as the  "previous  job"

JustinASmith:
@Alex-Bair I will clean up the commits into 2 commits after you have a chance to review again. Not sure what happened after I rebased locally. It got my commit history way out of wack, adding `source-navan` commit etc..

estuary-cdk/estuary_cdk/flow.py
```diff
@@ -133,6 +133,12 @@ class ValidationError(Exception):
 
 
 class ResourceOwnerPasswordOAuth2Credentials(abc.ABC, BaseModel):
+    # This configuration provides a "title" annotation for the UI to display
+    # instead of the class name.
+    model_config = ConfigDict(
+        title="OAuth",
+    )
```
Alex-Bair: I like these additions! However, doing so causes some of the connectors' spec snapshot tests to fail, like `source-looker`. If you're adding these `title`s to the various OAuth classes, please update the snapshots for the tests that are failing because of the `title` change.

.github/workflows/python.yaml
```diff
@@ -52,6 +52,9 @@ on:
       - "source-dynamics-365-finance-and-operations/**"
       - "source-incident-io/**"
       - "source-navan/**"
+      - "source-facebook-marketing-native/**"
+      - "source-navan/**"
+      - "source-facebook-marketing-native/**"
```
Alex-Bair: Noting that `source-facebook-marketing-native` and `source-navan` are listed twice in this file. I'm pretty sure this will get resolved when you clean up the commits.

PR #3413, opened 2025-10-20
source-google-analytics-data-api-native: handle metric aggregations for empty date ranges

Alex-Bair:
**Description:**

When a metric aggregation is used and no rows are returned, the API sometimes returns a list with a single empty object `[{}]` instead of leaving the aggregation field absent in the response. Converting this to `None` during validation simplifies handling in the connector.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed when using a metric aggregation & there are no rows returned & the aggregation field contains `[{}]` in the API response, the connector does not fail for a Pydantic validation error like it did previously.


JustinASmith (approved):
LGTM

PR #3422, opened 2025-10-22
source-looker: add `user_attributes` and `user_attribute_values` streams

Alex-Bair:
**Description:**

Adds the `user_attributes` and `user_attribute_values` streams to `source-looker`.

**Workflow steps:**

The connector's documentation should be updated to include the two new streams.

**Notes for reviewers:**

Tested on a local stack. Confirmed both new streams capture data.

### DIFF
--- a/source-looker/source_looker/models.py
+++ b/source-looker/source_looker/models.py
@@ -197,18 +197,30 @@ class Users(LookerStream):
     path: ClassVar[str] = "users"
 
 
-class UserCredentialsEmbed(LookerStream):
+class UserCredentialsEmbed(LookerChildStream):
     name: ClassVar[str] = "user_credentials_embed"
     path: ClassVar[str] = "credentials_embed"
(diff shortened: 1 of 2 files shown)


source-looker/source_looker/models.py
```diff
@@ -197,20 +197,32 @@ class Users(LookerStream):
     path: ClassVar[str] = "users"
 
 
-class UserCredentialsEmbed(LookerStream):
+class UserCredentialsEmbed(LookerChildStream):
```
Alex-Bair: Both `UserCredentialsEmbed` and `UserRoles` should subclass `LookerChildStream` instead of `LookerStream`; I missed this when I originally created the connector. There's no actual functional change because the resource creation logic in `build_resource_if_accessible_stream` looks at the structure of `STREAMS` rather than what class a stream subclasses, and that was wired up correctly so that these two streams behave like child streams rather than non-child streams.

PR #3434, opened 2025-10-24
source-hubspot-native: add `goals` and `workflows` streams

Alex-Bair:
**Description:**

This PR's scope includes:
- Redacting the `url` field in the capture snapshot (the value of this field was changing & causing our capture snapshot test to fail).
- Adding `goals` and `workflows` streams.
  - These streams use the `crm.objects.goals.read` and `automation` scopes, respectively.

**Workflow steps:**

The connector's documentation should be updated to include the two new streams.

**Notes for reviewers:**

Confirmed:
- both new streams are conditionally discovered if the provided credentials have access to the required endpoints.
- both new streams capture the expected amount of data.

### DIFF
--- a/source-hubspot-native/source_hubspot_native/resources.py
+++ b/source-hubspot-native/source_hubspot_native/resources.py
@@ -32,6 +32,7 @@ from .api import (
     fetch_delayed_marketing_emails,
     fetch_delayed_products,
     fetch_delayed_tickets,
+    fetch_delayed_workflows,
     fetch_email_events_page,
     fetch_form_submissions,
     fetch_forms,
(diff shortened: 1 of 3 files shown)


source-hubspot-native/source_hubspot_native/resources.py
```diff
@@ -80,11 +83,13 @@
...
     ResourceState,
     Ticket,
+    Workflow,
 )
 
 
 MISSING_SCOPE_REGEX = (
-    r"This app hasn't been granted all required scopes to make this call."
+    r"This app hasn't been granted all required scopes to make this call.|"
+    r"auth request is missing required '.+' scope"
```
Alex-Bair: The `automation/v3/workflows` endpoint can return a 403 error with a response body like "EXTERNAL auth request is missing required 'workflows-access-public-api' scope.". That didn't match against the previous `MISSING_SCOPE_REGEX`, so I expanded the regex to match against those types of 403 errors too.

PR #3436, opened 2025-10-24
source-facebook-marketing: add support for emitting sourced schemas in imported Airbte connectors

JustinASmith:
**Description:**

This adds support for emitting a sourced schema message from Airbyte connectors. This PR updates the `CaptureShim` in to accept an optional `should_emit_sourced_schemas` argument which will take the schema loaded from the `schemas` directory and recursively set `addtionalProperties` to `False` before emitting the sourced schema message and checkpointing.

**Workflow steps:**

**Notes for reviewers:**

- Tested on a local stack and verified that the inferred schema without these changes for this capture connector had various `format`s added to `*_id` type fields and after this change the inferred schema did not update to include these formats.


estuary-cdk/estuary_cdk/capture/response.py
```diff
@@ -43,6 +44,13 @@ class SourcedSchema(BaseModel):
     # We use a serialization alias to avoid Pydantic's complaints and serialize the field name correctly.
     schemaJson: dict = Field(serialization_alias="schema_json")
 
+    if _PYDANTIC_VERSION == "v2":
+        model_config = {"validate_by_name": True, "validate_by_alias": True}
+    else:
+        class Config:
+            allow_population_by_field_name = True
```
Alex-Bair: nit: could we add a comment explaining why we need to support Pydantic v1 with this `if/else` statement? I understand right now Pydantic v1 compatibility is needed so imported connectors using the `CaptureShim` can use this model, but I'm confident I'll forget that context in a month.

Alex-Bair (approved):
LGTM! Just had a couple small comments / suggestions.

PR #3445, opened 2025-10-28
materialize-elasticsearch: allow setting int field as keyword

danielnelson:
**Description:**

A user may wish to index numeric types as keywords to improve lookup in term searches if they do not need to the field with range queries.

closes: https://github.com/estuary/flow/issues/2477

**Workflow steps:**

Edit the field configuration using flowctl or advanced spec editor.  Backfill will be required if there is existing data as it is not possible to change the mapping of a field.

Field configuration docs would need tweaked to mention that keyword works with int fields as well.

https://docs.estuary.dev/reference/Connectors/materialization-connectors/Elasticsearch/#field-configuration

**Notes for reviewers:**

### DIFF
--- a/materialize-elasticsearch/driver_test.go
+++ b/materialize-elasticsearch/driver_test.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"os"
 	"slices"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


materialize-elasticsearch/driver_test.go
```diff
@@ -33,6 +34,9 @@ func testConfig() *config {
 
 func TestValidateAndApply(t *testing.T) {
 	flag.Parse()
+	if os.Getenv("TEST_DATABASE") != "yes" {
```
jgraettinger: nit: left over from manual testing? Or does this serve a functional role (and if so, add a comment)?

PR #3450, opened 2025-10-28
source-snowflake: Fallback collection key

willdonnelly:
**Description:**

Adds a fallback collection key for discovered Snowflake collections when the source table has no primary key. Previously no collection key was specified, because that used to be the only way we could let users set their own key in the UI.

Also includes an unrelated change to the endpoint spec snapshot, I assume somebody made changes to how login credentials work and just didn't update the snapshot to match.


jgraettinger (approved):
LGTM

PR #3473, opened 2025-11-03
source-postgres-batch: Chunked CTID scanning

willdonnelly:
**Description:**

Back in June we modified source-postgres so that instead of using a `WHERE ctid > $1 LIMIT k` query, keyless backfills would instead use `WHERE ctid > $1 AND ctid <= $2` with an explicit upper bound. This was done because in certain edge cases the former query may not actually be guaranteed to return results in ascending CTID order.

(If I recall correctly, there are two scenarios where this may be an issue. One is partitioned tables, and the other is PostgreSQL versions older than...I think v14? Regardless of the precise number the point is that the TID range scan query plan didn't always exist)

Anyway, we didn't make the analogous change to source-postgres-batch because the unbounded "just give me everything" query made that more difficult and we thought it was less pressing since it was more likely that we'd just backfill the whole table in a single query. This was probably unreasonably optimistic and we're seeing some issues with a production capture which may be explained by this.

So this commit modifies the XMIN-incremental polling behavior so that instead of issuing a single unbounded query we now do the same thing as source-postges and issue a stream of queries for individual chunks defined by a lower and an upper CTID bound, using the upper bound as the resume cursor for the next chunk instead of a result row, and determine that we've scanned the entire table when the upper-bound CTID grows larger than the table's estimated size.

Since the SQL batch connectors do not currently have any sort of chunk size tuning and in any event the tuning only has a very approximate relationship to CTID ranges, I have simply gone ahead and hardcoded that to the equivalent of 200k (which in practice could mean anywhere from 20k to 200k rows per chunk on typical tables).

**Workflow steps:**

In theory (and assuming I didn't screw anything up that the tests missed) this change should be unnoticeable to users, it's just improving the reliability of the backfill process in certain edge cases around task restarts.

### DIFF
--- a/source-boilerplate/testing/log.go
+++ b/source-boilerplate/testing/log.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"path/filepath"
+	"sort"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #3475, opened 2025-11-04
source-sqlserver: Query cdc.lsn_time_mapping directly

willdonnelly:
**Description:**

Previously we have used the `sys.fn_cdc_get_max_lsn()` helper function to query the maximum CDC LSN on a SQL Server database. But that function is literally just a wrapper around a `SELECT MAX(start_lsn) FROM cdc.lsn_time_mapping` query so it's not really doing anything we need, and now we have reason to suspect that this function is being repeatedly recompiled and that this compilation / caching is actually causing a significant amount of lock contention in a production capture.

To the best of my knowledge this function can trivially be swapped out for the aforementioned `MAX(start_lsn)` query in every version of SQL Server that we support, so this should be a risk-free change.


Alex-Bair (approved):
LGTM

PR #3480, opened 2025-11-05
materialize-sqlserver: use string column size info

danielnelson:
**Description:**

Create string columns with only enough size for the collection string.  If no size is provided then a MAX size varchar is created.  The `maxLength` in the schema is validated as the number of codepoints, not bytes, so when creating a column we reserve 4 bytes for every 1 length if creating a varchar column, or 2 bytes for nvarchar.  Note that with nvarchar the size is specified in byte-pairs.  This is almost certainly a massive over count for varchar, since very few characters would use the full 4 bytes.

A string column is compatible with columns with at least enough space as required by the schema.

Otherwise column can be migrated to any larger columns size.  Migration used the current strategy of creating a new column, copying the data, and renaming to the original name.  It may be faster to do an in-place alter column but this may also not scale very well for larger tables.

closes: #3387

**Workflow steps:**

No changes

Need to document the size factor.

**Notes for reviewers:**

My main concern with this PR is that we now require 4x the string maxLength.  If a user has manually altered columns to be less than this then they will be sized up.  I think we could also get into trouble with the primary key, which is limited to 900 bytes.

### DIFF
--- a/materialize-boilerplate/validate.go
+++ b/materialize-boilerplate/validate.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"slices"
+	"strconv"
 	"strings"
 
 	pf "github.com/estuary/flow/go/protocols/flow"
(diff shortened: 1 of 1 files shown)


mdibaiee (approved):
LGTM! thank you

PR #3499, opened 2025-11-14
source-outreach: reduce page size after 502 Bad Gateway response

Alex-Bair:
**Description:**

Requesting the `MAX_PAGE_SIZE` sometimes causes the Outreach API server to take too long to respond, and the connector ends up receiving a 502 Bad Gateway response instead (I assume from an intermediate gateway). When this happens, the connector needs to retry the same request but with a smaller page size. Eventually, the requested page size will be small enough that the Outreach API will respond in a reasonable amount of time.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed that after receiving a 502 Bad Gateway response, the connector retries the same request with a smaller page size, & that using a smaller page size eventually causes the connector to receive a 200 response from the API server.

### DIFF
--- a/source-outreach/source_outreach/api.py
+++ b/source-outreach/source_outreach/api.py
@@ -1,10 +1,10 @@
 from datetime import datetime
 from logging import Logger
-from typing import AsyncGenerator
+from typing import Any, AsyncGenerator
 from urllib.parse import urlparse, parse_qs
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM

PR #3503, opened 2025-11-16
estuary-cdk: refactor the placement of client creds in OAuth requests

nicolaslazo:
**Description:**

So far, each OAuth flavour in the CDK had a hardcoded place where OAuth client credentials were supposed to go -- either the headers, or the request body. However, in the process of developing the QuickBooks connector (#3468) we realised its rotating tokens system broke the assumed standard: client ids and secrets were to be concatenated and base64-encoded before being placed in the headers.

This PR adds a new class method to our OAuth models: `with_client_credentials_placement`, which allows for a builder pattern-like interface to override what we'd used as the standard so far. This means no pre-existing connector code needed to be updated.

Two smaller changes were effected too
* `OAuth2RotatingTokenSpec` got merged into `OAuth2Spec` and now the latter supports the `additionalTokenExchangeBody` field
* Redundancy in the way we construct `headers` and `form` dictionaries has been simplified

**Workflow steps:**

**Notes for reviewers:**

Changes have been tested on the following connectors:
* HubSpot
* Salesforce
* Outreach
* Genesys
* QuickBooks

### DIFF
--- a/estuary-cdk/estuary_cdk/flow.py
+++ b/estuary-cdk/estuary_cdk/flow.py
@@ -210,9 +210,6 @@ class ResourceOwnerPasswordOAuth2Credentials(_BaseOAuth2CredentialsData):
 
 class ClientCredentialsOAuth2Credentials(_BaseOAuth2CredentialsData):
     grant_type: ClassVar[str] = "client_credentials"
-    client_credentials_placement: ClassVar[OAuth2ClientCredentialsPlacement] = (
-        OAuth2ClientCredentialsPlacement.HEADERS
-    )
 
 
 class AuthorizationCodeFlowOAuth2Credentials(
--- a/source-genesys/source_genesys/models.py
+++ b/source-genesys/source_genesys/models.py
@@ -1,16 +1,19 @@
 from datetime import datetime, timezone, timedelta
 from enum import StrEnum
 from pydantic import AwareDatetime, BaseModel, Field
-from typing import ClassVar, Literal, Optional
+from typing import TYPE_CHECKING, ClassVar, Literal, Optional
 
 
 from estuary_cdk.capture.common import (
     BaseDocument,
     ConnectorState as GenericConnectorState,
-    ClientCredentialsOAuth2Credentials,
-    OAuth2TokenFlowSpec,
     ResourceConfig,
     ResourceState,
+)
+from estuary_cdk.flow import (
+    ClientCredentialsOAuth2Credentials,
+    OAuth2ClientCredentialsPlacement,
+    OAuth2TokenFlowSpec,
 )
 
 
@@ -23,6 +26,14 @@ OAUTH2_SPEC = OAuth2TokenFlowSpec(
     }
 )
 
+if TYPE_CHECKING:
+    OAuth2Credentials = ClientCredentialsOAuth2Credentials
+else:
+    OAuth2Credentials = (
+        ClientCredentialsOAuth2Credentials.with_client_credentials_placement(
+            OAuth2ClientCredentialsPlacement.HEADERS
+        )
+    )
 
 def default_start_date():
     dt = datetime.now(timezone.utc) - timedelta(days=30)
@@ -54,7 +65,7 @@ class EndpointConfig(BaseModel):
     ] = Field(
         title="Genesys Cloud Domain"
     )
-    credentials: ClientCredentialsOAuth2Credentials = Field(
+    credentials: OAuth2Credentials = Field(
         title="Authentication",
         discriminator="credentials_title"
     )
--- a/source-navan/source_navan/models.py
+++ b/source-navan/source_navan/models.py
@@ -1,4 +1,5 @@
 from typing import (
+    TYPE_CHECKING,
     Generic,
     TypeVar,
 )
@@ -12,6 +13,7 @@ from estuary_cdk.capture.common import (
 )
 from estuary_cdk.flow import (
     ClientCredentialsOAuth2Credentials,
+    OAuth2ClientCredentialsPlacement,
     OAuth2TokenFlowSpec,
 )
 from pydantic import (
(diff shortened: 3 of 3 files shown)


estuary-cdk/estuary_cdk/flow.py
```diff
@@ -13,6 +13,9 @@
     "LOCAL",  # We're running directly on the host as a local process.
 ]
 
+# Where to include specific request data
+RequestDataPlacement = Literal["headers", "form"]
```
Alex-Bair: nit: Can the comment and variable name be a little more descriptive so it's clearer what this represents? Maybe something like `OAuth2ClientCredentialsLocation` and `# Location of client id and client secret during the OAuth2 token exchange step`?

Alex-Bair: Also, what are your thoughts on making this a `StrEnum` instead of a `Literal`? Connector developers could then pick an option from enum to use in their connector instead of figuring out the correct string they should use (`"headers"` vs `"header"` vs `"Headers"` etc.).

estuary-cdk/estuary_cdk/http.py
```diff
@@ -314,11 +335,14 @@ async def _fetch_oauth2_token(
...
-        | AuthorizationCodeFlowOAuth2Credentials
-        | RotatingOAuth2Credentials,
+        credentials: (
+            BaseOAuth2Credentials
+            | ResourceOwnerPasswordOAuth2Credentials
+            | ClientCredentialsOAuth2Credentials
+            | AuthorizationCodeFlowOAuth2Credentials
+            | RotatingOAuth2Credentials
+        ),
+        client_creds_request_location: Literal["headers"] | Literal["form"],
```
Alex-Bair: In a formatting commit, I wouldn't expect there to be new additions or functional code changes. The `client_creds_request_location` param here is brand new, and if we reverted back to this commit thinking there were only formatting changes, things would break since invocations of `_fetch_oauth2_token` aren't providing a `client_creds_reqeust_location` argument.

Alex-Bair (commented):
Had a few small comments.

Additionally whenever I make changes that affect multiple connectors, I make sure that the associated tests are passing in the CI checks. That helps put me a little at ease & makes it easier to claim that my CDK level changes didn't break individual connectors. It looks like the `source-zendesk-support-native` tests are failing, so we should figure out why they're failing & get them passing before merging this PR.

estuary-cdk/estuary_cdk/flow.py
Alex-Bair: That's a reasonable assumption! I think for values internal to the CDK, we've used `Literal`s, but since it's expected individual connectors could specify these values, a `StrEnum` makes sense to me. Plus whenever I have a `match`/`case` that makes decisions based on a finite set of possible values (like you have in `_fetch_oauth2_token`), it's nice to use the enum's values in the `case`s.

estuary-cdk/estuary_cdk/http.py
```diff
@@ -345,53 +346,43 @@ async def _fetch_oauth2_token(
...
+                        f"{credentials.client_id}:{credentials.client_secret}".encode()
+                    ).decode()
+                )
+            case OAuth2ClientCredentialsPlacement.FORM:
+                form.update(
+                    {
+                        "client_id": credentials.client_id,
+                        "client_secret": credentials.client_secret,
+                    }
+                )
```
Alex-Bair: When doing a `match`/`case` statement that checks if a value is part of an enum, I like to include an [exhaustiveness check](https://typing.python.org/en/latest/guides/unreachable.html). This helps me catch that the `match`/`case` needs updated when a new member is added to an enum.

While we may not ever need to add a new member to the `OAuth2ClientCredentialsPlacement` enum, an exhaustiveness check here would be beneficial in case we do expand the enum later. You could use `assert_never` like mentioned in those Python docs, or you could raise an error like we do in the `match`/`case` a few lines below this point.

estuary-cdk/estuary_cdk/flow.py
```diff
@@ -79,10 +92,7 @@ class OAuth2Spec(BaseModel):
     accessTokenHeaders: dict[str, str]
     accessTokenResponseMap: dict[str, str]
     accessTokenUrlTemplate: str
-
-
-class OAuth2RotatingTokenSpec(OAuth2Spec):
-    additionalTokenExchangeBody: dict[str, str | int] | None
+    additionalTokenExchangeBody: dict[str, str | int] = Field(default={}, exclude=True)
```
Alex-Bair: Nice use of `exclude=True` to prevent `additionalTokenExchangeBody` from being serialized in the connector's `Spec` response. Can you also add a comment here explaining why we need to exclude `additionalTokenExchangeBody` from serialization? A short summary of what I mentioned in this [comment](https://github.com/estuary/connectors/pull/3503#discussion_r2546543283) would be good IMO.

estuary-cdk/estuary_cdk/flow.py
```diff
@@ -181,82 +161,69 @@ class AuthorizationCodeFlowOAuth2Credentials(abc.ABC, BaseModel):
...
+            cls.__name__,
+            (cls,),
+            {
+                "model_config": ConfigDict(
+                    json_schema_extra={"x-oauth2-provider": provider},
+                    title="OAuth",
+                ),
+                "_you_must_build_oauth2_credentials_for_a_provider": lambda _: None,
+            },
+        )
```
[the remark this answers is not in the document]
Alex-Bair: Hm, I'm not extremely fond of using `cast()` unless there aren't any better options. `cast()` doesn't verify the cast is valid. It tells `mypy` to blindly trust us, and I'd like to keep as many type checking guardrails in place as possible.

What do you think about this approach instead? Double check me, but I think it keeps the expressiveness with the `__name__` and `__qualname__` replacements:
```python
    @classmethod
    def for_provider(cls, provider: str) -> type[Self]:
        from pydantic import ConfigDict

        class _ProviderCredentials(cls):
            model_config = ConfigDict(
                json_schema_extra={"x-oauth2-provider": provider},
                title="OAuth",
            )

            def _you_must_build_oauth2_credentials_for_a_provider(self) -> ...

        _ProviderCredentials.__name__ = cls.__name__
        _ProviderCredentials.__qualname__ = cls.__qualname__

        return _ProviderCredentials
```

PR #3521, opened 2025-11-21
source-monday: decrease request limits

JustinASmith:
**Description:**

Decreases request limits to see if this helps a customer's capture with API timeout errors. I am unable to test this with a test account with the same size data and other limitations/factors.

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM as an experiment to see if the timeouts are related to requesting too much data.

If requesting less data does end up avoiding those `REQUEST_TIMEOUT` errors, it's worth considering dynamically reducing the page size / limit when we receive them instead of reducing the hardcoded page size / limit. We're able to surface specific `5xx` errors from the CDK into the connector now, and I've used that capability before to address extremely similar timeouts in other connectors. For example, https://github.com/estuary/connectors/pull/3499.

PR #3529, opened 2025-11-25
source-mongodb: fix race condition checking if streams are caught up

danielnelson:
**Description:**

Each time `startCatchingUp` was called, the done broadcast channel would be recreated.  If another goroutine was already waiting on this channel then it would never be closed.  Since we call the startCatchingUp function on a ticker, not synchronously with waiting, this was a race condition.

With this change the channels are created on demand, and stored so that the next time all streams are caught up they will be closed.

**Workflow steps:**

No changes

None

**Notes for reviewers:**

Switched to using testify for all checks in the test because t.Error does not fail immediately.  This could cause a later error to be reported instead of an earlier one.  Generally I prefer require/fatal errors for tests.


psFried (approved):
LGTM

PR #3531, opened 2025-11-25
source-facebook-marketing-native: Handle `5xx` errors by limiting retry attempts to 5

JustinASmith:
**Description:**

Facebook's Ads Insights API can return transient `5xx` errors under load or when requesting large amounts of data. We should retry these requests, but only a limited number of times to avoid a connector deadlock. I've limited these retries on the `client.py` class only since `permissions.py` `self.http.request` are for simple permission based information and shouldn't give us trouble in practice.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-facebook-marketing-native/source_facebook_marketing_native/client.py
+++ b/source-facebook-marketing-native/source_facebook_marketing_native/client.py
@@ -6,7 +6,7 @@ from enum import Enum
 from logging import Logger
 from typing import AsyncGenerator, Generic, TypeVar
 
-from estuary_cdk.http import HTTPSession
+from estuary_cdk.http import HTTPSession, Headers
 from pydantic import (
     BaseModel,
     ConfigDict,
@@ -22,6 +22,8 @@ from .permissions import PermissionManager
 from .fields import DELIVERY_INFO_FIELDS
 from .enums import Field
 
+
+MAX_500_ERROR_RETRY_ATTEMPTS = 5
 DEFAULT_PAGE_SIZE = 100
 UNSUPPORTED_FIELDS = {
     Field.UNIQUE_CONVERSIONS,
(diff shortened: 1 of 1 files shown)


source-facebook-marketing-native/source_facebook_marketing_native/client.py
```diff
@@ -338,7 +357,12 @@ async def fetch_all(
 
         while True:
             response: FacebookResourceType = await self._fetch_with_error_handling(
-                resource_model, adapter, url, request_params, include_deleted, account_id
+                resource_model,
+                adapter,
+                url,
+                request_params,
+                include_deleted,
+                account_id,
```
JustinASmith: Formatting artifacts - non-functional change.

source-facebook-marketing-native/source_facebook_marketing_native/client.py
```diff
@@ -203,6 +206,21 @@ def model_dump(self, **kwargs) -> dict:
...
+    status: int,
+    headers: Headers,
+    body: bytes,
+    attempt: int,
+) -> bool:
+    # Facebook's Ads Insights API can return transient 500 errors under load or when
+    # requesting large amounts of data. We should retry these requests, but only a limited
+    # number of times to avoid a connector deadlock.
+    if status >= 500 and attempt < MAX_500_ERROR_RETRY_ATTEMPTS:
+        return True
```
Alex-Bair: note: as the CDK is written now, `status` will always be `>= 500`  here since `should_retry` functions are only invoked when we receive a response with a `5xx` or greater status ([ref](https://github.com/estuary/connectors/blob/b078de09574a15a210c650695d3ee62a01e0658a/estuary-cdk/estuary_cdk/http.py#L555-L561)). Not something that has to be changed here, but I wanted to mention it for awareness.

Alex-Bair (approved):
LGTM

PR #3532, opened 2025-11-25
source-facebook-marketing-native: Remove `failed_delivery_checks` field from `ad_account` stream

JustinASmith:
**Description:**

This field was causing HTTP `500` responses and it was not clear why. Removing this until the field is needed or we can figure out the root cause. Regardless of how the field was requested in the `ad_account` request (by itself or with other fields) produced this error and FB's API response was empty other than the HTTP status.

**Workflow steps:**

**Notes for reviewers:**


Alex-Bair (approved):
LGTM.

Mentioning this for documentation & linking purposes, this problem with `failed_delivery_checks` is very similar to the one with `wish_bid` described in https://github.com/estuary/connectors/pull/2274. I wouldn't be surprised if we encounter similar situations with other fields later.

PR #3551, opened 2025-12-02
source-postgres: Request keepalives

willdonnelly:
**Description:**

I thought we were already doing this, but apparently not. I'm pretty sure this is a low risk change and it might help with some very frustrating replication connection timeouts.


Alex-Bair (approved):
LGTM

PR #3564, opened 2025-12-05
materialize-mysql: check for string before resizing varchar

danielnelson:
It is possible that the value is nil here, in that case we can skip any resizing.


jgraettinger (approved):
LGTM

PR #3594, opened 2025-12-07
materialize-motherduck: support azure blob staging

jameswinegar:
**Description:**

Add Azure Blob Storage as a Staging Bucket option for MotherDuck

https://docs.estuary.dev/reference/Connectors/materialization-connectors/motherduck/

**Notes for reviewers:**

Did not test flow with this. 

Test cases were auto-generated and might be too verbose. I've had annoyances with naming conventions in Azure before so decided to add those.

related: #3496

### DIFF
--- a/materialize-motherduck/azure_test.go
+++ b/materialize-motherduck/azure_test.go
@@ -1,0 +1,288 @@
+package main
+
+import (
+	"encoding/base64"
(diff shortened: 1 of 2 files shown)


danielnelson (approved):
Awesome, thank you!

PR #3595, opened 2025-12-07
materialize-motherduck: add r2 support for staging bucket

jameswinegar:
**Description:**

Add R2 and other S3 compatible storage support via endpoint. 

**Workflow steps:**

Should add an optional endpoint field to the config page.

https://docs.estuary.dev/reference/Connectors/materialization-connectors/motherduck/

**Notes for reviewers:**

Extended existing S3 support to allow for optional endpoint. Not tested in flow.

related: #3496

### DIFF
--- a/tests/materialize/custom-configs/.gitignore
+++ b/tests/materialize/custom-configs/.gitignore
@@ -1,4 +1,0 @@
-*
-!.gitignore
-!.gitkeep
-


materialize-motherduck/config.go
```diff
@@ -173,13 +175,27 @@ func (c *config) db(ctx context.Context) (*stdsql.DB, error) {
...
+			// Use path-style URLs since most S3-compatible services require this.
+			createTempSecret = fmt.Sprintf(`CREATE SECRET IF NOT EXISTS (
+				TYPE S3,
+				KEY_ID '%s',
+				SECRET '%s',
+				REGION '%s',
+				ENDPOINT '%s',
+				URL_STYLE 'path',
+				SCOPE 's3://%s'
+			);`, c.StagingBucket.AWSAccessKeyID, c.StagingBucket.AWSSecretAccessKey, c.StagingBucket.Region, c.StagingBucket.Endpoint, c.StagingBucket.BucketS3)
```
danielnelson: It appears the value for `ENDPOINT` should be just the host portion of the URL, without the leading `https://`.
```suggestion
            endpointURL, err := url.Parse(c.StagingBucket.Endpoint)
            if err != nil {
                return nil, err
            }
			// Custom endpoint for S3-compatible storage (e.g. Cloudflare R2).
			// Use path-style URLs since most S3-compatible services require this.
			createTempSecret = fmt.Sprintf(`CREATE SECRET IF NOT EXISTS (
				TYPE S3,
				KEY_ID '%s',
				SECRET '%s',
				REGION '%s',
				ENDPOINT '%s',
				URL_STYLE 'path',
				SCOPE 's3://%s'
			);`, c.StagingBucket.AWSAccessKeyID, c.StagingBucket.AWSSecretAccessKey, c.StagingBucket.Region, endpointURL.Host, c.StagingBucket.BucketS3)
```

PR #3596, opened 2025-12-08
source-mongodb: prefetch batches and use Rust bson-transcoder

mdibaiee:
**Description:**

Two main changes here:

1. Prefetch MongoDB cursor while processing documents to avoid serial network blocks. Each MongoDB cursor batch is a maximum of 16MB, so we hold up to 4 batches at the same time, consuming 64MB of memory (second commit can be reviewed separately)
2. Implement a new Rust BSON "transcoder" (first and third commits, best to review the final state of things as I have gone back and forth on a lot of things)

Transcoding is the process of converting data from one format to another without an intermediate layer. In our case, we transform directly from BSON to JSON using a custom JSON serializer implementation (our implementation wraps `serde_json::Serializer` with custom sanitization logic). According to my benchmarks BSON transcoding in Rust can be up to 3x faster than `bson.Unmarshal -> json.Marshal` in Go (see https://github.com/mdibaiee/bson-benchmarks for raw benchmarks)

This new transcoder is used both for Change Stream events and for backfills / incremental batch queries.

Here are my benchmarking results of running the connector against a MongoDB M30 cluster:

|           | 250 bytes documents | 20KB documents |
|------------|---------------------|----------------|
| Prior           | 6.5MB/s                    |    34MB/s          |
| Optimised  | 17MB/s                     |     57MB/s          |

Graph for 250 bytes documents. I first run the connector with `v2` image, then switch to this branch's latest build, and the throughput goes from 6.5MB/s to ~17.5MB/s.
image

Graph for 20KB documents (ran out of data to capture towards the end, hence the dip)

image

Note that we are still CPU-bound in the connector, so there is still room to improve.

**Workflow steps:**

**Notes for reviewers:**


source-mongodb/change_stream.go
```diff
@@ -170,16 +154,44 @@ func (c *capture) streamChanges(
...
+				if ev.err != nil {
+					return fmt.Errorf("change stream for %q: %w", s.db, ev.err)
+				}
+
+				opTime, _, err := c.processEvent(groupCtx, s, ev)
+				if err != nil {
+					return fmt.Errorf("processing event for %q: %w", s.db, err)
+				}
+
+				if catchup && coordinator.gotCaughtUp(s.db, opTime) {
```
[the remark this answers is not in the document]
mdibaiee: Good point, fixed!

bson-transcoder/src/serializer.rs
```diff
@@ -0,0 +1,1056 @@
...
+        let result = transcode_bson_to_sanitized_json(&doc, false);
+        let parsed: serde_json::Value = serde_json::from_str(&result).unwrap();
+        assert_eq!(parsed["refs"][0], "507f1f77bcf86cd799439011");
+        assert_eq!(parsed["refs"][1], "507f1f77bcf86cd799439012");
+    }
+
+    // ==========================================================================
+    // id_to_string tests (moved from sanitize.rs)
+    // ==========================================================================
+
```
danielnelson: What sanitize.rs?

bson-transcoder/src/serializer.rs
```diff
@@ -0,0 +1,1056 @@
...
+    use super::*;
+
+    fn serialize_to_string<T: Serialize>(value: &T, is_root: bool) -> String {
+        let mut buf = Vec::new();
+        let mut serializer = SanitizingSerializer::new(&mut buf, is_root);
+        value.serialize(&mut serializer).unwrap();
+        String::from_utf8(buf).unwrap()
+    }
+
+    /// Test integration with actual BSON transcoding
```
danielnelson: This comment doesn't really match the function, could probably just remove it since the function name is fine.

PR #3608, opened 2025-12-10
materialize-snowflake: retry load queries

danielnelson:
**Description:**

If an error occurs while running a load query, we would return an error and start over on the next run.  This can be a problem as you scale up the size of the transaction, as it becomes more likely for an error to occur and more expensive to rerun the transaction from the start.

In order to retry a query after it fails, we need to resume at the same position without loading documents multiple times.  I added an ORDER BY clause to ensure the results have the same ordering, and a row number so we can skip over results we already processed.

**Workflow steps:**

None

**Notes for reviewers:**

### DIFF
--- a/materialize-snowflake/.snapshots/TestSQLGeneration
+++ b/materialize-snowflake/.snapshots/TestSQLGeneration
@@ -97,7 +97,9 @@ ALTER TABLE "a-schema".key_value ALTER COLUMN
 --- End alter table drop not nulls ---
 
 --- Begin "a-schema".key_value loadQuery ---
-SELECT 0, TO_JSON("a-schema".key_value.flow_document)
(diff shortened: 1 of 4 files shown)


danielnelson:
It looks like we are well into processing a response by the time it fails.
```
querying Loads(5811973): chunk fetch: copying chunk stream: decoding row: read tcp ...: read: connection timed out
```

I'm not crazy about this change because if there are any out-of-band additions/removals it could result in missing data or duplicates load documents.  Retrying from a staged file, if even required, would be safer.  This would assume that these types of errors come up less frequently when writing to stage, which I would believe but I don't think it's a sure bet.

I think there is also a case for not attempting this at all and requiring smaller transactions that can be fully retried.

PR #3627, opened 2025-12-16
source-hubspot-native: use new sorted list memberships endpoint

nicolaslazo:
**Description:**

The `contact_list_memberships` stream has suffered of poor performance due to the only usable endpoint not being either sortable nor filterable, which forced us to traverse the entire dataset for incremental changes. However, Hubspot has recently exposed a new API endpoint (`/crm/v3/lists/{list_id}/memberships/join-order`) which allows us to sort by join order. This should be a vast improvement in incremental fetches' execution time.

**Workflow steps:**

**Notes for reviewers:**

Concurrency semaphores of 20 and 50 have been tested, with 20 showing more favourable results. Request delays tend to stay around ~1.4 seconds and peak at 1.7.

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api.py
+++ b/source-hubspot-native/source_hubspot_native/api.py
@@ -1746,8 +1746,8 @@ async def _request_contact_list_memberships(
     start: datetime,
     end: datetime | None = None,
 ) -> AsyncGenerator[ContactListMembership, None]:
-    url = f"{HUB}/crm/v3/lists/{contact_list.listId}/memberships"
-    params = {"limit": 250}
+    url = f"{HUB}/crm/v3/lists/{contact_list.listId}/memberships/join-order"
+    params: dict[str, int | str] = {"limit": 250}
 
     while True:
         response = ContactListMembershipResponse.model_validate_json(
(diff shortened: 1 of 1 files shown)


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -1756,16 +1756,16 @@ async def _request_contact_list_memberships(
         )
 
         for item in response.results:
-            if item.membershipTimestamp < start:
-                continue
             if end and item.membershipTimestamp >= end:
                 continue
+            if item.membershipTimestamp < start:
```
Alex-Bair: nit: Could we add a comment here or maybe at the top of the function stating that the way we're using the `/join-order` endpoint will return memberships in reverse chronological `membershipTimestamp` order, and that's why it's alright to end pagination as soon as `item.membershipTimestamp < start`? Since there's no query param that makes it obvious what the returned order is, a comment like that would be helpful.

Alex-Bair (approved):
LGTM % a couple small nits. Thanks! Excited to see how much quickly we'll be able to perform incremental sweeps once this change rolls out.

PR #3634, opened 2025-12-16
source-postgres-batch: Truncate text/byte/json columns

willdonnelly:
**Description:**

Copying over the same sort of oversized-column-value truncation logic which we already have in source-postgres.

It feels like there's a more clever solution which would hit all the batch captures at once, involving making the JSON encoder aware of the maximum size of an object and having it be able to "rewind" or "chop off" part of the buffer after serializing a value which puts it over the limit, but the devil's in the details there and it's not entirely clear how we'd identify strings versus JSON objects versus arrays or whatever and so the only _elegant_ solution along those lines would be to omit oversized fields entirely, so for now just implementing appropriate truncation for various column types seemed like the way to go.

### DIFF
--- a/source-postgres-batch/capture_test.go
+++ b/source-postgres-batch/capture_test.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


dyaffe (approved):
LGTM

PR #3637, opened 2025-12-17
source-chargebee-native: Handle Eventual Consistency

JustinASmith:
**Description:**

This PR adds eventual consistency handling to source-chargebee-native's incremental sync streams, following Chargebee's [documented recommendations](https://apidocs.chargebee.com/docs/api/list-ops) for incremental syncing. The changes ensure that recently updated records (which may not yet be available due to eventual consistency) are not missed during incremental syncs.

I've made the following changes into a handful of commits to try to break up some of the changes and their reasons.

Key changes:
- **Added `LAG` constant (5 minutes)**: Incremental queries now exclude records updated in the last 5 minutes, matching Chargebee's documented eventual consistency window
- **Added `MIN_INCREMENTAL_INTERVAL` (1 minute)**: Prevents unnecessary API calls when the time window is too small
- **Added empty result handling**: When no results are found in a time window, the cursor advances to `end_date` to avoid repeatedly querying the same empty range
- **Simplified date filtering**: Removed `[before]` and `[after]` filters in favor of exclusively using the `[between]` filter, which is inclusive and easier to reason about
- **Added validation**: Start and end dates must now be specified together, and cannot be equal
- **Added `limit` field validation**: EndpointConfig limit field now has an upper bound of 100 (Chargebee's API maximum)

**Workflow steps:**

No changes to user workflow. The connector operates the same way but now:
1. Incremental syncs query data up to 5 minutes in the past (instead of up to the current moment)
2. If the cursor is within 1 minute of current time, the sync waits rather than making a zero-window query
3. When a time window has no data, the cursor advances past it instead of re-querying on each poll
4. This applies to all incremental streams using the `[between]` filter on `updated_at` or similar cursor fields

None - this is an internal implementation change that doesn't affect the connector's external behavior or configuration options.

**Notes for reviewers:**

- Validated on a local stack that backfill and incremental sync work as intended.

### DIFF
--- a/source-chargebee-native/source_chargebee_native/api.py
+++ b/source-chargebee-native/source_chargebee_native/api.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta, UTC
 from logging import Logger
-from typing import AsyncGenerator, TypeVar
+from typing import AsyncGenerator, TypeVar, Literal
 
 import asyncio
 
@@ -16,9 +16,13 @@ from source_chargebee_native.models import (
     AssociationConfig,
 )
 
-# We cannot send a request with start and end dates that are the same, so
-# we enforce a minimum interval of 1 second between them.
-MIN_INCREMENTAL_INTERVAL = timedelta(seconds=1)
+# Chargebee's API has eventual consistency delays documented to be up to 5 minutes.
+# See: https://apidocs.chargebee.com/docs/api/list-ops
+LAG = timedelta(minutes=5)
+
+# Minimum interval between start and end dates for incremental queries.
+# This prevents us from making unnecessary API calls when the time window is too small.
+MIN_INCREMENTAL_INTERVAL = timedelta(minutes=1)
 
 ChargebeeResourceType = TypeVar("ChargebeeResourceType", bound=ChargebeeResource)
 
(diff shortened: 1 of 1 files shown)


source-chargebee-native/source_chargebee_native/api.py
JustinASmith: Something I may need to change. Possibly change to:

```python
if has_results:
    yield _dt_to_ts(max_updated_at + timedelta(seconds=1))
else:
    yield _dt_to_ts(end_date)
```

This would ensure we make progress forward and not get stuck waiting for data for a period of time we may never see data for. The interval will be capped at 30 days, so we could get stuck in a 30-day window where no data was changed on a account that is updated infrequently.

JustinASmith:
@Alex-Bair  Please note, that I just found where the documentation recommends a 5-minute lag (end of this page: https://apidocs.chargebee.com/docs/api/list-ops?lang=curl) - missed this before.

So I will update the LAG to 5 minutes.

JustinASmith:
I am also missing `sort_by` which is a bug - causing inconsistent sorting in paginated results. I will update this PR with that fix too.

source-chargebee-native/source_chargebee_native/api.py
```diff
@@ -208,12 +220,15 @@ async def fetch_resource_changes(
     assert isinstance(log_cursor, int)
 
     start_date = _ts_to_dt(log_cursor)
-    end_date = min(start_date + timedelta(days=30), datetime.now(tz=UTC))
+    # Apply LAG to handle eventual consistency - don't fetch records more recent than LAG ago
+    now_with_lag = datetime.now(tz=UTC) - LAG
+    end_date = min(start_date + timedelta(days=30), now_with_lag)
```
JustinASmith: FYI @Alex-Bair  - I will be force pushing to change the places where we use an arbitrary 30 day window for when the `log_cursor` is in the past due to a capture being disabled and re-enabled at a later date or some other reason. This has shown to take quite some time for large streams like `Transactions` and `Invoices`.

I will set this to: `end_date = min(start_date + MIN_INCREMENTAL_INTERVAL, now_with_lag)` so that we make incremental progress instead. This could have some impact on performance (querying smaller ranges), so I will try to monitor that.

source-chargebee-native/source_chargebee_native/api.py
```diff
@@ -44,9 +48,17 @@ async def _fetch_resource_data(
     resource_type: type[ChargebeeResourceType] = ChargebeeResource,
     filter_params: dict[str, str] | None = None,
     returns_list: bool = True,
+    sort_by: str | None = None,
+    sort_by_order: Literal["asc", "desc"] = "asc",
```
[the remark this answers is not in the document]
Alex-Bair: Great, thanks! Makes sense that since we're not trying to checkpoint in the middle of paging through results, it should be fine that the results haven't been ordered.

PR #3640, opened 2025-12-18
source-mongodb: lossy handling of utf-8 characters

mdibaiee:
**Description:**

In Go, `bson.Unmarshal` does not crap out about invalid UTF-8 characters, and it retains those invalid characters, and later on when we serialize they get replaced by UTF8 replacement characters.

In Rust, bson Deserialize is by default strict about UTF8 and this was breaking some customers' tasks. This pull-request updates the behaviour of bson-transcoder to use an additional layer of custom Deserializer, be relaxed about invalid UTF8 characters and replace them.

Note that we do not apply this lossy UTF8 deserialization on the `_id` field when it comes to deletion events (where we directly extract the `_id`. At first I thought I should do that too, but it somehow feels very wrong to have invalid UTF8 characters in the `_id` field.

Added benchmarks to make sure there is no noticable performance loss, and there isn't.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/bson-transcoder/benches/transcoder_bench.rs
+++ b/bson-transcoder/benches/transcoder_bench.rs
@@ -1,6 +1,7 @@
 use bson::raw::RawDocument;
-use bson_transcoder::bson_to_sanitized_json;
 use bson_transcoder::id_to_string;
+use bson_transcoder::serializer::SanitizingSerializer;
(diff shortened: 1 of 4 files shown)


jwhartley:
Merging this as approved, folks are offline, customer is waiting, and failing integration tests look unrelated.

PR #3641, opened 2025-12-18
source-iterable: add project_type config setting to fix `users` primary key

Alex-Bair:
**Description:**

Iterable supports three project types that determine how users are uniquely identified. See https://support.iterable.com/hc/en-us/articles/9216719179796:

- Email-based: `email` is the unique identifier
- UserID-based: `itblUserId` is the unique identifier
- Hybrid: `itblUserId` is the unique identifier

Previously, the `users` stream hardcoded `email` as the primary key, which is incorrect for UserID-based and Hybrid projects where users do not have to have an email address. Pre-existing captures won't have a `project_type` in their config, and the connector will maintain the current behavior for these captures by assuming an absent `project_type` means to use `email` as `users`' primary key.

Note: We use `itblUserId` (Iterable's internal identifier) instead of `userId` because `itblUserId` is always present on all user records for both Hybrid and UserId-based projects.
See https://support.iterable.com/hc/en-us/articles/217744303-User-Profile-Fields-Used-by-Iterable#itbluserid.

The spec snapshot change is expected due to the addition of the new `project_type` setting. See this [jsondiff](https://jsondiff.com/?left=https://raw.githubusercontent.com/estuary/connectors/e84b9a738bdb960cbfa111c93656d9e9e3e0094d/source-iterable/tests/snapshots/snapshots__spec__capture.stdout.json&right=https://raw.githubusercontent.com/estuary/connectors/4ef1aca2e9313671d11143281e5dcd4fb7a5f6e0/source-iterable/tests/snapshots/snapshots__spec__capture.stdout.json) for a cleaner diff than what GitHub's showing.

**Workflow steps:**

Connector documentation should be updated to mention the new `project_type` config option.

**Notes for reviewers:**

Confirmed via `flowctl raw discover` that the primary key for the `users` resource changes depending on the configured `project_type` according to the bullet points in the above description.

### DIFF
--- a/source-iterable/source_iterable/models.py
+++ b/source-iterable/source_iterable/models.py
@@ -1,0 +1,6 @@
+from enum import StrEnum
+
+class ProjectType(StrEnum):
+    EMAIL_BASED = "Email-based"
+    USER_ID_BASED = "UserID-based"
+    HYBRID = "Hybrid"
--- a/source-iterable/source_iterable/schemas/users.json
+++ b/source-iterable/source_iterable/schemas/users.json
@@ -310,6 +310,9 @@
     "userId": {
       "type": ["null", "string"]
     },
+    "itblUserId": {
+      "type": ["null", "string"]
+    },
     "shopify_created_at": {
       "type": ["null", "string"],
       "format": "date-time"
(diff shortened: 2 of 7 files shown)


JustinASmith (approved):
LGTM!

PR #3648, opened 2025-12-19
materialize-snowflake: add count of rows processed to errors

danielnelson:
This will allow us to determine how far along we were in the processing of a query if an error occurs.

related: https://github.com/estuary/connectors/pull/3608


williamhbaker (approved):
LGTM

PR #3650, opened 2025-12-23
source-facebook-marketing-native: Add `should_retry` handler for HTTP 500 responses

JustinASmith:
**Description:**

This addresses HTTP 500 errors that are produced from requesting too much data in a single API request affecting `ad_creatives` primarily, but also possible on other streams that are not `ads_insights*`. Without this, the streams may not make progress since the default page limit of 100 could always be too large.

The page size is now dynamically divided in half from 100 to the minimum of 1. These errors are retired up to 5 times.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack and observed the dynamic page size reducing from `100` -> `50` -> `25` and then succeeding at paging through the data for `ad_creatives`.

### DIFF
--- a/source-facebook-marketing-native/source_facebook_marketing_native/client.py
+++ b/source-facebook-marketing-native/source_facebook_marketing_native/client.py
@@ -6,7 +6,7 @@ from enum import Enum
 from logging import Logger
 from typing import AsyncGenerator, Generic, TypeVar
 
-from estuary_cdk.http import HTTPSession, Headers
+from estuary_cdk.http import HTTPSession, Headers, ShouldRetryProtocol
 from pydantic import (
     BaseModel,
     ConfigDict,
@@ -25,6 +25,7 @@ from .enums import Field
 
 MAX_500_ERROR_RETRY_ATTEMPTS = 5
 DEFAULT_PAGE_SIZE = 100
+MIN_PAGE_SIZE = 1
 UNSUPPORTED_FIELDS = {
     Field.UNIQUE_CONVERSIONS,
     Field.UNIQUE_CTR,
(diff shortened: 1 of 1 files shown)


source-facebook-marketing-native/source_facebook_marketing_native/client.py
```diff
@@ -220,6 +221,59 @@ def __init__(
...
+                    payload.get("error", {}).get("message")
+                    or payload.get("message")
+                    or ""
+                )
+            except Exception:
+                msg = body.decode("utf-8", errors="ignore")
+
+            should_reduce = params.limit > MIN_PAGE_SIZE and (
+                "reduce" in msg.lower() or "amount of data" in msg.lower()
+            )
```
JustinASmith: The actual message I have noticed in Production is `Please reduce the amount of data you're asking for, then retry your request` , but decided to match on bits oft he string that still indicate we are fetching too much data in case the exact message changes.

The response also has an error code of `1`, but according to [Facebook](https://developers.facebook.com/docs/graph-api/guides/error-handling/) that is a very generic and possible transient error. So matching on the message made more sense in this case so we know when we should reduce the page size, simply retry, or stop retrying early.

nicolaslazo (approved):

source-facebook-marketing-native/source_facebook_marketing_native/client.py
```diff
@@ -250,6 +304,7 @@ async def _fetch_resource_data(
                 self.log,
                 url,
                 params=request_params,
+                should_retry=self._make_should_retry(params),
```
danielnelson: Params is not the object being passed into self.http.request, so is it going to work when you modify the captured type in the callback?

PR #3653, opened 2025-12-23
source-iterable: remove `events` stream

Alex-Bair:
**Description:**

No active captures have successfully completed a sync of the `events` stream, even when it's been left to run for over half a day. The strategy of making a request for each user isn't scalable - it will take too long to complete an entire sync and too much data (50+ GB) can be stored on disk before checkpointing any of it.

Since this stream hasn't ever worked for us and implementing a different, truly incremental strategy will require re-writing as a native connector, I'd like to remove `events` from the imported `source-iterable` connector to avoid the undesirable failure mode of running reactors out of disk space.

The discover snapshot change is expected since the `events` stream no longer exists and therefore no longer discovered.

**Workflow steps:**

The connector's documentation should be updated to remove the `events` stream.

**Notes for reviewers:**

See https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1766513665335579 for context about `events` causing reactors to run out of disk space.


williamhbaker (approved):
LGTM

PR #3654, opened 2025-12-23
source-facebook-marketing-native: update retry logic

JustinASmith:
**Description:**

We are seeing elusive HTTP 500 errors in Prod captures still: `"error\":{\"code\":1,\"message\":\"An unknown error occurred\",\"error_subcode\":99}}`

From what I can find, this may mean we are still requesting too much data.

- Increase retry limit from 5 to 15
- Retry all retriable errors not just 500
- Retry even if we are at 1 for our page size
- Include `"unknown error"` in condition to reduce page size

**Workflow steps:**

**Notes for reviewers:**


danielnelson (approved):
This will retry 4xx errors as well?  Typically, I would retry all 5xx unless the API is doing something non-standard.

PR #3655, opened 2025-12-29
source-zoho: new connector

nicolaslazo:
**Description:**

This PR adds `source-zoho`, a new capture connector that reads module data from the Zoho CRM. It is designed to dynamically discover and collect all modules and fields on the user's account. Like other sources with formula fields, it regularly schedules backfills to merge in updated formula values.

**Workflow steps:**

**Notes for reviewers:**

Zoho offers multiple APIs to fetch data with, each varying in rate limits, response time, and output formats. The [Bulk Read API](https://www.zoho.com/crm/developer/docs/api/v8/bulk-read/overview.html) has a round-trip time of multiple minutes and is relegated to only working on backfills, while [COQL](https://www.zoho.com/crm/developer/docs/api/v8/COQL-Overview.html) is used in incremental fetches for its sub-second response time. The main conflict between these two is that the former returns records in CSV format while the latter uses JSON. To reconcile these differences, an exhaustive pattern matching field validator has been implemented. The connector will error out in the case that an unknown data type is found to prevent malformed data from being persisted.

The testing cycle for the field normalisation function was as follows:

1. Wait for all backfills to complete
2. For every module, create one record with all fields present
3. Wait for incremental fetches to complete. If any data types were not being handled, implement a new match case
4. Once all fetches are complete, make note of all the new documents
5. Issue a backfill to collect those same documents through the bulk API
6. Use JSON diffing to ensure output consistency between both methods

Test account credentials can be provided if required.

On a different note, this connector presents an issue that has not been solved. Zoho is very strict about the number of access token refreshes that can be performed in a small window of time and will respond with HTTP 400 errors once the rate limit is exceeded. By necessity, the CDK will refresh the token twice when publishing the capture, and backfills might fail because of this. The issue can be resolved by waiting a few minutes, but 400 error messages will be displayed on logs.


source-zoho/source_zoho/models.py
```diff
@@ -0,0 +1,335 @@
...
+    "ZohoCRM.bulk.READ",  # For Bulk Read API (backfills)
+    "ZohoCRM.coql.READ",  # For COQL queries (incremental)
+    "ZohoCRM.modules.READ",  # For reading module data
+    "ZohoCRM.settings.modules.READ",  # For reading module metadata
+    "ZohoCRM.settings.fields.READ",  # For field metadata
+]
+ALWAYS_REQUIRED_DOCUMENT_FIELDS = {"id", "Modified_Time"}
+
+UNSUPPORTED_MODULES = {
+    # Modules that get listed but aren't actually supported
```
Alex-Bair: nit: could we add some more information in this comment about why these listed modules aren't supported?

source-zoho/source_zoho/models.py
```diff
@@ -0,0 +1,335 @@
...
+    accessTokenHeaders={"content-type": "application/x-www-form-urlencoded"},
+    accessTokenBody=(
+        "grant_type=authorization_code"
+        + r"&client_id={{#urlencode}}{{{ client_id }}}{{/urlencode}}"
+        + r"&client_secret={{#urlencode}}{{{ client_secret }}}{{/urlencode}}"
+        + r"&redirect_uri={{#urlencode}}{{{ redirect_uri }}}{{/urlencode}}"
+        + r"&code={{#urlencode}}{{{ code }}}{{/urlencode}}"
+    ),
+    accessTokenResponseMap={
+        "access_token": "/access_token",
```
Alex-Bair: Including `access_token` and `access_token_expires_at` here in `accessTokenResponseMap` causes them to be included in the connector config in the spec. We don't need either of those fields in the config since the connector's using the CDK's `BaseOAuth2Credentials` class, and that just requires the `refresh_token`.

source-zoho/config.yaml
```diff
@@ -0,0 +1,18 @@
+credentials:
+    client_id_sops: ENC[AES256_GCM,data:kUhhnO+LNwzezY2wEWepFzR0Twz8sN13lqmlCaT2LKwTqe0=,iv:MZ6NdOs9s9uT97lAzzmDUgk4LlglezI2o12VOZdAx5k=,tag:CpaaBoIBAFJ7YQzyin8e9A==,type:str]
+    client_secret_sops: ENC[AES256_GCM,data:XtkkRkPyBJmQa/vA1czbhj2OqnP3mjM1Lh1hZuW5g8kGlVTcMrz22XBC,iv:xbwwKqr+cA1PdhSMTFmj7S35e7zDPQrkQBu5T2qU4Po=,tag:JYUvAC2Qq/Vn81X9/6y4wA==,type:str]
+    credentials_title: OAuth Credentials
+    refresh_token_sops: ENC[AES256_GCM,data:yb4s9nBFVeMLqxGwjPStk+MGuSo5Cql4lY14F55ltbdJNKcWkNo07rrBCvk32HG7U1aA0+C5EdxOqTj9hoKdTdCKaUpyiA==,iv:+hxT1/C1PVsv5Kw7dI2+a9nur1PE5vaMjBiQLnSBXQM=,tag:4lxK8cVx7jLg97QDJ0KHQA==,type:str]
+    access_token: 1000.cd0d8041f5020f86bc2fba518163e67e.582e86c1a14e74c8bfbed6edcd9947eb
+    access_token_expires_at: 1767011627
```
Alex-Bair: Related to my comment [here](https://github.com/estuary/connectors/pull/3655/changes#r2658134047), we won't want to include either `access_token` or `access_token_expires_at` in `config.yaml`.

Also, although we won't be including `access_token` in the config for this connector, when we do, we want to make sure it's encrypted with `sops` instead of in plaintext. The CDK typically handles this when we use one of the authentication classes it provides, but the `access_token` field here is unencrypted since it's an additional field specified in the `OAUTH2_SPEC.accessTokenResponseMap` that's not present in the `BaseOAuth2Credentials` model.

source-zoho/source_zoho/bulk_api.py
```diff
@@ -0,0 +1,308 @@
...
+
+CSV_CONFIG = CSVConfig(
+    delimiter=",",
+    quotechar='"',
+    lineterminator="\n",
+    encoding="utf-8",
+)
+
+
+def sleep_time_gen() -> Generator[float]:
```
Alex-Bair: nit: since this generator is internal to `bulk_api.py`, could we prepend its name with a `_` to make it `_sleep_time_gen` so it's a little clearer it's only used in this file?

source-zoho/source_zoho/bulk_api.py
```diff
@@ -0,0 +1,308 @@
...
+            for attempt, sleep_delay in enumerate(sleep_time_gen(), start=1):
+                job_details = await self._check_status(job_id)
+
+                if job_details.state == BulkJobState.COMPLETED:
+                    if job_details.result is None:
+                        self.log.warning(
+                            "Bulk job completed but no result available",
+                            {"job_id": job_id},
+                        )
+                        return
```
Alex-Bair: nit: I think it'd be a bit clearer if this early return happened outside of the loop where we're polling the job status. That'd help separate the different handling for fetching the job status and for taking an action when the job completes. I was thinking this return could be done after exiting the loop but before attempting to fetch results:
```python
assert job_details is not None
if job_details.result is None:
    # log and return early

download_url = self.base_url + job_details.result.download_path

```

source-zoho/source_zoho/bulk_api.py
```diff
@@ -0,0 +1,308 @@
...
+
+            if not job_details.result.more_records:
+                break
+
+            page_token = job_details.result.next_page_token
+            if not page_token:
+                self.log.warning(
+                    "more_records is True but no next_page_token provided",
+                    {"job_id": job_id},
+                )
```
Alex-Bair: Have you seen the response indicate there are more records with `more_records` but not provide a `next_page_token` when you've been testing? If so, instead of allowing the function to return & let the connector potentially miss data, we should figure out what this situation means and make sure we're not suppose to use some other way to fetch the remaining records.

source-zoho/source_zoho/models.py
```diff
@@ -0,0 +1,335 @@
...
+class FieldTypeFilter(StrEnum):
+    ALL = auto()
+    FORMULA = auto()
+
+
+class ZohoModule(BaseCSVRow, extra="allow"):
+    api_name: ClassVar[str]
+    generated_type: ClassVar[
+        str | None
+    ]  # "default", "linking", "subform", "web", "custom"
```
Alex-Bair: nit: I see both `ZohoModule` and `ModuleListing` have a `generated_type` class variable with similar comments that I assume indicates the possible values. What do you think about having a `StrEnum` whose members are those values to make it more explicit what the possible values are? Then we can use that `StrEnum` in the type annotations for `generated_type` instead of `str`?

source-zoho/source_zoho/resources.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def make_resource(module: type[ZohoModule]) -> Resource:
+        return Resource(
+            name=module.api_name,
+            key=["/id"],
+            model=module,
+            open=functools.partial(open, module),
+            initial_state=ResourceState(
+                backfill=ResourceState.Backfill(
+                    cutoff=cutoff, next_page=start_date.isoformat(timespec="seconds")
```
Alex-Bair: nit: Instead of seeding `next_page` with a non-`None` initial state, let's have `None` be the initial `next_page` state, then determine within the `backfill_module` function whether or not to use the start date in the config. See this [comment](https://github.com/estuary/connectors/pull/3372#discussion_r2417066148) for more context on the reasoning for doing this. tl;dr - we may build a feature into the platform that relies on the "beginning backfill" and "resuming backfill" logs from the CDK, and providing an initial `next_page` of `None` allows the CDK to determine when a backfill is starting vs. resuming.

source-zoho/source_zoho/bulk_api.py
```diff
@@ -0,0 +1,308 @@
...
+
+    timeout = datetime.now(tz=UTC) + TARGET_BACKFILL_INVOCATION_RUN_TIME
+    timeout_checkpoint: datetime | None = None
+
+    log.info(
+        "Initiating bulk backfill",
+        {"page_ts": page_ts, "cutoff": cutoff, "timeout": timeout},
+    )
+
+    bulk_manager = BulkJobManager(base_url, http, log)
```
Alex-Bair: Do we need a `bulk_manager` for each stream? I wouldn't anticipate so since the `bulk_manager` is general enough to work for any streams. I think we could just create a single `bulk_manager` instance in the `all_resources` function, then provide it to each stream's partial `fetch_page` function.

Alex-Bair (commented):
I had a handful of comments. Quite a few were because I don't have a clear idea of when we should use `page` vs. `page_token` to paginate through Bulk API results. Any clarity you can provide around that would be appreciated.

> On a different note, this connector presents an issue that has not been solved. Zoho is very strict about the number of access token refreshes that can be performed in a small window of time and will respond with HTTP 400 errors once the rate limit is exceeded. By necessity, the CDK will refresh the token twice when publishing the capture, and backfills might fail because of this.

We can talk about this problem more when we meet next week as well, but it looks like both `validate_credentials` and `all_resources` are instantiating a new `TokenSource`. Both `TokenSource`s will try to fetch a fresh access token the first time they make an API call, meaning we'll unnecessarily fetch a duplicate access token as part of `validate` within `all_resources`. We could add a guard clause in `all_resources` to only instantiate a new `TokenSource` if one doesn't already exist on the `http` client:
```python
async def all_resources(
    log: Logger,
    http: HTTPMixin,
    config: EndpointConfig,
) -> list[Resource]:
    if http.token_source is None:
        http.token_source = TokenSource(
            oauth_spec=OAUTH2_SPEC, credentials=config.credentials
        )

```
That'll remove at least one access token refresh request during the publication process.

source-zoho/source_zoho/bulk_api.py
nicolaslazo: I agree, I haven't seen this come up in testing. I'll turn it into an `assert` for typing reasons

source-zoho/source_zoho/models.py
```diff
@@ -0,0 +1,335 @@
...
+
+            case _, _:
+                raise ValueError(
+                    f"Unexpected field type found. Data type: {data_type}, JSON type: {json_type}"
+                )
+
+    @model_validator(mode="before")
+    @classmethod
+    def normalize_coql_data(cls, data: Any, context: ValidationInfo) -> Any:
+        """Convert COQL output to the bulk API format for consistency."""
```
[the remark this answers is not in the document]
nicolaslazo: I'll include the explanation in the function's docstring 💪

source-zoho/source_zoho/models.py
```diff
@@ -0,0 +1,335 @@
...
+            case "string" | "integer", _ if isinstance(value, str | None):
+                return value
+
+            case "double" | "integer", _ if isinstance(
+                value, float | int | Decimal | None
+            ):
+                return str(value) if value is not None else None
+
+            case "boolean", "boolean":
+                return "true" if value else "false"
```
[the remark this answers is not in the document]
nicolaslazo: Yup. We could have a Bulk API field normalisation function the same way we do with COQL data and implement the conversion there

PR #3718, opened 2026-01-06
materialize-motherduck: add max object size to store delete query

danielnelson:
**Description:**

If this value is too small then motherduck may produce an error, the current default size is 16MiB.  Since our documents can be up to 64MiB we increase this value to make sure all objects can be read.

**Workflow steps:**

No changes

None

**Notes for reviewers:**


williamhbaker (approved):
LGTM

PR #3752, opened 2026-01-13
Updating error messaging

jonwihl:
**Description:**

Updates the error message to be more actionable.

**Workflow steps:**

**Notes for reviewers:**


JustinASmith (approved):
LGTM

danielnelson (commented):
I don't think we should reference UI operations in the errors, can we rephrase it to be more interface agnostic?  Perhaps something like: `changing from delta updates to standard updates is not allowed without backfilling`.

PR #3758, opened 2026-01-15
source-iterable-native: new connector

Alex-Bair:
**Description:**

This commit introduces an initial implementation of a native Iterable capture connector. See the `README.md` for notable API features and connector design decisions.

After merging this connector, the `source-iterable` connector should be marked as deprecated.

**Workflow steps:**

Documentation should be created for the new connector.

**Notes for reviewers:**

Tested on a development stack. Confirmed all streams capture documents and complete their backfills, snapshots, and/or incremental replication.


source-iterable-native/source_iterable_native/export_job_manager/__init__.py
```diff
@@ -0,0 +1,184 @@
...
+            self.log.debug(f"Reading export job results.", {
+                "job_id": job_id,
+                "file": file.file,
+            })
+
+            _, lines = await self.http.request_lines(
+                self.log,
+                file.url,
+                # file.url is a pre-signed URL, and we shouldn't include
+                # the Iterable API key when reading results with that URL.
```
JustinASmith: Does the signed URL expire? Often times signed URLs do expire for security reasons and just want to make sure this was thought about.

source-iterable-native/source_iterable_native/models.py
```diff
@@ -0,0 +1,412 @@
...
+
+        parts = [
+            cls._get_id_part(data, "createdAt"),
+            cls._get_id_part(data, "email"),
+            cls._get_id_part(data, "itblUserId"),
+            cls._get_id_part(data, "campaignId"),
+            cls._get_id_part(data, "eventName"),
+            event_type,
+        ]
+        data["_estuary_id"] = xxhash.xxh128("|".join(parts).encode()).hexdigest()
```
JustinASmith: Good choice of hashing algorithm. Based on a brief review, `xxHash` appears fast and robust with respect to collision avoidance for this context.

source-iterable-native/source_iterable_native/export_job_manager/models.py
```diff
@@ -0,0 +1,81 @@
...
+class GetExportFilesResponse(BaseModel):
+    class ExportFile(BaseModel):
+        file: str
+        url: str
+
+    # exportTruncated is True if the total export size
+    # exceeded the max allowed 100GB size & was truncated
+    # (i.e. we need to request a smaller date range to
+    # ensure we don't miss data)
+    exportTruncated: bool
```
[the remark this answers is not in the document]
Alex-Bair: I'm 99% certain it's the latter, although I'm running a test right now to be completely certain. The test has processed ~10GB in an hour, which is a pretty long time for an export job to run, and it hasn't truncated the job yet. If it continues at the current pace, it could take about 10 hours for an export job to set `exportTruncated` to `True`.

The connector tries to keep the result set size to a more manageable amount by only requesting at most 5 days of data, which helps reduce both how long the export job takes to complete as well as how likely it is for the job to be truncated. It feels fairly unlikely that jobs requesting 5 days of data will get truncated or take very long to complete. If we do see that happen though, we can decrease the `INITIAL_EXPORT_DATE_WINDOW_SIZE` or expose it as a config setting; I wanted to avoid the config setting option for now though until we could see how well the 5 day `INITIAL_EXPORT_DATE_WINDOW_SIZE` works for users.

PR #3759, opened 2026-01-15
build-local: update for Python CDK-based connectors

Alex-Bair:
**Description:**

We've been passing around a `build-local-python.sh` script in back alleys for building Python CDK-based connectors during development. That's not ideal; ~~it'd be better if the `build-local.sh` script worked for all connectors in the `connectors` repo, so I've updated the script to work for Go, Rust, and Python CDK-based connectors.~~

Edit: I was wrong, `build-local.sh` already worked for Python CDK-based connectors by specifying the `./estuary-cds/common.Dockerfile` as the second argument. It'd still be nice to not have to specify it when using `build-local.sh`, which will be the main improvement of this PR.

**Workflow steps:**

The way to use the `build-local.sh` script is the same as it was before: `./build-local.sh <connector-name> [Dockerfile]`

**Notes for reviewers:**

Confirmed the script works to build:
- `materialize-iceberg` (Go) and `source-kafka` (Rust) that have Dockerfiles located in the connector's directory
- `source-hubspot-native` (Python CDK-based) that uses the `estuary-cdk/common.Dockerfile`

### DIFF
--- a/build-local.sh
+++ b/build-local.sh
@@ -1,33 +1,51 @@
 #!/bin/bash
-if [[ $# < 1 ]]; then
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+name="$1"
+
+if [[ -z "$name" ]]; then
     echo "This script builds \"<connector_name>:local\" images for testing."
-    echo "Usage: build-local.sh <connector_name> <Dockerfile>"
+    echo ""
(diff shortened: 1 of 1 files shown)


JustinASmith (approved):
LGTM! FWIW, this script worked for any connectors I'm pretty sure as long as you specify in the second argument the proper Dockerfile.

These changes are worth it though so we get exits on any non-zero exit codes, clear variable names, and automatic Dockerfile selection.

PR #3763, opened 2026-01-16
source-hubspot-native: add campaigns stream

### LINKED ISSUE #3762: source-hubspot-native: add `campaigns` stream
Docs URL: https://developers.hubspot.com/docs/api-reference/marketing-campaigns-public-api-v3/campaign/get-marketing-v3-campaigns-

nicolaslazo:
**Description:**

This PR implements a new `campaigns` stream for `source-hubspot-native`. Closes #3762 .

**Workflow steps:**

**Notes for reviewers:**

Campaigns have been excluded from snapshot tests as they require a HubSpot Sales subscription. The stream has been tested locally with a time limited free trial.


source-hubspot-native/source_hubspot_native/api.py
```diff
@@ -75,6 +76,27 @@
...
+    "hs_owner",
+    "hs_currency_code",
+    "hs_created_by_user_id",
+    "hs_campaign_status",
+    "hs_object_id",
+    "hs_name",
+    "hs_utm",
+    "hs_budget_items_sum_amount",
+    "hs_spend_items_sum_amount",
+}
```
Alex-Bair: nit: I assume there's not a way to determine these fields from the metadata returned in the `/crm/v3/properties/campaign` response. Is that right? It may not be possible, but I'd like to avoid having to update this set ourselves if/when HubSpot decides to add a new readable property to campaigns.

PR #3810, opened 2026-01-21
source-google-ads: make primary key fields non-nullable in custom query schemas

Alex-Bair:
**Description:**

Custom query streams dynamically generate their JSON schemas by querying the Google Ads API for field metadata. Previously, all fields were marked as nullable (e.g., `["integer", "null"]`), including fields specified as primary keys by the user.

This change checks if each field is part of the user-specified primary key and, if so, generates a non-nullable type (e.g., `"integer"` instead of `["integer", "null"]`).

Discover snapshot changes are expected - collection key fields for the example custom stream are no longer discovered as nullable.

**Workflow steps:**

**Notes for reviewers:**

Confirmed via `flowctl raw discover` that fields that are part of the collection key schema for custom streams are no longer nullable. You can see this in `acmeCo/my_custom_query.schema.yaml` with the `customer.id` and `segments.date` fields.

### DIFF
--- a/source-google-ads/acmeCo/my_custom_query.schema.yaml
+++ b/source-google-ads/acmeCo/my_custom_query.schema.yaml
@@ -46,9 +46,7 @@ properties:
     - boolean
     - 'null'
   customer.id:
-    type:
-    - integer
-    - 'null'
+    type: integer
   metrics.conversions:
     type:
     - number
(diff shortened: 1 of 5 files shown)


JustinASmith (approved):
LGTM

PR #3626, opened 2025-12-15
Shopify: Support Multiple Stores

JustinASmith:
**Description:**

#3508 

**Workflow steps:**

(How does one use this feature, and how has it changed)

**Documentation links affected:**

(list any [documentation links](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing) that you created, or existing ones that you've identified as needing updates, along with a brief description)

**Notes for reviewers:**

(anything that might help someone review this PR)

PR #3816, opened 2026-01-22
CDK: Support configuring config-encryption `URL` at build time for local dev

JustinASmith:
**Description:**

This adds an `ARG` and `ENV` variable to the CDK common.Dockerfile and consumes that as an optional value in the `BaseCaptureConnector` once at the module level. This provides the option to set this variable when testing on a local stack, which will not have access to the production KMS keys for encryption/decryption.

**Workflow steps:**

**Notes for reviewers:**

Tested along with #3626 in a GCP VM (`mise` local stack)

### DIFF
--- a/estuary-cdk/common.Dockerfile
+++ b/estuary-cdk/common.Dockerfile
@@ -26,6 +26,7 @@ FROM base AS runner
 ARG CONNECTOR_NAME
 ARG CONNECTOR_TYPE
 ARG DOCS_URL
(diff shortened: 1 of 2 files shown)


Alex-Bair (approved):
LGTM % is there a way to centralize the URL to a single spot rather than having it defined in both `common.Dockerfile` and the CDK?

Also, it'd be nice if the connector automatically knew if it was running in a development stack environment instead of production & automatically determined the appropriate `ENCRYPTION_URL` to use. But I'm not aware of anything currently in place that would let connectors figure that out, so the current strategy of requiring to manually override it in development sounds good.

PR #3837, opened 2026-01-28
source-iterable-native: convert `list_users` from a snapshot to a backfill only stream

Alex-Bair:
**Description:**

The `list_users` stream was causing OOMs and blocking graceful shutdowns because snapshotting all lists' users could take hours due to API limitations (5 req/min rate limit, responses taking 80+ minutes for large lists, no pagination).

This PR converts `list_users` to use periodic backfill, fetching a single list's users per `fetch_page` invocation. This makes the connector more responsive during graceful shutdowns, and the connector can make incremental progress while backfilling `list_users`. In order to capture new and updated `list_users`, a backfill schedule is used; by default, a backfill will be kicked off at most once a day.

The discover snapshot change is expected - the primary key for `list_users` is changing. All production captures will have to perform data flow resets for `list_users` due to the key change. This only applies to one capture, so I'll do that myself after merging this change.

The spec snapshot change is also expected due to the addition of the `schedule` field to the resource config.

**Workflow steps:**

The connector's documentation should be updated to reflect the new `schedule` resource config field and explain how it's used to capture new and updated `list_users`.

**Notes for reviewers:**

Tested on a local stack. I confirmed a single list's users is captured per `fetch_page` invocation, and the connector is able to make progress on a sweep of all lists' users across capture restarts.

### DIFF
--- a/source-iterable-native/README.md
+++ b/source-iterable-native/README.md
@@ -26,6 +26,11 @@ Some endpoints have different rate limits that aren't (currently) managed by the
 
 The connector manually sleeps between requests to these endpoints to avoid hitting rate limits.
 
+
+### `/lists/getUsers` challenges
+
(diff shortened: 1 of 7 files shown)


nicolaslazo (approved):
LGTM 👍

PR #3847, opened 2026-01-30
source-stripe-native: fix exempt streams with scheduled backfills

JustinASmith:
**Description:**

## Summary

This PR fixes two critical bugs affecting incremental capture and adds scheduled backfill support for streams with unreliable Stripe event generation.

## Bug Fixes

**Stuck cursor when `capture_connected_accounts=false`**: The incremental cursor never advanced during idle periods because the cursor advancement logic incorrectly required `connected_account_id` to be non-None. This caused the connector to repeatedly fetch the same time window indefinitely. Fixed by changing `elif connected_account_id and end > log_cursor` to `else` (or `elif end > log_cursor` for `fetch_incremental_no_events`).

**Duplicate backfills for Accounts when `capture_connected_accounts=true`**: The Accounts stream incorrectly spawned subtasks for each connected account. Since `/v1/accounts` always queries from the platform account to list all connected accounts, this caused N redundant backfills (e.g., 7,500 duplicates for 7,500 connected accounts). Fixed by adding Accounts to `CONNECTED_ACCOUNT_EXEMPT_STREAMS` to skip per-account subtasks and use non-dict state.

## Scheduled Backfills

Added `SCHEDULED_BACKFILL_STREAMS` with daily schedule (`0 0 * * *`) for eventual consistency, since Stripe's event generation is unreliable for these streams:

| Stream | Subtask Exempt | Incremental Mechanism | Backfill Schedule |
|--------|---------------|----------------------|-------------------|
| Accounts | Yes | List endpoint (`fetch_incremental_no_events`) | Daily |
| Persons | No | Events API (created/updated/deleted) | Daily |
| ExternalAccountCards | No | Events API (created/updated/deleted) | Daily |
| ExternalBankAccount | No | Events API (created/updated/deleted) | Daily |

**Accounts** is the only subtask-exempt stream. It uses the list endpoint because Stripe's event generation for accounts is unreliable—testing confirmed that webhooks receive `account.updated` events, but the Events API endpoint may not surface them.

**Persons, ExternalAccountCards, ExternalBankAccount** are child streams accessed via parent endpoints (e.g., `/accounts/{id}/persons`). They create per-account subtasks to query each connected account's events with the `Stripe-Account` header. They retain full EVENT_TYPES for best-effort event capture, with daily scheduled backfills as a safety net for missed events.

## State Migration

Accounts stream includes automatic state migration from dict format (per-account subtasks) to non-dict format for existing captures. The migration:
- Preserves the platform account's incremental cursor
- Triggers a re-backfill to ensure data consistency
- Uses `merge_patch=False` to fully replace dict state with non-dict state

**Workflow steps:**

https://github.com/estuary/flow/pull/2649

**Notes for reviewers:**

Performed a lot of different tests to build confidence in this.

**Unit tests** (`tests/test_state_migration_unit.py`):
- Migration uses platform account cursor when available
- Fallback to earliest cursor when platform cursor missing
- Handles None cursors in dict state
- Non-dict state passes through unchanged
- Empty dict state handled gracefully

**Manual verification** (live capture with 4 connected accounts):
- Confirmed state migrated from dict to non-dict format
- Verified cursor advanced after migration (13:08 → 13:12)
- Confirmed re-backfill triggered and completed
- Verified duplicate Accounts workers eliminated (4 → 1)
- Confirmed non-exempt streams (BankAccounts for e.g.) still use per-account subtasks
- Compared RocksDB state dumps before/after migration


source-stripe-native/source_stripe_native/resources.py
```diff
@@ -773,13 +936,19 @@ def fetch_page_factory(account_id: str):
...
-        initial_config=ResourceConfig(name=cls.NAME, interval=timedelta(minutes=5)),
+        initial_state=effective_initial_state,
+        initial_config=ResourceConfigWithSchedule(
+            name=cls.NAME,
+            interval=timedelta(minutes=5),
+            schedule=DEFAULT_SCHEDULE,
+        ) if needs_schedule else ResourceConfig(
+            name=cls.NAME,
+            interval=timedelta(minutes=5),
+        ),
```
JustinASmith: same as above

source-stripe-native/source_stripe_native/resources.py
```diff
@@ -312,6 +318,77 @@ def _create_initial_state(account_ids: str | list[str]) -> ResourceState:
...
+        task.log.info(
+            f"Migrating {stream_name} from dict state to non-dict state. "
+            "This will trigger a re-backfill."
+        )
+        state = _migrate_exempt_stream_state(state, platform_account_id)
+        # Uses merge_patch=False to fully replace dict state with non-dict state
+        await task.checkpoint(
+            ConnectorState(bindingStateV1={binding.stateKey: state}),
+            merge_patch=False,
+        )
```
Alex-Bair: What happens to the other bindings' previously checkpointed state after making this `merge_patch=False` checkpoint? I anticipate they get wiped; that's what I observed when I tried the same strategy for `source-jira-native` in this [PR](https://github.com/estuary/connectors/pull/3302#discussion_r2370259518).

Instead, a safer alternative that does not wipe out other bindings' checkpointed state would be using a `merge_patch=True` checkpoint that both adds the migrated non-subtask state and deletes state for the subtask keys in the same checkpoint. I used this strategy recently in `source-outreach` [here](https://github.com/estuary/connectors/blob/53d8a824743885887586f636e056f3458e99d5de/source-outreach/source_outreach/resources.py#L45-L64), and it worked well. Could the same strategy be used here?

Alex-Bair (commented):
Had a few comments/questions.

Also regarding the capture snapshot failure, is the issue that it's not getting replaced after running `pytest --insta=update`? The `source-stripe-native` capture snapshot has been relatively stable in the past year-ish and has only needed updated for the addition of new fields. It looks like the failures right now are due to missing documents, which makes me suspect that something in the connector has changed causing the test to miss those documents. We need to understand why that's happening and determine if it's acceptable before approving and merging.

source-stripe-native/source_stripe_native/resources.py
JustinASmith: I am looking into this more. I tested this thoroughly and even ran `flowctl-go shards recover` before and after migration and have logs dump showing we migrated with `merge_patch=False` for the `Accounts` binding and the other binding's states were preserved nonetheless... So I am confused. I reviewed the CDK too and you are right that it should wipe out the state...

FWIW, this is my logs and observations from RocksDB state dumps.
`[MIGRATION] Checkpoint payload (merge_patch=False): bindingStateV1={'Accounts': ResourceState(...)}`

Then the runtime logs show: `persisting updated connector state: {"bindingStateV1":{"Accounts":{...}}}`

The checkpoint sent to the runtime only contained the Accounts binding. Yet the final state dump shows all 40 other bindings preserved with their cursors advancing.

Alex-Bair: What were the conditions when you were testing? If all 40 other bindings emitted a checkpoint after the `merge_patch=False` one but before the connector restarted, that could explain why the remaining bindings' resource states were still present in the dump. Or if that didn't happen, it's possible the connector restarted, saw there was no state for the 40 other bindings, then started processing them with the `initial_state` & checkpointed; if this happened, I'd anticipate you'd also see the "beginning/resuming backfill" logs for those other bindings too.

When testing, it would help to add a log in the connector's `open` [method](https://github.com/estuary/connectors/blob/53d8a824743885887586f636e056f3458e99d5de/source-stripe-native/source_stripe_native/__init__.py#L82-L89) that logs out what `open.state` is. `open.state` is the `ConnectorState` the connector receives from the runtime, and that will show what the connector's actually receiving when it starts up before making any additional checkpoints.

source-stripe-native/source_stripe_native/models.py
```diff
@@ -816,11 +817,20 @@ class TransferReversals(BaseStripeChildObject):
...
+
+# Streams that should have scheduled backfills to ensure data completeness.
+# Stripe's Events API may not reliably generate events for these streams.
+# We don't fully understand why events appear for some users/scenarios and not others,
+# so scheduled backfills ensure we don't miss any data.
+SCHEDULED_BACKFILL_STREAMS = [
+    "Accounts",
     "ExternalAccountCards",
     "ExternalBankAccount",
     "Persons",
```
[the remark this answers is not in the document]
Alex-Bair: The connector itself doesn't need to handle re-initializing the state. This connector still uses the CDK's `open` function:
https://github.com/estuary/connectors/blob/4285491aadd5a332ec0bdbd69634af6bb24ae134/source-stripe-native/source_stripe_native/__init__.py#L89

and that already contains all the logic around checking if a resource's state should be re-initialized because of its schedule:
https://github.com/estuary/connectors/blob/e6aeb5c30e1d1a9baaf906b1e40e181096ce5b21/estuary-cdk/estuary_cdk/capture/common.py#L642-L674

That handling happens when the connector starts up, so the CDK handles the "when I start up, should I re-initialize the state?" piece. And the CDK's `open` function also sets a scheduled stop if any bindings do not have an ongoing backfill but should backfill soon per their schedule:
https://github.com/estuary/connectors/blob/e6aeb5c30e1d1a9baaf906b1e40e181096ce5b21/estuary-cdk/estuary_cdk/capture/common.py#L688-L691

But what happens if a binding's backfill took long enough that it missed a scheduled backfill? The CDK's `open` function wouldn't have set a scheduled stop since there was an ongoing backfill (we don't want to unnecessarily stop the connector!), so the CDK's `open_binding` / `_binding_backfill_task` functions handles stopping / scheduling a stop when an ongoing backfill completes:
https://github.com/estuary/connectors/blob/e6aeb5c30e1d1a9baaf906b1e40e181096ce5b21/estuary-cdk/estuary_cdk/capture/common.py#L1036-L1046

`source-stripe-native` doesn't use the CDK's `open_binding`/`_binding_backfill_task` functions, so it doesn't have that handling for stopping / scheduling a stop when an ongoing backfill finishes. That handling is what needs added to the priority capture logic in the connector.

PR #3875, opened 2026-02-05
source-hubspot-native: emit sourced schemas for CRM objects

Alex-Bair:
**Description:**

This PR's scope includes:
- updating the capture snapshot to include new fields HubSpot added recently
- updating the `properties` stream to use `BaseDocument` for its tombstone. This means we don't need default values on the `Property` model to create a `Property` instance for tombstones.
- emitting `SourcedSchema` messages for CRM object streams, like `deals` and `contacts`. The act of emitting a `SourcedSchema` bumps up the inferred schema's complexity limit from 1,000 to 10,000. This increases how many fields the inferred schema includes per the complexity limit, which is desired by users whose CRM objects have >1k properties.

I recommend reviewing commit by commit.

**Workflow steps:**

**Notes for reviewers:**

Tested on a local stack. Confirmed the `SourcedSchema` messages are constructed correctly, are accepted by the runtime, widen the inferred schema as I'd expect, and bump up the inferred schema's complexity limit.

### DIFF
--- a/source-hubspot-native/source_hubspot_native/models.py
+++ b/source-hubspot-native/source_hubspot_native/models.py
@@ -188,6 +188,17 @@ class Names(StrEnum):
     campaigns = auto()
 
 
+class PropertyTypes(StrEnum):
+    bool = "bool"
+    enumeration = "enumeration"
+    date = "date"
+    datetime = "datetime"
+    string = "string"
+    number = "number"
+    object_coordinates = "object_coordinates"
+    json = "json"
+
+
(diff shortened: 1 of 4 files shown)


Alex-Bair:
Moved back to draft. The main goal is to bump the complexity limit, and we don't need to include specific HubSpot properties to bump the limit. I'm going to rework this to not include any properties at all so schema inference can handle figuring out property types instead of relying on HubSpot's property metadata to figure it out.

PR #3879, opened 2026-02-06
Reapply "source-stripe-native: fix cursor advancement and Accounts subtask

JustinASmith:
**Description:**

This reverts commit 4bf727cfdcb36db29e5c41e7547aa047c50fe8bc.

Runtime related issues caused some issues when rolling this out originally. Thankfully we were able to revert pretty quick and no captures were truly impacted. This reapplies the reverted work and can be merged once we are confident things have settled.

## Summary

This PR fixes two critical bugs affecting incremental capture and adds scheduled backfill support for streams with unreliable Stripe event generation.

## Bug Fixes

**Stuck cursor when `capture_connected_accounts=false`**: The incremental cursor never advanced during idle periods because the cursor advancement logic incorrectly required `connected_account_id` to be non-None. This caused the connector to repeatedly fetch the same time window indefinitely. Fixed by changing `elif connected_account_id and end > log_cursor` to `else` (or `elif end > log_cursor` for `fetch_incremental_no_events`).

**Duplicate backfills for Accounts when `capture_connected_accounts=true`**: The Accounts stream incorrectly spawned subtasks for each connected account. Since `/v1/accounts` always queries from the platform account to list all connected accounts, this caused N redundant backfills (e.g., 7,500 duplicates for 7,500 connected accounts). Fixed by adding Accounts to `CONNECTED_ACCOUNT_EXEMPT_STREAMS` to skip per-account subtasks and use non-dict state.

## Scheduled Backfills

Added `SCHEDULED_BACKFILL_STREAMS` with daily schedule (`0 0 * * *`) for eventual consistency, since Stripe's event generation is unreliable for these streams:

| Stream | Subtask Exempt | Incremental Mechanism | Backfill Schedule |
|--------|---------------|----------------------|-------------------|
| Accounts | Yes | List endpoint (`fetch_incremental_no_events`) | Daily |
| Persons | No | Events API (created/updated/deleted) | Daily |
| ExternalAccountCards | No | Events API (created/updated/deleted) | Daily |
| ExternalBankAccount | No | Events API (created/updated/deleted) | Daily |

**Accounts** is the only subtask-exempt stream. It uses the list endpoint because Stripe's event generation for accounts is unreliable—testing confirmed that webhooks receive `account.updated` events, but the Events API endpoint may not surface them.

**Persons, ExternalAccountCards, ExternalBankAccount** are child streams accessed via parent endpoints (e.g., `/accounts/{id}/persons`). They create per-account subtasks to query each connected account's events with the `Stripe-Account` header. They retain full EVENT_TYPES for best-effort event capture, with daily scheduled backfills as a safety net for missed events.

## State Migration

Accounts stream includes automatic state migration from dict format (per-account subtasks) to non-dict format for existing captures. The migration:
- Preserves the platform account's incremental cursor
- Triggers a re-backfill to ensure data consistency
- Uses `merge_patch=False` to fully replace dict state with non-dict state

**Workflow steps:**

**Notes for reviewers:**

Performed a lot of different tests to build confidence in this.

**Unit tests** (`tests/test_state_migration_unit.py`):
- Migration uses platform account cursor when available
- Fallback to earliest cursor when platform cursor missing
- Handles None cursors in dict state
- Non-dict state passes through unchanged
- Empty dict state handled gracefully

**Manual verification** (live capture with 4 connected accounts):
- Confirmed state migrated from dict to non-dict format
- Verified cursor advanced after migration (13:08 → 13:12)
- Confirmed re-backfill triggered and completed
- Verified duplicate Accounts workers eliminated (4 → 1)
- Confirmed non-exempt streams (BankAccounts for e.g.) still use per-account subtasks
- Compared RocksDB state dumps before/after migration


Alex-Bair (approved):
LGTM

PR #3896, opened 2026-02-11
materialize-bigquery: add skip_cleanup feature flag

danielnelson:
**Description:**

When this flag is enabled, skip removal of staged files for later debugging.  Also added debug logging of the queries with their external data mappings, while the stage files are already logged it was difficult to determine how they were used in the query.

**Workflow steps:**

Add `skip_cleanup` to the feature flags to disable cleanup.  Use `logLevel: debug` for query logging.

None

**Notes for reviewers:**

### DIFF
--- a/materialize-bigquery/transactor.go
+++ b/materialize-bigquery/transactor.go
@@ -48,6 +48,7 @@ type transactor struct {
 
 	objAndArrayAsJson       bool
 	loggedStorageApiMessage bool
+	skipCleanup             bool
 }
 
 func prepareNewTransactor(
@@ -85,6 +86,7 @@ func prepareNewTransactor(
 			dialect:           ep.Dialect,
 			templates:         templates,
 			objAndArrayAsJson: featureFlags["objects_and_arrays_as_json"],
+			skipCleanup:       featureFlags["skip_cleanup"],
 			client:            client,
 			be:                be,
 			loadFiles:         boilerplate.NewStagedFiles(stagedFileClient{}, bucket, writer.DefaultJsonFileSizeLimit, cfg.effectiveBucketPath(), false, false),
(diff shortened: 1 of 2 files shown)


materialize-bigquery/transactor.go
```diff
@@ -392,6 +402,14 @@ func (t *transactor) Acknowledge(ctx context.Context) (*pf.ConnectorState, error
...
 			if err := t.client.queryIdempotent(groupCtx, b.storeSchema, item.Query, item.JobPrefix, item.SourceURIs, item.TempTableName); err != nil {
 				return fmt.Errorf("acknowledge query for %q: %w", b.target.Path, err)
+			}
+			log.WithFields(log.Fields{
+				"query":         item.Query,
+				"external_data": map[string][]string{item.TempTableName: item.SourceURIs},
+			}).Debug("acknowlege query executed")
+
+			if t.skipCleanup {
+				log.Warn("cleanup disabled; staged files retained")
```
williamhbaker: As a general pattern, disabling file cleanup may introduce a subtle correctness bug. Merge queries are idempotent in that they can be executed repeatedly and produce the same outcome, but if the table state changes in some other way between their executions they could actually cause some rows to revert back.

It's not like this would happen, but also not totally impossible that some invocation of the merge query would get "stuck" along the way somehow, and then more transactions are run that change the table, and then somehow that long-forgotten merge query gets unstuck and runs, reading from the still present staged files and mutating the table back to a prior state.

For BigQuery we don't have to worry about it because its idempotence mechanism relies on BigQuery's own query IDs being globally unique, but something to keep in mind for other systems that don't use that mechanism.

danielnelson: Thanks for the warning, I'll have to keep an eye out for this.  It sounds like this is the situation fencing is meant to prevent for materializations that implement it?

PR #3897, opened 2026-02-12
materialize-snowflake: log on encryption key change instead of error

danielnelson:
**Description:**

It's may be possible that an encryption key change isn't always going to result in an error, it's possible that some number of previous keys can still be read, so reducing this error to a warning.

**Workflow steps:**

**Notes for reviewers:**


danielnelson:
This won't ever work because we can't decrypt the blob if the key has changed.

PR #3921, opened 2026-02-21
estuary-cdk: enforce deadline for graceful shutdown

Alex-Bair:
**Description:**

When the `stopping.event` is set for any reason (ex: a stream encountered an exception, it's been 24 hours since the capture restarted), it's possible for other streams to block the graceful shutdown indefinitely. Captures have stalled in these situations and required manual intervention to get unstuck.

This commit adds `enforce_shutdown` that triggers after `stopping.event` is set. It waits 30 minutes for the graceful shutdown to complete. If those 30 minutes elapse and the connector hasn't exited, the `TaskGroup`'s running tasks are signaled to cancel. If the connector is _still_ running 5 minutes after attempting to cancel the tasks, then the CDK forces an exit with `os._exit(1)` as a last resort.

**Workflow steps:**

**Notes for reviewers:**

Tested on a development stack. Confirmed that:
- After 30 minutes, of waiting for a graceful shutdown, capture tasks are cancelled.
- 5 minutes after cancelling capture tasks, the connector is forcibly exited.

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
+++ b/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
@@ -33,6 +33,21 @@ from .common import _ConnectorState
 ENCRYPTION_URL = os.getenv("ENCRYPTION_URL") or "https://config-encryption.estuary.dev/v1/encrypt-config"
 
 
+PERIODIC_RESTART_INTERVAL = 24 * 60 * 60  # 24 hours
+GRACEFUL_SHUTDOWN_TIMEOUT = 30 * 60  # 30 minutes
+TASK_CANCELLATION_TIMEOUT = 5 * 60  # 5 minutes
+
+
+class TerminateTaskGroup(Exception):
+    """Exception raised to terminate a task group."""
+    pass
+
+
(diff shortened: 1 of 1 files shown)


estuary-cdk/estuary_cdk/capture/base_capture_connector.py
```diff
@@ -141,22 +197,28 @@ async def periodic_stop() -> None:
...
+                        ConnectorStatus(log, stopping),
+                        "capture",
+                        self.output,
+                        stopping,
+                        tg,
+                    )
+                    log.event.status("Capture started")
+                    await capture(task)
+            except* TerminateTaskGroup:
+                pass  # Expected when enforce_shutdown terminates the task group
```
Alex-Bair: Note: Terminating a task group isn't natively supported by the standard library, so I used a `TerminateTaskGroup` exception to trigger the task group to terminate itself. Usage of a `TermianteTaskGroup` exception to trigger task group termination was copied from the [Python docs](https://docs.python.org/3/library/asyncio-task.html#terminating-a-task-group).

PR #3924, opened 2026-02-23
source-firestore: use configured `database` when creating Firestore client

Alex-Bair:
**Description:**

The connector was always capturing from the `(default)` database regardless of the database config setting. This happened because we used the Firebase SDK's `app.Firestore()` method, which doesn't support specifying a database name.

This PR switches the connector to use `firestore.NewClientWithDatabase()` which respects the configured database path.

**Workflow steps:**

**Notes for reviewers:**

Added test coverage for the new helper functions. Confirmed with `flowctl raw discover` and `flowctl preview` that when configured with a non `(default)` database, the discovered resources and captured documents are those of the configured database instead of the `(default)` database.


willdonnelly (approved):
LGTM

PR #3930, opened 2026-02-25
source-mongodb: use FullDocument Required when changeStreamPreAndPost

mdibaiee:
**Description:**

`FullDocument: UpdateLookup` always does a lookup for every event in change stream, but if `changeStreamPreAndPostImages` is enabled on all enabled collections, then it is not necesary and the post image will be available without an extra lookup. This PR adds a check so if we find that the option is enabled on all collections, then we ues `FullDocument: Required`, trusting that it will be available.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-mongodb/change_stream_test.go
+++ b/source-mongodb/change_stream_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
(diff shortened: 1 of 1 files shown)


williamhbaker (approved):
LGTM % my comments about testing

PR #3932, opened 2026-02-25
source-oracle-batch: Fix custom cursor handling, and other stuff

willdonnelly:
**Description:**

This PR fixes the handling of custom user-specified polling cursors in `source-oracle-batch`, which was previously entirely broken.

Then I went ahead and fixed the test suite setup as well because it previously relied on a remote DB which I don't think even have access to, regenerated all test snapshots because there was some significant test snapshot drift (because, y'know, it's a lot harder to avoid this when people can't run the test suite), fixed a bug in `FLOAT` column handling which the test suite immediately surfaced now that it's runnable, and finally verified that the cursor fix didn't break anything.

I recommend reviewing this one commit-by-commit so it makes more sense.


danielnelson (approved):
LGTM

PR #3940, opened 2026-02-26
source-postgres: Fix LSN race on slot recreation

willdonnelly:
**Description:**

When (re)initializing a capture's start LSN for the first time we use the server's flush LSN as the initial value. The problem was, we were querying that LSN _before_ creating the replication slot, so if we got unlucky on a busy database we could end up with the slot having a confirmed_flush_lsn slightly ahead of the actual start LSN.

This was benign, since our `confirmedLSN <= startLSN` check is still in warn-only mode, but it'd be better not to produce spurious warnings if we get unlucky during a perfectly normal operation, so this commit just moves the flush LSN query after the slot creation.

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -170,13 +170,6 @@ func (db *postgresDatabase) ReplicationStream(ctx context.Context, startCursorJS
 
 	var slot, publication = db.config.Advanced.SlotName, db.config.Advanced.PublicationName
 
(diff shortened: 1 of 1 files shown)


willdonnelly:
Yeah, I'm not sure what's going on with source-postgres tests right now. I'm _pretty_ sure it's unrelated to this fix, but this fix isn't urgent so I'm going to set it aside until I can get CI green again.

PR #3949, opened 2026-03-03
materialize-s3-iceberg: add support for IAM auth

nicolaslazo:
**Description:**

Adds support for IAM auth by implementing a new `credentials` nested object in the connector's endpoint config. The old access token and secret fields are preserved, only hidden, for backwards compatibility with existing captures.

**Workflow steps:**

**Notes for reviewers:**

The UI has been tested through a local deployment, trying to leave required fields empty and ensuring the resulting error messages are descriptive.

The Pydantic dependency has been updated to gain access to the `primitive_type_array` union format, which helps simplify the JSON schema post-processing. A detailed explanation can be found in the documentation for `UICompatJsonSchema`.

I am aware of `S3StoreConfig`, which does most of the credentials configurations we need, but since I didn't want to break existing captures I kept my changes as minimal as possible.


materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
```diff
@@ -87,26 +89,81 @@ def run(
...
                     "client.region": cfg.region,
-                    "client.access-key-id": cfg.aws_access_key_id,
-                    "client.secret-access-key": cfg.aws_secret_access_key,
                     PY_IO_IMPL: "pyiceberg.io.fsspec.FsspecFileIO",  # use S3 file IO instead of Arrow
                 }
+
                 if cfg.catalog.glue_id:
                     glue_props["glue.id"] = cfg.catalog.glue_id
 
+                creds: dict[str, str] = {}
```
danielnelson: This creates a new dict, but it's never used because all of the branches create a new dict.  If this is just for the typing I think you could just not assign to it:
```
creds: dict[str, str]
```

materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
```diff
@@ -87,26 +89,81 @@ def run(
...
+class UICompatJsonSchema(GenerateJsonSchema):
+    """Customizes nullable type generation for UI compatibility.
+
+    The UI expects `oneOf` for discriminated unions (not `anyOf` with a null variant)
+    and `union_format='primitive_type_array'` handles nullable scalars. This subclass
+    fixes the discriminated union case by stripping the null wrapper.
+    """
+
+    @override
+    def nullable_schema(self, schema: dict[str, Any]) -> JsonSchemaValue:
```
danielnelson: I see this line comes up in mypy:
```
iceberg-ctl/iceberg_ctl/__main__.py:146: error: Argument 1 of "nullable_schema" is incompatible with supertype "GenerateJsonSchema"; supertype defines the argument type as "NullableSchema"  [override]
```

Should we use the `pydantic_core.core_schema.NullableSchema` type here?

materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/models.py
```diff
@@ -1,32 +1,99 @@
...
+        title="AWS Role ARN",
+        description="AWS Role which has access to the resource which will be assumed by Flow.",
+        json_schema_extra={"order": 2},
+    )
+    # Short-lived STS tokens injected at runtime by Flow's control plane.
+    aws_access_key_id: str = Field(
+        default="",
+        title="AWS Access Key ID",
+        description="AWS Access Key ID for accessing AWS services.",
+        json_schema_extra={"order": 3, "x-hidden-field": True},
```
danielnelson: Is it possible to exclude these injected fields from the Spec completely?  This might further discourage anyone from setting these manually.

danielnelson (approved):
LGTM % optional comments

danielnelson (approved):

PR #3957, opened 2026-03-04
source-posthog: turn persons into an incremental stream

nicolaslazo:
**Description:**

PostHog's persons stream is a snapshot type, which is currently filling up storage space in some reactors as it tries to accumulate thousands of documents in memory without checkpointing its progress.

This PR implements secondary cursor support for HogQL queries. This allows us to rely on `last_seen_at` for update timestamps, falling back to `created_at` when not available.

**Workflow steps:**

**Notes for reviewers:**

Tested locally through `flowctl preview`.

### DIFF
--- a/source-posthog/acmeCo/Persons.schema.yaml
+++ b/source-posthog/acmeCo/Persons.schema.yaml
@@ -15,6 +15,11 @@ $defs:
         description: Row ID of the Document, counting up from zero, or -1 if not known
         title: Row Id
         type: integer
+      project_id:
+        default: -1
+        description: The PostHog project this document belongs to
+        title: Project Id
+        type: integer
     title: Meta
     type: object
 additionalProperties: true
@@ -30,9 +35,16 @@ properties:
     format: date-time
     title: Created At
     type: string
+  last_seen_at:
+    anyOf:
+    - format: date-time
+      type: string
+    - type: 'null'
+    title: Last Seen At
 required:
 - id
 - created_at
+- last_seen_at
 title: Person
 type: object
 x-infer-schema: true
--- a/source-posthog/source_posthog/api.py
+++ b/source-posthog/source_posthog/api.py
@@ -2,10 +2,11 @@
 PostHog API client functions.
 """
 
-import functools
 from collections.abc import AsyncGenerator
 from datetime import UTC, datetime, timedelta
+from functools import wraps
 from logging import Logger
+from typing import Callable, ParamSpec
 from urllib.parse import urljoin
 
 import estuary_cdk.emitted_changes_cache as cache
@@ -29,7 +30,10 @@ from .models import (
     RestResponseMeta,
 )
 
+P = ParamSpec("P")
+
 HOGQL_PAGE_SIZE = 50_000
+BACKFILL_TIMEOUT_PERIOD = timedelta(minutes=5)
 
 
 # Cache for project IDs per organization (avoids re-fetching on retry).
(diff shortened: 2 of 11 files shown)


source-posthog/acmeCo/Persons.schema.yaml
```diff
@@ -15,6 +15,13 @@ $defs:
         description: Row ID of the Document, counting up from zero, or -1 if not known
         title: Row Id
         type: integer
+      project_id:
+        anyOf:
+        - type: integer
+        - type: 'null'
+        default: null
+        description: The PostHog project this document belongs to
+        title: Project Id
```
Alex-Bair: `_meta/project_id` is part of the composite key, and we shouldn't allow it to be nullable in order to make it easier to materialize `Persons` to an SQL destination. If we did have nullable keys, SQL materializations wouldn't be able to materialize `Persons` out of the box since SQL destinations reject nullable primary keys.

source-posthog/source_posthog/api.py
```diff
@@ -380,34 +390,98 @@ async def fetch_project_events(
...
-                yield item
+        if batch_count < HOGQL_PAGE_SIZE:
+            break
+
+        if datetime.now(tz=UTC) > backfill_timeout:
+            log.info(
+                f"{BACKFILL_TIMEOUT_PERIOD.total_seconds() / 60} "
+                + "minutes have elapsed, emitting a checkpoint"
+            )
+            break
```
Alex-Bair: Thinking through an edge case, what happens if there is a long run of records all with the same cursor value? Is it possible that we the following could happen?
- We fetch the first handful of records at cursor `N`.
- We stop fetching the remaining records at cursor `N` because the `BACKFILL_TIMEOUT_PERIOD` has elapsed.
- We yield `N` as the next cursor value.
- We reinvoke `backfill_persons` and have it start fetching records with cursor values greater than `N`. Meaning, we miss the remaining records at cursor `N` that would have taken more than `BACKFILL_TIMEOUT_PERIOD` to fetch.

That looks possible to me, but let me know if you disagree. I've handled the above situation in other connectors by only breaking out of the pagination loop early when both `BACKFILL_TIMEOUT_PERIOD` has elapsed _and_ the next document's cursor value is greater than the previous document's cursor value.  Here's an example from `source-klaviyo-native` where I implemented that logic:
https://github.com/estuary/connectors/blob/5bf47b1c861d5a9205f4210e956b9320e14fb8b1/source-klaviyo-native/source_klaviyo_native/api/common.py#L299-L312

source-posthog/source_posthog/resources.py
```diff
@@ -337,11 +343,66 @@ def open(
...
+                f"{project_id}": ResourceState.Backfill(cutoff=cutoff, next_page=None)
+                for project_id in project_ids
+            },
+        ),
+        initial_config=ResourceConfig(
+            name=Person.resource_name,
+            interval=timedelta(minutes=5),
+        ),
+        schema_inference=True,
+    )
```
Alex-Bair: Thinking through another edge case, what happens when a new project is added to a user's Posthog account after they've already set up a capture & the capture's state has already been initialized? As the code is now, that new project will have associated entries in `incremental_fetchers` and `backfill_fetchers`, but there will be no associated state for it in the resource state. (The `initial_state` is only used when initializing the resource's state; it's not referenced when there's a key present in the `fetch_changes` / `fetch_page` dicts but that key isn't present in the existing resource state). If there's no state for a `fetch_changes` subtask, the assertion [here](https://github.com/estuary/connectors/blob/5bf47b1c861d5a9205f4210e956b9320e14fb8b1/estuary-cdk/estuary_cdk/capture/common.py#L744-L746) will fail and the connector will crash.

We've addressed this challenge before by adding keys for new projects into the resource state in the `open` function & checkpointing so those state additions persist. This [snippet](https://github.com/estuary/connectors/blob/5bf47b1c861d5a9205f4210e956b9320e14fb8b1/source-shopify-native/source_shopify_native/resources.py#L318-L347) from `source-shopify-native` shows how we've done that elsewhere.

Alex-Bair (approved):
LGTM % small nit of regenerating the `schema.yaml` files with `flowctl raw discover` to reflect that `project_id` is now non-nullable.

Note: We'll need to perform dataflow resets for all production captures' `Persons` binding due to the change in the key field.

PR #3966, opened 2026-03-06
source-http-ingest: support SPKI DER format keys

nicolaslazo:
**Description:**

Some vendors like Twilio provide their public verification keys as a single line, base64-encoded SPKI DER format. This commit implements a check for the presence of the standard PEM headers (`-----BEGIN PUBLIC KEY-----`) and attempts to parse as a raw string of bytes if missing.

**Workflow steps:**

**Notes for reviewers:**

Tests have been conducted through `flowctl preview` with a real, disabled Twilio SendGrid verification key.


psFried (approved):
Looks good. Just to double check my understanding: We call `EcdsaVerifier::new` as part of handling the `Validate` RPC, so we're sure to get a validation error if the key can't be parsed, correct?

Issue #3469, opened 2025-11-03
Query Tag for Snowflake and Databricks


dyaffe:

jgraettinger:
we'd like to use this feature to surface the materialization name

PR #3974, opened 2026-03-10
materialize-databricks: set product identifier and query tag in Databricks SDK

jacobmarble:
**Description:**

Notifies the Databricks service that the client is Estuary, for metrics collection.
Helps #3469

**Notes for reviewers:**

~The semantic version `0.0.0` is required by `databricks.WithProduct()` by regex.~

### DIFF
--- a/materialize-databricks/config.go
+++ b/materialize-databricks/config.go
@@ -194,7 +194,9 @@ func (c config) ToURI(materializationName string) string {
 	if materializationName != "" {
 		// Query tag for Estuary task reporting.
 		// https://docs.databricks.com/aws/en/dev-tools/go-sql-driver#optional-parameters
-		params.Add("query_tags", "materialization-name:"+materializationName)
+		// Underscores are allowed in tag keys.
+		// https://docs.databricks.com/aws/en/sql/user/queries/query-tags#limitations
+		params.Add("query_tags", "materialization_name:"+materializationName)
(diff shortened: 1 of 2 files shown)


mdibaiee:
@jacobmarble @danielnelson my understanding of reading https://github.com/estuary/connectors/issues/3469 is different, I think the goal is to have the materialization task name as the ["tag"](https://docs.databricks.com/aws/en/sql/user/queries/query-tags), not the product. I think adding the product is a fair addition as well, but IIUC we want the materialization task name. Let's clarify, I will open a thread.

materialize-azure-fabric-warehouse/client.go
```diff
@@ -31,7 +31,7 @@ type client struct {
 	templates *templates
 }
 
-func newClient(ctx context.Context, ep *sql.Endpoint[config]) (sql.Client, error) {
+func newClient(ctx context.Context, _ string, ep *sql.Endpoint[config]) (sql.Client, error) {
```
mdibaiee: sorry this is going to be an annoying nit, I think go doesn't complain about unused arguments, but this means in the future developers won't know what this string really is until they look up the source implementation, so might as well just leave this as materializationName everywhere even if not used. I don't feel strongly about this, so if you prefer discarding unused arguments, keep it as-is. what do you think?

PR #4008, opened 2026-03-13
add govulncheck workflow and dependabot config

jacobmarble:
## Summary

Replace Go Dependabot alerts with [`govulncheck`](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck), which performs static reachability analysis to determine whether vulnerable symbols are actually called — eliminating false positives from unused code paths. See [Filippo Valsorda's "Turn Dependabot Off"](https://words.filippo.io/dependabot/) for the rationale.

## What's included

- **`.github/workflows/govulncheck.yaml`** — Runs `govulncheck ./...` weekly (Mondays at noon UTC). On failure, automatically creates a GitHub issue labeled `govulncheck` (with dedup to avoid repeated issues).
- **`.github/dependabot.yml`** — Configures Dependabot for `pip`, `docker`, and `github-actions` ecosystems only. Go (`gomod`) is intentionally excluded since `govulncheck` now covers it.

### DIFF
--- a/.github/workflows/govulncheck.yaml
+++ b/.github/workflows/govulncheck.yaml
@@ -5,6 +5,7 @@ permissions:
   issues: write
 
 on:
+  push:
   schedule:
     - cron: "0 12 * * 1"
 


danielnelson:
Overall I like this change and would like to have it merged.  I tried running govulncheck locally and did get a lot of errors, maybe I'm doing something wrong?

Here is some of the snipped output:
```
govulncheck ./...
-: # github.com/ibmdb/go_ibm_db/api                                                                                                                                                   
../../../usr/go-1.25/pkg/mod/github.com/ibmdb/go_ibm_db@v0.5.4/api/zapi_unix.go:22:11: fatal error: sqlcli1.h: No such file or directory                                              
   22 | // #include <sqlcli1.h>                                                                                                                                                       
      |           ^~~~~~~~~~~                                                                                                                                                         
compilation terminated.
```

Tried it with CGO disabled:
```
CGO_ENABLED=0 govulncheck ./...
/home/dbn/usr/go-1.25/pkg/mod/go.gazette.dev/core@v0.102.0/broker/codecs/zstandard_enable.go:13:73: undefined: zstd.NewReader
/home/dbn/usr/go-1.25/pkg/mod/github.com/marcboeker/go-duckdb/mapping@v0.0.21/mapping_linux_amd64.go:11:22: undefined: bindings.Type
/home/dbn/usr/go-1.25/pkg/mod/github.com/ibmdb/go_ibm_db@v0.5.4/api/api.go:17:9: undefined: SQLSMALLINT
```

jacobmarble:
> Overall I like this change and would like to have it merged. I tried running govulncheck locally and did get a lot of errors, maybe I'm doing something wrong?

Probably not. It needs to build the entire module, and the source-db2 README suggests it is not trivial:
> Note that unlike many connectors, the Db2 connector build requires various
> environment variables to be set. For simplicity, we have a `run_tests.sh`
> script which sets these variables before running the test suite.

I just tried to run the action in GitHub, and it fails with the same error. Not sure why I didn't catch that the first time, but here we are. If I can't find a solution quickly, then I'll close the PR.

PR #4021, opened 2026-03-16
source-pendo: aggregation query optimizations

Alex-Bair:
**Description:**

This PR's scope focuses on quick optimizations to how `source-pendo` requests data from Pendo's [aggregations endpoint](https://engageapi.pendo.io/#7c8479b8-3843-403c-94a9-04cbdf542db9).

Pendo has identified a few areas for query optimization. These include:
- using a 50,000 limit for all queries
- using a full visitor / account / track type journal scan to fetch a small number of documents

This PR addresses those by:
- decoupling the limit used during incremental replication from the limit used during backfills. These limits are also configurable via settings in the endpoint config.
- adding a `timeSeries` in the `source.entity` section of the request bodies. This lets Pendo avoid performing full journal scans and instead only scan the relevant sections of the journal.
- bounding cycle breaker queries (i.e. when we provide a `last_seen_id` to the body generation helper functions) to only the specific timestamp we care about in the associated `timeSeries`.

The spec snapshot change is expected due to the addition of the new advanced incremental and backfill limits.

I recommend reading commit-by-commit.

**Workflow steps:**

The connector's documentation should be updated to reflect the new advanced incremental and backfill limit endpoint config settings.

**Notes for reviewers:**

Tested on a development stack. Confirmed `source-pendo` starts up and captures data like before.

Confirmed the same data is returned for the `Visitor`, `Account`, and `TrackType` API requests both before and after a `timeSeries`  was added; in fact, after adding the `timeSeries`, the Pendo API returned results faster too, which directionally lines up with the anticipated performance improvement.

### DIFF
--- a/source-pendo/source_pendo/api.py
+++ b/source-pendo/source_pendo/api.py
@@ -39,6 +39,30 @@ def _ms_to_dt(ms: int) -> datetime:
 def base_url(host: PendoHost) -> str:
     return f"https://{host}/api/v1"
 
+def _generate_time_series(
+        period: str,
+        lower_bound: int,
+        upper_bound: int | None = None,
+        last_seen_id: str | None = None,
+) -> dict[str, str]:
+    """Builds a timeSeries block for Pendo aggregation queries.
+
+    When paginating within a single timestamp (last_seen_id is set), the scan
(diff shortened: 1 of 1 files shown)


nicolaslazo (approved):
Looks good. Thanks Alex!

PR #4066, opened 2026-03-24
materialize-s3-parquet: fix integration tests

danielnelson:
Switch from localstack to rustfs for s3-compat API and use docker compose for starting the container, based on work that @jacobmarble did for s3-iceberg.  Update the test snapshot since it has been a minute since it ran successfully and added to list of connectors to test.


tests/materialize/materialize-s3-parquet/snapshot.json
```diff
@@ -4,252 +4,314 @@
...
 }
 {
-  "_meta/flow_truncated": false,
-  "array_int": "[91,92]",
+  "array_int": "WzkxLDkyXQ==",
   "binary_field": "YWxvaGEK",
   "bool_field": false,
   "float_field": 99.99,
-  "flow_published_at": "1970-01-01T01:00:22Z",
+  "flow_published_at": 7222000000,
```
mdibaiee: Are these changes expected? have we actually changed the format of timestamps in s3-parquet at some point?

PR #4121, opened 2026-03-30
docs: materializations

mdibaiee:
**Description:**

**Workflow steps:**

**Notes for reviewers:**


materializations.md
```diff
@@ -0,0 +1,298 @@
...
+
+This is handled by `materialize-boilerplate`'s `Constrainter` and `Validator` interfaces, and `materialize-sql` implements `Constrainter` and `Validator`s on behalf of sql connectors.
+
+# Request.Apply
+
+Once the configuration and the various bindings have been validated, the runtime then asks the connector to `Request.Apply` the configurations to the destination. Apply essentially means _set up_ the environment, in most cases this means something like creating the necessary resources (schemas, tables, topics, folders) in the destination, or in cases where the schema of a collection has been updated, it means updating the resources to match the new schema (e.g. altering existing tables to add a new column or drop a colum or change a column's data type, etc.)
```
jacobmarble:
```suggestion
After the configuration and bindings have been validated, the runtime then asks the connector to `Request.Apply` the configurations to the destination.
Apply essentially means _set up or alter_ the destination environment.
In most cases the materialization connector creates the necessary destination resources (e.g. tables, folders, message queues), or when a collection schema has been updated, it means updating the destination resources to match (e.g. add or drop existing table columns, change column data types, etc.).
```

materializations.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Request.Flush
+
+Once all load requests are sent, the runtime sends `Request.Flush`, to which the connector responds with `Request.Flushed` with a connector state. The use case for this message is quite niche, but explained below.
+
+Consider a store like materialize-elastic. Elastic is a document DB suited for point lookup and point update. It doesn’t provide a meaningful bulk query API for reads or write (its "batch" operations don’t scale to the degree we’d need), so the basic strategy is to stream out Load’d documents, and stream in Store’d documents. We throw our hands up and say "this connector is at-least-once".
```
jacobmarble: This paragraph introduces `Store` and at-least-once without explaining them.

materializations.md
```diff
@@ -0,0 +1,298 @@
...
+
+Note that this pattern is at-least-once. A transaction may fail part-way through and be restarted, causing its effects to be partially or fully replayed.
+
+Care must be taken if the collection's schema has reduction annotations such as sum, as those reductions may be applied more than once due to a partially completed, but ultimately failed transaction.
+
+If the collection's schema is last-write-wins, this mode still provides effectively-once behavior. Collections which aren't last-write-wins can be turned into last-write-wins through the use of derivations.
+
+### Push-only Endpoints & Delta Updates
+
+Some systems, such as APIs, Webhooks, and Pub/Sub, are push-only in nature. Estuary's materializations can run in a "delta updates" mode, where loads are always skipped and Estuary does not attempt to store fully-reduced documents. Instead, during the store phase, the runtime sends delta updates which reflect the combined roll-up of collection documents processed only within this transaction.
```
jacobmarble: Much attention is given to the "combined roll-up" process. IIUC:
1. The process combines multiple changes for a key, described by multiple documents referencing that key, into a single change represented by a single document.
2. This process is implemented far upstream of materialization connectors, and the materialize connectors are not exposed to side effects of the process - documents are just documents.

It would be helpful to move these details into a final paragraph such that it is not so distracting. Without it, the delta update feature is trivial to understand, and the intermediate roll-up is a helpful FYI, not critical to understanding or implementing materializations.

materializations.md
```diff
@@ -0,0 +1,298 @@
+# Materializations
+
+Materialization connectors write Flow collections into a destination system, e.g. a database, warehouse, file storage, messaging queue, etc.
+
+This document's aim is to explain in as much detail as possible the materialization protocol (i.e. how does Flow runtime interact with a materialization connector), and explain the various constraints and considerations necessary when implementing a connector.
+
+The actual protocol is defined as protobuf messages [here](https://github.com/estuary/flow/blob/master/go/protocols/materialize/materialize.proto). Here we will reference some of those message types (in this doc these message types are denoted with a `Request.` or `Response.` prefix) and will explain their role. This document also references the [materialize-boilerplate](./materialize-boilerplate) implementation to create a real connection between the raw protocol and the actual boilerplate library we use to develop materializations.
+
```
jacobmarble: This is a good place to "set the table" with some definitions that could be clearer (enhance these suggestions with human or artificial intelligence):
- collection: the storage container from which documents are transported to materialization connectors
- document: a record belonging to a collection, received by the materializer; sometimes delivered as a complete record ("standard updates"), sometimes as a partial record ("delta updates")
- transaction: a batch of documents to be delivered to the destination, *ideally* exactly-once, *in certain cases* at-least-once
- recovery log: the durable Flow log, a record of every action taken by materializers, managed by the runtime

materializations.md
jacobmarble: What do you think about a new format:

`#` Definitions
- control plane, data plane
- collection, connector, runtime

`#` The materialize relationship
What is the runtime is doing, and what does it expect, when it sends Load/Flush/Store/StartCommit/Ack? We don't need runtime implementation details; we need contractual details.

`##` Definitions
- Flow recovery log
- transaction (in the materialize sense)
- materialization protocol: streaming gRPC service defined at ...
- connector state

`##` Spec and Validate: the Control Plane check

`##` Spec, Apply, Open: the Data Plane handshake

`##` Load and Flush: the Data Plane READ handler

`##` Store and StartCommit: the Data Plane WRITE handler

eg "After the runtime receives StartedCommit, it knows that A, B and C, so it commits X, Y, and Z to the recovery log. Now the transaction is (describe the state), so it notifies the materialization connector by sending Acknowledge."

`##` Acknowledge: the Data Plane WRITE OK signal

`#` Transactional patterns and delivery semantics

`##` Definitions
- exactly-once, at-least-once
(Table of the four patterns, column per attribute; processing guarantee, checkpoint authority, ...)

`##` Remote Store is Authoritative

`##` ...

materializations.md
```diff
@@ -0,0 +1,298 @@
...
+When using this pattern, `Request.Acknowledge` is a no-op implementation which is not necessary.
+
+The Recovery Log itself has fencing mechanism, however this fencing mechanism is not available to Destination Authoritative connectors (connectors which commit the data to destination in StartCommit), this is because in StartCommit, we either need to:
+
+- First commit the data to the destination, then tell runtime that the commit has been done so the recovery log can be updated
+- First commit to recovery log by runtime, then commit data to destination
+
+In case of Destinatio Authoritative (committing data to destination in StartCommit), we must be able to finish committing to store before telling runtime to update recovery log -- this means the data is already commited to destination before recovery log can validate fencing. So the zombie process will commit data to destination (write to tables), then ask for update to recovery log, which will fail because of fencing, but it is too late... data has already landed.
```
danielnelson:
```suggestion
In case of Destination Authoritative (committing data to destination in StartCommit), we must be able to finish committing to store before telling runtime to update recovery log -- this means the data is already commited to destination before recovery log can validate fencing. So the zombie process will commit data to destination (write to tables), then ask for update to recovery log, which will fail because of fencing, but it is too late... data has already landed.
```

materializations.md
mdibaiee: This is a "niche" hypothetical use case for the Flushed message having a state.

materializations.md
mdibaiee: I would like to defer to other docs for definitions of those parts. For connectors and collections we have https://docs.estuary.dev/concepts/, and for control plane and data plane, I assume Joaquin's docs on data plane and operations will include some docs on what a data plane is and we can refer to it.

I will add references to existing docs for as many of these words as do have an available reference the first time they are mentioned 👍🏽

materializations.md
```diff
@@ -0,0 +1,298 @@
...
+
+A connector creates staging tables where the data is loaded into before being commited to the destination table. These staging tables are named `flow_stage_<target_table_name>`. With this naming pattern, if two instances are spawned, they will both try to create, and write to the same table, resulting in a mixing of data between them, so instead the connector should create its staging tables as `flow_stage_<keyBegin>_<target_table_name>`.
+
+Once the materialization has been "Opened", the connector stays running and it receives the following messages from the runtime in a loop:
+
```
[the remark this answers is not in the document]
mdibaiee: I moved the Sequence Diagram here

materializations.md
```diff
@@ -0,0 +1,298 @@
...
+Without delta updates, Estuary would reduce documents -1, 3, and 2 by `sum` to arrive at document 4, which is stored. The next transaction, document 4 is loaded and reduced with 6, -7, and -1 to arrive at a new stored document 2. This document, 2, represents the full reduction of the collection documents materialized thus far.
+
+Compare to delta updates mode: collection documents -1, 3, and 2 are combined to store a delta-update document of 4. The next transaction starts anew, and 6, -7, and -1 combine to arrive at a delta-update document of -2. These delta updates are a windowed combine over documents seen in the current transaction only, and unlike before are not a full reduction of the document. If delta updates were written to pub/sub, note that a subscriber could further reduce over each delta update to recover the fully reduced document of 2.
+
+Note that many use cases require only `lastWriteWins` reduction behavior, and for these use cases delta updates does the "right thing" by trivially re-writing each document with its most recent version. This matches the behavior of Kafka Connect, for example.
```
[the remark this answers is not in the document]
mdibaiee: I believe what is meant is that if the destination system itself has a sort of `lastWriteWins` way of reducing documents (such as Kafka Connect), then the documents will end up with the fully reduced document when using delta updates as well, not that our materializations will have the same outcome with delta updates.

materializations.md
mdibaiee: Can you give your explanation of delta updates without the roll-up being involved? In my mind they are tangled

materializations.md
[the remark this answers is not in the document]
mdibaiee: Also added a cycle in text 👍🏽

materializations.md
```diff
@@ -0,0 +1,298 @@
...
+        "name": "estuary/source-something/my-lovely-collection",
+      },
+      "fields": {
+        "require": {
+          "some-optional-field-i-want-to-include": {}
+        },
+        "recommended": true
+      },
+      "backfill": 0,
+      "onIncompatibleSchemaChange": "backfill"
```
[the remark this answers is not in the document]
mdibaiee: Yeah this specific property is an extra one which is not part of the protocol, but rather is part of properties used by the control plane to manage bindings

danielnelson (approved):
Looks great, should we remove `docs/materialize/README.md` as part of this change?

mdibaiee:
@aeluce do you think we could replace https://docs.estuary.dev/reference/Connectors/materialization-protocol/ with this, or do you think this is too technical / too close to the source code of connectors?

mdibaiee:
I'm going to merge this, feel free to put comments post-merge and we can incorporate as a separate PR and also if you generally see the possibility of improving it further, feel free to open a PR anytime of course.

materializations.md
jacobmarble: For example, table `t` has columns `a`, `b`, `c`. The primary key is `a`. The collection stores these updates:
```
a=1, b=1
a=1, c=1
a=2, b=2
```

---

If spread across multiple transactions, then the materialize connector might receive:
```
Transaction A
a=1, b=1
```
```
Transaction B
a=1, c=1
a=2, b=2
```

---

If the upstream roll-up process does its thing, then the materialize connector might receive:
```
Transaction A
a=1, b=1, c=1
a=2, b=2
```

---

Whether the materialize connector gets the rolled-up transaction or not, it doesn't really matter, the logic from the materializer point of view is identical.

materializations.md
```diff
@@ -0,0 +1,298 @@
...
+At query time, the client needs to know a committed-through T. It then scans the store and takes the largest version of each key <= T.
+
+**Upsides**: No external storage, and this write pattern is well suited to some column-family storage architectures (thinking about a materialize-bigtable in particular).
+**Downsides**: Much more complexity for the client. Aside from internal usage, only very sophisticated users would be successful using this. (edited)
+
+## Request.Store + Request.StartCommit + Request.Acknowledge
+
+These three messages are highly related, and depending on the various patterns for writing a materialization connector, they will be used very differently. So here we explain some of the various patterns, use cases, scale-out strategies, limitations and nuances for each.
+
+Before diving into the patterns, one general constraint that is largely what dictates the patterns must be explained: transactionality. Flow sends documents to materializations as part of a transaction, which must either all be committed, or none at all.
```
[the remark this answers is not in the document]
jacobmarble: Sorry, I was terse. I'm thinking about two of the implementation patterns.

"Recovery Log with Non-Transactional Store":
> A transaction may fail part-way through and be restarted, causing its effects to be partially or fully replayed.

Elastic, MongoDB, and DynamoDB all commit incremental batches directly to the target, so a failed transaction could leave partial changes in the target.

"Push-only Endpoints":
Webhook, PubSub, and Slack similarly commit directly to target, so a failed transaction would cause documents to be sent twice.

materializations.md
mdibaiee: I think I would say this point is not entirely correct:

> This process is implemented far upstream of materialization connectors, and the materialize connectors are not exposed to side effects of the process - documents are just documents.

materialization connectors are part of this roll-up process by loading the documents from the destination so that the roll-up can happen.

Flow runtime does not go back and read the old documents of the collection when it sees a new document, it asks the materialization to `Load` the document, so that it can run reduction on that document, not on the history of the collection.

So it is not really an _upstream_ roll-up process, it's a roll-up process that happens between the runtime and the materialization.

Does that change things? Is there something we could improve in docs to clarify this?

materializations.md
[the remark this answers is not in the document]
mdibaiee: Yeah I think we mean that's the constraint we would _like_ to have, so long as it is possible. In the cases you mention it is not possible to have that sort of transactionality, so we fallback to a looser constraint for now.

Feel free to add an update to the docs to mention it as you see fit

materializations.md
jacobmarble: > > This process is implemented far upstream of materialization connectors, and the materialize connectors are not exposed to side effects of the process - documents are just documents.
> 
> materialization connectors are part of this roll-up process by loading the documents from the destination so that the roll-up can happen.
>
> Flow runtime does not go back and read the old documents of the collection when it sees a new document, it asks the materialization to `Load` the document, so that it can run reduction on that document, not on the history of the collection.

That's helpful, "far upstream" is obviously wrong. Do I have this right?

"Per key, **the runtime reduces (or rolls up) `Load` results with document updates** to create a **new document version**. The new document version is passed to `Store`, and the destination service replaces the previous document version with the new one."

---

> the runtime sends delta updates which reflect the combined roll-up of collection documents processed only within this transaction

The cited text sounds like the same reduction process, except the runtime doesn't call `Load` in Delta Updates mode. Do I have this part right?

"Per key, **the runtime reduces (or rolls up) document updates** to create a **single document update**. This document update is passed to `Store`, leaving final reduction, if any, to the destination service."

mdibaiee: Yes! I think you have both of them right.

In delta updates we only combine the various updates to the same key together, whereas in normal mode, we combine those with the loaded document

PR #4202, opened 2026-04-06
source-hubspot-native: improve memory usage & other misc. improvements

Alex-Bair:
**Description:**

We've observed `source-hubspot-native` consistently OOM when there are millions of delayed changes to catch up on in a single `fetch_delayed_changes` invocation. This PR attempts to avoid those OOMs by:
 - Refactoring ID fetching from a callback-based pagination pattern to async generators, allowing `fetch_changes_with_associations` to process IDs in bounded chunks instead of buffering all records in memory.
  - Stopping legacy API pagination once results are at or before the since cursor, since these APIs return results newest-first.                                                     
  - Replacing the global deduplication set in `fetch_search_objects` with a small set tracking only IDs at the current `max_updated` timestamp, used for deduplication at the 10k search boundary.       
  - Switching delayed streams from `should_yield` (which inserted into the emitted changes cache as a side effect) to a pure `has_as_recent_as` check, limiting cache growth.

Additionally, realtime streams now tolerate out-of-order search results by logging a warning and skipping the record instead of raising an exception.   

There are quite a few changes here, and a lot of the diff is a result of changing from callbacks to async generators & the resulting interface changes. I recommend reading commit-by-commit

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api/companies.py
+++ b/source-hubspot-native/source_hubspot_native/api/companies.py
@@ -42,7 +42,13 @@ def fetch_recent_companies(
                 await http.request(log, url, params=params)
             )
             for r in result.results:
-                yield (ms_to_dt(r.properties.hs_lastmodifieddate.timestamp), str(r.companyId))
+                ts = ms_to_dt(r.properties.hs_lastmodifieddate.timestamp)
+                # This API returns results newest-first, so once we
+                # see a record at or before `since` there's nothing
+                # left worth fetching.
+                if ts <= since:
+                    return
+                yield (ts, str(r.companyId))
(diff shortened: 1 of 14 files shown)


Alex-Bair:
Note for myself: This PR still needs tested out on a local stack to exercise backfills and incremental replication with CRM objects to flush out any bugs or behavior changes I haven't caught yet.

Alex-Bair:
This PR is superseded by fhttps://github.com/estuary/connectors/pull/4631. While it may not be as memory efficient as the solution in the PR aims to be, it should still address the memory concerns when there are millions of changes in a relatively small time window.

PR #4229, opened 2026-04-13
source-stripe-native: speed up fetching connected account ids with concurrent worker system

Alex-Bair:
**Description:**

The previous `_fetch_connected_account_ids` implementation was a single sequential paginator through GET `/v1/accounts`. For platforms with 36k+ connected accounts, this was taking over an hour. That's too slow since all connected account ids are fetched each time the capture starts up; we'd rather not spend an hour fetching connected account ids before capturing any data.

This commit replaces that sequential paginator with a concurrent worker system modeled after `source-klaviyo-native`'s [events backfill](https://github.com/estuary/connectors/blob/main/source-klaviyo-native/source_klaviyo_native/api/events.py). The concurrent worker system partitions the time range into chunks using Stripe's `created[gte]`/`created[lte]` query parameters and has multiple workers paginate through their respective chunks in parallel.

Workers detect dense time windows (chunks that take >30s to paginate) and, when idle workers are available, they subdivide the remaining unprocessed range into smaller chunks for other workers to pick up.

**Notes for reviewers:**

This is an isolated change that's a drop in replacement for how the connector fetches connected account ids. The very similar concurrent worker system in `source-klaviyo-native` has been working well for multiple months, and I anticipate it'll work well in `source-stripe-native` too to speed up fetching all connected account ids by at least 5x, likely more.

### DIFF
--- a/source-stripe-native/source_stripe_native/account_fetcher.py
+++ b/source-stripe-native/source_stripe_native/account_fetcher.py
@@ -1,0 +1,314 @@
+import asyncio
+import time
+import traceback
+from asyncio import CancelledError
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from logging import Logger
+
+from estuary_cdk.http import HTTPSession
+from estuary_cdk.utils import format_error_message
+
+from .api import API
+from .models import Accounts, ListResult
+
+
+# Stripe launched in 2011. No connected accounts can exist before this date.
+STRIPE_EPOCH = int(datetime(2011, 2, 1, tzinfo=UTC).timestamp())
+
+
+@dataclass
+class TimestampChunk:
+    start: int
+    end: int
+
+
+@dataclass
+class AccountFetchConfig:
+    initial_chunks: int = 50
+    num_workers: int = 10
+
+    # How many seconds a worker processes a chunk before it is considered
+    # a dense time window and should be divided into smaller chunks.
+    dense_chunk_threshold_seconds: int = 30
+
+    # Minimum chunk size in seconds (1 minute) to prevent over-subdivision.
+    minimum_chunk_seconds: int = 60
+
+    work_queue_size: int = 100
+
+    page_limit: int = 100
+
+
+DEFAULT_CONFIG = AccountFetchConfig()
+
+
+def split_timestamp_range(
+    start: int,
+    end: int,
+    num_chunks: int,
+    minimum_chunk_seconds: int,
+) -> list[TimestampChunk]:
+    """Split a unix timestamp range into roughly equal-sized chunks.
+
+    If minimum_chunk_seconds would be violated, fewer chunks are produced.
+    Returns a list of TimestampChunks covering the full range.
+    """
+    if num_chunks <= 0:
+        raise ValueError("num_chunks must be positive")
+
+    if start >= end:
+        raise ValueError("start must be before end")
+
+    total_duration = end - start
+    num_chunks = max(
+        1,
+        min(
+            num_chunks,
+            total_duration // minimum_chunk_seconds,
+        ),
+    )
+    chunk_duration = total_duration / num_chunks
+
+    chunks: list[TimestampChunk] = []
+    current_start = start
+
+    for i in range(num_chunks):
+        if i == num_chunks - 1:
+            chunk_end = end
+        else:
+            chunk_end = int(current_start + chunk_duration)
+
+        chunks.append(TimestampChunk(start=current_start, end=chunk_end))
+        current_start = chunk_end
+
+    return chunks
+
+
+class AccountWorkManager:
+    def __init__(
+        self,
+        http: HTTPSession,
+        log: Logger,
+        config: AccountFetchConfig = DEFAULT_CONFIG,
+    ):
+        self.http = http
+        self.log = log
+        self.config = config
+
+        self.work_queue: asyncio.Queue[TimestampChunk | None] = asyncio.Queue(maxsize=config.work_queue_size)
+
+        self._active_worker_count = 0
+        self.first_worker_error: str | None = None
+
+        self.account_ids: set[str] = set()
+
+    def mark_worker_active(self) -> None:
+        if self._active_worker_count >= self.config.num_workers:
+            raise Exception(f"A worker attempted to mark itself active when the active worker count is {self._active_worker_count}.")
+        self._active_worker_count += 1
+
+    def mark_worker_inactive(self) -> None:
+        if self._active_worker_count <= 0:
+            raise Exception(f"A worker attempted to mark itself inactive when the active worker count is {self._active_worker_count}.")
+        self._active_worker_count -= 1
+
+    def are_active_workers(self) -> bool:
+        return self._active_worker_count > 0
+
+    def has_idle_workers(self) -> bool:
+        return self._active_worker_count < self.config.num_workers
+
+    async def fetch_account_ids(self, start: int, end: int) -> set[str]:
+        initial_chunks = self._create_initial_chunks(start, end)
+        for chunk in initial_chunks:
+            self.work_queue.put_nowait(chunk)
+
+        # Purely diagnostic: logs task outcomes to aid debugging when a task
+        # fails or is cancelled unexpectedly. Not used for control flow —
+        # the TaskGroup handles exception propagation and cancellation.
+        def callback(task: asyncio.Task):
+            task_name = task.get_name()
+            status: str = ""
+            stack_trace: str | None = None
+
+            if task.cancelled():
+                status = "cancelled"
+            elif exc := task.exception():
+                status = f"failed with exception {format_error_message(exc)}"
+                if exc.__traceback__:
+                    stack_trace = "\nStack trace:\n" + "".join(traceback.format_list(traceback.extract_tb(exc.__traceback__)))
+            else:
+                status = "completed"
+
+            self.log.debug(f"Task {task_name} {status}.", {
+                "first_worker_error": self.first_worker_error,
+                "active_worker_count": self._active_worker_count,
+                "stack_trace": stack_trace,
+            })
+
+        self.log.debug("Starting concurrent account fetch workers.")
+        async with asyncio.TaskGroup() as tg:
+            for i in range(self.config.num_workers):
+                worker_id = i + 1
+                task = tg.create_task(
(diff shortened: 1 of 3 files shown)


Alex-Bair:
Note: most of the CI checks failed due to an intermittent GitHub issue, but the one for `source-stripe-native` did succeed once GitHub stopped returning 504s.

source-stripe-native/source_stripe_native/account_fetcher.py
```diff
@@ -0,0 +1,407 @@
...
+    if start >= end:
+        raise ValueError("start must be before end")
+
+    total_duration = end - start
+    chunk_duration = total_duration / num_chunks
+
+    if chunk_duration < minimum_chunk_seconds:
+        actual_num_chunks = max(1, total_duration // minimum_chunk_seconds)
+        chunk_duration = total_duration / actual_num_chunks
+        num_chunks = actual_num_chunks
```
nicolaslazo: Nit: this block could be turned into a one-liner by combining `min` and `max` clamps

```python
chunk_duration = total_duration // num_chunks
num_chunks = max(1, min(num_chunks, total_duration // minimum_chunk_seconds))
```

source-stripe-native/source_stripe_native/account_fetcher.py
```diff
@@ -0,0 +1,407 @@
...
+        log.debug(f"Account worker {worker_id} started.")
+
+        is_looking_for_work = False
+        while not shutdown_event.is_set():
+            chunk = await queue_manager.get_work_chunk(timeout=config.worker_timeout)
+            if chunk is None:
+                if not is_looking_for_work:
+                    log.debug(f"Account worker {worker_id} is idle and checking for new work.")
+                    is_looking_for_work = True
+                continue
```
nicolaslazo: Rather than polling for new work while they observe `shutdown_event`, workers could block indefinitely and shut down upon encountering a sentinel value in the work queue, similar to the pattern established in `source-hubspot-native` [here](https://github.com/estuary/connectors/blob/main/source-hubspot-native/source_hubspot_native/api/contact_list_memberships.py).

This is more or less what the shutdown monitor would do to propagate the signal:

```python
while True:
    await asyncio.sleep(1)
    if not self.are_active_workers() and self.queue_manager.are_all_queues_empty():
        for _ in range(self.config.num_workers):
            await self.queue_manager.work_queue.put(None)
        await self.queue_manager.subdivision_request_queue.put(None)
        break
```

This change would eliminate
  - `shutdown_event`
  - `worker_timeout` + `subdivision_timeout` config fields
  - The `is_looking_for_work` logic
  - The `get_work_chunk` + `get_subdivision_chunk` timeout wrappers

source-stripe-native/source_stripe_native/account_fetcher.py
```diff
@@ -0,0 +1,407 @@
...
+            minimum_chunk_seconds=config.minimum_chunk_seconds,
+        )
+        divided_chunks = [
+            TimestampChunk(start=s, end=e)
+            for s, e in divided_ranges
+        ]
+
+        await queue_manager.put_work_chunks(divided_chunks)
+
+    log.debug("Account subdivision worker exited.")
```
nicolaslazo: As far as I understand, the subdivision worker's job is to call `split_timestamp_range` on a dense time slice and push the subdivisions back into the main queue. IMO it would be more straightforward to inline this `split_timestamp_range` call and eliminate this worker type altogether. Though it would make the main worker's logic a little more complex, it would make the secondary queue and associated config params redundant and simplify the shutdown worker's trigger condition.

source-stripe-native/source_stripe_native/account_fetcher.py
```diff
@@ -0,0 +1,407 @@
...
+
+DEFAULT_CONFIG = AccountFetchConfig()
+
+
+def split_timestamp_range(
+    start: int,
+    end: int,
+    num_chunks: int,
+    minimum_chunk_seconds: int,
+) -> list[tuple[int, int]]:
```
nicolaslazo: Nit: every `split_timestamp_range` reference is followed by a `TimestampChunk` constructor call — we could make the return type of this definition `list[TimestampChunk]`.

If you'd still like to make use of tuple unpacking, we could replace the existing `TimestampChunk` dataclass definition with a `NamedTuple`.

source-stripe-native/source_stripe_native/account_fetcher.py
```diff
@@ -0,0 +1,407 @@
...
+        # the subdivision worker runs without the monitor getting a turn.
+        self.log.debug("Account fetch shutdown monitor started.")
+        while not self.shutdown_event.is_set():
+            if not self.are_active_workers() and self.queue_manager.are_all_queues_empty():
+                self.log.debug("All account fetch work complete - initiating shutdown.")
+                self.shutdown_event.set()
+                break
+
+            await asyncio.sleep(1)
+        self.log.debug("Account fetch shutdown monitor exited.")
```
nicolaslazo: If we end up with a single queue design, we might not need a shutdown monitor at all. After having populated the queue with the initial tasks, we could block on their completion and post sentinels afterwards

```python
async with asyncio.TaskGroup() as tg:
    for i in range(config.num_workers):
        tg.create_task(worker(...))

    await work_queue.join()
    for _ in range(config.num_workers):
        work_queue.put_nowait(None)
```

In order for this design to work, we'll have to call `work_queue.task_done()` after each time slice has been processed. This also solves the theoretical race condition: we can push the slice subdivisions to the queue before calling `.task_done()` for the parent.

This proposal should replace connector-specific synchronisation logic with standard library primitives.

nicolaslazo (changes requested):
It's great that we're tackling this issue, we're finding some sizable accounts out there. I left a few ideas on how we could reduce the code footprint

source-stripe-native/source_stripe_native/account_fetcher.py
Alex-Bair: I originally wrote `split_timestamp_range` to return `list[tuple[int, int]]` so it _could_ be considered a generic utility function that we could move into the CDK and use in multiple connectors. But as it is now, it doesn't live in the CDK, and if it did we would need to consider if we'd want to make it even more generic (i.e. accept either `int` or `datetime` bounds? etc.). And now that I'm thinking about it more, we'd probably want it to return `NamedTuple`s or a more fleshed out `TimestampChunk` dataclass anyway instead of raw `tuple`s.

I'm good with having it return `list[TimestampChunk]` instead since the function currently is an internal helper for just this file. I'll update it.

source-stripe-native/source_stripe_native/account_fetcher.py
Alex-Bair: Yeah, that's correct. The subdivision worker's job is pretty simple. I originally did that to reduce the responsibilities of chunk workers so I could reason about the design more easily. But I have a pretty good handle on how the design works now & your other suggestions help reduce the complexity further, so I'm good with adding "subdividing chunks and putting them back on the work queue" to the responsibilities of chunk workers. I'll update it.

source-stripe-native/source_stripe_native/account_fetcher.py
Alex-Bair: Nice, thanks for the suggestion. I didn't realize / think of using `.join()` and `.task_done()` to determine when all work has been completed. That helps simplify quite a bit, I'll update the code to use those primitives.

source-stripe-native/source_stripe_native/account_fetcher.py
Alex-Bair: Good point, I agree this is cleaner & simplifies the architecture a bit in combination with your other suggestions. I'll update.

source-stripe-native/source_stripe_native/account_fetcher.py
Alex-Bair: I actually prefer using multiple lines in this case to help with readability. I'm not against condensing it into a single assignment though:
```python
total_duration = end - start
num_chunks = max(
    1,
    min(
        num_chunks,
        total_duration // minimum_chunk_seconds,
    ),
)
chunk_duration = total_duration / num_chunks
```

source-stripe-native/source_stripe_native/account_fetcher.py
nicolaslazo: I just discovered it while working on this review. It might make my `source-hubspot-native` implementation simpler, too

Alex-Bair:
Thanks for the review & feedback! They helped simplify the code a bit & this should be ready for review again.

source-stripe-native/source_stripe_native/account_fetcher.py
nicolaslazo: I didn't consider it could become a CDK helper!

source-stripe-native/source_stripe_native/account_fetcher.py
```diff
@@ -0,0 +1,312 @@
...
+                        break
+
+                    # Stripe returns results in descending created order. After paginating
+                    # partway through chunk [S, E], the worker has processed accounts from E
+                    # down to last_created. The unprocessed range is [S, last_created].
+                    if (
+                        is_dense_chunk
+                        and last_created is not None
+                        and last_created > chunk.start
+                        and work_manager.has_idle_workers()
```
nicolaslazo, line 248: Now that we no longer check for active workers for our shutdown condition, this is the only non-logging reference to the worker counter left. I think we could get away with removing it—workers could pick these up after they finish what they're working on.

nicolaslazo (approved):
Looks great! I left one more potential refactor, but I'd say we are good to merge

source-stripe-native/source_stripe_native/account_fetcher.py
Alex-Bair: You're right that it's the only non-logging reference left to the worker count. But I'd prefer we keep it in since it prevents workers from endlessly subdividing their chunks and flooding the work queue. In the extreme case where _all_ chunks are dense, workers would continuously subdivide and the work queue would grow until the connector runs out of memory. This `has_idle_workers()` check helps prevent that by providing backpressure.

I'm also going to add a `maxsize` to the work queue so there's a safety net in case the queue does grow larger than we anticipate it should. It'd fail with a `QueueFull` exception, which is a lot easier to investigate than an ambiguous OOM.

PR #4233, opened 2026-04-13
source-*: check `rows.Err()` after all `rows.Next()` loops

Alex-Bair:
**Description:**

Per Go's [database/sql docs](https://pkg.go.dev/database/sql#Rows.Next), `rows.Next()` returns `false` both when the result set is exhausted and when an error is encountered. `rows.Err()` must be checked afterward to distinguish the two cases. Without this check, a mid-iteration error causes the function to return success with a partial result set.

**Notes for reviewers:**

There are also materializations that aren't checking `rows.Err()` after iterating through `rows.Next()`, but I'm going to split those off into a different PR for the materialization team to review.

### DIFF
--- a/source-oracle-batch/discovery.go
+++ b/source-oracle-batch/discovery.go
@@ -239,6 +239,9 @@ func discoverTables(ctx context.Context, db *sql.DB, discoverSchemas []string) (
 			Name:  tableName,
 		})
 	}
+	if err := rows.Err(); err != nil {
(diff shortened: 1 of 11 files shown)


willdonnelly (approved):
LGTM

If I'm reading this correctly it looks like there's one significant data-correctness issue fixed here (in the source-oracle backfill code) and the rest is all discovery/prerequisites/diagnostics (which are still good to detect errors on, but significantly less likely to cause real user harm). That's better than I expected!

PR #4235, opened 2026-04-13
materialize-snowflake: skip rename if no encryption key

danielnelson:
Checkpoints created before the encryption key was added do not have the information required to be rewritten and renamed.  The previous change guessed that the current channel key was the one used to write the files originally but it did not always work.  With this change we use the prior strategy of doing a registration on the original file.

related: #4047

**Workflow steps:**

No changes

None

**Notes for reviewers:**


mdibaiee (approved):

williamhbaker (approved):
LGTM

PR #4236, opened 2026-04-13
fix(clickhouse): keep store table schema in sync with target

jacobmarble:
**Description:**

When the runtime changes the destination schema by calling `Apply()`, and store tables exist for the target table, then the next call to `Acknowledge()` fails because it implements the store-commit operation with `ALTER TABLE ... MOVE PARTITION ...`, which requires that the tables have a common schema.

This fixes the bug by applying every schema change to both the target and store tables.

**Workflow steps:**

This is a bug fix that requires no change to customer workflows.

**Notes for reviewers:**

### DIFF
--- a/materialize-clickhouse/.snapshots/TestSQLGeneration
+++ b/materialize-clickhouse/.snapshots/TestSQLGeneration
@@ -46,6 +46,22 @@ ORDER BY (theKey)
 ;
 --- End delta_updates createTargetTable ---
 
+--- Begin key_value queryStoreTablesAllRangeKeys ---
+
+SELECT name FROM system.tables
+WHERE database = currentDatabase()
+  AND name LIKE 'flow_temp_store_%_key_value'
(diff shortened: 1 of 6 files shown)


mdibaiee:
Posting this from Slack here for the record:

This actually leads to correctness issues: it is possible that the table changes in a way that requires re-creating the table. Your PR clones all actions on the target table, on the temporary stage tables. This means if a transaction was staged for Acknowledge, but now the target table is modified in a way that requires re-creating the table, the DropTable call will delete the staged table and continue as if the transaction has been committed, but it hasn’t in reality, and there will be no error, but there will be missing data as a result.

We can’t handle Apply before Acknowledge until the runtime fixes it for it. For the time being, backfilling is the only way to deal with it.

mdibaiee:
I think this can be closed yeah? @jacobmarble

PR #4240, opened 2026-04-13
materialize-*: handle null valued keys in MergeBounds

mdibaiee:
**Description:**

- Compound keys can have null values in one of their key parts. For merge bounds, what that means is we should include that in our merge bound. I have updated MergeBounds so that it detects existence of null values for a specific key, and if that's the case, we add `OR IS NULL` as a condition.
- Previously we would panic on seeing null values:
```
runTransactions: app.ConsumeMessage: panic: minKey unhandled key type <nil> (value: <nil>)
```

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-azure-fabric-warehouse/.snapshots/TestSQLGeneration
+++ b/materialize-azure-fabric-warehouse/.snapshots/TestSQLGeneration
@@ -166,9 +166,9 @@ WITH (
 SELECT 0, r.flow_document
 FROM flow_temp_table_load_0 AS l
 JOIN "a-warehouse"."a-schema".key_value AS r
-	 ON  l.key1 = r.key1 AND r.key1 >= 10 AND r.key1 <= 100
-	 AND l.key2 = r.key2
-	 AND BASE64_DECODE(l."key!binary") = r."key!binary"
+	 ON  (l.key1 = r.key1 AND r.key1 >= 10 AND r.key1 <= 100)
+	 AND (l.key2 = r.key2)
+	 AND (BASE64_DECODE(l."key!binary") = r."key!binary")
 --- End "a-warehouse"."a-schema".key_value loadQuery ---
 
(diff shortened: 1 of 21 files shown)


materialize-sql/templating.go
```diff
@@ -96,8 +99,9 @@ type MergeBoundsBuilder struct {
 	keyColumns []Column
 	literaler  func(any) string
 
-	lower []any
-	upper []any
+	lower  []any
+	upper  []any
+	isNull []bool
```
danielnelson, line 104: Maybe we should call these new fields hasNull?  The key could have some normal values and some null, and this is true if there are any null keys.

materialize-azure-fabric-warehouse/.snapshots/TestSQLGeneration
```diff
@@ -257,6 +257,98 @@ WHERE _flow_delete = 0;
 DROP TABLE flow_temp_table_store_0;
 --- End "a-warehouse"."a-schema".key_value storeMergeQuery ---
 
+--- Begin "a-warehouse"."a-schema".key_value loadQuery (with null bounds) ---
+SELECT 0, r.flow_document
+FROM flow_temp_table_load_0 AS l
+JOIN "a-warehouse"."a-schema".key_value AS r
+	 ON  (l.key1 = r.key1 AND r.key1 >= 10 AND r.key1 <= 100 OR r.key1 IS NULL)
```
danielnelson, line 264: I love the parens, maybe go even further and put them around the different operators?

```suggestion
	 ON  ((l.key1 = r.key1 AND r.key1 >= 10 AND r.key1 <= 100) OR r.key1 IS NULL)
```

danielnelson (approved):

dyaffe:
Let me know if it's possible or worthwhile to test this first on the broken task.  Would like to get it fixed before it permanently breaks.

mdibaiee:
This turned out to be an issue with control plane, closing for now

PR #4243, opened 2026-04-14
materialize-snowflake: detect QUOTED_IDENTIFIERS_IGNORE_CASE during validation

jwhartley:
## Summary

When `QUOTED_IDENTIFIERS_IGNORE_CASE` is enabled at the Snowflake account or user level, Snowflake uppercases quoted identifiers even when they were created with lowercase casing. This breaks the connector, which uses quoted identifiers for column names containing special characters (e.g. `"_meta/op"`), and expects Snowflake to preserve their case exactly as created.

The failure mode is confusing: the connector creates `"_meta/op"`, reads back the table schema and sees `_META/OP`, thinks the column doesn't exist, tries to `ALTER TABLE ... ADD COLUMN "_meta/op"`, and hits a `column '_META/OP' already exists` error deep in `completeRecovery`. This has come up multiple times with different customers.

This PR adds a prereq check that detects the setting during validation and surfaces a clear, actionable error with remediation instructions rather than silently failing mid-run.

Disclaimer: I am not a software engineer

## Test plan

- [ ] Verify the prereq check fires with a descriptive error when connecting to a Snowflake account/user with `QUOTED_IDENTIFIERS_IGNORE_CASE = TRUE`
- [ ] Verify normal materialization setup is unaffected when the setting is `FALSE` (default)

### DIFF
--- a/materialize-snowflake/client.go
+++ b/materialize-snowflake/client.go
@@ -315,7 +315,10 @@ func preReqs(ctx context.Context, cfg config) *cerrors.PrereqErr {
 		if err != nil {
 			errs.Err(fmt.Errorf("checking for QUOTED_IDENTIFIERS_IGNORE_CASE: %w", err))
 		} else {
-			if strings.EqualFold(quotedIdentifiersIgnoreCase, "true") {
+			ignoreCase, err := strconv.ParseBool(value)
+			if err != nil {
+				errs.Err(fmt.Errorf("unable to parse QUOTED_IDENTIFIERS_IGNORE_CASE value: %w", err))
+			} else if ignoreCase {
 				errs.Err(fmt.Errorf(
 					"the Snowflake parameter QUOTED_IDENTIFIERS_IGNORE_CASE is enabled, which is incompatible with this connector: " +
 						"Estuary uses quoted identifiers for column names containing special characters (e.g. \"_meta/op\"), " +
--- a/materialize-snowflake/client.go
+++ b/materialize-snowflake/client.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"regexp"
 	"slices"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -319,13 +320,7 @@ func preReqs(ctx context.Context, cfg config) *cerrors.PrereqErr {
 			if err != nil {
 				errs.Err(fmt.Errorf("unable to parse QUOTED_IDENTIFIERS_IGNORE_CASE value: %w", err))
 			} else if ignoreCase {
-				errs.Err(fmt.Errorf(
-					"the Snowflake parameter QUOTED_IDENTIFIERS_IGNORE_CASE is enabled, which is incompatible with this connector: " +
-						"Estuary uses quoted identifiers for column names containing special characters (e.g. \"_meta/op\"), " +
-						"and requires Snowflake to preserve their case exactly as created. " +
-						"Disable this parameter for the Estuary user with: " +
-						"ALTER USER <estuary_user> SET QUOTED_IDENTIFIERS_IGNORE_CASE = FALSE",
-				))
+				errs.Err(fmt.Errorf("QUOTED_IDENTIFIERS_IGNORE_CASE must be FALSE, see go.estuary.dev/mat-snow"))
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


materialize-snowflake/client.go
```diff
@@ -304,6 +304,23 @@ func preReqs(ctx context.Context, cfg config) *cerrors.PrereqErr {
 				errs.Err(fmt.Errorf("no warehouse configured and default warehouse not set for user '%s': must set a value for 'Warehouse' in the endpoint configuration", cfg.Credentials.User))
 			}
 		}
+
+		// Check that QUOTED_IDENTIFIERS_IGNORE_CASE is not enabled. When this Snowflake parameter
+		// is enabled, quoted identifiers are returned uppercased regardless of how they were
+		// created. Estuary uses quoted identifiers for column names containing special characters
+		// (such as "_meta/op"), and requires that Snowflake preserve their case exactly as created.
+		var quotedIdentifiersIgnoreCase string
+		if err := db.QueryRowContext(ctx, "SELECT SYSTEM$GET_SESSION_PARAMETER('QUOTED_IDENTIFIERS_IGNORE_CASE');").Scan(&quotedIdentifiersIgnoreCase); err == nil {
```
danielnelson, line 313: This doesn't seem to be a valid query, I think it should be:
```
SHOW PARAMETERS LIKE 'QUOTED_IDENTIFIERS_IGNORE_CASE' IN SESSION;
```

You won't see any error because you test for `err == nil` and so only return an error if this query ran successfully.  What we should do is check for a query error and report it, and if there is none we can continue to check for the desired value.  (The error handling in this function is a little abnormal because we try to report multiple errors.)


```suggestion
		var key, value, default_, level, desc, type_ string
		query := "SHOW PARAMETERS LIKE 'QUOTED_IDENTIFIERS_IGNORE_CASE' IN SESSION;"
		err := db.QueryRowContext(ctx, query).Scan(&key, &value, &default_, &level, &desc, &type_)
		if err != nil {
			errs.Err(fmt.Errorf("checking for QUOTED_IDENTIFIERS_IGNORE_CASE: %w", err))
		} else {
```

materialize-snowflake/client.go
```diff
@@ -304,6 +304,23 @@ func preReqs(ctx context.Context, cfg config) *cerrors.PrereqErr {
...
+			if strings.EqualFold(quotedIdentifiersIgnoreCase, "true") {
```
danielnelson, line 314: If we parse the value as a boolean we make sure it came back in the form we expected, just a little bit more defensive programming:
```suggestion
			ignoreCase, err := strconv.ParseBool(value)
			if err != nil {
				errs.Err(fmt.Errorf("unable to parse QUOTED_IDENTIFIERS_IGNORE_CASE value: %w", err))
			} else if ignoreCase {
```

materialize-snowflake/client.go
```diff
@@ -304,6 +304,23 @@ func preReqs(ctx context.Context, cfg config) *cerrors.PrereqErr {
...
+				errs.Err(fmt.Errorf(
+					"the Snowflake parameter QUOTED_IDENTIFIERS_IGNORE_CASE is enabled, which is incompatible with this connector: " +
+						"Estuary uses quoted identifiers for column names containing special characters (e.g. \"_meta/op\"), " +
+						"and requires Snowflake to preserve their case exactly as created. " +
+						"Disable this parameter for the Estuary user with: " +
+						"ALTER USER <estuary_user> SET QUOTED_IDENTIFIERS_IGNORE_CASE = FALSE",
+				))
```
danielnelson, line 321: This error is definitely on the long side, I would say its uncommon for an error to be longer than 80 chars.  In Go errors are often concatenated as well so individual errors are typically on the shorter side, I notice that this error goes off the right side of my screen in the webapp so I need to scroll horizontally.

I can see an argument for errors in the prerequisite checks being special.  In the future, I'd love if we returned additional metadata here that the UI could use to render the errors, including maybe a shortened link to relevant docs, but the interface doesn't allow for it currently.

So for now I'd just try to shrink this down as much as you feel you can.  I don't typically include why I think things happened or how to address them, I like to stick just to the facts about what happened.  ~~Maybe it's worth adding a section to the docs?~~ (edit: I see it)

jwhartley:
Hi @danielnelson I've accepted your suggestions, merged the doc update https://github.com/estuary/flow/pull/2853, and used a short link to make the error shorter here.
For this PR, changing the snowflake materialization is not something I'm game to do as a non-engineer, can you squash and merge if you're happy with it?

danielnelson (approved):

PR #4247, opened 2026-04-14
source-dynamics-365-finance-and-operations: emit sourced schemas

Alex-Bair:
**Description:**

When a connector relies solely on schema inference, the runtime can only learn a field's type from non-null values observed in captured documents. If a field has only ever been null in captured documents, the runtime has no type information for it, and materializations cannot create the corresponding column in the destination. This is especially problematic for sparse D365 entities where many fields may remain null for long periods. We've solved this problem before in other captures by emitting sourced schema messages - messages that inform the runtime about the source system's understanding of the schema, usually derived from source system provided metadata.

This PR crafts sourced schema messages from the table attribute metadata present in the top level `model.json`. This gives the Flow runtime upfront knowledge of all field types and constraints, so materialization columns can be created before any data arrives.

**Notes for reviewers:**

See this [PR](https://github.com/estuary/connectors/pull/2732) for some more context around how sourced schema messages are used by capture connectors to pre-widen inferred schemas. The use case described in that PR for `source-salesforce-native` is the same use case here for `source-dynamics-365-finance-and-operations`; pre-populating columns in destinations before any non-null values have been captured for those columns.

Additional source: [official docs](https://learn.microsoft.com/en-us/power-apps/maker/data-platform/export-data-lake-faq#what-date-and-time-formats-can-be-expected-in-exported-dataverse-tables) describing the format for different `datetime` and `datetimeOffset` fields

### DIFF
--- a/source-dynamics-365-finance-and-operations/source_dynamics_365_finance_and_operations/models.py
+++ b/source-dynamics-365-finance-and-operations/source_dynamics_365_finance_and_operations/models.py
@@ -1,5 +1,7 @@
 from dataclasses import dataclass
-from typing import ClassVar, Literal, Annotated
+from enum import StrEnum
+from logging import Logger
+from typing import Any, ClassVar, Literal, Annotated
 
 from estuary_cdk.capture.common import (
     ConnectorState as GenericConnectorState,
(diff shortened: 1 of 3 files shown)


nicolaslazo (approved):
LGTM

PR #4248, opened 2026-04-14
materialize-snowflake: use an atomic for bdec filename counter

danielnelson:
It seems possible that two files could get the same file count since this is the a shared counter.  While this hasn't been seen in practice it makes sense to prevent.

**Workflow steps:**

No changes

None

**Notes for reviewers:**

### DIFF
--- a/materialize-snowflake/stream.go
+++ b/materialize-snowflake/stream.go
@@ -6,6 +6,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mdibaiee (approved):

dyaffe:
Is there any risk here or migration needed?  Just want to be extra careful with any change to this since it's so sensitive.

danielnelson:
I would classify this update as very safe, and recommend merging so that this aspect is not in question on any future issues.

PR #4252, opened 2026-04-15
materialize-clickhouse: add column type migration support

### LINKED ISSUE #4246: materialize-clickhouse: add type migrations
The Clickhouse connector doesn't currently support type migration. For example, when a source column type changes from Int64 to Float64, the connector should automatically change the target column, [as described here](https://docs.estuary.dev/guides/schema-evolution/#auto-inference-only-widens).

Let's do this after https://github.com/estuary/connectors/pull/3284 is merged.

jacobmarble:
**Description:**

Add type migration to the Clickhouse materializer.
Closes #4246

**Workflow steps:**

We override 3 of the 5 standard steps.

**Notes for reviewers:**


materialize-clickhouse/sqlgen.go
```diff
@@ -103,6 +112,15 @@ var clickHouseDialect = func(database string) sql.Dialect {
 		Placeholderer:       sql.PlaceholderFn(func(index int) string { return "?" }),
 		TypeMapper:          mapper,
 		MaxColumnCharLength: 256,
+		MigratableTypes: sql.MigrationSpecs{
+			"int64":                {sql.NewMigrationSpec([]string{"Float64", "String"})},
+			"float64":              {sql.NewMigrationSpec([]string{"String"})},
+			"bool":                 {sql.NewMigrationSpec([]string{"String"})},
+			"date32":               {sql.NewMigrationSpec([]string{"String"})},
+			"datetime64(6, 'utc')": {sql.NewMigrationSpec([]string{"String"}, sql.WithCastSQL(datetimeToStringCast))},
+			"string":               {sql.NewMigrationSpec([]string{"Float64"})},
```
danielnelson: This means a string field can be updated to float64?  How does it convert a text value like foo?

materialize-clickhouse/sqlgen.go
```diff
@@ -103,6 +112,15 @@ var clickHouseDialect = func(database string) sql.Dialect {
 		Placeholderer:       sql.PlaceholderFn(func(index int) string { return "?" }),
 		TypeMapper:          mapper,
 		MaxColumnCharLength: 256,
+		MigratableTypes: sql.MigrationSpecs{
+			"int64":                {sql.NewMigrationSpec([]string{"Float64", "String"})},
```
danielnelson: How are integers larger than the float64 significand handled?

danielnelson (commented):
I haven't done much with migrations so just a few questions.

materialize-clickhouse/sqlgen.go
jacobmarble: This is how "hugeint" is converted to "number". Looking at other materializers, this is not how we normally do things. I'll iterate on this, probably use Clickhouse `Int256` type for integers larger than 64 bits.

materialize-clickhouse/sqlgen.go
jacobmarble: When an integer size exceeds 64 bits, the column is migrated to `String`.
When an integer becomes a float, the column is migrated to `Float64`.

materialize-clickhouse/client.go
```diff
@@ -130,6 +130,55 @@ func (c *client) DeleteTable(_ context.Context, path []string) (string, boilerpl
 	}, nil
 }
 
+// clickHouseMigrationSteps customizes the standard 5-step column type migration
+// for ClickHouse-specific SQL syntax:
+//   - Step 0: ADD COLUMN with Nullable() wrapper (ClickHouse is NOT NULL by default)
+//   - Step 1: ALTER TABLE UPDATE mutation syntax instead of standard UPDATE
+//   - Step 2: DROP COLUMN (standard)
+//   - Step 3: RENAME COLUMN (standard)
+//   - Step 4: No-op (ClickHouse manages nullability via Nullable() wrapper)
```
mdibaiee, line 139: Why is this a no-op? at this step we should revert `Nullable()` on the new column if it is supposed to be NOT NULL.

Is it possible in Clickhouse to change the type of a column from `Nullable(T)` to `T`?

materialize-clickhouse/sqlgen.go
jacobmarble: Values are converted by the database, so int64 => float64 cast/conversion is as defined in ClickHouse. TL;DR: the result is quietly lossy.

```sql
ALTER TABLE t
UPDATE col_name_flowtmp1 = IF(col_name IS NULL, NULL, CAST(col_name AS Int64)
```

I've tested this empirically:
```sql
create table t (a Int64, b Float64) engine = MergeTree() order by a;
insert into t values (36028797018963968, 9.9);
alter table t update b = CAST(a AS Float64) WHERE true;
select * from t;
```

The result is lossy:
```text
   ┌─────────────────a─┬─────────────────b─┐
1. │ 36028797018963968 │ 36028797018963970 │ -- 36.03 quadrillion
   └───────────────────┴───────────────────┘
```

ClickHouse provides a special function [`accurateCast`](https://clickhouse.com/docs/sql-reference/functions/type-conversion-functions#accurateCast), which yields the same result.

---

However, this happens so rarely that it isn't accounted for in other materializers -- here's why.

`FlatTypeMapping` for `sql.INTEGER` yields ClickHouse type `Int64` or `Int128`, and `sql.STRING_INTEGER` yields `Int128` or `String`. The only integer-to-float trigger the Collection column type changing from `FlatType` `sql.INTEGER` to `sql.NUMBER`.

To be clear, ClickHouse doesn't have a wider float type, and other materializers don't accommodate with `VARCHAR` or similar, so this is adequate IMO.

materialize-clickhouse/sqlgen.go
jacobmarble: I've improved this in a new PR https://github.com/estuary/connectors/pull/4254 which this PR is rebased on. Integers wider than 64 bits are stored as `Int128` which is the same width as most other SQL targets' `HUGEINT` and similar.

materialize-clickhouse/client.go
jacobmarble: This is a no-op because ClickHouse allows creating non-nullable columns without populating them. Comments are rewritten to help the reader.

Related to this, the logic in step 0 was flawed. Columns are now created as `Nullable()` or not, so the nullability attribute normally modified at step 4 is now correctly established in step 0.

ClickHouse has a very handy [`ALTER TABLE ... ALTER COLUMN ... TYPE`](https://clickhouse.com/docs/sql-reference/statements/alter/column#modify-column) statement that converts column types in a single statement. It does work for all but the cast from `DateTime` to `String`; shimming in that exception was very ugly, so I decided to keep the current code instead.

materialize-clickhouse/sqlgen.go
```diff
@@ -251,23 +260,35 @@ var clickHouseDialect = func(database string) sql.Dialect {
 		Placeholderer:       sql.PlaceholderFn(func(index int) string { return "?" }),
 		TypeMapper:          mapper,
 		MaxColumnCharLength: 256,
+		MigratableTypes: sql.MigrationSpecs{
+			"int64":                {sql.NewMigrationSpec([]string{"Float64", "Int128", "Int256", "String"})},
```
mdibaiee: can we use all lowercase for both sides of these please? that's the convention. We use `EqualFold` and `ToLower` in the migration mapping so we can always compare lowercases

materialize-clickhouse/sqlgen.go
```diff
@@ -343,13 +364,38 @@ SETTINGS
 
 {{ define "alterTargetColumns" }}
 {{- range $ind, $col := $.AddColumns }}
-ALTER TABLE {{$.Identifier}} ADD COLUMN IF NOT EXISTS {{$col.Identifier}} Nullable({{$col.NullableDDL}});
+ALTER TABLE {{$.Identifier}} ADD COLUMN {{$col.Identifier}} {{ if not $col.MustExist }}Nullable({{ end }}{{$col.NullableDDL}}{{ if not $col.MustExist }}){{ end }};
```
mdibaiee: Can we directly use `{{ $col.DDL }}` which should include nullability in it conditionally?

materialize-clickhouse/sqlgen.go
```diff
@@ -343,13 +364,38 @@ SETTINGS
...
+
+-- ClickHouse has a special syntax for this operation, which normally
+-- runs as a background process like merges. The mutations_sync=1
+-- setting causes statement execution to block until the process is
+-- complete. The WHERE true clause is required syntax.
+
+{{ define "alterTargetMigratePopulateColumn" }}
+{{- range $ins := $.Instructions }}
+ALTER TABLE {{$.Table.Identifier}}
+UPDATE {{ $ins.TempColumnIdentifier }} = {{ if not $ins.TypeMigration.MustExist }}if(isNull({{ $ins.TypeMigration.Identifier }}), NULL, {{ end }}{{ call $ins.TypeMigration.CastSQL $ins.TypeMigration }}{{ if not $ins.TypeMigration.MustExist }}){{ end }}
```
mdibaiee: Not sure if this works or not, just occurred to me that it might:

Can we do `CAST($identifier, Nullable($TypeMigration))` so that there is no need to check for nullability before casting?

materialize-clickhouse/driver_clickhouse_test.go
```diff
@@ -1052,3 +1053,122 @@ func TestMovePartitionMissingTarget(t *testing.T) {
 	require.ErrorAs(t, moveErr, &exc)
 	require.EqualValues(t, chproto.ErrUnknownTable, exc.Code)
 }
+
+// TestMigrationInt256Paths exercises column type migrations involving Int256,
+// which aren't reached by the shared RunMigrationTest fixture: that fixture's
+// widest integer has maximum 2^64-1, mapping to Int128.
```
mdibaiee: Can we add this case to RunMigrationTest instead?

materialize-clickhouse/sqlgen.go
mdibaiee: There are `WithNullableText` and `WithNotNullText` in `type_mapping.go` that should be used to make nullability added automatically to column DDLs depending on whether they are nullable or not...

I think the current way it is defined is not really compatible with Clickhouse's paranethesis syntax, but it could be made compatible I guess.

Up to you if you think it is worth refactoring that part to allow Clickhouse to delegate setting of `Nullable(T)` in materialize-sql, or you want to keep it the way it is now.

The templates for Clickhouse are just very verbose right now because of this

materialize-clickhouse/driver_clickhouse_test.go
jacobmarble: Opened https://github.com/estuary/connectors/pull/4295 for this

materialize-clickhouse/sqlgen.go
jacobmarble: Opened https://github.com/estuary/connectors/pull/4294 for this

materialize-clickhouse/sqlgen.go
jacobmarble: That does not work. We can remove the condition and handle everything through `if(foo IS NULL, NULL, CAST(...))` without side effect.

jacobmarble:
This PR is rebase to include these:
- https://github.com/estuary/connectors/pull/4307
- https://github.com/estuary/connectors/pull/4295

mdibaiee (approved):

PR #4254, opened 2026-04-15
materialize-clickhouse: handle wide integers properly

jacobmarble:
**Description:**

If possible, ClickHouse should follow the pattern in most SQL materializers:
- Store integers wider than 64 bits as actual integers, if possible.
- Store string integers as integers, if possible.
- Fallback on String after that.

~ClickHouse has `Int128`, which matches the largest integer size in most other SQL materializers. ClickHouse also has `Int256`, but I've not used that, as no other SQL materializer uses such a wide type.~

This change steps integer types through `Int64`, `Int128`, `Int256`, and finally `String`, as required by the Collection schema.

**Workflow steps:**

No changes to customer workflow.

**Notes for reviewers:**

If any customer had been storing integers as string (`sql.STRING_INTEGER`) then we would need to bump the version of this connector. Because only one customer is using materialize-clickhouse today, I've audited their schema and found no such columns, so we're safe to merge without a new version.


materialize-clickhouse/main_test.go
```diff
@@ -13,10 +13,13 @@ func TestMain(m *testing.M) {
 		if err := exec.Command("docker", "compose", "-f", "docker-compose.yaml", "up", "--wait").Run(); err != nil {
 			panic("starting docker compose: " + err.Error())
 		}
-		defer exec.Command("docker", "compose", "-f", "docker-compose.yaml", "down", "-v").Run()
 	}
 
-	os.Exit(m.Run())
+	exitCode := m.Run()
+	if !isShortMode() {
+		_ = exec.Command("docker", "compose", "-f", "docker-compose.yaml", "down", "-v").Run()
```
danielnelson, line 20: Why no more defer?  Could be nice if the test will shutdown the container on a panic.  I would love a check of error and a log message or panic.

Side note, I haven't tested this but I think `testing.Short()` would work if you call `flag.Parse()` first.

danielnelson (approved):
As this changes some mapped type, I imagine it could require a backfill to updated to.  This materialization is probably new enough that this isn't a concern, but something to start thinking about.

materialize-clickhouse/main_test.go
jacobmarble: This was necessary because `os.Exit()` doesn't allow deferred tasks to execute, so the docker compose wouldn't clean up.

We could use a better pattern, but this is the only `TestMain` function that needs to `defer`.

materialize-clickhouse/sqlgen.go
```diff
@@ -48,21 +50,59 @@ var clickHouseClampDatetime = sql.StringCastConverter(func(str string) (interfac
...
-			sql.MULTIPLE:       sql.MapStatic("String", sql.UsingConverter(sql.ToJsonString)),
-			sql.STRING_INTEGER: sql.MapStatic("String", sql.UsingConverter(sql.ToStr)),
-			sql.STRING_NUMBER:  sql.MapStatic("Float64", sql.UsingConverter(sql.StrToFloat(math.NaN(), math.Inf(1), math.Inf(-1)))),
+			sql.NUMBER:   sql.MapStatic("Float64"),
+			sql.BOOLEAN:  sql.MapStatic("Bool"),
+			sql.OBJECT:   sql.MapStatic("String", sql.UsingConverter(sql.ToJsonString)),
+			sql.ARRAY:    sql.MapStatic("String", sql.UsingConverter(sql.ToJsonString)),
+			sql.BINARY:   sql.MapStatic("String"),
+			sql.MULTIPLE: sql.MapStatic("String", sql.UsingConverter(sql.ToJsonString)),
+			sql.STRING_INTEGER: sql.MapStringMaxLen(
```
mdibaiee: This is not specific to Clickhouse and I think it affects all connectors, but I just realised, if there are negative numbers in `type: string, format: integer`, we will end up converting to string _earlier_, because there is 1 extra character for the minus sign.

I don't think we need to address it for now, but something to keep in mind. @danielnelson

materialize-clickhouse/sqlgen.go
```diff
@@ -48,21 +50,59 @@ var clickHouseClampDatetime = sql.StringCastConverter(func(str string) (interfac
...
+				sql.MapStatic("Int128", sql.UsingConverter(toBigInt)),
+				sql.MapStatic("String", sql.UsingConverter(sql.ToStr)),
```
mdibaiee: Is there a reason not to use Clickhouse's Int256?

jacobmarble: The other materializers' destination integer data types stop at 64 bits (names such as `long` and `bigint`) or 128 bits (names like `hugeint`). After that, materializers promote to fixed-point (`numeric(38,0)` or `decimal(38,0)`) and string types (`varchar`, `text`, etc).

Yesterday this pattern seemed clear, so I selected the progression `Int64` => `Int128` => `String`. However, in order to write this comment, I've audited with fresh eyes, and [built a spreadsheet](https://docs.google.com/spreadsheets/d/1aCVOWhBMqcvy7Vy7CnaAtIrd3X8ROfmxgGzz10cvcZc/edit?usp=sharing).

My conclusion is that better options are `Int256` (256 integer bits / ~76 decimal digits) or `Decimal256(0)` (76 decimal digits / ~256 integer bits).

jacobmarble: 7bc900a01

mdibaiee: Is it possible to do step-by-step use of types? That is, use Int128, then if maximum is higher, use 255, then if that's not enough, use Decimal

The reason I ask is we have had customers ask us about table size and if a customer is only storing numbers that fit under 128bit, they may complain about us making tables that use 256 bits for all of their numeric columns (assume the worst case scenario: a huge table with tons of numbers)

I believe we can stack a few MaxStrLen's on top of each other for this

materialize-sql/type_mapping.go
```diff
@@ -18,6 +18,21 @@ import (
...
+//
+// Guidelines for integer conversion:
+// - INTEGER:
+//   - Signed 64-bit integer, such as BIGINT
+//
+// - STRING_INTEGER:
+//   - A string representation, such as VARCHAR or TEXT
+//
+// - Wide INTEGER; more than 64 bits:
+//   - The widest available signed integer type, at least 128-bits wide, such as HUGEINT
```
mdibaiee: I think this is confusing to distinguish between `INTEGER` and `Wide INTEGER` here: they are the same type as it concerns `FlatType`

I think _narrow_ and _wide_ are arbitrary here and are really dependent on the connector

materialize-clickhouse/sqlgen.go
jacobmarble: FYI the [Clickhouse `Decimal` type](https://clickhouse.com/docs/sql-reference/data-types/decimal) is limited to 76 digits - same as a 256-bit integer.

materialize-sql/type_mapping.go
jacobmarble: Yes, I see what you mean. I'll just revert this, maybe later we'll think of a better framework. 24e479618

materialize-clickhouse/sqlgen.go
jacobmarble: This is done as 0176c836d. I'm updating `materialize-sql/testdata/flow.yaml` to add more steps to the test battery used in other places, will add that commit when it's ready.

jacobmarble: Additional integer width steps tested across SQL materializers in 93c8c7b09. Note the commit makes only additive changes to snapshots, and makes no code changes.

materialize-motherduck/.snapshots/TestSQLGeneration
```diff
@@ -10,6 +10,8 @@ CREATE TABLE IF NOT EXISTS db."a-schema".key_value (
 	field_with_projection VARCHAR,
 	flow_published_at TIMESTAMP WITH TIME ZONE NOT NULL,
 	"integer" BIGINT,
+	"integerGt128Bit" HUGEINT,
+	"integerGt256Bit" HUGEINT,
```
jacobmarble, line 14: This illustrates that some materializers may try to store values wider than the projected type allows. A problem for another day.

materialize-clickhouse/sqlgen.go
```diff
@@ -48,20 +50,117 @@ var clickHouseClampDatetime = sql.StringCastConverter(func(str string) (interfac
...
+	return func(p *sql.Projection) (sql.DDLer, sql.CompatibleColumnTypes, sql.ElementConverter) {
+		// STRING_INTEGER projections (format: "integer") route by declared
+		// maxLength — values arrive as JSON strings and can exceed the
+		// precision that Numeric.Minimum/Maximum can carry in float64.
+		// MaxLength == 0 means no length was declared; fall back to the
+		// widest bounded integer rather than the narrowest.
+		if p.Inference.String_ != nil && p.Inference.String_.Format == "integer" {
+			switch maxLen := p.Inference.String_.MaxLength; {
+			case maxLen == 0:
+				return fn256(p)
```
jacobmarble, line 117: I'm uncertain about this - safer to use `fnString`?

materialize-clickhouse/sqlgen.go
```diff
@@ -48,20 +50,117 @@ var clickHouseClampDatetime = sql.StringCastConverter(func(str string) (interfac
...
+
+		// INTEGER projections dispatch by the inferred numeric range.
+		// Numeric Minimum and Maximums are stated as powers of 10, so
+		// comparisons to exact integer values aren't precise, but this
+		// logic works out since 1e19 is the next power of 10 past int64
+		// max, 1e39 past int128, and 1e77 past int256.
+		if p.Inference.Numeric != nil {
+			switch {
+			case p.Inference.Numeric.Minimum >= math.MinInt64 && p.Inference.Numeric.Maximum <= math.MaxInt64:
+				return fn64(p)
```
jacobmarble, line 137: Similar concern here - if min/max values are zero, should we use `Int64` or `String`?

materialize-clickhouse/sqlgen.go
mdibaiee: I think we go with the basic type here, so `fn64`

materialize-clickhouse/sqlgen.go
mdibaiee: I think int64

materialize-clickhouse/sqlgen.go
jacobmarble: Cool. This is consistent with your recent comment in Slack. 👍

mdibaiee (approved):
Thank you!

PR #4257, opened 2026-04-16
materializations: benchmark suite and gcp setup

mdibaiee:
**Description:**

(Can be reviewed commit-by-commit)

- Add a new set of tools for running benchmarks of materialization connectors. This tool supports both local (running connector on the machine) and docker runs, and there are _scenarios_ defined which are common benchmarks we may want to run on our connectors.
To run a benchmark:

```
./tests/benchmark/materialize/run.sh \
          --connector materialize-postgres \
          --scenario tests/benchmark/materialize/scenarios/large-docs.yaml
```

The scenarios are defined in YAML in a way that allows you to specify the size of transactions by combining document count and document size. It also allows specifying "overlaps", so that a transaction may include updates or deletions of documents from a previous transaction, so we can test insertions as well as merges.

- There is also another tool for provisioning GCP VMs suitable for running benchmarks on servers to avoid internet instability / latency from affecting benchmarks. To set up a vm, simply:
```
./tests/benchmark/gcp-vm.sh create
./tests/benchmark/gcp-vm.sh ssh
```

Then instructions will be provided for running the benchmarks on this machine (same as running on host). There is also a warning for stale machines that have been created more than a week ago whenever anyone runs any command on this CLI to help us clean up.

The clean-up is easy:
```
./tests/benchmark/gcp-vm.sh destroy
```

To iterate on benchmarks is also done by syncing your files to the server, so you can run tests again after making changes:
```
./tests/benchmark/gcp-vm.sh sync
```

Each run results in a JSON file under `tests/benchmark/materialize/runs` which includes metrics about the run, so different runs can be compared easily.

Those runs can also be pulled down to your local machine using:
```
./tests/benchmark/gcp-vm.sh pull-runs
```

I have stored some baseline benchmarks under runs/baselines for each connector which can serve as comparison points for us.

**Workflow steps:**

**Notes for reviewers:**


tests/benchmark/materialize/runs/baselines/.gitkeep
danielnelson, line 1: Remove since you have added the baseline files.

tests/benchmark/materialize/generate.py
```diff
@@ -0,0 +1,582 @@
...
+
+_SUFFIXES = {"B": 1, "KB": 1024, "MB": 1024**2, "GB": 1024**3, "TB": 1024**4}
+
+
+def parse_size(value: Any) -> int:
+    """Parse a size value: int (bytes) or string with B/KB/MB/GB/TB suffix."""
+    if isinstance(value, int):
+        return value
+    if not isinstance(value, str):
+        raise ValueError(f"size must be int or string, got {type(value).__name__}")
```
danielnelson: I like to put the acceptable types in the type hint and only handle those types at runtime.  If another type is passed then not my problem and I just let it crash.

```suggestion
def parse_size(value: int | str) -> int:
    """Parse a size value: int (bytes) or string with B/KB/MB/GB/TB suffix."""
    if isinstance(value, int):
        return value
```

tests/benchmark/materialize/generate.py
```diff
@@ -0,0 +1,582 @@
...
+    key_field: str = "id"
+    payload_field: str = "payload"
+    key_type: str = "integer"  # "integer" or "uuid"
+
+
+@dataclass
+class OverlapSpec:
+    with_tx: int
+    fraction: float
+    op: str  # "u" or "d"
```
danielnelson:
```suggestion
    op: Literal["u", "d"]
```

tests/benchmark/materialize/generate.py
```diff
@@ -0,0 +1,582 @@
...
+    p.add_argument("--output", default="-", help="Output path (default: stdout)")
+    p.add_argument("--state", default=None, help="Optional path to write state JSON")
+    args = p.parse_args(argv)
+
+    collections, transactions = load_scenario(args.scenario)
+
+    if args.output == "-":
+        state = emit(collections, transactions, sys.stdout, seed=args.seed)
+    else:
+        with open(args.output, "w", buffering=128 * 1024) as f:
```
danielnelson: This is probably the default buffering value?

jacobmarble (approved):
I'm excited to try this out!

tests/benchmark/materialize/generate_spec.py
```diff
@@ -0,0 +1,220 @@
...
+
+  --docker    use a Docker image instead of a local binary
+  --image     full image ref (overrides default ghcr.io/estuary/<connector>:<tag>)
+  --tag       image tag (default: $VERSION env or "local")
+
+The scenario's `collections[].resource` (if present) is passed through to the
+binding; otherwise we derive `{table: <basename>}` from the collection name.
+"""
+
+from __future__ import annotations
```
danielnelson: This import is deprecated, there is a new method for deferred annotations in 3.14 which I don't have experience with but will be the new way forward.

tests/benchmark/materialize/generate.py
```diff
@@ -0,0 +1,582 @@
...
+# Knuth-style multiplier used to scatter per-doc pool offsets.
+_POOL_MIX = 2654435761
+# Fixed namespace for deterministic UUID generation from integer keys.
+_UUID_NAMESPACE = uuid.UUID("a3f2b8c1-7d4e-4f9a-b6c8-1e2d3f4a5b6c")
+
+
+def _int_key_to_uuid(key: int) -> str:
+    """Deterministic UUID from an integer key. Uses UUID v5 (SHA-1 based)
+    so the same key always produces the same UUID across runs."""
+    return str(uuid.uuid5(_UUID_NAMESPACE, struct.pack("<q", key)))
```
danielnelson: I wouldn't use struct for this, if you really want it to be 8 bytes little endian you could use `key.to_bytes(8, byteorder='little')`, but here I don't think it really matters it just needs to be bytes:
```suggestion
    return str(uuid.uuid5(_UUID_NAMESPACE, key.to_bytes())
```

tests/benchmark/materialize/generate.py
```diff
@@ -0,0 +1,582 @@
...
+    return plans, fresh_start, fresh_count
+
+
+def emit(
+    collections: dict[str, CollectionSpec],
+    transactions: list[TxSpec],
+    out,
+    seed: int = 0,
+) -> dict:
+    """Write the fixture to `out` (a file-like in text mode). Returns a state dict."""
```
danielnelson: Instead of the comment for `out` type hint you could use `typing.TextIO`.  Could almost use `io.Writer` but I don't think you can call flush on it, but you could flush in the caller or just leave it off because the file is closed as soon as this function ends?

danielnelson (approved):
Looks good, all comments optional.

PR #4258, opened 2026-04-16
materialize-snowflake: add retry limit for token persistance

danielnelson:
Currently we wait forever for the token to be persisted potentially causing the connector to be stuck.  With this change after 5m we exit with an error, when the connector restarts it will reopen the channel which will fence any ongoing changes from other clients.

Additionally, when a temporary error occurred we would retry it but there was no logging that it occurred, which made it impossible to know if errors were the cause of retries or if the token was not updated.  related: https://github.com/estuary/connectors/pull/3959

Increased the max retry backoff to 2s, in case we are polling to quickly.

**Workflow steps:**

No changes

None

**Notes for reviewers:**


williamhbaker (approved):
LGTM

PR #4263, opened 2026-04-17
materialize-mysql: implement 'Additional Table Create SQL'

### LINKED ISSUE #4028: Add "additional table creation sql" to the MySQL connector
Similar to how we do it for postgres

jacobmarble:
**Description:**

Closes https://github.com/estuary/connectors/issues/4028

Implement the "Additional Table Create SQL" materialize-mysql feature.

The feature is slightly different to the materialize-postgres equivalent - MySQL cannot perform SQL statements after `CREATE TABLE` in one transaction, hence "Additional SQL statement(s) to be run after the table is created."

**Workflow steps:**

Docs updated at https://github.com/estuary/flow/pull/2869

**Notes for reviewers:**

### DIFF
--- a/materialize-mysql/.snapshots/TestConfigURI-IncorrectSSL
+++ b/materialize-mysql/.snapshots/TestConfigURI-IncorrectSSL
@@ -1,2 +1,2 @@
-will:secret1234@tcp(example.com:3306)/somedb?clientFoundRows=true&connectionAttributes=program_name%3AEstuary+materialize-mysql
+will:secret1234@tcp(example.com:3306)/somedb?clientFoundRows=true&connectionAttributes=program_name%3AEstuary+materialize-mysql&multiStatements=true
 invalid 'sslmode' configuration: unknown setting "whoops-this-isnt-right"
--- a/materialize-mysql/client.go
+++ b/materialize-mysql/client.go
@@ -15,6 +15,7 @@ import (
 	pf "github.com/estuary/flow/go/protocols/flow"
 	pm "github.com/estuary/flow/go/protocols/materialize"
 	"github.com/go-sql-driver/mysql"
+	log "github.com/sirupsen/logrus"
 )
 
 type client struct {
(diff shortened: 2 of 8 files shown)


materialize-mysql/.snapshots/TestConfigURI-IncorrectSSL
```diff
@@ -1,2 +1,2 @@
-will:secret1234@tcp(example.com:3306)/somedb?clientFoundRows=true&connectionAttributes=program_name%3AEstuary+materialize-mysql
+will:secret1234@tcp(example.com:3306)/somedb?clientFoundRows=true&connectionAttributes=program_name%3AEstuary+materialize-mysql&multiStatements=true
```
jacobmarble, line 1: FYI `multiStatements=true` allows users to provide multiple statements to the additional SQL value. Postgres allows multiple statements by default, MySQL requires it to be explicitly enabled.

materialize-mysql/client.go
```diff
@@ -221,8 +222,27 @@ func (c *client) AlterTable(ctx context.Context, ta sql.TableAlter) (string, boi
...
 
 func (c *client) CreateTable(ctx context.Context, tc sql.TableCreate) error {
-	_, err := c.db.ExecContext(ctx, tc.TableCreateSql)
-	return err
+	var res tableConfig
+	if tc.Resource != nil {
+		res = tc.Resource.(tableConfig)
+	}
+
+	if _, err := c.db.ExecContext(ctx, tc.TableCreateSql); err != nil {
```
danielnelson, line 234: Can these statements be in a transaction so all tables are created atomically?

danielnelson (approved):

materialize-mysql/client.go
mdibaiee: Jacob mentions in the PR description that MySQL does not support that apparently. It seems it does support certain operations being done atomically across tables (e.g. `DROP TABLE t1, t2, t3`), but I don't see it supporting multiple separate DDL statements in the same transaction. https://dev.mysql.com/doc/refman/8.4/en/atomic-ddl.html

mdibaiee (approved):

materialize-mysql/client.go
jacobmarble: Not really. I've added a transaction wrapping the additional SQL statements, and a comment to clarify. 104664a7630382a1584a7e9b5ff521d63f2d612c

PR #4266, opened 2026-04-19
estuary-cdk: support connector-specific dockerfiles

nicolaslazo:
**Description:**

Builders will first search for a dockerfile at each connector's project root before defaulting to `common.Dockerfile`. Developers wishing to add support for webhook captures may symlink to the included `webhook-capture.Dockerfile` file to expose port 8080.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/.claude/skills/create-webhook-connector/SKILL.md
+++ b/.claude/skills/create-webhook-connector/SKILL.md
@@ -108,3 +108,14 @@ async def all_resources(log, http, config):
 Ask the user which discriminator strategy fits their webhook provider and whether event types can be discovered via API or need to be statically defined.
 
 If the provider also has pull API endpoints (not just webhooks), use `/classify-stream-types` to determine the appropriate stream type for each endpoint.
+
+## Dockerfile setup
+
+Webhook connectors need a `Dockerfile` symlink in their connector directory pointing to the shared webhook Dockerfile:
+
+```bash
+cd source-<name>
+ln -s ../estuary-cdk/webhook-capture.Dockerfile Dockerfile
+```
+
+CI and `build-local.sh` automatically detect connector-specific Dockerfiles
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -416,21 +416,20 @@ jobs:
           cd ..
           pytest ${{ matrix.connector.name }}
 
-      # Uncomment to allow each connector to have its own Dockerfile
-      #- name: Find Dockerfile for ${{ matrix.connector.name }}
-      #  id: dockerfile
-      #  run: |
-      #    DOCKERFILE_PATH="${PWD}/estuary-cdk/common.Dockerfile"
-      #    if [ -f "${{ matrix.connector.name }}/Dockerfile" ]; then
-      #      DOCKERFILE_PATH="${PWD}/${{ matrix.connector.name }}/Dockerfile"
-      #    fi
-      #    echo "path=${DOCKERFILE_PATH}" >> $GITHUB_OUTPUT
+      - name: Find Dockerfile for ${{ matrix.connector.name }}
+        id: dockerfile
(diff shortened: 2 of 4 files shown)


.github/workflows/python.yaml
```diff
@@ -430,7 +430,7 @@ jobs:
         uses: docker/build-push-action@v6
         with:
           context: .
-          file: estuary-cdk/common.Dockerfile
+          file: ${{ matrix.connector.type == 'webhook-capture' && 'estuary-cdk/webhook-capture.Dockerfile' || 'estuary-cdk/common.Dockerfile' }}
```
nicolaslazo: This wouldn't scale well as more types get added and I found a commented out snippet of code indicating we've already considered adding support for connector-specific Dockerfile paths. However, these dockerfiles are largely static and this approach keeps changes as simple as possible.

Alex-Bair (commented):
Unfortunately, I don't think connectors built with a `webhook-capture` type will work. Connector images' types must be one of the `RuntimeProtocol` members defined [here ](https://github.com/estuary/flow/blob/cfeef660ca5d297ae1ffa968170fdde096252242/crates/runtime/src/lib.rs#L51-L55)in the runtime. If the connector type is something other than `"capture"`, `"derive"`, or `"materialize"`, the runtime will [error out](https://github.com/estuary/flow/blob/cfeef660ca5d297ae1ffa968170fdde096252242/crates/runtime/src/lib.rs#L58-L65) and refuse to run the image. So, we can't use a different type to differentiate between captures that need to expose a port & captures that don't. We _could_ use another build argument in the GitHub Action to differentiate between the two, but then how does the `build-local.sh` script know when to expose a port? It'd have to read the workflow file or we'd have to tell `build-local.sh` to expose a port when running the script.

I think until we have more time to investigate the best way to add `EXPOSE 8080` to the `common.Dockerfile`, I suggest requiring connectors to use their own `Dockerfile` when they need to expose a port. We can update the GitHub Actions workflow and `build-local.sh` script to prefer a local `Dockerfile` over any shared one, and that would enable the current `source-appsflyer` use case. There will be some duplication between `Dockerfile`s in the short term, but I'm optimistic we'll be able to find a more extensible solution once another connector comes along that needs to expose a port.

nicolaslazo:
I wasn't aware of that runtime constraint, good catch.

I followed your advice with a small experiment of my own, I tweaked our scripts to support resolving real filepaths and then symlinked to `../estuary_cdk/webhook-capture.Dockerfile` from a connector's project root. Seems to work fine, and should be less redundant than copy and pasting the same Dockerfile repeatedly. 

Are there any circumstances where symlinks wouldn't work?

.claude/skills/create-webhook-connector/SKILL.md
```diff
@@ -108,3 +108,14 @@ async def all_resources(log, http, config):
...
+## Dockerfile setup
+
+Webhook connectors need a `Dockerfile` symlink in their connector directory pointing to the shared webhook Dockerfile:
+
+```bash
+cd source-<name>
+ln -s ../estuary-cdk/webhook-capture.Dockerfile Dockerfile
+```
+
+CI and `build-local.sh` automatically detect connector-specific Dockerfiles — no matrix `type` field is needed.
```
Alex-Bair: nit: We don't need to mention the other approach we tried here.
```suggestion
CI and `build-local.sh` automatically detect connector-specific Dockerfiles.
```

Alex-Bair (approved):
LGTM! I think the symlinks will work. I tested out symlinking to `webhook-capture.Dockerfile` locally, and it worked for me as well. The connector images built in CI for this change also don't have `"8080/tcp": {}` in their `docker inspect` output, so the behavior for Python connectors that don't symlink to another `Dockerfile` remains the same. 👍

PR #4268, opened 2026-04-20
materialize-s3-iceberg: add alternative naming feature flag

### LINKED ISSUE #4136: Add nested_dot_hash_location_style formatting to materialize-s3-iceberg (like parquet)

danielnelson:
Similar to the flag added to the materialize-iceberg connector, this flag changes the table data paths in object storage to:
```
  <base_location>/<namespace>/<table>.<hash>
```

I brought the logic for table location naming into its own file for both iceberg connectors to better be able to compare them.

There are some notable differences between them, unlike materialize-iceberg this connector allows hyphens in namespace and table names and it always uses a nested layout with a directory for the namespace.

closes: #4136

**Workflow steps:**

To enable set the `nested_dot_hash_location_style` feature flag.

None planned

**Notes for reviewers:**

### DIFF
--- a/materialize-iceberg/config.go
+++ b/materialize-iceberg/config.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
-	"regexp"
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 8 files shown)


mdibaiee (commented):
it would be nice if we could have unit tests for the location generation so it can be compared before and after the change, and the two styles are in snapshots so we don't break them in the future

danielnelson:
I added some unit tests, but I haven't changed any snapshots.  We could add some additional configurations for TestIntegration with the flag set, but have some reservations because I don't want to blow out test times trying to hit all options in integration tests.  Any thoughts on how to best handle this?

mdibaiee:
@danielnelson I think unit tests are fine, I meant unit test snapshots, not necessarily integration test snapshots 👍🏽 

I personally would run integration tests with the new feature flag locally just to confirm it works fine and the snapshot updates for integration tests are as expected, but no need to commit it.

mdibaiee (approved):

jacobmarble (approved):

PR #4282, opened 2026-04-21
source-hubspot-native: coerce empty-string properties to `None`

### LINKED ISSUE #4272: source-hubspot-native: cleared properties don't propagate as null to destinations
### Summary

When a property is cleared in HubSpot, the connector strips it from the captured document, causing `MERGE` reduction to preserve the stale value. The clearance never reaches the destination.

### Root cause

`source_hubspot_native/models.py:319` removes all empty-string property values:

```python
self.properties = {k: v for k, v in self.properties.items() if v != ""}
```

HubSpot's API returns `""` for cleared properties. With the key absent from the new document, `MERGE` reduction on `properties` sees the key only in the LHS (older doc), so the old value is preserved. Downstream materializations never observe the clearance.

### Reproduction

1. Clear a property (e.g. `closedate`) on a HubSpot record
2. Wait for the capture to pick up the update
3. Observe: destination table still shows the old value
4. Check the raw document: `jq '.properties | has("closedate")'` returns `false`

### Proposed fix

Convert empty strings to `None` instead of stripping the key:

```python
self.properties = {k: (None if v == "" else v) for k, v in self.properties.items()}
```

Apply the equivalent treatment to `propertiesWithHistory`.

### Why this is safe

- The Pydantic field is already typed `dict[str, str | None]` (line 283), so null is a first-class value in the schema — no schema migration needed
- `null` doesn't interfere with format inference the way `""` does. Format inference examines non-null string values to derive `date-time` / `date` / etc., and null values are ignored
- `False`, `0`, and `[]` (preserved by #3520) continue to pass through untouched
- `MERGE` + `lastWriteWins` (the default per-property strategy) will propagate explicit nulls to the destination

### Prior context

- #1665 (original empty-string stripping — schema-inference motivation)
- #3520 (Nov 2025 — narrowed from "all falsy" to "only empty strings" to fix calculated-property refresh)
- The older Airbyte-based `source-hubspot` connector already does the equivalent conversion in `streams.py:553-556` — but only when a format is declared. Since `source-hubspot-native`'s schema declares `str | None` uniformly, we can safely apply the conversion unconditionally.

### Workaround

A capture backfill temporarily fixes stuck rows (re-reads current HubSpot state; with no prior value to merge against, the destination row ends up NULL). The bug recurs on the next clearance event.

### Semantic bonus

Calculated properties that legitimately return `""` (e.g. `hs_sequences_is_enrolled` when the formula can't compute) would be represented as null in collections going forward, which better conveys "present but unknown" vs. "absent from this partial document."

Alex-Bair:
**Description:**

HubSpot returns `""` for cleared properties and for formatted fields without values. The previous strip-filter removed those keys entirely, which combined with the merge reduction strategy (commit 0bd589b40) meant clearances never propagated to destinations; with a merge reduction strategy, an absent key preserves the LHS value.

Coercing `""` to `None` keeps the key present while leaving format inference intact (`null` values are ignored by format inference) and lets the reduced document read by materializations overwrite stale destination values with `null`.

The capture snapshot changes are expected. The newly capture properties have `""` values when we receive them from the API and were previously filtered out by the connector; with the new "coerce `""` values to `None`" behavior, these properties are now included in documents with `null`s instead.

Fixes #4272.

**Notes for reviewers:**

Here's the originating [Slack thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1776745707813859) for some context.


nicolaslazo (approved):
LGTM!

PR #4291, opened 2026-04-22
source-oracle: fix oracle 11g compatibility issues

nicolaslazo:
**Description:**

This PR replaces the three features implemented in Oracle 12+ with 11g-friendly alternatives:

- CDB checks are only queried if the source's major version >= 12
- `FETCH NEXT` clauses are replaced with `ROWNUM` subqueries
- Uses `NLSSORT` in place of `COLLATE BINARY` for sorting string-like columns

**Workflow steps:**

**Notes for reviewers:**

The docker image has been replaced with an 11g alternative locally to reproduce the reported exceptions. After tests passed, the original test suite was restored to ensure no regressions were present.

### DIFF
--- a/source-oracle/backfill.go
+++ b/source-oracle/backfill.go
@@ -324,15 +324,16 @@ func (db *oracleDatabase) fetchBackfillRowIDRange(ctx context.Context, schemaNam
 	return nil
 }
 
-// The set of column types for which we need to specify `COLLATE BINARY` to get
+// The set of column types for which we need to use `NLSSORT` with binary sorting to get
 // proper ordering and comparison. Represented as a map[string]bool so that it can be
 // combined with the "is the column typename a string" check into one if statement.
-// See https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/COLLATE-Operator.html
+// See https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/NLSSORT.html
 var columnBinaryKeyComparison = map[string]bool{
-	"varchar2": true,
 	"char":     true,
-	"nvarchar": true,
 	"nchar":    true,
+	"nvarchar2": true,
+	"varchar": true,
+	"varchar2": true,
 }
 
 // render a "cast" expression for a column so that we can cast it to the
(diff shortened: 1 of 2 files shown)


source-oracle/backfill.go
```diff
@@ -438,24 +438,30 @@ func (db *oracleDatabase) buildScanQuery(start bool, info *sqlcapture.DiscoveryI
...
+		quotedName := quoteColumnName(colName)
+		arg := fmt.Sprintf(":p%d", idx+1)
 		// If a precise backfill is requested *and* the column type requires binary ordering for precise
-		// backfill comparisons to work, add the 'COLLATE BINARY' qualifier to the column name.
-		if columnTypes[colName].JsonType == "string" && columnBinaryKeyComparison[colName] {
-			pkey = append(pkey, quotedName+" COLLATE BINARY")
+		// backfill comparisons to work, use NLSSORT with binary sorting so that comparisons
+		// and ordering are consistent regardless of database NLS settings. NLSSORT is used
+		// instead of the COLLATE operator for compatibility with Oracle 11g.
+		if columnTypes[colName].JsonType == "string" && columnBinaryKeyComparison[strings.ToLower(columnTypes[colName].Original)] {
```
Alex-Bair: It looks like you fixed a subtle bug here, too. Previously, this line had `&& columnBinaryKeyComparison[colName]`, but that always returned `false` since the keys in `columnBinaryKeyComparison` are column types, not column names, so this whole branch was never reached. Now we're correctly using the column type for the lookup. 👍 

nit: For readability purposes, I'd recommend storing the column type in a separate variable, like:

```suggestion
        var colType = strings.ToLower(columnTypes[colName].Original)
		if columnTypes[colName].JsonType == "string" && columnBinaryKeyComparison[colType] {
```

source-oracle/backfill.go
Alex-Bair, line 337: Since you fixed the `columnBinaryKeyComparison` lookup bug mentioned in this [comment](https://github.com/estuary/connectors/pull/4291/changes#r3125332886), we should make sure this map is complete. Looking at the list of supported string types in `discovery.go`:
https://github.com/estuary/connectors/blob/28ba70ecf858e2b0e7075a75e2474ea65e05ea6c/source-oracle/discovery.go#L286

The map is missing `varchar` and `nvarchar2`, and the existing `nvarchar` key doesn't match any supported type (although that was likely a typo that was meant to be `nvarchar2`). Could you please update this map so `VARCHAR` and `NVARCHAR2` primary key columns go through the binary sorting path?

Alex-Bair (commented):
I had a couple small items to address before approving. Looks good to me otherwise.

Could you please also include some more detail in the actual commit message about what changes are made in the commit? Copying the PR description in there is fine. That'll help us quickly understand what was done in the commit without having to look up the PR description.

Alex-Bair (approved):
LGTM, thanks!

PR #4295, opened 2026-04-22
migration: add test cases

### LINKED ISSUE #4293: add wide integer migration cases to RunMigrationTest
Can we add this case to RunMigrationTest instead?

_Originally posted by @mdibaiee in https://github.com/estuary/connectors/pull/4252#discussion_r3123342516_

jacobmarble:
**Description:**

Closes #4293
Helps https://github.com/estuary/connectors/issues/4246

Adds type migration test cases explicitly covering:
- 32, 64, 65, 128, 129, 256, and 257-bit integers
- 18, 19, 38, 39, 76, and 77-digit integers stored as strings
- unbounded forms of the above

Also renames test case properties to be clearer about "int", "bigInt", "hugeInt" - now the cases are named eg `int128ToInt256`.

**Workflow steps:**

**Notes for reviewers:**


materialize-s3-iceberg/testdata/migrate.flow.yaml
jacobmarble, line 1: Note that these are the only exclusion changes.

materialize-boilerplate/testdata/integration/collections.migrate-base.flow.yaml
```diff
@@ -4,22 +4,27 @@ collections:
       type: object
       properties:
         id: { type: integer }
-        bigIntToNum: { type: integer }
-        bigIntToHugeInt: { type: integer }
-        hugeIntToNum: { type: integer, maximum: 18446744073709551615 }
+        int64ToNum: { type: integer }
```
mdibaiee: I was initially confused about the names and the propertyes. There is `int64ToNum` which is originally `type: integer`, and then below there is `int64ToString`, which is originally `type: string, format: integer` -- these are two distinct types, but they are both called `int64`.

Perhaps `strInt64` vs `int64` or something like that?

mdibaiee: On a similar line, I think we should also check conversions between these two: from `int64` to `strInt64` and vice versa, etc.

materialize-s3-iceberg/testdata/migrate.flow.yaml
jacobmarble: With added conversion paths in 4af35aa98, exclusions are also added to DynamoDB, Redshift, S3-Iceberg, Spanner.

materialize-boilerplate/testdata/integration/collections.migrate-base.flow.yaml
```diff
@@ -5,27 +5,53 @@ collections:
...
         id: { type: integer }
         int64ToNum: { type: integer }
+        int64ToNumStr: { type: integer }
         int64ToInt128: { type: integer }
         int64ToInt256: { type: integer }
+        int64ToIntStr64: { type: integer }
+        int64ToIntStr128: { type: integer }
+        int64ToIntStr256: { type: integer }
         int128ToNum: { type: integer, maximum: 18446744073709551615 }
+        int128ToNumStr: { type: integer, maximum: 18446744073709551615 }
```
mdibaiee: This occurred to me now: why are we using unsigned integer max here instead of signed integer?

In the code we are using signed integers, so this should also be

`9,223,372,036,854,775,807` for a maximum 64 bit signed integer

materialize-boilerplate/testdata/integration/collections.migrate-base.flow.yaml
```diff
@@ -5,27 +5,53 @@ collections:
...
         int128ToInt256: { type: integer, maximum: 18446744073709551615 }
+        int128ToIntStr128: { type: integer, maximum: 18446744073709551615 }
+        int128ToIntStr256: { type: integer, maximum: 18446744073709551615 }
         int256ToNum: { type: integer, maximum: 1e39 }
```
mdibaiee: Similarly here we should use the precise boundary of maximum 128 bit signed integer:

```
170,141,183,460,469,231,731,687,303,715,884,105,727
```

materialize-boilerplate/testdata/integration/collections.migrate-base.flow.yaml
jacobmarble: This is meant to push just beyond the signed 64-bit limit. Similarly, the 256-bit values are `1e39` which is just beyond the signed 128-bit limit.

I've made two changes based on this feedback:
- int128 cases have `maximum: 1e38` and int256 have `maximum: 1e76` - just under the maximum bit-width threshold, instead of just above the minimum
- intStr128 cases have `maxLength: 38` and intStr256 cases have `maxLength: 76` - same threshold adjustments, drop the unused `maximum` value limit

After these changes, the MySQL snapshots changed because the `NUMERIC`=>`LONGTEXT` threshold is 65 digits.

jacobmarble:
In order to push the integer type `maximum` value higher, I've replaced `maximum: 1e38` with `maximum: 170141183460469231731687303715884105727.0` and `maximum: 1e76` with `maximum: 57896044618658097711785492504343953926634992332820282019728792003956564819967.0`. Why those silly `.0`s? Because the yaml parser doesn't handle such wide integer values, but it does handle very large float values. 🤷‍♂️

materialize-boilerplate/testdata/integration/collections.migrate-base.flow.yaml
```diff
@@ -4,23 +4,54 @@ collections:
...
-        bigIntToHugeInt: { type: integer }
-        hugeIntToNum: { type: integer, maximum: 18446744073709551615 }
+        int64ToNum: { type: integer }
+        int64ToNumStr: { type: integer }
+        int64ToInt128: { type: integer }
+        int64ToInt256: { type: integer }
+        int64ToIntStr64: { type: integer }
+        int64ToIntStr128: { type: integer }
+        int64ToIntStr256: { type: integer }
+        int128ToNum: { type: integer, maximum: 170141183460469231731687303715884105727.0 }
```
mdibaiee: The current checks are checking that for values that fit in 128, we use 128bit integers, it is not checking that we switch to 128 at the correct boundary.

I think one way to achieve checking the boundaries is to change the way we set the maximums, so for example for `int64`s we would set `math.MaxInt64`:

```
maximum: 9,223,372,036,854,775,807
```

For `int128` we would set `math.MaxInt64 + 1`
```
maximum: 9,223,372,036,854,775,808
```

For int256 we would set `math.MaxInt128 + 1`:
```
maximum: 170,141,183,460,469,231,731,687,303,715,884,105,728
```

This is checking that we actually switch from one to the other as soon as the boundary has been crossed. What do you think?

mdibaiee: To put it in perspective, the current tests would still pass if we were wrongly switching between integers at widely wrong boundaries. For example the current tests would pass if we chose int64 for a field with `maximum: math.MaxInt128 - 1`

jacobmarble: If we set the `maximum` attribute on the `int64`s then we don't exercise the path `maximum: 0`. We can add more test cases to cover both, of course.

In earlier states `maximum: 1e39` was intended to just exceeded the 128 bit threshold. Because these larger values lose precision in floating point, I don't think we can do that trick with `MaxInt128 + 1`, but I'll try and update here.

jacobmarble:
Claude wrote a Python script to help with the limits of 64-bit float values.
[tight_f64_bounds.py](https://github.com/user-attachments/files/27063733/tight_f64_bounds.py)

```text
N = 64
  int min       : -9223372036854775808
  f64 lower     : -9223372036854775808  (slack from int min: 0)
  int max       : 9223372036854775807
  f64 upper     : 9223372036854774784  (slack from int max: 1023)
  yaml literal  : minimum: -9223372036854775808
                  maximum: 9223372036854775807

N = 65
  int min       : -18446744073709551616
  f64 lower     : -18446744073709551616  (slack from int min: 0)
  int max       : 18446744073709551615
  f64 upper     : 18446744073709549568  (slack from int max: 2047)
  yaml literal  : minimum: -18446744073709551616.0
                  maximum: 18446744073709549568.0

N = 128
  int min       : -170141183460469231731687303715884105728
  f64 lower     : -170141183460469231731687303715884105728  (slack from int min: 0)
  int max       : 170141183460469231731687303715884105727
  f64 upper     : 170141183460469212842221372237303250944  (slack from int max: 18889465931478580854783)
  yaml literal  : minimum: -170141183460469231731687303715884105728.0
                  maximum: 170141183460469212842221372237303250944.0

N = 129
  int min       : -340282366920938463463374607431768211456
  f64 lower     : -340282366920938463463374607431768211456  (slack from int min: 0)
  int max       : 340282366920938463463374607431768211455
  f64 upper     : 340282366920938425684442744474606501888  (slack from int max: 37778931862957161709567)
  yaml literal  : minimum: -340282366920938463463374607431768211456.0
                  maximum: 340282366920938425684442744474606501888.0

N = 256
  int min       : -57896044618658097711785492504343953926634992332820282019728792003956564819968
  f64 lower     : -57896044618658097711785492504343953926634992332820282019728792003956564819968  (slack from int min: 0)
  int max       : 57896044618658097711785492504343953926634992332820282019728792003956564819967
  f64 upper     : 57896044618658091284033315468382851758786622968169871930916816872785223614464  (slack from int max: 6427752177035961102167848369364650410088811975131171341205503)
  yaml literal  : minimum: -57896044618658097711785492504343953926634992332820282019728792003956564819968.0
                  maximum: 57896044618658091284033315468382851758786622968169871930916816872785223614464.0

N = 257
  int min       : -115792089237316195423570985008687907853269984665640564039457584007913129639936
  f64 lower     : -115792089237316195423570985008687907853269984665640564039457584007913129639936  (slack from int min: 0)
  int max       : 115792089237316195423570985008687907853269984665640564039457584007913129639935
  f64 upper     : 115792089237316182568066630936765703517573245936339743861833633745570447228928  (slack from int max: 12855504354071922204335696738729300820177623950262342682411007)
  yaml literal  : minimum: -115792089237316195423570985008687907853269984665640564039457584007913129639936.0
                  maximum: 115792089237316182568066630936765703517573245936339743861833633745570447228928.0
```

materialize-boilerplate/testdata/integration/collections.migrate-base.flow.yaml
jacobmarble: I've got a much more complete set of migration paths now, including positive and negative integers with bit widths 0, 64, 65, 128, 129, 256, 257. Updating tests will take some time, however, because this uncovers a few uncovered paths that need to be investigated and possibly excluded.

For example, Postgres fails to migrate `int000` (unbounded `sql.INTEGER` == `BIGINT`) to `numStr` (`sql.STRING_NUMBER` == `DECIMAL`).

materialize-boilerplate/testdata/integration/collections.migrate-base.flow.yaml
```diff
@@ -3,44 +3,231 @@ collections:
...
+        emailToString: *email
+        hostnameToString: *hostname
+        uriToString: *uri
+        uriReferenceToString: *uriReference
+        uriTemplateToString: *uriTemplate
+        jsonPointerToString: *jsonPointer
+        relativeJsonPointerToString: *relativeJsonPointer
+        regexToString: *regex
+        uint32ToString: *uint32
+        uint64ToString: *uint64
```
mdibaiee: Here in this file this is called `uint64ToString`, but in `migrate-migrated.flow.yaml` it is `uint064ToString`

materialize-s3-iceberg/testdata/migrate.flow.yaml
```diff
@@ -16,8 +16,8 @@ materializations:
...
         recommended: true
         exclude:
-          - hugeIntToMulti
-          - hugeIntToNum
-          - hugeIntToString
-          - bigIntToHugeInt
+          - int128ToMulti
+          - int128ToNum
+          - int64ToInt128
+          - intStr128ToString
```
mdibaiee, line 22: These field names have also changed

materialize-boilerplate/testdata/integration/collections.migrate-migrated.flow.yaml
```diff
@@ -3,44 +3,237 @@ collections:
...
+        uriToString: *string
+        uriReferenceToString: *string
+        uriTemplateToString: *string
+        jsonPointerToString: *string
+        relativeJsonPointerToString: *string
+        regexToString: *string
+        uint32ToString: *string
+        uint064ToString: *string
+        iriToString: *string
+        iriReferenceToString: *string
```
mdibaiee, line 245: It seems like we are no longer exercising `type: integer` to `type: string`, and only `type: string, format: integer` to `type: string`

we previouly were also exercising that, it is a possible migration path that we could support

mdibaiee (approved):
this is awesome! LGTM % some minor comments

(let's hold off on merging it due to demo)

materialize-boilerplate/testdata/integration/collections.migrate-base.flow.yaml
jacobmarble: Hmm, I meant to keep `uint64` and `uint32`, but since this one slipped through, I'll rename them to `uint064` and `uint032` throughout.

materialize-boilerplate/testdata/integration/collections.migrate-migrated.flow.yaml
jacobmarble: Dang, that's a good catch.

jacobmarble:
The `materialize-mysql` test failure is defined by this error message:
```
creating endpoint: could not create a connection to database "test" at "svc-e1649f3e-a3e3-4b22-806e-6502c8b7d30e-dml.aws-oregon-4.svc.singlestore.com:3306": Error 1049 (42000): Unknown database 'test'
```

This is fixed in #4325

PR #4296, opened 2026-04-22
source-braintree-native: strip microseconds from cursors

Alex-Bair:
**Description:**

This PR is a follow-up to 0bc7c7b741. That commit fixed `disputes` but the other incremental streams (`subscriptions`, `customers`, `credit_card_verifications`, `transactions`) were still yielding cursors with microseconds via `datetime.now(tz=UTC) - LAG`. When a scheduled re-initialization fires, the CDK coordinates the new backfill cutoff with the current incremental cursor, so the microseconds leaked into `ResourceState.Backfill.cutoff` even though `resources.py` initializes it with `.replace(microsecond=0)`. This PR fixes this by preventing microseconds from making it into the incremental cursor in the first place.

Also, we're now stripping microseconds from the backfill window `end` so existing backfills with microsecond-laden cutoffs can complete instead of infinitely yielding the same page cursor on the final backfill iteration where `start < cutoff` but `floor(cutoff) == start`. The termination check is now `start >= end` instead of `start >= cutoff`.

### DIFF
--- a/source-braintree-native/source_braintree_native/api/__init__.py
+++ b/source-braintree-native/source_braintree_native/api/__init__.py
@@ -76,7 +76,7 @@ async def fetch_transactions(
         initial_end=min(
             log_cursor + timedelta(hours=window_size),
             datetime.now(tz=UTC) - LAG,
-        ),
+        ).replace(microsecond=0),
         log=log,
     )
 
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
LGTM

PR #4303, opened 2026-04-23
estuary-cdk: include status code in 5xx retry logs

Alex-Bair:
**Description:**

This PR adds the status code to the log emitted by the CDK when it automatically retries after receiving a 5xx response. This can be useful when troubleshooting APIs that consistently return 5xx errors for reasons we don't yet fully understand.


nicolaslazo (approved):
👍

PR #4304, opened 2026-04-23
source-outreach: use date windows for incremental replication

Alex-Bair:
**Description:**

Despite returning results in mostly sorted order per the `sort` query parameter, Outreach can still return slightly unsorted results. The connector's current replication strategies rely on the API returning results in non-decreasing order of a given cursor field, and we've seen both real-time and lookback subtasks for incremental streams consistently fail since that assumption doesn't hold. This PR makes `source-outreach` use an alternative replication strategy that uses date windows & doesn't rely on a specific ordering of results.

This PR's scope includes:
- revising `backfill_resources` and `fetch_resources` to not assume the API returns results in any particular order
- ensuring the `backfill_resources` function captures records at the `cutoff`
- allowing users to configure the window size used by `fetch_resources` via a new endpoint config setting

The spec snapshot change is expected due to the addition of the `advanced.window_size` endpoint config setting.

The connector's documentation should be updated to reflect the new config setting.

**Notes for reviewers:**

The changes are best reviewed commit-by-commit.

### DIFF
--- a/source-outreach/source_outreach/api.py
+++ b/source-outreach/source_outreach/api.py
@@ -23,6 +23,12 @@ MAX_PAGE_SIZE = 1000
 MIN_PAGE_SIZE = 1
 NEXT_PAGE_QUERY_PARAMETER = "page[after]"
 
+# MAX_WINDOW_SIZE bounds how far forward a single fetch_resources invocation
+# will advance. The Outreach API sometimes returns results unsorted within a
+# query, so we process records in explicit [lower, upper] date windows and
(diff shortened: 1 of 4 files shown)


nicolaslazo (approved):
Looks good!

Alex-Bair:
Post-approval, I removed the use of the `gt` operator in requests' query parameters. Outreach doesn't support the `gt` operator to filter on the cursor fields we use, so we need to continue using `gte` and filter out duplicate records client-side in the connector.

PR #4305, opened 2026-04-23
materialize-*: native_binary_column_type feature flag

### LINKED ISSUE #4302: Consistent handling of BINARY data types in materializations
So to sum up:

1. We want to start using `contentMediaType` to know more details about the type of the field. For binaries, we would like to have `application/octet-stream`. For other types we can also utilise the same
2. In materializations, for historical reasons, we are handling `contentEncoding: base64` inconsistently: some connectors materialize to string, some to binary. We want to add a new feature flag (enabled by default) so that all connectors now materialize to binary. Existing tasks will have that feature flag set to off to retain backward compatibility.

Going forward that means materializations will materialize any field with either:

1. `contentEncoding: base64`
2. `contentEncoding: base64` and `contentMediaType: application/octet-stream`

As BINARY columns in their destination

mdibaiee:
**Description:**

- Changes the behaviour of all connectors so that they use an actual BINARY data type column for materializing columns which have either:
  - `contentEncoding: base64` with no `contentMediaType`
  - `contentEncoding: base64` and `contentMediaType: application/octet-stream`
- We will set `no_native_binary_column_type` on existing tasks to disable the feature flag for existing tasks, retaining behaviour as-is. New tasks will have `native_binary_column_type` set by default.
- There is now a migration path bidirectionally between string and binary fields, so that customers can switch the feature flag on or off and be automatically migrated.

**Workflow steps:**

**Notes for reviewers:**


materialize-mysql/infile.go
```diff
@@ -80,31 +82,55 @@ func (i *infile) drain(ctx context.Context, txn *stdsql.Tx, drainQuery string) e
...
+		}
+		writeEscapedField(buf, string(b))
+		return nil
 	}
+}
 
-	return record, nil
+// writeEscapedField writes a string field, escaping characters that the
+// LOAD DATA parser would otherwise interpret as separators or NULL markers.
+func writeEscapedField(buf *bytes.Buffer, s string) {
```
danielnelson, line 126: suggestion: Consider using strings.Replacer in this function, I think it should have better performance and arguably style advantages?

materialize-mysql/sqlgen.go
```diff
@@ -363,13 +394,20 @@ ALTER TABLE {{$.Identifier}}
...
+-- work on SingleStore: the LOAD DATA SET clause that would decode @v_i runs
+-- after the row's NOT NULL constraint is enforced, which fails because the
+-- binary column is part of the primary key.
 
 {{ define "createLoadTable" }}
 CREATE TEMPORARY TABLE {{ template "temp_load_name" . }} (
 	{{- range $ind, $key := $.Keys }}
 		{{- if $ind }},{{ end }}
-		{{ $key.Identifier }} {{ $key.DDL }}
+		{{ $key.Identifier }} {{ if IsBinary $key }}VARCHAR(384) NOT NULL{{ else }}{{ $key.DDL }}{{ end }}
```
danielnelson: Can you tell me about the choice of 384 in VARCHAR(384)?  This is enough space for up to a 256 byte binary field to be encoded as a base64 field, so is 256 our limit for key length?

materialize-redshift/sqlgen.go
```diff
@@ -177,6 +197,26 @@ func toJsonCast(migration sql.ColumnTypeMigration) string {
...
+// which is the canonical textual representation of binary data in Flow.
+func varbyteToStringCast(migration sql.ColumnTypeMigration) string {
+	return fmt.Sprintf(`FROM_VARBYTE(%s, 'base64')`, migration.Identifier)
+}
+
+// stringToVarbyteCast decodes a base64-encoded VARCHAR/TEXT column into native
+// VARBYTE bytes. Used when the native_binary_column_type feature flag is
+// enabled on a task that previously stored binary fields as base64 text.
+func stringToVarbyteCast(migration sql.ColumnTypeMigration) string {
+	return fmt.Sprintf(`FROM_BASE64(%s)`, migration.Identifier)
```
danielnelson: Other queries are using `TO_VARBYTE`, should this also use it? `TO_VARBYTE(%s, 'base64')`

materialize-sql/element_converter.go
```diff
@@ -190,3 +192,50 @@ var ClampDate ElementConverter = StringCastConverter(func(str string) (interface
...
+// Base64Decoder decodes a base64-encoded tuple element into raw bytes. It is
+// used by materializations that store binary fields in native byte-typed
+// columns and bind the decoded bytes directly to their driver, rather than
+// decoding via SQL on the server side.
+var Base64Decoder ElementConverter = func(te tuple.TupleElement) (any, error) {
+	switch v := te.(type) {
+	case nil:
+		return nil, nil
+	case []byte:
+		return base64.StdEncoding.DecodeString(string(v))
```
danielnelson: This should avoid a copy:
```suggestion
		return base64.StdEncoding.Decode(v)
```

materialize-sql/element_converter.go
```diff
@@ -190,3 +192,50 @@ var ClampDate ElementConverter = StringCastConverter(func(str string) (interface
...
+// emitting hex from the connector, the loader can decode binary directly into
+// the native column without an intermediate staging step.
+var Base64ToHex ElementConverter = func(te tuple.TupleElement) (any, error) {
+	var raw []byte
+	var err error
+	switch v := te.(type) {
+	case nil:
+		return nil, nil
+	case []byte:
+		raw, err = base64.StdEncoding.DecodeString(string(v))
```
danielnelson:
```suggestion
		raw, err = base64.StdEncoding.Decode(v)
```

danielnelson (approved):

materialize-mysql/sqlgen.go
mdibaiee: That was the intention, yes... but I actually realised if we are considering 256 as the limit of varchar primary keys, then even in this staging table we can't create a larger primary key... This would limit binary keys to be even less than 256 because the string representation would have to fit under 256 and the binary representation would be even smaller... This is sort of historical: I think the reason we have this limit is to support older versions of MySQL with InnoDB (pre-5.7.7)

The reason we had this intermediate string format in staging tables was due to an issue with SingleStore that I hit during development, but that actually no longer applies and has been refactored away. So I reworked this to use the same type as the final primary key type in the staging table as well to avoid all of this conditional work. It should have been that way from the beginning. Thanks for noticing!

materialize-motherduck/sqlgen.go
```diff
@@ -163,13 +176,13 @@ JOIN read_json(
 	columns={
 	{{- range $ind, $bound := $.Bounds }}
 		{{- if $ind }},{{ end }}
-		{{$bound.Identifier}}: '{{$bound.DDL}}'
+		{{$bound.Identifier}}: '{{ if IsBinary $bound }}VARCHAR{{ else }}{{$bound.DDL}}{{ end }}'
```
jacobmarble: Does the nullability attribute matter here? The snapshot drops `NOT NULL`, but I'm not sure how `read_json()` works.

Ditto this comment 3-4 more times below.

materialize-mysql/sqlgen.go
```diff
@@ -162,6 +170,19 @@ var mysqlDialect = func(tzLocation *time.Location, database string, product stri
...
+	migrationSpecs["varchar"] = append(
+		migrationSpecs["varchar"],
+		sql.NewMigrationSpec([]string{"varbinary(256)"}, sql.WithCastSQL(stringToBinaryCast)),
+		sql.NewMigrationSpec([]string{"longblob"}, sql.WithCastSQL(stringToBinaryCast)),
+	)
+	migrationSpecs["longtext"] = append(
+		migrationSpecs["longtext"],
+		sql.NewMigrationSpec([]string{"varbinary(256)"}, sql.WithCastSQL(stringToBinaryCast)),
+		sql.NewMigrationSpec([]string{"longblob"}, sql.WithCastSQL(stringToBinaryCast)),
+	)
```
jacobmarble: Since this is unconditional, could it be moved to the `migrationSpecs` initialization? The `mariadb` conditional block could do the appending.

materialize-redshift/client.go
```diff
@@ -294,36 +294,32 @@ func (c *client) DeleteCheckpointsEntry(ctx context.Context, taskName string) er
...
 
-	// Redshift VARBYTE columns are returned as hex-encoded strings by the
-	// driver. Checkpoint data is stored as base64(gzip(bytes)) in VARBYTE,
-	// so the driver returns hex(utf8(base64(gzip(bytes)))). Decode this chain
-	// back to base64(bytes) so that snapshots match the common format used by
-	// other SQL materializations.
-	checkpointIdx := -1
+	// Checkpoint data is stored as base64(gzip(bytes)) in a VARBYTE column.
+	// snapshotTestTable queries VARBYTE values via FROM_VARBYTE(col, 'base64'),
+	// so the value already comes back as base64(utf8(base64(gzip(bytes)))).
```
jacobmarble, line 305: 🤯

materialize-redshift/sqlgen.go
```diff
@@ -261,16 +303,33 @@ CREATE TEMPORARY TABLE {{ template "temp_name" $.Target }} (
 -- Idempotent creation of the store table for staging new records.
 
 {{ define "createStoreTable" }}
+{{- if HasBinary $.Columns }}
```
jacobmarble: Is the condition worth the complexity? Does `CREATE TEMPORARY TABLE LIKE` do anything better than explicitly declaring the schema?

materialize-snowflake/sqlgen.go
```diff
@@ -179,6 +186,18 @@ func toJsonCast(migration sql.ColumnTypeMigration) string {
 	return fmt.Sprintf(`TO_VARIANT(%s)`, migration.Identifier)
 }
 
+func binaryToStringCast(migration sql.ColumnTypeMigration) string {
```
jacobmarble, line 189: The migration path binary-to-string is not in the integration tests, should we add it?

materialize-sql/element_converter.go
```diff
@@ -190,3 +193,50 @@ var ClampDate ElementConverter = StringCastConverter(func(str string) (interface
...
+		return nil, nil
+	case []byte:
+		raw, err = base64.StdEncoding.AppendDecode(nil, v)
+	case string:
+		raw, err = base64.StdEncoding.DecodeString(v)
+	case json.RawMessage:
+		raw, err = base64.StdEncoding.AppendDecode(nil, bytes.Trim(v, `"`))
+	default:
+		return nil, fmt.Errorf("Base64ToHex: unexpected value type %T", te)
+	}
```
jacobmarble: Maybe delegate to `Base64Decoder`:
```suggestion
	raw, err := Base64Decoder(te)
```

jacobmarble (approved):

materialize-snowflake/sqlgen.go
mdibaiee: I added both binaryToString and stringToBinary to integration tests

materialize-motherduck/sqlgen.go
mdibaiee: Important point I had missed, thanks!

mdibaiee:
I ran some tests to make sure that if I disable the feature flag, the snapshots of integration tests match the ones prior to the pull-request. I found some backward-incompatibilities this way and fixed those in [0ca8279](https://github.com/estuary/connectors/pull/4305/commits/0ca8279c8bb6d2f997ce6b79ac994fbe179b2093) -- now the snapshots match pre- and post-PR with the feature flag disabled.

materialize-mysql/.snapshots/TestSQLGeneration_Singlestore
jacobmarble, line 1: Did you intend to include this?

jacobmarble (approved):

materialize-mysql/.snapshots/TestSQLGeneration_Singlestore
mdibaiee: Yes, SingleStore has some conditional templating bits which are different from MySQL and MariaDB, so I think it is worth snapshotting it

mdibaiee:
Found two more bugs:

Migrations in Redshift and SQLServer were broken for specific edge cases. In case of redshift, if the base64-encoded string passed a certain number of characters, it would have `\n` which would break the `json_parse` due to newlines.

For SQLServer if the column name included quote characters it would break the migration cast.

Fixed both and added test cases to the integration tests to cover these

mdibaiee:
More updates to:
1. Simplify redshift's staging table by directly using a `VARBINARY`, same as destination table
2. Add migration support from binary to string to all connectors and update their snapshots

mdibaiee:
Found another bug in `no_flow_document` + `native_binary_column` in postgres which would return binary columns as hex instead of base64, fixed in [04fd71e](https://github.com/estuary/connectors/pull/4305/commits/04fd71e3bc8b81d105b8879e427e9541b19dad8a)

PR #4306, opened 2026-04-23
source-outreach: retry 500 Network Client Error with smaller page sizes

Alex-Bair:
**Description:**

We're seeing Outreach respond to specific `mailings` requests with a 500 status code and a "Network Client Error" body. I'm wondering if using smaller page sizes for these requests would let the connector get around these network client errors; the theory is that if we request less data, Outreach doesn't have to do as much work before responding. That approach worked for 502 Bad Gateway responses, so hopefully it works for 500 Network Client Error responses too.

**Notes for reviewers:**

I don't plan to merge this into `main` unless I see that it does work to get past those 500 Network Client Errors after placing the troublesome capture on the PR image tag. I'll confirm with a comment on this PR either way before merging or closing it.

### DIFF
--- a/source-outreach/source_outreach/api.py
+++ b/source-outreach/source_outreach/api.py
@@ -43,17 +43,35 @@ def _extract_page_cursor(links: OutreachResponse.Links | None) -> str | None:
     return cursor_list[0] if cursor_list else None
 
 
+def _is_502_bad_gateway_response(
+    status: int,
+    body: str,
+) -> bool:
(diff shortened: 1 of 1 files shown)


nicolaslazo (approved):
Approving so we can use the image tag in testing

Alex-Bair:
It didn't work, so I'm closing this PR. Outreach will need to troubleshoot why they're responding with a 500 Network Client Error for these requests.

PR #4307, opened 2026-04-23
materialize-sql: split BareDDL from NullableDDL

jacobmarble:
**Description:**

Until ClickHouse, `MappedType.NullableDDL` was always the unmodified "bare" form (`INTEGER` and never `INTEGER NULLABLE`) because no materializer sets a nullable modifier in practice. However, some contexts in which `NullableDDL` is read aren't actually interested in the nullable form, but in the bare form.

This splits the bare form from the nullable form so that ClickHouse can provide the two forms to the appropriate contexts.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-bigquery/sqlgen.go
+++ b/materialize-bigquery/sqlgen.go
@@ -216,7 +216,7 @@ CLUSTER BY {{ range $ind, $key := $.Keys }}
 ALTER TABLE {{$.Identifier}}
 {{- range $ind, $col := $.AddColumns }}
 	{{- if $ind }},{{ end }}
-	ADD COLUMN {{$col.Identifier}} {{$col.NullableDDL}}
+	ADD COLUMN {{$col.Identifier}} {{$col.BareDDL}}
 {{- end }};
 {{- end -}}
 {{- if $.DropNotNulls -}}
--- a/materialize-bigquery/sqlgen.go
+++ b/materialize-bigquery/sqlgen.go
@@ -216,7 +216,7 @@ CLUSTER BY {{ range $ind, $key := $.Keys }}
 ALTER TABLE {{$.Identifier}}
 {{- range $ind, $col := $.AddColumns }}
 	{{- if $ind }},{{ end }}
-	ADD COLUMN {{$col.Identifier}} {{$col.BareDDL}}
+	ADD COLUMN {{$col.Identifier}} {{$col.NullableDDL}}
 {{- end }};
 {{- end -}}
 {{- if $.DropNotNulls -}}
--- a/materialize-clickhouse/sqlgen.go
+++ b/materialize-clickhouse/sqlgen.go
@@ -341,6 +341,11 @@ SETTINGS
 ;
 {{ end }}
 
+-- ClickHouse permits adding non-nullable columns. The default value for a
+-- non-nullable column is the zero value (0, "", etc) for the column type.
+-- For that reason, this ALTER TABLE ... ADD COLUMN statement might look a bit
+-- strange compared to other SQL dialects.
+
 {{ define "alterTargetColumns" }}
 {{- range $ind, $col := $.AddColumns }}
 ALTER TABLE {{$.Identifier}} ADD COLUMN {{$col.Identifier}} {{$col.DDL}};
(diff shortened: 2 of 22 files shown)


materialize-bigquery/sqlgen.go
```diff
@@ -216,7 +216,7 @@ CLUSTER BY {{ range $ind, $key := $.Keys }}
 ALTER TABLE {{$.Identifier}}
 {{- range $ind, $col := $.AddColumns }}
 	{{- if $ind }},{{ end }}
-	ADD COLUMN {{$col.Identifier}} {{$col.NullableDDL}}
+	ADD COLUMN {{$col.Identifier}} {{$col.BareDDL}}
```
mdibaiee: Why is this changed to `BareDDL`? When adding a column to a table, it must be nullable.

materialize-databricks/sqlgen.go
```diff
@@ -161,7 +161,7 @@ CREATE TABLE IF NOT EXISTS {{$.Identifier}} (
 ALTER TABLE {{$.Identifier}} ADD COLUMN
 {{- range $ind, $col := $.AddColumns }}
 	{{- if $ind }},{{ end }}
-	{{$col.Identifier}} {{$col.NullableDDL}}
+	{{$col.Identifier}} {{$col.BareDDL}}
```
mdibaiee: Same comment here: columns added to tables must be nullable.

mdibaiee: Practically speaking these two are equal for these connectors, hence no issues being seen in the tests, however semantically, it is important that we use the nullable form of columns when adding them to a table.

materialize-sql/migration_mapping.go
```diff
@@ -40,7 +40,7 @@ type MigrationTarget interface {
 type StringTarget string
 
 func (t StringTarget) CanMigrate(_ boilerplate.ExistingField, mapped MappedType) bool {
-	return strings.EqualFold(string(t), mapped.NullableDDL)
+	return strings.EqualFold(string(t), mapped.BareDDL)
```
mdibaiee, line 43: 👍🏼

materialize-bigquery/sqlgen.go
jacobmarble: You're right!

materialize-databricks/sqlgen.go
jacobmarble: Right again! There are others besides these two; I'll audit before the next commit.

jacobmarble: Fun fact: Because (1) ClickHouse bare DDL implies not-null, and (2) ClickHouse not-null columns are zero-if-missing instead of error-if-missing, its type migration step 0 `ALTER TABLE ... ADD COLUMN` statement uses `$col.DDL` and does not need a step 4.

This is like playing chess with a different set of rules for each destination. 😄

materialize-clickhouse/sqlgen.go
```diff
@@ -343,7 +343,7 @@ SETTINGS
 
 {{ define "alterTargetColumns" }}
 {{- range $ind, $col := $.AddColumns }}
-ALTER TABLE {{$.Identifier}} ADD COLUMN IF NOT EXISTS {{$col.Identifier}} Nullable({{$col.NullableDDL}});
+ALTER TABLE {{$.Identifier}} ADD COLUMN {{$col.Identifier}} {{$col.DDL}};
```
jacobmarble, line 351: I'm leaving this as-is, but adding a comment to explain ClickHouse to future readers.

materialize-sqlserver/client.go
```diff
@@ -189,7 +189,7 @@ func (c *client) AlterTable(ctx context.Context, ta sql.TableAlter) (string, boi
 				return "", nil, fmt.Errorf("could not find column info for %q", dn.Name)
 			}
 
-			ddl, err := col.nullableDDL()
+			ddl, err := col.bareDDL()
```
jacobmarble: This name change stays because the task is explicitly dropping the not-null attribute from a column type.

mdibaiee: But semantically, we are trying to state the nullable form of the type so that it is now nullable (its `NOT NULL` has been dropped, so now it is nullable) -- I think `nullableDDL` works here 🤔

mdibaiee (approved):
LGTM % comment about sqlserver:

> semantically, we are trying to state the nullable form of the type so that it is now nullable (its NOT NULL has been dropped, so now it is nullable) -- I think nullableDDL works here 🤔

mdibaiee (approved):

PR #4309, opened 2026-04-24
source-ashby: implement snapshot and child streams

nicolaslazo:
**Description:**

All Ashby streams were treated as incremental and requiring no parent entity id parameters. This change

- Implements snapshot streams
- Sets flags to pull additional data when available
- Implements a guard for the scenario when docs are yielded but no sync token is provided by Ashby at the end of pagination

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-ashby/source_ashby/api.py
+++ b/source-ashby/source_ashby/api.py
@@ -5,13 +5,14 @@ from estuary_cdk.capture.common import LogCursor
 from estuary_cdk.flow import ValidationError
 from estuary_cdk.http import HTTPSession
 from estuary_cdk.incremental_json_processor import IncrementalJsonProcessor
+from pydantic import JsonValue
 
 from .models import (
     ApiKeyInfoResponse,
     AshbyEntity,
+    ChildEntityMixin,
     ResponseMeta,
 )
-
 
 API_BASE_URL = "https://api.ashbyhq.com"
 
@@ -19,7 +20,9 @@ API_BASE_URL = "https://api.ashbyhq.com"
 async def fetch_api_key_scopes(http: HTTPSession, log: Logger) -> set[str]:
     # Ashby requires Content-Type: application/json on all POST requests.
     # Passing json={} ensures aiohttp sets the header automatically.
-    response = await http.request(log, f"{API_BASE_URL}/apiKey.info", method="POST", json={})
+    response = await http.request(
+        log, f"{API_BASE_URL}/apiKey.info", method="POST", json={}
+    )
     info = ApiKeyInfoResponse.model_validate_json(response)
 
     if info.errorInfo is not None:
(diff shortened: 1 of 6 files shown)


source-ashby/source_ashby/resources.py
```diff
@@ -86,6 +107,45 @@ def open(binding, binding_index, state, task, all_bindings):
...
+        key=["/_meta/row_id"],
+        model=entity_cls,
+        open=open,
+        initial_state=ResourceState(),
+        initial_config=ResourceConfig(
+            name=entity_cls.name,
+            interval=timedelta(minutes=5),
+        ),
+        schema_inference=True,
+    )
```
Alex-Bair: question: Would it work to use the recently added [`SnapshotResource`](https://github.com/estuary/connectors/blob/2649a97ea0121e2223c308ded998f03262a193cd/estuary-cdk/estuary_cdk/capture/common.py#L478-L489) here instead of `Resource`? It's not necessary, but should remove the need to specify the `key` and `initial_state` during instantiation. Using `Resource` instead is fine, but if there's some rough edge that's preventing the use of `SnapshotResource` instead I'd like to know so I can smooth it out.

source-ashby/source_ashby/resources.py
```diff
@@ -86,6 +107,45 @@ def open(binding, binding_index, state, task, all_bindings):
...
+        task: Task,
+        all_bindings,
+    ):
+        open_binding(
+            binding,
+            binding_index,
+            state,
+            task,
+            fetch_snapshot=functools.partial(snapshot_fn, entity_cls, http),
+            tombstone=BaseDocument(_meta=BaseDocument.Meta(op="d")),
```
Alex-Bair, line 141: note: Similar to below, you shouldn't have to specify a `tombstone` here any longer as long as you're fine with the CDK [defaulting to use `BaseDocument` for the tombstone](https://github.com/estuary/connectors/blob/2649a97ea0121e2223c308ded998f03262a193cd/estuary-cdk/estuary_cdk/capture/common.py#L861-L865) like you are here. You can still specify it if you want, but it _should_ work without it.

source-ashby/source_ashby/api.py
```diff
@@ -87,11 +116,58 @@ async def fetch_entity(
...
+
+
+async def snapshot_child_entity(
+    entity_cls: type[ChildEntity],
+    http: HTTPSession,
+    log: Logger,
+) -> AsyncGenerator[ChildEntity, None]:
+    url = f"{API_BASE_URL}/{entity_cls.path}"
+
+    async for parent in snapshot_entity(entity_cls.parent_entity, http, log):
```
Alex-Bair: Sometimes we've seen that keeping a response open while we're iterating through parent ids & making requests for child resources can lead to `TimeoutErrors` ([here's the spot in `source-zendesk-support-native` where we fixed that issue](https://github.com/estuary/connectors/commit/1c0d4ed4fd48ab81978375bef3106a91238a780f)). That seems to be possible here too - the `snapshot_entity` call streams the API response for parent records, and we're keeping that response open while we make requests for child resources. We've handled this in other connectors by fetching an entire response's worth of parent records at a time, buffering the parent ids in a list, closing that response, then fetching the child records for those in memory parent ids. That'd be a good pattern to use here too to avoid potential `TimeoutError`s.

source-ashby/source_ashby/api.py
```diff
@@ -38,6 +40,32 @@ async def fetch_api_key_scopes(http: HTTPSession, log: Logger) -> set[str]:
...
+    async for item in processor:
+        yield item
+
+    meta = processor.get_remainder()
+    if not meta.success:
+        log.error(
+            "Ashby API error during snapshot",
+            extra={"entity": entity_cls.name, "errors": meta.errors},
+        )
+        raise RuntimeError(f"Ashby API error for {entity_cls.name}: {meta.errors}")
```
Alex-Bair: Both here and in `snapshot_child_entity`, do we need to paginate through results using `meta.nextCursor` like we do in `fetch_entity`? The API docs' `interviewEvent.list` [page](https://developers.ashbyhq.com/reference/intervieweventlist) seems to suggest pagination is necessary. It feels like the "paginate through all pages" logic could be abstracted to a helper and re-used across all three functions, although that's not always worth the extra effort if the code's fairly simple.

Alex-Bair (approved):
LGTM % the pagination question for snapshot resources and the timeout concern around concurrently fetching child records while streaming the response containing parent ids.

source-ashby/source_ashby/resources.py
nicolaslazo: No reason, it's just that it's a new change and I forgot it was there! I'll also update our Claude skills to reference it so it gets brought up in PR reviews

source-ashby/source_ashby/api.py
nicolaslazo: Good catch, will add it to the child streams skill as well

source-ashby/source_ashby/api.py
nicolaslazo: Some streams do, some don't. I'll have all functions use the same paginator

Alex-Bair (approved):
LGTM!

PR #4314, opened 2026-04-24
source-gong: treat Gong's 404-on-empty-filter as empty results

Alex-Bair:
**Description:**

Gong returns an HTTP 404 response with a body like:
```json
{"errors":["No calls found corresponding to the provided filters"]}
```
when a filtered list query matches zero records, instead of the conventional 200 with an empty array. This was bubbling up as a fatal `HTTPError` in `fetch_changes`, crashing the capture task on any empty window. Wrap request+parse in a helper that recognizes the "No ... found ... provided filters" error shape and returns an empty page.

### DIFF
--- a/source-gong/source_gong/api.py
+++ b/source-gong/source_gong/api.py
@@ -1,4 +1,5 @@
 import asyncio
+import re
 import time
 from collections import deque
 from datetime import UTC, datetime, timedelta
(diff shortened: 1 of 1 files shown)


nicolaslazo (approved):
LGTM!

PR #4315, opened 2026-04-27
materialize-s3-iceberg: switch integration tests to Polaris + rustfs

jacobmarble:
**Description:**

Replaces the `tip-catalog` (+ Postgres) integration test stack with `apache/polaris:latest` (REST catalog) backed by `rustfs/rustfs:latest` (S3-compatible). The integration tests now run end-to-end against a fully self-hosted Iceberg catalog — no encrypted real-AWS credentials required.

**Workflow steps:**

`go test ./materialize-s3-iceberg/...` — `TestIntegration` brings the compose stack up, creates the bucket on rustfs, creates a Polaris catalog with `stsUnavailable: true` so direct S3 creds are used, and tears down at the end. `TestMain` auto-resolves the iceberg-ctl Poetry venv so `PYTHON_PATH` no longer has to be set by hand.

None.

**Notes for reviewers:**

- `driver_test.go` replaces sops decryption + the tip-catalog warehouse-creation API call with plain config + Polaris management API (OAuth2 client_credentials → POST `/api/management/v1/catalogs`, then grant `service_admin → catalog_admin`).
- `icebergctl.go`/`iceberg-ctl/__main__.py` pass `s3.endpoint`/`s3.access-key-id`/`s3.secret-access-key`/`s3.region` through to pyiceberg's `RestCatalog` and strip the `X-Iceberg-Access-Delegation` header so we don't request vended credentials. `create_namespace` is now idempotent (catches `NamespaceAlreadyExistsError`).
- `driver.go`: `s3StoreConfig()` enables `UsePathStyle` automatically when a custom `s3_endpoint` is set — needed for any S3-compatible backend (rustfs here, also helpful for MinIO/R2).
- `testdata/config.yaml` is now plain YAML pointing at the local stack and sets `allow_existing_tables_for_new_bindings` because the apply test re-applies with `lastBinding=nil` while the table exists from the prior iteration.


materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
```diff
@@ -84,10 +84,26 @@ def run(
...
                     "default",
                     **{k:v for k, v in properties.items() if v is not None},
                 )
+                # We pass S3 credentials directly, so disable the catalog's
+                # credential-vending path. Some catalogs (e.g. Polaris) require
+                # a separate privilege for *_WITH_WRITE_DELEGATION operations
+                # that we don't need when not using vended credentials.
+                if hasattr(catalog, "_session"):
+                    catalog._session.headers.pop("X-Iceberg-Access-Delegation", None)
+                ctx.obj["catalog"] = catalog
```
mdibaiee: I'm a bit worried about these changes, and whether they will affect customer authentication that is already set up.

To make sure we are not going to break existing tasks two things we should test:
1. Running integration tests against the remote S3 as we used to, but with this code. So essentially have this piece of code change, but still use the previous test harness and see if it works -- this one we can do now
2. Release this slowly to a few tasks and monitor them before releasing this to everyone. After releasing to everyone as well we need to monitor the iceberg tasks closely to make sure there is no regression -- once 1 is done we can start this with an approval and using the image tag

mdibaiee: I think one option to minimise risk is to only change the auth method and headers if and only if direct S3 credentials are provided. i.e. the condition to strip `X-Iceberg-Access-Delegation` should be under `if cfg.aws_access_key_id and secret and region`

Also noting that `aws_access_key_id` and `aws_access_secret_key` are deprecated and customers do not use them anymore, and instead they specify `credentials.aws_access_key_id`.

In the current setup both of them are being passed to the catalog, I'm not surew hat are the implications of that

jacobmarble: I've split the commit into two:

9cc959f64 changes `__main__.py` and was force-pushed to get a passing test on the record: https://github.com/estuary/connectors/actions/runs/25073314574/job/73459468113?pr=4315

1ee7f2065 switches the test environment to polaris and rustfs.

jacobmarble: >     1. Running integration tests against the remote S3 as we used to, but with this code. So essentially have this piece of code change, but still use the previous test harness and see if it works -- this one we can do now

I believe this is done (see my last comment).

>     2. Release this slowly to a few tasks and monitor them before releasing this to everyone. After releasing to everyone as well we need to monitor the iceberg tasks closely to make sure there is no regression -- once 1 is done we can start this with an approval and using the image tag

This is something we do before merging, correct?

mdibaiee: > This is something we do before merging, correct?

Yes, we can try it together, also for the other PR for materialize-iceberg

mdibaiee (approved):
let's release it to a few tasks first and then merge 👍🏽

PR #4316, opened 2026-04-28
materialize-iceberg: reduce test duration from 15+ min to ~48s

jacobmarble:
**Description:**

Replaces the cloud-only integration test path with a hermetic docker-compose stack (Apache Polaris + RustFS + a local Spark cluster) so `go test ./materialize-iceberg/...` runs end-to-end without AWS/SOPS access. To make the test fast, PySpark jobs are submitted to a long-lived daemon holding a single `SparkSession` instead of a fresh `spark-submit` per job.

End-to-end `go test` wall: **~15:15 → ~48s** (~19× speedup; before this change, jobs hit a per-submit ~13s JVM cold-start, dominated by Iceberg classpath + Polaris OAuth).

**Workflow steps:**

`go test ./materialize-iceberg/...` brings up the compose stack, runs materialize/apply/migrate against it, then tears down. `ICEBERG_TEST_ALL=1` (or `-iceberg.test-all`) opts in to the existing cloud (REST/Glue/S3 Tables) configurations alongside the local stack.

None.

**Notes for reviewers:**

- Production EMR Serverless code path (`emr.go`, EMR-side python script entrypoints) is unchanged. The python scripts now expose a `run(spark, input)` function so the daemon can call them in-process; the `__main__` wrappers preserve the EMR `spark-submit` contract verbatim.
- New `computeType: "Spark Standalone"` in `config.go` is intentionally hidden from the JSON schema — only test configs use it.
- New compose services (`polaris`, `rustfs`, `spark-master`/`spark-worker`/`spark-daemon`) plus a Polaris bootstrap script that provisions a catalog + principal at `up` time.
- Existing cloud snapshots renamed to `TestIntegration-*`; cloud configs in `testdata/` are still gated by `ICEBERG_TEST_ALL`.


materialize-iceberg/spark.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (s *sparkClient) client() *http.Client {
+	if s.httpClient != nil {
+		return s.httpClient
+	}
+	// No timeout; jobs can run for tens of seconds and the connector's own
+	// context already carries the cancellation signal we care about.
+	s.httpClient = &http.Client{Timeout: 0}
+	return s.httpClient
```
mdibaiee: This lazy initialisation of http Client might be done in a race. I don't think it will necessarily affect things in practice, but `runJob` is called from multiple places and it is theoretically possible that multiple callers try to initialise httpClient at the same time. Can we just initialise httpClient when we initialise sparkClient or is there a reason to do it lazily?

materialize-iceberg/spark.go
```diff
@@ -0,0 +1,147 @@
...
+// rather than a process spawn.
+type sparkClient struct {
+	cfg                 sparkConfig
+	emrCfg              emrConfig
+	catalogAuth         catalogAuthConfig
+	catalogURL          string
+	warehouse           string
+	materializationName string
+	bucket              blob.Bucket
+	tokenURL            string
```
mdibaiee: none of these seem to be used at all?

materialize-iceberg/scripts/polaris-bootstrap.sh
```diff
@@ -0,0 +1,192 @@
...
+  i=0
+  while [ "$i" -lt 300 ]; do
+    resp=$(curl -fsS -X POST "${TOKENS}" \
+      -H "Content-Type: application/x-www-form-urlencoded" \
+      --data-urlencode "grant_type=client_credentials" \
+      --data-urlencode "client_id=${ROOT_CLIENT_ID}" \
+      --data-urlencode "client_secret=${ROOT_CLIENT_SECRET}" \
+      --data-urlencode "scope=PRINCIPAL_ROLE:ALL" 2>/dev/null) || resp=""
+    if [ -n "$resp" ]; then
+      token=$(echo "$resp" | sed -n 's/.*"access_token":"\([^"]*\)".*/\1/p')
```
mdibaiee: nit: this is fine, but we could use `jq` instead for a cleaner interface. I think it is already available in CI and in most environments because we use it in integration tests as well

materialize-iceberg/scripts/polaris-bootstrap.sh
```diff
@@ -0,0 +1,192 @@
...
+client_id=$(echo "$principal_resp" | sed -n 's/.*"clientId":"\([^"]*\)".*/\1/p')
+client_secret=$(echo "$principal_resp" | sed -n 's/.*"clientSecret":"\([^"]*\)".*/\1/p')
+
+if [ -z "$client_id" ] || [ -z "$client_secret" ]; then
+  # Principal already existed (409). Rotate its credentials so we have a known
+  # secret to hand back to the test harness.
+  log "rotating credentials for existing principal ${PRINCIPAL_NAME}"
+  rotate_resp=$(mgmt_post "${MGMT}/principals/${PRINCIPAL_NAME}/rotate" "{}")
+  client_id=$(echo "$rotate_resp" | sed -n 's/.*"clientId":"\([^"]*\)".*/\1/p')
+  client_secret=$(echo "$rotate_resp" | sed -n 's/.*"clientSecret":"\([^"]*\)".*/\1/p')
```
mdibaiee: If you decide to go with jq, these are also affected

materialize-iceberg/scripts/polaris-bootstrap.sh
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }
+}
+EOF
+)" >/dev/null
+
+log "creating principal ${PRINCIPAL_NAME}"
+principal_resp=$(mgmt_post "${MGMT}/principals" "{\"principal\": {\"name\": \"${PRINCIPAL_NAME}\"}}")
+client_id=$(echo "$principal_resp" | sed -n 's/.*"clientId":"\([^"]*\)".*/\1/p')
+client_secret=$(echo "$principal_resp" | sed -n 's/.*"clientSecret":"\([^"]*\)".*/\1/p')
```
mdibaiee: ditto

materialize-iceberg/spark.go
jacobmarble: Fixed by eagerly constructing the http.Client in `newMaterialization` and dropping the lazy accessor in 88c1a97e6.

materialize-iceberg/spark.go
jacobmarble: Right, none of those were read by any sparkClient method after the daemon refactor — dropped them all in 6506cd9b2.

materialize-iceberg/scripts/polaris-bootstrap.sh
jacobmarble: Switched the bootstrap container to alpine:3.20 (~5MB) so we can `apk add curl jq` and parse the responses with explicit jq paths in 0088c0243.

materialize-iceberg/scripts/polaris-bootstrap.sh
jacobmarble: Updated alongside the other two extractions in 0088c0243.

materialize-iceberg/scripts/polaris-bootstrap.sh
jacobmarble: Same — updated in 0088c0243.

materialize-iceberg/python/common.py
```diff
@@ -115,37 +124,137 @@ def common_args() -> argparse.Namespace:
...
+    """
+    use_local_s3 = bool(c.get("s3_endpoint"))
+
     builder = (
         SparkSession.Builder()
-        .appName("Iceberg Test")
-        .config(
-            "spark.jars.packages",
-            "org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.4.1,software.amazon.awssdk:bundle:2.20.160,software.amazon.awssdk:url-connection-client:2.20.160",
-        )
```
jacobmarble, line 128: Claude removed this, but I want to make sure it doesn't break EMR. I've forked a temporary branch which only removes `spark.jars.packages` from the EMR path (this change) and now running the tests with ICEBERG_TEST_ALL=true.

jacobmarble: Test ran ~51min on a branch that branched off main with **only** `spark.jars.packages` removed (`jgm-emr`).

**Result**: every Spark-driven path passes — `materialize` (load.py + merge.py) and `migrate` (exec.py column migration) all 9 cloud variants ✅. If the EMR runtime weren't shipping Iceberg, those would fail at session creation with `ClassNotFoundException: org.apache.iceberg.spark.SparkCatalog`. They didn't, including a 1m26s `evaluating loads` step on glue-catalog-auth — real Iceberg + Spark work.

So the production EMR Serverless application's release-label runtime supplies the Iceberg + AWS SDK jars, and removing `spark.jars.packages` is safe (and saves ~5–10s of Ivy resolution per cold start).

**Unrelated noise**: 3 of 9 `apply` variants failed — all s3tables — with AWS returning `400 Bad Request: The specified metadata is not valid` from the s3tables REST endpoint on a schema diff that simultaneously demotes `required` columns to `optional` and adds new columns. No Spark involved (Go connector calling AWS directly), so unrelated to this change. Same failure presumably reproduces on main.

Suggest landing the removal with a short comment in `_build_session` noting that the jars come from the EMR release-label runtime + the test docker image, so future readers don't wonder where they are.

materialize-iceberg/python/common.py
```diff
@@ -115,37 +124,137 @@ def common_args() -> argparse.Namespace:
...
         .config("spark.sql.catalog.estuary", "org.apache.iceberg.spark.SparkCatalog")
         .config("spark.sql.catalog.estuary.type", "rest")
-        .config(
+        .config("spark.sql.catalog.estuary.uri", c["catalog_url"])
+        .config("spark.sql.catalog.estuary.warehouse", c["warehouse"])
+        # The long-lived daemon shares a SparkSession across submissions.
+        # Without this, Iceberg's REST catalog caches table metadata for
+        # 30s, masking schema changes the connector makes between jobs
+        # (notably the column migration's add-temp-columns step).
+        .config("spark.sql.catalog.estuary.cache-enabled", "false")
```
mdibaiee: I'm a bit worried about performance implications of this for the non-migration path.

Maybe we can benchmark this a bit to see if there is a noticeable difference with this disabled, and if so, only disable it during Apply RPC?

mdibaiee (approved):
LGTM % note about caching and performance

Let's try this out on a few tasks with the image tag (the image tag as built by the CI) -- we can do this together

materialize-iceberg/python/common.py
jacobmarble: Good call — gated the cache-disable on a `shared_across_submissions` flag that only the daemon sets, so EMR entrypoints keep the default catalog cache. See 96ddb4625.

Reasoning: each EMR script is a fresh JVM with a session that exits when the job ends, so the 30s metadata cache is bounded by job lifetime and there's no cross-job staleness. The daemon is the only path that reuses a session across submissions, so it's the only path that needs the cache off.

materialize-iceberg/driver_test.go
```diff
@@ -2,26 +2,136 @@ package connector
...
+
+	// Ensure a clean slate: any leftover compose state plus a stale credentials
+	// file from a prior run would silently produce credentials that don't match
+	// the current Polaris instance. Tearing down explicitly is cheap relative
+	// to debugging an "unauthorized_client" failure later.
+	exec.Command("docker", "compose", "-f", composeFile, "-p", composeProject, "down", "-v").Run()
+	_ = os.Remove(credsPath)
+
+	upCmd := exec.Command("docker", "compose", "-f", composeFile, "-p", composeProject, "up", "--wait", "--wait-timeout", "180")
+	dockerUp = true
```
mdibaiee, line 67: This should be set after the error check below I think

mdibaiee (approved):
Thanks! minor comment below

materialize-iceberg/driver_test.go
jacobmarble: I put it here on purpose. `docker compose up` can return a non-zero exit status, and still leave some resources behind. Setting `dockerUp = true` here triggers the cleanup in `TestMain().

The docker compose setup/teardown is inconsistent throughout the materializer implementations, so a refactor is imminent.

PR #4325, opened 2026-04-28
materialize-mysql: replace SingleStore cloud workspace with dev container

jacobmarble:
**Description:**

Replaces the shared SingleStore cloud workspace dependency in `materialize-mysql` integration tests with the [SingleStore dev container](https://github.com/singlestore-labs/singlestoredb-dev-image), matching how mysql and mariadb already run. Eliminates a recurring source of test flakiness — most recently the `Unknown database 'test'` failure when the workspace's test DB was deleted.

**Workflow steps:**

`go test ./materialize-mysql/...` now spins up `mysql`, `mariadb`, and `singlestore` services from `docker-compose.yaml` and tears them down on cleanup. No external credentials, no `sops` decrypt, no API token. Also drops the `local.` segment from per-product config filenames since they are all local now.

None.

**Notes for reviewers:**

- The connector behavior is unchanged; only the test harness moved from cloud to local container.
- Snapshot diff in `.snapshots/TestIntegration-materialize` is 4 hunks, all in the singlestore section: each resource's `applied.actionDescription` now starts with the `flow_checkpoints_v1` `CREATE TABLE` (which the connector has always emitted on a fresh DB — the cloud workspace had a leftover table from prior runs, suppressing it via `IF NOT EXISTS`). The `materialization TEXT` column type comes from the existing product branch in `sqlgen.go:78-80`. The mariadb/mysql sections of the snapshot already contained the same DDL.
- Removed: `testdata/setup-singlestore.sh`, `testdata/api-token.singlestore.yaml`, the encrypted `testdata/config.singlestore.yaml`, and the cloud-workspace bringup block in `driver_test.go`.


materialize-mysql/.snapshots/TestIntegration-materialize
```diff
@@ -287,7 +287,7 @@ Table Data:
...
...SON at: /stringRelativeJsonPointerField with inferred types: [string]',\n\t\t`stringTimeField` TIME(6) COMMENT 'auto-generated projection of JSON at: /stringTimeField with inferred types: [string]',\n\t\t`stringUint32Field` NUMERIC(65,0) COMMENT 'auto-generated projection of JSON at: /stringUint32Field with inferred types: [string]',\n\t\t`stringUint64Field` NUMERIC(65,0) COMMENT 'auto-generated projection of JSON at: /stringUint64Field with inferred types: [string]',\n\t\t`stringUriField` LONGTEXT COMMENT 'auto-generated projection of JSON at: /stringUriField with inferred types: [string]',\n\t\t`stringUriReferenceField` LONGTEXT COMMENT 'auto-generated projection of JSON at: /stringUriReferenceField with inferred types: [string]',\n\t\t`stringUriTemplateField` LONGTEXT COMMENT 'auto-generated projection of JSON at: /stringUriTemplateField with inferred types: [string]',\n\t\t`stringUuidField` LONGTEXT COMMENT 'auto-generated projection of JSON at: /stringUuidField with inferred types: [string]',\n\n\t\tPRIMARY KEY (id)\n) CHARACTER SET=utf8mb4 COLLATE=utf8mb4_bin COMMENT='Generated for materialization acmeCo/tests/materialize-mysql/singlestore of collection tests/data-types';\n"]
```
danielnelson, line 290: Question: It looks like this adds the creation of `flow_checkpoints_v1`, why is this table new to the snapshot?

danielnelson (approved):

materialize-mysql/.snapshots/TestIntegration-materialize
jacobmarble: Singlestore tests previously ran against a shared cloud workspace where `flow_checkpoints_v1` already existed from prior runs. `materialize-sql`'s `Setup` (driver.go:270) only emits the checkpoints `CREATE TABLE` when the table isn't in the info schema, so on the cloud workspace it was a no-op and never appeared in `actionDescription`. The dev container starts empty each run, so Setup now emits it — matching what the mysql and mariadb sections of this snapshot already contain.

mdibaiee (approved):
nice!

PR #4327, opened 2026-04-28
materialize-iceberg: use uri from catalog config

danielnelson:
If the catalog config from /config specifies a uri, it should be used for future requests instead of the catalog url provided by the user.

One example of a catalog using this is Nessie.  With this catalog you connect to a branch specific catalog URL such as /iceberg/my-branch, the catalog config it will return a branch prefix and a new catalog URI. When the uri and prefix are combined it will be the correct URL for future requests to this branch.

**Workflow steps:**

No changes

None

**Notes for reviewers:**

Nessie example:

To use a branch, the initial catalog url might be:
```
https://example.org/iceberg/my-branch
```

The connector checks the iceberg catalog for the catalog config at:
```
https://example.org/iceberg/my-branch/v1/config
```

Nessie replies with:
```
{
    "defaults": {
        "prefix": "my-branch",
    },
    "overrides": {
        "uri": "https://example.org/iceberg/"
    }
}
```

In this case for future requests, such as listing namespaces, we should use:
```
https://example.org/iceberg/v1/my-branch/namespaces
```

### DIFF
--- a/materialize-iceberg/catalog/catalog.go
+++ b/materialize-iceberg/catalog/catalog.go
@@ -278,16 +278,29 @@ func New(ctx context.Context, catalogUrl string, warehouse string, opts ...Catal
 		if p, ok := catalogCfg.Defaults["prefix"]; ok {
 			prefix = p
 		}
+		if u, ok := catalogCfg.Overrides["uri"]; ok {
+			uri, err := url.Parse(u)
+			if err != nil {
+				return nil, fmt.Errorf("unable to parse defaults.uri %q: %w", uri, err)
+			}
+			baseURL = uri.JoinPath("v1")
+		}
 	}
 	if catalogCfg.Overrides != nil {
 		if p, ok := catalogCfg.Overrides["prefix"]; ok {
 			prefix = p
-
+		}
+		if u, ok := catalogCfg.Overrides["uri"]; ok {
+			uri, err := url.Parse(u)
+			if err != nil {
+				return nil, fmt.Errorf("unable to parse overrides.uri %q: %w", uri, err)
+			}
+			baseURL = uri.JoinPath("v1")
 		}
 	}
-	if prefix != "" {
-		rc.rHttp = rc.rHttp.SetBaseURL(baseURL.JoinPath(prefix).String())
-	}
+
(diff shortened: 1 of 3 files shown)


materialize-iceberg/catalog/catalog.go
```diff
@@ -278,16 +278,29 @@ func New(ctx context.Context, catalogUrl string, warehouse string, opts ...Catal
 		if p, ok := catalogCfg.Defaults["prefix"]; ok {
 			prefix = p
 		}
+		if u, ok := catalogCfg.Overrides["uri"]; ok {
```
mdibaiee: Is this supposed to be `catalogCfg.Defaults["uri"]`? Right now this is exactly the same code as below with `Overrides["uri"]`

danielnelson: Yes, thank you.  Fixed now.

materialize-iceberg/catalog/catalog_test.go
```diff
@@ -0,0 +1,117 @@
...
+				require.NoError(t, err)
+				w.Write(respBody)
+			},
+			check: func(catalogURL, baseURL *url.URL) {
+				require.Equal(t, baseURL.Path, "/iceberg/v1/main")
+				require.Equal(t, catalogURL.Host, baseURL.Host)
+			},
+		},
+		{
+			name: "with uri",
```
mdibaiee: Let's also add a test for defaults uri
```suggestion
		{
			name: "with defaults uri",
			handler: func(w http.ResponseWriter, r *http.Request) {
				w.Header().Set("Content-Type", "application/json")
				w.WriteHeader(http.StatusOK)
				respBody, err := json.Marshal(map[string]any{
					"defaults": map[string]string{
						"uri": "http://example.org/iceberg/",
					},
					"overrides": map[string]string{},
				})
				require.NoError(t, err)
				w.Write(respBody)
			},
			check: func(catalogURL, baseURL *url.URL) {
				require.Equal(t, baseURL.String(), "http://example.org/iceberg/v1")
			},
		},
		{
			name: "with overrides uri",
```

mdibaiee (approved):
LGTM % small suggestion

PR #4336, opened 2026-04-29
source-appsflyer: relax restrictions on event_value field

nicolaslazo:
It has been observed that the `event_value` field being used for the construction of document surrogate keys is not present in some streams. To summarise:

1. All Android apps report `event_value` in all event types
2. The following event types include `event_value` on iOS apps:
    a. organic-install-in-app-event
    b. re-attribution-in-app-event
    c. re-engagement-in-app-event
3. All other iOS streams never include it

This change relaxes the field requirement only where it's strictly necessary, while enforcing our capture setup requirements as much as we can.

**Description:**

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-appsflyer/source_appsflyer/models.py
+++ b/source-appsflyer/source_appsflyer/models.py
@@ -77,6 +77,13 @@ class EventTypesResponse(BaseModel):
     event_types: list[str]
 
 
+KNOWN_STREAMS_WITH_EVENT_VALUE_FIELD = frozenset({
+    "organic-install-in-app-event",
+    "re-attribution-in-app-event",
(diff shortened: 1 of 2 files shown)


Alex-Bair (approved):
LGTM % is it intentional to make `event_type` a required field? I see that's being done in this PR too.

nicolaslazo:
> is it intentional to make event_type a required field? I see that's being done in this PR too

That's a deliberate choice. We need it to determine if a message belongs to one of the event types that always contain `event_value`. It's the same field we use to route messages to their respective collections, so it has always been implicitly required.

I'll update the documentation to reflect this.

PR #4343, opened 2026-04-30
source-shopify-native: add some fields to `orders`

Alex-Bair:
**Description:**

- `app.name`
- `title` for the concrete types that have it and implement the `discountApplications` interface
- actually fill out the `value` object for `discountApplications`

**Notes for reviewers:**

These should all be free to add since they don't require any new scopes or additional connections beyond what we were already using.


nicolaslazo (approved):
LGTM

PR #4344, opened 2026-04-30
source-appsflyer: extend `event_value` special case to all platforms

nicolaslazo:
The document validator carried the wrong assumption that only iOS apps could lack the `event_value` field. This change relaxes that restriction, leaving it active for just the three streams we know to hold it across all platforms:

```python
KNOWN_STREAMS_WITH_EVENT_VALUE_FIELD = frozenset(
    {
        "organic-install-in-app-event",
        "re-attribution-in-app-event",
        "re-engagement-in-app-event",
    }
)
```

Additionally, we now introspect the app id to know if we need to check for the presence of `app_type`.


Alex-Bair (approved):
LGTM

PR #4347, opened 2026-04-30
materialize-s3-iceberg: s/python/iceberg-go/

jacobmarble:
**Description:**

Replaces the embedded Python `iceberg-ctl` subprocess with the native Go `iceberg-go` library; the connector now creates and writes Iceberg tables in-process. Behavior and config schema are kept intentionally equivalent to the Python implementation.

**Workflow steps:**

No user-visible changes.

None.

**Notes for reviewers:**

- Two commits: a shared `go/schema-gen` enabler (traverse `oneOf`/`anyOf` when applying fixups) and the connector migration itself. The Python `iceberg-ctl/` tree and the Python Docker build stage are removed.
- Behavior held equivalent to the old tool: date/timestamp clamp to years 1–9999 (connector-local, in `clamp.go`), legacy top-level `aws_access_key_id`/`aws_secret_access_key` promoted into `credentials` on decode, info-schema lists existing namespaces first and skips absent ones, and `appendFiles` keeps the `flow_checkpoints_v1` idempotency fence plus the pyiceberg-default duplicate-file check.
- Config-schema parity: nullable scalars still emit `["string","null"]` and `upload_interval` keeps `format: duration` (`schema_compat.go`). The credentials block now uses the shared `filesink`/`iam` schema instead of a bespoke one.


mdibaiee:
@jacobmarble do we still want to do this?
I'm just afraid of the risk of regressions with this, seems like a sudden lift, but I like the path... mixing Python and Go is quite the mess

jacobmarble:
> @jacobmarble do we still want to do this? I'm just afraid of the risk of regressions with this, seems like a sudden lift, but I like the path... mixing Python and Go is quite the mess

I still want to do it, sooner is better IMHO.

The Python Iceberg implementation is more mature than the Golang impl, but we don't push it very hard.

The PR hasn't been updated recently because no one had reviewed it yet. If you support the effort (with caveat to mitigate regression risks) then I'll rebase and squash to make the review easier.

mdibaiee:
@jacobmarble yeah I think this is worth doing if we can mitigate the risk 👍🏽

jacobmarble:
> @jacobmarble yeah I think this is worth doing if we can mitigate the risk 👍🏽

This change has been deployed to two tasks for several weeks now.
- `jacob/asdb-exchange-ice/materialize-s3-iceberg` - this PR
- `jacob/asdb-exchange-ice/materialize-s3-iceberg-partitioned` - this PR, plus Iceberg table partitioning WIP commit

I'll rebase and cleanup any regressions. Also need to test the change to `go/writer/parquet.go`.

materialize-s3-iceberg/catalog.go
```diff
@@ -4,139 +4,368 @@ import (
...
+			props[icebergio.S3AccessKeyID] = cfg.Credentials.AWSAccessKeyID
+			props[icebergio.S3SecretAccessKey] = cfg.Credentials.AWSSecretAccessKey
+			props[icebergio.S3Region] = cfg.Region
+		case filesink.AWSIAM:
+			// AWSAccessKeyID / AWSSecretAccessKey / AWSSessionToken under
+			// IAMTokens are runtime-injected session credentials produced by
+			// assuming the configured role — not user-provided keys.
+			props[icebergio.S3AccessKeyID] = cfg.Credentials.AWSAccessKeyID
+			props[icebergio.S3SecretAccessKey] = cfg.Credentials.AWSSecretAccessKey
+			props[icebergio.S3SessionToken] = cfg.Credentials.AWSSessionToken
```
mdibaiee:
```suggestion
			props[icebergio.S3AccessKeyID] = cfg.Credentials.IAMTokens.AWSAccessKeyID
			props[icebergio.S3SecretAccessKey] = cfg.Credentials.IAMTokens.AWSSecretAccessKey
			props[icebergio.S3SessionToken] = cfg.Credentials.IAMTokens.AWSSessionToken
```

materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
```diff
@@ -1,525 +0,0 @@
...
-):
-    log(f"alter_table: altering table {table}")
-    catalog = ctx.obj["catalog"]
-    assert isinstance(catalog, Catalog)
-
-    table_alter = TableAlter.model_validate_json(table_alter_json)
-
-    tbl = catalog.load_table(table)
-
-    with tbl.update_schema() as update:
```
mdibaiee, line 389: this `update_schema` defaults to `case_sensitive=True` I think

materialize-s3-iceberg/catalog.go
```diff
@@ -183,23 +411,36 @@ func (c *catalog) UpdateResource(_ context.Context, bindingUpdate boilerplate.Bi
...
 
 	return fmt.Sprintf("alter table %q", fqn), func(ctx context.Context) error {
-		if _, err := runIcebergctl(ctx, c.cfg, "alter-table", fqn, string(input)); err != nil {
+		tbl, err := c.cat.LoadTable(ctx, bindingUpdate.Binding.ResourcePath)
+		if err != nil {
+			return fmt.Errorf("loading table %q: %w", fqn, err)
+		}
+
+		tx := tbl.NewTransaction()
+		us := tx.UpdateSchema(false, false)
```
mdibaiee: But this one is doing `caseSensitive=false` so this seems like an incompatibility

mdibaiee (changes requested):
Claude thinks Glue catalogs will fail with this new implementation. So I would suggest that we set up the Glue integration tests (I think we have some Glue catalog setup for materialize-iceberg, we could maybe use similar setup here), and then run the integration tests and iterate until we can verify it works well. I also noticed that actually in our [docs for this connector](https://docs.estuary.dev/reference/Connectors/materialization-connectors/amazon-s3-iceberg/) we explicitly advertise this connector for use with AWS Glue Catalog, but we are running our tests against Polaris REST catalog. So this is definitely something we need to do, to run tests against a Glue catalog

materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
jacobmarble: Fixed

materialize-s3-iceberg/catalog.go
jacobmarble: Fixed

jacobmarble:
> Claude thinks Glue catalogs will fail with this new implementation. So I would suggest that we set up the Glue integration tests (I think we have some Glue catalog setup for materialize-iceberg, we could maybe use similar setup here), and then run the integration tests and iterate until we can verify it works well.

Added Glue integration tests, selected with env var `S3_ICEBERG_TEST_GLUE=1`.

mdibaiee:
@jacobmarble  can we have both glue and rest be run by default in the CI? I'm afraid of drift between the two

jacobmarble:
> @jacobmarble can we have both glue and rest be run by default in the CI? I'm afraid of drift between the two

Done

mdibaiee:
@jacobmarble sorry for being annoying, but if we can have the namespace names match between REST and Glue catalogs, then we can expect that the snapshots are byte-identical. Right now I think the only difference between the snapshots is the namespace (`tests` vs `s3_iceberg_tests`)

jacobmarble:
> @jacobmarble sorry for being annoying, but if we can have the namespace names match between REST and Glue catalogs, then we can expect that the snapshots are byte-identical. Right now I think the only difference between the snapshots is the namespace (`tests` vs `s3_iceberg_tests`)

Done. Snapshots are byte-identical now. This surfaced a glitch in the Glue integration test - it depended on an existing namespace, and didn't clean it up. Making the snapshots identical required both to either create and cleanup their namespace, or not.

mdibaiee (approved):
Thanks a lot! Let's merge on Monday given that this can be a risky deployment 🤞🏽

PR #4348, opened 2026-05-01
materialize-s3-iceberg: add additional_table_properties resource field

### LINKED ISSUE #3984: Materialize-s3-iceberg additonal table creation statement
Iceberg table metadata can easily exceed the 5mb file size limit if its retaining a lot of old snapshots. Example:

Table metadata file s3://<redacted>/metadata/<>.metadata.json exceeds the maximum allowed size. To reduce metadata size, consider: expiring old snapshots, reducing schema complexity, limiting table properties, and removing unnecessary partition specs or sort orders

Users can set up a job that cleans up old snapshots, but for quality of life, it would be easier if we were able to create the table with an expiration limit: history.expire.min-snapshots-to-keep.

Suggestion:

- Configuration field option for setting history.expire.min-snapshots-to-keep
- Table creation includes the value of this field

jacobmarble:
Closes #3984

**Description:**

Adds an `additional_table_properties` field to the materialize-s3-iceberg resource config. The map is forwarded through `iceberg-ctl create-table` and merged into the property dict passed to `pyiceberg.catalog.create_table` at table-creation time. Properties are set only when the table is created and cannot be changed afterward.

`pyiceberg.table.TableProperties.DEFAULT_NAME_MAPPING` is filtered out of user input so a binding cannot clobber the name mapping that pyiceberg requires to read the table.

Example resource config:

```json
{
  "table": "my_table",
  "additional_table_properties": {
    "write.metadata.compression-codec": "gzip",
    "write.metadata.delete-after-commit.enabled": "true",
    "write.metadata.previous-versions-max": "10",
    "history.expire.min-snapshots-to-keep": "1",
    "history.expire.max-snapshot-age-ms": "86400000"
  }
}
```

**Workflow steps:**

Applies to newly-created tables. Existing tables are unaffected.

None.

**Relationship to `materialize-iceberg`:**

`materialize-iceberg` already exposes an identical `additional_table_properties` resource field (same JSON tag, same title, same description, same "set only at creation" semantics). This PR brings the same user-facing config knob to `materialize-s3-iceberg` so the two connectors behave consistently from the catalog perspective.

Implementation differences are limited to how the property map reaches the catalog:

- `materialize-iceberg` passes the map directly to its Go REST-catalog client in the `createTableRequest.properties` field.
- `materialize-s3-iceberg` JSON-marshals the map into the `iceberg-ctl create-table` invocation; the Python side merges it into the dict handed to `pyiceberg.catalog.create_table` and filters out `DEFAULT_NAME_MAPPING` (which `materialize-iceberg` does not need to guard against because it constructs the catalog request directly).

**Notes for reviewers:**

The first commit on this branch hardcoded a fixed set of metadata-compaction and snapshot-expiration properties; the second commit replaces that with the user-configurable field. Net effect vs main is the new resource field only.

### DIFF
--- a/materialize-s3-iceberg/.snapshots/TestSpec
+++ b/materialize-s3-iceberg/.snapshots/TestSpec
@@ -339,6 +339,14 @@
         "title": "Delta Update",
         "description": "Should updates to this table be done via delta updates. Currently this connector only supports delta updates.",
         "default": true
+      },
+      "additional_table_properties": {
+        "additionalProperties": {
+          "type": "string"
+        },
+        "type": "object",
+        "title": "Additional Table Properties",
+        "description": "Additional Iceberg table properties to set when the table is created. These are set only at creation time and cannot be changed afterwards. Example: {'write.parquet.compression-codec': 'zstd'}"
       }
     },
     "type": "object",
--- a/materialize-s3-iceberg/catalog.go
+++ b/materialize-s3-iceberg/catalog.go
@@ -109,8 +109,11 @@ func (c *catalog) createNamespace(ctx context.Context, namespace string) error {
 	return err
 }
 
-func (c *catalog) CreateResource(ctx context.Context, b *pf.MaterializationSpec_Binding) (string, boilerplate.ActionApplyFn, error) {
-	tc := tableCreate{Location: tablePath(c.cfg.Bucket, c.cfg.Prefix, b.ResourcePath[0], b.ResourcePath[1], c.locationStyle)}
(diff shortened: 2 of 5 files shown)


mdibaiee:
@jacobmarble I think we want to allow customers to set their own properties on tabled as well as being able to override our defaults here, otherwise they will feel stuck with defaults that may not suit them. I'm not sure about the product impact of these new defaults and whether customers may be surprised by them. Worth checking with product people to see if they agree with these new defaults.

mdibaiee (approved):

PR #4349, opened 2026-05-01
source-iterable-native: expand `list_users` handling for skipping lists when there are consistent transport-level errors

Alex-Bair:
**Description:**

This PR's scope includes:
- removing the 4 hour max limit for the configurable `list_users_timeout` setting
- skipping lists that have consistent transport-level errors when backfilling `list_users`. The `list_users` binding defaults to constantly backfilling anyway, so any lists that are skipped in the current backfill will be retried on future backfills.

The connector's documentation should be updated to reflect that there's no longer a 4 hour maximum limit on the `list_users_timeout` setting.


nicolaslazo (approved):
LGTM

PR #4351, opened 2026-05-01
source-oracle: flag NUMBER/FLOAT primary keys as having unpredictable ordering

Alex-Bair:
**Description:**

Bare `NUMBER`, `NUMBER(p, s>0)`, and `FLOAT` columns are discovered as
`JsonType="string"` + `Format="number"`, which `encodeKeyFDB` encodes as a raw
string. Bytewise lexicographical order on these keys disagrees with Oracle's numeric
`ORDER BY` (ex: "10000" < "9999"), tripping the precise-backfill ordering
sanity check.

This PR avoids tripping the precise-backfill ordering assertion by
marking tables that have `format: number` primary keys with
`UnpredictableKeyOrdering` so they use the unfiltered backfill path.


source-oracle/discovery.go
```diff
@@ -71,10 +71,36 @@ func (db *oracleDatabase) DiscoverTables(ctx context.Context) (map[sqlcapture.St
...
 
+// Returns true if the bytewise lexicographic ordering of serialized row keys for
+// this column type matches the database backfill ordering.
+func predictableColumnOrder(colType any) bool {
+	// NUMBER columns without a scale-zero constraint are discovered as
+	// JsonType="string" + Format="number" and encoded as raw strings,
+	// whose bytewise lex order disagrees with Oracle's numeric ORDER BY
+	// (ex: "10000" < "9999").
+	if t, ok := colType.(oracleColumnType); ok {
+		if t.JsonType == "string" && t.Format == "number" {
```
mdibaiee, line 97: should `Format == "integer"` be also included?

Alex-Bair: The connector is already zero-padding `Format == "integer"` columns here so they aren't hitting the same lexicographical order vs. numeric order issue as `Format == "number"` columns.
https://github.com/estuary/connectors/blob/14b093aa10189ef97c6ab0702b8d6b3d46df461b/source-oracle/main.go#L380-L388

 I can include them though if you think the current zero-padding handling isn't sufficient for `Format == "integer"` columns?

mdibaiee: oh I see, so `format: number` is the only type used for unknown precision numeric values. Makes sense 👍🏽

mdibaiee (approved):

PR #4355, opened 2026-05-04
estuary-cdk: implement transactor for ACK management

### LINKED ISSUE #4320: estuary-cdk: add support for explicit ACKs on webhook connectors
Webhook captures must confirm that Flow has persisted incoming messages before replying to HTTP requests. Responding with a 200 status code signals that data has been properly persisted and we should assume it will never be resent, therefore these confirmations must be accurate.

Setting `explicitAcknowledgements: true` in connectors' "opened" messages makes Flow respond with

```
{checkpoints: N}
```

meaning that the last N checkpoints have been saved. This kind of state tracking is not currently supported in the CDK. If the control plane crashes and some pull API messages are dropped, the connector would restart from an earlier cursor and those messages would be re-fetched.

The solution is to implement an analogue of the `source-http-ingest`'s [transactor](https://github.com/estuary/connectors/blob/main/source-http-ingest/src/transactor.rs). The transactor should:

- Hold every checkpoint requester and receive all ACKs
- Block webhook checkpoints until these are received
- Not block pull API `checkpoint()` calls. Callers do not need ACKs to progress and they should be silently dropped

nicolaslazo:
**Description:**

Webhook servers were responding with 200s to incoming requests without waiting for the Flow runtime to confirm they have been persisted; this change makes `checkpoint()` calls block on ACKs for webhook tasks.

The Flow runtime responds with `{checkpoints: N}` to signal that the first N pending checkpoints have been processed, so a deque was implemented to track the order in which they've been made.

The transactor is designed to keep pull API checkpoints fire-and-forget to not harm performance.

**Workflow steps:**

**Notes for reviewers:**

Recommended file reading order:

1. `base_capture_connector.py`
2. `transactor.py`
3. `common.py`
4. Tests
5. `source-appsflyer/resources.py`

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
+++ b/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
@@ -28,6 +28,7 @@ from ..utils import format_error_message, get_running_tasks_info, sort_dict
 from . import Request, Response, Task, request, response
 from ._emit import emit_bytes
 from .common import _ConnectorState
+from .transactor import Transactor
 
 # Default encryption service URL, can be overridden via ENCRYPTION_URL environment variable
 # for local testing with docker gateway addresses like http://172.18.0.1:port
@@ -57,6 +58,12 @@ class BaseCaptureConnector(
     metaclass=abc.ABCMeta,
 ):
     output: BinaryIO = sys.stdout.buffer
+
+    _transactor: Transactor
+    """Constructed during `open`. Coordinates stdout writes for every Task and,
+    when the capture has at least one binding requiring Flow's explicit
+    Acknowledge protocol, lets ACK-requiring Tasks block their checkpoints
+    until Flow confirms persistence."""
 
     @abc.abstractmethod
     async def spec(self, log: FlowLogger, _: request.Spec) -> ConnectorSpec:
@@ -108,7 +115,7 @@ class BaseCaptureConnector(
         return None # No-op by default.
 
     async def acknowledge(self, acknowledge: request.Acknowledge) -> None:
-        return None  # No-op.
+        self._transactor.deliver_ack(acknowledge.checkpoints)
 
     async def handle(
         self,
@@ -140,6 +147,11 @@ class BaseCaptureConnector(
 
             opened, capture = await self.open(log, open)
             await self._emit(Response(opened=opened))
+
+            self._transactor = Transactor(
+                self.output,
+                requires_explicit_acks=opened.explicitAcknowledgements,
+            )
 
             stopping = Task.Stopping()
 
@@ -233,6 +245,7 @@ class BaseCaptureConnector(
                         self.output,
                         stopping,
                         tg,
+                        self._transactor,
                     )
                     log.event.status("Capture started")
                     await capture(task)
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -17,6 +17,7 @@ from typing import (
     TypeVar,
     cast,
 )
+from typing_extensions import override
 
 from pydantic import AwareDatetime, BaseModel, Field, NonNegativeInt
 
@@ -38,6 +39,7 @@ from ..flow import (
 from ..pydantic_polyfill import GenericModel
 from ..utils import json_merge_patch
 from . import Task, request, response
+from .webhook.match import CollectionMatchingSpec, UrlMatch
 
 LogCursor = tuple[str | int] | AwareDatetime | NonNegativeInt
 """LogCursor is a cursor into a logical log of changes.
@@ -189,6 +191,21 @@ class ResourceConfig(BaseResourceConfig):
 
 
 _ResourceConfig = TypeVar("_ResourceConfig", bound=ResourceConfig)
+
+
+class WebhookResourceConfig(BaseResourceConfig):
+    PATH_POINTERS: ClassVar[list[str]] = ["/name"]
+
+    name: str = Field(description="Name of this resource")
+    match_rule: CollectionMatchingSpec = Field(
+        default_factory=lambda: UrlMatch(value="*"),
+        description="Matching spec for routing incoming webhooks to this collection",
+        discriminator="type",
+    )
+
+    @override
+    def path(self) -> list[str]:
+        return [self.name]
 
 
 CRON_REGEX = (
@@ -473,6 +490,7 @@ class Resource(Generic[_BaseDocument, _BaseResourceConfig, _BaseResourceState]):
     schema_inference: bool
     reduction_strategy: ReductionStrategy | None = None
     disable: bool = False
+    requires_explicit_acks: bool = False
 
 
 @dataclass(kw_only=True)
@@ -558,8 +576,6 @@ def resolve_bindings(
     errors: list[str] = []
 
     # Check for routing conflicts between webhook match rules.
-    from .webhook.resources import WebhookResourceConfig
-
     webhook_match_rules = [
         resource.initial_config.match_rule
         for resource in resources
@@ -751,7 +767,11 @@ def open(
             # TaskGroup stays alive until graceful shutdown begins.
             await task.stopping.event.wait()
 
-    return (response.Opened(explicitAcknowledgements=False), _run)
+    requires_explicit_acks = any(
+        rsc.requires_explicit_acks for _, rsc in resolved_bindings
+    )
+
+    return (response.Opened(explicitAcknowledgements=requires_explicit_acks), _run)
 
 
 def open_binding(
--- a/estuary-cdk/estuary_cdk/capture/task.py
+++ b/estuary-cdk/estuary_cdk/capture/task.py
@@ -23,7 +23,7 @@ from ..flow import (
 )
 from ..pydantic_polyfill import GenericModel
 from . import request, response
-from ._emit import emit_from_buffer
+from .transactor import Transactor
 
 
 class Request(GenericModel, Generic[EndpointConfig, ResourceConfig, ConnectorState]):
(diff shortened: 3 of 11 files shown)


estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -730,7 +749,11 @@ async def _run(task: Task):
             # TaskGroup stays alive until graceful shutdown begins.
             await task.stopping.event.wait()
 
-    return (response.Opened(explicitAcknowledgements=False), _run)
+    requires_explicit_acks = any(
+        rsc.is_webhook_resource for _, rsc in resolved_bindings
+    )
```
Alex-Bair, line 772: Making explicit acks only possible when there's a webhook resource enabled is what we currently need and want, but I'd suggest de-coupling how we do that a bit so it's possible to require acknowledgements for non-webhook resources. I'm thinking something like:
1. Adding a `requires_explicit_acknowlegements` field to the [`Resource` dataclass](https://github.com/estuary/connectors/blob/7c1f847308118d6fec099003d33f9c09ef175d8c/estuary-cdk/estuary_cdk/capture/common.py#L453-L496) that defaults to `False`.
2. Create a `WebhookResource` class that subclasses `Resource` and defaults `requires_explicit_acknowledgements` to `True` (and possible any other defaults that seem sensible?). `WebhookResource` would be a convenience class, similar to `SnapshotResource`.
3. Key off `resource.requires_explicit_acknowledgements` when determining whether or not explicit acknowledgements are required.

How does that sound to you?

estuary-cdk/estuary_cdk/capture/base_capture_connector.py
```diff
@@ -58,6 +59,12 @@ class BaseCaptureConnector(
 ):
     output: BinaryIO = sys.stdout.buffer
 
+    _transactor: Transactor | None = None
+    """Set on `open` when the capture has at least one binding requiring
+    Flow's explicit Acknowledge protocol. Forwards Acknowledge messages from
+    Flow to the capture's Tasks so webhook checkpoints can block until
+    persistence is confirmed."""
```
Alex-Bair: What would it look like if `_transactor` were never `None`? There are a lot of `if self._transactor is not None` scattered throughout the code since we're using the presence of `self._transactor` to determine if we need to perform the checkpoint accounting and acknowledgement awaiting. IMO it'd be conceptually cleaner to separate those two concerns by always having a `_transactor` present, then using something else, like a `_tracks_acks` field on `Transactor`, to determine whether or not to perform that accounting & awaiting.

estuary-cdk/estuary_cdk/capture/task.py
```diff
@@ -169,14 +187,55 @@ def sourced_schema(self, binding_index: int, schema: dict[str, Any]):
...
+            await emit_from_buffer(self._buffer, self._output)
+            self.reset()
+
+            return
+
+        completion: asyncio.Future[None] | None = None
+        if self.requires_ack:
+            completion = asyncio.get_running_loop().create_future()
+
+        await self.transactor.commit(self._buffer, completion)
```
Alex-Bair: nit: what do you think about having `transactor.commit` mint and return the `Future` instead? That could tighten up this check here & pushes the responsibility of determining "what do we do when we require acknowledgements?" onto the transactor.

estuary-cdk/estuary_cdk/capture/task.py
```diff
@@ -169,14 +187,55 @@ def sourced_schema(self, binding_index: int, schema: dict[str, Any]):
...
+        self.reset()
```
Alex-Bair: Making the `transactor` always present like I mentioned in this [comment](https://github.com/estuary/connectors/pull/4355/changes/7c1f847308118d6fec099003d33f9c09ef175d8c#r3196628557) would let `Task.checkpoint` collapse to a single path here: write the checkpoint, hand the buffer to `transactor.commit`, and optionally await the returned `Future`. The "do we need to wait for this checkpoint to be ACK'd?" decision moves into the `Transactor` itself via a new parameter (ex: `transactor.commit(self._buffer, await_ack=True)`). In combination with the helper mentioned in this comment, this section could look like:
```python
checkpoint = Response[Any, Any, ConnectorState](
    checkpoint=response.Checkpoint(                                                                                                                                                                          
        state=ConnectorStateUpdate(updated=state, mergePatch=merge_patch)                                                                                                                                    
    )
)                                                                                                                                                                                                            
self._buffer.write(    
    checkpoint.model_dump_json(by_alias=True, exclude_unset=True).encode()
)                                                                                                                                                                                                            
self._buffer.write(b"\n")
                                                                                                                                                                                                             
completion = await self._drain_buffer_into(                                                                                                                                                                  
    self.transactor, await_ack=self.requires_ack
)                                                                                                                                                                                                            

```

What are your thoughts on that approach instead?

estuary-cdk/estuary_cdk/capture/task.py
```diff
@@ -219,12 +280,3 @@ async def run_task(parent: Task):
...
         return task
-
-    async def _emit(self, response: Response[EndpointConfig, ResourceConfig, ConnectorState]):
-        self._buffer.write(
-            response.model_dump_json(by_alias=True, exclude_unset=True).encode()
-        )
-        self._buffer.write(b"\n")
-        self._buffer.seek(0)
-        await emit_from_buffer(self._buffer, self._output)
-        self.reset()
```
Alex-Bair, line 230: nit: it would be nice to consolidate the various buffer draining interactions in a single helper function, and while we can't keep using the `_emit` method as it is if all emissions will start going through the `transactor`, I think having a helper consolidate a lot of our interactions with a `Task`'s buffer is helpful. Maybe something like the below would work?
```python
async def _drain_buffer_into(
    self, transactor: Transactor, await_ack: bool
) -> asyncio.Future[None] | None:
    """Drain everything in this Task's buffer through `transactor`, then
    reset the buffer to accept the next checkpoint's bytes."""
    self._buffer.seek(0)
    completion = await transactor.commit(self._buffer, await_ack=await_ack)
    self._buffer.truncate(0)
    self._buffer.seek(0)
    self._hasher.reset()
    return completion
```

Alex-Bair (commented):
Looks pretty good! Most of my comments were around trying to simplify the code so it's easier to reason about & maintain. My main ask is that instead of conditionally creating the `transactor` when there's a resource that requires acknowledgements, it'd be simpler to always create the transactor, always perform emissions through it, and have it return a `Future` for the `Task` to await only when an acknowledgement is required.

Let me know if I was unclear anywhere; we can always discuss during our 1:1s too.

estuary-cdk/estuary_cdk/capture/common.py
nicolaslazo: I like the idea, it's an elegant way to both do polymorphism and simplify the webhook resource definition interface in one move 🤔

estuary-cdk/estuary_cdk/capture/base_capture_connector.py
nicolaslazo: We discussed this on our 1:1, it's addressed 👍

estuary-cdk/estuary_cdk/capture/task.py
nicolaslazo: I couldn't inline `Task.reset()` as it's used in other places, but I extracted all the other checkpoint-related, byte level operations to a helper function. `Task.checkpoint()` now only concerns itself with calling it and waiting on any returned `Future`s

estuary-cdk/estuary_cdk/capture/webhook/resources.py
```diff
@@ -85,7 +84,7 @@ def open_webhook_binding(
         webhook_bindings = {
             idx: rsc
             for idx, (_, rsc) in enumerate(all_bindings)
-            if isinstance(rsc.initial_config, WebhookResourceConfig)
+            if rsc.requires_explicit_acks
```
Alex-Bair: Since this check is now looking at `rsc.requires_explicit_acks`, it's possible that a non-webhook binding that requires explicit acknowledgements to be included in this set. Is there a way we can distinguish webhook vs. non-webhook bindings, like with `isinstance(rsc, WebhookResource)` or something similar?

estuary-cdk/estuary_cdk/capture/task.py
```diff
@@ -220,11 +271,61 @@ async def run_task(parent: Task):
         task.set_name(child_name)
         return task
 
-    async def _emit(self, response: Response[EndpointConfig, ResourceConfig, ConnectorState]):
-        self._buffer.write(
-            response.model_dump_json(by_alias=True, exclude_unset=True).encode()
+
+class ConcurrentTask:
```
Alex-Bair: Was `ConcurrentTask` suppose to be included in this PR? I thought it was getting introduced in a separate PR. If you want to include it in this one, LMK, but I held off reviewing the related code for now since I suspect it might have slipped through accidently during a rebase.

Alex-Bair (commented):
Thanks for addressing my feedback! I had two more questions after reviewing again. I held off approving since I _think_ `ConcurrentTask` is suppose to be in your other PR, not this one. But if you do want to include it here, LMK, and I can review the relevant code in this PR.

estuary-cdk/estuary_cdk/capture/webhook/resources.py
nicolaslazo: Good catch!

estuary-cdk/estuary_cdk/capture/task.py
nicolaslazo: Ugh, in the rush to fix the botched rebase I accidentally let this in. It's fixed

Alex-Bair (approved):
LGTM!

PR #4357, opened 2026-05-04
materialize-bigtable: new connector

### LINKED ISSUE #2918: stats-next: new bigtable materialization

williamhbaker:
**Description:**

New `materialize-bigtable` connector. Materializes Flow collections to Google Cloud Bigtable as standard updates.

A few interesting design choices to be aware of:

#### Exactly-once via MVCC
The connector stamps every cell write with a monotonically increasing counter as the cell timestamp. The column family is created with `MaxVersions(2)`, so a crashed transaction's "dirty" cell is preserved alongside the prior committed value. Loads request the latest 2 versions and walk newest-first, skipping any cell at the in-flight timestamp—this surfaces the last *committed* value even if a previous Store partially landed. 
> **Note:** This relies on idempotent runtime transactions; without them, reductions other than LWW are incorrect on retry.

#### Row keys are FDB-packed tuples
These are used directly as the Bigtable row key, which preserves lexicographic ordering. This enables efficient range scans with appropriate composite key structures.

#### Value Encoding
Everything is stored as bytes to match Bigtable's native model:

| Type | Encoding | Notes |
| :--- | :--- | :--- |
| **Booleans** | `0x00` / `0x01` | |
| **Integers** (int64) | 8-byte big-endian | Bigtable's spec'd format. Enables atomic increment via `ReadModifyWrite` (unused here). |
| **Integers** (wider than int64) | Decimal text (e.g. `"99999999999999999999"`) | Used when schema inference indicates the range or string length exceeds int64. |
| **Numbers** (floats) | 8-byte big-endian IEEE 754 | Not formally Bigtable-specified, but matches standard tool behavior. |
| **Null** | Empty byte slice | Ambiguous with zero-length string/binary, but unlikely to matter in practice. |
| **Strings**  | UTF-8 bytes        | Passed through. |
| **Binary**   | Raw bytes          | Base64-decoded from JSON. |
| **Arrays / objects / multi-typed** | JSON bytes | Stored as the original JSON encoding. |

#### No delta-updates mode
There is no meaningful semantic for a key/value store; same as `materialize-dynamodb`.

Closes https://github.com/estuary/flow/issues/2918

**Workflow steps:**

See https://github.com/estuary/flow/pull/2919

**Notes for reviewers:**


materialize-bigtable/type_mapping.go
```diff
@@ -0,0 +1,300 @@
...
+	out := make([]byte, 8)
+	switch v := te.(type) {
+	case int:
+		binary.BigEndian.PutUint64(out, uint64(int64(v)))
+	case int64:
+		binary.BigEndian.PutUint64(out, uint64(v))
+	case uint64:
+		binary.BigEndian.PutUint64(out, v)
+	case float64:
+		binary.BigEndian.PutUint64(out, uint64(int64(v))) // Truncate trailing zero fraction
```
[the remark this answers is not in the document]
williamhbaker: Fields with `type: integer` are known to not contain the nan / inf values, otherwise they'd be `type: number`.

materialize-bigtable/testdata/apply.flow.yaml
```diff
@@ -0,0 +1,19 @@
...
+  # See README.md for instructions on running against a real Cloud Bigtable
+  # instance.
+
+  # acmeCo/tests/materialize-bigtable-gcp:
+  #   endpoint:
+  #     local:
+  #       command: ["go", "run", "."]
+  #       protobuf: true
+  #       config: config.gcp.yaml
+  #   bindings: []
```
jacobmarble: Would you mind following this pattern?

https://github.com/estuary/connectors/blob/2f4876c60954ca207fbf7c4517e799f214b7c83e/materialize-iceberg/driver_test.go#L159-L168

FWIW I'm using the pattern in the EventBridge PR as well, much simpler though: https://github.com/estuary/connectors/pull/4359

materialize-bigtable/transactor.go
```diff
@@ -0,0 +1,327 @@
...
+	// Usually the default `flow_document` but a user may project the root
+	// document under an alternate name.
+	docField string
+}
+
+type state struct {
+	// CommittedTimestamp is the last successfully committed cell timestamp;
+	// the timestamp stamped on writes for the in-progress transaction is
+	// always `CommittedTimestamp + 1`.
+	CommittedTimestamp uint64 `json:"committed_timestamp"`
```
jacobmarble: How about `int64` to exactly match `bigtable.Timestamp`?

jacobmarble (approved):
Solid!

materialize-bigtable/testdata/apply.flow.yaml
williamhbaker: Cool, that is simpler! Updated.

PR #4358, opened 2026-05-04
estuary-cdk: add if/then/else schema block to support hard deletes

Alex-Bair:
**Description:**

Standard-mode materializations issue real `DELETE`s only when the collection schema marks `op="d"` docs to reduce as `merge + delete: true`, producing a tombstone the materialization translates to a `DELETE`, while letting other docs reduce normally. JSON Schema's if/then/else is how Flow expresses that per-document reduction switch on a single schema; the Go SQL framework already injects this block during discovery, but the Python CDK did not, so connectors emitting `_meta/op="d"` produced soft-delete markers instead of hard deletes.

This PR adds that if/then/else block to all discovered resources' schemas.

The changes to connectors' discover snapshots are expected due to the addition of the if/then/else to support hard deletes.

**Workflow steps:**

Dataflows that have a CDK-based capture that was previously emitting deletions (`_meta/op = "d"`) and a materialization with `hardDelete` set to `true` will begin using hard deletes instead of soft deletes.

Note: This change only affects dataflows that emit deletions, like many native connector's full refresh streams that take a snapshot of the entire data set in the source system. Dataflows where the capture side does not emit deletions are not affected.

**Notes for reviewers**

After this PR is merged into `main`, I'll open a PR for `source-netsuite-v2` that pulls in the latest CDK changes, including the if/then/else bit in discovered resource schemas added in this PR.

I confirmed via `flowctl raw combine` that the new if/then/else schema block correctly routes documents with `_meta.op == "d"` to the then branch and all other documents, including those missing `_meta` or `_meta/op`, to the else branch, with reductions applied as specified on each side.

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -503,8 +503,29 @@ def discovered(
         if resource.schema_inference:
             schema["x-infer-schema"] = True
 
+        # Materializations require this if/then/else conditional to support
+        # hard deletes. It tells Flow to use a top-level merge reduction
+        # strategy for deletions and mark them with `delete: True` so materializations
+        # know to issue DELETEs to destinations.
+        #
+        # `_meta` and `_meta.op` are both `required` in the `if` clause because
+        # JSON Schema's `properties` keyword passes when a key is absent.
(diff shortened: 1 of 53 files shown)


nicolaslazo (approved):
LGTM!

PR #4359, opened 2026-05-04
materialize-eventbridge: new connector

### LINKED ISSUE #4262: Request a connector to materialize to Eventbridge
### System Name

Amazon Eventbridge

### Type

Materialize

### Details

A large customer wants to push event data to Eventbridge.  This should be a delta updates connector and at least once is likely good enough.

jacobmarble:
**Description:**

New `materialize-eventbridge` connector — Pattern 4 (push-only, delta-updates) materialization to AWS EventBridge.

- One bus per endpoint; each binding sets `Source` + `DetailType`. The full materialized document is the event `Detail` payload.
- Auth: discriminated `credentials` block — `AWSAccessKey` (static keys) or `AWSIAM` (assume-role + STS, same pattern as `materialize-dynamodb`).
- `Apply` verifies the bus via `DescribeEventBus`; at-least-once delivery; 256 KB per-entry limit enforced client-side.
- Implements `boilerplate.Materializer` (the other Pattern-4 connectors hand-roll), so it runs on the standard `bptest` rig and produces the same snapshot files SQL materializers do.

Closes #4262.

**Workflow steps:**

User picks region, bus name, and auth mode; adds bindings with `source` + `detail_type`. Documents are published via `PutEvents` batched 10/call with bounded `errgroup` parallelism (`storeConcurrency=8`); partial-failure responses are classified per-entry — throttle codes retry with exponential backoff, permanent codes fail-fast.

User docs for `materialize-eventbridge` will follow in a separate doc PR.

**Notes for reviewers:**

- LocalStack tests run in CI by default; AWS-side tests pass locally with `-eventbridge.test-all` (resources documented in `materialize-eventbridge/testdata/README.md` — account `076183946664`, region `us-east-2`).
- Endpoint resolver uses the deprecated `aws.EndpointResolverWithOptionsFunc` for consistency with four other AWS-touching connectors; deferred to a repo-wide migration.


materialize-eventbridge/driver.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	var constraint pm.Response_Validated_Constraint
+	switch {
+	case p.IsRootDocumentProjection():
+		constraint.Type = pm.Response_Validated_Constraint_LOCATION_REQUIRED
+		constraint.Reason = "The root document must be materialized"
+	case p.IsPrimaryKey:
+		constraint.Type = pm.Response_Validated_Constraint_LOCATION_REQUIRED
+		constraint.Reason = "Document keys must be included"
```
jacobmarble: from @mdibaiee:
> I believe keys should also be forbidden since we are doing the full document and there is no keying of the events at all?
>
> Looking at google-pubsub it seems we have it as Optional there, but that’s because the key can be used to order the messages published, but here it seems we don’t use the key at all

I did try to switch primary key to forbidden, but the integration tests fail because the flow runtime requires key fields not be forbidden, so the best we can do is optional.

---

I've checked the other pattern-4 materializers:
- slack requires and uses keys
- pinecone requires and uses keys
- webhook requires, but does not use keys, only `RawJSON`
- google-pubsub makes keys optional, and uses `.PackedKey`

[Notably, this comment tags the pinecone logic](https://github.com/estuary/connectors/blob/main/materialize-pinecone/driver.go#L204-L210):
> We require collection keys be materialized because it seems pretty reasonable to
> require they be included as metadata since the composite key is used as the basis for
> the vector ID, and also to avoid complications from
> https://github.com/estuary/flow/issues/1057.

materialize-eventbridge/transactor.go
```diff
@@ -0,0 +1,190 @@
...
+		batch = nil
+		errGroup.Go(func() error {
+			return t.putEvents(ctx, entries)
+		})
+	}
+
+	for it.Next(false) {
+		b := t.bindings[it.Binding]
+
+		if len(it.RawJSON) > putEventsMaxEntrySize {
```
jacobmarble: from @mdibaiee:
> Does this limit include the source and detail_type as well?

No, let's fix that.

Some important docs, which contradict each other re per-event limits:
- https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEvents.html
  - > You can batch multiple event entries into one request for efficiency. However, the total entry size must be less than 1MB.
- https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-putevents.html#eb-putevent-size
  - > This limit applies to the request as a whole, not to individual entries. A single event can use up to the full 1 MB if it is the only entry in the request. You can calculate the entry size before you send the events.
- https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/eventbridge#Client.PutEvents
  - > You can batch multiple event entries into one request for efficiency. However, the total entry size must be less than 256KB.

What I've done:
- limit the entire batch size to 1MB
- limit each event to 256KB
- formulate the size properly
- keep the 10 events-per-batch limit, as this is a service-side hard limit

materialize-eventbridge/driver.go
```diff
@@ -0,0 +1,389 @@
...
+		), nil
+	case AWSIAM:
+		return c.Credentials.IAMTokens.AWSCredentialsProvider()
+	}
+	return nil, fmt.Errorf("unknown %q: %s", "auth_type", c.Credentials.AuthType)
+}
+
+type resource struct {
+	Source     string `json:"source" jsonschema:"title=Event Source,description=Source field set on every event published from this binding (e.g. \"my.app\")." jsonschema_extras:"x-collection-name=true,order=1"`
+	DetailType string `json:"detail_type" jsonschema:"title=Detail Type,description=DetailType field set on every event published from this binding (e.g. \"OrderPlaced\")." jsonschema_extras:"order=2"`
```
jacobmarble: from @mdibaiee:
> Seems like this is going to be a painful connector to configure for customers because they need to manually set `detail_type` for each binding, which could be a lot... Worth checking with product people what they think we should do here for a better experience: a default? some other form of inference?

I've added `jsonschema_extras:"x-collection-name=true"` to the `DetailType` field to signal default value == collection name in the UI.

materialize-eventbridge/driver.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	return nil
+}
+
+func (d *materialization) CheckPrerequisites(ctx context.Context) *cerrors.PrereqErr {
+	prereqs := &cerrors.PrereqErr{}
+	if err := describeBus(ctx, d.client, d.cfg.EventBusName); err != nil {
+		prereqs.Err(err)
+	}
+	return prereqs
```
jacobmarble, line 283: from @mdibaiee:
> We need to also check we have write permission for the bus

How do you propose we perform this check? I can think of two ways.

1) Send a test event. This works, but creates clutter. Should we clean the clutter ourselves? Create a dead queue for the clutter? Ignore it?

2) Invoke `sts.NewFromConfig(cfg).GetCallerIdentity(...)` and `iam.NewFromConfig(cfg).SimulatePrincipalPolicy(...)` to directly query for the needed permission. This is elegant, but (1) has no precedent in the repository, (2) there are edge cases like assumed-role ARNs returned by `GetCallerIdentity` must be converted to the underlying role ARN (3) the role needs permission to call `GetCallerIdentity` and `SimulatePrincipalPolicy`.

materialize-eventbridge/driver.go
jacobmarble: 59f526295

materialize-eventbridge/driver.go
jacobmarble: 3ce45279e

materialize-eventbridge/transactor.go
jacobmarble: ad28a8b32

materialize-eventbridge/driver.go
mdibaiee: @jacobmarble let's ask a potential / first customer of this connector what they think about having the collection name as the DetailType

materialize-eventbridge/driver.go
mdibaiee: Sending a test event is off the table: our Validate step must be fully read-only
I think the second option is a good best-effort check, doesn't need to be exhaustive

jacobmarble: 90d03436c

materialize-eventbridge/driver.go
```diff
@@ -387,9 +395,67 @@ func (d *materialization) SnapshotTestResource(ctx context.Context, path []strin
...
+func checkPutEventsPermission(ctx context.Context, cfg config, busARN string, prereqs *cerrors.PrereqErr) {
+	awsCfg, err := awsConfigFor(ctx, cfg)
+	if err != nil {
+		log.WithField("error", err).Warn("could not build AWS config for permission check; skipping")
+		return
+	}
+
+	callerOut, err := sts.NewFromConfig(awsCfg).GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+	if err != nil {
+		log.WithField("error", err).Warn("GetCallerIdentity failed; skipping write-permission check")
```
mdibaiee, line 415: Shouldn't this also be a cause for failure? Is it possible not to have this permission, but still have the permission to write?

materialize-eventbridge/driver.go
```diff
@@ -387,9 +395,67 @@ func (d *materialization) SnapshotTestResource(ctx context.Context, path []strin
...
+	}
+
+	principalARN := iamPrincipalARN(aws.ToString(callerOut.Arn))
+	simOut, err := awsiam.NewFromConfig(awsCfg).SimulatePrincipalPolicy(ctx, &awsiam.SimulatePrincipalPolicyInput{
+		PolicySourceArn: &principalARN,
+		ActionNames:     []string{"events:PutEvents"},
+		ResourceArns:    []string{busARN},
+	})
+	if err != nil {
+		log.WithField("error", err).Warn("SimulatePrincipalPolicy failed; skipping write-permission check")
```
mdibaiee, line 426: Same question here

mdibaiee (approved):
LGTM % question regarding the check

materialize-eventbridge/driver.go
jacobmarble: > Sending a test event is off the table: our Validate step must be fully read-only
> I think the second option is a good best-effort check, doesn't need to be exhaustive

My understanding of this comment is to log a warning if the caller lacks permission-to-check-permission. Just confirmed the behavior with a new unit test in be249b0f8.

FWIW this connector does invoke `DescribeEventBus`, which doesn't check for write permission, but does confirm read permission (better than nothing). Also the S3 connectors don't explicitly check for object creation permissions AFAICT.

materialize-eventbridge/driver.go
jacobmarble: I've changed the default values for both source and detail type after brainstorming with a customer. e1fad958b80687d9d357803f67b7177086ecf6ae

PR #4360, opened 2026-05-04
source-postgres: Add skip_prerequisites and skip_table_prerequisites

willdonnelly:
**Description:**

Adds two feature flags as escape hatches for cases where the connector's prerequisite checks should be skipped.

The immediate motivation here is a production capture where the prerequisite checks appear to be taking approximately an hour and a half to run. This is anonymously slow, and we should figure out why it is so slow, but it would be very useful to be able to just say "skip the checks" for now.


Alex-Bair (approved):
LGTM

PR #4361, opened 2026-05-04
materialize-(s3-)iceberg: clamp out-of-range dates and timestamps

jacobmarble:
**Description:**

Fixes a production failure in materialize-s3-iceberg where committing a transaction crashes with `OverflowError: date value out of range`. `pyiceberg.add_files` reads the parquet file's column min/max stats through pyarrow, which materializes them as Python `datetime` — so any year > 9999 or < 1 (well within Iceberg's INT64-micros / INT32-days range) aborts the append.

The fix is a small clamp in the Store loop that rewrites out-of-range timestamp/date strings to the appropriate boundary before they reach the parquet writer. materialize-iceberg routes through Spark/Iceberg-Java and is unaffected; this PR adds parallel regression tests there as a baseline.

**Workflow steps:**

No user-visible changes. Out-of-range values are clamped to year bounds the rest of the Python ecosystem can read.

None.

**Notes for reviewers:**

- 3 commits: mat-iceberg tests (baseline), s3-iceberg tests (fail without clamp), s3-iceberg clamp (tests pass).
- Clamp is connector-local; shared `go/writer` is untouched.
- `clampTimestamp` compares `t.UTC().Year()` not `t.Year()` — "9999-12-31T23:59:59-14:00" looks like 9999 locally but reaches parquet as year 10000.
- Tests require docker compose; both regression subtests pass.

### DIFF
--- a/materialize-s3-iceberg/clamp.go
+++ b/materialize-s3-iceberg/clamp.go
@@ -6,6 +6,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/estuary/flow/go/protocols/fdb/tuple"
 )
 
 // Bounds of Python `datetime`. pyiceberg.add_files reads parquet column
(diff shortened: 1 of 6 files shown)


materialize-s3-iceberg/transactor.go
```diff
@@ -174,6 +174,25 @@ func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
 			row = append(row, it.RawJSON)
 		}
 
+		// Clamp out-of-range timestamps/dates so pyiceberg.add_files can
+		// read the parquet column stats without overflowing Python datetime.
+		for i, v := range row {
```
danielnelson: Would it make sense for the clamping function to be an `ElementConverter` and return it from `func (d *materialization) MapType(`?

materialize-s3-iceberg/clamp.go
```diff
@@ -0,0 +1,63 @@
...
+	case y < pyMinYear:
+		return time.Date(pyMinYear, 1, 1, 0, 0, 0, 0, time.UTC).Format(time.RFC3339Nano), true
+	}
+	return "", false
+}
+
+// clampDate also rescues > 4-digit years, which time.Parse(time.DateOnly, ...)
+// refuses to parse at all.
+func clampDate(s string) (string, bool) {
+	m := dateLikeRe.FindStringSubmatch(s)
```
danielnelson, line 58: Presumably clampTimestamp(...) also has the issue where it can't handle >4-digit years too?  It may not be an issue that we should even handle though, I believe the date and date-time types are defined to be rfc3339 which does not allow years greater than 9999.

danielnelson (approved):

materialize-s3-iceberg/transactor.go
jacobmarble: I ass-u-me'd that `ElementConverter` was a `materialize-sql` thing, but there are two:
- `materialize-sql`: `type ElementConverter func(tuple.TupleElement) (interface{}, error)`
- `materialize-boilerplate`: `type ElementConverter func(tuple.TupleElement) (any, error)`

Similarly, there are two `ConvertAll()` methods, which are both key to using the respective `ElementConverter`. Now the clamping functions implement the boilerplate `ElementConverter` function.

d28d6c5aa

materialize-s3-iceberg/clamp.go
jacobmarble: Indeed, the sql, SingleStore and ClickHouse timestamp clamps all return error on RFC3339 parse failure, so we should do the same here. I've also improved test coverage of these clamp functions.

7152653cd

PR #4372, opened 2026-05-06
estuary-cdk: implement MultipleWritersTask for concurrent webhook writes

### LINKED ISSUE #4322: estuary-cdk: instantiate independent `Task`s for webhook streams
Unlike pull API streams, webhook streams are currently designed to reuse the same `Task` instance in a capture. This is suboptimal for two reasons:

- Streams must acquire a [mutex](https://github.com/estuary/connectors/blob/main/estuary-cdk/estuary_cdk/capture/webhook/server.py#L96-L134) to write to the shared buffer
- Since all streams have to share the same limited memory space, this is likelier to overflow

Quoting @Alex-Bair 's comment:

> aiohttp runs request handlers concurrently, but every invocation of
> webhook_handler shares the same Task, and therefore the same underlying
> buffer (Task._buffer) and hasher. task.captured() writes to that buffer
> synchronously, and task.checkpoint() awaits an emit that seeks to 0,
> reads the buffer from a worker thread, then truncates it. Without this
> lock, two concurrent requests can corrupt each other in at least two
> ways:
> 
>   1. Silent data loss. R1 enqueues docs and awaits checkpoint(); while
>      R1 is inside _emit, R2 enqueues its own docs via captured(). When
>      R1's emit finishes and calls reset(), R2's queued docs are
>      truncated away. R2 then checkpoints an empty buffer, returns 200
>      to the webhook sender, and Flow never sees those documents.
> 
>   2. Corrupt stdout. While _copy_buffer is reading the buffer from a
>      worker thread (after seek(0)), another request's captured() call
>      writes to the same file object from the event-loop thread.
>      SpooledTemporaryFile has a single shared file position and is not
>      thread-safe, so this can produce partial/interleaved JSON on
>      stdout. The Flow runtime consumes that stream, and invalid JSON
>      there has blast radius beyond a single capture task.
> 
> Holding a lock around captured() + checkpoint() serializes only the
> emit path; JSON parsing, match-rule evaluation, and document enrichment
> above still run concurrently across requests. That is fine for typical
> webhook loads but will bottleneck sustained high-throughput workloads.
> 
> PROPER FIX: mirror how pull-API bindings work (see _binding_incremental_task
> et al. in common.py) — give each webhook binding its own child Task via
> task.spawn_child, each with its own buffer. webhook_handler would group
> enriched_docs by doc.binding, hand each group to that binding's child
> Task, and checkpoint each touched binding under a per-binding lock.
> Requests targeting different bindings would then emit concurrently
> without ever sharing a buffer, removing this class of bug structurally
> instead of patching it. We should fix this whenever we touch this code next.

nicolaslazo:
**Description:**

The current Task implementation gives callers manual control over how
many documents are buffered before each checkpoint. That model works
when there is a single document producer per Task, but our aiohttp
webhook server handles an arbitrary number of concurrent requests across
the the same Task; this exposes us to the risk that concurrent
checkpoint() calls may interleave on the shared buffer.

This change introduces a MultipleWritersTask class with an `emit()`
method that atomically writes buffers and checkpoints together.

MultipleWritersTask also gives each binding its own buffer, replacing
the previous stopgap where a single mutex was shared across every
binding in the webhook server.

Closes #4322.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/task.py
+++ b/estuary-cdk/estuary_cdk/capture/task.py
@@ -55,7 +55,7 @@ def orjson_default(obj):
     if isinstance(obj, decimal.Decimal):
         return str(obj)
     if isinstance(obj, (bytes, bytearray)):
-        return base64.b64encode(obj).decode('utf-8')
+        return base64.b64encode(obj).decode("utf-8")
     raise TypeError
 
 
@@ -147,16 +147,21 @@ class Task:
         Or, reset() will discard any queued documents."""
 
         if isinstance(document, dict):
-            b = orjson.dumps({
-                "captured": {
-                    "binding": binding,
-                    "doc": document,
-                }
-            }, default=orjson_default)
+            b = orjson.dumps(
+                {
+                    "captured": {
+                        "binding": binding,
+                        "doc": document,
+                    }
+                },
+                default=orjson_default,
+            )
         else:
-            b = Response(
-                captured=response.Captured(binding=binding, doc=document)
-            ).model_dump_json(by_alias=True, exclude_unset=True).encode()
+            b = (
+                Response(captured=response.Captured(binding=binding, doc=document))
+                .model_dump_json(by_alias=True, exclude_unset=True)
+                .encode()
+            )
 
         self._buffer.write(b)
         self._buffer.write(b"\n")
@@ -172,9 +177,15 @@ class Task:
         """Write a SourcedSchema message for the given binding to the buffer.
         SourcedSchema messages won't be emitted until checkpoint() is called."""
 
-        b = Response(
-            sourcedSchema=response.SourcedSchema(binding=binding_index, schema_json=schema)
-        ).model_dump_json(by_alias=True, exclude_unset=True).encode()
+        b = (
+            Response(
+                sourcedSchema=response.SourcedSchema(
+                    binding=binding_index, schema_json=schema
+                )
+            )
+            .model_dump_json(by_alias=True, exclude_unset=True)
+            .encode()
+        )
 
         self._buffer.write(b)
         self._buffer.write(b"\n")
--- a/estuary-cdk/estuary_cdk/capture/task.py
+++ b/estuary-cdk/estuary_cdk/capture/task.py
@@ -3,7 +3,7 @@ import base64
 import decimal
 import tempfile
 import traceback
-from collections.abc import Awaitable
+from collections.abc import Awaitable, Iterable
 from dataclasses import dataclass, field
 from logging import Logger
 from typing import Any, BinaryIO, Callable, Generic
@@ -13,6 +13,7 @@ import xxhash
 from pydantic import Field
 
 from estuary_cdk.capture.connector_status import ConnectorStatus
+from estuary_cdk.capture.document import AssociatedDocument
 
 from ..flow import (
     ConnectorSpec,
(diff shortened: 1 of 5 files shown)


Alex-Bair (commented):
I know I originally provided the direction of instantiating separate `Task`s for each webhook binding, but the more I think about the problem we're trying to solve plus seeing that the existing `source-appsflyer` captures aren't throughput limited by sharing a single `Task`, I'd rather we not take the "one `Task` per webhook binding" approach right now; it's an optimization I thought we'd need, but based on empirical evidence, we don't need to take on that additional complexity & improve throughput here yet. I apologize for changing my mind & switching direction late on this one. :bow:

The messy bit remaining to clean up is ensuring a `Task` doesn't attempt to write to its buffer while its buffer is being drained. We're using locks to coordinate this whenever we're using the various `Task` methods, but IMO it'd be cleaner to push this complexity down into the buffer itself; if the buffer guaranteed no write would ever happen when it's being drained, that'd simplify a lot of the logic elsewhere in the CDK now & in the future.

So, what I'm proposing is we:
- still use a single `Task` for all webhook bindings
- wrap the `Task`'s buffer in a `ConcurrentBuffer` class whose writes and drains are safe for concurrent producers. A `ConcurrentBuffer` would be an overall improvement for all connectors, albeit it's an improvement that wasn't needed until we started capturing webhooks.

I've pushed a spike implementation of this up to the [`bair/estuary-cdk-concurrent-buffer-spike`](https://github.com/estuary/connectors/compare/main...bair/estuary-cdk-concurrent-buffer-spike) branch for reference of what I'm thinking. Do you think that's a suitable alternative or see something I'm missing?

nicolaslazo:
We've discussed this over VC but, to summarise our conversation:

I like the idea of a `ConcurrentBuffer` class that encapsulates the tightly-coupled byte buffer and hasher. It could get polymorphically swapped for a `Buffer` alternative, depending on whether we need an internal lock to guarantee mutual exclusivity or it's guaranteed by design. Unfortunately, the buffer level does not have the degree of oversight to guarantee that webhook captures can atomically drain documents and then checkpoint. We are not thrilled about adding more nuance to the `Task` class, but it feels like the correct layer at which to branch the logic out. I'll keep thinking of more elegant alternatives.

estuary-cdk/estuary_cdk/capture/task.py
```diff
@@ -270,3 +270,86 @@ async def run_task(parent: Task):
         task = self._tg.create_task(run_task(self))
         task.set_name(child_name)
         return task
+
+
+class AssociatedDocument(Protocol):
+    binding: int
+    doc: Any
```
Alex-Bair: There's already an `AssociatedDocument` class in `capture/common.py`, and it looks like we're redefining a same-named protocol here to avoid a circular import with `from .common import AssociatedDocument`, is that right? Having two separate classes with the same name & purpose is a bit confusing and dances around the real problem of import paths. I think it'd be better to instead re-organize the CDK to make it possible to import `AssociatedDocument` in `task.py` without causing circular import issues. I suggest we do this by extracting the classes and types around documents (so `BaseDocument`, `_BaseDocument`, and `AssociatedDocument`) from `common.py` and relocate them into a standalone `documents.py` file, then both `common.py` and `task.py` can import whatever classes they need from `documents.py` without creating a circular import.

estuary-cdk/estuary_cdk/capture/task.py
```diff
@@ -270,3 +270,86 @@ async def run_task(parent: Task):
...
+            name=name,
+            output=output,
+            stopping=stopping,
+            tg=tg,
+            transactor=transactor,
+            requires_ack=requires_ack,
+        )
+        self._log: Logger = log
+        self._name: str = name
+        self._buffer_lock: asyncio.Lock = asyncio.Lock()
```
Alex-Bair: nit: It's slightly confusing that `MultipleWritersTask` has an internal `_task` object it's wrapping instead of being a task itself, and that causes us to reach into that internal `_task`'s private attributes elsewhere like in `webhook/server.py`. What do you think about extracting a bunch of the common attributes and behaviors to a `BaseTask` class, then have `Task` and `MultipleWritersTask` inherit from it and define whatever public methods they need to expose for their intended usage pattern?

Alex-Bair (commented):
Thanks for documenting what we talked about over VC here too, I appreciate it!

After I reviewed again, I had one blocking item around the duplicate `AssociatedDocument` class (I think we can avoid this duplication with a little re-organization) & an optional nit about lifting common attributes & behaviors between `Task` and `MultipleWritersTask` into a shared spot.

estuary-cdk/estuary_cdk/capture/task.py
nicolaslazo: Agreed, I was messing with a ton of `Task` private attributes. It's fixed.

Alex-Bair (approved):
LGTM, thank you!

PR #4375, opened 2026-05-06
source-postgres: Replication idle timeouts redux

willdonnelly:
**Description:**

Replaces the 6 hour "stream-to-fence watchdog" with a 15 minute receive timeout inside relayChanges. The previous WDT was reset by decoded change/commit events, and so we set the failure threshold _very high_ to ensure that even long periods of restart_lsn-to-confirmed_flush_lsn catchup decoding couldn't cause it to fail.

The new logic enforces a 15-minute upper bound on how long the connector will wait for _any_ message from the server during replication. This now includes Primary Keepalive Messages, which should be arriving several times a minute no matter what the server is doing [1], so this method of checking idleness should be more robust, which in turn lets us make it much more aggressive. In theory 15 minutes is actually still absurdly generous relative to the actual timescales involved.

[1] We send StandbyStatusUpdate messages every 10s with `replyRequested` set. So long as the server is processing these, it should be sending back a stream of PrimaryKeepaliveMessages every 10s as well. I have confirmed that walsender has special cases in all the places where a long-running operation might block the message dispatch loop, and in those cases the server should send keepalives every 15s instead.

### DIFF
--- a/source-postgres/replication.go
+++ b/source-postgres/replication.go
@@ -28,11 +28,14 @@ import (
 )
 
 const (
-	// streamToFenceWatchdogInterval is the length of time to wait between checking whether
-	// we're still processing change events. This should never be hit in normal operation,
-	// and exists only so that certain rare failure modes produce an error rather than
-	// blocking forever.
-	streamToFenceWatchdogInterval = 6 * time.Hour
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #4376, opened 2026-05-06
source-zendesk-support-native: add side conversations stream

asmith-octopus:
## Description

Adds a new `side_conversations` capture stream to the Zendesk Support native connector.

Side conversations are private conversations that agents can have with external parties (e.g. third-party vendors or internal teams) without the end-user seeing them. They are scoped per-ticket and accessible via `GET /tickets/{ticket_id}/side_conversations`.

## API approach

The Zendesk side conversations API does not have its own incremental export endpoint — it can only be queried per-ticket. To avoid full re-scanning all tickets on every sync, the implementation mirrors the pattern used by `ticket_comments` and `ticket_audits`:

1. Use the tickets incremental cursor export to find recently-updated tickets.
2. For each updated ticket, fetch its side conversations via `GET /tickets/{ticket_id}/side_conversations`.
3. Yield side conversations along with updated cursors as the stream progresses.

## Feature flag check

Side conversations are an optional Zendesk feature that must be enabled in the account settings. During discovery, the connector checks `GET /account/settings` for the `side_conversations.show_in_context_panel` flag and only surfaces the stream if the feature is active on the account.

## Risks

Side conversations would be captured in full each time and ticket is updated.

## Testing

Added specific test suite for side coversations stream. Snapshots may need updating by team with access to the Estuary account.


Alex-Bair:
Thank you for the contribution, Andrew! It looks good! You're right that there are a couple small items the Estuary team has to do before this can be merged, which in this case is just updating our snapshot tests to include documents from the `side_conversations` stream. I've opened https://github.com/estuary/connectors/pull/4431 to do that & get your contributions merged. I'll follow up here once that PR's merged & the `side_conversations` stream is available.

Alex-Bair:
Following up, #4431 has been merged & the `side_conversations` stream is available now.

PR #4384, opened 2026-05-07
materialize-iceberg: restore default catalog setting

danielnelson:
In #4222, the default catalog was removed and was switched to using a fully qualified query.  However this introduced a possible error if the materialization is restarted during the post-commit apply.

The scripts are sent once on each startup but the queries are staged during Load and applied in Acknowledge. If interrupted and the connector is upgraded you will have newer scripts with an older version of the queries.

Adding back the default catalog is safe, because it will only be used if the queries do not have 3-part fully qualified names.

**Workflow steps:**

No changes needed

None

**Notes for reviewers:**


SeanWhelan (approved):
LGTM, thanks!

jacobmarble (approved):

PR #4387, opened 2026-05-08
source-dynamics-365-finance-and-operations: defer reading deletes to reduce memory usage

Alex-Bair:
**Description:**

Captures with hundreds of bindings catching up on years of changes have been OOM killed shortly after starting. The primary memory cost was in the new `defer_deletes` function added in https://github.com/estuary/connectors/pull/4337. That function maintained two `dict`s that cached information, including entire deletion rows in order to emit them after emitting all non-deletion rows, and that cached data is enough to tip captures over the 1 GiB memory limit when catching up on lots of historical data.

This PR updates the connector to use an alternative strategy. Instead of caching deletion rows in memory then yielding them at the end, we defer reading any files whose first row is a deletion until all other files are read. That means we don't need to cache rows at all and can yield them as we read them. That does rely on the assumption that files either contain entirely deletions _or_ non-deletions; that's consistent with the CSVs I've observed from the source system, and the connector fails loudly if that assumption doesn't hold.

**Workflow steps:**

**Notes for reviewers:**


source-dynamics-365-finance-and-operations/source_dynamics_365_finance_and_operations/api.py
```diff
@@ -188,93 +185,143 @@ async def get_folder_contents_for_table(
...
-    # latest_upsert_key tracks the highest order_key for non-deletion events.
-    latest_upsert_key: dict[str, tuple[int, datetime]] = {}
-    # pending_deletes buffers all deletions. It gets drained after yielding
-    # all non-deletions.
-    pending_deletes: dict[str, TransformedRow] = {}
+def _row_version(row: TransformedRow, csv_name: str) -> int:
+    version = row.get("versionnumber")
+    if not isinstance(version, str):
+        raise ValueError(f"{csv_name} contains a row missing required field 'versionnumber'.")
+    return int(version)
```
nicolaslazo: I'd say that these extraction + validation fns could be turned into a Pydantic model... if we weren't running out of memory already

nicolaslazo (approved):
LGTM!

source-dynamics-365-finance-and-operations/source_dynamics_365_finance_and_operations/api.py
Alex-Bair: Yeah... I ended up getting rid of these too. Even caching row versions and ids caused OOMs when processing lots of data. I ended up relying on the same assumptions as I did before the recent changes; changes are ordered sequentially within a file by their `SinkModifiedOn` values, and changes across files are sequential per the files' modification times. The only change now is that if a file contains deletes, we defer reading the file until all other non-delete-containing files in the folder have been read.

Issue #3628, opened 2025-12-16
source-bigquery-batch: Google Analytics Daily Reporting


jonwihl:
A common use case for BigQuery is exporting Google Analytics daily event data. Ideally, all of the data would end up in the same collection instead of being broken out per day.

It should:

- [ ] Push all new bindings to a single collection
- [ ] Routinely delete older bindings. After the daily report is read into Estuary, it can be removed. We could do a rolling week of bindings at a time

jwhartley:
Also had a customer using postgres table inheritance with 89 child tables of each parent, and wanting to CDC stream data just for the parents (single collection for all child changes), where something general might come in handy
Will B ([slack](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1767640436470089?thread_ts=1767326824.349109&cid=C03Q2NRFKDL)): 
> might be a useful platform-level feature we're missing (that is, something to say "take all discovered bindings matching <regex> and map them to a single output collection" and have it happen automatically, for any connector

Alex-Bair:
After the discussion today, it sounds like this _might_ be possible by discovering these tables with the same recommended name. The thought is that the recommended name is usually the discriminator determining which collection receives documents from which bindings, so using the same recommended name for multiple bindings would cause them to send documents to the same collection. This wouldn't be the default behavior for `source-bigquery-batch`, so we'd probably need some kind of option to tell the connector to use the same recommended name for a set of discovered bindings.

willdonnelly:
I'm looking into this. My initial thought is that we probably don't want to duplicate `source-bigquery-batch` in its entirety, so we'll probably want to build this feature as a special case toggled by some advanced option or feature flag. If we specifically want this to look like a distinct connector, perhaps we'd feature-flag it and then have a simple mechanism for applying different feature-flag defaults based on the connector variant.

@dyaffe you mentioned creating this as a "G4A Data Export" connector variant. Was this (a distinct connector for users to pick in the UI, which then automatically does the appropriate Google Analytics special cases) specifically the desired solution or just what you thought would be simplest?

dyaffe:
I was thinking it would be the same connector and we'd just use a variant tag to provide different setup instructions, etc if that works?

willdonnelly:
Okay, I've been doing some more thinking and investigation. TL;DR: there are two distinct options for how we might want to implement this, the lazy but potentially inefficient way and the fancy but more performant way.

Background info:

[...]

PR #4390, opened 2026-05-08
source-ga4-bigquery: New connector for ingesting GA4 BigQuery exports

willdonnelly:
**Description:**

This PR adds a new capture connector for Google Analytics daily BigQuery export tables. Discovers `events_YYYYMMDD` / `users_YYYYMMDD` / `pseudonymous_users_YYYYMMDD` tables as a single binding per stream type.

Each binding is captured by reading out daily tables individually (with row-level checkpointing for incremental resumability), using a table scheduling strategy which reads each table twice: once when it's new and once at a delay to catch any late-arriving events. Within a particular stream, table readout proceeds from newest to oldest table, in order to prioritize recent data even when large historical backlogs need to be captured.

Overall backfill size can be bounded using a `min_date` advanced setting. If this is not set, all historical data will be backfilled, which could be rather a lot of data.

Currently the connector will not capture `events_intraday_YYYYMMDD` or `events_fresh_YYYYMMDD`. It might be desirable to add support for them in the future, as both of these would give us lower-latency streaming. However it's tricky to define an incremental polling cursor, so this would entail refreshing the latest table multiple times throughout the day:
- The `events_intraday_YYYYMMDD` table is created when "Streaming export" is enabled. It is written continuously as events are recorded throughout the day, and then deleted after `events_YYYYMMDD` is created.
- The `events_fresh_YYYYMMDD` tables are created when using "Fresh Daily" export, and the tables are repeatedly updated in ~hourly batches throughout the day. It is definitely possible to have ordinary `events_YYYYMMDD` daily tables alongside "fresh daily" ones because one of our prospective users has both.

Part of https://github.com/estuary/connectors/issues/3628

https://docs.estuary.dev/reference/Connectors/capture-connectors/google-analytics-4-bigquery-native/ should be deleted or replaced with the documentation for this new connector, as this connector was built because that guidance just didn't really work.

**Notes for reviewers:**

This connector departs from the usual pattern of stimulus-response snapshotting I like to use, because there's just no real way to set something like that up for a Google Analytics to BigQuery daily export pipeline unless we're okay with our test runs taking multiple days.

Instead I've tested this against the `bigquery-public-data.ga4_obfuscated_sample_ecommerce` public dataset to verify stream discovery and events capture work correctly and reliably. This dataset only contains `events` tables and only for a 3-month window back in 2021, so unfortunately that means that the user-data tables and the actual behavior when polling a live dataset that's having new tables added daily are less tested than I would like, however the actual scheduling and readout logic are largely shared between events and users tables, and the scheduling behavior is covered by the `TestEvaluatePollCycle` unit test so I am cautiously optimistic.

Anyway I think that unless I've missed something obvious the only thing we can really do is release this and keep an eye on things when the first users try it in production.


source-ga4-bigquery/main_test.go
```diff
@@ -0,0 +1,451 @@
...
+			// to the dataset's oldest suffix.
+			Name:               "small_dataset_clamps_window",
+			Initial:            streamState{},
+			Suffixes:           []string{"20260504", "20260505", "20260506"},
+			WindowDays:         windowDays,
+			WantPending:        []string{"20260506", "20260505", "20260504"},
+			WantPrimaryThrough: "20260506",
+			WantFinalThrough:   "20260504",
+			WantAdded:          3,
+		},
```
Alex-Bair, line 171: I'm thinking through a similar test case where a capture is set up two days after a BigQuery export was started & there's only two days worth of data in the BigQuery tables. What would happen in that case? I _think_ the earliest day gets `FinalThrough` set to it on the first poll, then subsequent polls would skip it even though it hasn't aged out of the 72 hour late event window yet.

source-ga4-bigquery/discovery.go
```diff
@@ -0,0 +1,329 @@
...
+			"/event_bundle_sequence_id",
+			"/batch_event_index",
+		},
+		DataFields: map[string]*jsonschema.Schema{
+			"event_timestamp":          {Type: "integer"},
+			"event_name":               {Type: "string"},
+			"user_pseudo_id":           nullableStringSchema(),
+			"event_bundle_sequence_id": nullableIntegerSchema(),
+			"batch_event_index":        nullableIntegerSchema(),
+		},
```
Alex-Bair, line 90: I couldn't figure out how to comment on commit 3573cfa6e7e8bd07a3995bd9397c98fda28fe486, so I'll comment here instead.

I'm 99% sure that even if the Flow runtime is ok with nullable key fields, materializations built with `materialize-boilerplate` will fail to validate them unless there's a default annotation present.
https://github.com/estuary/connectors/blob/e35f1121b947c4eabd2ec52927765eaa9243f47e/materialize-boilerplate/validate.go#L77-L89

Does that line up with your understanding? If so, would there be sensible defaults to pick for these nullable key fields so users can capture & materialize their pre-2024 data? Maybe using `-1` and `"unknown"` as sentinels could work?

Alex-Bair (approved):
LGTM % a couple questions.

source-ga4-bigquery/main_test.go
willdonnelly: Hmm, that's a good question. I'll think through that scenario a bit more and add some more tests covering it. I suspect the right answer is actually going to be that `FinalThrough` should stay null or something until there are at least `WindowDays` tables present.

source-ga4-bigquery/discovery.go
willdonnelly: Ah, I _knew_ there was some other issue with collection key nullability which I couldn't recall. Yes, we can probably come up with a reasonable default value for each of these nullable properties.

willdonnelly: How are default values applied in this context, is it just when the properties are omitted or will explicit `"foo": null` entries be automatically replaced with the default value as well in a collection-key context?

If it's just the former it seems like we'd also need some logic on our end to suppress explicit nulls from the `user_pseudo_id` and `event_bundle_sequence_id` columns since those ones exist but may be null in the source DB.

source-ga4-bigquery/main_test.go
willdonnelly: This should be fixed now, with basically the solution described above.

source-ga4-bigquery/discovery.go
willdonnelly: This should be fixed, I'm _fairly_ sure the answer to my previous question is that JSON schema defaults only apply to omitted properties so I added some logic to omit those nullable collection key properties if/when we see null values in a particular row.

Alex-Bair: My recollection is the same, that the `default` annotation is only applied for omitted properties. Omitting those key fields when they contain null values sounds good to me.

source-ga4-bigquery/main_test.go
Alex-Bair: Looks good to me!

PR #4393, opened 2026-05-11
ci: delayed tag 2 weeks for each connector

### LINKED ISSUE #4060: Connector release channels

mdibaiee:
**Description:**

- Add a new `:delayed` tag which is automatically tagged ~2 weeks after an image has been released.
- This is going to be used to be cautious for some of the more sensitive customers.
- Versioned tags are still our stable channel.
- There is a Claude safety pass in the CI that checks subsequent PRs to make sure that if we are building a delayed image, there has been no roll-forward fix for that connector in later PRs, to prevent a known bug from being released to `:delayed` channel
- I think at the moment we can merge this and have the images published, and we can monitor the CI every once in a while to see if the roll-forward fix check is working fine, and then we can have our support folks put customers on this new image
- Note that there is currently an issue where the rendered form on the UI may be from the versioned tag, whereas the delayed tag is used... causing possible incompatibilities between UI and the delayed image.

For a run of this see: https://github.com/estuary/connectors/actions/runs/26176409516?pr=4393

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/.github/scripts/delayed_tag.py
+++ b/.github/scripts/delayed_tag.py
@@ -1,0 +1,748 @@
+#!/usr/bin/env python3
+"""
+delayed_tag.py — compute the ':delayed' image tag plan for Estuary connectors.
+
+Subcommands
+-----------
+plan   Read ci.yaml / python.yaml, query GitHub for merged PRs, then for each
+       connector pick the most recent PR that (a) touched that connector's
+       directory, (b) was merged at least CUTOFF_DAYS ago, and (c) has not
+       been reverted on main. Writes plan.tsv and later.tsv.
+
+check  Read plan.tsv / later.tsv, call Claude Sonnet for each connector that
+       has LATER PRs to detect roll-forward fixes, then write safe_plan.tsv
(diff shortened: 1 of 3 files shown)


jacobmarble (approved):
This looks brittle, but definitely worth a try.

jacobmarble:
> This looks brittle, but definitely worth a try.

Brittle because there are no tests, and because asking Claude anything is non-deterministic. I vote we try it; if its value is proven, and if it does turn out to be brittle, then we could rewrite some of the logic in unit-tested Go or Python.

mdibaiee:
@jacobmarble I thought about this too, and I guess I needed a push... I'm going to have it rewritten in a proper script file with tests so we can rely on it more easily

jacobmarble (approved):
Lovely!

mdibaiee:
I'm going to merge, but I propose we don't use these images yet, and we wait a bit and watch the CI pipeline to see how it goes with roll-forward fixes. We also inevitably need to wait a bit for all connectors to have a 2-weeks-old PR that we can use for the delayed tag, some stale connectors don't.

PR #4400, opened 2026-05-11
{source-materialize}-postgres: support JSONB

### LINKED ISSUE #4369: JSONB mapping from source-postgres to materialize-postgres
We want to make sure that JSONB fields in source-postgres are correctly mapped to JSONB in materialize-postgres as well.

The suggested idea is to use `contentMediaType` or a similar annotation in JSONSchema to pass information about the underlying type of the JSON field in the source connector, and use that info in the materialization to materialize it as JSONB.

mdibaiee:
**Description:**

- Use a custom `contentMediaType: "application/vnd.postgresql.jsonb+json"` annotation in JSONSchema to mark an object as originating from a postgresql `JSONB` column type
- In materializations, use `JSONB` for creating such a column
- Add automatic column migrations from `JSON` to `JSONB` and vice versa

I have identified customers that would be affected by this and I am coordinating with our support team to make sure we can transition them to this new change without breaking their pipelines down the line.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-postgres/sqlgen.go
+++ b/materialize-postgres/sqlgen.go
@@ -42,7 +42,7 @@ func createPgDialect(featureFlags map[string]bool) sql.Dialect {
 	jsonbContentMediaType := "application/vnd.estuary.postgresql.jsonb+json"
 	jsonOrJsonb := func(jsonMapper, jsonbMapper sql.MapProjectionFn) sql.MapProjectionFn {
 		return func(p *sql.Projection) (sql.DDLer, sql.CompatibleColumnTypes, sql.ElementConverter) {
-			if p.Inference.String_ != nil && p.Inference.String_.ContentType == jsonbContentMediaType {
+			if p.Inference.ContentMediaType == jsonbContentMediaType {
 				return jsonbMapper(p)
 			}
 			return jsonMapper(p)
--- a/materialize-postgres/sqlgen.go
+++ b/materialize-postgres/sqlgen.go
@@ -34,18 +34,36 @@ func createPgDialect(featureFlags map[string]bool) sql.Dialect {
 		binaryMapping = sql.MapStatic("BYTEA", sql.UsingConverter(sql.Base64Decoder))
 	}
 
+	// Fields originating from a PostgreSQL JSONB column carry this
+	// (non-standard, vendor-tree) contentMediaType so they can be recreated
+	// as JSONB at the destination instead of collapsing onto JSON. Anything
+	// without the annotation defaults to JSON, preserving the historical
+	// behavior for existing materializations and non-Postgres sources.
+	jsonbContentMediaType := "application/vnd.postgresql.jsonb+json"
+	jsonOrJsonb := func(jsonMapper, jsonbMapper sql.MapProjectionFn) sql.MapProjectionFn {
+		return func(p *sql.Projection) (sql.DDLer, sql.CompatibleColumnTypes, sql.ElementConverter) {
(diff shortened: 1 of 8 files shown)


source-postgres/.snapshots/TestDiscoveryComplex
```diff
@@ -36,10 +36,12 @@ sql> COMMENT ON COLUMN test.discoverycomplex_934635.k1 IS 'I think this is a key
...
           },
           "doc": {
-            "description": "(source type: json)"
+            "description": "(source type: json)",
+            "contentMediaType": "application/json"
           },
           "doc/bin": {
-            "description": "(source type: non-nullable jsonb)"
+            "description": "(source type: non-nullable jsonb)",
+            "contentMediaType": "application/vnd.postgresql.jsonb+json"
```
danielnelson: I noticed that PostgREST uses a similar media type format, but they define it on their own vendor: `application/vnd.pgrst.object+json`.  Should we instead use something specific to us as a vendor such as `application/vnd.estuary.jsonb+json`?

I know this specifically meant to handle same db round-tripping, but I wonder if there really won't be any ideas to have things like mysql -> postgres customizations based on the source type, will we have a combinatorial issue?

danielnelson (commented):
Both json and jsonb are encoded in the collection in the same way, so it doesn't seem like the media type is describing anything about the actual format of the data.  It also see that `contentMediaType` is described only for strings.

I feel like we potentially could need to invent media types for every possible source type?  I'm also imagining things like a request for mirroring aspects such as string collation.  We do already have the information in the description and while it's not super machine readable we could try parsing.

Or perhaps we would be better served with a new custom annotation for source type?

materialize-postgres/sqlgen.go
```diff
@@ -80,7 +98,12 @@ func createPgDialect(featureFlags map[string]bool) sql.Dialect {
 			"text":                     {sql.NewMigrationSpec([]string{"bytea"}, sql.WithCastSQL(stringToByteaCast))},
 			"character varying":        {sql.NewMigrationSpec([]string{"bytea"}, sql.WithCastSQL(stringToByteaCast))},
 			"bytea":                    {sql.NewMigrationSpec([]string{"text"}, sql.WithCastSQL(byteaToStringCast))},
-			"*":                        {sql.NewMigrationSpec([]string{"json"}, sql.WithCastSQL(toJsonCast))},
+			// PostgreSQL accepts implicit casts in both directions between
+			// json and jsonb, so columns can move freely as the upstream
+			// source type annotation flips.
+			"json":  {sql.NewMigrationSpec([]string{"jsonb"})},
+			"jsonb": {sql.NewMigrationSpec([]string{"json"})},
+			"*":     {sql.NewMigrationSpec([]string{"json"}, sql.WithCastSQL(toJsonCast))},
```
jacobmarble: Should `*` to `jsonb` also be permitted?

jacobmarble (approved):

mdibaiee:
@danielnelson that was a great point I had overlooked. I had a chat with data plane folks and they were happy to have `contentMediaType` for all types, so I'm working on a PR for that https://github.com/estuary/flow/pull/2945

source-postgres/.snapshots/TestDiscoveryComplex
mdibaiee: @danielnelson I think you are right, I was also a bit worried about overlapping with another vendor (postgres itself), so I think we can do `vnd.estuary.postgres.jsonb+json` or something like this

PR #4431, opened 2026-05-11
source-zendesk-support-native: add `side_conversations` stream

Alex-Bair:
**Description:**

Adds the `side_conversations` stream. There _may_ be more elegant ways of adding this stream and reducing some code duplication, but it's a little more involved than usual since the `id` field for this stream is a string instead of the usual integer. For now, I'm opting for a little code duplication until we see if there are other streams that have string `id`s and can figure out how best to re-use the code within the connector.

I evaluated using the [time-based incremental side conversations event export](https://developer.zendesk.com/api-reference/ticketing/side_conversation/side_conversation_event/#incremental-side-conversation-event-export) instead of the cursor-based incremental ticket export as the parent stream. That'd reduce how much duplicate data we process (we'd only fetch side coversations of tickets who had some side conversation event happen), but the limit for the side conversation endpoint is 100 and since it's a time-based export, the connector would end up missing data if more than 100 side conversation events happened in a single second. So, we're using the cursor-based incremental ticket export to avoid that potential missing data issue.

Snapshot changes are expected due to:
- me muddling around in our dev Zendesk account to test out the various ways the API exposes side conversations
- the addition of the `side_conversations` stream

The connector's documentation should be updated to include the `side_conversations` stream.


nicolaslazo (approved):
LGTM. Whenever we decide to refactor this, the fix might be parametrising the primary key type like I did in [source-posthog](https://github.com/estuary/connectors/blob/main/source-posthog/source_posthog/models.py#L124-L131) with `T` being bound to `str | int`.

Alex-Bair:
Ooo, thanks for mentioning that, I agree that'd be a good thing to do when refactoring.. Another thing we'd want to figure out during that refactor is how to reuse the existing `fetch_ticket_child_resources` and `backfill_ticket_child_resources` for the `side_conversations` stream too. There are ways to do that, but they're a little more involved than I felt was needed right now.

PR #4439, opened 2026-05-12
source-sage-intacct: surface result level errors

Alex-Bair:
**Description:**

Sage Intacct returns content level failures with the description nested in `result.errormessage` rather than at the response or operation level. `raise_for_error` previously only inspected the upper two scopes, so these failures were raised as the bare string "result status: failure" with the actual error description discarded. It also required `errorno` to be a non-null string, which crashed `model_validate` when Sage returned a null `errorno` on an otherwise-descriptive error.

This commit makes `Error` fields optional, moves first error selection and formatting onto `ErrorMessage` via `__str__`, and checks `response.operation.result.errormessage` in `raise_for_error` before falling back to the bare status error message.


SeanWhelan (approved):
LGTM, thanks!

PR #4452, opened 2026-05-12
source-sage-intacct: only discover bindings the user has permission to query

Alex-Bair:
**Description:**

Sage gates QUERY access per-object via the user's role and the tenant's subscription. Previously, a single PL04000005 denial (ex: "you do not have permission" or "operation cannot be performed on objects of type X") were discovered even though the provided credentials lack the capability to query them. The connector would then try to query them and predictably fail.

This PR omits non-query-able objects from the list of discovered resources by detecting PL04000005 errors when probing which objects are queryable.

**Notes for reviewers:**

Tested with a Sage Intacct account that does not have permission to query all of our defined resources. Confirmed that those resources that weren't query-able were excluded from the list of discovered resources.

### DIFF
--- a/source-sage-intacct/source_sage_intacct/models.py
+++ b/source-sage-intacct/source_sage_intacct/models.py
@@ -65,6 +65,11 @@ class EndpointConfig(BaseModel):
 ConnectorState = GenericConnectorState[ResourceState]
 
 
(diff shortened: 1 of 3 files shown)


nicolaslazo (approved):
LGTM

PR #4454, opened 2026-05-12
source-greenhouse-native: make demographic questions a full refresh stream

Alex-Bair:
**Description:**

This PR's scope includes:
- Adding sensible defaults for `model` and `schema_inference` to the CDK's helper `SnapshotResource` class.
- Changing `source-greenhouse-native`'s `demographic_questions` stream to be a full refresh stream instead of an incremental one. The API doesn't support filtering by the `updated_at` field, so we have to fetch all records on each sweep. I opted to make this a full refresh stream instead of a client-side incremental one since I don't have sufficient evidence that the `updated_at` field always exists for `demographic_questions` documents (mostly, if it always exists, why doesn't Greenhouse let us filter by it?).

**Notes for reviewers:**

We don't have a Greenhouse dev account, so I couldn't test out this change locally. But it re-uses a lot of the existing, working code for paginating through the API's responses, so it _should_ work.

Existing capture tasks will need to re-discover the `demographic_questions` binding in order to pick up the key change from `id` to `_meta/row_id`. This should happen automatically via auto-discovers.

### DIFF
--- a/.claude/skills/classify-stream-types/SKILL.md
+++ b/.claude/skills/classify-stream-types/SKILL.md
@@ -95,15 +95,12 @@ def open(
         state,
         task,
         fetch_snapshot=functools.partial(list_all, http, ...),
-        tombstone=MyDocument(_meta=MyDocument.Meta(op="d")),
     )
 
 resource = SnapshotResource(
     name=MyDocument.name,
-    model=MyDocument,
     open=open,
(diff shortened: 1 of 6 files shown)


nicolaslazo (approved):
LGTM 👍

Alex-Bair:
Following up for my own future reference, it looks like there isn't an `updated_at` field on the captured documents, so we aren't able to make `demographic_questions` a client-side incremental stream & it'll have to remain a full refresh stream.

PR #4456, opened 2026-05-12
source-quickbooks: Rework incremental fetch bounds

willdonnelly:
**Description:**

Changes the lower-bound timestamp for queries to be an _exclusive_ lower bound so that when doing incremental fetches we never receive the same item which was the source of that particular cursor value. To avoid issues with late-arriving data or same-second-as-query events, the upper-bound timestamp is changed to `now() - DELAY` so that we're never looking at items modified more recently than some interval. I have arbitrarily picked 30s as that interval but I'm happy to adjust it if anyone feels like a larger or smaller value would be better.

Since the same query function is used for backfills as well, I have tweaked the backfill page cursor / checkpoint logic such that the same overall behavior _should_ be preserved with the change from `>=` to `>` in the query. This would invalidate any existing state checkpoints for ongoing backfills but there aren't any so that's fine.

### DIFF
--- a/source-quickbooks/source_quickbooks/api.py
+++ b/source-quickbooks/source_quickbooks/api.py
@@ -15,6 +15,13 @@ from .models import QuickBooksEntity
 PAGE_SIZE = 1000
 TARGET_FETCH_PAGE_INVOCATION_RUN_TIME = timedelta(minutes=5)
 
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM, thanks for tackling this!

PR #4460, opened 2026-05-13
materialize-redshift: load bounds and DISTSTYLE ALL for temp load table

mdibaiee:
**Description:**

- We have noticed redshift load queries can be the most time-consuming part of pipelines for target tables that are large. We seem to scan all of the target table unfortunately (there are no "indices"). Here I am adding Key Bounds as well as a recommended "DISTSTYLE ALL" to distribute the load table across compute slices so they are all local during the lookup.
- I would like to try this out on a customer task to see how it fares.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-redshift/sqlgen.go
+++ b/materialize-redshift/sqlgen.go
@@ -172,7 +172,12 @@ func createRsDialect(caseSensitiveIdentifierEnabled bool, featureFlags map[strin
 				},
 				sql.QuoteTransform(`"`, `""`),
 			))),
-		Literaler: sql.ToLiteralFn(sql.QuoteTransform("'", "''")),
+		// Redshift treats backslashes as escape characters in single-quoted
+		// string literals (with no STANDARD_CONFORMING_STRINGS option exposed,
+		// at least on Redshift Serverless). The literal 'foo\bar' parses to
+		// "fooar" — the backslash and following char are consumed. So we
+		// double every backslash in literal output to round-trip correctly.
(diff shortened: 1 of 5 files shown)


danielnelson (approved):
LGTM.  Would it be worth applying and testing the merge bounds change separately from changing the diststyle in order to be sure they are both giving improvements?

materialize-redshift/sqlgen.go
```diff
@@ -172,7 +172,12 @@ func createRsDialect(caseSensitiveIdentifierEnabled bool, featureFlags map[strin
 				},
 				sql.QuoteTransform(`"`, `""`),
 			))),
-		Literaler: sql.ToLiteralFn(sql.QuoteTransform("'", "''")),
+		// Redshift treats backslashes as escape characters in single-quoted
+		// string literals (with no STANDARD_CONFORMING_STRINGS option exposed,
+		// at least on Redshift Serverless). The literal 'foo\bar' parses to
+		// "foobar" — the backslash and following char are consumed. So we
```
jacobmarble: If the backslash and following character were consumed, then wouldn't `foo\bar` parse to `fooar`?

jacobmarble (approved):

materialize-redshift/sqlgen.go
mdibaiee: @jacobmarble yes! you are right

PR #4461, opened 2026-05-13
source-oracle: Additional replication logging

willdonnelly:
**Description:**

Some production source-oracle captures are having issues where they will go and just keep on streaming replication events indefinitely, until an error or the runtime forces a task restart. We've seen this persist for up to 22h and there's no reason to believe it wouldn't go on for longer in the absence of errors or runtime interference.

When this happens, there is basically nothing in the logs for that whole period of time, even though we can infer from various things that the connector is still actively doing work and _probably_ it just missed the fence watermark change event somehow.

This PR adds some more logging and promotes a few DEBUG logs to INFO level so that we can hopefully have a clearer picture of what is going on during these prolonged streaming periods.


Alex-Bair (approved):
LGTM

PR #4465, opened 2026-05-13
source-sage-intacct: capture records with a `null` `WHENMODIFIED` & capture deletions only when we have permission

Alex-Bair:
**Description:**

This PR's scope includes:
- handling records that have a `null` `WHENMODIFIED` field. A `null` `WHENMODIFIED` is evidently possible. I'm not exactly sure _why_ it's possible, but regardless of why `WHENMODIFIED` is `null`, I'm pretty sure the fix is to capture these `null` `WHENMODIFIED` records based on their `WHENCREATED` field instead.
- not attempting to capture deletions if we can't access the `AUDITHISTORY` object required to figure out what's been deleted

See the commit messages for more details.

The connector's documentation should be updated to mention that the connector needs permission to read the `AUDITHISTORY` object in order to capture deletions.

**Notes for reviewers:**

Sanity checks with `flowctl raw discover` and `flowctl preview` before and after my changes with our testing account gave the same outputs.

I also tested out the equivalent API calls for the new `get_creations_since_request` and `get_creations_at_request` with our test account, and they returned 200 OK as expected. They didn't return any records, but that's because there are no records in our test account without a non-`null` `WHENMODIFIED` field.

See this Slack [thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1778618894907279?thread_ts=1778583494.583519&cid=C03Q2NRFKDL) for more context on the specific capture task hitting errors that will get unstuck after these changes are merged.

As part of this change, we've added a 2 minute lag to capturing updates to guard against the connector querying for updated and created records close together, then emitting the create after the update due to network or Sage API races. 2 minutes seems acceptable to me, but I can adjust it lower if we think 2 minutes is too much lag & the creation / update race doesn't seem valid.

### DIFF
--- a/source-sage-intacct/source_sage_intacct/models.py
+++ b/source-sage-intacct/source_sage_intacct/models.py
@@ -198,6 +198,28 @@ class IncrementalResource(BaseDocument, extra="allow"):
     RECORDNO: int
     WHENMODIFIED: AwareDatetime
 
+    def cursor_value(self) -> AwareDatetime:
+        return self.WHENMODIFIED
+
+
+# CreationRecord captures records that exist in Sage with a null WHENMODIFIED.
+# They cannot be captured by the WHENMODIFIED-keyed incremental query, so a
+# parallel sub-task keyed on WHENCREATED picks them up.
+class CreationRecord(BaseDocument, extra="allow"):
+    RECORDNO: int
+    WHENCREATED: AwareDatetime
+
+    def cursor_value(self) -> AwareDatetime:
+        return self.WHENCREATED
+
+
+def parse_backfill_record(raw: dict) -> "IncrementalResource | CreationRecord":
+    """Routes to IncrementalResource when WHENMODIFIED is present, else
+    CreationRecord."""
+    if raw.get("WHENMODIFIED") is not None:
+        return IncrementalResource.model_validate(raw)
+    return CreationRecord.model_validate(raw)
+
 
 class DeletionRecord(BaseDocument, extra="forbid"):
     RECORDNO: int = Field(alias="OBJECTKEY", serialization_alias="RECORDNO")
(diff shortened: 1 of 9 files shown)


source-sage-intacct/source_sage_intacct/models.py
```diff
@@ -198,6 +204,28 @@ class IncrementalResource(BaseDocument, extra="allow"):
...
+    def cursor_value(self) -> AwareDatetime:
+        return self.WHENCREATED
+
+
+def parse_backfill_record(raw: dict) -> "IncrementalResource | CreationRecord":
+    """Routes to IncrementalResource when WHENMODIFIED is present, else
+    CreationRecord."""
+    if "WHENMODIFIED" in raw:
+        return IncrementalResource.model_validate(raw)
+    return CreationRecord.model_validate(raw)
```
nicolaslazo, line 235: > CreationRecord captures records that exist in Sage with a null WHENMODIFIED

Shouldn't `parse_backfill_record` also be checking if `WHENMOFIED` is `None`?

Alex-Bair: We don't necessarily need to. `parse_backfill_record` is used to parse dumped `SageRecord`s, and when `SageRecord`s are dumped any `None` values are excluded:
https://github.com/estuary/connectors/blob/3ab49933eaaac6eca4dd6ac357377a5c3b8f56bd/source-sage-intacct/source_sage_intacct/sage.py#L79

But it is safer to use `raw.get("WHENMODIFIED")` instead of `"WHENMODIFIED" in raw` when determining whether to validate with `IncrementalRecord` or `CreationRecord` in case how `SageRecord` serialization behavior changes later. I've updated this check to cover if `WHENMODIFIED` is `None` too.

nicolaslazo (approved):
Looks good, thanks!

PR #4471, opened 2026-05-14
go/writer: improve parquet write performance

jacobmarble:
**Description:**

Improves parquet write per a new benchmark:
- 15% CPU
- 41% memory
- 75% memory allocations

Two-part refactor of `ParquetWriter`'s buffer-to-sink pipeline:

1. **Column-oriented buffer** (`b55c7158d`): changes the write buffer from `[][]any` to `[]any` of typed slice pointers (`*[]int64`, `*[]parquet.ByteArray`, etc.) with a parallel `[]int16` definition-level slice. Values are converted to their Parquet physical types at `Write` time instead of during the flush loop, eliminating the per-flush transposition pass and reducing GC scanning pressure.

2. **Page-copy** (`cf5e9e76d`): `transferColumnValues` previously decoded every column value from the scratch file and re-encoded it into the sink. Now the scratch writer uses the same codec as the sink and disables dictionary encoding, so compressed data pages are copied byte-for-byte through a new `mergeWriter` type that accepts pre-encoded pages and tracks row counts explicitly.

**Benchmark results** (A = baseline, B = column-oriented buffer, C = page-copy):

Step 1 (A→B): cpu +7%, mem −25%, allocs ~unchanged.
Step 2 (B→C): cpu −9–25%, mem −6–33%, allocs −75%.

```
                                   │      A (sec/op)      │          C (sec/op)          │
                                   │        sec/op        │   sec/op     vs base         │
ParquetWriter/small/uncompressed-8            699.4m ± 2%   559.9m ± 2%  -19.95% (p=0.000 n=8+7)
ParquetWriter/small/snappy-8                  718.6m ± 3%   636.6m ± 6%  -11.41% (p=0.000 n=8+7)
ParquetWriter/large/uncompressed-8             3.271 ± 2%    2.653 ± 2%  -18.92% (p=0.000 n=8+7)
ParquetWriter/large/snappy-8                   3.318 ± 4%    3.036 ± 2%   -8.48% (p=0.000 n=8+7)
geomean                                        1.528         1.302       -14.83%

                                   │     A (B/op)     │          C (B/op)           │
                                   │      B/op        │    B/op      vs base        │
ParquetWriter/small/uncompressed-8          638.3Mi ± 0%   333.0Mi ± 0%  -47.82% (p=0.000 n=8+7)
ParquetWriter/small/snappy-8               651.8Mi ± 0%   444.2Mi ± 0%  -31.85% (p=0.000 n=8+7)
ParquetWriter/large/uncompressed-8         2.945Gi ± 0%   1.464Gi ± 0%  -50.29% (p=0.000 n=8+7)
ParquetWriter/large/snappy-8               2.967Gi ± 0%   1.990Gi ± 0%  -32.94% (p=0.000 n=8+7)
geomean                                    1.364Gi        819.8Mi       -41.32%

                                   │  A (allocs/op)  │       C (allocs/op)       │
                                   │   allocs/op     │ allocs/op    vs base      │
ParquetWriter/small/uncompressed-8       4.013M ± 0%   1.020M ± 0%  -74.58% (p=0.000 n=8+7)
ParquetWriter/small/snappy-8             4.013M ± 0%   1.021M ± 0%  -74.56% (p=0.000 n=8+7)
ParquetWriter/large/uncompressed-8      19.997M ± 0%   5.046M ± 0%  -74.77% (n=8+7)
ParquetWriter/large/snappy-8            19.997M ± 0%   5.048M ± 0%  -74.75% (p=0.000 n=8+7)
geomean                                  8.958M        2.269M       -74.67%
```

Net: **−15% CPU, −41% memory, −75% allocations** (all p=0.000).

**Workflow steps:**

No user-visible change.

None.

**Notes for reviewers:**

- `makeColumnBuffer` stores a pointer-to-slice in `[]any` to avoid boxing a 3-word slice header on every append.
- `appendVal[T parquetValue]` is a 5-line generic the compiler inlines at each call site, allowing devirtualization of the `getValFn[T]` parameter.
- `mergeWriter` (`merge_writer.go`) is the new sink type; it accepts `WriteDataPage` calls and tracks row counts via `SetNumRows`.

### DIFF
--- a/go/writer/merge_writer.go
+++ b/go/writer/merge_writer.go
@@ -1,0 +1,266 @@
+package writer
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io"
+
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet"
+	"github.com/apache/arrow-go/v18/parquet/file"
+	"github.com/apache/arrow-go/v18/parquet/metadata"
+	"github.com/apache/arrow-go/v18/parquet/schema"
+)
+
+// parquetMagic is the 4-byte file marker that opens and closes every parquet file.
+var parquetMagic = []byte{'P', 'A', 'R', '1'}
+
+// mergeWriter writes a parquet file by importing already-encoded data pages into row groups. It
+// is built on top of arrow-go's metadata builders and PageWriter, but skips arrow-go's
+// *file.Writer entirely. *file.Writer assumes data is fed via WriteBatch — it owns dictionary
+// emission, page buffering, and chunk-stats aggregation, none of which apply when we are pumping
+// pre-encoded pages from one parquet file into another. mergeWriter exposes a smaller surface
+// shaped to that import use case: callers feed pre-encoded data pages directly and tell the row
+// group its row count explicitly.
+//
+// The sink must satisfy arrow-go's parquet/internal/utils.WriterTell (io.Writer + Tell() int64);
+// countingWriteCloser does after we add Tell. Go's structural typing lets us pass a value of an
+// external concrete type into arrow-go's internal-interface-typed parameter, since interface
+// satisfaction depends only on method sets.
+type mergeWriter struct {
+	sink     mergeWriterSink
+	schema   *schema.Schema
+	props    *parquet.WriterProperties
+	metadata *metadata.FileMetaDataBuilder
+
+	rowGroupCount int
+	open          bool
+	rowGroup      *mergeRowGroupWriter
+}
+
+// mergeWriterSink is the contract the sink must satisfy for arrow-go's PageWriter to use it.
+// In practice this is countingWriteCloser, which adds Tell() returning the byte count.
+type mergeWriterSink interface {
+	io.Writer
+	Tell() int64
+}
+
+func newMergeWriter(sink mergeWriterSink, sc *schema.GroupNode, props *parquet.WriterProperties, kv metadata.KeyValueMetadata) (*mergeWriter, error) {
+	if props == nil {
+		props = parquet.NewWriterProperties()
+	}
+	fileSchema := schema.NewSchema(sc)
+	if _, err := sink.Write(parquetMagic); err != nil {
+		return nil, fmt.Errorf("writing parquet magic header: %w", err)
+	}
+	return &mergeWriter{
+		sink:     sink,
+		schema:   fileSchema,
+		props:    props,
+		metadata: metadata.NewFileMetadataBuilder(fileSchema, props, kv),
+		open:     true,
+	}, nil
+}
+
+// AppendRowGroup begins a new row group. Any currently-open row group is closed first.
+func (mw *mergeWriter) AppendRowGroup() (*mergeRowGroupWriter, error) {
+	if !mw.open {
+		return nil, fmt.Errorf("mergeWriter: AppendRowGroup on closed writer")
+	}
+	if mw.rowGroup != nil {
+		if err := mw.rowGroup.Close(); err != nil {
+			return nil, fmt.Errorf("closing prior row group: %w", err)
+		}
+	}
+	rgMeta := mw.metadata.AppendRowGroup()
+	rg := &mergeRowGroupWriter{
+		parent:   mw,
+		metadata: rgMeta,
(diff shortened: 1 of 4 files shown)


go/writer/merge_writer.go
```diff
@@ -0,0 +1,266 @@
...
+		metadata: rgMeta,
+		ordinal:  int16(mw.rowGroupCount),
+	}
+	mw.rowGroupCount++
+	mw.rowGroup = rg
+	return rg, nil
+}
+
+// Close finalizes any open row group, writes the parquet footer (file metadata, footer length,
+// and trailing magic), and marks the writer closed. The underlying sink is not itself closed.
```
mdibaiee, line 88: Why is the underlying sink not closed?

go/writer/merge_writer.go
```diff
@@ -0,0 +1,266 @@
...
+	metadata   *metadata.RowGroupMetaDataBuilder
+	ordinal    int16
+	nextCol    int
+	currentCol *mergeColumnWriter
+	nrows      int
+	closed     bool
+}
+
+// SetNumRows records the row count for this row group; required before Close. The same value
+// must also describe every column written into the row group.
```
mdibaiee, line 140: Not sure what this means "The same value must also describe every column written into the row group"

go/writer/merge_writer.go
```diff
@@ -0,0 +1,266 @@
...
+	return mw.metadata.Snapshot()
+}
+
+// mergeRowGroupWriter writes columns sequentially into one row group. Use NextColumn to advance
+// through columns in schema order. Set the row count via SetNumRows before Close.
+type mergeRowGroupWriter struct {
+	parent     *mergeWriter
+	metadata   *metadata.RowGroupMetaDataBuilder
+	ordinal    int16
+	nextCol    int
```
mdibaiee, line 133: nit: The name of this property threw me off a bit... if I understand correctly this is essentially `colOrdinal` of the column being written by `NextColumn`

also maybe we can store it as int16 since it seems to be used as such

go/writer/parquet.go
```diff
@@ -210,47 +215,119 @@ func NewParquetWriter(w io.WriteCloser, sch ParquetSchema, opts ...ParquetOption
 	schemaRoot := schema.MustGroup(schema.NewGroupNode("schema", parquet.Repetitions.Required, fields, -1))
 
 	cwc := &countingWriteCloser{w: w}
+	props, kvmeta := sinkWriterProperties(cfg)
+	// Matching arrow-go's *file.Writer constructor, we panic on the magic-bytes write failure
+	// rather than threading an error through NewParquetWriter — a sink that can't accept 4 bytes
+	// at construction time will fail again immediately on the first real write.
+	sinkWriter, err := newMergeWriter(cwc, schemaRoot, props, kvmeta)
+	if err != nil {
+		panic(fmt.Sprintf("creating sink writer: %s", err))
```
mdibaiee, line 224: Would we get better error presentation for users if we bubble up the error or we need the panic stacktrace?

go/writer/merge_writer.go
```diff
@@ -0,0 +1,266 @@
...
+func (cw *mergeColumnWriter) WriteDataPage(page file.DataPage) error {
+	if cw.closed {
+		return fmt.Errorf("mergeColumnWriter: WriteDataPage on closed column")
+	}
+	if cw.pageWriter.HasCompressor() {
+		var buf bytes.Buffer
+		compressed := cw.pageWriter.Compress(&buf, page.Data())
+		compressedData := make([]byte, len(compressed))
+		copy(compressedData, compressed)
+		newBuf := memory.NewBufferBytes(compressedData)
```
mdibaiee, line 232: Do you mind adding a comment as to why the copy is necessary?

mdibaiee (commented):
Amazing results!!

A few small comments, otherwise LGTM

One note: I think we can now remove `WithDisableDictionaryEncoding` here: https://github.com/estuary/connectors/blob/8da94b35392744324de0751d412f2f940c0c9995/go/writer/parquet.go#L171-L175

mdibaiee:
@jacobmarble it seems like tests are timing out on this branch while they work on `main`... perhaps something is hanging with the new implementation?

dyaffe:
Just commenting directly here as per the slack thread.  Lets hold off on this change until after we switch to the Snowpipe streaming SDK.

PR #4472, opened 2026-05-14
capture/bigquery: Translate infs to strings

willdonnelly:
**Description:**

Impacts both source-ga4-bigquery and source-bigquery-batch, but I only added a test to source-bigquery-batch since the actual value translation logic is shared and source-bigquery-batch is the one where we can easily do round-trip testing.


Alex-Bair (approved):
LGTM

PR #4474, opened 2026-05-14
source-oracle: Replication logging and self-checks and bugfix

willdonnelly:
**Description:**

We have observed a production capture looping repeatedly with `endSCN == startSCN` for each iteration. This condition persisted for about 40 minutes and ~1k loop iterations until, presumably, something about the set of available redo logs changed in a way which got it unstuck. I also strongly suspect that the way this changed was that some older archived logs were deleted, and that once unstuck we _silently skipped_ an uncaptured portion of the SCN range as a result.

This PR does three things:
- Adds some more chatty logging about redo log files being added in each polling iteration. We'll probably want to turn this back down at some point but right now I want as much info as possible about what's going on here.
- Adds a self-check function `validateLogFileCoverage` which checks the selected list of redo log files from `addLogFiles` and verifies that this list represents a **contiguous SCN range** which **includes the startSCN/endSCN endpoints**. Previously if these properties were violated we would just silently miss some data we thought we were capturing, and now it will be a noisy failure probably requiring a backfill to get past. Which is how things should be, for unrecoverable breaks in replication continuity.
- Modifies `addLogFiles` so that the "we selected >= 2GiB of data" break condition can only trigger when the selected set of log files also constitutes _forward progress_. This should prevent the `endSCN == startSCN` infinite looping behavior we observed.

### DIFF
--- a/source-oracle/replication.go
+++ b/source-oracle/replication.go
@@ -174,7 +174,8 @@ var MaxReplicationLogFilesSize = 2 * 1024 * 1024 * 1024
 func (s *replicationStream) addLogFiles(ctx context.Context, startSCN, maxSCN SCN) (SCN, error) {
 	logrus.WithFields(logrus.Fields{
 		"startSCN": startSCN,
-	}).Debug("adding log files")
+		"maxSCN":   maxSCN,
+	}).Info("logminer: selecting redo log files")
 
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #4477, opened 2026-05-14
source-pendo: escape `"` and `\` in filter expressions

Alex-Bair:
**Description:**

Some users have identifiers in Pendo that contain the special character `"`, and it's suspected that not escaping `"` in these identifiers is causing issues with aggregation requests:

```
capture.Account.incremental: Encountered HTTP error status 400 which cannot be retried.
URL: https://us1.app.pendo.io/api/v1/aggregation
Response:
Cannot parse filter expression
```

This commit escapes `"` and `\` when using identifiers in aggregation requests, which should fix the issue.

Note: I saw this [Pendo doc](https://support.pendo.io/hc/en-us/articles/21326198721563-Choose-IDs-and-metadata) states:
> Don't use \ or " or invalid UTF-8 characters in an Account ID. These characters can result in errors in Pendo.

So... it's possible some other Pendo side error will occur & this won't unblock captures for users who use `\` or `"` in their account ids. If that's the case, we'll need to revisit & investigate further.

**Notes for reviewers:**

I'm unable to reproduce the issue since we don't have a dev Pendo account, so this is based purely off of the details provided in the user's bug report. The bug makes sense & definitely seems like an issue, but I wasn't able to directly confirm that the bug is what's causing the `Cannot parse filter expression` failure.

### DIFF
--- a/source-pendo/source_pendo/api.py
+++ b/source-pendo/source_pendo/api.py
@@ -39,6 +39,15 @@ def _ms_to_dt(ms: int) -> datetime:
 def base_url(host: PendoHost) -> str:
     return f"https://{host}/api/v1"
 
+def _escape_filter_string(value: str) -> str:
+    # Pendo's filter grammar is documented as "basic C expression syntax" but doesn't
+    # spell out string escape rules, so we only escape the two characters we know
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
LGTM

PR #4478, opened 2026-05-14
source-*-batch: Checkpoint between cursor values

willdonnelly:
**Description:**

Tweaks the polling readout logic in various batch SQL captures to only emit incremental-progress checkpoints in between dissimilar cursor values so that long runs of identical cursors have to be read out in full. This is the same rule we use in a bunch of other places, just applied to the batch SQL polling operations which didn't previously use it.

None, our documentation doesn't go into that level of detail regarding cursor update timing and users shouldn't really need to care, this will just work more reliably in certain edge cases.

### DIFF
--- a/source-bigquery-batch/driver.go
+++ b/source-bigquery-batch/driver.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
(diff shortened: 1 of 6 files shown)


Alex-Bair (approved):
LGTM!

Issue #4476, opened 2026-05-14
SQL Captures: Improve Discovery Scalability and Filter Options


willdonnelly:
Some people have absolutely massive SQL databases, such that our discovery logic tends to time out or OOM when trying to process everything.

Our SQL database discovery operations are typically structured as a set of "list all tables; list all columns of all tables; list all keys of all tables" queries followed by an aggregation step which combines all of those results. This is O(1) in the number of tables and thus theoretically attractive, but we're seeing the downsides of that approach here.

In the past we have implemented stopgap solutions to this by pushing down specific filter into the database discovery queries, typically in the form of a whitelist of schema names. However this isn't scalable to more fancy filtering because it requires that any filter options be something which we can push off onto the database **and** that we be able to combine them all into every discovery query.

### Listing Tables and Chunking

Part of the solution here is that we need to stop trying to fetch all info for all tables in parallel and instead we must be more deliberate about what we need and when.

Concretely, we should list all tables in the database (optionally pushing down some or all filter options into _that_ query) and then we should fetch anything else (columns, keys, indices, etc) for just the filtered list of tables which we intend to discover, possibly using chunking if required (e.g. "list columns for the next 100 tables in this list").

The upshot of this is that simply handling a list of table names isn't all that expensive even at typical scales of ~100k tables (we're talking like, 5-10MiB worst case), our filter options can basically all be applied to that list, and then the other queries don't have to care about filtering beyond table identity any more.

### Additional Filter Options

Once the fetching of full table details is decoupled from the selection of what tables we care about, we can use that to implement more powerful filtering. Currently "Discover Schemas" lives in "Advanced Options" but it already feels a bit out of place and adding two more discovery-filter settings in there will just make that worse, so the plan is:
- Add a new "Discovery Filters" section to the endpoint config with options:
  - "Include Schemas" and "Exclude Schemas": Whitelist/blacklist literal strings.
  - "Table Patterns": A list of table-name regexes. Applied with implicit `^$` anchors to the bare table names so that `foo_bar` will by default match any table named `<anyschema>.foo_bar`
  - Duplicates of connector-specific options like "Discover Unpublished Tables", where present
- Leave the existing advanced options where they are for the moment, but union the old `Discover Schemas` setting with the new `Discovery Filters > Include Schemas` for compatibility
- Plan on hiding the old option(s) and doing a bulk-edit to move any custom settings to the new location(s) in the near future.

willdonnelly:
[...]

PR #4481, opened 2026-05-14
sqlcapture+source-mysql: Discovery filtering improvements

willdonnelly:
**Description:**

Implements https://github.com/estuary/connectors/issues/4476 in the shared `sqlcapture` logic and in `source-mysql`. Other databases remain largely unchanged with only minimal changes as required to fit the new list-then-details interface.

**Workflow steps:**

Nothing should change for databases other than MySQL.

Users of `source-mysql` will see a new "Discovery Filters" section in the task configuration, where they can configure more flexible discovery selection options.

The configuration sections in all source-mysql docs should be updated:
- https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/
- https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/amazon-rds-mysql/
- https://docs.estuary.dev/reference/Connectors/capture-connectors/MySQL/google-cloud-sql-mysql/

**Notes for reviewers:**

It might help to go commit-by-commit here, this is a conceptually straightforward refactor-then-optimize-then-extend sequence and seeing everything at once makes it look more complicated than it really is.


Alex-Bair (commented):
(This is mostly a note for myself for when I take another look later). The commits through the chunked discovery queries LGTM. I held off reviewing the last commit & approving since, based off your Slack thread, it sounds like we're considering using globbing instead of regex to match table patterns.

willdonnelly:
Yeah, I think we settled on globbing so I've gone ahead and implemented that now.

I ended up adding a helper package in `go/tableglob` for that because this didn't seem worth pulling in a non-stdlib dependency for, using `filepath.Match` would have been a little awkward, and the translate-to-regex-and-compile code ended up being not too hairy.

Alex-Bair (approved):
LGTM!

PR #4482, opened 2026-05-14
source-impact-native: support millisecond datetime parsing in _s_to_dt function

SeanWhelan:
**Description:**

Fix source-impact-native cursor parsing to accept timestamps with fractional seconds. The Impact API's Tasks endpoint returns CreationDate values like 2024-07-17T21:55:46.000+0000, which _s_to_dt could not parse because its strptime format string (%Y-%m-%dT%H:%M:%S%z) had no .%f component. The helper now tries the fractional-second format first and falls back to the original, raising a clear ValueError if neither matches. Other Impact endpoints that omit fractional seconds continue to parse as before.

**Workflow steps:**

N/A

N/A

**Notes for reviewers:**

N/A

### DIFF
--- a/source-impact-native/source_impact_native/api.py
+++ b/source-impact-native/source_impact_native/api.py
@@ -412,13 +412,8 @@ async def fetch_snapshot_child(
 
 
 
-def _s_to_dt(date: str):
-    for fmt in ("%Y-%m-%dT%H:%M:%S.%f%z", "%Y-%m-%dT%H:%M:%S%z"):
-        try:
-            return datetime.strptime(date, fmt)
-        except ValueError:
-            continue
-    raise ValueError(f"unrecognized datetime format: {date}")
+def _s_to_dt(date: str) -> datetime:
+    return datetime.fromisoformat(date)
 
 def _cursor_dt(name, logcursor):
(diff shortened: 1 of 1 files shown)


source-impact-native/source_impact_native/api.py
```diff
@@ -413,8 +413,12 @@ async def fetch_snapshot_child(
...
 
 def _s_to_dt(date: str):
-    date = datetime.strptime(date, "%Y-%m-%dT%H:%M:%S%z")
-    return date
+    for fmt in ("%Y-%m-%dT%H:%M:%S.%f%z", "%Y-%m-%dT%H:%M:%S%z"):
+        try:
+            return datetime.strptime(date, fmt)
+        except ValueError:
+            continue
+    raise ValueError(f"unrecognized datetime format: {date}")
```
Alex-Bair: If these strings are always in ISO 8601 format (which seems to be true based off the two formats here?), this could be simplified with [`datetime.fromisoformat`](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat):
```suggestion
def _s_to_dt(date: str) -> datetime:
    return datetime.fromisoformat(date)
```

Alex-Bair (approved):
LGTM % a potential simplification

PR #4483, opened 2026-05-15
materialize-s3-iceberg: add nanosecond timestamps

### LINKED ISSUE #4462: Iceberg: support nanosecond timestamps
Let's add a `materialize-iceberg` task-level option to use Iceberg `timestamptz_ns` column types instead of `timestamptz`.

### Iceberg v3

[Iceberg table format v3 adds support for](https://iceberg.apache.org/spec/#version-3) nanosecond timestamps, serialized as "8-byte little-endian long".

The range of values is narrower than the more common microsecond timestamps:
- microsecond timestamps: 290,308 BCE -> 294,246 CE
- nanosecond timestamps: 1677 CE -> 2262 CE

**Blocker for the `materialize-s3-iceberg` path:** pyiceberg ≤0.11.1 (latest) cannot write to v3 tables — `write_manifest_list` has no V3 writer and `data_file_statistics_from_parquet_metadata` rejects ns-precision parquet column stats, so `Transaction.add_files` fails at append time.

### Flow

Timestamps are serialized and transferred as RFC3339 strings, whose range is 1 CE -> 9999 CE, and technically has no precision limit. The following connectors explicitly handle nanosecond precision in code:

- `source-oracle` — parses `TIMESTAMP WITH TIME ZONE` with 9-digit fractional seconds: [`replication.go:29-34`](https://github.com/estuary/connectors/blob/main/source-oracle/replication.go#L29-L34), [`main.go:60-63`](https://github.com/estuary/connectors/blob/main/source-oracle/main.go#L60-L63)
- `source-kafka` — handles Avro `TimestampNanos` and `LocalTimestampNanos` logical types: [`pull.rs:514-536`](https://github.com/estuary/connectors/blob/main/source-kafka/src/pull.rs#L514-L536)
- `materialize-snowflake` — persists to `TIMESTAMP_LTZ/NTZ/TZ` at scale 9 via nanosecond-scaled binary decimal encoding: [`bdec.go:819-832`](https://github.com/estuary/connectors/blob/main/materialize-snowflake/bdec.go#L819-L832)

Note that "parses nanos from source" (Oracle, Kafka) and "persists nanos to a ns-precision column" (Snowflake) are distinct claims — most other materializers can carry a 9-digit RFC3339 string through opaquely but truncate when writing to the destination column type.

### Out-of-range timestamp semantics

Flow's wire format allows year 0001–9999, but `timestamptz_ns` (int64 Unix nanos) is only valid 1677-09-21 → 2262-04-11. Out-of-range values are clamped to the ns min/max.

### Reader compatibility

`timestamptz_ns` requires Iceberg v3 readers. Validate writes with a **DuckDB-based unit test**.

### Schema evolution

Iceberg doesn't allow in-place promotion between `timestamptz` and `timestamptz_ns` (different physical encodings). The implementation must support **bidirectional migration** — both `timestamptz` → `timestamptz_ns` and `timestamptz_ns` → `timestamptz`.

jacobmarble:
**Description:**

Fixes https://github.com/estuary/connectors/issues/4462

Map timestamps to Iceberg `timestamptz_ns` in `materialize-s3-iceberg`.

Tables are created as format v3 when the option is enabled, and a v2 table that predates the option is upgraded to v3 in the same transaction that adds its first `timestamptz_ns` column. Out-of-range values are clamped to the int64-nanoseconds band (~1677–2262).

**Workflow steps:**

A task-level advanced option (`nanosecond_timestamps`) causes all timestamps to be materialized as `timestamptz_ns` instead of `timestamptz`. Toggling it on an existing materialization requires a backfill: Iceberg has no in-place promotion between the two timestamp encodings.

None yet — the new `nanosecond_timestamps` advanced option should be added to the materialize-s3-iceberg docs page.

**Notes for reviewers:**

Two commits: the first adds the regression tests (deliberately does not compile on its own), the second is the implementation that makes them pass. Full suite is green, including the Glue integration run.

### DIFF
--- a/docs/reference/Connectors/materialization-connectors/amazon-s3-iceberg.md
+++ b/docs/reference/Connectors/materialization-connectors/amazon-s3-iceberg.md
@@ -66,6 +66,7 @@ Estuary collections to your tables.
 | **`/region`**                     | Region                | AWS Region.                                                                                                                                         | string | Required         |
 | **`/namespace`**                  | Namespace             | Namespace for bound collection tables (unless overridden within the binding resource configuration).                                                | string | Required         |
 | `/upload_interval`                | Upload Interval       | Frequency at which files will be uploaded. Must be a valid ISO8601 duration string no greater than 4 hours.                                         | string | PT5M             |
+| `/advanced/nanosecond_timestamps` | Nanosecond Timestamps | Use nanosecond precision (Iceberg format v3) for date-time columns instead of microsecond precision (format v2).                                    | bool   | false            |
 | **`/catalog/catalog_type`**       | Catalog Type          | Either "Iceberg REST Server" or "AWS Glue".                                                                                                         | string | Required         |
 | `/catalog/glue_id`                | Glue Catalog ID       | Glue Catalog ID to use. If not specified, defaults to the account ID of the configured credentials. This enables cross-account Glue catalog access. | string |                  |
 | **`/catalog/uri`**                | URI                   | URI identifying the REST catalog, in the format of 'https://yourserver.com/catalog'.                                                                | string | Required         |
@@ -129,6 +130,13 @@ Estuary collection fields with `{type: string, format: time}` and `{type: string
 materialized as **string** columns rather than **time** and **uuid** columns for compatibility with
 Apache Spark. **[Nested types](https://iceberg.apache.org/spec/#nested-types)** are not currently
 supported.
+
+With the advanced option `nanosecond_timestamps` enabled, fields with `{format: date-time}` are
+instead materialized as **timestamptz_ns** columns with nanosecond precision, and tables are created
+using [Iceberg format v3](https://iceberg.apache.org/spec/#version-3-extended-types-and-capabilities).
+Make sure your query engine supports format v3 tables before enabling this option. Toggling it on an
+existing materialization requires a backfill of its bindings, because Iceberg has no in-place
+promotion between the two timestamp encodings.
 
 ## Table Maintenance
 
--- a/docs/reference/Connectors/materialization-connectors/amazon-s3-iceberg.md
+++ b/docs/reference/Connectors/materialization-connectors/amazon-s3-iceberg.md
@@ -136,9 +136,11 @@ instead materialized as **timestamptz_ns** columns with nanosecond precision, an
 using [Iceberg format v3](https://iceberg.apache.org/spec/#version-3-extended-types-and-capabilities).
 Make sure your query engine supports format v3 tables before enabling this option.
 
-Toggling the option on an existing materialization applies to data going forward: Iceberg has no
-in-place conversion between the two timestamp encodings, so each timestamp column is re-created
-under the new type, and existing rows read as **null** for those columns. Prior values remain
+Toggling the option in either direction on an existing materialization applies to data going
+forward: Iceberg has no in-place conversion between the two timestamp encodings, so each timestamp
+column is re-created under the new type, and existing rows read as **null** for those columns.
+Enabling the option on an existing format v2 table also upgrades the table to format v3 in place.
+Prior values remain
 readable in the table's earlier snapshots via time travel. To repopulate history under the new
 type, trigger a backfill of the binding, which re-materializes the table from the Flow collection.
 
--- a/docs/reference/Connectors/materialization-connectors/amazon-s3-iceberg.md
+++ b/docs/reference/Connectors/materialization-connectors/amazon-s3-iceberg.md
@@ -134,9 +134,13 @@ supported.
 With the advanced option `nanosecond_timestamps` enabled, fields with `{format: date-time}` are
 instead materialized as **timestamptz_ns** columns with nanosecond precision, and tables are created
 using [Iceberg format v3](https://iceberg.apache.org/spec/#version-3-extended-types-and-capabilities).
-Make sure your query engine supports format v3 tables before enabling this option. Toggling it on an
-existing materialization requires a backfill of its bindings, because Iceberg has no in-place
-promotion between the two timestamp encodings.
+Make sure your query engine supports format v3 tables before enabling this option.
+
+Toggling the option on an existing materialization applies to data going forward: Iceberg has no
+in-place conversion between the two timestamp encodings, so each timestamp column is re-created
+under the new type, and existing rows read as **null** for those columns. Prior values remain
+readable in the table's earlier snapshots via time travel. To repopulate history under the new
+type, trigger a backfill of the binding, which re-materializes the table from the Flow collection.
 
 ## Table Maintenance
 
--- a/materialize-s3-iceberg/catalog.go
+++ b/materialize-s3-iceberg/catalog.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"slices"
 	"time"
 
 	"github.com/apache/iceberg-go"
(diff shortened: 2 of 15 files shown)


materialize-s3-iceberg/catalog.go
```diff
@@ -398,9 +399,15 @@ func (c *catalog) CreateResource(ctx context.Context, b *pf.MaterializationSpec_
...
+		}
 		for k, v := range res.AdditionalTableProperties {
-			// Never let a user-supplied key override the name mapping.
-			if k == icebergtable.DefaultNameMappingKey {
+			// Never let a user-supplied key override the name mapping, or
+			// downgrade the format version below what timestamptz_ns columns
+			// require.
+			if k == icebergtable.DefaultNameMappingKey ||
+				(k == icebergtable.PropertyFormatVersion && c.cfg.nanosecondTimestamps()) {
 				continue
```
danielnelson: I could see situations where this becomes confusing to users, especially the more complicated our rules become.  Consider if we need to return an error or log a warning when changes are made.

docs/reference/Connectors/materialization-connectors/amazon-s3-iceberg.md
```diff
@@ -66,6 +66,7 @@ Estuary collections to your tables.
 | **`/region`**                     | Region                | AWS Region.                                                                                                                                         | string | Required         |
 | **`/namespace`**                  | Namespace             | Namespace for bound collection tables (unless overridden within the binding resource configuration).                                                | string | Required         |
 | `/upload_interval`                | Upload Interval       | Frequency at which files will be uploaded. Must be a valid ISO8601 duration string no greater than 4 hours.                                         | string | PT5M             |
+| `/advanced/nanosecond_timestamps` | Nanosecond Timestamps | Use nanosecond precision (Iceberg format v3) for date-time columns instead of microsecond precision (format v2).                                    | bool   | false            |
```
danielnelson, line 69: I think we need to add documentation to the flow repo for now.  Check the link I added to our discussion couple days back on this.

materialize-s3-iceberg/testdata/config-glue-ns.yaml
```diff
@@ -0,0 +1,36 @@
...
+#ENC[AES256_GCM,data://L3Z9nGP/vCyUaiatkDvnXnaLCoOOlk+XnfHhd5h5G7CPr6JQda0uqI8CqlBERVtzaUVenySw/xoZVvpMeMrq2hp0hXgEHb+HmI,iv:m202PXu+Z+V6WzYeY8PcWuX4fVJpPsD9oNfG/nASzQw=,tag:OMdzm/im8U9aR05XwBBuTg==,type:comment]
+#ENC[AES256_GCM,data:PdvcV+YjlVZIUZnmeOWkPLOQNKaatDtXVHh5S6qJ8kxMyDoCGfvD,iv:8Gv5N9XFSF6/8RDniwqmCkc/UHhOUAZbZQvaw3iGZ/Q=,tag:vR++naj9fPQrJ4C2lmsH1w==,type:comment]
+#ENC[AES256_GCM,data:eW96lfGNpmDyvDcblQwRJ6dCsdMybORdzcBjEP5xE/ZwT14mn1Z5bz5SP7iHmV12JY4aHrvSTdlSi50M3t0voFUmxMWGQt34Y/CbZ4/2Pw==,iv:cqWzipUdCD7lH+wtFvve3eh8qDwTLAwNaU2HlxJmAjI=,tag:2/mb/Q7l607FEhgHPxTvNg==,type:comment]
+#ENC[AES256_GCM,data:5BaBzeocVZfoYUjMnpog9TZ1jiKOlC4dL6QS1QK3ycXCX9yjm8na7F9zMemEP5qWSiVDic/uCeZ3lg975uTPcOuBryYaGKfKKv2N5GM72Co=,iv:phti6mz+CLMkrnaVhyKDGnmgvxZzwV5RpNjrbWU6i5c=,tag:yXNEbAUmLFFZcFMFDZwWWg==,type:comment]
+#ENC[AES256_GCM,data:4385uqgBBsLOFMHTYQVcJd6XtFqMFSLEduTG7VfzI6HF4Tlm9gsvySC5Wvjx6fEDZXRRJDJ7QaI=,iv:a3xF9mh0YexSkQaR+z0zO0pgf4BUSTnkuge/c8F46M0=,tag:Jcvf6zXox00RiH7mwUobHw==,type:comment]
```
danielnelson: Maybe we should use `encrypted_suffix: _sops` so the comments and other non-sensitive fields will be visible in pull requests.

danielnelson (dismissed):

mdibaiee:
@jacobmarble is it possible for a customer to attempt this transition, but their iceberg does not support v3 and they get stuck?

docs/reference/Connectors/materialization-connectors/amazon-s3-iceberg.md
jacobmarble: Done https://github.com/estuary/flow/pull/3199

jacobmarble:
> @jacobmarble is it possible for a customer to attempt this transition, but their iceberg does not support v3 and they get stuck?

Yes, but "stuck" is resolved by switching the feature off and triggering another backfill. Switching nanoseconds on requires a backfill anyhow.

mdibaiee:
@jacobmarble this kind of fallible upgrades shouldn't happen automatically on toggling of a feature, because the user may be unaware of the necessity of a backfill (even if we write docs or input description), and a backfill can be catastrophic for a customer with large tables where the tables are production load-bearing.

So the pattern we use is usually one where if a backfill is requires for this toggle to work, we expect and require the user to both toggle this checkbox, AND trigger a backfill themselves. If the user toggles this checkbox but does not trigger a backfill, we error out telling the user about this requirement. This way users won't unknowingly get stuck with a truncated/re-created table

I think more generally, we need to consider backfills as potentially catastrophic data loss for users who really don't want their data to be lost.

It's also useful to think about how this interacts with `retain_data_on_backfill`

jacobmarble:
> @jacobmarble this kind of fallible upgrades shouldn't happen automatically on toggling of a feature, because the user may be unaware of the necessity of a backfill (even if we write docs or input description), and a backfill can be catastrophic for a customer with large tables where the tables are production load-bearing.
> 
> So the pattern we use is usually one where if a backfill is requires for this toggle to work, we expect and require the user to both toggle this checkbox, AND trigger a backfill themselves. If the user toggles this checkbox but does not trigger a backfill, we error out telling the user about this requirement. This way users won't unknowingly get stuck with a truncated/re-created table
> 
> I think more generally, we need to consider backfills as potentially catastrophic data loss for users who really don't want their data to be lost.
> 
> It's also useful to think about how this interacts with `retain_data_on_backfill`

The weekend cleared my mind and now I realize that I was wrong to suggest backfill as necessary here.

A given Iceberg table can have multiple schemas. Enabling nanoseconds on an existing table should be very simple:
- change the table format from v2 to v3 (if necessary)
- create a new schema, replacing the microseconds column(s) with nanoseconds column(s)
- resume writing to the new schema

Iceberg query/compute engines may or may not be OK with this. If the customer needs all of the timestamp data converted to nanoseconds, then they can backfill, but otherwise it's a very quick metadata change.

I'll make this change.

jacobmarble:
> this kind of fallible upgrades shouldn't happen automatically on toggling of a feature

Resolved by removing the fallible upgrade entirely. Toggling `nanosecond_timestamps` no longer implies a backfill: the connector now migrates via Iceberg schema evolution (drop + re-add the column under a new field ID, in the same transaction as the v3 upgrade). It's metadata-only and non-destructive: existing table data is untouched, rows read `null` for the migrated column going backward, and new writes carry the new type. A backfill remains available as an explicit, user-initiated choice for repopulating history — never something the toggle does implicitly. This also sidesteps the `retain_data_on_backfill` interaction since no backfill occurs.

mdibaiee:
> existing table data is untouched, rows read null for the migrated column going backward, and new writes carry the new type.

this still seems _destructive_ in the sense that the customer loses the ability to access their existing data, and that is also can break their downstream consumers of the data. I can't think of a regular case where a customer would be okay with not being able to read the data of the column before the switch 🤔 it seems inevitable that they would then want a backfill so they can have a full view of their data

mdibaiee:
Had a chat offline with @jacobmarble -- my understanding of Iceberg was insufficient. Now my understanding is that Iceberg retains the data as it was, and there will be multiple schemas that exist for the table (the primary one being the new v3 schema with nanoseconds), but the readers of the table can choose to query both the new schema and the previous schema to have the full view of the data with pre-nanosecond timestamps and nanosecond timestamps. 

This can be a real use case where a customer with large tables may want to retain their table data and not do a backfill, and still switch to nanoseconds.

mdibaiee (approved):

PR #4484, opened 2026-05-15
source-pendo: add debug logging before aggregation API requests

Alex-Bair:
**Description:**

We've received a report of a source-pendo capture failing with the error:
```
runTransactions: readMessage: Task capture.Account.incremental: Encountered HTTP error status 400 which cannot be retried.
URL: https://us1.app.pendo.io/api/v1/aggregation
Response: Cannot parse filter expression
```

We suspect this failure is due to the connector not escaping `"` or `\` in the filter expression, and these debug logs will help confirm whether or not that's the issue.


nicolaslazo (approved):
👌

PR #4488, opened 2026-05-15
source-mysql: Add skip_prerequisites and skip_table_prerequisites

willdonnelly:
**Description:**

Adds two feature flags as escape hatches for cases where the connector's prerequisite checks should be skipped.

**Notes for reviewers:**

I have no reason to believe that we need this right now, but we recently introduced these flags for source-postgres and we have a user doing a POC right now which could plausibly struggle with the per-table validation work, so it seems reasonable to just go ahead and add them for symmetry and do it now just in case we need the escape hatch.


Alex-Bair (approved):
LGTM

PR #4490, opened 2026-05-15
source-oracle: Per-thread log file selection for RAC

willdonnelly:
**Description:**

In Oracle RAC, each instance writes its own thread of redo, and SCN ranges across threads overlap arbitrarily while sequence numbers are independent. The previous log selection logic assumed a single thread, and could misbehave in all kinds of interesting ways when that wasn't true.

The new selection logic sorts candidate files by `FIRST_CHANGE#` and tracks `NEXT_CHANGE#` per-thread then takes the minimum across threads to find the global covered SCN upper-bound after adding each file. This logic reduces to the old behavior in single-threaded cases.

I have not tested this against a real RAC deployment but I believe that it should work in theory, and as previously mentioned it reduces to the single-thread logic the rest of the time so I believe it should be safe to try in production.

### DIFF
--- a/source-oracle/replication.go
+++ b/source-oracle/replication.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"cmp"
 	"context"
 	"database/sql"
 	"encoding/base64"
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #4494, opened 2026-05-18
source-jira-native: support scheduled backfills

Alex-Bair:
**Description:**

Some users have expressed interest in scheduling backfills for various resources for `source-jira-native`, mostly for the streams that the connector isn't inferring/capturing deletions for (ex: `issues`, `issue_coments`, etc.). The idea is that the user will be able to infer that any documents _not_ captured since the most recent backfill have been deleted & can be manually marked with `_meta/op = "d"` in their destination.

This commit updates the connector to use `ResourceConfigWithSchedule` to support these uses cases. In the (possibly near?) future, our platform plans on suporting this type of deletion inference & schedule backfills automatically, and using `ResourceConfigWithSchedule` is directionally correct to support that longer term goal.

The connector's documentation should be updated to reflect the new `schedule` field on resource configs.

### DIFF
--- a/source-jira-native/source_jira_native/__init__.py
+++ b/source-jira-native/source_jira_native/__init__.py
@@ -17,35 +17,35 @@ from .resources import all_resources, validate_credentials, validate_projects
 from .models import (
     ConnectorState,
     EndpointConfig,
-    ResourceConfig,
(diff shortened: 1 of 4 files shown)


nicolaslazo (approved):
LGTM

PR #4497, opened 2026-05-19
materialize-redshift: add is not null to update statement

danielnelson:
When updating a column with many rows and null values, we ran into the error: `Invalid input (SQLSTATE XX000)`.

With psql the error was:
```
error:  Invalid input
  code:      8001
  context:   String value exceeds the max size of 65535 bytes
```

This was odd since the column only contained several short simple strings:
```
 column    |   count
-----------+-----------
           | 426004506
 active    |   7177705
 unknown   |   2428474
 inactive  |   2459190
```

By skipping rows that are null the statement was able to complete.  In order to add the where clause I broke out the update statements into one per column.

**Workflow steps:**

No changes

None

**Notes for reviewers:**


mdibaiee:
@danielnelson breaking the queries into one per column can be devastatingly slow for large table migrations, we have previouly hit this, when a customer for example has updates to a large number of Salesforce fields, and the materialization chokes on trying to migrate all of them. It is important to do all of this in one query so there is a single table scan pass.

jacobmarble:
> In order to add the where clause I broke out the update statements into one per column.

> It is important to do all of this in one query so there is a single table scan pass.

We'll need a more complicated where clause, but it should be possible to consolidate into a single update statement.

danielnelson:
I updated the query pattern to be like:
```sql
UPDATE mytable
SET
    {1} = {2},
    {3} = {4},
    {5} = {6},
WHERE
    {2} IS NOT NULL
    OR {4} IS NOT NULL
    OR {6} IS NOT NULL;
```

It seems possible that we could get stuck in a situation where it fails together but going one at a time would work.  If we have problems with this in the future we could consider an advanced option controlling how many columns to update per query.

jacobmarble:
> It seems possible that we could get stuck in a situation where it fails together but going one at a time would work. If we have problems with this in the future we could consider an advanced option controlling how many columns to update per query.

I agree.

Also this doesn't help at all where one column is never null. 🤔

danielnelson:
It isn't fully clear why `is not null` allowed the update query so I am going to close this PR until we have better understanding of the root cause.

PR #4506, opened 2026-05-19
materialize-clickhouse: query parts with sequential_consistency

danielnelson:
Since the clickhouse Conn is a connection pool, we may have switch replicas since writing the partitions.  In this case the newly written partitions may not be visible without this setting.

**Workflow steps:**

No changes

None

**Notes for reviewers:**

I looked to see if we can easily get a single connection to a single replica, similar to if we called `sql.DB.Conn()`, but didn't see an way to do this.


jacobmarble (approved):
I'm sure this was fun to diagnose

jacobmarble:
**Post-incident audit note (2026-06-05)** — recording findings here since this PR is the closest prior art to the data-loss incident investigated this week.

**What this PR got right.** The diagnosis in the description is exactly correct: the `Conn` pool can switch replicas between writing staged partitions and querying `system.parts`, making freshly written parts invisible to the commit path. A customer incident (Salesforce → ClickHouse Cloud materialization, ~Jun 2–4) confirmed this mechanism in production: `system.query_log` showed 107 stage-table INSERTs but only 44 `MOVE PARTITION` statements over 48h — for every lost transaction the INSERT succeeded and the MOVE was never issued, i.e. `queryStoreParts` returned an empty set and the subsequent `DROP` destroyed the staged rows. Roughly half of all transactions in that window were lost, silently, with clean task logs.

**Why this mitigation turned out to be ineffective** (losses were observed both before and after this change shipped in the 2026-05-21 build):

1. `select_sequential_consistency` applies to reads *of replicated/SharedMergeTree tables*. The query here reads `FROM system.parts` — a virtual system table that dumps the replica's local in-memory part registry. There is no replicated table in the `FROM` clause to synchronize, so the setting is effectively a no-op for this statement.
2. Even where the setting does force a metadata sync, it covers *data parts within a table the replica knows about* — not DDL. The stage table is `CREATE OR REPLACE`'d every transaction, producing a new table UUID each time. A replica lagging on the DDL log resolves the stage-table *name* to the previous incarnation (already moved + dropped), for which zero active parts is a "correct" answer. Independent corroboration: the same cluster surfaced `code 517: metadata version on replica is 0, while common metadata is 1` on an `ALTER` during the incident window — ClickHouse detects the lag for DDL, but a `system.parts` scan just answers stale.

**Verification.** A detection build (#4623) deployed to the affected task on 2026-06-04 caught the race live on 2026-06-05 12:32:37Z — `refusing to commit store table: system.parts reports 0 staged rows but 1 rows were stored` — and the preserved stage table was committed successfully on retry three minutes later. Zero rows lost with the guard in place.

**Follow-up.** Re the reviewer note ("didn't see a way to get a single connection to a single replica"): setting `MaxOpenConns: 1, MaxIdleConns: 1` on `clickhouse.Options` for the transactor's load/store `Conn`s achieves exactly the `sql.DB.Conn()`-style pinning wanted here — every statement of the temp-table lifecycle then runs on one session, which always observes its own DDL and inserts. That fix, plus a hard row-accounting check (refuse to MOVE/DROP unless `system.parts` accounts for every staged row), is in #4623 and its follow-on commit.

jacobmarble:
https://github.com/estuary/connectors/pull/4623 has been merged as a stopgap.

---

The `select_sequential_consistency` query setting is useful for querying MergeTree tables with strong consistency, so it can be used to query the staging tables directly. However, that staging table is `CREATE OR REPLACE`'d with every transaction, so the in-memory system tables can easily have a stable **table reference**, meaning **a query could read the wrong table with strong consistency**.

The obvious mitigation for stale table references is to truncate, rather than drop, the staging tables; the system reference to the staging table is stable. In our case, truncation shouldn't be necessary because when we've completed a transaction, the stage table *should* be empty. We can confirm this with:

```sql
SELECT _partition_id, count() FROM <stage_table>
GROUP BY _partition_id
SETTINGS select_sequential_consistency = 1;
```

Or even just:

```sql
SELECT count() FROM <stage_table>
SETTINGS select_sequential_consistency = 1;
```

PR #4509, opened 2026-05-19
sqlcapture: Periodic progress logging

willdonnelly:
**Description:**

During normal operation we log `processed replication events` every 60 seconds or thereabouts as stream-to-fence cycles complete. But if things go wrong and a large WAL backlog starts to accumulate we just sit there silently, potentially for hours. This is not ideal.

This PR adds an every-5-minutes ticker to drive periodic progress reporting during long-running stream-to-fence cycles. I've also gone ahead and moved the replication diagnostics into that same ticker, since AFAIK there was never any principled reason to only run it once and if streaming takes several hours to finish it would be really nice if we could get updated diagnostics periodically as well.

I've also thrown in another diagnostic logging change for source-postgres to keep a tally of ignored streams for which it's receiving change events, and to log that information as part of replication diagnostics. One way that streaming cycles can get bogged down is if the connector is being spammed with a ton of changes on tables we're not capturing and this can help to surface that information.

### DIFF
--- a/source-postgres/diagnostics.go
+++ b/source-postgres/diagnostics.go
@@ -3,8 +3,10 @@ package main
 import (
 	"context"
 	"fmt"
+	"slices"
 	"time"
 
+	"github.com/estuary/connectors/sqlcapture"
 	"github.com/jackc/pgx/v5/pgxpool"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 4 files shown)


Alex-Bair (approved):
LGTM

PR #4520, opened 2026-05-19
materialize-iceberg: don't add required columns to existing tables

jacobmarble:
**Description:**

Fixes Snowflake's `errno=100478, "non-nullable column without default missing data"` on iceberg materializations whose source collection inferred a new MUST-exist field (e.g. `html_body`) after the destination already had data.

`computeSchemaForUpdatedTable` writes new fields with `Required: p.MustExist || p.IsPrimaryKey`. Pre-existing data files have no data for the new field id, so their manifest entries have no `null_value_counts` for it. Spark/pyiceberg fill the missing id with null on read; Snowflake refuses. The fix forces `Required: false` on every field appended by `computeSchemaForUpdatedTable` (both `NewProjections` and `tempMigrateProjections`).

**Workflow steps:**

No user-visible change. Inferred-MUST fields added by schema evolution now land as `optional`. Fields added at table-creation time still take `required` from `MustExist`. Bump the binding's `backfill` counter to recreate the table if `required` is desired post-evolution.

None.

**Notes for reviewers:**

Test-before-fix per the TDD convention:
- `2d05b953f` - failing regression for the add-required-column case
- `c5c9355a9` - demote direction passes, confirming `NewlyNullableFields` is safe
- `39e0fd441` - refactors the add-test to drive `computeSchemaForUpdatedTable` directly
- `434daf0ca` - fix + `TestComputeSchemas` snapshot update (migration temp columns now optional too)

Reproducer: `TestIntegration/add-required-column-regression`. Asserts conditionally - if the connector emits the field as `required`, the manifest must carry `null_value_counts` for it; post-fix the field is `optional` and the assertion is vacuously satisfied.

### DIFF
--- a/materialize-iceberg/.snapshots/TestComputeSchemas
+++ b/materialize-iceberg/.snapshots/TestComputeSchemas
@@ -24,7 +24,7 @@ table {
 	13: new: optional string
 	14: dateToStr_flow_tmp: optional string
 	15: intToDecimal_flow_tmp: optional decimal(38, 0)
-	16: decimalToFloat_flow_tmp: required double
+	16: decimalToFloat_flow_tmp: optional double
 	17: timestampToStr_flow_tmp: optional string
 }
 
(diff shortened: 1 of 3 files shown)


jacobmarble:
Closing for now while we reconsider the approach. Branch left as-is locally.

PR #4533, opened 2026-05-21
materialize-boilerplate: enforce root document location for new bindings over existing tables

jacobmarble:
**Description:**

Closes the publish-time gap reported by a customer where `materialize-bigquery` published bindings with `FieldSelection.Document = ""` and silently degenerated to INSERT-only writes (duplicate rows on every update).

Root cause: when a new binding is adopted over an existing table via `allow_existing_tables_for_new_bindings` and the existing `flow_document` column has an incompatible type, `validateMatchesExistingResource` returned a constraint set with `flow_document=INCOMPATIBLE` and either `FIELD_FORBIDDEN` or nothing on alternate root projections. The control plane drops bare `INCOMPATIBLE` in the no-live-spec path ([`field_selection.rs:269-287`](https://github.com/estuary/flow/blob/master/crates/validation/src/field_selection.rs#L269-L287)) and has no `LOCATION_REQUIRED` signal for the root pointer, so the publish proceeds with no document column.

Two changes, gated on `lastBinding == nil`:
- Emit `LOCATION_REQUIRED` on alternate root projections instead of `FIELD_FORBIDDEN`, so the control plane has the protocol-level signal documented at [`proto-flow/src/materialize.rs:359-364`](https://github.com/estuary/flow/blob/master/crates/proto-flow/src/materialize.rs#L359-L364).
- Hard-error from the validator when no root projection ends up with a selectable constraint (single-root spec with incompatible existing column — the customer's actual case).

**Workflow steps:**

No user-facing workflow change. Affected publishes that previously succeeded silently with broken bindings will now either select a viable alternate root projection or fail at publish with a clear "incompatible column" error.

None.

**Notes for reviewers:**

Four commits, paired (test then fix). Each test fails at its own commit and is unlocked by the next commit. Pre-existing snapshot updates for five scenarios where `second_root` previously came back `FIELD_FORBIDDEN`; now `LOCATION_REQUIRED`.

The `lastBinding != nil` paths (`cannot_add_new_root_document_without_backfill`, `change_root_document_projection_for_standard_updates`) are intentionally untouched — they continue to return constraints and rely on the control plane to surface conflicts.

### DIFF
--- a/materialize-boilerplate/.snapshots/TestValidate
+++ b/materialize-boilerplate/.snapshots/TestValidate
@@ -125,7 +125,7 @@ increment backfill counter for disabled -> enabled binding:
 
 table already exists with identical spec:
 {"Field":"_meta/flow_truncated","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"The projection has a single scalar type"}
-{"Field":"flow_document","Type":1,"TypeString":"FIELD_REQUIRED","Reason":"This field is the document in the current materialization"}
+{"Field":"flow_document","Type":2,"TypeString":"LOCATION_REQUIRED","Reason":"The root document is required for a standard updates materialization"}
 {"Field":"flow_published_at","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
 {"Field":"key","Type":1,"TypeString":"FIELD_REQUIRED","Reason":"This field is a key in the current materialization"}
 {"Field":"multiple","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
@@ -138,7 +138,7 @@ table already exists with identical spec:
 
 table already exists with incompatible proposed spec:
 {"Field":"_meta/flow_truncated","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"The projection has a single scalar type"}
-{"Field":"flow_document","Type":1,"TypeString":"FIELD_REQUIRED","Reason":"This field is the document in the current materialization"}
+{"Field":"flow_document","Type":2,"TypeString":"LOCATION_REQUIRED","Reason":"The root document is required for a standard updates materialization"}
 {"Field":"flow_published_at","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
 {"Field":"key","Type":1,"TypeString":"FIELD_REQUIRED","Reason":"This field is a key in the current materialization"}
 {"Field":"multiple","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"This location is part of the current materialization"}
--- a/materialize-boilerplate/validate.go
+++ b/materialize-boilerplate/validate.go
@@ -205,7 +205,25 @@ func (v Validator) validateMatchesExistingResource(
 			// Only the originally selected root document projection is allowed to be selected for
 			// changes to a standard updates materialization. If there is no previously persisted
 			// spec, the first root document projection is selected as the root document.
-			if (lastBinding != nil && p.Field == lastBinding.FieldSelection.Document) || (lastBinding == nil && len(docFields) == 1) {
+			if lastBinding == nil && len(docFields) == 1 {
+				// New binding adopted via `allow_existing_tables_for_new_bindings`. Emit
+				// LOCATION_REQUIRED so the control plane can enforce the root document slot
+				// (proto-flow/src/materialize.rs:359-364). Returning INCOMPATIBLE here when
+				// the existing column type doesn't match would silently drop in the
+				// no-live-spec path (validation/src/field_selection.rs:269-287), letting the
+				// user publish with no document column and INSERT-only data corruption.
+				//
+				// A document-column type mismatch is not surfaced at validate time anymore
+				// in this case; the connector's Apply / runtime path will surface it instead.
+				// See materializer.go:654-657 — that loop does not check the Document column,
+				// so users with an incompatible flow_document column will see a runtime
+				// error rather than a publish-time one. A follow-up should extend the Apply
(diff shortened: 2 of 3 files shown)


materialize-boilerplate/validate_test.go
```diff
@@ -265,6 +265,80 @@ func TestValidate(t *testing.T) {
...
+			return out
+		}
+		is := NewInfoSchema(transformPath, simpleTestTransform, simpleTestTransform, false, false)
+		res := is.PushResource(transformPath([]string{"key_value"})...)
+		for _, f := range binding.FieldSelection.AllFields() {
+			proj := *binding.Collection.GetProjection(f)
+			fieldType := strings.Join(proj.Inference.Types, ",")
+			if f == binding.FieldSelection.Document {
+				fieldType = "string"
+			}
```
danielnelson, line 307: This gives the document an existing field with the wrong type, so it feels like we can't use it as the document.  It seems then if we say the location is required that field selection will accept it as-is and it will fail during load?

materialize-boilerplate/validate_test.go
```diff
@@ -265,6 +265,80 @@ func TestValidate(t *testing.T) {
 		require.NotNil(t, cs)
 	})
 
+	t.Run("new binding with existing table - single root projection with incompatible column", func(t *testing.T) {
+		// Regression test for the customer's actual collection shape: only flow_document
+		// is declared as a root projection. Per the protocol contract documented at
+		// flow/crates/proto-flow/src/materialize.rs:359-364, connectors materializing the
```
danielnelson: I'm uneasy about comments with line references to another repo, it feels like its going to get out of sync.  Maybe try to reword the comments towards the expected contract as best you can?  I know this is easier said than done...

materialize-boilerplate/validate.go
```diff
@@ -204,8 +211,48 @@ func (v Validator) validateMatchesExistingResource(
...
+					selectedRootField = p.Field
+				} else {
+					c = &pm.Response_Validated_Constraint{
+						Type: pm.Response_Validated_Constraint_FIELD_FORBIDDEN,
+						Reason: fmt.Sprintf(
+							"Cannot materialize root document projection '%s' because field '%s' is already being materialized as the document",
+							p.Field,
+							selectedRootField,
+						),
+					}
```
danielnelson, line 253: Should we return `LOCATION_REQUIRED` for all root document projections that are acceptable?  I think then the runtime will pick one of them to be the FieldSelection.Document in Apply and Open.

Possible counterargument: could this make which field is selected unpredictable and if so would this be a problem?

materialize-boilerplate/validate.go
```diff
@@ -283,6 +324,17 @@ func (v Validator) validateMatchesExistingResource(
 		constraints[p.Field] = c
 	}
 
+	if lastBinding == nil && !deltaUpdates && len(docFields) > 0 && selectedRootField == "" {
```
danielnelson, line 327: question: `len(docFields) == 0` when `no_flow_document`?

danielnelson (approved):

materialize-boilerplate/validate.go
mdibaiee: @danielnelson I don't think so: `no_flow_document` is transparent to control-plane, so I think control plane anyways selects a root document projection (e.g. `flow_document`), and every document will have this `flow_document` sent by the runtime, as well as the connector has to send back this `flow_document` in Load... so from the perspective of the runtime and the control plane, `flow_document` IS there, it is just _constructed_ by us rather by being a column in the database

mdibaiee: Actually, on checking further, yes... `len(docFields)` is most likely going to be zero on `no_flow_document`, because in [materialize-sql](https://github.com/estuary/connectors/blob/main/materialize-sql/driver.go#L207-L210) we emit `FIELD_OPTIONAL` for the root document projection if `no_flow_document` is enabled.

The user may still choose to include it but most likely it is going to be zero

materialize-boilerplate/.snapshots/TestValidate
```diff
@@ -125,7 +125,7 @@ increment backfill counter for disabled -> enabled binding:
 
 table already exists with identical spec:
 {"Field":"_meta/flow_truncated","Type":3,"TypeString":"LOCATION_RECOMMENDED","Reason":"The projection has a single scalar type"}
-{"Field":"flow_document","Type":1,"TypeString":"FIELD_REQUIRED","Reason":"This field is the document in the current materialization"}
+{"Field":"flow_document","Type":2,"TypeString":"LOCATION_REQUIRED","Reason":"The root document is required for a standard updates materialization"}
```
mdibaiee, line 128: If there is an existing flow_document with the name `my_document`, on a subsequent apply to the same materialization, we expect that specific `my_document` to be present, rather than the root projection existing _somewhere_.

`LOCATION_REQUIRED` means that the location as it relates to the collection itself (in case of `flow_document`, that is the "root" location, which is essentially the path `/` -- the top-level root of all properties of a collection)

`FIELD_REQUIRED` means that the specific column name that is being presented (in this case `my_document` specifically) must exist.

I think when re-applying a materialization, we want the same root projection to exist. i.e. if a root projection has changed between two Applys, we can't validate that: it means that so far we have materialized the root projection into `my_document`, and on a next apply we might be materializing it into `some_other_field_name`, which is not something we support AFAIK. At least not explicitly, although _it might_ just work.

materialize-boilerplate/validate_test.go
mdibaiee: @danielnelson Yes, I think this is the reason Johnny said we need to present both constraints to the control plane: This field is INCOMPATIBLE as-is, and at the same time, it is REQUIRED, which ought to tell the control plane: instead of silently skipping this incompatible field, surface an error that says `field X is incompatible, but required`

mdibaiee:
This issue has been fixed by #4648

PR #4537, opened 2026-05-22
materialize-postgres: support ENUM column type

### LINKED ISSUE #4370: Support materializing enum fields in materialize-postgres
JSONSchema already has the concept of enums. In postgres we can map it to the postgres enum type.

We could additonally do this for other connectors if it is readily possible.

mdibaiee:
**Description:**

- Add support for materializing `type: string, enum: []` values in postgres
- Supports extending enums to add new values, as well as migrating from string to enum
- Added test cases to cover this

**Workflow steps:**

**Notes for reviewers:**


materialize-boilerplate/info_schema.go
```diff
@@ -63,12 +67,12 @@ func (r *ExistingResource) PushField(field ExistingField) {
...
-		if r.caseInsensitiveFields && strings.EqualFold(f.Name, translated) {
-			return &f
-		} else if f.Name == translated {
-			return &f
+	translated := r.translateField(name)
+	for i := range r.fields {
+		if r.caseInsensitiveFields && strings.EqualFold(r.fields[i].Name, translated) {
+			return &r.fields[i]
+		} else if r.fields[i].Name == translated {
+			return &r.fields[i]
```
mdibaiee, line 75: This change allows us to get a mutable pointer to the underlying field so that we can modify it, for example to set a custom `Meta` on it

materialize-postgres/enum_type.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+
+		data, err := json.Marshal(p.Inference.EnumJsonVec)
+		if err != nil {
+			return fallback(p)
+		}
+		var values []string
+		if err := json.Unmarshal(data, &values); err != nil {
+			return fallback(p)
+		}
```
danielnelson, line 105: These should never error so perhaps we should panic here?

materialize-postgres/enum_type.go
```diff
@@ -0,0 +1,211 @@
...
+	base = tableName + "_" + field + "_flow_enum"
+
+	if len([]byte(base)) > 63 {
+		// Hash mode: <tableName>_<truncatedField>_<8hexHash>_flow_enum
+		// Fixed overhead (2 separators + 8-char hash + "_flow_enum"): 20 bytes.
+		// Cap tableName to 42 bytes so the field portion gets at least 1 byte.
+		const maxTableBytes = 42
+		tableName = truncateToBytes(tableName, maxTableBytes)
+		h := sha256.Sum256([]byte(tableName + "_" + field))
+		hash := fmt.Sprintf("%x", h[:4])
```
danielnelson, line 158: I suppose it is possible that we could have a naming conflict with this code. Perhaps we should use the full untruncated tableName as hash input and an encoding with more values to further reduce the odds of this?

Another thought is that maybe we can detect and surface if this does happen.

materialize-postgres/client.go
```diff
@@ -158,6 +222,40 @@ func (c *client) TruncateTable(ctx context.Context, path []string) (string, boil
...
+
+	// ALTER TYPE ADD VALUE for existing enum columns
+	for _, col := range ta.Table.Columns() {
+		e, ok := col.MappedType.TargetType.(*PgEnum)
+		if !ok || e.TypeName == "" {
+			continue
+		}
+		for _, v := range e.Values {
+			stmts = append(stmts, fmt.Sprintf(`ALTER TYPE %s ADD VALUE IF NOT EXISTS %s`,
+				e.TypeName, c.ep.Dialect.Literal(v)))
```
danielnelson, line 255: Do you think this could end up being a lot of queries in practice?  What do you think the upper limit might be for number of enum variants?  Do we need to detect what is new?

materialize-postgres/enum_type_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Long field name triggers hash mode
+	longField := strings.Repeat("x", 50)
+	schema, base, typeName, err := enumTypeNameParts(dialect, []string{"public", "orders"}, longField)
+	require.NoError(t, err)
+	require.Equal(t, "public", schema)
+	require.LessOrEqual(t, len([]byte(base)), 63, "base must fit within Postgres 63-byte identifier limit")
+	require.True(t, strings.HasSuffix(base, "_flow_enum"), "base must end with _flow_enum")
+	require.NotEmpty(t, typeName)
+	_ = schema
```
danielnelson:
```suggestion
```

materialize-sql/migration_mapping.go
```diff
@@ -92,3 +92,4 @@ func WithCastSQL(castSQL CastSQLFunc) MigrationSpecOption {
 		m.CastSQL = castSQL
 	}
 }
+
```
danielnelson:
```suggestion
```

danielnelson (approved):

materialize-postgres/client.go
mdibaiee: I think it is unlikely we will have customers with _many_ possible enumeration values... but almost always as we make these assumptions they are broken by a customer 😁 

so probably at some stage it will be worth it to filter out only new ones. I initially implemented it actually, but removed it later, because the code had just too much complexity for such a simple feature (enums)... so I tried to simplify it a lot

Let's see how far we can live with this simpler code

materialize-postgres/enum_type.go
mdibaiee: The chance of collision is really slim here (0.000001%), given that two fields are actually long enough and share a lot in common, they are still not likely to collide. I left the 4 characters (32 bits) of hash but I made sure the hash is calculated on the full tableName which is an important oversight 👍🏽

PR #4543, opened 2026-05-26
source-oracle: fix deadlock when making recursive queries

Alex-Bair:
**Description:**

`source-oracle` has been observed deadlocking when capturing a multi-window transaction. The outer LogMiner cursor and the recursive `capturePendingTransaction` poll share the same pinned `*sql.Conn`, and running `END_LOGMNR` / `ADD_LOGFILE` / `START_LOGMNR` on it while the outer cursor's `Rows` is still streaming eventually trips `driver.ErrBadConn`. `database/sql`'s cleanup of the bad connection then blocks acquiring the writer lock on the connection's `closemu`, which the outer `Rows` still holds as an `RLock`. The single goroutine driving both ends up waiting on itself, stuck in permanent self-deadlock.

This PR's scope includes three commits:
1. Refactor all helper functions to accept an `*sql.Conn` parameter instead of always reading `s.conn` off the receiver. This is a refactor that makes no behavioral changes in order to more clearly illustrate the behavioral changes in the next commit.
2. Use a separate `*sql.Conn` (`innerConn`) for `capturePendingTransaction`'s recursive poll. Each connection has an independent `closemu`, so an `ErrBadConn` on one session doesn't collide with the other session. Independent LogMiner state on the inner session also very likely prevents a cause of `ErrBadConn` entirely since the inner queries no longer mutate the outer cursor's session state.
3. Defer `rows.Close()` in `receiveMessages` to ensure we always call `rows.Close()` even if we encounter an error while iterating through `rows.Next()`.

I recommend reading commit-by-commit. This PR looks like a lot of changes, but a good chunk of them are due to the refactor in the first commit.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-oracle/replication.go
+++ b/source-oracle/replication.go
@@ -719,9 +719,15 @@ func (s *replicationStream) StartReplication(ctx context.Context, discovery map[
 		// Context cancellation typically occurs only in tests, and for test stability
 		// it should be considered a clean shutdown and not necessarily an error.
 		close(s.events)
-		s.transactionsStmt.Close()
-		s.conn.Close()
-		s.innerConn.Close()
+		if cerr := s.transactionsStmt.Close(); cerr != nil {
+			logrus.WithError(cerr).Warn("closing transactions statement")
+		}
+		if cerr := s.conn.Close(); cerr != nil {
+			logrus.WithError(cerr).Warn("closing replication connection")
+		}
+		if cerr := s.innerConn.Close(); cerr != nil {
+			logrus.WithError(cerr).Warn("closing inner replication connection")
+		}
 		s.errCh <- err
 	}()
 
@@ -1108,7 +1114,11 @@ func (s *replicationStream) receiveMessages(ctx context.Context, startSCN, endSC
 	if err != nil {
 		return fmt.Errorf("logminer query: %w", err)
 	}
-	defer rows.Close()
+	defer func() {
+		if cerr := rows.Close(); cerr != nil {
+			logrus.WithError(cerr).Warn("closing logminer rows")
+		}
+	}()
 
 	var totalMessages = 0
 	var fullMessages = 0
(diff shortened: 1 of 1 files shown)


source-oracle/replication.go
```diff
@@ -43,22 +65,21 @@ func (db *oracleDatabase) ReplicationStream(ctx context.Context, cpJSON json.Raw
 	if err != nil {
 		return nil, fmt.Errorf("unable to connect to database: %w", err)
 	}
```
nicolaslazo, line 67: This is not related to your PR, but I wanted to surface it for awareness. I think we are leaking connection slots in source dbs. Per `(sql.DB).Conn`'s documentation:

> Every Conn must be returned to the database pool after use by calling `Conn.Close`.

Adding `defer conn.Close()` should be enough to address this.

source-oracle/replication.go
```diff
@@ -1064,6 +1108,7 @@ func (s *replicationStream) receiveMessages(ctx context.Context, startSCN, endSC
 	if err != nil {
 		return fmt.Errorf("logminer query: %w", err)
 	}
+	defer rows.Close()
```
nicolaslazo: Nit: `rows.Close()` is a fallible operation, so we could log its result for a little added visibility:


```suggestion
	defer func() {
        if cerr := rows.Close(); cerr != nil {
            logrus.WithError(cerr).Warn("closing logminer rows")
        }
    }()
```

source-oracle/replication.go
```diff
@@ -688,6 +721,7 @@ func (s *replicationStream) StartReplication(ctx context.Context, discovery map[
 		close(s.events)
 		s.transactionsStmt.Close()
 		s.conn.Close()
+		s.innerConn.Close()
```
nicolaslazo: Nit: like mentioned [here](https://github.com/estuary/connectors/pull/4543/changes#r3310957472), it would be nice to log if anything goes wrong while attempting to close connections

source-oracle/replication.go
```diff
@@ -1319,6 +1362,7 @@ func (s *replicationStream) Close(ctx context.Context) error {
 	s.cancel()
 	s.transactionsStmt.Close()
 	s.conn.Close()
+	s.innerConn.Close()
```
nicolaslazo: Nit: this would be another instance where conn.Close() calls could fail, like mentioned here

nicolaslazo (approved):
Looks good! I just left a few nits around error visibility

source-oracle/replication.go
Alex-Bair: We're actually already calling `Close()` on the connections in the `replicationStream`'s public `Close` method here.
https://github.com/estuary/connectors/blob/42a6ce35a076797468c5921b1b5189d676487c7c/source-oracle/replication.go#L1317-L1323

The `sqlcapture` framework is already deferring calling the replication stream's `Close` here, so we should be cleaning up connection slots whenever the capture stops running.
https://github.com/estuary/connectors/blob/42a6ce35a076797468c5921b1b5189d676487c7c/sqlcapture/capture.go#L220-L224

source-oracle/replication.go
Alex-Bair: Good point, I've added logging whenever a `Close()` call returns an error.

PR #4545, opened 2026-05-27
source-mysql: Fix latin-1 transcoding in VARCHAR columns

willdonnelly:
**Description:**

Turns out the value decoding path for VARCHAR columns is subtly different from TEXT and we mostly exercised TEXT in our collation test cases, so there was a path for non-ASCII characters in VARCHAR columns with the latin-1 character set to get plumbed directly into JSON serialization (as strings containing the raw latin1 bytes) which would replace them with U+FFFD. Fixed by improving the test coverage and fixing the transcoding pipeline so those strings get translated according to the latin1 character set as well.

Also fixed another issue while I was here, turns out that MySQL "latin1" is not in fact ISO latin1 but rather the Windows cp1252 character set. Because of course it is. The two are mostly identical except for codepoints 0x80-0x9F which are only defined in cp1252, so we should use that character set when decoding these bytes.

**Workflow steps:**

Backfills may be required in some cases to capture the correctly transcoded data, however this edge case isn't terribly common in practice (the default collation in MySQL is UTF-8 these days) and in most cases where latin1 is usable at all the incorrect transcoding result only produces infrequent U+FFFD replacements where there should be an accented character. I don't believe we need to proactively identify impacted dataflows at this time, backfilling collections as needed if/when users complain should be sufficient.

### DIFF
--- a/source-mysql/.snapshots/TestUnicodeText-latin1_char
+++ b/source-mysql/.snapshots/TestUnicodeText-latin1_char
@@ -1,0 +1,14 @@
+# ================================
+# Collection "acmeCo/test/test/unicodetext_latin1_char_78412948_07": 6 Documents
+# ================================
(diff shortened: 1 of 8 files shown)


Alex-Bair (approved):
LGTM

PR #4546, opened 2026-05-27
source-salesforce-native: add My Domain login host support

Alex-Bair:
**Description:**

Organizations that require their My Domain login URL and block `login.salesforce.com` could not authenticate with either method since all authentication methods were hardcoded to use the standard `login.salesforce.com` or `test.salesforce.com` URLs.

This PR fixes this by adding an optional top-level `my_domain` field to the endpoint config that is used as the login host for both auth methods when set and falls back to the existing login/test behavior when blank:

- OAuth: the authorize and access-token URL mustache templates select the My Domain host when present, and the runtime refresh-token URL is built from it as well.
- Username/Password: the SOAP login is reduced to the subdomain `simple_salesforce` expects (e.g. mycompany.my.salesforce.com -> mycompany.my).

The field is strictly validated as a full `*.my.salesforce.com` host because the control plane interpolates it verbatim into the OAuth URLs & does not normalize it. An empty or absent value takes the `login`/`test` fallback.

Spec snapshot changes are expected due to the addition of the `my_domain` field to the endpoint config as well as the changes to the mustache templates in the OAuth URLs.

The connector's documentation should be updated to include the new `my_domain` endpoint config field. It'd be worthwhile to update the connector's documentation to indicate when users would want to fill out the new `my_domain` field.

**Notes for reviewers:**

Tested on a dev stack. Confirmed:
- Both authentication methods plus the OAuth flow still work without filling out the `my_domain` field.
- Both authentication methods plus the OAuth flow work when the `my_domain` field is filled out.

Relevant Salesforce documentation:
[Customize Your Login Process with My Domain](https://trailhead.salesforce.com/content/learn/modules/identity_login/identity_login_my_domain)
[Log In To Salesforce with Code](https://help.salesforce.com/s/articleView?id=xcloud.domain_name_login_code.htm&type=5)
[Set the My Domain Login Policy](https://help.salesforce.com/s/articleView?id=xcloud.domain_name_setting_login_policy.htm&type=5)

### DIFF
--- a/source-salesforce-native/source_salesforce_native/auth.py
+++ b/source-salesforce-native/source_salesforce_native/auth.py
@@ -21,22 +21,27 @@ from simple_salesforce.login import SalesforceLogin
 from .shared import VERSION
 
 
+# _OAUTH_LOGIN_ORIGIN is a mustache fragment rendered by the control plane that selects the OAuth login origin.
+# The org's My Domain host is used when provided, otherwise the standard login/test host is used.
+_OAUTH_LOGIN_ORIGIN = (
+    "https://"
+    r"{{#config.my_domain}}{{{config.my_domain}}}{{/config.my_domain}}"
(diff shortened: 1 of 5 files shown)


nicolaslazo (approved):
LGTM!

PR #4547, opened 2026-05-27
source-hubspot-native: add marketing events snapshot stream 

nicolaslazo:
**Description:**

Adds support for the [marketing events](http://developers.hubspot.com/docs/api-reference/marketing-marketing-events-v3/guide) API endpoint. Though the documentation lacks precision, the behaviour is consistent with other v3 endpoints.

**Workflow steps:**

**Notes for reviewers:**

Tested through `flowctl preview` on a test account with dummy data.

This new stream requires the `crm.objects.marketing_events.read` scope to be added to our production app before merging.


source-hubspot-native/test.flow.yaml
```diff
@@ -65,6 +65,10 @@ captures:
     - resource:
         name: marketing_emails
       target: acmeCo/marketing_emails
+    - resource:
+        name: marketing_event
+        interval: PT5M
+      target: acmeCo/marketing_event
```
Alex-Bair: The CI check is failing because our test credentials don't have the correct scope to hit the `marketing_events` endpoint:
```
estuary_cdk.http.HTTPError: Encountered HTTP error status 403 which cannot be retried.
URL: https://api.hubapi.com/marketing/v3/marketing-events
Response:
***"status":"error","message":"This app hasn't been granted all required scopes to make this call. Read more about required scopes here: https://developers.hubspot.com/scopes.","errors":[***"message":"All of the following scopes are required.","context":***"requiredGranularScopes":["crm.objects.marketing_events.read"]***],"links":***"scopes":"https://developers.hubspot.com/scopes"***,"category":"MISSING_SCOPES"***
```

This is a bit of a chicken or the egg situation: we can't capture `marketing_events` in the connector until our OAuth app is requesting the correct scopes, but we don't want to update our OAuth app's requested scopes until we update the connector. I suggest disabling the `marketing_events` stream here for now in `flow.yaml` so the connector doesn't attempt to hit the associated endpoint, then re-enable it later after we update our OAuth app's scopes. We won't have immediate coverage for what a marketing event document should look like in the capture snapshot, but as long as we follow back up afterwards we'll have test coverage shortly after the OAuth app scope changes have landed.

source-hubspot-native/test.flow.yaml
```diff
@@ -65,6 +65,10 @@ captures:
     - resource:
         name: marketing_emails
       target: acmeCo/marketing_emails
+    - resource:
+        name: marketing_event
```
Alex-Bair: The new stream's name is singular (`marketing_event`) whereas most all other streams are plural. Could we update the new stream to have a plural name (`marketing_events`) to align with how the other streams are named? We may not be able to reuse the existing `Names.marketing_event` since it's already used for an association elsewhere, so it may be cleanest to add a new `Names.marketing_events` member to the enum to use as this stream's name.

source-hubspot-native/source_hubspot_native/resources.py
```diff
@@ -640,6 +651,35 @@ def open(
...
+            binding_index,
+            state,
+            task,
+            fetch_snapshot=functools.partial(fetch_marketing_events, http),
+            tombstone=MarketingEvent(_meta=MarketingEvent.Meta(op="d")),
+        )
+
+    return SnapshotResource(
+        name=Names.marketing_event,
+        model=MarketingEvent,
```
Alex-Bair: nit: It doesn't look like there are any fields on the `MarketingEvent` model beyond what's already defined on the standard `BaseDocument` one, so we should be able to omit specifying both the `model` and `tombstone` arguments here - the CDK will automatically use `BaseDocument` as the model and `BaseDocument.Meta(op="d")` for the tombstone for any `SnapshotResource`s by default.

Alex-Bair (commented):
Looks pretty good! I had a couple small items:
- Looks like the CI check is failing, and we'll need to make sure it passes before merging & making the scope changes to our OAuth app.
- Could the stream name be plural to align with how the other streams are named?

source-hubspot-native/test.flow.yaml
nicolaslazo: Yeah, my plan was to wait until the PR was approved, re-run tests updating snapshots if necessary, and only then merging. Your plan sounds good to me too

Alex-Bair: Gotcha, yeah, that'd work too.

Alex-Bair (approved):
LGTM!

PR #4548, opened 2026-05-27
materialize-postgres: add "enum" flag that is enabled by default

mdibaiee:
**Description:**

- We released this change but then realised that `_meta/op` is an enum and all tasks triggered migrations. Now some tasks are stuck because of not having enough disk to do the migration, and some other tasks have views that depend on `_meta/op` and during migration they cannot be dropped.

**Workflow steps:**

**Notes for reviewers:**


dyaffe (commented):
LGTM

dyaffe (approved):
LGTM

PR #4550, opened 2026-05-27
source-salesforce-native: add client credentials authentication method

Alex-Bair:
**Description:**

Users who have SSO enabled in their organization are still hitting authentication errors with the `UserPass` authentication option even after specifying their My Domain via the `my_domain` field introduced in https://github.com/estuary/connectors/pull/4546.

This PR adds a new OAuth 2.0 Client Credentials authentication option that uses an Salesforce External Client App to avoid requiring interactive login (and therefore the SSO process as well). I'm optimistic this will be an acceptable authentication option for users whose organization's SSO & other login restrictions make `UserPass` and authentication options requiring interactive login fail.

Spec snapshot changes are expected due to the addition of the new authentication method.

The connector's documentation should be updated to reflect the new authentication option. The steps to create an External Client App in Salesforce aren't super obvious, so we should add details to the documentation describing how to do so as well.

**Notes for reviewers:**

I confirmed with my own dev Salesforce account:
- the new client credentials authentication option works by creating an External Client App in my dev Salesforce account, updating `config.yaml` to use that new app's credentials, & confirming all tests still pass with `pytest`. I also verified the same thing on a dev stack in a GCP VM too.
- the existing authentication options still work like they did previously.


nicolaslazo (approved):
LGTM!

PR #4557, opened 2026-05-28
source-hubspot-native: sleep in fetch_delayed_changes until it is time to poll again

williamhbaker:
**Description:**

Prevent a CPU-intensive hot loop when `fetch_delayed_changes` is not yet ready to poll.

This is a targeted change in source-hubspot-native to mitigate the immediate issue. I could imagine a higher-level and more extensible addition to the CDK's multi-stream mechanism to allow for a separate interval to be specified for multiple different streams, but I'm leaving that for a later implementation if its warranted.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api/shared.py
+++ b/source-hubspot-native/source_hubspot_native/api/shared.py
@@ -1,4 +1,5 @@
-from datetime import datetime, timedelta, UTC
+import asyncio
+from datetime import UTC, datetime, timedelta
 from logging import Logger
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM

As long as the configured resource interval is greater than 0, I thought the CDK should handle sleeping until it's time to poll again.
https://github.com/estuary/connectors/blob/60037150e98520ae4d6133f2f94a645e553bd492/estuary-cdk/estuary_cdk/capture/common.py#L1214-L1216

But if the resource interval is 0 (which is currently the default), then yeah, it'll keep looping and churn through CPU.

PR #4592, opened 2026-05-29
source-mongodb: probe `startAfter` support instead of inferring from version

Alex-Bair:
**Description:**

We've observed some `source-mongodb` captures connecting to a DocumentDB instance through a VCP endpoint attempt to use change streams but fail with the error "BSON field '$changeStream.startAfter' is an unknown field". Since the endpoint didn't end in `docdb.amazonaws.com`, the connector didn't detect it was connecting to a DocumentDB instance with the `isDocumentDB` check & attempted to use change streams instead of defaulting to batch mode. It turns out that Amazon DocumentDB can report a recent version and supports change streams, post-batch resume tokens, and `resumeAfter`, but does not implement `startAfter`, and that caused the error mentioned above.

This PR fixes that by replacing the version-based `supportsStartAfter` gate with `probeChangeStreamSupport`, which determines capabilities by attempting them rather than trusting the reported version:

- It opens a change stream and reads the post-batch resume token. A server that produces no PBRT cannot checkpoint reliably, so it is captured in batch mode.
- If a PBRT is present, it tests `startAfter` with that token and falls back to `resumeAfter` when the option is rejected.

I verified the new behavior against a real DocumentDB 5.0.0 instance. It keeps change-stream mode and uses `resumeAfter`, while genuine MongoDB continues to use `startAfter`.

### DIFF
--- a/source-mongodb/pull.go
+++ b/source-mongodb/pull.go
@@ -19,6 +19,7 @@ import (
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


nicolaslazo (approved):
LGTM

PR #4593, opened 2026-05-29
source-mongodb: treat transcoder shutdown as cancellation, not failure

Alex-Bair:
**Description:**

Capture tests intermittently failed with spurious "transcoding ... reading frame: EOF" errors on shutdown, which I hit running tests locally and in CI.

The transcoder subprocess is tied to the capture's context via `exec.CommandContext`, so when the capture shuts down and cancels its context, the subprocess is killed. A batch that's mid-transcode then fails its pipe read or write with EOF, and `transcodeBatch` returned that as a hard error instead of recognizing the cancellation.

Pipe reads and writes aren't context-aware, so on a read or write error `transcodeBatch` now checks `t.ctx.Err()` first, returning the cancellation when the context is done and falling through to the original error otherwise. The EOF is causally downstream of the cancellation, so the context error is always set by then. A genuine crash while the context is live is still surfaced.

This was a regression from moving to an out-of-process transcoder in 5c28c4aaa; previously transcoding was in-process with no subprocess to kill.

### DIFF
--- a/source-mongodb/transcoder.go
+++ b/source-mongodb/transcoder.go
@@ -195,6 +195,14 @@ func (t *Transcoder) transcodeBatch(count int, writeAll func() error) ([]*Transc
 	for i := 0; i < count; i++ {
 		resp, err := t.readResponse()
 		if err != nil {
+			// Pipe reads aren't context-aware, so an EOF here is ambiguous. It's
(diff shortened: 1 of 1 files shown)


nicolaslazo (approved):
LGTM

PR #4594, opened 2026-05-29
source-sentry: support custom explore queries

### LINKED ISSUE #4318: source-sentry: add new streams
- Traces: /api/0/organizations/{organization_id_or_slug}/traces/{trace_id}/ -> Traces

- Events: /api/0/projects/{organization_id_or_slug}/frontend/events/{trace_id}/ -> Events

Alex-Bair:
**Description:**

This PR adds an `explore_queries` config option. Each entry runs a [Sentry Explore events query](https://docs.sentry.io/api/explore/query-explore-events-in-table-format/) for a chosen dataset and field list, and becomes its own `custom_explore_`-prefixed incremental stream.

Incremental capture uses a rolling `statsPeriod=30d` full-fidelity window with server-side timestamp bounds and a forward-walk that advances the cursor through Sentry's 10k pagination cap. Backfill covers history older than 30 days via explicit start/end windows.

Snapshot changes are all expected due to the addition of the new endpoint config fields & specifying an `explore_queries` stream for our snapshot tests.

The connector's documentation should be updated to reflect the new config options & point to [Sentry's documentation](https://docs.sentry.io/api/explore/query-explore-events-in-table-format/) for details on what values are valid for each `explore_queries` field, especially since the UI doesn't appear to show descriptions for these fields.

image

**Notes for reviewers:**

Tested on a development stack & with `flowctl preview`. Confirmed data is captured for the `custom_explore_my_spans` query added in `config.yaml`.


source-sentry/source_sentry/models.py
```diff
@@ -20,6 +28,156 @@ def default_start_date():
...
+    def field_list(self) -> list[str]:
+        """User-specified fields (deduped, original order) followed by the
+        required PK/cursor fields the connector always queries."""
+        seen: set[str] = set()
+        result: list[str] = []
+        for field in _split_csv(self.fields) + required_fields(self.dataset):
+            if field not in seen:
+                seen.add(field)
+                result.append(field)
+        return result
```
nicolaslazo: Nit: if we don't care about required fields being located at the end of the list, we could replace the dual accumulator loop with a one liner:

```python
return list(set(_split_csv(self.fields) + required_fields(self.dataset)))

nicolaslazo (approved):
LGTM!

source-sentry/source_sentry/models.py
Alex-Bair: Good point, I like that. I updated it to be a one liner. I used `dict.fromkeys` instead of `set` to keep a deterministic ordering of the user-specific fields. That may not ultimately matter too much, but I figure anything we can keep the same across runs will make troubleshooting later a smidge easier.

PR #4595, opened 2026-06-01
source-hubspot-native: let real-time streams skip out of order results

Alex-Bair:
**Description:**

When searching for records updated very close to the present, we've observed HubSpot's Search API return records slightly out-of-order. This has been causing capture tasks to crash somewhat frequently.

This PR makes the real-time streams skip any out-of-order results they receive from the Search API. This is ok because the delayed stream will come along later and emit any records that the real-time stream missed.


nicolaslazo (approved):
LGTM

PR #4596, opened 2026-06-01
source-hubspot-native: use `/modified/after` endpoint for delayed engagements

Alex-Bair:
**Description:**

The delayed `engagements` stream used the `/engagements/v1/engagements/recent/modified`, endpoint which returns results most-recent-first, is offset-paginated, and caps at ~10k records. Because the delayed stream trails realtime by an hour and that endpoint has no time filter, it had to page through the entire most-recent hour of modifications before reaching the date window it cares about. On a high-volume account, that trailing hour alone exceeds 9,900 records, so every delayed poll raised `MustBackfillBinding`, which triggered a full-connector restart. The result was a constant re-backfill loop that prevented the engagements binding from making progress.

This PR fixes that by switching the delayed stream to use the [`/engagements/v1/engagements/modified/after` endpoint](https://developers.hubspot.com/docs/api-reference/legacy/crm/activities/engagements/get-engagements-v1-engagements-modified-after), which filters by modification time, is cursor-paginated, and has no 10k offset cap. Because it reads forward from the cursor's timestamp there is no recent-zone to page past. Its results are unordered, so the fetcher fully enumerates the cursor and returns everything at once, letting `fetch_changes_with_associations` apply the `[since, until]` windowing. The endpoint returns the same `/engagements/v1` id space, so the existing batch-read, association, and backfill paths are unchanged.

**Notes for reviewers:**

Tested with `flowctl preview`. Confirmed the delayed stream for `engagements` captures data as expected.

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api/engagements.py
+++ b/source-hubspot-native/source_hubspot_native/api/engagements.py
@@ -11,18 +11,53 @@ from estuary_cdk.http import HTTPSession
 
 from ..models import (
     Engagement,
+    EngagementsModifiedAfter,
     Names,
     OldRecentEngagements,
 )
 from .object_with_associations import fetch_changes_with_associations
 from .shared import (
+    dt_to_ms,
     ms_to_dt,
     HUB,
     MustBackfillBinding,
 )
 
 
-async def _fetch_engagements(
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
Cool find!

PR #4601, opened 2026-06-02
source-hubspot-native: add marketing event participants stream

nicolaslazo:
**Description:**

Adds a new stream for the [Retrieve participant breakdown by HubSpot ID](https://developers.hubspot.com/docs/api-reference/legacy/marketing/marketing-events/participant-state/get-breakdown) endpoint.

**Notes for reviewers:**

- Tested using `flowctl preview`
- `MarketingEvent.objectId` has been made a mandatory field. It is used to construct event participation URLs
- Filtering events by `updatedAt` has been considered, but adding new participants has not caused a timestamp refresh. All events will have to be iterated through unconditionally.

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api/marketing_events.py
+++ b/source-hubspot-native/source_hubspot_native/api/marketing_events.py
@@ -5,12 +5,10 @@ from estuary_cdk.http import HTTPSession
 from estuary_cdk.incremental_json_processor import IncrementalJsonProcessor
 from pydantic import BaseModel, JsonValue
 
-from ..models import MarketingEvent, PageResult
-from .shared import (
-    HUB,
-)
+from ..models import MarketingEvent, MarketingEventParticipant, PageResult
+from .shared import HUB
 
-MARKETING_EVENTS_PAGE_SIZE = 100
+PAGE_SIZE = 100
 
 
 class _ResponseRemainder(BaseModel, extra="allow"):
@@ -24,9 +22,7 @@ async def fetch_marketing_events(
     url = f"{HUB}/marketing/v3/marketing-events"
     after: str | None = None
 
-    input: dict[str, JsonValue] = {
-        "limit": MARKETING_EVENTS_PAGE_SIZE,
-    }
+    input: dict[str, JsonValue] = {"limit": PAGE_SIZE}
 
     while True:
         if after:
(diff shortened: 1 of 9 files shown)


source-hubspot-native/source_hubspot_native/api/marketing_events.py
```diff
@@ -48,3 +44,37 @@ async def fetch_marketing_events(
             return
 
         after = paging.next.after
+
+
+async def fetch_marketing_event_participants(
+    http: HTTPSession,
+    log: Logger,
+) -> AsyncGenerator[MarketingEventParticipant, None]:
+    async for event in fetch_marketing_events(http, log):
```
Alex-Bair: nit: It may be unlikely to be an issue here, but since we're holding the API response for marketing events open as we're iterating through the parent ids & making requests for child resources, it's possible we hit `TimeoutErrors` like I mentioned in this [comment](https://github.com/estuary/connectors/pull/4309#discussion_r3138846747). Maybe we're able to fetch all participants of all marketing events before a `TimeoutError` is raised, but if we can't, a simple fix is to cache all parent ids in memory before fetching any child resources.

source-hubspot-native/source_hubspot_native/resources.py
```diff
@@ -678,6 +685,33 @@ def open(
...
+            state,
+            task,
+            fetch_snapshot=functools.partial(fetch_marketing_event_participants, http),
+        )
+
+    return SnapshotResource(
+        name=Names.marketing_event_participants,
+        open=open,
+        initial_config=ResourceConfig(
+            name=Names.marketing_event_participants, interval=timedelta(minutes=5)
```
Alex-Bair: nit: Do you think we could make the default interval for `marketing_event_participants` more around 15 minutes or more to reduce how frequently it runs & consumes some of the API limit? Since it's a child snapshot stream, I can imagine it's more likely to take a fair number of API requests to complete a snapshot than a non-child stream, and marketing event participants sounds like something that doesn't need a 5 minute interval by default.

Alex-Bair (approved):
LGTM % a couple optional nits

source-hubspot-native/source_hubspot_native/api/marketing_events.py
nicolaslazo: Fair enough, this should become standard practice. On it

Alex-Bair (approved):
LGTM, thanks!

PR #4495, opened 2026-05-18
source-impact-native: fix timestamp bugs

nicolaslazo:
**Description:**

- Use `>=` to include actions at incremental checkpoint boundaries
- Do not overwrite `max_ts` on every campaign and yielded document for `Actions` incremental fetches
- Do not return early when multiple documents are present at the start date

**Workflow steps:**

(How does one use this feature, and how has it changed)

**Documentation links affected:**

(list any [documentation links](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing) that you created, or existing ones that you've identified as needing updates, along with a brief description)

**Notes for reviewers:**

No testing credentials were available

PR #4602, opened 2026-06-02
source-impact-native: fix `Contracts` stream bugs

Alex-Bair:
**Description:**

This PR's scope includes:
- Fixing the same `max_ts` bug for all child streams of `Campaigns` that was fixed for `Actions` in https://github.com/estuary/connectors/pull/4495.
- Fixing `Contracts` to look at `DateLastUpdated` for cursor values instead of `DateCreated`.

All active tasks (of which there are only 2) should backfill their `Contracts` streams to capture whatever data may have been missed due to the `DateLastUpdated` vs. `DateCreated` mismatch bug.

### DIFF
--- a/source-impact-native/source_impact_native/api.py
+++ b/source-impact-native/source_impact_native/api.py
@@ -281,7 +281,7 @@ async def fetch_incremental_child(
     async for campaign in campaigns:
         campaign_list.add(campaign.Id)
 
-    
+    max_ts = log_cursor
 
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
LGTM! 

Since we're all using the `pr-review-toolkit:review-pr` skill before we push our PRs, I'll update our `CLAUDE.md` to automatically check if bug fixes apply to other existing streams.

PR #4608, opened 2026-06-03
source-oracle: tolerate redo log SCN gaps from database restarts

Alex-Bair:
**Description:**

`validateLogFileCoverage` required each redo log to be exactly adjacent to the previous one (`curr.FIRST_CHANGE# == prev.NEXT_CHANGE#`), which catches when archive logs are deleted before we can mine them. But it also causes the connector to fail when an Oracle restart leaves behind an SCN gap after routine maintenance, with errors like:
```
redo log SCN gap in thread 1 between "..._seq_357345..." (NEXT_CHANGE#=91403872085) and "..._seq_357346..." (FIRST_CHANGE#=91403893034)
```

Within a thread, an unbroken `SEQUENCE#` chain is what guarantees no redo was lost. This PR relaxes the check to tolerate a gap only when it carries the restart signature: consecutive `SEQUENCE#`s, a forward SCN jump, and an unchanged incarnation (`RESETLOGS_ID`, ruling out an `OPEN RESETLOGS` rewind that discards redo).

**Notes for reviewers:**

I confirmed with a dockerized Oracle instance that restarts keep `SEQUENCE#` consecutive and `RESETLOGS_ID` unchanged while the SCN jumps forward across a range that contained no changes to user tables.

### DIFF
--- a/source-oracle/replication.go
+++ b/source-oracle/replication.go
@@ -191,6 +191,11 @@ type redoFile struct {
 	DictEnd     string
 	Bytes       int
 	Thread      int
+	// ResetlogsID identifies the database incarnation this log belongs to. It
+	// changes when the database is opened with RESETLOGS, which discards redo
+	// and forks a new timeline. Used by validateLogFileCoverage to distinguish
+	// a benign restart gap from a data-losing incarnation change.
+	ResetlogsID int
 }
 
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
LGTM!

PR #4609, opened 2026-06-03
source-sqlserver{,-batch}: send real server name when tunneling

Alex-Bair:
**Description:**

When an SSH tunnel is configured, both connectors rewrite the DSN host to `localhost` to point the TCP dial at the tunnel's local listener. But `go-mssqldb` also derives the advertised server name (TDS LOGIN7 `ServerName` and TLS SNI) from the DSN host, so the connectors were advertisiung `localhost` too.

Self-hosted SQL Server ignores that name, but Azure SQL uses it to route each connection to the correct backend behind its shared gateway. Receiving `localhost`, the gateway rejects the login (`error 40532: Cannot open server "localhost" requested by the login` or `Connection was denied because Deny Public Network Access is set to Yes`), which breaks tunneled Azure captures.

This PR fixes this by decoupling the dial target from the advertised name: `ToURI` keeps the real address in the DSN, and a custom `go-mssqldb` `HostDialer` redirects only the TCP dial to the tunnel.

**Notes for reviewers:**

Tested discovery with both connectors and an Azure SQLServer instance accessed via an SSH network tunnel. Before this PR's changes, login failed with `Connection was denied because Deny Pubic Network Access is set to Yes` (despite that setting not actually being enabled). After the changes, both connectors connected through the tunnel and completed discovery successfully.

Regarding all the CI failures, I anticipate they're related to the issue discussed in this [thread](https://estuaryworkspace.slack.com/archives/C03PF2FSM5H/p1780515277736809) and will be resolved once this [PR](https://github.com/estuary/flow/pull/2995) is merged. Edit: That PR's merged & the relevant CI runs are green now. 👍

### DIFF
--- a/source-sqlserver-batch/main.go
+++ b/source-sqlserver-batch/main.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"net/url"
 	"strings"
 	"text/template"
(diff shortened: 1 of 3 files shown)


nicolaslazo (approved):
👍

willdonnelly:
Just noting that we should probably do this for source-sqlserver-ct as well, whether in this PR or a separate one.

Alex-Bair:
Thanks for catching that, I'll add `source-sqlserver-ct` to this PR too.

PR #4622, opened 2026-06-04
source-kafka: include additional information when failing to parse a message payload

Alex-Bair:
**Description:**

When a message payload fails to parse, it'd be helpful to have more information around the message than only its topic. This PR adds that additional info (like the partition, offset, payload length, first few bytes covering schema registry framing), which will be helpful when diagnosing why the connector failed to parse a specific message.


nicolaslazo (approved):
LGTM

PR #4623, opened 2026-06-04
materialize-clickhouse: hard check against silent staged-row loss at commit

jacobmarble:
**Description:**

The connector commits by staging rows into a `flow_temp_store_*` table, enumerating its parts via `system.parts`, moving them to the target table, and dropping the stage table. On ClickHouse Cloud, the parts query can land on a replica with stale metadata for the stage table (re-created with a new UUID every transaction). That replica reports zero parts, the move is skipped, and the `DROP` silently destroys the staged rows. Observed in production: ~half of all transactions lost over multiple days, clean logs throughout.

This PR adds detection only — it does not fix the underlying race:

- `queryStoreParts` also returns per-partition row counts.
- `Store` records rows staged per binding in the persisted state (`stateItem.StoredRows`).
- The commit fails before moving if `system.parts` doesn't account for every staged row, and before dropping if any rows remain after the moves — preserving the stage table for retry.

**Workflow steps:**

No happy-path change. A transaction that would previously lose data now fails with `refusing to commit store table: system.parts reports N staged rows but M rows were stored` and retries.

None.

**Notes for reviewers:**

- `StoredRows == 0` (state from a prior version's checkpoint) disables the check — rollout is backward-compatible.
- On recovery, a mismatch with the stage table gone is treated as an already-completed commit. A crash between MOVE and DROP now fails loudly on recovery instead of assuming success — deliberate.
- New integration test `TestMoveRefusesOnStoredRowsMismatch` covers fail/preserve/succeed/recover.


jacobmarble:
This has run overnight for a customer, and caught one instance of the theorized data loss. It also prevented data loss by yielding an error instead of silently dropping the data; the error was retried and the second attempt succeeded.

danielnelson (approved):

PR #4625, opened 2026-06-05
source-stripe-native: add `prices` and `value_lists` streams; Claude improvements

nicolaslazo:
**Description:**

Adds three new streams:
- Prices
- Value lists
- Value list items

Multiple miscellaneous changes have been bundled in. Markedly, the refactoring of resource definitions to reduce code duplication and the inclusion of new Claude skills to aid in the development of new streams.

**Workflow steps:**

**Notes for reviewers:**

Tested through `flowctl preview`. Best reviewed commit by commit.

### DIFF
--- a/.claude/skills/add-stream/SKILL.md
+++ b/.claude/skills/add-stream/SKILL.md
@@ -1,0 +1,112 @@
+---
+name: add-stream
+description: Add a new stream to an existing estuary-cdk connector — classify the endpoint, register it, regenerate flow discovery, refresh snapshots. Use when extending a `source-*` connector with another endpoint.
+argument-hint: "[connector-name] [stream-name]"
+allowed-tools: Bash Read Write Edit Glob Grep WebFetch WebSearch
+---
+
+Add a `stream-name` stream to the `source-$1` connector. Read a few neighboring streams in the same connector before designing the new one — local conventions outrank any generic pattern.
+
+## Laws
+
+These apply in every phase. Re-read them before each phase boundary.
+
+1. Open the session with a TODO list. One task per phase below, plus any provider-specific work that comes up.
+2. **Budget rule for captures and previews:** Phase 1 (Rate Limit Survey) sets the budget. If every endpoint this stream and its siblings will touch allows **more than 20 requests/hour**, you may run `flowctl preview`, `pytest` (including the snapshot tests that drive `flowctl preview` under the hood), and capture-running commands as often as needed without further consent. If any required endpoint is tighter than that, ask before each run, and consider Law #6.
+3. **Cost-saver:** `disable: true` on a binding in `test.flow.yaml` skips that stream during preview/capture. When the budget is tight or you only care about the new stream's behavior, disable unrelated bindings before running tests. Restore them before committing.
+4. **`flowctl raw discover` is always free.** It exercises only the connector's discovery path and doesn't burn provider API budget. No consent needed.
+5. Prefer snapshot tests over fine-grained assertions when checking outputs.
+6. State out loud when you're deliberately following an existing convention. State equally out loud when you're deviating, and justify it. Silence is worse than confirmation.
+7. When you copy a pattern from another stream, cite the exact `file:line` you copied from so the user can sanity-check.
+8. If you propose a function/field/flag that you remember seeing in the connector, grep for it before recommending — the codebase may have changed since memory was formed.
+
+## Phase 0 — Reconnoiter
+
+Confirm the connector exists. Locate its `models.py`, `resources.py`, `api.py` (or equivalents — naming varies between Python connectors). Read streams in the same connector before designing anything.
+
+**Clean-room check.** Kick off the connector's existing test suite (typically `pytest` from the connector directory) in the background as soon as Phase 0 begins. Phases 1–3 are read-only / planning and may proceed in parallel while the suite runs — do **not** block on it. The clean-room result is a **gate on Phase 4**: before writing any code for the new stream, confirm the baseline passes so later failures are attributable to the new stream, not pre-existing drift.
+
+If the suite fails:
+
+- **Simple schema drift** (the connector spec/discover output has shifted but the code is unchanged): you may run the `regenerate-flow-discovery` sub-skill to refresh the snapshots, then commit the result on its own with the message `<connector-name>: update tests`. Keep this commit separate from the new-stream work so the PR diff stays scoped.
+- **Flakey fields in the diff** (timestamps like `updated_at`, ETags, anything that changes between runs): surface them to the user and ask whether to add them to the connector's `FIELDS_TO_REDACT` list in `tests/test_snapshots.py` (name varies between connectors — grep for `FIELDS_TO_REDACT` to find the convention).
+- **Anything else:** stop and surface the failure to the user before entering Phase 4.
+
+## Phase 1 — Rate Limit Survey
+
+Before researching the requested stream, figure out the provider's API rate limits.
+
+**First, check for cached notes.** Read `<connector>/CLAUDE.md` (create the file if it doesn't exist). If it has an "API Rate Limits" section, use those values — don't re-derive. They're authoritative until proven stale (and you'll only spot staleness if a run starts 429-ing).
+
+**If the section is absent**, look up the limits:
+
+- The published limit for the endpoint family (typically requests/second or requests/hour, sometimes per-key, sometimes per-account).
+- Any _tighter_ limits that apply to the specific endpoints this stream will touch — search endpoints, expensive list endpoints, and event firehoses often have lower budgets than the general bucket.
+- Any soft-quota / overage / 429 backoff behavior the provider documents.
+
+Then **write the findings to `<connector>/CLAUDE.md`** under an `## API Rate Limits` section so the next stream addition doesn't repeat the work.
+
+```markdown
+## API Rate Limits
+
+- General: <N> req/sec per key (per provider docs, link).
+- Tighter buckets:
+  - `/v1/<endpoint>`: <M> req/min
+  - `/v1/<events-firehose>`: <K> req/hr
+- 429 behavior: <retry-after / exponential backoff / hard-cut>.
+```
+
+**Set the budget for the rest of the skill**:
+
+- **All required endpoints > 20 req/hr:** run `flowctl preview` / `pytest` / captures freely.
+- **Any required endpoint ≤ 20 req/hr:** ask before each run; plan to disable unrelated bindings in `test.flow.yaml` (`disable: true`) before testing.
+- **`flowctl raw discover` is always allowed**, regardless of budget.
+
+## Phase 2 — Endpoint Survey
+
+State _every_ viable provider endpoint that could back this stream (list, search, expand, related event streams, etc.). For each, summarize what it offers and what it lacks (filterability, sorting, cursors, freshness guarantees, regional availability).
+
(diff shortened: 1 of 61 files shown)


source-stripe-native/source_stripe_native/resources.py
```diff
@@ -399,21 +403,25 @@ async def _maybe_migrate_exempt_state(
...
-    http: HTTPSession,
-    start_date: datetime,
-    incremental_window_size: timedelta,
+def _build_resource(
+    *,
+    name: str,
+    model: type[BaseDocument],
+    key: list[str],
+    fetch_changes_factory,
+    fetch_page_factory,
```
Alex-Bair: nit: Can we add some type annotations for `fetch_changes_factory` and `fetch_page_factory`? You may be able to re-use the `FetchChangesFnFactory` and `FetchPageFnFactory` [protocols](https://github.com/estuary/connectors/blob/eac4a45c1c8cf18ae2eb4ae0b0c3193912475915/source-stripe-native/source_stripe_native/priority_capture.py#L24-L35) in `priority_capture.py` for this if you also widen them a little to allow the `account_id` paramter to be `str | None`.

.gitignore
```diff
@@ -1,5 +1,5 @@
 venv
-.claude/scheduled_tasks.lock
+**/.claude/
```
Alex-Bair: Does this cause the top level `.claude` directory to be ignored? If so, is that intentional?

.claude/skills/add-stream/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Laws
+
+These apply in every phase. Re-read them before each phase boundary.
+
+1. Open the session with a TODO list. One task per phase below, plus any provider-specific work that comes up.
+2. **Budget rule for captures and previews:** Phase 1 (Rate Limit Survey) sets the budget. If every endpoint this stream and its siblings will touch allows **more than 20 requests/hour**, you may run `flowctl preview`, `pytest` (including the snapshot tests that drive `flowctl preview` under the hood), and capture-running commands as often as needed without further consent. If any required endpoint is tighter than that, ask before each run, and consider Law #6.
+3. **Cost-saver:** `disable: true` on a binding in `test.flow.yaml` skips that stream during preview/capture. When the budget is tight or you only care about the new stream's behavior, disable unrelated bindings before running tests. Restore them before committing.
+4. **`flowctl raw discover` is always free.** It exercises only the connector's spec/discovery path and doesn't burn provider API budget. No consent needed.
```
Alex-Bair: nit: It's more precise to say `flowctl raw discover` exercises the discovery path. `flowctl raw spec` exercises the spec path.
```suggestion
4. **`flowctl raw discover` is always free.** It exercises only the connector's discovery path and doesn't burn provider API budget. No consent needed.
```

.claude/skills/classify-stream-types/SKILL.md
```diff
@@ -14,11 +14,28 @@ Evaluate each endpoint in order:
...
 2. **Does the endpoint support date-range or cursor-based filtering with a cursor that persists over time?** (e.g., `updated_at`, monotonic ID, event timestamp, sequence number — not just `created_at`) → **Incremental + Backfill**. Prefer endpoints that also support sorting, but filtering without document sorting is fine.
 3. **No filtering, but supports sorting in reverse chronological order?** → **Incremental only**. Use `fetch_changes` to walk from the latest document backward to the cursor (initially `start_date`). Only yield a cursor checkpoint once the walk reaches the cursor — if interrupted mid-walk, the next invocation restarts from the top. After the initial catch-up, subsequent invocations only walk back to the last checkpoint.
+   - **If the resource is mutable AND no webhook events exist AND there's no `updated_at`-style filter** (only a `created` cursor): incremental-only on `created` catches new rows but **silently misses every update/delete** to existing rows. Pair the incremental stream with a **scheduled backfill** as the only mechanism that recovers update/delete state. See "Scheduled backfill as update fetcher".
```
Alex-Bair: The other option in this situation is a full refresh / snapshot stream. If we're able to complete the snapshot in a reasonable amount of time & the snapshots aren't egregiously large, that's sometimes preferred to the "capture creations incrementally & backfill periodically for updates" method since snapshot streams are able to infer deletions.

source-stripe-native/source_stripe_native/models.py
```diff
@@ -533,6 +533,21 @@ class Payouts(BaseStripeObjectWithEvents):
     }
 
 
+# Stripe's docs for GET /v1/prices claim it "returns a list of your active
+# prices," but the bare list endpoint actually returns both active and
+# inactive prices. The `active` query param only filters when explicitly set;
+# omitting it captures everything — verified empirically on 2026-06-01 against
+# an account with an archived price.
```
Alex-Bair, line 576: Thanks for checking this behavior and noting it!

source-stripe-native/source_stripe_native/models.py
```diff
@@ -456,6 +456,17 @@ class EarlyFraudWarning(BaseStripeObjectWithEvents):
     }
 
 
+# Stripe documents no value_list* event types and no `updated` field
+# on the resource (verified 2026-06-01).
+# Value lists are mutable (rename, item add/remove, delete), so the
+# scheduled backfill in SCHEDULED_BACKFILL_STREAMS is required to
+# eventually observe updates and deletes — incremental on `created`
```
Alex-Bair: nit: The wording here is a little misleading. The scheduled backfill only lets the stream observe updates to previously created value lists, it doesn't capture any deletions of value lists.

source-stripe-native/source_stripe_native/api.py
```diff
@@ -491,6 +491,173 @@ async def fetch_backfill_substreams(
...
+        async for doc in child_data:
+            doc_ts = _s_to_dt(doc.created)
+
+            if doc_ts > max_ts:
+                max_ts = doc_ts
+
+            if doc_ts >= log_cursor:
+                if account_id:
+                    doc.account_id = account_id
+                yield doc
```
Alex-Bair, line 666: The strategy for incrementally capturing newly created list items here is:
1. Iterate through all value lists
2. Yield the items in each value list that were created after the `log_cursor`.

Do you have an idea of how many value lists users may have? If it's a lot, the current `fetch_incremental_list_items` function's strategy could block for a while and prevent timely shutdowns as it paginates through all lists and potentially through each lists' items. But if we don't anticipate there will be many value lists or many items in each list, then this strategy could be fine.

source-stripe-native/source_stripe_native/resources.py
```diff
@@ -81,6 +81,14 @@ async def check_accessibility(
         )
         is_accessible = not is_permission_blocked and not is_disabled
 
+        if is_permission_blocked:
+            log.warning(
+                (
+                    f"Removing inaccessible resource for {url} due to missing API key permissions. "
+                    "Update your restricted API key to grant read access if this resource should be captured."
+                )
+            )
```
Alex-Bair, line 91: Would it be helpful to always log this warning `if not is_accessible` and include the `err.message` in the log too? That way we'd always know why a stream is being excluded even if it's a Stripe account scoped issue rather than a permissions issue.

source-stripe-native/test.flow.yaml
```diff
@@ -230,3 +230,8 @@ captures:
         interval: PT5M
         schedule: 0 0 * * *
       target: acmeCo/ValueLists
+    - resource:
+        name: ValueListItems
+        interval: PT5M
+        schedule: 0 0 * * *
+      target: acmeCo/ValueListItems
```
Alex-Bair, line 237: Since the `ValueListItems` stream has the potential to block the event loop for a long time, what do you think about disabling this binding by default to avoid it rolling out to all existing captures? That way if we do see that it causes issues, the impact is limited to those captures that explicitly enabled it.

source-stripe-native/source_stripe_native/api.py
```diff
@@ -491,6 +491,173 @@ async def fetch_backfill_substreams(
...
+) -> AsyncGenerator[StripeObjectNoEvents | LogCursor, None]:
+    """
+    Incrementally captures a list item stream — a child stream of items listed
+    under a parent that has no events at the /events endpoint (e.g.
+    ValueListItems under ValueLists). Since Stripe emits no events for these
+    resources, "incremental" re-lists every parent and its items each sweep and
+    emits items whose `created` is at or after the cursor — the same
+    list-and-filter approach fetch_incremental_no_events uses for its base
+    streams, applied per parent. Because items are immutable, this captures all
+    new items; deletes are surfaced by the scheduled backfill.
```
Alex-Bair: Deletes aren't captured by the scheduled backfill. Scheduled backfills would only capture updates. But based on this docstring, list items can't be updated, so do we need a scheduled backfill for this stream?

Alex-Bair (commented):
Looks pretty good! Thanks for building out the Claude skills and investigating how to add those new streams. I had a few questions, mostly around the `ValueListItems` stream.

.gitignore
nicolaslazo: This is intentional and applies to all directories recursively, starting from the repo base dir. This is because Claude now generates `.claude/` directories anywhere a session is started and may include other user-specific files. For example, I have a number of `settings.local.json`s indicating which permissions I've permanently allowed.

I'm now realising that developers might want to know when skill files have been modified in the project-level `.claude/` directory. I'll update the glob to `*/**/.claude/`.

source-stripe-native/source_stripe_native/models.py
nicolaslazo: Right. I thought the runtime would be able to infer deletions, but a backfill trigger wipes all previous connector state.

source-stripe-native/source_stripe_native/resources.py
nicolaslazo: We discussed this on call, but I've added specific warnings for each logical branch. I've also noticed that we're still accepting resources that raise `HTTPError` if `is_accessible is True` and we decided it wasn't worth fixing if it wasn't broken.

source-stripe-native/source_stripe_native/api.py
nicolaslazo: Not really, we use created timestamp-based filtering now. It's fixed

source-stripe-native/source_stripe_native/api.py
nicolaslazo: We discussed this on call, but expected data volumes are unclear and we should prepare for the worst case scenario.

source-stripe-native/source_stripe_native/resources.py
```diff
@@ -74,6 +84,24 @@ async def check_accessibility(
...
+                f"Removing resource the account is not configured to use: {url}."
+            )
+        else:
+            warning_msg = (
+                f"Encountered HTTP error while checking accessibility for {url}"
+            )
+
+        log.warning(
+            warning_msg, extra={"err.code": err.code, "err.message": err.message}
+        )
```
Alex-Bair: nit: wanted to note that we're using the keyword argument `extra={}` in this log instead of passing in a raw dict like we do elsewhere. That'll cause logs to have the `err.code` and `err.message` fields in the `fields` object instead of in the `fields.args` object. That's fine, but I wanted to make sure that was intentional.

Alex-Bair (approved):
LGTM % one optional nit

source-stripe-native/source_stripe_native/resources.py
nicolaslazo: No good reason beyond force of habit. Let's follow convention instead 💪

PR #4627, opened 2026-06-05
materialize-clickhouse: persistent temp tables with authoritative reads

jacobmarble:
**Description:**

Permanent fix for the silent staged-row loss on clustered ClickHouse deployments (#4623 added detection; this removes the race):

- Do not `CREATE OR REPLACE` the stage table on every commit, to keep the table reference from becoming stale. Stage tables are kept for the lifetime of the binding; truncate cleans up after aborted transactions.
- Query the partition IDs from the stage table directly, using the MergeTree virtual column `_partition_id`, so that `select_sequential_consistency` works as intended (it is a no-op against `system.parts`).
- Keep the checks from #4623 in place, so that correctness is not lost in any event.

The load path gets the same treatment: persistent load tables truncated per round, a key-count check before the join, and the join now runs with `select_sequential_consistency = 1` so it also observes target rows committed by prior transactions' moves.

**Workflow steps:**

No happy-path change. The first Acknowledge of a session ensures each binding's temp tables (create if missing, re-create if schema drifted, truncate otherwise); MOVE PARTITION empties the stage table on every commit, so steady state issues no DDL at all.

None.

**Notes for reviewers:**

- Connector state no longer persists rendered SQL — only `StoredRows`. Acknowledge renders SQL from the live binding, which also fixes recovery of pre-#4623 checkpoints (their persisted single-column parts query can't satisfy the two-column scan).
- The pre-move row check is skipped during recovery: a crash mid-move legitimately leaves fewer staged rows, and an empty stage during recovery is a completed commit. This removes two potential crash-loops.
- `Destroy` no longer drops anything: stage tables may hold a committed-but-unacknowledged transaction's rows, and load tables follow the same lifecycle for symmetry. The only drop path is schema-drift re-creation in the ensure pass.
- Known limitation: a removed binding's temp tables are stranded (a removed binding is indistinguishable from a disabled one, whose pending staged rows must survive for re-enablement). Safe cleanup needs control-plane knowledge; deliberately out of scope.
- `TestAcknowledgeRecoveryMatrix` covers full/partial/already-completed recovery, leftover truncation, schema-drift re-creation, and pre-#4623 checkpoint compatibility.


materialize-clickhouse/driver.go
```diff
@@ -217,7 +213,21 @@ type transactor struct {
...
+	// completed. It is cleared after the first Acknowledge.
+	recovery bool
+	// ensured is set once the first Acknowledge of the session has run the
+	// ensure pass: recovering any pending commits and creating / truncating /
+	// re-creating the persistent temp tables for every binding. It is
+	// deliberately independent of the recovery flag, which is only set when
+	// the Open request carried prior connector state -- a brand-new task has
+	// no state but still needs its temp tables ensured before the first
+	// transaction.
+	ensured           bool
```
danielnelson, line 230: I wonder if the ensured steps should be appended to the Apply handling?

danielnelson (approved):

materialize-clickhouse/driver.go
jacobmarble: I looked at moving these into Apply but I don't think they fit there cleanly:

1. **The temp tables are range-scoped.** Every temp-table name embeds `_range.KeyBegin`, which comes from `open.Range` — i.e. the shard's key range. Apply operates on the materialization spec and has no shard range, so it doesn't know which ranges exist. A shard split creates a new range (and new temp-table names) with no corresponding Apply, so anything Apply created would be the wrong set. Only the session knows its own range.

2. **Two of the three things the ensure pass does are inherently per-session.** Recovering a committed-but-unacknowledged transaction needs the connector state (which Apply doesn't receive), and truncating rows left by a transaction that staged but never committed is per-session cleanup. Neither has a meaning in Apply.

3. **Create-if-missing / schema-drift re-create is the only piece that could plausibly go in Apply** (migrations only happen there), but splitting it out would put temp-table DDL in two places — potentially on different processes/connections — with ordering coupling, for no real gain. The ensure pass is idempotent, runs once at the first Acknowledge of a session (which the runtime serializes ahead of the first Store), and steady state issues no DDL at all since MOVE PARTITION empties the stage on every commit.

So I think keeping it in the session is the right call.

PR #4631, opened 2026-06-08
source-hubspot-native: allow delayed streams to checkpoint between chunks

Alex-Bair:
**Description:**

Delayed CRM streams previously emitted an entire `[since, until]` window as a single unit without intermediate checkpoints. If a date window was particularly dense, all records in that window had to be fetched before any checkpoints were yielded. This made it very difficult for delayed streams to make progress when there were 300k+ changes in a single hour; the connector couldn't fetch all records in the window before getting restarted.

This PR makes each delayed stream that uses the `fetch_search_objects` ID fetcher emit intermediate checkpoints after a chunk of ~10k records, which will let the delayed stream make incremental progress when working through a large chunk of changes.

The main changes in this PR include:
- `fetch_search_objects` now returns chunks of ~10k results in ascending timestamp order plus a `PageCursor` to allow the caller to fetch the next chunk of results. If there are no more results in the requested `[since, until]` window, a falsy `PageCursor` is returned to signal that pagination has ended.
- `_fetch_id_chunks` no longer ends pagination when it sees records with a timestamp below `since`. Instead, it relies on the fetcher function to signal the end of pagination with a falsy `PageCursor`.
- `fetch_chunked_changes_with_associations` works with the new chunking strategy by emitting a chunk of documents then the maximum timestamp within that chunk as an intermediate checkpoint. It assumes that chunks provided to it contain all records between its minimum and maximum timestamps & successive chunks advance in time, which `fetch_search_objects` does now.
- `fetch_delayed_changes` passes those intermediate datetime checkpoints from `fetch_chunked_changes_with_associations` through with some defensive guards to ensure checkpoints are strictly increasing.

**Notes for reviewers:**

We _may_ want to rewrite the real-time streams that use `fetch_search_objects` to use the chunking and intermediate checkpoint strategy as well, but right now they're not the ones struggling to make progress & I'd like to ensure the chunking strategy works well for the delayed streams before rolling it out for the real-time streams.

### DIFF
--- a/source-hubspot-native/source_hubspot_native/api/object_with_associations.py
+++ b/source-hubspot-native/source_hubspot_native/api/object_with_associations.py
@@ -35,12 +35,15 @@ _FetchIdsFn = Callable[
     Awaitable[tuple[Iterable[tuple[datetime, str]], PageCursor]],
 ]
 """
-Returns a stream of object IDs that can be used to fetch the full object details
-along with its associations. Used in `fetch_changes_with_associations`.
+Returns a page of object IDs that can be used to fetch the full object details
+along with its associations, plus a PageCursor for the next page (or a falsy
+value, e.g. None, when no more pages remain). Used by
+`fetch_changes_with_associations`.
 
-IDs may be returned in any order, but iteration will be stopped upon seeing an
-entry that's as-old or older than the datetime cursor. Entries newer than the
-until datetime will be discarded.
+IDs may be returned in any order. The fetcher is responsible for signaling
+completion via a falsy next_page: fetchers that page newest-first should do so
+once they reach an entry as-old-or-older than `since`, and window-bounded
+fetchers do so when their window is exhausted.
 """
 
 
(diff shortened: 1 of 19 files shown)


nicolaslazo (approved):
LGTM!

source-hubspot-native/source_hubspot_native/api/object_with_associations.py
```diff
@@ -217,27 +232,22 @@ async def fetch_batch_with_associations(
...
-    http: HTTPSession,
-    with_history: bool,
     since: datetime,
     until: datetime | None,
-) -> AsyncGenerator[tuple[datetime, str, CRMObject], None]:
-
-    # Walk pages of recent IDs until we see one which is as-old
-    # as `since`, or no pages remain.
-    recent: list[tuple[datetime, str]] = []
+) -> AsyncGenerator[tuple[list[tuple[datetime, str]], bool], None]:
```
nicolaslazo: Nit: we should consider turning some of these tuples into `NamedTuple`s to make call chains easier to follow through function signatures

Alex-Bair: Great suggestion, I've refactored a number of them into some `NamedTuple`s to improve readability in another commit.

nicolaslazo (approved):
This helps a lot, thanks!

Issue #3163, opened 2025-08-13
source-kafka: skip tables with schema references in discovery


jwhartley:
Can we skip discovery of schemas that use schema references? It's what we do for protobuf schemas, since we don't support those either.

(I'm assuming this is a small amount of work - if big don't bother)

Context From ChatGPT:
> **How common schema references are in Kafka**
> If using Schema Registry:
> - References are very common, especially in Avro and Protobuf formats.
> - Confluent added schema references in 2020 specifically so teams could avoid duplicating shared types.
> - Large orgs often have “base” schemas (e.g., CommonEventHeader, Address, UserIdentity) stored as separate schema IDs, and then topic-specific schemas reference them.
> - This pattern is almost the default in enterprises using Kafka with registry for multiple teams.
> 
> If not using Schema Registry (schemas embedded in the app or sent raw):
> - References are rare — messages usually have a fully inlined schema because there’s no central store to resolve them.
> - Developers just paste the full schema or generate it at build time.

PR #4634, opened 2026-06-08
source-kafka: resolve protobuf well-known type imports and harden dis…

### LINKED ISSUE #4633: source-kafka: protobuf schemas importing google well-known types break Schema Registry discovery
### Summary

When a Confluent Schema Registry contains a protobuf schema that imports a google well-known type such as `google/protobuf/timestamp.proto`, `source-kafka` fails during discovery:

```
Could not connect to the configured schema registry. Double check your configuration and try again.

Caused by:
    0: failed to create DescriptorPool from FileDescriptorSet
    1: imported file 'google/protobuf/timestamp.proto' has not been added
```

The "Could not connect" wrapper is misleading. The registry connects fine; the failure happens while building the protobuf `DescriptorPool`.

Discovery fetches and parses the schemas for every topic in the registry, so one such schema fails discovery for the whole registry, including unrelated Avro and JSON topics. An Avro-only user is blocked when another team publishes a protobuf topic with a well-known type import into the same registry.

### Root cause

In `schema_registry.rs`, `resolve_protobuf_references_inner` does not fetch `google/protobuf/*` imports from the registry, because prost-reflect bundles the well-known types:

```rust
if reference.name.starts_with("google/protobuf/") {
    continue;
}
```

`fetch_protobuf_schema` then builds the pool from an empty pool:

```rust
let descriptor_pool = DescriptorPool::from_file_descriptor_set(file_descriptor_set)?;
```

`from_file_descriptor_set` starts empty, so the skipped well-known types are never present. Any schema whose `dependency` list includes `google/protobuf/timestamp.proto` then fails to resolve. The read path in `pull.rs` builds the pool the same way, so decoding is affected too.

### Fix

Build from `DescriptorPool::global()`, which already contains the google well-known types, and add the fetched files into it. `global()` returns a copy-on-write clone, so per-schema additions stay local and do not leak across topics.

### Reproduction

1. Run a Confluent Schema Registry and Kafka (for example `confluentinc/cp-schema-registry` and `confluentinc/cp-kafka`).
2. Register a protobuf value schema that imports the well-known type. No `references` entry is needed; the registry resolves google well-known types itself:
   ```protobuf
   syntax = "proto3";
   import "google/protobuf/timestamp.proto";
   message Event { string id = 1; google.protobuf.Timestamp ts = 2; }
   ```
3. Run `flowctl raw discover` against the registry. It fails with the error above and returns zero bindings, even when there are valid Avro topics present.

### Impact

This blocks any user whose registry contains protobuf schemas importing google well-known types, which is common given how often `Timestamp`, `Duration`, and the wrapper types are used.

### Also addressed alongside this

Discovery is currently all-or-nothing: one unparseable or unsupported schema fails discovery for the whole registry. The fix PR also makes discovery skip such a schema with a warning, and the affected topic falls back to schema inference with the default collection key, so the other topics stay discoverable. This covers the concern in #3163, which proposed skipping reference-using schemas. The PR keeps the topic and degrades it rather than dropping it. Protobuf references are now supported, so the remaining unsupported case is Avro and JSON references.

SeanWhelan:
**Description:**

Two fixes to `source-kafka` Schema Registry handling (details in #4633):

1. Protobuf schemas importing google well-known types (e.g. `Timestamp`) failed discovery and decoding with `imported file 'google/protobuf/timestamp.proto' has not been added`. The pool was built from an empty `from_file_descriptor_set`; it now builds from `DescriptorPool::global()`, which already contains the well-known types.
2. Discovery resilience: one unsupported schema (e.g. Avro/JSON with references) no longer fails discovery for the whole registry. The affected topic is skipped with a warning and falls back to the default collection key. Registry connectivity failures still fail discovery.

**Notes for reviewers:**

- `global()` returns a copy-on-write clone, so per-schema additions don't leak across topics.
- Regression tests in `tests/test.rs`: `protobuf-wkt-topic` (imports `timestamp.proto`; discovered and decoded to RFC3339) and `avro-ref-topic` (unsupported Avro reference; proves discovery survives). Existing topics' output is unchanged. These integration tests aren't in the Docker build (`tests/` isn't copied), same as the existing protobuf tests; validated locally, full suite green.
- Re #3163: this addresses the concern (a reference schema no longer breaks discovery) by degrading the topic rather than skipping it. Leaving for James to decide whether to close.

Closes #4633. Related to #3163.


nicolaslazo (approved):
LGTM 👍

PR #4636, opened 2026-06-09
source-shopify-native: conditionally add `retailLocation` to `orders` stream

### LINKED ISSUE #4342: source-shopify-native: add `retailLocation` field to `orders` stream
## Problem

The `orders` stream doesn't include the [`retailLocation` field](https://shopify.dev/docs/api/admin-graphql/latest/queries/orders#returns-nodes.fields.retailLocation). Only the [`id` field](https://shopify.dev/docs/api/admin-graphql/latest/queries/orders#returns-nodes.fields.retailLocation) on `retailLocation` needs included for now.

Note that this field has access requirements beyond the `QUALIFYING_SCOPES` the `orders` stream currently requires - reading `retailLocation.id` requires the `read_locations` scope.

## Possible solution

Conditionally include the `retailLocation.id` field in the `orders` stream if and only if the `read_locations` scope is present.

Alex-Bair:
**Regression?**

No

**Description:**

Addresses #4342. 

The `retailLocation` can only be queried on `orders` when the `read_locations` scope is granted, which the `orders` stream doesn't need otherwise. Instead of forcing every user that enables `orders` to grant the `read_locations` scope, this PR adds a general mechanism for conditionally including query fields based on various per-store capabilities (mostly scopes and the store type), and the connector now uses that mechanism to include `retailLocation.id` in `orders` when we have the `read_locations` scope.

**Notes for reviewers:**

The various infrastructure around conditionally including fields in queries may seem a little overkill when only the `retailLocation` field in `orders` uses it, but I'm going to update the connector to conditionally include more fields over the next week & the more general handling will pay off then.

I recommend reviewing commit-by-commit.

### DIFF
--- a/source-shopify-native/source_shopify_native/api.py
+++ b/source-shopify-native/source_shopify_native/api.py
@@ -13,6 +13,7 @@ from source_shopify_native.graphql.bulk_job_manager import BulkJobManager
 from source_shopify_native.models import (
     BaseResponseData,
     ShopifyGraphQLResource,
+    StoreCapabilities,
     StoreValidationContext,
     TShopifyGraphQLResource,
 )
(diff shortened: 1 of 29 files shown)


nicolaslazo (approved):
LGTM!

PR #4641, opened 2026-06-09
source-mysql: Fix bug in MariaDB JSON column DDL parsing

### LINKED ISSUE #4638: source-mariadb: JSON columns discovered as `string` but captured as parsed JSON → write-schema validation failure
## Summary

source-mariadb discovers MariaDB `JSON` columns with `type: [string, null]`, but at capture time the binlog value arrives as parsed JSON (object/array). The captured document then fails validation against its own connector-provided write schema:

```
document failed validation against its collection JSON Schema
  absoluteKeywordLocation: flow://connector-schema#/$defs/.../properties/<col>
  detail: "Type mismatch: expected a null or string"
  instanceLocation: /<col>
  instanceValue: "<array>"
```

This crash-loops the capture shard (shardFailure), which blocks every binding on the task, not just the affected one.

## Root cause

MariaDB has no native JSON type — `JSON` is an alias for `LONGTEXT COLLATE utf8mb4_bin` with a `CHECK (json_valid(col))` constraint, and `information_schema.columns.DATA_TYPE` reports it as `longtext` (https://mariadb.com/kb/en/json-data-type/).

Discovery maps types purely from `DATA_TYPE` and never inspects the `json_valid` CHECK constraint:
- `source-mysql/discovery.go` `mysqlTypeToJSON`: `longtext → {jsonType:"string"}`, `json → {}`.

So a MariaDB JSON column is discovered as `string`, but the value streaming through replication is parsed JSON that no longer matches. This is long-known:
- #404 (2022): "JSON is an alias for the LONGTEXT column type, so discovery and captures will just treat that as a string" — deliberately left unfixed.
- `source-mysql/datatype_test.go:125` carries the explicit "this test fails on MariaDB" comment.

The materialize side was addressed (PR #2053 — treat JSON as compatible with longtext for MariaDB), but the source/discovery side was not.

> Note: #404 expected MariaDB JSON to be captured as a *string* (which would validate). In practice we are now seeing the value captured as a parsed *array/object*, which fails. Worth confirming the exact path by which a `longtext`-discovered column ends up transcoded as raw JSON at replication time (binlog column metadata vs. the discovered type), since that detail decides between the two fixes below.

## Impact

- Any MariaDB customer who adds/uses a JSON column on a captured table hits a hard, recurring capture failure.
- Re-discovery does not help (discovery re-reads `longtext` → `string` every time).
- Backfill only masks it (SELECT returns text → validates; the next CDC insert re-breaks).
- The only customer workarounds today are: disable auto-discovery + manually widen the write schema, or change the source column to a non-JSON text type. Both are unsatisfying.

## Proposed fix

Detect MariaDB JSON columns during discovery and type them as free-form JSON (`{}`), matching MySQL's `json` handling:
- Inspect `information_schema.CHECK_CONSTRAINTS` / `information_schema.TABLE_CONSTRAINTS` (or the column's check clause) for a `json_valid(<col>)` predicate, and/or
- Use the binlog table-map column metadata to recognize the JSON type, and ensure the replication transcoder and the discovered schema agree.

Either way, discovery and the runtime transcoder must agree so the captured value matches the write schema.

## Repro
- MariaDB source, table with a `JSON` column.
- Insert a row whose JSON value is an array or object.
- Capture fails write-schema validation as above.

willdonnelly:
**Description:**

In MariaDB the "JSON" column type is an alias for "LONGTEXT COLLATE utf8mb4_bin" and so normally `source-mariadb` (`source-mysql`) discovers such columns as strings and captures them as strings.

However, we didn't previously have any logic to resolve that alias when processing a _DDL alteration_ on such a column, so if one were to issue an `ALTER TABLE ... ADD COLUMN foo JSON` on a table which was already part of an ongoing capture, the internal metadata would treat that as a `json` column type, which happens to mean the _MySQL_ JSON column type. Once this happened, subsequent capture outputs would contain bare JSON values which wouldn't match the discovered type.

This PR fixes that by resolving the "JSON" column type to "LONGTEXT" when the target database is MariaDB, so that DDL tracking again matches the discovery-initialization metadata pathway.

(Note: We should _also_ consider adding a feature flag and eventually flipping the default behavior so that MariaDB JSON columns are captured as JSON values consistent with most other databases. However the immediate issue being solved in this PR is just that we can get into a state where we capture values which don't match our own discovered schemas)

Fixes https://github.com/estuary/connectors/issues/4638

### DIFF
--- a/source-mysql/.snapshots/TestMariaDBAddJSONColumn
+++ b/source-mysql/.snapshots/TestMariaDBAddJSONColumn
@@ -2,10 +2,10 @@
 # Collection "acmeCo/test/test/mariadbaddjsoncolumn_31415926": 3 Documents
 # ================================
 {"_meta":{"op":"c","source":{"schema":"test","snapshot":true,"table":"MariaDBAddJSONColumn_31415926","cursor":"backfill:0"}},"data":"before","id":1}
-{"_meta":{"op":"c","source":{"schema":"test","table":"MariaDBAddJSONColumn_31415926","cursor":"binlog.000123:56789:123"}},"data":"obj","id":2,"j":{"bucket_id":1,"id":1}}
-{"_meta":{"op":"c","source":{"schema":"test","table":"MariaDBAddJSONColumn_31415926","cursor":"binlog.000123:56789:123"}},"data":"arr","id":3,"j":[{"bucket_id":1,"id":1}]}
+{"_meta":{"op":"c","source":{"schema":"test","table":"MariaDBAddJSONColumn_31415926","cursor":"binlog.000123:56789:123"}},"data":"obj","id":2,"j":"{\"bucket_id\": 1, \"id\": 1}"}
+{"_meta":{"op":"c","source":{"schema":"test","table":"MariaDBAddJSONColumn_31415926","cursor":"binlog.000123:56789:123"}},"data":"arr","id":3,"j":"[{\"bucket_id\": 1, \"id\": 1}]"}
 # ================================
 # Final State Checkpoint
 # ================================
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM

PR #4642, opened 2026-06-09
Adding a feature to prevent sub account fan out for main api tokens

jonwihl:
**Description:**

Introduces a change that prevents sub-account fan out when users are authenticating with a main api token. Sub accounts require using a sub-account api token. This allows a user to set up a Twilio capture and only target resources that belong the main account sid defined in the configuration.

**Workflow steps:**

[Twilio Sub-Account limitations](https://www.twilio.com/docs/iam/api/subaccounts#authentication)

**Notes for reviewers:**


twilio-patch.patch
Alex-Bair, line 1: This file isn't needed, it should be removed before merging.

source-twilio/VERSION
Alex-Bair, line 1: I don't see any backwards-incompatible changes in this PR, so we don't need to increment the connector's version.

Alex-Bair (approved):
LGTM % removing the `.patch` file and not bumping the `VERSION.

The connector's documentation should also be updated to reflect the new field in the endpoint config.

PR #4643, opened 2026-06-09
source-mongodb: debug log when skipping change events with null fullDocument

Alex-Bair:
**Regression?**

No

**Description:**

In `updateLookup` mode the connector drops non-delete change events whose `fullDocument` is `null`, assuming the document was deleted between the update and the lookup, and advances the cursor past them. If that assumption is ever wrong, an update is silently lost. That's hard to diagnose because the skip currently leaves no trace.

This commit adds a debug log at the skip recording identifying info so a skipped event can be correlated against a reported missing document.


nicolaslazo (approved):
LGTM

Issue #4479, opened 2026-05-14
SQL and Warehouse Materializations: support feature flags as advanced configuration options


jonwihl:
The following feature flags should become permanent configuration options in the 'Advanced Configuration' section:
- allow_existing_tables_for_new_bindings
- retain_existing_data_on_backfill
- always_drop_tables_on_backfill

Support considerations:

- Should support migrate all existings feature flags to configuration options?

jwhartley:
[Internal thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1778788849805369)

PR #4646, opened 2026-06-10
automatic feature flags setting for materializations

### LINKED ISSUE #4599: Automatic feature flag releases
Work with the team to design a new process for releasing “potentially” breaking changes (even after verifications have been done to ensure no *known* incompatibilities), so that we can release changes without affecting existing tasks whenever possible

See https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1779973873783469?thread_ts=1779973707.705819&cid=C03Q2NRFKDL

mdibaiee:
**Description:**

Materialization feature flags gain a third kind of default, alongside enabled-for-all and disabled-for-all: **enabled only for tasks created on or after a cutoff date**. This is how we change a default behavior without disturbing existing tasks, replacing the manual "bulk-edit `no_<flag>` into every production config, then merge a second PR flipping the default" process.

A connector author declares it in the defaults map:

```go
var featureFlagDefaults = map[string]common.FlagDefault{
	"existing_thing": common.FlagEnabled,
	"new_thing":      common.FlagEnabledForTasksCreatedAfter("2026-06-10"),
}
```

Resolution uses the `created_at` date that estuary/flow@4d66c8db stamps onto built task specs (requires flow v0.6.12, bumped here). Precedence per flag: an explicit `<flag>` / `no_<flag>` in `/advanced/feature_flags` wins, then the cutoff comparison, then the fixed default. `created_at` is empty only during a task's very first build, before its creation is committed — that means brand new, and resolves every cutoff-gated flag on.

Open then emits a `configUpdate` restating the endpoint config with each cutoff-gated flag written out explicitly, so a task's flags can be read off its config instead of inferred from its creation date. The restatement writes the flags into the sealed config's plaintext `sops.overlay` (#4941 marks `feature_flags` `nonsensitive`), so nothing is decrypted, re-encrypted, or re-keyed.

No flag uses a cutoff yet; this PR only wires up the mechanism.

**Workflow steps:**

For connector authors only — no user-visible change. To change a default behavior, add the flag with `FlagEnabledForTasksCreatedAfter(<ship date>)` and merge. Tasks that exist on that date keep the old behavior for the rest of their lives; tasks created afterwards get the new behavior from birth. Support can still override either way per-task via `/advanced/feature_flags`, which takes priority over the cutoff.

The cutoff is a date rather than a timestamp because a date is all a spec records about when a task was created, so the two are compared directly. A flag takes effect for the whole of the day it ships on.

None user-facing. Internal `docs/feature_flags.md` gains a "Changing Default Behaviors (Materializations)" section covering resolution and the config recording; the existing bulk-edit workflow is relabelled as the fallback for captures.

**Notes for reviewers:**

- Resolution is a pure function of `(feature_flags string, spec.CreatedAt)`, so Validate, Apply and Open agree without persisting anything. Validate takes the date from `LastMaterialization` (the proposed spec isn't in the request), Apply and Open from `req.Materialization`. All three agree because the date is derived from the task's immutable control-plane ID.
- **Recording happens at Open** because `sealed_config_json` is an Open-only request field; Apply has no sealed config to restate. `WithConfigUpdates` is correspondingly an `OpenOption`.
- **The `configUpdate` is a record, not a decision.** It writes the value the connector already resolved, so the config can never contradict runtime behavior and the task behaves identically before and after it lands. That is what lets it be asynchronous and best-effort — failures are logged and ignored, and nothing is persisted in connector state.
- A `configUpdate` racing a user edit or another publication can't clobber it: the control plane only applies one whose build matches the task's current `last_build_id`, and publishes with `expect_pub_id`. So at most one config update takes effect per build, and stale ones are dropped.
- Only cutoff-gated flags are recorded with `configUpdate`. Recording fixed defaults would freeze them per-task and defeat any later fleet-wide change of a default.
- Encrypted values are copied across untouched; the restatement is otherwise re-serialized, which sorts its properties. That is safe even though sops MACs a document's values in the order it encounters them, because the control plane parses the emitted config into a `serde_json::Value` and re-serializes it when applying the update, sorting it the same way. An unencrypted config has no ciphertext to preserve and is restated directly, rather than being given a `sops` stanza the runtime would then look for ciphertext in.
- Emitting no longer contacts the config-encryption service, so no secret leaves the connector and a config keeps the key it was sealed with — including a customer-managed key Estuary cannot encrypt to.
- A spec built before flow stamped `created_at` has no date and would resolve as brand new. This means this PR has to wait 20 days since the release of the `created_at` change before we can release, to make sure all tasks that have been enabled, do have a `created_at`. Disabled tasks will end up having `created_at` when their enabling publication goes through.

**Verified end-to-end** on a local data plane (materialize-postgres, age-sealed config, cutoff-gated probe flag): the connector emits one `configUpdate` carrying `sops.overlay.advanced.feature_flags` with the ciphertext untouched; the control plane republishes; the task's next Open decrypts the round-tripped config and resolves the flag from it, so nothing is re-emitted on later builds. Publishing an overlay that also rewrites `/address` is rejected with `overlay modifies location '/address', which is not marked nonsensitive`.

### DIFF
--- a/docs/feature_flags.md
+++ b/docs/feature_flags.md
@@ -32,6 +32,17 @@ var featureFlagDefaults = map[string]bool{
 }
 ```
 
+Materialization connectors instead hold `common.FlagDefault` values, so that a flag's
+default can be gated on the task's creation date — see
+[Changing Default Behaviors (Materializations)](#changing-default-behaviors-materializations):
+
+```go
+var featureFlagDefaults = map[string]common.FlagDefault{
+  // A short (one or two line) description of what this flag does when set/unset.
+  "do_something": common.FlagDisabled,
+}
+```
+
 Then at the appropriate point(s) in the code, introduce the new logic conditional on the
 flag setting:
 
@@ -51,9 +62,79 @@ Typically this should be default-off, but if it makes things clearer there shoul
 be anything wrong with adding a default-on flag instead and having `no_foobar` be the
 setting which opts into the new behavior.
 
-## Changing Default Behaviors
+## Changing Default Behaviors (Materializations)
 
-The process for changing a default connector behavior is as follows:
+Materialization connectors can gate a flag's default on when a task was created,
+which changes the default for new tasks only, with no config edits and no
+coordinated merge. This is the preferred approach; the manual bulk-editing process in
+the next section remains for captures, which don't yet resolve flags this way.
+
+The default-values map holds `common.FlagDefault` values rather than plain booleans:
+
+```go
+var featureFlagDefaults = map[string]common.FlagDefault{
+  // A short (one or two line) description of what this flag does when set/unset.
+  "do_something": common.FlagDisabled,
+  // Applies only to tasks created on or after the date this flag was released,
+  // so that tasks already running at that point keep the old behavior.
+  "new_thing": common.FlagEnabledForTasksCreatedAfter("2026-06-10"),
+}
+```
+
+Set the cutoff to the date the flag ships. Every task in existence at that point is
+older than the cutoff and keeps the old behavior for the rest of its life, while every
+task created afterwards gets the new behavior from its first publication. A user can
+still opt either way per-task with `new_thing` / `no_new_thing` in
+`/advanced/feature_flags`, which always wins over the cutoff.
+
+The cutoff is a date, not a timestamp, because a date is all a spec records about when
+a task was created — the two dates are compared directly. A flag therefore takes effect
+for the whole of the day it ships on, so pick the date accordingly if that matters.
+
+### How it resolves
+
+Built capture and materialization specs carry a `created_at` date (UTC, `YYYY-MM-DD`),
+derived by the control plane from the task's control-plane ID. The materialization
+boilerplate resolves flags from that date in every RPC:
+
+| RPC | Creation date taken from |
+| --- | --- |
+| Validate | `req.LastMaterialization` — absent for a task that has never been published |
+| Apply | `req.Materialization` |
+| Open | `req.Materialization` |
+
+Resolution is therefore a pure function of the spec: it is stable for the life of the
+task, identical across the three RPCs, and needs nothing persisted in connector state.
+
+The date is empty only during a task's very first build, before its creation is
+committed and an ID assigned. That means "brand new", and resolves as though the task
+were created now — enabling every cutoff-gated flag, which is correct because a cutoff
+is never in the future. Since the date is derived from the task's ID rather than
+recorded at build time, it is present on every spec built since the field was
+introduced, including those of long-running tasks.
+
+### Recording the resolved value in the config
+
+So that a task's flags can be read off its config rather than inferred from its
+creation date, Apply emits a [`configUpdate`](observable_logs.md) event restating the
+endpoint config with each cutoff-gated flag written out explicitly — `new_thing` or
+`no_new_thing` in `/advanced/feature_flags`. It is recomputed on every Apply and stops
+being emitted once the config contains the flag.
+
+This is a *record*, not a decision. The value written is the one the connector already
+resolved from the creation date, so the config can never contradict the connector's
+behavior, and a task behaves identically whether or not the update has landed yet —
+which is what makes it safe for the write to be asynchronous and best-effort. Failures
+are logged and ignored.
+
+Only cutoff-gated flags are recorded. Flags with a fixed default are deliberately left
+out: writing those into every config would freeze them per-task and defeat any later
+fleet-wide change of the default.
+
+## Changing Default Behaviors Manually
+
+For captures, and for materialization flags that can't be expressed as a creation-date
+cutoff, the process for changing a default connector behavior is as follows:
 
  - Add the feature flag `new_thing` with a default value corresponding to the old behavior (typically false).
  - Edit the endpoint configs of all tasks in production to add the `no_new_thing` flag setting.
--- a/docs/feature_flags.md
+++ b/docs/feature_flags.md
@@ -116,10 +116,26 @@ introduced, including those of long-running tasks.
 ### Recording the resolved value in the config
 
 So that a task's flags can be read off its config rather than inferred from its
-creation date, Apply emits a [`configUpdate`](observable_logs.md) event restating the
+creation date, Open emits a [`configUpdate`](observable_logs.md) event restating the
 endpoint config with each cutoff-gated flag written out explicitly — `new_thing` or
-`no_new_thing` in `/advanced/feature_flags`. It is recomputed on every Apply and stops
+`no_new_thing` in `/advanced/feature_flags`. It is recomputed on every Open and stops
 being emitted once the config contains the flag.
+
+The restatement never decrypts and re-encrypts the config. Open carries the task's
+sealed config alongside its decrypted one, and the flag is written into the plaintext
+`sops.overlay` of that document: the runtime merges the overlay over the decrypted
+config, having first checked that every location it touches is annotated
+`nonsensitive: true` in the connector's config schema — which is why
+`/advanced/feature_flags` must carry `jsonschema_extras:"nonsensitive=true"`. The
+encrypted values are copied across untouched, so a config stays under the key it was
+encrypted with, including a customer-managed one that Estuary cannot itself encrypt to.
+A config that isn't encrypted has no ciphertext to preserve and is restated directly.
+
+Property order does matter to sops, which computes its MAC over a document's values in
+the order it encounters them — but it is not something the connector has to preserve.
+The control plane parses the emitted config and re-serializes it when applying the
+update, which sorts its properties, so the restatement is sorted here as well and a
+config that verifies once stored verifies as emitted.
 
 This is a *record*, not a decision. The value written is the one the connector already
 resolved from the creation date, so the config can never contradict the connector's
--- a/docs/feature_flags.md
+++ b/docs/feature_flags.md
@@ -81,11 +81,11 @@ var featureFlagDefaults = map[string]common.FlagDefault{
 }
 ```
 
-Set the cutoff to the date the flag ships. Every task in existence at that point is
-older than the cutoff and keeps the old behavior for the rest of its life, while every
-task created afterwards gets the new behavior from its first publication. A user can
-still opt either way per-task with `new_thing` / `no_new_thing` in
-`/advanced/feature_flags`, which always wins over the cutoff.
+Set the cutoff to the day after the pull request merges. Every task in existence at
+that point is older than the cutoff and keeps the old behavior for the rest of its
+life, while every task created afterwards gets the new behavior from its first
+publication. A user can still opt either way per-task with `new_thing` /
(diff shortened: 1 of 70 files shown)


docs/observable_logs.md
```diff
@@ -0,0 +1,42 @@
+# Operator-observable logs
```
mdibaiee, line 1: Docs for https://github.com/estuary/flow/pull/3010 -- included in this PR just for convenience

go/common/config_update.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, fmt.Errorf("encoding request body: %w", err)
+	}
+
+	var url = defaultEncryptionURL
+	if fromEnv := os.Getenv("ENCRYPTION_URL"); fromEnv != "" {
+		url = fromEnv
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
```
jacobmarble: It's often nicer to find this at the top of a function.

materialize-bigquery/bigquery.go
```diff
@@ -22,14 +23,14 @@ import (
...
-	"objects_and_arrays_as_json":       true,
-	"datetime_keys_as_string":          true,
-	"retain_existing_data_on_backfill": false,
-	"skip_cleanup":                     false,
-	"native_binary_column_type":        true,
+	"objects_and_arrays_as_json":       common.FlagEnabled,
+	"datetime_keys_as_string":          common.FlagEnabled,
+	"retain_existing_data_on_backfill": common.FlagDisabled,
+	"skip_cleanup":                     common.FlagDisabled,
+	"native_binary_column_type":        common.FlagEnabled,
```
jacobmarble, line 33: This is very nice!

materialize-boilerplate/flags_test.go
```diff
@@ -0,0 +1,198 @@
...
+	raw string
+}
+
+func (c flagsTestConfig) Validate() error          { return nil }
+func (c flagsTestConfig) DefaultNamespace() string { return "" }
+func (c flagsTestConfig) FeatureFlags() (string, map[string]common.FlagDefault) {
+	return c.raw, map[string]common.FlagDefault{
+		"fixed_on":  common.FlagEnabled,
+		"fixed_off": common.FlagDisabled,
+		"gated":     common.FlagEnabledForTasksCreatedAfter("2026-06-10"),
```
jacobmarble, line 28: ❤️

jacobmarble (dismissed):

go/common/config_update.go
```diff
@@ -0,0 +1,123 @@
...
+		"schema": schema,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("encoding request body: %w", err)
+	}
+
+	var url = defaultEncryptionURL
+	if fromEnv := os.Getenv("ENCRYPTION_URL"); fromEnv != "" {
+		url = fromEnv
+	}
```
danielnelson: Will this variable be automatically set on a local stack, or do we need to arrange for it?

go/common/config_update_test.go
```diff
@@ -0,0 +1,141 @@
...
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "bad schema", http.StatusBadRequest)
+	}))
+	defer server.Close()
+	t.Setenv("ENCRYPTION_URL", server.URL)
+
+	hook := logtest.NewGlobal()
+	defer hook.Reset()
+
+	err := EmitConfigUpdate(context.Background(), "updating config", json.RawMessage(`{}`), json.RawMessage(`{}`))
```
danielnelson:
```suggestion
	err := EmitConfigUpdate(t.Context(), "updating config", json.RawMessage(`{}`), json.RawMessage(`{}`))
```

docs/feature_flags.md
```diff
@@ -51,9 +62,79 @@ Typically this should be default-off, but if it makes things clearer there shoul
...
+
+Set the cutoff to the date the flag ships. Every task in existence at that point is
+older than the cutoff and keeps the old behavior for the rest of its life, while every
+task created afterwards gets the new behavior from its first publication. A user can
+still opt either way per-task with `new_thing` / `no_new_thing` in
+`/advanced/feature_flags`, which always wins over the cutoff.
+
+The cutoff is a date, not a timestamp, because a date is all a spec records about when
+a task was created — the two dates are compared directly. A flag therefore takes effect
+for the whole of the day it ships on, so pick the date accordingly if that matters.
```
danielnelson, line 92: Setting the date is a little awkward but doable if the flag is new, do you suggest we land the code first in the disabled position and then switch to a date as a second commit?  Or perhaps we use date cutoff flags only when changing the default of an existing flag?

go/common/feature_flags.go
```diff
@@ -1,6 +1,188 @@
...
+// The date is derived from the task's control-plane ID rather than recorded at
+// build time, so every spec built since the field was introduced carries it,
+// including those of long-running tasks. A spec built before the field existed
+// has no date at all and resolves as brand new (see Resolve) — but a task can
+// only reach a connector version that knows about this flag by being
+// republished, which rebuilds its spec and stamps the date.
+//
+// Panics if date is not a valid date in CreatedAtLayout form, which is a
+// connector programming error.
+func FlagEnabledForTasksCreatedAfter(date string) FlagDefault {
```
danielnelson, line 87: We can do this later, but I guess we will want the inverse flag? `DisabledForTasksCreatedAfter`

go/common/config_update.go
```diff
@@ -0,0 +1,123 @@
...
+	"net/http"
+	"os"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// defaultEncryptionURL is Estuary's config encryption service, which encrypts
+// any config properties marked as secret in the schema using sops.
+const defaultEncryptionURL = "https://config-encryption.estuary.dev/v1/encrypt-config"
```
danielnelson: Not sure where to hang this.  Did we figure out what happens if the task is using a private KMS?

danielnelson (commented):
Do you have guidance on when we should use date cutoff feature flags vs simple enable/disable flags?

go/common/config_update.go
mdibaiee: See https://github.com/estuary/flow/pull/3291

go/common/feature_flags.go
mdibaiee: yeah we can do that later when the need arises

mdibaiee:
> Do you have guidance on when we should use date cutoff feature flags vs simple enable/disable flags?

I think most of the time we want a date cutoff feature flag so new tasks pick up the new behaviour, unless something really is an opt-in... which usually should instead be an advanced config instead of a feature flag IMO

go/common/config_update.go
mdibaiee: I'll wait for #4941 to be merged and then rebase this with sops.overlay use so that it works for customers with custom KMS too

mdibaiee:
I verified this end-to-end with a local mise stack of flow.

danielnelson:
> > Do you have guidance on when we should use date cutoff feature flags vs simple enable/disable flags?
> 
> I think most of the time we want a date cutoff feature flag so new tasks pick up the new behaviour, unless something really is an opt-in... which usually should instead be an advanced config instead of a feature flag IMO

I haven't thought too deeply about this, but perhaps feature flags should be used only for rollout?  We have some flags currently that don't meet this standard, I think they are flags to make them less visible in the endpoint.  Examples are `retain_existing_data_on_backfill`, `always_drop_tables_on_backfill`, `nested_dot_hash_location_style`.  Do you think these ought to be made into advanced options?

mdibaiee:
@danielnelson yeah I think those are better off as advanced knobs and not feature flags. I believe there was a ticket about this actually, see: https://github.com/estuary/connectors/issues/4479

docs/feature_flags.md
mdibaiee: I think we set the date of a feature flag to a date in the future, and merge it before that date. So for example we set:

my_flag: 2026-09-09

Then we merge the PR on the last hours of 2026-09-08.

Any tasks created on and after 09-09 will get the feature flag enabled.

danielnelson (dismissed):

materialize-boilerplate/materializer.go
```diff
@@ -1028,16 +1112,43 @@ func (c *constrainterAdapter[EC, FC, RC, MT]) DescriptionForType(p *pf.Projectio
 	return mt.String(), nil
 }
 
+// ParseFlags resolves feature flags without a task spec to take the creation
```
willdonnelly: I don't think this fallback should exist. If we're using date-based resolution it should apply everywhere, and if there are places it can't be used that suggests a fundamental problem with this approach.

docs/feature_flags.md
```diff
@@ -51,9 +62,95 @@ Typically this should be default-off, but if it makes things clearer there shoul
...
+  // A short (one or two line) description of what this flag does when set/unset.
+  "do_something": common.FlagDisabled,
+  // Applies only to tasks created on or after the date this flag was released,
+  // so that tasks already running at that point keep the old behavior.
+  "new_thing": common.FlagEnabledForTasksCreatedAfter("2026-06-10"),
+}
+```
 
-The process for changing a default connector behavior is as follows:
+Set the cutoff to the date the flag ships. Every task in existence at that point is
```
willdonnelly: Maybe obvious but just to make sure it's noted: this instruction should be updated if the recommendation is now "make sure the date is set to _tomorrow_ before merging"

docs/feature_flags.md
```diff
@@ -51,9 +62,95 @@ Typically this should be default-off, but if it makes things clearer there shoul
...
+| RPC | Creation date taken from |
+| --- | --- |
+| Validate | `req.LastMaterialization` — absent for a task that has never been published |
+| Apply | `req.Materialization` |
+| Open | `req.Materialization` |
+
+Resolution is therefore a pure function of the spec: it is stable for the life of the
+task, identical across the three RPCs, and needs nothing persisted in connector state.
+
+The date is empty only during a task's very first build, before its creation is
```
willdonnelly: This paragraph is not very helpful to a casual reader. What does "during a task's very first build" mean in terms of actual validation/publication/discovery behavior? More precisely: will the first actual "for real" execution of a capture/materialization have a created_at date or not? It sounds like the answer is a restatement of the table above as "the very first Validate() run won't have a date but everything afterwards including the first real data-moving invocation will", right?

willdonnelly (dismissed):
LGTM, I guess. I'm still not a fan of this approach, but my objections are primarily aesthetic (the feature_flags list will be cluttered with entries no human ever added manually, every new flag will cause a bunch of publication churn as every task updates itself, small windows of time around date changes where I can't _prove_ things would misbehave but it's hard to rule out entirely, and we have to take a bit of extra care around enabled-after dates to avoid mistakes) and nothing that makes this fundamentally unsound.

materialize-boilerplate/materializer.go
mdibaiee: yep 100% agree, fixing this 👍🏽

danielnelson (approved):

PR #4647, opened 2026-06-10
source-shopify-native: add `staff_members` stream

### LINKED ISSUE #4226: source-shopify-native: add a `staff_members` stream
Add a `staff_members` stream that captures from the [`StaffMember` object](https://shopify.dev/docs/api/admin-graphql/latest/objects/staffmember).

Access to the `StaffMember` object requires the `read_users` scope, which I think is only available on Shopify Plus / Advanced stores.

There doesn't appear to be an updated at or created at field for `StaffMember`, so this will likely need to be a full refresh or periodic backfill stream.

Alex-Bair:
**Regression?**

No.

**Description:**

This PR's scope includes:
- Updating the CDK to support multiple `fetch_snapshot` subtasks. We previously weren't sure how this would work since the CDK automatically manages the `_meta/row_id` field and emits tombstones to infer deletions between snapshots, but multiple `fetch_snapshot` subtasks should actually work fine as long as the collection key includes an additional discriminator that identifies which subtask documents and tombstones came from. ex: If the collection key is `["_meta/subtask_id", "_meta/row_id"]`, then the CDK's deletion inference mechanism should work still since deletions between each subtask's snapshots will have a unique identifier per subtask.
- Updating `source-shopify-native` to support full refresh streams, using the CDK's `fetch_snapshot` subtask support & a collection key of `["_meta/store", "_meta/row_id"]`.
- Adding `staff_members` as a full refresh stream in `source-shopify-native` that captures Shopify's [`StaffMember`](https://shopify.dev/docs/api/admin-graphql/latest/objects/StaffMember) object.

See individual commits for more details. Best reviewed commit-by-commit.

The connector's documentation should be updated to include that we can capture `staff_members` and include the `read_users` scope as required for that stream.

**Notes for reviewers:**

I haven't been able to get the `read_users` scope for my dev Shopify app, so I haven't been able to test out this stream with actual API calls to confirm everything works end-to-end. I'm fairly confident it should ✨ just work ✨ , but there may be some bumps to iron out once the stream gets used in production.

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -747,6 +747,28 @@ def open(
     return (response.Opened(explicitAcknowledgements=requires_explicit_acks), _run)
 
 
+def _tombstone_discriminator_values(
+    tombstone: BaseDocument, key: list[str]
+) -> tuple[object, ...]:
+    """Return a tombstone's collection-key values other than the CDK-managed
+    /_meta/row_id, in key order. These are what must differ between snapshot
+    subtasks: two subtasks sharing them would emit tombstones into the same
+    (discriminator, row_id) keyspace."""
+    dump = tombstone.model_dump(by_alias=True)
+    values: list[object] = []
+    for key_pointer in key:
+        # row_id is assigned per-pass by the CDK and is identical across
+        # subtasks, so it can't distinguish them.
+        if key_pointer == "/_meta/row_id":
+            continue
+        node: Any = dump
+        for field in key_pointer.strip("/").split("/"):
+            # A KeyError here means the tombstone is missing a key field.
+            node = node[field]
+        values.append(node)
+    return tuple(values)
+
+
 def open_binding(
     binding: CaptureBinding[_ResourceConfig],
     binding_index: int,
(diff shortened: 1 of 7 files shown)


nicolaslazo (approved):
LGTM

estuary-cdk/estuary_cdk/capture/common.py
```diff
@@ -870,24 +874,78 @@ async def backfill_closure(
...
+            )
+            # The CDK can set _meta/row_id but not the connector-specific discriminator, so
+            # each subtask must supply its own tombstone with that field pre-set.
+            assert (
+                isinstance(tombstone, dict)
+                and tombstone.keys() == fetch_snapshot.keys()
+            ), (
+                "fetch_snapshot subtasks requires a tombstone dict with "
+                "the same keys, each carrying the subtask discriminator. "
+            )
```
Alex-Bair, line 932: Looking at this with fresh eyes, it's possible for a connector author to accidently provide tombstones with the same discriminator values, which would cause the keyspaces for deletions to be the same (deletions from one subtask would actually show up as deletions for a different subtask). That'd be really confusing to troubleshoot, so I'm going to add another assertion here that verifies tombstones contain unique discriminator values.

PR #4648, opened 2026-06-10
materialize-boilerplate: set multiple constraints for projections

### LINKED ISSUE #4540: projections bug causing flow_document to not be read

mdibaiee:
**Description:**

- Fixes the bug where a flow_document projection may already exist and be incompatible, while at the same time being required
- Did a refactor of `validateMatchesExistingResource` -- it was too hard to follow specially after adding the new logic. It is a separate commit

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-boilerplate/validate.go
+++ b/materialize-boilerplate/validate.go
@@ -3,6 +3,7 @@ package boilerplate
 import (
 	"encoding/json"
 	"fmt"
+	"maps"
 	"slices"
 	"strconv"
 	"strings"
@@ -62,6 +63,10 @@ func NewValidator(
 }
 
 // ValidateBinding calculates the constraints for a new binding or a change to an existing binding.
+// Each field may have more than one constraint: notably, a required field whose existing
+// materialized column is incompatible with the proposed projection carries both the requiredness
+// constraint and the INCOMPATIBLE constraint, so that the control plane fails the build with a
+// clear error.
 func (v Validator) ValidateBinding(
 	path []string,
 	deltaUpdates bool,
@@ -69,7 +74,7 @@ func (v Validator) ValidateBinding(
 	boundCollection pf.CollectionSpec,
 	fieldConfigJsonMap map[string]json.RawMessage,
 	lastSpec *pf.MaterializationSpec,
-) (map[string]*pm.Response_Validated_Constraint, error) {
+) (map[string][]*pm.Response_Validated_Constraint, error) {
 	lastBinding := findLastBinding(path, lastSpec)
 	existingResource := v.is.GetResource(path)
 
@@ -105,13 +110,18 @@ func (v Validator) ValidateBinding(
 	}
 
 	var err error
-	var constraints map[string]*pm.Response_Validated_Constraint
+	var constraints map[string][]*pm.Response_Validated_Constraint
 	if existingResource == nil || (lastBinding != nil && backfill != lastBinding.Backfill) {
 		// Always validate as a new table if the existing table doesn't exist, since there is no
 		// existing table to be incompatible with. Also validate as a new table if we are going to
 		// be replacing the table.
-		if constraints, err = v.validateNewBinding(boundCollection, deltaUpdates, fieldConfigJsonMap); err != nil {
+		newConstraints, err := v.validateNewBinding(boundCollection, deltaUpdates, fieldConfigJsonMap)
+		if err != nil {
 			return nil, err
+		}
+		constraints = make(map[string][]*pm.Response_Validated_Constraint, len(newConstraints))
+		for field, c := range newConstraints {
+			constraints[field] = []*pm.Response_Validated_Constraint{c}
 		}
 	} else {
 		if lastBinding != nil && lastBinding.DeltaUpdates && !deltaUpdates {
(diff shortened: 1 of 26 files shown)


materialize-boilerplate/validate.go
```diff
@@ -183,122 +213,216 @@ func (v Validator) validateMatchesExistingResource(
...
+			len(cs) == 1 &&
+			cs[0].Type == pm.Response_Validated_Constraint_FIELD_OPTIONAL &&
 			slices.Contains(lastBinding.FieldSelection.AllFields(), p.Field) {
-			c = &pm.Response_Validated_Constraint{
+			cs = []*pm.Response_Validated_Constraint{{
 				Type:   pm.Response_Validated_Constraint_LOCATION_RECOMMENDED,
 				Reason: "This location is part of the current materialization",
-			}
+			}}
 		}
```
danielnelson: `LOCATION_RECOMMENDED` is documented as deprecated and equivalent to `FIELD_OPTIONAL`.  If they are equivalent then this shouldn't change anything, could we just remove this block?

materialize-boilerplate/validate.go
```diff
@@ -183,122 +213,216 @@ func (v Validator) validateMatchesExistingResource(
...
-				"The root document must be materialized as field '%s'",
-				lastBinding.FieldSelection.Document,
-			),
+		// with a constraint on a root document projection that it does have. The paired
+		// LOCATION_REQUIRED constraint guarantees the conflict is surfaced as a clean error.
+		constraints[docFields[0]] = []*pm.Response_Validated_Constraint{
+			{
+				Type:   pm.Response_Validated_Constraint_LOCATION_REQUIRED,
+				Reason: "The root document must be materialized for standard updates",
+			},
```
danielnelson, line 252: It seems like the actual required location is the document field from the last binding.  It probably doesn't exist in this binding, but it seems like it possibly could as a non-document field?  I suppose we can't set a constraint on a projection that doesn't exist?

materialize-boilerplate/validate.go
```diff
@@ -183,122 +213,216 @@ func (v Validator) validateMatchesExistingResource(
...
+	// requiredness when the base requires the projection, so the control plane
+	// surfaces a clear conflict; for a non-required projection it returns only
+	// the incompatibility, which the control plane drops.
+	withIncompatibility := func(incompatible *pm.Response_Validated_Constraint) []*pm.Response_Validated_Constraint {
+		if base.Type == pm.Response_Validated_Constraint_LOCATION_REQUIRED ||
+			base.Type == pm.Response_Validated_Constraint_FIELD_REQUIRED {
+			return []*pm.Response_Validated_Constraint{base, incompatible}
+		}
+		return []*pm.Response_Validated_Constraint{incompatible}
+	}
```
danielnelson: I would move this to a module level function since its being called from another method.

materialize-boilerplate/validate.go
```diff
@@ -183,122 +213,216 @@ func (v Validator) validateMatchesExistingResource(
...
-) (map[string]*pm.Response_Validated_Constraint, error) {
-	constraints := make(map[string]*pm.Response_Validated_Constraint)
+) (map[string][]*pm.Response_Validated_Constraint, error) {
+	// Root document projections, in projection order. The first one is the
+	// document when there is no prior selection to defer to.
+	var docFields []string
+	for _, p := range boundCollection.Projections {
+		if p.IsRootDocumentProjection() {
+			docFields = append(docFields, p.Field)
+		}
```
danielnelson, line 223: This would be a pre-existing issue if it were not true but are we guaranteed that `len(docFields) > 0` is valid when using standard updates?  Later on this is accessed in a few spots.  Since it's not enforced by the protobuf, perhaps we should add a check for this and return an error?

materialize-boilerplate/validate.go
```diff
@@ -183,122 +213,216 @@ func (v Validator) validateMatchesExistingResource(
...
+		}}, nil
+	}
+
+	// The key location was already part of the selection: validate it against
+	// its existing column, or recommend it if the destination doesn't report
+	// columns.
+	existingField := existingResource.GetField(p.Field)
+	if existingField == nil {
+		return []*pm.Response_Validated_Constraint{{
+			Type:   pm.Response_Validated_Constraint_LOCATION_RECOMMENDED,
```
danielnelson: Should we switch to `FIELD_OPTIONAL` because `LOCATION_RECOMMENDED` is deprecated?

danielnelson (approved):

materialize-boilerplate/validate.go
mdibaiee: Added check to this

materialize-boilerplate/validate.go
mdibaiee: Did a refactor to remove the deprecated `LOCATION_RECOMMENDED` everywhere 👍🏽

materialize-boilerplate/validate.go
mdibaiee: we actually can set a constraint on a project that doesn't exist (to say it must exist, for example)

PR #4649, opened 2026-06-10
source-shopify-native: add `staffMember.id` field to `orders` stream

### LINKED ISSUE #4227: source-shopify-native: include `staffMember` field in the `orders` stream
Add the [`staffMember` field](https://shopify.dev/docs/api/admin-graphql/latest/objects/Order#field-Order.fields.staffMember) to the `orders` stream.

I'm 95% sure access to the `staffMember` field on the `Order` object will require the `read_users` scope on a Shopify Plus / Advanced store, so we can't unconditionally include the `staffMember` field for all captures. We'll need to dynamically add it only when we'll be able to successfully query it.

Alex-Bair:
**Regression?**

No

**Description:**

Reading `staffMember` requires a Shopify Plus/Advanced store or a finance embedded app _plus_ the `read_users` scope, so we only include the `staffMember.id` field in `orders` queries when those conditions are met. Only the `id` field of `staffMember` is captured; to get other `staffMember` related details, users can join on the soon-to-be-added `staff_members` stream.

**Notes for reviewers:**

Similar to https://github.com/estuary/connectors/pull/4647, I couldn't get the `read_users` scope for my dev Shopify account, so I haven't tested this change out with actual API queries. But it should work since it aligns with [Shopify's documentation](https://shopify.dev/docs/api/admin-graphql/2026-01/objects/Order#field-Order.fields.staffMember) on how to query this field.

Also, the `read_users` scope isn't added to the list of scopes in the connector in this PR because https://github.com/estuary/connectors/pull/4647 is adding it.

### DIFF
--- a/source-shopify-native/source_shopify_native/graphql/orders/orders.py
+++ b/source-shopify-native/source_shopify_native/graphql/orders/orders.py
@@ -23,6 +23,7 @@ class Orders(ShopifyGraphQLResource):
         id
         name
     }
+    # {{ staffMember }}
     billingAddress {
         id
         address1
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
LGTM

PR #4650, opened 2026-06-12
source-shopify-native: resolve nested connections in non-bulk queries

### LINKED ISSUE #4225: source-shopify-native: add the `refundLineItems` to the `order_refunds` stream
Add the [`refundLineItems` field](https://shopify.dev/docs/api/admin-graphql/latest/objects/Refund#field-Refund.fields.refundLineItems) to the `order_refunds` stream. Fields on `refundLineItem` should include:
- id
- location
- priceSet
- quantity
- restocked
- restockType
- subtotalSet
- totalTaxSet
- lineItem (include all non-deprecated subfields that are not connections)

### LINKED ISSUE #4339: source-shopify-native: add `fulfillmentLineItems` to the `fulfillments` stream
## Problem
The `fulfillments` stream in `source-shopify-native` doesn't capture [`fulfillmentLineItems`](https://shopify.dev/docs/api/admin-graphql/latest/objects/Fulfillment#field-Fulfillment.fields.fulfillmentLineItems). Add this nested field so downstream consumers can tie fulfillments to the specific order line items they shipped.

Minimum field set: `id`, `quantity`, `discountedTotalSet`, `originalTotalSet`, `lineItem.id` (final shape TBD).

## Possible Solution
          
  `fulfillmentLineItems` is a GraphQL connection, and the connector's non-bulk path forbids nested connections. So the `fulfillments` stream will be moved to bulk queries:                                                      
  
  1. Drop `SHOULD_USE_BULK_QUERIES = False` in `graphql/orders/fulfillments.py`.                                                                                                                                                
  2. Add the `fulfillmentLineItems { edges { node { ... } } }` selection.
  3. Rewrite `process_result` for bulk JSONL. Verify behavior against a real bulk job.                                                                          
  4. Add a state migration. Flipping to bulk queries collapses the state from {backfill, incremental} to {incremental}.                

## Reference
[Originating Slack thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1777553486396779?thread_ts=1777473909.054049&cid=C03Q2NRFKDL)

Alex-Bair:
**Regression?**

No

**Description:**

Some Shopify fields are GraphQL connections nested under a list. For example, a refund's `refundLineItems` lives under `Order.refunds`, and a fulfillment's `fulfillmentLineItems` under `Order.fulfillments`. Bulk operations reject a connection nested inside a list, and the connector's non-bulk path queries didn't support nested connections (mainly because paginating through nested connections is complex & we hadn't needed to tackle that complexity since bulk queries are likely more efficient when queries contain connections), so these fields couldn't be captured with the existing functionality.

This PR addresses this by adding a `NestedConnection` resolver for non-bulk queries. The strategy for resolving a `NestedConnection` is:
1. Inline the first page into the outer query. In the common case where a parent has few children, no extra HTTP requests are made to paginate through the connection.
2. Drain the remaining records only when the first page reports there are more records via the `pageInfo`. The remaining records are fetched via `node(parent_id:)` queries.
3. Flatten the result into a list under the field name, dropping the `edges`/`pageInfo` wrapper that's only used to paginate through a connection.

The new `NestedConnection` resolver is used for adding the `refundLineItems` field to the `order_refunds` stream & the `fulfillmentLineItems` field to the `fulfillments` stream.

Capture snapshot changes are expected. I did a lot of modifications & tweaking in the dev Shopify account we use for snapshot tests, so the baseline changed quite a bit in the first commit. The remaining snapshot changes are expected due to the addition of the new fields.

**Notes for reviewers:**

Tested with `flowctl preview`. Confirmed the `NestedConnection` is drained by artificially setting the page size to `1` in order to for the additional records to be fetched via follow up queries.

I put a little bit of effort into making `NestedConnection` & the resolver logic somewhat general purpose, although they're still pretty specific to Shopify's GraphQL API due to the specific pagination fields. If we need to resolve connections for other GraphQL APIs in the future, we _may_ be able to polish the `NestedConnection` infrastructure in this connector so it's fit for use in any connector hitting a GraphQL API then pull it into the CDK.

### DIFF
--- a/source-shopify-native/source_shopify_native/api.py
+++ b/source-shopify-native/source_shopify_native/api.py
@@ -1,7 +1,8 @@
+import asyncio
 import time
 from datetime import UTC, datetime, timedelta
 from logging import Logger
-from typing import AsyncGenerator
+from typing import Any, AsyncGenerator
 
 from estuary_cdk.capture.common import (
     LogCursor,
@@ -19,6 +20,7 @@ from source_shopify_native.models import (
 )
 
 from .graphql.client import ShopifyGraphQLClient
+from .graphql.nested_resolver import NESTED_CONNECTION_CONCURRENCY, resolve_document
 from .utils import dt_to_str, str_to_dt
 
 CHECKPOINT_INTERVAL = 1_000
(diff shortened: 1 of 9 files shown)


nicolaslazo (approved):
👍

PR #4652, opened 2026-06-12
source-impact-native: avoid eventually consistent results by lagging incremental replication behind by an hour

Alex-Bair:
**Regression?**

No.

**Description:**

We've observed Impact's API be eventually consistent, meaning a record can become queryable via the API after its timestamp cursor. When that happens, the connector has already seen records with later timestamps & progressed its cursor past the record affected by eventual consistency, so that record gets missed.

This PR adds an `INCREMENTAL_LAG` to hopefully combat the Impact API's eventual consistency. All incremental `fetch_changes` functions ignore any records with cursor timestamps more recent than `now - INCREMENTAL_LAG`. Ideally, this gives the Impact API enough time to return consistent results over the time range we care about & keeps the connector out of eventual consistency territory.

We may need to update this connector to use a real-time and delayed stream strategy if a 1 hour lag is unacceptable for users, but that'll require some more work. We'd ideally polish up the connector at the same time so it uses the same patterns & conventions we've been using in more recent connectors too.

**Notes for reviewers:**

Pylance was flagging a lot of errors due to incorrect / missing type hints, so I fixed a lot of those complaints in the first commit. There are still a handful of type related complaints, but addressing those will require more involved refactoring of classes in `models.py` than I wanted to tackle right now.

Best reviewed commit-by-commit.

### DIFF
--- a/source-impact-native/source_impact_native/api.py
+++ b/source-impact-native/source_impact_native/api.py
@@ -15,6 +15,15 @@ from source_impact_native.models import ActionInquiries, Actions
 API = "https://api.impact.com"
 API_CATALOG = "Advertisers"
 
+# Impact's endpoints are eventually consistent.A record can become
+# queryable minutes after its timestamp, after the connector has
+# already seen records with later timestamps. Since the incremental
+# cursor is a high-water mark, anything that surfaces below it would
+# be filtered out and lost forever. To avoid that, we never advance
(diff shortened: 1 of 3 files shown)


nicolaslazo (approved):
Looks good! I didn't know Python supported chained comparison operators.

PR #4661, opened 2026-06-15
materialize-hubspot: add initial connector

### LINKED ISSUE #3978: Connector to materialize to hubspot
### System Name

Hubspot

### Type

Materialize

### Details

We would love the capability to push data from Estuary to Hubspot via a materialization.  There are a lot of UX concerns here but ideally, a materialization to a SaaS would allow for:

1. Pushing to any main Hubspot stream
2. Helping the user understand the schema requirements.  They won't be pushing every field, but should be able to search for fields and map them to equivalents in their collection.
3. Control the latency where required.  In general we should try to push data as quickly as we can unless there's a cost associated.

danielnelson:
**Regression?**

No

**Description:**

Add a materialization connector for HubSpot.

This connector always uses delta updates to do partial updates with last-write-wins to HubSpot records.  We decided to never create properties automatically, so you will need to setup any custom properties ahead of time in the webapp.  Also worth noting is that you must select a single value as the key.

Because you cannot determine the `id` of a record in HubSpot, the key field is used to match to the record instead.  A property (roughly a column) can be unique or non-unique, and this determines the set of APIs that the connector will use.  A unique property is heavily preferred, since non-unique is more expensive and its unfortunately not guaranteed to not create duplicate records due to a race condition between creating a record and it appearing in the search API.  Still it is allowed for convenience and by far the majority of pre-existing records are non-unique.

https://github.com/estuary/flow/pull/3042

closes: #3978


materialize-hubspot/config.go
```diff
@@ -0,0 +1,208 @@
...
+	ClientSecret         Secret    `json:"client_secret" jsonschema_extras:"secret=true"`
+	RefreshToken         Secret    `json:"refresh_token" jsonschema_extras:"secret=true"`
+	AccessToken          Secret    `json:"access_token" jsonschema_extras:"secret=true"`
+	AccessTokenExpiresAt time.Time `json:"access_token_expires_at"`
+}
+
+// ServiceKeyCredentials allows authentication with a service key.  You would
+// only want to use this method for debugging and development.
+type ServiceKeyCredentials struct {
+	ServiceKey Secret `json:"service_key" jsonschema:"title=Service Key,description=HubSpot Service Key with required scope grants.", jsonschema_extras:"secret=true"`
```
mdibaiee:
```suggestion
	ServiceKey Secret `json:"service_key" jsonschema:"title=Service Key,description=HubSpot Service Key with required scope grants." jsonschema_extras:"secret=true"`
```

materialize-hubspot/config.go
```diff
@@ -0,0 +1,208 @@
...
+				},
+			},
+			orderedmap.Pair[string, *jsonschema.Schema]{
+				Key: "delta_updates",
+				Value: &jsonschema.Schema{
+					Type:        "boolean",
+					Title:       "Delta Update",
+					Description: "Should updates to this table be done via delta updates. This connector always uses delta updates.",
+					ReadOnly:    true,
+					Default:     true,
```
mdibaiee, line 171: I don't know if our UI recognises `ReadOnly: true`, but I think it does recognise `Const: true` which makes the value a constant.

materialize-hubspot/driver.go
```diff
@@ -0,0 +1,415 @@
...
+	schemagen "github.com/estuary/connectors/go/schema-gen"
+	boilerplate "github.com/estuary/connectors/materialize-boilerplate"
+	pf "github.com/estuary/flow/go/protocols/flow"
+	pm "github.com/estuary/flow/go/protocols/materialize"
+	log "github.com/sirupsen/logrus"
+	"go.gazette.dev/core/consumer/protocol"
+)
+
+var (
+	flowPropertyGroupName = "flow"
```
mdibaiee: `flow` seems like a very generic name, can we make this `estuary_flow` or something more specific?

materialize-hubspot/driver.go
```diff
@@ -0,0 +1,415 @@
...
+
+				actions = append(actions, fmt.Sprintf("Create Property %q", field.Property.Name))
+
+				err := materializer.CreateProperty(ctx, b.object, field.Property)
+				if err != nil {
+					return nil, fmt.Errorf("unable to create property for field %q: %w", field.Property.Name, err)
+				}
+			}
+		}
+	}
```
mdibaiee, line 159: Are these unused since we forbid any constraints for any projections?

materialize-hubspot/objects.go
```diff
@@ -0,0 +1,103 @@
...
+	DealsObject      CRMObject = "deals"
+	EmailsObject     CRMObject = "emails"
+	LineItemsObject  CRMObject = "line_item"
+	MeetingsObject   CRMObject = "meetings"
+	PostalMailObject CRMObject = "postal_mail"
+	ProductsObject   CRMObject = "products"
+	QuotesObject     CRMObject = "quotes"
+	TasksObject      CRMObject = "tasks"
+	TicketsObject    CRMObject = "tickets"
+	UnknownObject    CRMObject = "unknown"
```
mdibaiee, line 23: calls, emails, meetings, postal_mail, tasks are missing from our OAuth scopes, is that intentional?

materialize-hubspot/objects.go
```diff
@@ -0,0 +1,103 @@
...
+
+type CRMObject string
+
+const (
+	CallsObject      CRMObject = "calls"
+	CompaniesObject  CRMObject = "companies"
+	ContactsObject   CRMObject = "contacts"
+	DealsObject      CRMObject = "deals"
+	EmailsObject     CRMObject = "emails"
+	LineItemsObject  CRMObject = "line_item"
```
mdibaiee:
```suggestion
	LineItemsObject  CRMObject = "line_items"
```

materialize-hubspot/transactor.go
```diff
@@ -0,0 +1,340 @@
...
+					if len(items) != 0 {
+						if !yield(&Batch{BindingIdx: activeBinding, Items: items}, nil) {
+							return
+						}
+						items = make([]BatchItem, 0, MaxBatchRecords)
+					}
+				}
+			}
+
+			changedKey := activeKey != key
```
mdibaiee: I'm not sure if in delta updates keys arrive ordered (e.g. it may be possible in delta updates to receive two documents with the same "key" apart from each other in the transaction, not necessarily contiguous) -- worth checking with data plane team

materialize-hubspot/type_mapping.go
```diff
@@ -0,0 +1,256 @@
...
+	name = startLowerLetter.ReplaceAllString(name, "")
+
+	if len(name) > MaxPropertyNameLength {
+		name = name[:MaxPropertyNameLength]
+	}
+
+	if name == "" {
+		return "", fmt.Errorf("unable to convert to a valid property name: %q", fieldName)
+	}
+	return name, nil
```
mdibaiee, line 264: I think this means that if the user creates a property called `myName`, and they have a field called `my-name`, and another `myName` they will both match, right? maybe we need a validator to avoid multiple matches

mdibaiee (commented):
Thanks! A first review below

I think we really need to have a similar coverage of tests as with normal connectors, e.g. TestSpec, TestIntegration, etc. to allow us to develop the connector with more confidence

materialize-hubspot/objects.go
danielnelson: Yeah, the scopes aren't quite consistent.

materialize-hubspot/type_mapping.go
danielnelson: This is basically correct though this particular example maps to slightly different fields: `my-name -> my_name` and `myName -> myname`.  Slightly changing your example to the fields `my-name` and `my_name` and they both map to `my_name`.

There are other restrictions on the name, it is a little confusing in their webapp because there you enter the "label" and the "name" is generated.   I notice that I don't use the exact same mapping as their webapp, maybe I should, here are some examples from their webapp:
- `myName -> myname`
- `my-name -> myname`
- `2myname -> n2myname`
- `my/name -> myname`
- `Straße -> strae`
- `שלום -> slwm`

During validate I set `FoldedField` on the constraint and if there is a conflict the runtime handles it by selecting only one.  You can change which one is selected but you can't have both.
2026-06-22-161753_938x110_scrot

materialize-hubspot/config.go
danielnelson: This seemed to work but I'll check out `const` and see if it behaves differently.

materialize-hubspot/driver.go
danielnelson: I'm not sure I'm understanding your comment correctly, but this is dead code.  We decided that we wouldn't create properties dynamically later on in the process and I left it in.  We don't validate fields without a corresponding property and we no longer request the scopes.  In the case that a property is removed between validate and apply it might be activated.

Normally I would take this out, and I could remove it from client.go as well, but I think if we had a few changes in the UI then we might want it back.  Now it feels like I should remove it though, agreed?

materialize-hubspot/driver.go
danielnelson: That would make sense, though if I remove the property creation code fully this can go.

materialize-hubspot/transactor.go
danielnelson: I have talked to @jgraettinger and the StoreIterator is always sorted by (binding, key).

materialize-hubspot/config.go
danielnelson: With `readOnly` the control is disabled, so you can't turn it off.  The idea with this was just to make it more clear that it was using delta updates under the hood, since it does have some implications such as no hard deletions.

On the other hand `const` doesn't seem to do anything for a bool, I think perhaps it sets the initial value?

materialize-hubspot/driver.go
mdibaiee: @danielnelson yeah probably we can remove it as a separate commit so that it stays in git history for quick revival once we want to support it

materialize-hubspot/config.go
mdibaiee: interesting, good to know `const` doesn't seem to work for bool

materialize-hubspot/type_mapping.go
danielnelson: I make the translation more similar, handling the first 5 examples above.

danielnelson:
All comments should be addressed, ready for another round of reviews.

materialize-hubspot/transactor.go
```diff
@@ -0,0 +1,353 @@
...
+
+	// If we received more results than we have ids, there are duplicate
+	// records with this idProperty.  A user will need to go in and merge the
+	// records to fix this, or better yet switch to a unique property.
+	if len(response.Results) != len(ids) {
+		log.WithFields(log.Fields{
+			"object":      b.object,
+			"id_property": b.idProperty.Name,
+		}).Warn("duplicate records matched")
+	}
```
mdibaiee, line 141: Is it possible that `len(response.Results) < len(ids)` meaning there are missed records? the comment seems to want to check for `len(response.Results) > len(ids)` but we are using `!=`

materialize-hubspot/validator.go
```diff
@@ -0,0 +1,179 @@
...
+	if !mappedField.Compatible(property) {
+		return &pm.Response_Validated_Constraint{
+			Type:        pm.Response_Validated_Constraint_FIELD_FORBIDDEN,
+			Reason:      fmt.Sprintf("The existing property for this field has an incompatible type: %q", property.Type),
+			FoldedField: name,
+		}, nil
+	}
+
+	switch {
+	case projection.IsPrimaryKey:
```
mdibaiee, line 144: Should we check that the key is not `NumberPropertyType` (i.e. float64)?

materialize-hubspot/type_mapping.go
```diff
@@ -0,0 +1,264 @@
...
+type MappedField struct {
+	Name string
+	Ptr  string
+
+	// The Property we would create if one does not exist.
+	Property *Property
+}
+
+func NewMappedField(projection pf.Projection, fc FieldConfig) (*MappedField, error) {
+	types := projection.Inference.Types
```
mdibaiee: nit: it may be a better idea to clone this slice instead of modifying in place, although there is no bug with the way it is today, it could be a pitfall in the future

materialize-hubspot/config.go
```diff
@@ -0,0 +1,218 @@
...
+		schemagen.OneOfSubSchema("Service Key", ServiceKeyCredentials{}, string(ServiceKeyAuthType)),
+	}
+	return schemagen.OneOfSchema("Authentication", "", "auth_type", string(OAuth2AuthType), subSchemas...)
+}
+
+type AdvancedConfig struct {
+	Limit       float64 `json:"limit" jsonschema:"title=Request Limit,description=Maximum requests per second across all bindings excluding search.,default=10,exclusiveMinimum=0,maximum=250"`
+	Burst       int     `json:"burst" jsonschema:"title=Request Burst Count,description=Burst requests across all bindings excluding search.,default=100,minimum=1,maximum=1000"`
+	SearchLimit float64 `json:"search_limit" jsonschema:"title=Search Request Limit,description=Maximum search requests per second across all bindings.,default=5,exclusiveMinimum=0,maximum=250"`
+	SearchBurst int     `json:"search_burst" jsonschema:"title=Search Request Burst Count,description=Burst search requests across all bindings.,default=5,minimum=1,maximum=1000"`
```
mdibaiee: we probably want `omitempty` for these

materialize-hubspot/transactor.go
danielnelson: Yes, thank you.

materialize-hubspot/validator.go
danielnelson: I don't want to disallow `NumberPropertyType`, because this type is integers and floats, and I want to support integer primary keys.  I don't think it is possible for a primary key to be anything other than integer, string, bool; this came up recently.  We aren't getting any help in the type system here, I would love to think of a way we can.

I added a check here disallowing numeric, let me know if you think I should leave it.

Here is a link about [HubSpot types](https://developers.hubspot.com/docs/api-reference/latest/crm/properties/guide#types).

mdibaiee (approved):

PR #4664, opened 2026-06-16
source-sqlserver-ct: Detect replicas and incoherent CT metadata

willdonnelly:
**Description:**

Change Tracking must be captured from the primary writable database. I was under the impression that our existing prerequisites ought to error out on a replica but apparently they don't, and actually trying to capture from a replica leads to pain as the version metadata is incoherent in a way which resulted in repeatedly backfilling all tables over and over.

Adds a prerequisite check that fails when DATABASEPROPERTYEX reports the database is read-only, so the misconfiguration is reported to the user up front.

Also adds an internal sanity check in pollChanges that errors out if the current version is below any table's min valid version, so that any future cases of incoherent replicas not caught by the prerequisite check might fail cleanly instead of repeatedly backfilling.

We need to update the SQL Server CT documentation to note the primary-DB requirement but that isn't really affected by this, that was always the requirement.

### DIFF
--- a/source-sqlserver-ct/prerequisites.go
+++ b/source-sqlserver-ct/prerequisites.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"strings"
 
 	"github.com/estuary/connectors/go/capture/sqlserver/backfill"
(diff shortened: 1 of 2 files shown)


Alex-Bair (approved):
LGTM

PR #4654, opened 2026-06-15
source-shopify-native: add `disputes` and `order_returns` streams

Alex-Bair:
**Regression?**

No

**Description:**

This PR's scope includes:
- Updating the capture snapshot with a new baseline because I changed the data in our dev Shopify account used in tests during development of the other features in this PR.
- Preparing for the addition of some new streams. These preparations include:
  - Switching to use `ResourceConfigWithSchedule` so periodic backfills can be scheduled for bindings.
  - Centralizing the sanity check that the API always returns a page cursor when it says there's more pages.
  - Adding a `backfill_incremental_unsorted` function for unsorted incremental streams, like `locations`, `inventory_items`, and `disputes`. This makes it easier for unsorted incremental streams to complete their backfills when they contain lots of records.
- Adding two new streams, including:
  - `disputes`: an unsorted, incremental stream that uses scheduled backfills to capture updates
  - `order_returns`: a sorted, incremental stream that uses bulk queries

Snapshot changes are all expected either due to the re-baselining after changing data in our dev Shopify account, the switch to use `ResourceConfigWithSchedule`, or the addition of the two new streams.

**Documentation links affected:**

The connector's documentation should be updated to reflect:
- The addition of the new `schedule` field for resource configs.
- The two new streams, `disputes` and `order_returns`
- The scopes required to capture data for those two new streams.

**Notes for reviewers:**

I was able to test out that `order_returns` works, and you can see evidence of that in the `order_returns` document included in the capture snapshot. I was unable to create any `disputes` documents in our dev Shopify account, so I was unable to verify this stream works with actual responses from the API. After merging this PR, I plan to keep a close eye on any captures that discover & use the `disputes` stream to ensure it works as anticipated.

Best reviewed commit-by-commit.

PR #4668, opened 2026-06-16
source-shopify-native: make `disputeEvidence` a `ConditionalField`

Alex-Bair:
**Regression?**

No

**Description:**

This PR is a fast follow up to https://github.com/estuary/connectors/pull/4654 that introduced the `disputes` stream.

Turns out that `disputeEvidence` can only be queried when the `read_shopify_payments_dispute_evidences` scope is granted. This means this field need to be a `ConditionalField` that's only included in queries when that specific scope is granted.


nicolaslazo (approved):
LGTM

PR #4673, opened 2026-06-16
source-shopify-native: handle `UnverifiedReturnLineItem` in `order_returns` JSONL results

Alex-Bair:
**Regression?**

No

**Description:**

This PR is another fast-follow up to https://github.com/estuary/connectors/pull/4654 that introduced the `order_returns` stream.

The `returnLineItems` connection on a `Return` is typed as the [`ReturnLineItemType` interface](https://shopify.dev/docs/api/admin-graphql/latest/interfaces/ReturnLineItemType), which has two concrete implementations: `ReturnLineItem` and `UnverifiedReturnLineItem`. In a bulk operation, each node's id reflects its concrete type, so unverified line items arrive as `gid://shopify/UnverifiedReturnLineItem/...` rather than `gid://shopify/ReturnLineItem/....`

The JSONL parser only matched the `ReturnLineItem` prefix, so unverified line items fell through to the catch-all branch and raised "Unidentified line in JSONL response." errors & crashed the connector. This PR fixes that by handling `UnverifiedReturnLineItem`s like `ReturnLineItem`s, routing both concrete types into the same branch so they're appended to their parent return's `returnLineItems` list.

### DIFF
--- a/source-shopify-native/source_shopify_native/graphql/orders/returns.py
+++ b/source-shopify-native/source_shopify_native/graphql/orders/returns.py
@@ -129,7 +129,10 @@ class OrderReturns(ShopifyGraphQLResource):
                 current_order = record
                 current_order[RETURNS_KEY] = []
 
-            elif "gid://shopify/ReturnLineItem/" in id:
+            elif (
+                "gid://shopify/ReturnLineItem/" in id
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

PR #4676, opened 2026-06-17
source-stripe-native: implement API version pinning

nicolaslazo:
**Description:**

The status=all filter in Subscriptions / SubscriptionItems inadvertently set a hard requirement on how old an API version we can use. This commit adds an API version endpoint config field, falling back to the account's default to preserve the original behavior.

`BaseStripeObject` children can now define a minimum required API version. If any streams fall below the floor they'll be disabled, rather than removed, to handle when an existing capture gets downgraded.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-stripe-native/source_stripe_native/resources.py
+++ b/source-stripe-native/source_stripe_native/resources.py
@@ -17,11 +17,14 @@ from estuary_cdk.capture.document import BaseDocument
 from estuary_cdk.flow import CaptureBinding
 from estuary_cdk.http import HTTPError, HTTPMixin, HTTPSession, TokenSource
 
+from source_stripe_native.http import StripeHTTPMixin
+
 from .protocols import FetchChangesFnFactory, FetchPageFnFactory
 
 from .account_fetcher import fetch_connected_account_ids
 from .api import (
     API,
+    fetch_api_version,
     fetch_backfill,
     fetch_backfill_substreams,
     fetch_backfill_list_items,
@@ -42,6 +45,7 @@ from .models import (
     STREAMS,
     Accounts,
     BaseStripeChildObject,
+    BaseStripeObject,
     BaseStripeObjectNoEvents,
     BaseStripeObjectWithEvents,
     ConnectorState,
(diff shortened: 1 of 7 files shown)


source-stripe-native/source_stripe_native/resources.py
```diff
@@ -331,9 +340,33 @@ async def all_resources(
...
+
+        minimum = model.MINIMUM_API_VERSION
+        if minimum and effective_api_version < minimum:
+            resource.disable = True
+            log.warning(
+                (
+                    f"Disabling stream '{resource.name}': requires Stripe API version "
+                    f"{minimum}+, but the effective version is {effective_api_version}."
+                )
+            )
```
Alex-Bair, line 370: nit: the `resource.disable` field only really matters the first time a resource is discovered. For any captures that have already discovered the resource, setting `resource.disable` doesn't flip the binding from enabled -> disabled. So, scenarios like the following are possible:
- existing captures continue attempting to capture data for the resource despite the minimum API version not being met.
- users re-enable an initially disabled resources, and the capture fails on start up since the resource won't work with the default API version.

IMO, it's a better UX to omit resources from discovery when we know they won't work instead of disabling them so we can avoid those scenarios above.

Alex-Bair (approved):
LGTM % an optional nit + the spec snapshot needs updated.

source-stripe-native/source_stripe_native/resources.py
nicolaslazo: I agree, that's how we usually handle this sort of scenarios, but I'm worried about downgrades. What happens if a user goes back to an old API version on an existing capture and a resource no longer exists?

Alex-Bair: The resource would no longer be discovered, so we'd stop capturing data for it. The collection would still exist though with all documents captured up until that point, and the resource state would still exist in the connector state so we theoretically could pick back up where we left off if they upgrade to a supported API version later.

PR #4677, opened 2026-06-17
source-sqlserver{,-ct}: Log backfill queries

willdonnelly:
**Description:**

Adds logic to log the backfill query for a table the first time it is scanned during a connector invocation.

In other connectors the analogous logic runs an EXPLAIN, and I was going to do the same thing here with `SET SHOWPLAN`, but this was such an absolute pain that I gave up. Turns out that query parameters via go-mssqldb don't play nicely with SHOWPLAN. Also SHOWPLAN requires a special permission which we likely wouldn't have anyway, so it just wasn't worth implementing a workaround at this time.


Alex-Bair (approved):
LGTM

PR #4678, opened 2026-06-17
materialize-elasticsearch: retry transient per-item bulk store errors

jacobmarble:
**Description:**

Makes `materialize-elasticsearch` ride out transient destination-cluster blips instead of crashing the shard. When an index's primary shard is briefly unavailable, Elasticsearch returns a retryable per-item error (e.g. `retry_on_primary_exception`) inside an HTTP 200 bulk response. Previously the transactor treated the first item error as fatal and failed the shard; now it re-sends just the affected items with bounded exponential backoff.

Re-sending only the failed items is safe for delta bindings too — items that succeeded (including generated-ID creates) are never re-sent, so no duplicate rows.

**Workflow steps:**

No config/API changes; behavior-only. Tasks stop crash-looping on transient `shard is not in primary mode` errors.

**Notes for reviewers:**

- Scope: only the 200-with-item-errors path. True HTTP 429/5xx are still handled by the existing go-elasticsearch transport retry (`driver.go` `RetryOnStatus`) — not duplicated.
- `firstError()` is replaced by `bulkResponse.classify()` (terminal error → fail; retryable indices → resend); store batches are now `[][]byte` segments so a subset can be re-sent.
- The separate `create`→`index` idempotency change is **split out to a follow-up PR** per review discussion — it would fix the `version_conflict` failures but also removes a consistency tripwire, so it's worth deciding on its own.
- Verified: `go vet` + unit suite, and `TestIntegration` against ES 8.9.0 and OpenSearch 2.13.0 (snapshots unchanged). The retryable-item-error path can't be reproduced in the docker ES, so it's unit-tested via an injected sender.
- Heads-up: an unrelated `.github/workflows/ci.yaml` duckdb chore rode in on this branch.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -308,7 +308,7 @@ jobs:
 
       - name: Install duckdb
         if: matrix.connector == 'materialize-motherduck' || matrix.connector == 'materialize-s3-iceberg'
-        uses: opt-nc/setup-duckdb-action@v1.0.8
+        uses: opt-nc/setup-duckdb-action@v1.2.9
         with:
           version: v1.3.1
 
--- a/materialize-elasticsearch/transactor.go
+++ b/materialize-elasticsearch/transactor.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/elastic/go-elasticsearch/v8/esapi"
 	"github.com/elastic/go-elasticsearch/v8/esutil"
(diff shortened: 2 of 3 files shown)


mdibaiee:
@jacobmarble is this understanding correct that the second part of your PR is enough to fix the issue we saw for the customer?

> When a 200 bulk response carries only retryable per-item errors (e.g. retry_on_primary_exception from a briefly non-primary shard), re-send just those items with bounded backoff, rather than surfacing the first item error as fatal.

I am worried about the first part, because I think it _hides_ issues with missing/broken Loads. What do you think about landing the re-sending of bounded backoff first, and then we can try out the other part as a separate PR? I just think it would be helpful to confirm the issue is what we think it is by bounding the retry, and then revisit the create-upsert part

jacobmarble:
> @jacobmarble is this understanding correct that the second part of your PR is enough to fix the issue we saw for the customer?
> 
> > When a 200 bulk response carries only retryable per-item errors (e.g. retry_on_primary_exception from a briefly non-primary shard), re-send just those items with bounded backoff, rather than surfacing the first item error as fatal.
> 
> I am worried about the first part, because I think it _hides_ issues with missing/broken Loads. What do you think about landing the re-sending of bounded backoff first, and then we can try out the other part as a separate PR? I just think it would be helpful to confirm the issue is what we think it is by bounding the retry, and then revisit the create-upsert part

I agree, let's just do the second part for now. The explicit insert acts like a deliberate tripwire - the expected state doesn't match, don't ignore.

mdibaiee (approved):

PR #4682, opened 2026-06-18
source-salesforce-native: accept ISO 8601 duration time windows

nicolaslazo:
**Description:**

Adds support for ISO 8601 duration time windows. Bare integers are interpreted as days to preserve the existing behaviour.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/source-salesforce-native/source_salesforce_native/models.py
+++ b/source-salesforce-native/source_salesforce_native/models.py
@@ -1,10 +1,11 @@
-from datetime import datetime, UTC
+from datetime import datetime, timedelta, UTC
 from enum import StrEnum
-from typing import Any, Iterator, Annotated, ClassVar
+from typing import Any, Iterator, ClassVar, Literal
 
 from pydantic import (
     AwareDatetime,
     BaseModel,
+    ConfigDict,
     Field,
     ValidationInfo,
     create_model,
@@ -28,6 +29,7 @@ EARLIEST_VALID_DATE_IN_SALESFORCE = datetime(1700, 1, 1, tzinfo=UTC)
 # Salesforce was founded on 03FEB1999. As long as cursor values aren't backdated before this date (which only seems possible
 # for a select few of them), Salesforce's founding date should be a good start date for most users.
 SALESFORCE_FOUNDING_DATE = datetime(1999, 2, 3, tzinfo=UTC)
+MIN_INCREMENTAL_WINDOW_SIZE = timedelta(minutes=1)
 
 
 class SalesforceResourceConfigWithSchedule(ResourceConfigWithSchedule):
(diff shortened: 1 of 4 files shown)


source-salesforce-native/tests/snapshots/snapshots__spec__capture.stdout.json
```diff
@@ -6,11 +6,12 @@
...
-              "default": 18250,
-              "description": "Date window size for Bulk API 2.0 queries (in days). Typically left as the default unless Estuary Support or the connector logs indicate otherwise.",
-              "exclusiveMinimum": 0,
+              "default": "P50Y",
+              "description": "Date window size for Bulk API 2.0 queries. A bare integer is interpreted as a number of days; otherwise an ISO 8601 duration (e.g. PT1H) is accepted. Typically left as the default unless Estuary Support or the connector logs indicate otherwise.",
+              "format": "duration",
+              "gt": "PT1M",
               "title": "Window size",
-              "type": "integer"
+              "type": "string"
```
Alex-Bair: To preserve backwards compatibility with existing specs, we should only widen the `window_size` field's type in the endpoint spec's schema. If we change it in a backwards incompatible way, there are workflows in the UI that break, like when trying to edit a config that had an integer value for `window_size` but the schema says it should be a string.

For example, the UI threw an error when I did the following:
- Created a `source-salesforce-native` capture without the changes in this PR. `window_size` was set to an integer.
- Re-built `source-saleforce-native` with the changes in this PR & re-queued the `connector_tags` job to refresh the value of the `endpoint_spec_schema`.
- Navigated to the edit page of the existing capture. The input field for `window_size` showed this error:

Image
and attempting to edit the field triggered this error:

Image

I _think_ something like this would work so the JSON schema allows both integers and string durations but inside the connector we'd treat it like a `timedelta` regardless, but I haven't tested this out to confirm:
```python
from pydantic import GetJsonSchemaHandler
from pydantic.json_schema import JsonSchemaValue
from pydantic_core import CoreSchema

class _WindowSizeJsonSchema:
    @classmethod
    def __get_pydantic_json_schema__(
        cls, core_schema: CoreSchema, handler: GetJsonSchemaHandler
    ) -> JsonSchemaValue:
        schema = handler(core_schema)
        # Modify rather than replace
        return {"oneOf": [{"type": "integer"}, schema]}

DurationField = Annotated[
    Union[str, timedelta],
    BeforeValidator(_window_size_days_or_timedelta),
    _WindowSizeJsonSchema,
]

```

source-salesforce-native/source_salesforce_native/models.py
```diff
@@ -43,6 +45,18 @@ def default_start_date():
     return SALESFORCE_FOUNDING_DATE
 
 
+def _window_size_days_or_timedelta(value: Any) -> Any:
```
Alex-Bair: nit: Are we able to tighten these type hints, or are our hands tied because Pydantic requires the use of `Any` type hints here? It'd be nice if the function signature was:
```python
def _window_size_days_or_timedelta(value: int | timedelta) -> timedelta:

```

Alex-Bair (commented):
Looks good & pretty close to what I expected. The only change needed before merging is still allowing integer fields in the `window_size`'s JSON schema to avoid an error in the UI when editing an existing capture.

source-salesforce-native/source_salesforce_native/models.py
```diff
@@ -86,20 +126,31 @@ def _require_my_domain_for_client_credentials(self) -> "EndpointConfig":
         return self
 
     class Advanced(BaseModel):
-        window_size: Annotated[int, Field(
-            description="Date window size for Bulk API 2.0 queries (in days). Typically left as the default unless Estuary Support or the connector logs indicate otherwise.",
+        window_size: WindowSizeInDays | WindowSizeAsInterval = Field(
```
Alex-Bair: super tiny nit: could we swap the order here to be `WindowSizeAsInterval | WindowSizeInDays`? That way the interval option shows up first in the UI & it's more likely users will specify an interval instead of an integer. Doing this should also swap the `order` of the options in the associated `oneOf` in the spec snapshot too.

Alex-Bair (approved):
LGTM, thanks! It'd be nice to have the interval option show up first to encourage users to use it instead of the integer one, but that's just a small optional improvement.

source-salesforce-native/source_salesforce_native/models.py
nicolaslazo: Agreed, we should nudge users to use the new format

PR #4688, opened 2026-06-18
source-mysql: MariaDB binlog flavor

willdonnelly:
**Description:**

We've always hard-coded `Flavor: "mysql"` in the binlog syncer config but strictly speaking that's incorrect and we're seeing a failure in production which seems likely to be related to this.

I'm pretty sure using the MariaDB flavor will result in a few new-to-us / previously-unhandled replication event types which we don't care about, but if we don't do anything they'll hit the `unknown event type` default case and fail the task. To hopefully avoid that I've whitelisted them proactively.


Alex-Bair (approved):
LGTM

PR #4656, opened 2026-06-15
"observable" logs for backfills and column migrations

mdibaiee:
**Description:**

- add a few observable logs to begin with: backfills and column migrations. This allows us to get a sense of how many backfills and column migrations we are running, and from the task name we can infer the connector to see if a specific connector is doing it too much. We can create dashboards for these in Grafana.

**Workflow steps:**

(How does one use this feature, and how has it changed)

**Documentation links affected:**

(list any [documentation links](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing) that you created, or existing ones that you've identified as needing updates, along with a brief description)

**Notes for reviewers:**

(anything that might help someone review this PR)

PR #4689, opened 2026-06-18
source-{misc. CDK captures}: start backfills with `next_page=None` so observable logs are more accurate

Alex-Bair:
**Regression?**

No.

**Description:**

Some connectors have been seeding their initial backfill state with non-`None` `next_page` values. Functionally, that's perfectly fine, but it prevents the CDK from determining between a freshly started backfill & an ongoing backfill. We've recently made the `beginning backfill` log in the CDK an [observable log](https://github.com/estuary/connectors/pull/4656) so we can track which connectors are performing backfills & how frequently. It'd be nicer if as many CDK based connectors as possible didn't seed their backfills' initial `next_page`s with non-`None` values so those observable logs were more accurate / useful.

This PR updates many of the CDK connectors to stop seeding `next_page` with non-`None` values so fresh backfills trigger the `beginning backfill` log. There are still some more improvements we can make to certain connectors so these diagnostic logs are even more accurate to reality (ex: updating `source-shopify-native` to have distinct backfill / `fetch_page` tasks for incremental streams using bulk queries), but the current PR scope felt fairly well scoped by itself & tackled enough of the low hanging fruit for now.

**Notes for reviewers:**

No functional changes. They're all pretty mechanical changes of pushing the initial `page` determination from the `Resource` construction flow down into each `fetch_page` function.

### DIFF
--- a/source-braintree-native/source_braintree_native/api/__init__.py
+++ b/source-braintree-native/source_braintree_native/api/__init__.py
@@ -135,14 +135,18 @@ async def backfill_transactions(
     base_url: str,
     braintree_gateway: BraintreeGateway,
     window_size: int,
+    start_date: datetime,
     log: Logger,
     page: PageCursor | None,
     cutoff: LogCursor,
 ) -> AsyncGenerator[IncrementalResource | PageCursor, None]:
-    assert isinstance(page, str)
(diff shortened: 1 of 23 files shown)


nicolaslazo (approved):
LGTM!

PR #4692, opened 2026-06-21
add "task_name" to observable logs

mdibaiee:
**Description:**

- Add task_name to improve Grafana dashboards for these logs

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/task.py
+++ b/estuary-cdk/estuary_cdk/capture/task.py
@@ -115,6 +115,14 @@ class _BaseTask:
     _output: BinaryIO
     _tg: asyncio.TaskGroup
 
+    catalog_task_name: str
+    """The catalog task name of this capture (e.g. "acmeCo/my-capture"). Shared
+    by all Tasks of a capture and inherited by children via spawn_child. Used for
+    observability logging.
+
+    Distinct from the `_name` field, which is a connector-internal name for the
+    Python/asyncio task (e.g. "capture", "capture.webhook-server")."""
+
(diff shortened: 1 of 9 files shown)


estuary-cdk/estuary_cdk/capture/task.py
```diff
@@ -129,6 +134,7 @@ def __init__(
         tg: asyncio.TaskGroup,
         transactor: Transactor,
         requires_ack: bool = False,
+        task_name: str = "",
```
Alex-Bair: Do we need a default value for this new parameter? Any running catalog task should have a name, and I imagine we'd always want that name logged out instead of defaulting to an empty string.

estuary-cdk/estuary_cdk/capture/task.py
```diff
@@ -115,6 +115,11 @@ class Stopping:
     _output: BinaryIO
     _tg: asyncio.TaskGroup
 
+    task_name: str
```
Alex-Bair: "Task" is already a concept in Python & the CDK, and there's already a `_name` field on `_BaseTask`s that intuitively sounds like it should refers to the same thing as this new `task_name` field (it doesn't, which is why this new field is needed). Could we rename this new field so it's clearer that this refers to the catalog task name, not a Python task name? Maybe something like `catalog_task_name` would be enough of a distinction?

Alex-Bair (approved):
LGTM % renaming the variable internally to distinguish between catalog task names & connector internal task names.

PR #4695, opened 2026-06-22
source-hubspot-native: don't let the real-time engagements subtask trigger a backfill

Alex-Bair:
**Regression?**

No.

**Description:**

For a Hubspot account with lots of `engagments` activity, the real-time `engagements` subtask frequently hits the `/recent/modified` endpoint's 10k limit and triggers backfills via `MustBackfillBinding`. This was previously necessary when the `engagments` incremental task was _only_ using the `/recent/modified` endpoint. But now that there's a delayed subtask capturing from the `/modified/after` endpoint that doesn't have a limit on how many records can be returned, the real-time subtask no longer needs to trigger backfills when it loses consistency. When that happens, we rely on the more complete delayed subtask to capture any records the real-time subtask missed.

Additionally, I removed all the `MustBackfillBinding` handling since no stream in the connector uses it anymore & it was dead code.


willdonnelly (approved):
LGTM

PR #4697, opened 2026-06-22
source-mongodb: fix backfill dropping mixed-_id-type documents on resume

### LINKED ISSUE #4693: source-mongodb: mixed `_id` types -> resumed backfill silently skips an entire BSON type cohort
## Summary

A collection with mixed-type `_id` values (e.g. mostly `string`, a minority `ObjectId`) loses the entire minority-type cohort during the initial backfill whenever the backfill resumes, and the capture still reports the backfill as complete, so the loss is silent. Seen on a customer's MongoDB → Snowflake eval: 5.97M docs in Mongo, 5.91M landed, and the missing ~64k were exactly the `ObjectId`-`_id` documents.

## Root cause

The backfill is resumable: it checkpoints the last `_id` (`LastCursorValue`) and on resume rebuilds the cursor with a `$gt` filter in `backfill.go`:

```go
filter = bson.D{{Key: cursorField, Value: bson.D{{Key: "$gt", Value: v}}}}
```

MongoDB query operators type-bracket, so `{_id: {$gt: "<string>"}}` only matches values of the same BSON type and excludes every `ObjectId` ([type bracketing](https://www.mongodb.com/docs/manual/reference/method/db.collection.find/#type-bracketing)). In BSON canonical order `String` sorts before `ObjectId`, so the strings are read first. Once the cursor is re-established while still in the string range (connector restart, shard reassignment, server cursor timeout), the resume filter is `{_id: {$gt: "<a string>"}}` and can never reach the `ObjectId` bracket. The string cohort exhausts, the resume query returns 0 docs, and `doBackfill` treats the scan as finished and sets `Backfill.Done = true` having never queried an `ObjectId`.

Both the change-stream `_id`-hint path (`SetHint(bson.M{idProperty: 1})`) and the explicit-sort path are affected, since the bracketing is in the `$gt` filter, not the sort. This is separate from the sharded-ordering fixes, which correct ordering rather than the resume filter.

## Impact

- Silent data loss: a whole BSON `_id` type cohort is dropped while the backfill reports complete, so counts come up short by exactly that cohort.
- Hits any collection with mixed-type `_id` (common after partial migrations between `ObjectId` and string keys). No customer-side fix: it isn't a config/schema/key issue, and a plain re-backfill repeats the skip.
- An uninterrupted from-scratch backfill captures everything (the first query has no `$gt` filter), which is why small/local tests pass and large Atlas backfills fail.

## Proposed fix (for discussion)

1. Resume with a cross-type comparison, `{$expr: {$gt: ["$_id", v]}}`, instead of the bracketed `$gt` (needs a check on index-bound behavior under `$expr`).
2. When a resume query returns 0 docs, advance to the start of the next BSON type instead of declaring the backfill done.

Same cursor-recreation-skips-values family as #3723.

## Repro

Minimal, in mongosh:

```js
db.c.insertMany([{_id:"u-1"},{_id:"u-2"},{_id:new ObjectId()},{_id:new ObjectId()}])
db.c.find({}).sort({_id:1})    // all 4 (strings then ObjectIds)
db.c.find({_id:{$gt:"u-2"}})   // 0 rows; both ObjectIds excluded by type bracketing
```

End to end: seed a collection with string and `ObjectId` `_id`s, run a change-stream capture from a checkpoint whose `last_id` is a string (the state an interrupted backfill leaves), and it logs `completed backfill` while capturing none of the ObjectId docs.

Alex-Bair:
**Regression?**

No.

**Description:**

MongoDB's query comparison operators are type-bracketed: `{_id: {$gt: v}}` only matches values of the same BSON type as `v`. So when a collection's `_id` field holds more than one BSON type, a backfill that resumes from a value of one type drains the rest of that type, gets an empty result, and wrongly concludes the backfill is complete, silently skipping every document of a later-sorting type (ex: ObjectId-`_id` docs in a mostly string-`_id` collection).

This commit fixes this by replacing the single type-bracketed `$gt` resume filter with a union over BSON type brackets. The remainder of the resume value's own bracket via `$gt`, plus all of each later-sorting bracket via\ `{$gte: <bracketMin>}`, in [canonical BSON order](https://www.mongodb.com/docs/manual/reference/bson-type-comparison-order/).

This uses only `$or`/`$gt`/`$gte`, which allows MongoDB servers to uses indexes (confirmed with `explain()`) and avoids `$expr`, which appears to only use an index on MongoDB 5.0+ and is not fully supported on all [DocumentDB](https://docs.aws.amazon.com/documentdb/latest/devguide/expr.html) versions.

Cursor values of types that can't be range-scanned and are unlikely to be a good cursor field (regex, JS/code, array, min/max key) are rejected with an explicit error.

Closes https://github.com/estuary/connectors/issues/4693


willdonnelly (approved):
LGTM

PR #4698, opened 2026-06-22
delayed_tag: retry Claude API errors instead of assuming safe

mdibaiee:
**Description:**

- We assume safety when there is a Claude error, which is quite dangerous. This changes it so we first attempt retry, and if that doesn't work, we fail the pipeline

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/.github/scripts/delayed_tag.py
+++ b/.github/scripts/delayed_tag.py
@@ -765,13 +765,18 @@ def build_claude_prompt(
     return '\n\n'.join(b['text'] for b in blocks)
 
 
-def call_claude(content, api_key: str, model: str = 'claude-sonnet-4-6') -> str:
-    """Call the Anthropic Messages API and return the assistant's text.
+# HTTP statuses worth retrying: rate-limit (429), request timeout/conflict, and
+# server/gateway/overload errors. Any other 4xx (e.g. 401 auth, 400 bad request)
+# won't be fixed by retrying, so we fail fast and loudly on those.
+_RETRYABLE_STATUS = frozenset({408, 409, 429, 500, 502, 503, 504, 529})
 
-    `content` is a Messages API content value — either a plain string or a list
-    of content blocks (the cache-aware form from build_claude_content). Token
-    usage (including cache reads/writes) is logged to stderr so the cache hit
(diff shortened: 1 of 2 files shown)


.github/scripts/delayed_tag.py
```diff
@@ -786,20 +789,67 @@ def call_claude(content, api_key: str, model: str = 'claude-sonnet-4-6') -> str:
...
+    for attempt in range(1, attempts + 1):
+        try:
+            return _claude_request(content, api_key, model)
+        except urllib.error.HTTPError as exc:
+            if exc.code not in _RETRYABLE_STATUS:
+                raise RuntimeError(
+                    f'Claude API error: HTTP {exc.code} {exc.reason}'
+                ) from exc
+            last_exc = exc
+        except (urllib.error.URLError, KeyError, json.JSONDecodeError) as exc:
```
danielnelson: I would skip catching the KeyError, I don't expect it will happen if the status code is right.

.github/scripts/delayed_tag.py
```diff
@@ -786,20 +789,67 @@ def call_claude(content, api_key: str, model: str = 'claude-sonnet-4-6') -> str:
...
+    rate is visible in CI logs.
+
+    Transient failures (network errors, HTTP 429/5xx, malformed responses) are
+    retried up to `attempts` times with exponential backoff; non-retryable
+    failures (e.g. HTTP 401/400) raise immediately. When every attempt is
+    exhausted the final RuntimeError propagates. Callers MUST NOT downgrade this
+    to a 'safe' verdict: a roll-forward check we never ran is not a check that
+    passed, and treating it as safe could ship a known-buggy image to
+    ':delayed'."""
+    last_exc: Optional[BaseException] = None
```
danielnelson: Suggest using this more modern style instead of the old Optional, Python 3.9 is EOL so you can count on it everywhere.  Use Exception instead of BaseException because all the errors caught are Exception and it seems unlikely you would ever want to catch a BaseException for purposes of retrying.
```suggestion
    last_exc: Exception | None = None
```

danielnelson (approved):

PR #4699, opened 2026-06-22
source-facebook-marketing-native: remove `ad_insights_dma` stream due to out-of-cycle API change

Alex-Bair:
**Regression?**

No.

**Description:**

Meta made an [out-of-cycle API change](https://developers.facebook.com/documentation/ads-commerce/marketing-api/out-of-cycle-changes/occ-2026#transitioning-from-nielsen-dmas-to-comscore-markets) to prevent anyone from using `breakdowns=dma` when requesting ad insights. This permanently breaks our pre-built `ad_insights_dma` stream, so we should stop discovering it.

The recommended alternative is `breakdowns=comscore_market`, so I also added a pre-built `ad_insights_comscore_market` stream. The underlying data is grouped differently from `breakdowns=dma`, which is why this is a completely different stream that will feed a completely different collection.


willdonnelly (approved):
LGTM

PR #4700, opened 2026-06-22
source-mysql: More efficient chunked discovery queries

willdonnelly:
**Description:**

Modifies the discovery query chunking behavior and the query filter construction from `WHERE (table_schema, table_name) IN ((?, ?), ...` to `WHERE table_schema = ? AND table_name IN (?, ?, ...)`.

The former construction basically always forces a full-table scan over the relevant system tables because the query optimizer isn't smart enough to decompose the tuple constructors and push it down as individual equality checks. The new one also can't do that for individual tables but it behaves much more efficiently on schemas so I think it's a solid improvement.

If it turns out we really need point-access performance it might be possible to decompose discovery queries into a big hairy `UNION ALL` across individual `WHERE schema = ? AND table = ?` queries but the result just looks so awful I can't stand it and so I'd rather not go that far unless we're forced to.


Alex-Bair (approved):
LGTM

PR #4707, opened 2026-06-23
sqlcapture: Lower rediscovery interval

willdonnelly:
**Description:**

During the main capture loop we re-run discovery every 5 minutes to drive several behaviors like sourcedSchema generation and re- enabling dropped tables if/when they come back. The specific 5m interval was somewhat arbitrary and predicated on the assumption that discovery is tolerably cheap.

Since we're seeing some production databases struggling with the burden of all the discovery queries we send (from multiple captures) let's turn the obvious tuning knob up to 15m. This is still plenty fast for our purposes but means 1/3rd as much discovery work, so it's a cheap win while we finish the real performance improvements.

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -174,7 +174,7 @@ type Capture struct {
 }
 
 const (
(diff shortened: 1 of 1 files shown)


Alex-Bair (approved):
LGTM

PR #4708, opened 2026-06-23
source-sentry: bound incremental explore window size

Alex-Bair:
**Regression?**

No.

**Description:**

The incremental cursor for custom explore streams is seeded at the trailing edge of the 30-day full-fidelity window. `fetch_explore_query` only checkpoints a new cursor after draining the entire window, so a high-volume stream's first poll may never finished before the connector restarts. This causes the incremental task to get stuck attempting to process an ever growing amount of data.

This PR fixes this by bounding each `fetch_explore_query`'s incremental poll to be at max `_MAX_INCREMENTAL_EXPLORE_WINDOW` (1 day) wide. This will hopefully allow captures to work through larger swaths of changes more easily. If it doesn't, I'll look into adding a configurable `window_size` setting in the endpoint config instead of hardcoding it to 1 day.

### DIFF
--- a/source-sentry/source_sentry/api.py
+++ b/source-sentry/source_sentry/api.py
@@ -185,6 +185,8 @@ _EXPLORE_INGESTION_LAG = timedelta(minutes=10)
 
 _EXPLORE_BACKFILL_WINDOW = timedelta(days=1)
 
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM

PR #4709, opened 2026-06-23
fetch-flow.sh: Use latest flow-connector-init

willdonnelly:
**Description:**

I have no idea what's up with this whole "pull a stale release tarball which was last updated December 2025 and then overwrite some of the binaries with newer ones" thing, but it looks like adding `flow-connector-init` to the list of newer binaries is the most minimal fix to unbreak CI at the moment.


Alex-Bair (approved):
LGTM

PR #4715, opened 2026-06-23
materialize-mongodb: preserve integer precision in stored documents

jacobmarble:
**Regression?**

Latent since the connector's first commit; surfaced as a crash by the strict-insert path added later. Brief history:
- `ed8b7ac41` (2023-03-06) — first commit stores docs via `json.Unmarshal` into `bson.M`, which decodes JSON numbers as `float64`, silently rounding integers > 2^53. Harmless then: every store was an idempotent upsert.
- `ae246b889` (2023-09-08) — switched to load-driven strict `InsertOne` on `!Exists`. From here the rounding can cause a duplicate-key crash.

**Description:**

Preserve integer precision when building MongoDB documents to store. A large integer key (e.g. an `EVENT_ID` near 2^63) was stored as a rounded `float64`. The runtime re-derives a store's key from the document body returned by Load (a `Loaded` carries no key), so the rounded value no longer matched the requested key → `Exists=false` → strict `InsertOne` of an `_id` that already exists → `E11000`, in a permanent crash loop (a backfill can't fix it; the key re-corrupts immediately).

Fix: decode with `json.Decoder.UseNumber()` and keep integral values that fit as `int64` (recursively). Keys now round-trip exactly. The `E11000` tripwire is untouched — no idempotent masking.

**Workflow steps:**

Behavior-only; no config/API change. Standard materializations with integer keys beyond 2^53 stop diverging.

**Notes for reviewers:**

- Three commits: behavior-preserving `storeDocument` extraction (green) → failing regression test (red) → fix (green).
- `TestStoreDocumentPreservesLargeIntegerKey` drives the full store→BSON→load-sanitize→re-extract round-trip.
- Out of scope: values genuinely outside the int64 range still fall back to `float64` — BSON has no native uint64; pre-existing.
- Affected destinations should be re-backfilled after deploy to rewrite the lossy doubles already stored.

### DIFF
--- a/materialize-mongodb/transactor.go
+++ b/materialize-mongodb/transactor.go
@@ -18,6 +18,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
@@ -72,10 +73,21 @@ type transactor struct {
 type binding struct {
 	collection   *mongo.Collection
 	deltaUpdates bool
-	// keyTokens holds the parsed RFC6901 JSON pointer tokens of the binding's
-	// key fields, in the same order as a store's key tuple, used to restore
-	// exact integer key values in the stored document.
-	keyTokens [][]string
+	// keyRestorations holds, for each of the binding's key fields declared as
+	// JSON-schema integer (the only type whose stored precision needs
+	// restoring — see storeDocument), its index into a store's key tuple and
+	// parsed RFC6901 JSON pointer tokens. Key fields of other declared types
+	// are omitted so the per-document restore loop never re-derives their
+	// (never-integer) type.
+	keyRestorations []keyRestoration
+}
+
+// keyRestoration is a single entry of binding.keyRestorations. tupleIndex
+// indexes into a store's key tuple; tokens is the pre-parsed JSON pointer to
+// that key field's location in the stored document.
+type keyRestoration struct {
+	tupleIndex int
+	tokens     []string
 }
 
 func (t *transactor) RecoverCheckpoint(ctx context.Context, spec pf.MaterializationSpec, rangeSpec pf.RangeSpec) (m.RuntimeCheckpoint, error) {
@@ -191,7 +203,7 @@ func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
 			batchSize += len(key)
 		} else {
 			b := t.bindings[it.Binding]
-			doc, err := storeDocument(it.RawJSON, key, b.deltaUpdates, it.Key, b.keyTokens)
+			doc, err := storeDocument(it.RawJSON, key, b.deltaUpdates, it.Key, b.keyRestorations)
 			if err != nil {
 				return nil, err
 			}
@@ -466,7 +478,7 @@ func storeRetryDelay(ctx context.Context, attempt int) error {
 // document body a rounded key field would no longer match — provoking a
 // duplicate-key crash loop. Non-key numbers are intentionally left as float64 so
 // a field's BSON type stays stable regardless of its value.
-func storeDocument(rawJSON json.RawMessage, idValue string, deltaUpdates bool, key tuple.Tuple, keyTokens [][]string) (bson.M, error) {
+func storeDocument(rawJSON json.RawMessage, idValue string, deltaUpdates bool, key tuple.Tuple, keyRestorations []keyRestoration) (bson.M, error) {
 	var doc bson.M
 	if err := json.Unmarshal(rawJSON, &doc); err != nil {
 		return nil, fmt.Errorf("unmarshalling json doc: %w", err)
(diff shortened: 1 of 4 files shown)


materialize-mongodb/transactor.go
```diff
@@ -461,6 +449,88 @@ func storeRetryDelay(ctx context.Context, attempt int) error {
...
+		for i, vv := range t {
+			t[i] = convertJSONNumbers(vv)
+		}
+		return t
+	case json.Number:
+		if i, err := t.Int64(); err == nil {
+			return i
+		}
+		f, _ := t.Float64()
+		return f
```
danielnelson: I'm concerned about type stability, depending on the size of the value the type of the field changes.  Perhaps we should use a fixed conversion for the field based on the inference?

When the number is larger than int64, does the error come back?

materialize-mongodb/transactor.go
```diff
@@ -461,6 +449,88 @@ func storeRetryDelay(ctx context.Context, attempt int) error {
...
+// Integer precision must be preserved. The runtime determines a store's Exists
+// flag by re-deriving the document key from the document body it gets back from
+// a Load (a Loaded response carries only the document, not its key). A plain
+// json.Unmarshal into a map decodes every JSON number as float64, so an integer
+// key beyond float64's exact range (2^53) — e.g. a large EVENT_ID — is rounded
+// when stored. On the next load round-trip the runtime re-derives a different
+// key, never matches the document, reports the key as absent (Exists=false), and
+// the connector issues a strict InsertOne for an _id that already exists →
+// E11000. Decoding with UseNumber and keeping integral values as int64 keeps the
+// stored document faithful so the key round-trips exactly.
```
danielnelson: This comment makes sense now but it feels backwards facing, you have to know the history of the function to understand it.  Since this is a function level comment, I'd rather have this comment only describe the current behavior of the function, not previous bugs or things that could happen if it were implemented differently.

Inside the function, we could add comments about potential issues we are handling, but things like EVENT_ID or "Unmarshal into a map" might be confusing.  I think some of the details of this specific bug could be added to the regression test case.

materialize-mongodb/transactor.go
jacobmarble: The behavior for values outside of the int64 range is **quietly lossy**, but the full int64 range is now exactly round-trippable. This is similar to SQL Server, Azure Fabric, and others that don't explicitly handle wide integer values.

danielnelson: My main concern is type changes.  One thing I'm thinking about is if a collection have key values like 0.5, 1, 1.5, 2 in a float field, some with decimal points and some without but all floats.  If this can happen it would insert float, int, float, int into the database, instead of a consistent type based on the collection types.

I agree about the out of range float, it still doesn't work well but is not a regression.

danielnelson (approved):

materialize-mongodb/transactor.go
mdibaiee: I think as a general rule of thumb (which might again be broken by a customer), but we tend to not treat floats as first-class primary keys. They are an unusual edge case at best, since floats are inherently imprecise, they are a really bad choice for a primary key. I don't remember any customer having float keys (which could any day be broken by a new customer), so we shouldn't worry too much about float keys.

That said, I think your point still stands that the difference in typing which is unstable could cause issues down the road in other scenarios

materialize-mongodb/transactor.go
```diff
@@ -461,6 +456,136 @@ func storeRetryDelay(ctx context.Context, attempt int) error {
...
+	if !deltaUpdates {
+		if len(key) != len(keyTokens) {
+			panic(fmt.Sprintf("store key tuple has %d elements but binding declares %d key fields", len(key), len(keyTokens)))
+		}
+		for i, tokens := range keyTokens {
+			if v, ok := key[i].(int64); ok {
+				setAtPointer(doc, tokens, v)
+			}
+		}
+	}
```
mdibaiee: I'm a bit worried about the performance implication of running this check every time, and also the setting of the value in a nested fashion goes through a lot of type reflections and checks which are usually slow in Go.

I think ideally we check for integer keys specifically when we are creating `keyTokens` (we also check their types and persist), so here we only loop over known-integer keys.

An additional optimization may be that if the jsonPointer is pointing at a field in the root of the document (most common case), we directly set that field instead of iterating through jsonPointer and spending time doing type checks for nested objects / arrays.

materialize-mongodb/transactor.go
```diff
@@ -461,6 +456,136 @@ func storeRetryDelay(ctx context.Context, attempt int) error {
...
+	return tokens
+}
+
+// setAtPointer writes val at the location addressed by the parsed JSON pointer
+// tokens within doc, traversing both JSON objects and arrays. It no-ops when the
+// location is absent or a token does not address its parent's type, degrading to
+// the value json.Unmarshal already decoded rather than altering the document's
+// shape.
+func setAtPointer(doc bson.M, tokens []string, val any) {
```
mdibaiee: I think it would be better for us to use an existing jsonpointer library to reduce the code we have to maintain ourselves: https://pkg.go.dev/github.com/go-openapi/jsonpointer#section-readme

mdibaiee (commented):
Thanks! I'm a bit worried about looping over all keys of all documents doing type inference over `interface{}` which can be quite slow...

materialize-mongodb/transactor.go
jacobmarble: Following up: checked whether a Flow collection key can actually be typed as bare JSON Schema `number` (the mixed 0.5/1/1.5/2 scenario above). It can't — Flow's catalog validation rejects it at publish time. `is_keyable_type()` (`flow/crates/json/src/schema/types.rs`) only permits `BOOLEAN | INTEGER | STRING` for a key location; a bare `number` key fails with `KeyWrongType` (`flow/crates/validation`) before the collection spec is ever accepted. So this isn't just rare — it's structurally unreachable.

materialize-mongodb/transactor.go
jacobmarble: Fixed — precomputed which key fields are integer-typed at bind time (`isIntegerKeyType` in driver.go → `binding.keyRestorations`), so the per-document loop only touches eligible keys. Added a root-level fast path in `storeDocument` too, skipping `setAtPointer` for the common single-segment case.

Benchmarked before/after: delta is within noise — `json.Unmarshal` dominates cost either way. Still the structurally correct fix, and matches the precompute pattern `materialize-elasticsearch` already uses.

materialize-mongodb/transactor.go
jacobmarble: Done.

mdibaiee (approved):

PR #4720, opened 2026-06-25
go/blob: validate bucket path is a bare prefix, not a full URI

jwhartley:
## Problem

A materialization's staging-prefix config field (`bucket_path` or `prefix`) is meant to be a bare key prefix within the bucket; the bucket itself is configured separately. If it is mistakenly set to a full URI like `s3://my-bucket/my/prefix`, the connector silently produces corrupt object keys.

The prefix is joined onto object keys with `path.Join`, whose `path.Clean` collapses `s3://` to `s3:/`. The staged-file keys then start with a stray `s3:` segment, and the downstream reader fails. In `materialize-iceberg` the Spark merge job throws:

```
java.lang.IllegalArgumentException: java.net.URISyntaxException: Expected scheme-specific part at index 3: s3:
```

This is hard to diagnose: the config publishes cleanly (existing validation only rejected a leading `/`) and fails much later, deep inside a Spark job, with an opaque Java URI error.

## Fix

Add a shared `blob.ValidateBucketPath(field, prefix)` helper (`go/blob/path.go`) that rejects:

- a leading `/` (preserving the existing behavior and message), and
- an embedded URI scheme (`://`), with an actionable message pointing at the correct format.

Wire it into every materialization that takes a user-configurable staging prefix:

| Connector | Field | Prior validation |
|-----------|-------|------------------|
| filesink (s3/gcs/azure parquet + csv) | `prefix` | leading `/` only |
| materialize-iceberg | `bucket_path` | leading `/` only |
| materialize-s3-iceberg | `prefix` | leading `/` only |
| materialize-redshift | `bucketPath` | none |
| materialize-bigquery | `bucket_path` | none |

`materialize-redshift` and `materialize-bigquery` had no prefix validation before. They pass the leading-slash-trimmed value (`effectiveBucketPath()`), so their existing tolerance of a leading slash is preserved while the new scheme check still applies.

## Why `go/blob`

`go/blob` is the lowest shared layer that owns object-key/URI construction (`URI()` per backend) and is already imported by `filesink`, `materialize-boilerplate`, and the materializations. Putting the helper there both fixes the footgun and collapses the duplicated `prefix %q cannot start with /` check that previously lived in five separate files.

## Testing

- New table test `TestValidateBucketPath` (10 cases) passes: `go test ./go/blob/...`.
- `go build` and `go vet` pass for all touched packages.

## Note for reviewers

Adding validation where there was none (redshift, bigquery) means an already-broken config that contains a URI scheme will now fail at publish with a clear message instead of failing later at runtime. That is the intent, but flagging it as a behavior change. A natural follow-up is to surface the same guard at the config-schema/UI layer so it is caught before publish.

### DIFF
--- a/filesink/store.go
+++ b/filesink/store.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"strings"
 	"time"
 
 	storage "cloud.google.com/go/storage"
@@ -59,8 +58,8 @@ func (c GCSStoreConfig) Validate() error {
 	} else if _, err := time.ParseDuration(c.UploadInterval); err != nil {
 		return fmt.Errorf("parsing upload interval %q: %w", c.UploadInterval, err)
 	} else if c.Prefix != "" {
-		if strings.HasPrefix(c.Prefix, "/") {
-			return fmt.Errorf("prefix %q cannot start with /", c.Prefix)
+		if err := blob.ValidateBucketPath("prefix", c.Prefix); err != nil {
+			return err
 		}
 	} else if c.FileSizeLimit < 0 {
 		return fmt.Errorf("fileSizeLimit '%d' cannot be negative", c.FileSizeLimit)
@@ -142,8 +141,8 @@ func (c AzureBlobConfig) Validate() error {
 	if _, err := time.ParseDuration(c.UploadInterval); err != nil {
 		return fmt.Errorf("parsing upload interval %q: %w", c.UploadInterval, err)
 	} else if c.Prefix != "" {
-		if strings.HasPrefix(c.Prefix, "/") {
-			return fmt.Errorf("prefix %q cannot start with /", c.Prefix)
+		if err := blob.ValidateBucketPath("prefix", c.Prefix); err != nil {
+			return err
 		}
 	} else if c.FileSizeLimit < 0 {
 		return fmt.Errorf("fileSizeLimit '%d' cannot be negative", c.FileSizeLimit)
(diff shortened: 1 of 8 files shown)


filesink/store.go
```diff
@@ -59,8 +58,8 @@ func (c GCSStoreConfig) Validate() error {
 	} else if _, err := time.ParseDuration(c.UploadInterval); err != nil {
 		return fmt.Errorf("parsing upload interval %q: %w", c.UploadInterval, err)
 	} else if c.Prefix != "" {
-		if strings.HasPrefix(c.Prefix, "/") {
-			return fmt.Errorf("prefix %q cannot start with /", c.Prefix)
+		if err := blob.ValidateBucketPath("prefix", c.Prefix); err != nil {
+			return err
```
mdibaiee: Passing the field name as an argument feels weird, I would prefer something like this:

```suggestion
		if err := blob.ValidateBucketPath(c.Prefix); err != nil {
			return fmt.Errorf("prefix %w", err)
```

materialize-bigquery/bigquery.go
```diff
@@ -136,6 +137,10 @@ func (c config) Validate() error {
 		return err
 	}
 
+	if err := blob.ValidateBucketPath("bucket path", c.effectiveBucketPath()); err != nil {
```
mdibaiee: let's use the actual `bucket_path` field name

mdibaiee (commented):
Thanks! two comments below

filesink/store.go
jwhartley: Done in 67455619c. Dropped the field argument; ValidateBucketPath now returns a field-agnostic error and each caller wraps it with fmt.Errorf("<field> %w", err).

materialize-bigquery/bigquery.go
jwhartley: Done in 67455619c. Using the actual JSON field names now: bucket_path here, prefix for filesink and s3-iceberg, bucketPath for redshift.

mdibaiee (approved):

PR #4721, opened 2026-06-25
Add custom endpoint support for BigQuery emulator and proxy use

Shareed2k:
**Description:**

Adds an advanced `endpoint` option to BigQuery source and materialization connectors so they can connect to a BigQuery emulator or proxy server.

When `advanced.endpoint` is set, BigQuery clients are configured with `option.WithEndpoint(...)` and `option.WithoutAuthentication()`.

**Workflow steps:**

Set `advanced.endpoint` in the connector endpoint config to the BigQuery emulator or proxy server URL.

Example:

```yaml
advanced:
  endpoint: http://localhost:9050
```

For normal BigQuery usage, leave `advanced.endpoint` unset. Existing authenticated behavior is unchanged.

BigQuery connector docs should mention the new advanced `endpoint` option for emulator/proxy usage.

**Notes for reviewers:**

- BigQuery source connectors skip credential validation when `advanced.endpoint` is set.
- Materialize BigQuery still uses credentials for GCS staging, but the BigQuery client itself uses the custom endpoint without auth.
- Storage Read client setup is skipped for custom materialize endpoints.
- Spec snapshots were updated for the new advanced field.
- Verified with:
  - `go test ./go/capture/bigquery/client ./source-bigquery-batch ./source-ga4-bigquery`
  - `go test ./materialize-bigquery -run 'TestSpecification|TestSQLGeneration'`
- Full `go test ./materialize-bigquery` requires `flowctl` in `PATH`.

[bento PR](https://github.com/warpstreamlabs/bento/pull/887)
[bqemulator PR](https://github.com/jjviscomi/bqemulator/pull/149)


jacobmarble (commented):
This looks good to me so far. What emulator or implementation have you tested this against? I assume bqemulator because it is mentioned; any others?

jacobmarble (changes requested):
Before we can merge this change, we need the feature covered by automated tests, because right now nothing verifies it works or keeps working.

**Merge bar for this PR:** `go test ./materialize-bigquery/...` — with **no special flags, no env vars, and no access to Google Cloud** — must pass against a local emulator started via `docker-compose`, the same way our other emulator-backed connectors test. (Any tests that need a real BigQuery project should become *opt-in*, not the default.)

We have two solid in-repo templates for exactly this:

- **`materialize-bigtable`** — closest model. It runs against the Cloud SDK emulator **by default** and only touches real Cloud Bigtable when you opt in. See:
  - [`materialize-bigtable/driver_test.go`](https://github.com/estuary/connectors/blob/main/materialize-bigtable/driver_test.go) — `docker compose up --wait` + `t.Cleanup(down -v)`, emulator-by-default with a `-bigtable.test-all` / `BIGTABLE_TEST_ALL` opt-in for real cloud, and `Endpoint` set in the test config.
  - [`materialize-bigtable/docker-compose.yaml`](https://github.com/estuary/connectors/blob/main/materialize-bigtable/docker-compose.yaml) — emulator service + healthcheck on the `flow-test` network.
- **`materialize-spanner`** — another GCP-emulator example: [`docker-compose.yaml`](https://github.com/estuary/connectors/blob/main/materialize-spanner/docker-compose.yaml), [`main_test.go`](https://github.com/estuary/connectors/blob/main/materialize-spanner/main_test.go), [`testdata/config.local.yaml`](https://github.com/estuary/connectors/blob/main/materialize-spanner/testdata/config.local.yaml).

### The easy part: the harness

Mirror the bigtable layout. Roughly:

`materialize-bigquery/docker-compose.yaml`:
```yaml
services:
  bigquery:
    image: ghcr.io/goccy/bigquery-emulator:latest
    command: ["--project=test-project", "--dataset=testing"]
    networks: [flow-test]
    ports:
      - "9050:9050"   # REST
      - "9060:9060"   # gRPC
    environment:
      STORAGE_EMULATOR_HOST: "http://gcs:4443"   # see "the hard part" below
    healthcheck:
      test: ["CMD-SHELL", "wget -qO- http://localhost:9050/bigquery/v2/projects/test-project/datasets || exit 1"]
      interval: 2s
      timeout: 2s
      retries: 30
networks:
  flow-test:
    name: flow-test
    external: true
```

`testdata/config.local.yaml` (note the endpoint base path — see gotcha):
```yaml
project_id: test-project
dataset: testing
region: US
bucket: test-bucket
advanced:
  endpoint: http://localhost:9050/bigquery/v2/
```

`*_test.go` (start/stop compose; emulator-only by default):
```go
require.NoError(t, exec.Command("docker", "compose", "-f", "docker-compose.yaml", "up", "--wait").Run())
t.Cleanup(func() { exec.Command("docker", "compose", "-f", "docker-compose.yaml", "down", "-v").Run() })
```

Two connector gaps to close so an emulator config is even accepted (you already did the equivalent in the two `source-*` connectors, but the materializer was missed):
- [`materialize-bigquery/bigquery.go`](https://github.com/estuary/connectors/blob/main/materialize-bigquery/bigquery.go) — `config.Validate()` still requires credentials; skip that block when `Advanced.Endpoint != ""`, same pattern you used in `source-bigquery-batch`/`source-ga4-bigquery`.
- [`materialize-bigquery/client.go`](https://github.com/estuary/connectors/blob/main/materialize-bigquery/client.go) — `preReqs()` verifies a real GCS bucket; short-circuit it when an endpoint is set.

**Gotcha I hit:** the BigQuery Go client treats `option.WithEndpoint(...)` as the *full* API base path (normally `.../bigquery/v2/`). Pointing it at `http://host:9050` gives 404s; it needs `http://host:9050/bigquery/v2/`. Please confirm the exact base path your emulator serves.

### The hard part: the load/merge path won't run on a BigQuery emulator as written

I ran the real BigQuery client against `jjviscomi/bqemulator` to check. DDL/DML/datasets work, **but the connector's actual load path does not**, because Load ([`transactor.go`](https://github.com/estuary/connectors/blob/main/materialize-bigquery/transactor.go)) and Acknowledge ([`query.go`](https://github.com/estuary/connectors/blob/main/materialize-bigquery/query.go)) build queries with `Query.TableDefinitions` — i.e. **temporary external tables over `gs://` URIs**. No BigQuery emulator I'm aware of implements external data configuration:

- `jjviscomi/bqemulator`: silently ignores `tableDefinitions`; the job fails with `"Not found: Table <name> ..."`. (It also rejects multi-clause `ALTER TABLE` — `"Only one ALTER command per statement is supported"` — which breaks the standard apply test's multi-column `DROP NOT NULL`.)
- `goccy/bigquery-emulator`: external tables are "not implemented" either — **but** per its docs it does support native **load jobs from GCS** and a `fsouza/fake-gcs-server` via `STORAGE_EMULATOR_HOST`.

So to get the materialize/merge tests passing on an emulator, the feature likely needs an emulator-compatible load path: when a custom endpoint is configured, stage to a GCS emulator and load via a **native load job into a temporary native table**, then merge — instead of the external-table query. That's why the compose sketch above pairs `goccy/bigquery-emulator` with `fake-gcs-server`:

```yaml
  gcs:
    image: fsouza/fake-gcs-server:latest
    command: ["-scheme", "http", "-host", "0.0.0.0", "-port", "4443", "-public-host", "gcs:4443"]
    networks: [flow-test]
    ports: ["4443:4443"]
```

(Treat the `goccy` + `fake-gcs-server` snippets as a starting point — I verified the `bqemulator` behavior hands-on but not these two; check the upstream [goccy](https://github.com/goccy/bigquery-emulator) and [fake-gcs-server](https://github.com/fsouza/fake-gcs-server) docs for exact flags.)

### Scope

The custom-endpoint *load* path is the heart of this feature, so it really should be exercised end-to-end on the emulator. If that turns out to be a large lift, let's talk about scope on the PR — but at minimum `go test ./materialize-bigquery/...` needs to run green against docker-compose only, with no GCP access, before we merge. Happy to help review the harness once you have a draft.

PR #4727, opened 2026-06-29
source-monday: reduce `items` page size used for subsequent pages during backfills

Alex-Bair:
**Regression?**

No.

**Description:**

The first page when fetching `items` during a backfill is small (`5`), but subsequent pages always requested `100` items. I suspect that's causing 500 errors from Monday's API during an ongoing backfill, so I'd like to try reducing the subsequent page size and see if that helps avoid those 500 errors.

**Notes for reviewers:**

I'm not confident `25` is low enough, and we may need to dynamically reduce the page size when the connector receives 500 responses from the API, but that's a pretty complex change I'd like to postpone if dropping the page size to `25` is sufficient to address the current issue.


nicolaslazo (approved):
LGTM

Alex-Bair:
This change does not address the 500 internal server errors the capture task is hitting. I'm closing this PR & will get a different one up for review to address the actual issue causing those 500 errors.

PR #4730, opened 2026-06-30
source-kafka: route output through a buffered writer

Alex-Bair:
**Regression?**

No

**Description:**

An active `source-kafka` capture appears resource constrained - throughput rose from ~60 GB / hr -> ~90 GB / hr just from moving it to a beefier reactor. Reviewing the connector, it does per-message I/O:
- It emits a `Captured` document & a `Checkpoint` after every message.
- Output goes to a raw `std::io::Stdout`, [which is a `LineWriter`](https://doc.rust-lang.org/src/std/io/stdio.rs.html#608-613) so every line forces a `write` syscall and the existing `if response.captured.is_none() { flush }` guard never actually batched anything.

The changes in this PR include:
- Buffering `stdout` in a 256 KiB `BufWriter` so captured docs accumulate and flush on the next checkpoint.

### DIFF
--- a/source-kafka/src/pull.rs
+++ b/source-kafka/src/pull.rs
@@ -28,6 +28,7 @@ use rdkafka::{
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Map};
 use std::collections::{hash_map::Entry, HashMap};
+use std::io::{BufWriter, Stdout};
 use time::{format_description, OffsetDateTime};
 
 #[derive(Debug, Deserialize, Serialize, Default)]
@@ -73,7 +74,7 @@ enum MetaTimestamp {
     LogAppendTime(String),
 }
 
-pub async fn do_pull(req: Open, mut stdout: std::io::Stdout) -> Result<()> {
+pub async fn do_pull(req: Open, mut stdout: BufWriter<Stdout>) -> Result<()> {
     let spec = req.capture.expect("open must contain a capture spec");
 
     let state = if req.state_json == "{}" {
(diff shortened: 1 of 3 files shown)


source-kafka/src/pull.rs
```diff
@@ -28,29 +28,70 @@ use rdkafka::{
...
 use serde_json::{json, Map};
 use std::collections::{hash_map::Entry, HashMap};
+use std::io::{BufWriter, Stdout};
+use std::time::{Duration, Instant};
 use time::{format_description, OffsetDateTime};
 
+/// How long a single `poll` blocks waiting for a message before returning
+/// `None`, which triggers a flush of any pending batch. Bounds output latency
+/// on low-traffic topics so batching never strands pending documents.
+const POLL_TIMEOUT: Duration = Duration::from_millis(250);
```
Alex-Bair: We're holding off merging this PR until we see how much performance other improvements give us. We'd rather avoid accepting a 250ms latency tradeoff if possible.

williamhbaker:
A very small change that would unambiguously be worth doing: With the current scheme, we flush after every document written to stdout, and then flush again after the checkpoint. As you've noted, the check [here](https://github.com/estuary/connectors/blob/383fa59b56c20ff1fe4edac046f060d3378842fd/source-kafka/src/lib.rs#L119-L130) doesn't actually prevent a flush syscall. Keeping the bit about wrapping stdout in a `BufWriter` would fix that. So I'd say we should do that at least and see how much it helps.

source-kafka/src/pull.rs
Alex-Bair: The change introducing the 250ms latency has been dropped from the PR.

Alex-Bair:
That makes sense to me. I updated the PR to only include the `BufWriter` change.

All the CI checks are failing per the issue mentioned in https://estuaryworkspace.slack.com/archives/C03Q9A3GFEE/p1782864119458959, but the did succeed previously. So, I'll have to wait until that issue's resolved before merging any changes.

Edit: That issue's resolved now & the `source-kafka` CI check passed after another retry.

williamhbaker (approved):
LGTM

PR #4731, opened 2026-06-30
source-kafka: use `jemalloc` as the global allocator

Alex-Bair:
**Regression?**

No

**Description:**

Profiling with `perf` showed allocation pressure on the read path for a capture with steady throughput at ~90 GB / hr. This PR switches the connector to `jemalloc` via the shared `allocator` crate.

Using that crate required a current `flow` rev, and the latest rev moved `flow` onto `prost` 0.14, so the first commit bumps the `flow` deps and aligns the connector's `prost` stack (`prost-reflect` 0.14→0.16, `prost-types` 0.13→0.14) onto a single `prost` 0.14.

We're hoping for a small (~5%) performance gain from these changes on the bottlenecked capture, and we'll measure the actual improvement after rolling out this change.

**Notes for reviewers**

A transient HTTP error failed the first run of the `source-kafka` CI check. I cancelled the remaining checks so I could retry the relevant `source-kafka` one sooner, and that's why so many of the checks are cancelled.

### DIFF
--- a/source-kafka/Cargo.lock
+++ b/source-kafka/Cargo.lock
@@ -775,7 +775,7 @@ dependencies = [
 [[package]]
 name = "doc"
 version = "0.0.0"
-source = "git+https://github.com/estuary/flow#4360bd459bbf993715893f5c8132bcea175de5c6"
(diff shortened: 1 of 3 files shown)


williamhbaker (approved):
LGTM

PR #4732, opened 2026-06-30
source-kafka: serialize protobuf messages directly to a JSON value

Alex-Bair:
**Regression?**

No.

**Description:**

The protobuf decode path serialized each `DynamicMessage` to a byte buffer and then re-parsed those bytes into a `serde_json::Value`. This was a redundant round-trip that profiling flagged as a hot spot.

This PR avoids that extra serialization by passing a `serde_json::value::Serializer` to `serialize_with_options` so the message is built straight into the `Value` in one pass, dropping the intermediate buffer and the re-parse.


williamhbaker (approved):
LGTM

PR #4734, opened 2026-07-01
source-kafka: serialize protobuf messages without an intermediate `serde_json::Value`

Alex-Bair:
**Regression?**

No.

**Description:**

The per-message path decoded a protobuf message into a `DynamicMessage`, serialized that into a `serde_json::Value` (a `BTreeMap`-backed DOM) so `_meta` and the message key could be merged in, then re-serialized the `Value` to a string. Profiling a live capture showed the `Value` DOM was ~10% of samples; together with the second serialize pass and the allocations backing the tree, it was a significant share of CPU.

This PR addresses that by routing the `DynamicMessage` through `serde_json`'s streaming serializer directly into the output buffer instead. Payload fields stream out as-is, the message key's fields are merged over them, and `_meta` is appended, in a single pass with no intermediate `Value`.

Only the protobuf payload path changes; Avro and JSON payloads are already `serde_json::Value`s and keep the existing merge. Captured documents are unchanged (the Flow runtime normalizes object key order), so there are no snapshot updates.

**Notes for reviewers:**

Documents emitted by the connector through the protobuf path no longer have sorted keys. The runtime sorts keys anyway, so I think that's fine? We're shifting the sorting work from the connector to the runtime, and since the connector is CPU constrained, it seems reasonable to rely on the runtime to sort document keys to free up the connector to do other work.

### DIFF
--- a/source-kafka/src/document.rs
+++ b/source-kafka/src/document.rs
@@ -1,0 +1,442 @@
+//! Streaming serialization of captured documents.
+//!
+//! [`MergeSerializer`] wraps another [`Serializer`]. When the value being
+//! serialized is a JSON object, it forwards the object's fields to the inner
+//! serializer, then appends the Kafka message key's fields and a trailing
+//! `_meta` object.
+//!
+//! This lets `pull.rs` serialize a protobuf `DynamicMessage` straight to output
+//! bytes without first materializing it as a `serde_json::Value`, skipping the
+//! intermediate DOM and the per-field allocations it requires.
+
+use std::marker::PhantomData;
+
+use serde::ser::{Error as _, Impossible, Serialize, SerializeMap, Serializer};
+use serde_json::{Map, Value};
+
+const EXPECT_OBJECT: &str = "captured document must serialize as a JSON object";
+const KEY_NOT_STRING: &str = "captured document field name must be a string";
+
+/// Generates `Serializer` methods that reject the value. Our top-level document
+/// must be a JSON object, so any other kind is an error.
+macro_rules! reject_scalars {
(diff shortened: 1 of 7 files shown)


source-kafka/src/document.rs
```diff
@@ -0,0 +1,474 @@
...
+            for (name, value) in key_fields {
+                if name == "_meta" {
+                    key_has_meta = true;
+                }
+                inner.serialize_entry(name, value)?;
+            }
+        }
+        if !key_has_meta {
+            inner.serialize_entry("_meta", self.meta)?;
+        }
```
Alex-Bair: Like the comment says, this song & dance around merging in our `_meta` field is so we preserve the connector's existing behavior of what `_meta` field makes it into the document (our `_meta` overrides a same named field in the payload but not a same named key field). I'm not convinced that's the desired behavior we actually want, but I didn't want to introduce a behavioral change on top of all the performance-focused changes I've been merging.

williamhbaker (approved):
LGTM

One thing that would be good:  a differential unit test old-path-vs-new over a corpus of real DynamicMessages. Maybe not quite to the extent of a full-blown fuzz test but it would be good to confirm / lock in the equivalency of the two paths, since there's a lot that can subtly differ with something like this (like float fields).

Alex-Bair:
Good point. I added a differential test in `src/document.rs` comparing the old and new paths. It's all equivalent except float32 fields. The old path widened them to f64 (so `0.1` printed as `0.10000000149011612`), and the new path prints the shortest f32 form (`0.1`). I kept the new behavior since it aligns with the official protobuf implementation's JSON serialization.

PR #4737, opened 2026-07-01
estuary-cdk: gate hard-delete reduction behind schema inference

Alex-Bair:
**Regression?**

No.

**Description:**

867d1df76 added an `if op=="d" -> then {reduce: merge, delete: true}` block to every discovered binding so standard-updates materializations issue real DELETEs. On a delete, that merge reduction re-inflates the last-known full record and re-validates it against the connector's schema. If a field has narrowed since that record was written, the older stored value no longer matches and the materialization fails at runtime.

This only bites dataflows whose captures do not use schema inference. Before commit 867d1df76 their deletes reduced with the default `lastWriteWins`, replacing the record with the sparse tombstone and never re-inflating old fields, so a source-side type narrowing was harmless. After it, the same narrowing turns the next delete into a hard crash. Dataflows with captures using schema inference tolerate the narrowing and are unaffected.

This PR fixes this by gating the whole if/then/else block behind `resource.schema_inference`. Non-schema inference resources fall back to the pre-commit behavior: a plain top-level reduction when a `reduction_strategy` is set, or the default `lastWriteWins` otherwise.

Only the four imported Airbyte connectors not using schema inference are affected within this repo. Their discover snapshots are regenerated to drop the if/then/else block.

**Notes for reviewers:**

In practice, the four imported Airbyte connectors are unable to replicate the described failure mode since they never emit documents with `_meta/op = "d"`. But this change to the CDK is still needed regardless so connectors that emit deletions + don't use schema inference (like `source-netsuite` that lives outside this repo) will behave appropriately.

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -495,29 +495,43 @@ def discovered(
         if resource.schema_inference:
             schema["x-infer-schema"] = True
 
-        # Materializations require this if/then/else conditional to support
-        # hard deletes. It tells Flow to use a top-level merge reduction
-        # strategy for deletions and mark them with `delete: True` so materializations
-        # know to issue DELETEs to destinations.
-        #
-        # `_meta` and `_meta.op` are both `required` in the `if` clause because
-        # JSON Schema's `properties` keyword passes when a key is absent.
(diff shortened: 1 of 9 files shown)


nicolaslazo (approved):
LGTM

PR #4739, opened 2026-07-01
source-kafka: log consumer lag every minute

Alex-Bair:
**Regression?**

No.

**Description:**

Lag among a single topic's partitions are rolled up into a single log. Log lines look like:
```jsonl
{"level":"INFO","message":"consumer lag","topic":"events","partitions":1,"partitions_behind":1,"total_messages_behind":50,"max_messages_behind":50,"max_lag_partition":0}
{"level":"INFO","message":"consumer lag","topic":"orders","partitions":3,"partitions_behind":2,"total_messages_behind":854,"max_messages_behind":812,"max_lag_partition":2}
```
We can identify the total number of messages a topic / binding is behind as well as which partition for a topic is lagging the most & by how many messages. I think this level of granularity makes sense, but if you think we should have more granular detail (ex: log stats for each partition so we can determine how far behind we are for each), I'm happy to adjust.

I also added some basic connector status events so it's easier to answer the "is my capture behind & if so, by how many messages?" question by glancing at the dashboard.

### DIFF
--- a/source-kafka/src/configuration.rs
+++ b/source-kafka/src/configuration.rs
@@ -1,9 +1,16 @@
 use anyhow::Result;
 use rdkafka::client::{ClientContext, OAuthToken};
 use rdkafka::consumer::{BaseConsumer, ConsumerContext};
+use rdkafka::statistics::Statistics;
 use rdkafka::ClientConfig;
 use schemars::{schema::RootSchema, JsonSchema};
 use serde::{de, Deserialize, Deserializer, Serialize};
+
+// How often in milliseconds librdkafka computes consumer statistics and
+// delivers them to the `stats` callback, which logs per-partition lag. This is
+// serviced on the same thread that polls for messages, so the interval is kept
+// coarse to keep the logging overhead negligible.
+const STATISTICS_INTERVAL_MS: &str = "300000"; // 5 minutes
 
 #[derive(Serialize, Deserialize)]
 pub struct EndpointConfig {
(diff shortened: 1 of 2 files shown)


source-kafka/src/configuration.rs
```diff
@@ -1,9 +1,23 @@
...
 use rdkafka::ClientConfig;
 use schemars::{schema::RootSchema, JsonSchema};
 use serde::{de, Deserialize, Deserializer, Serialize};
+use std::sync::Mutex;
+
+// How often in milliseconds librdkafka computes consumer statistics and
+// delivers them to the `stats` callback, which logs per-partition lag. This is
+// serviced on the same thread that polls for messages, so the interval is kept
+// coarse to keep the logging overhead negligible.
+const STATISTICS_INTERVAL_MS: &str = "60000"; // 1 minute
```
williamhbaker: 1 minute gets surprisingly spammy...we use 5 minutes elsewhere for things like this.

source-kafka/src/configuration.rs
```diff
@@ -1,9 +1,23 @@
...
+
+// Total messages behind at or below which the capture is considered caught up:
+// the status stays a plain "Streaming messages" with no lag detail appended. A
+// small nonzero value avoids flapping the status on a busy topic's trivially
+// small lag.
+const CAUGHT_UP_THRESHOLD: i64 = 10;
```
williamhbaker: This seems finicky. I'd probably abandon the idea of trying to distinguish "caught up" from "backfilling" and just report what offset we're at vs. the max offset.

source-kafka/src/configuration.rs
```diff
@@ -299,10 +316,129 @@ impl ClientContext for FlowConsumerContext {
...
+        if last_status.as_deref() != Some(status.as_str()) {
+            tracing::info!(eventType = "connectorStatus", "{status}");
+            *last_status = Some(status);
+        }
+    }
+}
+
+// Formats a non-negative count with comma thousands separators, e.g. 4530 ->
+// "4,530", so large lag values in the connector status are easier to read.
+fn with_commas(n: i64) -> String {
```
williamhbaker: I think I would keep it without commas, to maximize machine readability. We generally report huge numbers as-is (stats, etc.) without trying to prettify them elsewhere.

williamhbaker (approved):
LGTM, can't hurt.

I did leave some suggestions for your consideration.

source-kafka/src/configuration.rs
Alex-Bair: Yep, 5 minutes sounds good. I'll update it.

source-kafka/src/configuration.rs
Alex-Bair: Since the `connectorStatus`es are presented in the UI for humans, I figured adding commas would head off user complaints about readability when there are millions of messages to catch up on. But I don't feel strongly about that. I'll remove them.

source-kafka/src/configuration.rs
Alex-Bair: Note for myself: when there are multiple partitions per topic, we're reading different offsets in each partition. There's not a single offset to report per topic, so I report the number of messages we're behind across all partitions of all topics instead.

Issue #4725, opened 2026-06-29
materialize-sql: ClampDatetime rejects valid timestamps that use a space separator instead of "T", crashing the shard


jwhartley:
## 1. Priority
Medium - recurring and triggered by normal operation, not a one-off. Seen across three tenants in a few days; each occurrence crash-loops an entire BigQuery materialization shard (one bad value blocks all bindings). It fires whenever re-inference tags a mixed-format timestamp field as `date-time` (e.g. after a collection reset), so it recurs without special conditions. A clean workaround exists (exclude the offending field and republish), which is why not High.

## 2. Scope and prevalence
Theoretical reach: broad. `ClampDatetime` lives in the shared `materialize-sql` library, so every SQL materialization (BigQuery, Snowflake, Redshift, Databricks, Postgres, etc.) uses it.

Observed: recurring and recent. A 14-day cross-data-plane Loki scan for this signature shows zero occurrences before 2026-06-24, then three unrelated tenants begin within ~3 days (first hits 2026-06-24, 06-26, 06-27), all materializing to BigQuery on the same data plane (`ops/dp/public/gcp-us-central1-c2`). Two different capture sources are involved (a MongoDB capture and Klaviyo captures), so it is not source-specific. Each occurrence crash-loops the whole materialization (one unparseable value fails the shard, blocking every binding).

## 3. Trigger (testable)
A `format: date-time` field receives a value that uses a space separator instead of `T`, e.g. `2025-11-29 01:05:28+00:00` or `2026-06-23 00:15:23.918411+00:00`. The materialization fails when storing that value.

## 4. Root cause (confirmed)
`ClampDatetime` (`materialize-sql/element_converter.go:177-185`) gates every date-time value through strict `time.Parse(time.RFC3339Nano, str)`:

```go
var ClampDatetime ElementConverter = StringCastConverter(func(str string) (interface{}, error) {
	str = strings.Replace(str, "z", "Z", 1)
	if parsed, err := time.Parse(time.RFC3339Nano, str); err != nil {
		return nil, err
	} else if parsed.Year() == 0 {
		return minimumTimestamp, nil
	}
	return str, nil
})
```

Two things make this a converter bug rather than bad data:
1. The converter already normalizes one class of almost-RFC3339 input (it repairs a lowercase trailing `z` to `Z` on the line before the parse). A space-vs-`T` separator is the same class of input and is simply not handled.
[...]

PR #4742, opened 2026-07-02
source-klaviyo: normalize consent timestamps to RFC3339

### LINKED ISSUE #4726: source-klaviyo: consent timestamp fields emitted as non-RFC3339 (space-separated), violating the date-time contract
## 1. Priority
Low - first observed occurrence, rare trigger, customer workaround in place. The companion converter-side fix (#4725) covers existing data on the read/materialize side immediately.

## 2. Scope and prevalence
Theoretical reach: any consumer that respects the `format: date-time` contract. SQL materializations reject the value at conversion time (#4725); non-SQL consumers mishandle it too (Elasticsearch `date` mappings reject it, Iceberg/Parquet timestamp logical types expect a parseable instant, derivations/file consumers inherit the malformed value).
Observed prevalence: low. Surfaced once, on a Klaviyo -> BigQuery pipeline.

## 3. Trigger (testable)
source-klaviyo emits Klaviyo consent timestamps as space-separated strings (e.g. `2026-06-23 00:15:23.918411+00:00`) rather than RFC3339 (`2026-06-23T00:15:23.918411Z`). When schema inference tags these fields `format: date-time`, the collection advertises a contract (date-time = RFC3339) that the values violate.

Affected fields (documented Klaviyo consent timestamps):
- `attributes/properties/$consent_timestamp`
- `attributes/properties/subscriptions/email/marketing/consent_timestamp`
- `attributes/properties/subscriptions/sms/marketing/consent_timestamp`
- `attributes/properties/subscriptions/sms/transactional/consent_timestamp`

## 4. Root cause (likely)
The connector passes Klaviyo's consent timestamp values through as-is, and Klaviyo formats them with a space separator instead of `T`. Confidence: likely (observed in the emitted values; not yet traced to the exact connector code path).

## 5. Investigation / reproduction (testable)
Observed downstream as the materialize-sql conversion failure in #4725 (`cannot parse ... as "T"`). The value in the collection is space-separated; the write schema does not constrain `attributes/properties/*`, so it is accepted, while the inferred read schema carries `format: date-time` from earlier RFC3339 values.

## 6. Possible fixes
Normalize the documented consent timestamp fields to RFC3339 in the connector before emitting, so the collection holds clean RFC3339 for every downstream consumer.

Scope note: `attributes/properties` is a free-form object of Klaviyo custom properties, so the connector cannot generically know which arbitrary props are timestamps. Limit this to the documented consent fields above, not a blanket "normalize all timestamps."

Caveat: existing collection data is already space-separated; a backfill is required to clean historical rows. The companion #4725 converter fix handles existing data on the read side immediately. If the broader "shared normalization layer for CDK/API connectors" direction (see #4725 open question) is adopted, this per-connector fix may be subsumed.

## 7. References
- Slack thread (#integrations): https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1782706029810379
- #4725 - materialize-sql `ClampDatetime` rejects space-separated timestamps; the defensive, cross-destination fix.

jacobmarble:
**Description:**

source-klaviyo passed Klaviyo consent timestamps through as-is, and Klaviyo formats them space-separated (`2026-06-23 00:15:23+00:00`) instead of RFC3339 `T`-separated. Once schema inference tags these `format: date-time`, the collection advertises an RFC3339 contract the values violate, breaking downstream consumers (surfaced as a materialize-sql `cannot parse ... as "T"` failure).

Normalizes the four documented consent fields on the `Profiles` stream via a `map_record` override (`GlobalExclusions` inherits it): `attributes.properties.$consent_timestamp` and `attributes.subscriptions.{email.marketing, sms.marketing, sms.transactional}.consent_timestamp`. Scope is limited to documented fields since `attributes/properties` is free-form. Follows the existing `Events` stream precedent (`.replace(" ", "T")`).

Fixes estuary/connectors#4726.

**Workflow steps:**

No user-facing change. Newly emitted profile records carry RFC3339 consent timestamps. Historical rows remain space-separated until backfilled (companion #4725 handles existing data on the read side).

**Notes for reviewers:**

- The issue listed subscription paths under `attributes/properties/subscriptions/...`; the real structure is `attributes/subscriptions/...` (per `GlobalExclusions` + fixtures) — corrected here.
- Kept the `+00:00` offset rather than converting to `Z`; both are valid RFC3339 and this matches the connector's other timestamps.
- Tests: added a regression test in `TestProfilesStream`; `poetry run pytest tests/test_streams.py` → 47 passed. Discover/spec snapshots unaffected.

### DIFF
--- a/source-klaviyo/source_klaviyo/streams.py
+++ b/source-klaviyo/source_klaviyo/streams.py
@@ -34,9 +34,9 @@ LAG = 110
 # else untouched. The anchored regex is what keeps free-text values (e.g. "May 5, 2021")
 # safe from mangling.
 #
-# Matches "YYYY-MM-DD HH:MM:SS[.ffffff][ ](Z | +HH:MM | +HHMM)".
+# Matches the observed Klaviyo format "YYYY-MM-DD HH:MM:SS[.ffffff]+HH:MM".
 _SPACE_SEPARATED_DATETIME_RE = re.compile(
-    r"^(\d{4}-\d{2}-\d{2}) (\d{2}:\d{2}:\d{2}(?:\.\d+)?) ?(Z|[+-]\d{2}:?\d{2})$"
+    r"^(\d{4}-\d{2}-\d{2}) (\d{2}:\d{2}:\d{2}(?:\.\d+)?)([+-]\d{2}:\d{2})$"
 )
 
 
@@ -45,8 +45,6 @@ def _normalize_datetime(value: str) -> str:
     if not match:
         return value
     date, time, tz = match.groups()
-    if tz != "Z" and ":" not in tz:
-        tz = f"{tz[:3]}:{tz[3:]}"  # +0000 -> +00:00
     return f"{date}T{time}{tz}"
 
 
--- a/source-klaviyo/source_klaviyo/streams.py
+++ b/source-klaviyo/source_klaviyo/streams.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
+import re
 import urllib.parse
 from abc import ABC, abstractmethod
 from datetime import datetime, timedelta, UTC
@@ -24,12 +25,46 @@ from .exceptions import KlaviyoBackoffError
 LAG = 110
 
 
-def _normalize_rfc3339(value: Any) -> Any:
-    # Klaviyo emits consent timestamps space-separated (e.g. "2026-06-23 00:15:23+00:00"),
-    # but the `format: date-time` contract requires the RFC3339 "T" separator.
-    if isinstance(value, str) and " " in value:
-        return value.replace(" ", "T")
-    return value
+# Klaviyo emits some datetime fields (e.g. consent timestamps) space-separated
+# (e.g. "2026-06-23 00:15:23.918411+00:00") rather than RFC3339. When schema inference
+# tags such a field `format: date-time`, the collection advertises an RFC3339 contract
+# the value violates, breaking downstream consumers. Rather than enumerate the known
+# offenders (Klaviyo custom properties are free-form, so new ones surface unpredictably),
+# we normalize any string that matches the space-separated shape and leave everything
+# else untouched. The anchored regex is what keeps free-text values (e.g. "May 5, 2021")
+# safe from mangling.
+#
+# Matches "YYYY-MM-DD HH:MM:SS[.ffffff][ ](Z | +HH:MM | +HHMM)".
+_SPACE_SEPARATED_DATETIME_RE = re.compile(
+    r"^(\d{4}-\d{2}-\d{2}) (\d{2}:\d{2}:\d{2}(?:\.\d+)?) ?(Z|[+-]\d{2}:?\d{2})$"
+)
+
+
+def _normalize_datetime(value: str) -> str:
+    match = _SPACE_SEPARATED_DATETIME_RE.match(value)
+    if not match:
+        return value
+    date, time, tz = match.groups()
+    if tz != "Z" and ":" not in tz:
+        tz = f"{tz[:3]}:{tz[3:]}"  # +0000 -> +00:00
+    return f"{date}T{time}{tz}"
+
+
+def _normalize_datetimes(data: Any) -> None:
+    """Recursively normalize space-separated datetime strings in place."""
+    if isinstance(data, dict):
+        for key, value in data.items():
+            if isinstance(value, str):
+                data[key] = _normalize_datetime(value)
+            elif isinstance(value, (dict, list)):
+                _normalize_datetimes(value)
+    elif isinstance(data, list):
+        for i, item in enumerate(data):
+            if isinstance(item, str):
+                data[i] = _normalize_datetime(item)
+            elif isinstance(item, (dict, list)):
+                _normalize_datetimes(item)
+
 
 class KlaviyoStream(HttpStream, ABC):
     """Base stream for api version v2023-10-15"""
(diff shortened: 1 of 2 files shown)


source-klaviyo/source_klaviyo/streams.py
```diff
@@ -331,6 +339,31 @@ def request_params(
...
+
+        # Klaviyo returns the documented consent timestamps space-separated rather than
+        # RFC3339. We only normalize these known fields: `attributes/properties` is a
+        # free-form custom-property object, so we can't generically tell which arbitrary
+        # props are timestamps.
+        attributes = record.get("attributes") or {}
+
+        properties = attributes.get("properties")
+        if isinstance(properties, dict) and "$consent_timestamp" in properties:
+            properties["$consent_timestamp"] = _normalize_rfc3339(properties["$consent_timestamp"])
```
Alex-Bair: This works for the current issue, but I'm 99.99% sure there are other properties other than `$consent_timestamp` that would benefit from this normalization too. Instead of narrowing in on just `$consent_timestamp` then playing whack-a-mole whenever another property pops up that's an issue, I'd suggest recursively checking all `attributes` & normalizing those that look like they're almost compliant datetime strings.

Determining "is this a datetime string?" for an arbitrary field can be tricky, but we've used regexes elsewhere to do this type of thing when we know a source system doesn't naturally format datetimes like we want. You could take some inspiration from how we do that in `source-iterable-native` here. You may need to adapt the regex to match against how Klaviyo formats datetime strings:
https://github.com/estuary/connectors/blob/589b85e6420fd3af9ab92f514552fcc6d950083e/source-iterable-native/source_iterable_native/models.py#L262-L314

(Note: the recursive check + replace would incur a performance hit, but `source-klaviyo` isn't CPU constrained, is a deprecated connector, and any users concerned about connector performance should be using `source-klaviyo-native` anyway since it's much faster).

source-klaviyo/source_klaviyo/streams.py
```diff
@@ -23,6 +23,14 @@
 # we only ask for data that older than LAG minutes in the `events` stream.
 LAG = 110
 
+
+def _normalize_rfc3339(value: Any) -> Any:
+    # Klaviyo emits consent timestamps space-separated (e.g. "2026-06-23 00:15:23+00:00"),
+    # but the `format: date-time` contract requires the RFC3339 "T" separator.
+    if isinstance(value, str) and " " in value:
+        return value.replace(" ", "T")
+    return value
```
Alex-Bair: note: Addressing my other [comment](https://github.com/estuary/connectors/pull/4742/changes#r3516150128) about normalizing fields only when they match against specific regexes will address what I'm about to say, but I wanted to mention another concern here too with the strategy as-is.

Right now, `_normalize_rfc3339` is only invoked for the `$consent_timestamp` field. We're assuming `$consent_timestamp` is _always_ formatted like `2026-06-23 00:15:23.918411+00:00`  so the strategy of "replace any spaces with a T" works. But I don't have 100% confidence that `$consent_timestamp` is always formatted like that. SaaS systems can have a surprising amount of customizability, and it wouldn't surprise me if `$consent_timestamp` was some custom field whose contents could differ between Klaviyo accounts. For example, if it is a custom field & the contents were strings like `2026-06-23 00:15:23 +00:00` or `"May 5, 2021"`, `_normalize_rfc3339` would transform it into `2026-06-23T00:15:23T+00:00` or `"MayT5,T2021"`. We've seen SaaS sources exhibit lots of unusual behavior & I wouldn't be surprised if the scenario I described is possible.

Alex-Bair (changes requested):
It'd be nice to extend the normalization to any timestamp looking field within `properties` so we don't need to play whack-a-mole later whenever we discover another one that's not formatted like we need them to be to maintain `format: date-time` in the inferred schema.

Additionally for myself later, I'm curious if `source-klaviyo-native` is already performing an equivalent normalization. I doubt it, so the capture team should look into updating that connector's handling too.

source-klaviyo/source_klaviyo/streams.py
jacobmarble: Good call — switched to exactly this. Added an anchored regex (`_SPACE_SEPARATED_DATETIME_RE`) plus a recursive in-place walk (`_normalize_datetimes`) over `attributes`, modeled on `source-iterable-native`'s `_fix_iterable_datetime` / `_fix_iterable_datetimes`. `Profiles.map_record` now just calls the walker instead of poking at named fields, so any space-separated datetime is normalized, not just `$consent_timestamp`.

I kept the regex scoped to the one format we've actually observed from Klaviyo — `YYYY-MM-DD HH:MM:SS[.ffffff]+HH:MM` (space separator, optional fractional seconds, colon offset). Happy to broaden it (e.g. `Z`, colon-less offsets) if/when we see those from Klaviyo, but didn't want to assert formats we haven't confirmed.

source-klaviyo/source_klaviyo/streams.py
jacobmarble: Agreed, and the regex approach resolves this. Normalization only fires when a string matches the full anchored datetime pattern, so a custom/free-text value like `"May 5, 2021"` — or a differently-shaped timestamp like your `2026-06-23 00:15:23 +00:00` example — simply doesn't match and is passed through **untouched** rather than mangled into `"MayT5,T2021"` / `"2026-06-23T00:15:23T+00:00"`. The regression test asserts exactly that: free-text and already-compliant values are unchanged, and only the observed space-separated form is rewritten.

Alex-Bair (approved):
LGTM, thanks!

PR #4743, opened 2026-07-02
materialize-sql: tolerate near-RFC3339 timestamps in datetime converters

### LINKED ISSUE #4725: materialize-sql: ClampDatetime rejects valid timestamps that use a space separator instead of "T", crashing the shard

jacobmarble:
**Description:**

Fixes #4725. `sql.ClampDatetime` validated `format: date-time` values with a strict `time.Parse(time.RFC3339Nano, ...)`, so a near-RFC3339 value (space separator instead of `T`, missing offset) crash-looped the whole materialization shard when schema inference tagged a mixed-format field as `date-time`.

- New `sql.ParseFlexibleTimestamp` accepts RFC3339 plus the variants flow's file-parser datetime sanitizer already tolerates (space separator, lowercase `z`, missing offset assumed UTC), emitting canonical RFC3339. Already-valid inputs pass through byte-identical.
- `ClampDatetime` / `NormalizeDatetimeString` / `ClampDate` use it; `NormalizeDatetimeString` still passes unparseable values through rather than newly erroring. `ClampDate` truncates datetime values to their date part.
- Same bug pattern fixed in the connector-local clones: materialize-clickhouse, materialize-mysql (SingleStore clamps and `rfc3339ToTZ`, which had the identical strict parse for MySQL/MariaDB DATETIME columns), and materialize-s3-iceberg.

**Workflow steps:**

No user-facing change; previously-rejected near-RFC3339 timestamps now materialize as normalized RFC3339.

None.

**Notes for reviewers:**

Tests-first: commit 1 adds failing regression tests (reproducing the exact error from the issue), commit 2 the shared fix, then one commit per connector clone. Unit + docker integration suites pass for all touched packages.

### DIFF
--- a/materialize-sql/element_converter.go
+++ b/materialize-sql/element_converter.go
@@ -166,16 +166,14 @@ const (
 	minimumDate      = "0001-01-01"
 )
 
-// flexibleTimestampLayouts are the near-RFC3339 variants accepted in addition to RFC3339Nano:
-// a space separator instead of "T", and/or a missing offset (assumed UTC). Flow's schema
-// inference can tag a field as `format: date-time` based on some values while other values in
-// the same field use these variants, so rejecting them crash-loops otherwise-healthy shards.
-// The set of layouts is aligned with flow's file-parser datetime sanitizer
-// (crates/parser/src/format/sanitize/datetime.rs), which normalizes the same variants.
+// flexibleTimestampLayouts are the near-RFC3339 variants accepted in addition to RFC3339Nano.
+// Flow's schema inference can tag a field as `format: date-time` based on some values while
+// other values in the same field use these variants, so rejecting them crash-loops
+// otherwise-healthy shards. Only variants that fully specify the instant are accepted: a
+// timestamp without a timezone offset may be local time, and assuming a zone for it would
+// silently shift the data.
 var flexibleTimestampLayouts = []string{
 	"2006-01-02 15:04:05.999999999Z07:00",
-	"2006-01-02T15:04:05.999999999",
-	"2006-01-02 15:04:05.999999999",
 }
 
 // ParseRFC3339Nano parses str as RFC3339 or a near-RFC3339 variant, returning the parsed
@@ -194,8 +192,8 @@ func ParseRFC3339Nano(str string) (time.Time, string, error) {
 	return time.Time{}, "", fmt.Errorf("could not parse %q as RFC3339 date-time", str)
 }
 
-// NormalizeDatetimeString rewrites near-RFC3339 datetimes (lowercase "z", space separator,
-// missing offset) to canonical RFC3339. Values it cannot parse are passed through unmodified,
+// NormalizeDatetimeString rewrites near-RFC3339 datetimes (lowercase "z", space separator)
+// to canonical RFC3339. Values it cannot parse are passed through unmodified,
 // since this converter has always been a pass-through and endpoints may accept formats it does
 // not recognize.
 var NormalizeDatetimeString ElementConverter = StringCastConverter(func(str string) (interface{}, error) {
--- a/materialize-sql/element_converter.go
+++ b/materialize-sql/element_converter.go
@@ -166,22 +166,55 @@ const (
 	minimumDate      = "0001-01-01"
 )
 
-// NormalizeDatetimeString replaces a lowercase trailing "z" with uppercase "Z" to ensure
-// RFC3339 compliance.
+// flexibleTimestampLayouts are the near-RFC3339 variants accepted in addition to RFC3339Nano:
+// a space separator instead of "T", and/or a missing offset (assumed UTC). Flow's schema
+// inference can tag a field as `format: date-time` based on some values while other values in
+// the same field use these variants, so rejecting them crash-loops otherwise-healthy shards.
+// The set of layouts is aligned with flow's file-parser datetime sanitizer
+// (crates/parser/src/format/sanitize/datetime.rs), which normalizes the same variants.
+var flexibleTimestampLayouts = []string{
+	"2006-01-02 15:04:05.999999999Z07:00",
+	"2006-01-02T15:04:05.999999999",
+	"2006-01-02 15:04:05.999999999",
+}
+
+// ParseRFC3339Nano parses str as RFC3339 or a near-RFC3339 variant, returning the parsed
+// time and its canonical RFC3339 rendering. Inputs that are already valid RFC3339 are returned
+// byte-identical (after repairing a lowercase "z"); only fallback layouts are reformatted.
+func ParseRFC3339Nano(str string) (time.Time, string, error) {
+	str = strings.Replace(str, "z", "Z", 1)
+	if parsed, err := time.Parse(time.RFC3339Nano, str); err == nil {
+		return parsed, str, nil
+	}
+	for _, layout := range flexibleTimestampLayouts {
+		if parsed, err := time.Parse(layout, str); err == nil {
+			return parsed, parsed.Format(time.RFC3339Nano), nil
+		}
+	}
+	return time.Time{}, "", fmt.Errorf("could not parse %q as RFC3339 date-time", str)
+}
+
+// NormalizeDatetimeString rewrites near-RFC3339 datetimes (lowercase "z", space separator,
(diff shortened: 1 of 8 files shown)


jacobmarble:
One gap - this doesn't touch Snowflake or Redshift, they've got their own copy of the same strict-RFC3339 parse (`materialize-snowflake/sqlgen.go:50` `rfc3339ToUTCString()`, `materialize-redshift/sqlgen.go:36` inline in `datetimeMapping`). Same fix pattern as clickhouse/mysql/s3-iceberg here would close them too.

jacobmarble:
We won't address these extra cases. The intention is to parse incoming data strictly — the flexible parsing added here is only a bandaid to handle data that already exists in collections. A future data-plane change will reject such data going forward, so the scope stays limited to the converters implicated in the documented issue.

materialize-sql/element_converter.go
```diff
@@ -166,22 +166,55 @@ const (
...
+// flexibleTimestampLayouts are the near-RFC3339 variants accepted in addition to RFC3339Nano:
+// a space separator instead of "T", and/or a missing offset (assumed UTC). Flow's schema
+// inference can tag a field as `format: date-time` based on some values while other values in
+// the same field use these variants, so rejecting them crash-loops otherwise-healthy shards.
+// The set of layouts is aligned with flow's file-parser datetime sanitizer
+// (crates/parser/src/format/sanitize/datetime.rs), which normalizes the same variants.
+var flexibleTimestampLayouts = []string{
+	"2006-01-02 15:04:05.999999999Z07:00",
+	"2006-01-02T15:04:05.999999999",
+	"2006-01-02 15:04:05.999999999",
```
danielnelson: Why accept without timezone?  I feel like this has more risk because it is possible they time is in local time.

danielnelson (approved):

materialize-sql/element_converter.go
jacobmarble: Good point — agreed. A timestamp with an explicit offset (even space-separated) fully specifies the instant, but one without an offset may be local time, and assuming UTC would silently shift the data. That's a worse failure mode than the crash this PR fixes, and every observed value in #4725 carried an explicit offset anyway. Dropped the offset-less layouts in 372f34d5c: only the space-separator variant with an explicit Z/offset is accepted now, and offset-less timestamps are rejected as before.

PR #4745, opened 2026-07-02
materialize-clickhouse: retry transient connection drops during Load/Store

### LINKED ISSUE #4744: materialize-clickhouse: retry transient connection drops (read: EOF) during Load/Store
## Context

`foo/mysql-cdc-prod/clickhouse-direct` (ClickHouse Cloud target) has been
flapping with shard failures caused by the ClickHouse connection dropping
mid-query:

```
transactor.Load: preparing load batch: query processing: failed to read first block packet from <ip>:9440 (conn_id=N): read: EOF
transactor.Store: preparing store batch: query processing: failed to read first block packet from <ip>:9440 (conn_id=N): read: EOF
```

#4724 (per-binding Load queries) fixed the original cause on the Load JOIN
path (code 241 OvercommitTracker kills from a combined `UNION ALL` across all
bindings). Deployed and confirmed working: zero Load-path `code 241` since.

However, EOF failures continue on both Load and Store, now attributed to the
ClickHouse server itself (replica OOM/recycle under memory pressure, likely
from background `ReplacingMergeTree` merge/compaction — unconfirmed, customer
CH-side `system.query_log`/`system.part_log` for the incident window were
lost to a replica recycle before they could be pulled). This is tracked
separately as a ClickHouse capacity issue.

Each occurrence currently causes a full shard failure + restart (crash-loop
style recovery), which is expensive and generates alert noise for what is
often a transient, self-healing blip (the replica recovers within seconds to
minutes).

## Ask

Add a bounded retry/backoff around the ClickHouse operations that see these
transient connection drops, so a single blip doesn't take down the whole
shard:

- `materialize-clickhouse/driver.go:415` — `preparing load batch` (`PrepareBatch`
  for the load-keys insert)
- `materialize-clickhouse/driver.go:507` / `:526` — `querying Load documents for
  %s` (the per-binding Load JOIN query, post-#4724)
- `materialize-clickhouse/driver.go:553` — `preparing store batch` (`PrepareBatch`
  for the store insert)
- `materialize-clickhouse/driver.go:769` (`moveStorePartitionsToTarget`) — `Exec`
  of `MOVE PARTITION`

## Why these points are safe to retry

- `PrepareBatch` fails while reading the first block packet, i.e. before any
  rows have been sent — retrying is a clean no-op on the ClickHouse side.
- `MOVE PARTITION` is already guarded by the existing hard row-count checks in
  `moveStorePartitionsToTarget` (pre- and post-move counts must reconcile), so
  a retry after a transient failure is safe; a genuinely stuck/partial move
  still surfaces as a hard error rather than silently double-applying.
- `clickhouse.Open` returns a pooled `Conn`, so a retry naturally dials a fresh
  connection — on ClickHouse Cloud's load-balanced endpoint this can land on a
  healthy replica.

## Design notes

- Bound both attempt count and total elapsed time; after exhausting retries,
  fail as today (let the shard restart) rather than retrying forever against a
  persistently unhealthy cluster.
- Only retry a well-defined class of transient errors (connection reset/EOF/
  i/o timeout on the native protocol) — do not blanket-retry all ClickHouse
  errors (e.g. don't retry `code: 241` OOM kills or schema/DDL errors; those
  need the query itself to change, not a retry).
- No existing retry/backoff helper exists in this connector or in
  `materialize-sql` today — this will likely need a small new helper.

## Evidence / references

- Fix for the original Load-path cause: https://github.com/estuary/connectors/pull/4724
- Task: `foo/mysql-cdc-prod/clickhouse-direct`, ClickHouse Cloud service
  `rustaws-bc-61`, target `estuary_direct_prod`.

jacobmarble:
**Description:**

Closes #4744. Adds bounded retry/backoff around ClickHouse operations that fail on transient connection drops (`read: EOF`, resets, i/o timeouts) when a ClickHouse Cloud replica is recycled, so a blip no longer causes a full shard failure.

- New `retryPolicy` helper: max 8 attempts, exponential backoff 1s→30s, 5 min elapsed cap; retries only connection-level errors — never server exceptions (e.g. code 241 OOM) or context cancellation.
- Applied to Load/Store `PrepareBatch` (fails before rows are sent), the per-binding Load JOIN query, and the `MOVE PARTITION` exec (idempotent; guarded by existing row-accounting checks).
- The Load query retry stops once any document has been emitted, since re-delivery would duplicate loaded documents.
- `Batch.Send` is not retried: a drop mid-send leaves the insert outcome unknown.

**Workflow steps:**

No user-facing change; existing materializations pick this up on upgrade.

**Notes for reviewers:**

Retry loop and error classification are unit-tested (`retry_test.go`), including the exact wrapped-`io.EOF` shape clickhouse-go produces. Full suite incl. `TestIntegration/*` passes locally.

### DIFF
--- a/materialize-clickhouse/retry.go
+++ b/materialize-clickhouse/retry.go
@@ -1,0 +1,94 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net"
+	"syscall"
+	"time"
+
+	clickhouseproto "github.com/ClickHouse/clickhouse-go/v2/lib/proto"
+	log "github.com/sirupsen/logrus"
+)
+
+// retryPolicy bounds retries of transient ClickHouse failures by both attempt
+// count and total elapsed time. After the bounds are exhausted the last error
+// is returned and the shard fails as it would without retries, so a
+// persistently unhealthy cluster still surfaces promptly.
+type retryPolicy struct {
+	maxAttempts    int
+	initialBackoff time.Duration
+	maxBackoff     time.Duration
+	maxElapsed     time.Duration
+}
+
+// transientRetryPolicy rides out a ClickHouse Cloud replica recycle, which
+// typically resolves within seconds to a couple of minutes.
+var transientRetryPolicy = retryPolicy{
+	maxAttempts:    8,
+	initialBackoff: time.Second,
+	maxBackoff:     30 * time.Second,
+	maxElapsed:     5 * time.Minute,
+}
+
+// retry invokes op, retrying with exponential backoff while retryable reports
+// the error as worth retrying. Because the underlying clickhouse.Conn is a
+// pool that discards broken connections, each retry naturally dials fresh --
+// on a load-balanced endpoint this can land on a healthy replica.
+func (p retryPolicy) retry(ctx context.Context, desc string, retryable func(error) bool, op func() error) error {
+	var deadline = time.Now().Add(p.maxElapsed)
+	var backoff = p.initialBackoff
+
+	for attempt := 1; ; attempt++ {
+		var err = op()
+		if err == nil || !retryable(err) {
+			return err
+		}
+		if attempt >= p.maxAttempts || !time.Now().Add(backoff).Before(deadline) {
+			return err
+		}
+
+		log.WithFields(log.Fields{
+			"op":      desc,
+			"attempt": attempt,
+			"backoff": backoff.String(),
+			"error":   err.Error(),
+		}).Warn("retrying transient ClickHouse error")
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(backoff):
+		}
+		if backoff *= 2; backoff > p.maxBackoff {
+			backoff = p.maxBackoff
+		}
+	}
+}
+
+// isTransientErr reports whether err is a transient connection-level failure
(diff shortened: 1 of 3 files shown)


materialize-clickhouse/retry.go
```diff
@@ -0,0 +1,94 @@
...
+
+// retryPolicy bounds retries of transient ClickHouse failures by both attempt
+// count and total elapsed time. After the bounds are exhausted the last error
+// is returned and the shard fails as it would without retries, so a
+// persistently unhealthy cluster still surfaces promptly.
+type retryPolicy struct {
+	maxAttempts    int
+	initialBackoff time.Duration
+	maxBackoff     time.Duration
+	maxElapsed     time.Duration
```
danielnelson, line 23: I like the `retryPolicy` type, might add something like this to HubSpot PR.  Thinking about `maxAttempts` vs `maxElapsed`, they seem similar in that they both limit the max time spent retrying.  Do you think it makes sense to have both?

danielnelson (approved):

materialize-clickhouse/retry.go
jacobmarble: Because this is single-use/hard-coded, no, I don't think it makes sense to have both. However, if it's going to be reused, then it's best to have both.

PR #4747, opened 2026-07-02
source-monday: fetch `status` of `subitems` in the `items` stream separately to avoid 500 errors

Alex-Bair:
**Regression?**

No.

**Description:**

Monday.com can return an `INTERNAL_SERVER_ERROR` when `state` is requested on subitems through the nested `subitems` selection, which fails the whole `items_page`/`next_items_page` request and stalls the `items` backfill on any board with an affected subitem.

This PR fixes that by removing `state` from the shared item fragment so subitems no longer request it and instead populate it from a top-level `items(ids:)` query, which returns subitem `state` without error. This runs for both the backfill and incremental item fetches and leaves the emitted document shape unchanged.

I also included some additional guidance in the top-level `CLAUDE.md` on working with GraphQL APIs & added the `pytest-asyncio` dep to `source-monday` so the tests I added run in our CI checks.


nicolaslazo (approved):
LGTM

PR #4748, opened 2026-07-02
estuary-cdk: time out of long-running OAuth token exchanges

nicolaslazo:
**Description:**
We found that `source-quickbooks` OAuth token exchanges can get stuck on infinite 429 backoff loops until the control plane kills our `Validate` RPCs. Since we don't expect rotations to take more than 10 minutes, it's safe to error out early and provide a more specific error instead.

This commit also logs warnings at regular intervals when specific function calls take longer than expected.

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
+++ b/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
@@ -24,7 +24,12 @@ from ..flow import (
 )
 from ..http import HTTPMixin, TokenSource
 from ..logger import FlowLogger
-from ..utils import format_error_message, get_running_tasks_info, sort_dict
+from ..utils import (
+    format_error_message,
+    get_running_tasks_info,
+    sort_dict,
+    surface_slow,
+)
 from . import Request, Response, Task, request, response
 from ._emit import emit_bytes
 from .common import _ConnectorState
@@ -39,6 +44,8 @@ PERIODIC_RESTART_INTERVAL = 24 * 60 * 60  # 24 hours
 GRACEFUL_SHUTDOWN_TIMEOUT = 30 * 60  # 30 minutes
 TASK_CANCELLATION_TIMEOUT = 5 * 60  # 5 minutes
 WEBHOOK_SHUTDOWN_TIMEOUT = 60  # 1 minute
+
+OAUTH2_ROTATION_DEADLINE = 10 * 60  # 10 minutes
 
 
 class TerminateTaskGroup(Exception):
@@ -346,11 +353,16 @@ class BaseCaptureConnector(
         if access_token_expiration < now + timedelta(minutes=5):
             # Exchange for new access token & refresh token.
             log.info("Attempting to rotate OAuth2 tokens.")
-            token_exchange_response = await self.token_source.initialize_oauth2_tokens(
-                log, self
+            token_exchange_response = await surface_slow(
+                log,
+                "OAuth2 token exchange",
+                self.token_source.initialize_oauth2_tokens(log, self),
+                deadline=OAUTH2_ROTATION_DEADLINE,
             )
 
-            # Replace tokens in the config.
+            # Replace tokens in the config. From here until the control plane
+            # publishes the emitted config update, these in-memory tokens are
+            # the only valid copy.
             credentials = self.token_source.credentials
             credentials.access_token = token_exchange_response.access_token
             credentials.refresh_token = token_exchange_response.refresh_token
--- a/estuary-cdk/estuary_cdk/http.py
+++ b/estuary-cdk/estuary_cdk/http.py
@@ -34,6 +34,7 @@ from .utils import format_error_message
 
 DEFAULT_AUTHORIZATION_HEADER = "Authorization"
 DEFAULT_AUTHORIZATION_TOKEN_TYPE = "Bearer"
+OAUTH2_TOKEN_EXCHANGE_TIMEOUT = 30.0  # seconds, per attempt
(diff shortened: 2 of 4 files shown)


estuary-cdk/estuary_cdk/capture/base_capture_connector.py
```diff
@@ -346,11 +353,16 @@ async def _rotate_oauth2_tokens(
...
+                log,
+                "OAuth2 token exchange",
+                self.token_source.initialize_oauth2_tokens(log, self),
+                deadline=OAUTH2_ROTATION_DEADLINE,
             )
 
-            # Replace tokens in the config.
+            # Replace tokens in the config. From here until the control plane
+            # publishes the emitted config update, these in-memory tokens are
+            # the only valid copy.
```
Alex-Bair, line 365: > From here until the control plane publishes the emitted config update, these in-memory tokens are the only valid copy.

Does this comment mean "the only valid copy within the running connector instance"? The row in Supabase's `config_updates` table plus the `configUpdate` connector status log should also contains the same valid, encrypted credentials.

estuary-cdk/estuary_cdk/utils.py
```diff
@@ -1,13 +1,70 @@
...
+            return await asyncio.wait_for(asyncio.shield(task), timeout=next_warn)
+        except asyncio.CancelledError:
+            # Our caller was cancelled; don't leak the shielded task.
+            _ = task.cancel()
+            raise
+        except TimeoutError:
+            if task.done():
+                # The awaitable itself raised TimeoutError, as opposed to our
+                # polling timeout having elapsed; propagate its outcome.
+                return task.result()
```
Alex-Bair, line 47: [`asyncio.wait_for`](https://docs.python.org/3.11/library/asyncio-task.html#asyncio.wait_for) attempts to cancel the passed in awaitable once the timeout elapses, but `asyncio.shield` is being used to prevent the awaitable from being cancelled. Is there a way we could simplify this by using `asyncio.wait` instead of `asyncio.wait_for`? [`asyncio.wait`](https://docs.python.org/3.11/library/asyncio-task.html#asyncio.wait) doesn't attempt to cancel the passed in awaitable, which is the behavior we want here.

```suggestion
        try:
            done, _ = await asyncio.wait({task}, timeout=next_warn)
        except asyncio.CancelledError:
            # asyncio.wait leaves the task running on cancellation; our caller
            # was cancelled, so cancel the task rather than leaking it.
            _ = task.cancel()
            raise

        if done:
            # Propagates the awaitable's value or its exception.
            return task.result()
```

estuary-cdk/estuary_cdk/utils.py
```diff
@@ -1,13 +1,70 @@
...
+    not be cancelled once started, such as work following an irreversible
+    side effect.
+    """
+    task = asyncio.ensure_future(aw)
+    started = time.monotonic()
+    next_warn = warn_after
+
+    while True:
+        try:
+            return await asyncio.wait_for(asyncio.shield(task), timeout=next_warn)
```
Alex-Bair, line 38: Should we stop waiting if the deadline has elapsed before `next_warn`?
```suggestion
        timeout = next_warn
        if deadline is not None:
            timeout = min(next_warn, deadline - (time.monotonic() - started))
        try:
            return await asyncio.wait_for(asyncio.shield(task), timeout=timeout)

```

estuary-cdk/estuary_cdk/http.py
```diff
@@ -432,6 +433,10 @@ async def _fetch_oauth2_token(
             headers=headers,
             form=form,
             with_token=False,
+            # Bound each attempt and don't retry server errors forever, so that
+            # a misbehaving provider surfaces as a loggable failure.
+            should_retry=lambda status, headers, body, attempt: attempt < 5,
+            timeout=aiohttp.ClientTimeout(total=OAUTH2_TOKEN_EXCHANGE_TIMEOUT),
```
Alex-Bair, line 439: I'm not aligned with lowering the total timeout from `aiohttp`'s default of 5 minutes to 30 seconds. The scenario I'm concerned about is the refresh token is single use & it takes over 30 seconds for the connector to receive a response (due to slow processing on the provider side, the network being inherently unreliable, or a combination of both). The source system would receive & consume the single use refresh token but the connector stops waiting for the response after 30 seconds & makes another separate request with the now invalidated refresh token.

We're still vulnerable to this scenario when we're relying on `aiohttp`'s 5 minute default total timeout, but I'd anticipate the scenario I'm concerned about is much less likely to occur with a 5 minute timeout than a 30 second one.

estuary-cdk/estuary_cdk/utils.py
```diff
@@ -1,13 +1,70 @@
...
+from collections.abc import Awaitable, Mapping
 from dataclasses import dataclass
-from typing import Any
+from logging import Logger
+from typing import Any, TypeVar
+
+T = TypeVar("T")
+
+
+async def surface_slow(
```
Alex-Bair, line 12: nit: `surface_slow` doesn't immediately make me realize this has a potential side effect of cancelling the `aw` if `deadline` is exceeded. Could we rename this function so it's clearer that the `aw` may get cancelled? Maybe something like `run_with_deadline`?

Alex-Bair (commented):
There's some additional scope in this PR beyond what's stated in the description (meaning, the changes in `http.py`). I don't think those changes in `http.py` are needed right now, and I think we could take them out while still achieving the intent of the PR.

PR #4750, opened 2026-07-02
parquet writer memory improvements

mdibaiee:
**Description:**

- A bug in arrow-go was causing a memory leak, making our connector OOM pretty quickly when row groups were above a certain size, see https://github.com/apache/arrow-go/pull/885
- This bug has been fixed upstream. This pull-request switches to this latest version of arrow-go.
- Moreover, I am adding a new `bufferSize` advanced knob that is necessary for certain sizes of records
- Adds a new `dropPagecache` function which asks the kernel to clean up the page cache of scratch files we are done with. In my local tests this is useful for preventing the cgroup memory usage metric from blowing up past `max` and getting OOM killed

**Workflow steps:**

**Notes for reviewers:**


go/writer/parquet.go
```diff
@@ -257,43 +267,24 @@ func writerOpts(cfg parquetConfig) []file.WriteOption {
...
-
 	w.buffer = append(w.buffer, row)
 	w.bufferSizeBytes += w.rs.estSize(row)
-	w.scratch.rowCount += 1
 
-	if w.bufferSizeBytes >= maxBufferSize {
-		// Write out the buffer as a single row group to the scratch file.
+	if w.bufferSizeBytes >= w.cfg.bufferSize {
+		// Write out the buffer as a single row group, to the scratch file or directly to the
+		// output (see flushBuffer).
```
jacobmarble: Is it always to the scratch file, or is it really possible to write directly to the output?

go/writer/parquet.go
```diff
@@ -257,43 +267,24 @@ func writerOpts(cfg parquetConfig) []file.WriteOption {
...
 		if err := w.flushBuffer(); err != nil {
 			return fmt.Errorf("write flushing buffer based on buffer size: %w", err)
 		}
 	}
 
-	if w.scratch.sizeBytes >= w.cfg.rowGroupByteLimit ||
+	if w.scratch.writer != nil && (w.scratch.sizeBytes >= w.cfg.rowGroupByteLimit ||
 		w.scratch.rowCount >= w.cfg.rowGroupRowLimit ||
```
jacobmarble, line 286: The value of `rowCount` is checked on every call to `Write()`, but incremented only in batches. Perhaps better to increment just before `Write()` returns nil?

go/writer/parquet.go
```diff
@@ -354,13 +345,29 @@ func (w *ParquetWriter) flushScratchFile() error {
...
+	if err := transferColumnValues(scratchReader, w.sinkWriter); err != nil {
 		return fmt.Errorf("transferring column values: %w", err)
-	} else if err := sr.Close(); err != nil {
+	}
+
+	// The scratch file's page cache is charged against the process's cgroup memory limit like
+	// heap memory is, and the kernel won't necessarily reclaim it before it's needed elsewhere.
+	// Drop it explicitly now that its contents have been transferred to the sink writer and the
+	// file is about to be removed.
+	dropPageCache(sr)
```
jacobmarble, line 368: Claude is convinced that this incorrect, and that further performance gains are available by using `fadvise` a little differently. I'll investigate and prove with a benchmark test.

jacobmarble:
I pushed two commits to this branch that measure and extend the page-cache work here:

- e198fc11c `go/writer: add scratch file page cache residency benchmark`
- 501b7c93a `go/writer: sync and drop scratch file page cache incrementally`

**Why:** the `dropPageCache` call in `flushScratchFile` runs at the one point where it can't help. `FADV_DONTNEED` skips dirty pages, and the scratch file is never synced, so most of its pages are still dirty when the hint fires — and the `os.Remove` two statements later evicts the inode's entire page cache anyway. Meanwhile the actual cgroup charge accrues *while the scratch file is being written* (up to `rowGroupByteLimit`, 512 MiB by default) and again while `transferColumnValues` reads it all back.

**The benchmark** (`BenchmarkScratchPageCache`, Linux-only) streams ~200 MiB of rows through a `ParquetWriter` while a background goroutine samples the scratch file's actual page-cache residency via `mincore(2)`, reporting peak residency for the write phase and the close (read-back) phase separately. Standard benchmark metrics only see the Go heap, so this is the only way to observe what the cgroup sees. It skips itself on tmpfs, where fadvise can't drop anything (worth double-checking that production scratch files don't land on a tmpfs mount — if they do, none of this fadvise work has any effect).

**The fix** does two things:

1. `flushBuffer` now calls `sync_file_range(WAIT|WRITE|WAIT)` + `FADV_DONTNEED` on the scratch file after each row group, capping write-phase residency at ~one buffer instead of the full scratch file.
2. `transferColumnValues` drops each column chunk's byte range right after transferring it (those pages are clean by then, so no sync needed), capping the read-back phase the same way.

**benchstat**, before = benchmark on this branch as-is, after = with the fix (n=6, `-benchtime 3x`, all deltas p=0.002):

| metric | before | after | vs base |
|---|--:|--:|--:|
| time per op | 832.6 ms ± 8% | 1204.2 ms ± 3% | +44.6% |
| throughput | 234.6 MiB/s ± 8% | 162.2 MiB/s ± 4% | −30.9% |
| write-phase peak page cache | 187.8 MiB ± 0% | 23.5 MiB ± 0% | **−87.5%** |
| close-phase peak page cache | 199.3 MiB ± 0% | 25.1 MiB ± 0% | **−87.4%** |

The scratch file is 199.3 MiB in both cases: before, essentially all of it is resident in cgroup-charged page cache at peak (187.8 MiB during writes, 199.3 MiB during the Close read-back); after, both phases are pinned at ~25 MiB — one `bufferSize` — independent of scratch file size. With the default 512 MiB `rowGroupByteLimit` the production reduction would be ~20x.

**The honest trade-off:** the benchmark shows ~31% lower throughput, but that's a worst case — the sink is `io.Discard`, so the synchronous scratch-file writeback is the only I/O being measured. In production the sink is a network upload that should dominate. If the cost matters, the standard mitigation is the "one-behind" pattern: kick `SYNC_FILE_RANGE_WRITE` (async) on row group N at write time and only wait+drop on row group N−1 at the next flush, overlapping writeback with encoding.

All existing `go/writer` tests pass with byte-identical snapshots. Feel free to drop, squash, or rework these commits however you like.

go/writer/parquet.go
jacobmarble: See my comment below. The throughput penalty is real, but may not matter and is expected.

go/writer/parquet.go
mdibaiee: I backed out on the direct write actually, leftover comment, thanks!

go/writer/parquet.go
mdibaiee: Thanks!

jacobmarble (approved):

PR #4754, opened 2026-07-03
materialize-s3-iceberg: reload table metadata on retry

### LINKED ISSUE #4751: materialize-s3-iceberg: append_files commit retry reuses stale table, so one concurrent commit aborts the whole transaction
## 1. Priority

Medium - a production materialization is degraded with customer-visible impact: dataflow is delayed, several collections are days behind in the destination and not self-clearing at current throughput. No data loss (Iceberg commits are atomic) and a partial workaround exists (binding `priority` to protect individual tables), but the workaround does not restore overall throughput.

## 2. Scope and prevalence

Theoretical blast radius: any `materialize-s3-iceberg` task whose target Iceberg catalog receives a concurrent commit to a table this connector is appending to. Concurrent commits are a normal Iceberg condition, not misuse: table maintenance (compaction / rewrite, snapshot expiration), catalog-side operations, a second writer, or this connector's own crash-restart process overlap (already flagged in a code comment at `iceberg-ctl/iceberg_ctl/__main__.py:482-486`) all produce them. Large-binding tasks amplify the impact because one failed table commit aborts the whole multi-binding transaction, and restart re-reads every table's metadata before any write resumes.

Observed prevalence: 1 tenant, 1 task, 252 bindings, single shard. Crashing roughly every 30 min during a backfill. First reported occurrence. The connector is a single writer here (single shard confirmed, all log lines carry `keyBegin:00000000`), so the competing commits originate outside this task. Impact is chronic: ~14 shard failures/day (one every ~1.7h), continuous since at least 2026-06-08 (log-retention limit) with no clean period. Consequence: the destination lags the source by days on high-volume tables (one is ~3 days behind), and the backlog does not clear on its own at current throughput.

## 3. Trigger (testable)

An Iceberg table that receives a commit from any other writer during the window between when `append_files` loads the table and when it commits. The connector's 3 retries then all re-assert the stale base snapshot and fail identically, so a single competing commit is unrecoverable and raises, aborting the transaction.

Minimal repro input (synthetic, no customer data):
1. Create an Iceberg table in a REST catalog.
2. Have writer A run `append_files` against it (load table, stage add-files).
3. Before A commits, have writer B commit any change to the same table's `main` branch.
4. Let A attempt its commit. Observe A retries 3x against the pre-B snapshot and fails on all attempts with `branch main has changed`, rather than reloading and succeeding.

## 4. Root cause (confirmed - code ref)

`iceberg-ctl/iceberg_ctl/__main__.py` `append_files` loads the table once at line 457, before the retry loop. Inside `while True:` (line 490) each attempt calls `txn = tbl.transaction()` on that same stale `tbl` (line 493), so every attempt asserts the same base snapshot id. On a commit conflict the retries cannot succeed because the base is never refreshed. Confirmed by logs: the asserted `expected id` is unchanged across attempts 2 and 3 (see section 5). The reload at line 513 happens only after a successful commit, for verification.

- Confirmed source data was present: the upstream capture backfill of the affected table completed and the collection held the data; nothing was pending on the source side.
- Materialization was crash-looping. `append_files` output for multiple distinct tables:
```
append_files: exception on attempt 1: CommitFailedException: Cannot commit to table <table> metadata location from <...>.metadata.json to <...>.metadata.json because it has been concurrently modified to <...>.metadata.json
append_files: sleeping 2 seconds before retry
append_files: exception on attempt 2: CommitFailedException: Requirement failed: branch main has changed: expected id <A> != <B>
append_files: sleeping 4 seconds before retry
append_files: exception on attempt 3: CommitFailedException: Requirement failed: branch main has changed: expected id <A> != <B>
append_files: max retries reached, re-raising exception
```
  Note `expected id <A> != <B>` is identical on attempts 2 and 3: the table state (`<B>`) did not change between them, yet the connector kept asserting the stale base `<A>`. That is the non-rebasing retry.
- Each failure propagated to a shard crash:
```
runTransactions: store.StartCommit: ... pyiceberg.exceptions.CommitFailedException: ... branch main has changed
```
- Crash timeline (exact `ts` for log grep; each cycle is `commit failed: appendFiles` -> `connector failed` -> `shard failed`, each on a different table):
```
2026-07-02T19:30:46.292543345Z  commit failed: appendFiles  <table-1>
2026-07-02T20:00:53.359278771Z  commit failed: appendFiles  <table-2>
2026-07-02T20:33:23.370781639Z  commit failed: appendFiles  <table-3>
2026-07-02T21:53:06.125697021Z  commit failed: appendFiles  <table-4>
2026-07-02T23:35:07.232109364Z  shard failed (post-priority restart, right after recovery commit)
```
  The four failing tables are distinct (different schemas), consistent with an external process committing across many tables rather than one hot table. Grep strings: `commit failed: appendFiles`, `branch main has changed`, `has been concurrently modified`.
- Single shard confirmed (all log lines `keyBegin:00000000`), so the concurrent committer is external to this task.
- On restart, the connector `Open` phase took ~2m40s (re-reading metadata for all 252 bindings) before any append resumed, so each crash cost a full sweep plus the Open tax.
- Workaround verified live: setting a high binding `priority` on the stuck table moved it to the front of every transaction; it committed early (before the sweep reaches a conflicting table) and drained cleanly.

1. Customer-immediate stopgap (ops, no code change): set binding `priority` on latency-critical tables so they commit first, ahead of any table that might fail the sweep; and reduce or schedule concurrent maintenance/writes (compaction, snapshot expiry, other jobs) on the target tables during backfills.
2. Data-preserving: not applicable, no data is lost; commits are atomic and the stalled data stays in the collection.
3. Code direction: reload the table (`catalog.load_table(table)`) at the top of each retry attempt so `add_files` rebases onto the current snapshot before re-committing; an append is commutative with other appends, so a rebased retry should succeed. Consider more generous bounded retries with backoff for busy tables, or delegating to pyiceberg's built-in commit retry (`commit.retry.num-retries`) instead of the hand-rolled loop. Related follow-up (separate concern): a single table's commit failure aborts the entire multi-binding transaction, so the blast radius of one conflict is the whole sweep plus a full-metadata Open on restart; decoupling per-table commit failure from the whole-transaction abort would bound the damage.

Any code fix is drafted for human review, not self-merged or published.

## 7. Definition of done

- Acceptance: with a competing writer committing to the same table between `append_files`'s table-load and commit, `append_files` reloads and eventually commits successfully; an intermittent concurrent commit no longer escalates to a connector/shard failure.
- Regression test: simulate a concurrent commit to the target table in the window between table-load and commit-attempt; assert `append_files` retries by reloading the table and commits, and that both writers' data files are present afterward (no rows dropped).
- Constraints / do not break: preserve the `flow_checkpoints_v1` idempotency (do not double-append the same files on retry, keep the "checkpoint already set, skip" short-circuit); do not drop or null data on conflict; do not weaken exactly-once semantics.

## 8. References

- Slack thread (internal, escalation system of record): https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1783037180783329

mdibaiee:
**Description:**

- When retrying a failed transaction due to a conflict, we should reload table metadata so that the new retry is no longer on a stale metadata that would conflict again. Tested this locally by reproducing the issue and verifying with this change the issue is resolved.

Resolves https://github.com/estuary/connectors/issues/4751

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
+++ b/materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
@@ -507,6 +507,8 @@ def append_files(
                 raise
             log(f"append_files: sleeping {attempt * 2} seconds before retry")
             time.sleep(attempt * 2)
+            log(f"append_files: reloading table before retry (attempt {attempt + 1})")
+            tbl = catalog.load_table(table)
             attempt += 1
 
     log(f"append_files: reloading table to verify update")
--- a/materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
+++ b/materialize-s3-iceberg/iceberg-ctl/iceberg_ctl/__main__.py
@@ -457,37 +457,41 @@ def append_files(
     tbl = catalog.load_table(table)
     log(f"append_files: table loaded successfully")
 
-    log(f"append_files: parsing checkpoints from table properties")
-    checkpoints = TypeAdapter(dict[str, str]).validate_json(tbl.properties.get("flow_checkpoints_v1", "{}"))
-    cp = checkpoints.get(materialization, "")  # prev_checkpoint will be unset if this is the first commit to the table
-    log(f"append_files: current checkpoint from table is '{cp}'")
-
-    if cp == next_checkpoint:
-        log(f"append_files: checkpoint is already '{next_checkpoint}', skipping append")
-        print(f"checkpoint is already '{next_checkpoint}'")
-        return  # already appended these files
-    # TODO(whb): Re-enable this sanity check after any tasks effected by the
-    # disabled bindings state tracking bug have moved past it.
-    # elif cp != "" and cp != prev_checkpoint:
-    #     # An absent checkpoint table property is allowed to accommodate cases
-    #     # where the user may have manually dropped the table and the
-    #     # materialization automatically re-created it, outside the normal
-    #     # backfill counter increment process.
-    #     raise Exception(
-    #         f"checkpoint from snapshot ({cp}) did not match either previous ({prev_checkpoint}) or next ({next_checkpoint}) checkpoint"
-    #     )
-
-    # Files are only added if the table checkpoint property has the prior checkpoint. The checkpoint
-    # property is updated to the current checkpoint in an atomic operation with appending the files.
-    # Note that this is not 100% correct exactly-once semantics, since there is a potential race
-    # between retrieving the table properties and appending the files, where a zombie process could
-    # append the same files concurrently. In principal Iceberg catalogs support the atomic
-    # operations necessary for true exactly-once semantics, but we'd need to work with the catalog
-    # at a lower level than PyIceberg currently makes available.
-    checkpoints[materialization] = next_checkpoint
-
+    # The checkpoint is re-parsed from the table properties and re-validated on every attempt, not
+    # just the first, so that a retry after a reload always rebases its idempotency check against
+    # the latest table state. Without this, a retry could recommit over a checkpoint transition that
+    # a concurrent (e.g. zombie) process already completed, silently duplicating the appended files.
     attempt = 1
     while True:
+        log(f"append_files: parsing checkpoints from table properties (attempt {attempt})")
+        checkpoints = TypeAdapter(dict[str, str]).validate_json(tbl.properties.get("flow_checkpoints_v1", "{}"))
+        cp = checkpoints.get(materialization, "")  # prev_checkpoint will be unset if this is the first commit to the table
+        log(f"append_files: current checkpoint from table is '{cp}'")
+
+        if cp == next_checkpoint:
+            log(f"append_files: checkpoint is already '{next_checkpoint}', skipping append")
+            print(f"checkpoint is already '{next_checkpoint}'")
+            return  # already appended these files
+        # TODO(whb): Re-enable this sanity check after any tasks effected by the
+        # disabled bindings state tracking bug have moved past it.
+        # elif cp != "" and cp != prev_checkpoint:
+        #     # An absent checkpoint table property is allowed to accommodate cases
+        #     # where the user may have manually dropped the table and the
+        #     # materialization automatically re-created it, outside the normal
+        #     # backfill counter increment process.
+        #     raise Exception(
+        #         f"checkpoint from snapshot ({cp}) did not match either previous ({prev_checkpoint}) or next ({next_checkpoint}) checkpoint"
+        #     )
+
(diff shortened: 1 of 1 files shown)


jwhartley:
Reloading `tbl` on retry fixes the stale-snapshot conflict, but the retry loop reuses the `checkpoints` dict computed before the conflict, which opens a narrower version of the race this connector already documents.

Risk: crash-restart overlap (called out in #4751 as a real concurrent-commit source, not just external compaction/maintenance jobs) can produce a zombie old task instance and a newly restarted instance both attempting the same checkpoint transition. If the old instance commits first, the new instance's commit conflicts. With this fix, its retry reloads the table and rebases the transaction, but never re-checks whether the checkpoint was already advanced by the other instance before recommitting. That turns what used to be a loud crash-loop (data-safe, since a fresh invocation after restart would hit the checkpoint short-circuit and skip) into a silent double-append of the same files.

Suggestion: move the checkpoint read, skip-check, and dict update inside the retry loop so every attempt re-validates against the freshly loaded table, not just the first attempt:

```python
attempt = 1
while True:
    checkpoints = TypeAdapter(dict[str, str]).validate_json(tbl.properties.get("flow_checkpoints_v1", "{}"))
    cp = checkpoints.get(materialization, "")
    if cp == next_checkpoint:
        log(f"append_files: checkpoint is already '{next_checkpoint}', skipping append")
        print(f"checkpoint is already '{next_checkpoint}'")
        return
    checkpoints[materialization] = next_checkpoint
    try:
        txn = tbl.transaction()
        txn.add_files(file_paths.split(","))
        txn.set_properties({"flow_checkpoints_v1": json.dumps(checkpoints)})
        txn.commit_transaction()
        break
    except Exception as e:
        if attempt == 3:
            raise
        time.sleep(attempt * 2)
        tbl = catalog.load_table(table)
        attempt += 1
```

This isn't a new trust assumption, it's the same idempotency gate the function already applies on entry, just applied on every attempt instead of only the first. It also means the disabled sanity check a few lines up would automatically cover retries too if it's ever re-enabled. Extra cost is one cheap JSON parse per retry, negligible next to the `load_table` round trip already added.

mdibaiee:
@jwhartley thanks, reproduced locally and applied the fix, verifying this scenario is also covered

jwhartley (approved):
The updated commit (83e31c0) correctly closes the gap from the first version: checkpoint read + idempotency skip-check + the checkpoints dict update now run inside the retry loop, so every attempt (including retries) re-validates against the freshly reloaded table instead of a stale pre-conflict snapshot. That fixes the double-append risk a zombie/concurrent process could have triggered.

Fixes the reported stale-snapshot bug in #4751.

PR #4757, opened 2026-07-04
materialize-boilerplate: retain_existing_data should disable load opt

### LINKED ISSUE #4702: Reject retain_existing_data_on_backfill without allow_existing_tables_for_new_bindings
> Priority: Low (control-plane triage, 2026-06-18)
>
> Few users in practice, the docs are being clarified in estuary/flow#3049, and we only found 2 tenants who had configured this incorrectly.

---

## Summary

When `retain_existing_data_on_backfill` is enabled without `allow_existing_tables_for_new_bindings`, backfilling a table that already contains data produces duplicate rows. We should guard against this combination so users cannot silently end up with duplicates.

## Mechanism

- `retain_existing_data_on_backfill` skips the truncate on backfill (materialize-boilerplate/materializer.go, retain branch ~L582), so existing rows remain.
- A backfill increments the binding's backfill counter, which is part of its `state_key` (`assemble::encode_state_key(resource_path, backfill)`), so the runtime's persisted per-binding max-key (the high-water mark behind the load-key optimization) is no longer found and resets to empty (`crates/runtime/src/materialize/protocol.rs`).
- The optimization then skips the Load for any key above that now-empty mark (`protocol.rs`, `if !disable_load_optimization { return Ok(None) }`). Because a backfill replays in key order, every key reads as a new high, so all loads are skipped.
- With loads skipped, the store takes the insert path (`it.Exists == false`) rather than merge, inserting rows next to the retained ones, producing duplicates.
- `allow_existing_tables_for_new_bindings` sets `DisableLoadOptimization` on the Opened response (materialize-boilerplate/materializer.go:772), which forces loads, so the retain backfill merges instead of inserting. That is why the two flags must be used together.

## Proposed change

Reject the combination at publish/validation time: a materialization with `retain_existing_data_on_backfill` set but `allow_existing_tables_for_new_bindings` not set should fail with a message telling the user to enable both (per the discussion: retain should require allow, and error if it is not set). Docs now spell out that the two flags must be used together: estuary/flow#3049.

Alternatives:
- Make the runtime load-key optimization retain-aware so it does not skip loads when the table was not truncated (fixes the root cause rather than guarding the config combination).
- Have `retain` imply the load-skip-disable automatically.

## Migration note

Existing tasks may already be running with `retain` set and `allow` unset (this is how the duplicates surface in the field). A hard validation error would appear on their next publish. Decide whether to hard-error, warn, or grandfather existing tasks; at minimum the error message should point users to the missing flag.

## Confirmation

Traced through connector and runtime source and confirmed with engineering: `allow` is known to force loads correctly, and `retain` should require it.

mdibaiee:
**Description:**

- [materialize-boilerplate: retain_existing_data should disable load opt](https://github.com/estuary/connectors/pull/4757/changes/5aa4a868066e07676ec884c95f5aa24302431121)
- Resolves #4702 

**Workflow steps:**

**Notes for reviewers:**

### DIFF
--- a/materialize-boilerplate/materializer.go
+++ b/materialize-boilerplate/materializer.go
@@ -796,8 +796,15 @@ func RunNewTransactor[EC EndpointConfiger, FC FieldConfiger, RC Resourcer[RC, EC
 	}
 
 	return transactor, &pm.Response_Opened{
-		RuntimeCheckpoint:       cp,
-		DisableLoadOptimization: featureFlags["allow_existing_tables_for_new_bindings"],
+		RuntimeCheckpoint: cp,
+		// retain_existing_data_on_backfill skips truncating a table on
+		// backfill, but the runtime's per-binding load-key optimization is
+		// reset by a backfill and would skip loads, causing backfills to
+		// insert duplicate rows alongside the retained data instead of
+		// merging with it. Disabling the load optimization forces loads so
+		// that retained data is merged correctly.
+		DisableLoadOptimization: featureFlags["allow_existing_tables_for_new_bindings"] ||
+			featureFlags["retain_existing_data_on_backfill"],
 	}, &mCfg.MaterializeOptions, nil
 }
 


danielnelson (approved):
Approved, but could we `DisableLoadOptimization` for `retain_existing_data_on_backfill` instead of requiring the other feature flag? This keeps all the flags independent which seems preferable to having this interaction.

materialize-boilerplate/materializer.go
```diff
@@ -962,6 +970,25 @@ func (c *constrainterAdapter[EC, FC, RC, MT]) DescriptionForType(p *pf.Projectio
...
+// backfill, but the runtime's per-binding load-key optimization is reset by a
+// backfill and will skip loads unless allow_existing_tables_for_new_bindings
+// is also set to force loads via DisableLoadOptimization. Without that,
+// skipped loads cause backfills to insert duplicate rows alongside the
+// retained data instead of merging with it.
+func checkFeatureFlagCombination(parsedFlags map[string]bool) error {
+	if parsedFlags["retain_existing_data_on_backfill"] && !parsedFlags["allow_existing_tables_for_new_bindings"] {
+		return fmt.Errorf(
+			"the 'retain_existing_data_on_backfill' feature flag requires the 'allow_existing_tables_for_new_bindings' feature flag to also be enabled, otherwise backfills will produce duplicate rows. " +
+				"Add 'allow_existing_tables_for_new_bindings' to 'advanced.feature_flags' in the endpoint configuration, or remove 'retain_existing_data_on_backfill' if it is not needed",
```
jacobmarble: This error's advice can contradict the pre-existing error at line 610 (the drop/recreate path). A user who sets only `retain_existing_data_on_backfill` hits this message and **adds** `allow_existing_tables_for_new_bindings` as instructed — but if their backfill also requires dropping and re-creating the table (e.g. a key type change), they then hit the line-610 error telling them to **remove** `retain_existing_data_on_backfill`.

jacobmarble (approved):

materialize-boilerplate/materializer.go
mdibaiee: I think if we follow @danielnelson 's suggestion of just disabling load optimization for `retain_existing_data_on_backfill` we also don't fall into this trap...

these two really do seem to be distinct ones and they don't always mix well 👍🏽

mdibaiee:
@danielnelson updated the PR to just disable load optimization on `retain_existing_data` feature flag

danielnelson (approved):

jwhartley:
Docs update for this: the feature-flags page had a stale caveat telling users to pair `retain_existing_data_on_backfill` with `allow_existing_tables_for_new_bindings` to avoid duplicates. Updated in [estuary/flow#3125](https://github.com/estuary/flow/pull/3125).

PR #4760, opened 2026-07-05
Trigger a cross-repo docs preview build on connector-docs PRs

jwhartley:
## What

Triggering side of the cross-repo PR preview mechanism. Companion PR: estuary/docs#7.

`docs/` here is aggregated into estuary/docs via a git submodule, so a PR against this repo that touches `docs/**` has no self-contained Docusaurus build to preview with — that's a real UX regression from what flow/site has today (a sticky "preview deployed" comment on every docs PR). This workflow closes that gap: on `opened`/`reopened`/`synchronize`/`closed` for a PR touching `docs/**`, it mints a GitHub App token scoped to estuary/docs and fires a `repository_dispatch` there with the PR number, head SHA, and action. estuary/docs#7 receives it, builds against that commit, deploys a preview, and comments back on this PR.

## Why gated, not wired live yet

The job checks `vars.PREVIEW_APP_ID != ''` and skips (not fails) if unset. The GitHub App backing this (`estuary-docs-previews`) still needs to be created via the GitHub UI — no API path for downloading a new App's private key — so merging this now is safe; it just won't fire anything until the App exists and its credentials are added here and in estuary/docs. Setup steps logged in the internal migration tracker.

### DIFF
--- a/.github/workflows/docs-preview-trigger.yaml
+++ b/.github/workflows/docs-preview-trigger.yaml
@@ -1,0 +1,56 @@
+name: "Trigger connector docs preview"
+
+# estuary/docs owns the Docusaurus build (it aggregates this repo's docs/
+# via a git submodule), so a PR here that touches docs/ can't build its own
+# preview. This dispatches the build to estuary/docs, which reports back
+# with a sticky comment on this PR. See estuary/docs
+# .github/workflows/connector-docs-preview.yaml for the receiving side.
+#
+# Requires the "estuary-docs-previews" GitHub App (PREVIEW_APP_ID repo
+# variable + PREVIEW_APP_PRIVATE_KEY secret, installed on both estuary/docs
+# and estuary/connectors). Until those exist, this job skips rather than
+# failing, so it's safe to merge ahead of the App being created.
+
+on:
+  pull_request:
+    paths:
+      - docs/**
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
+
+concurrency: connector-docs-preview-${{ github.event.pull_request.number }}
+
+jobs:
+  dispatch:
+    if: ${{ vars.PREVIEW_APP_ID != '' }}
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Mint token for estuary/docs
+        uses: actions/create-github-app-token@v2
+        id: app-token
+        with:
+          app-id: ${{ vars.PREVIEW_APP_ID }}
+          private-key: ${{ secrets.PREVIEW_APP_PRIVATE_KEY }}
+          owner: estuary
+          repositories: docs
+
+      - name: Dispatch preview build to estuary/docs
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          PR_ACTION: ${{ github.event.action }}
+        run: |
+          gh api repos/estuary/docs/dispatches --input - <<EOF
+          {
+            "event_type": "connector-docs-preview",
+            "client_payload": {
+              "pr_number": ${PR_NUMBER},
+              "head_sha": "${HEAD_SHA}",
+              "action": "${PR_ACTION}"
+            }
+          }
+          EOF
--- a/.github/workflows/docs-preview-trigger.yaml
+++ b/.github/workflows/docs-preview-trigger.yaml
@@ -23,6 +23,10 @@ on:
 
 concurrency: connector-docs-preview-${{ github.event.pull_request.number }}
 
+# The job only ever uses the App-minted token (scoped to estuary/docs), never
+# the default GITHUB_TOKEN, so it needs none of the ambient token's scope.
+permissions: {}
+
 jobs:
   dispatch:
     if: ${{ vars.PREVIEW_APP_ID != '' }}
--- a/.github/workflows/docs-preview-trigger.yaml
+++ b/.github/workflows/docs-preview-trigger.yaml
@@ -7,13 +7,30 @@ name: "Trigger connector docs preview"
 # `docs/build` commit status. See estuary/docs
 # .github/workflows/connector-docs-preview.yaml for the receiving side.
 #
-# The commit status is the part worth protecting: two thirds of the pages on
-# docs.estuary.dev come from this repo, the docs build sets
-# onBrokenLinks: 'throw', and the production deploy is an unattended hourly
-# cron with no failure alerting. A broken relative link merged here therefore
-# fails every subsequent deploy and freezes docs.estuary.dev at its last good
-# build, with nothing to say so. Making `docs/build` a required check on this
-# repo closes that.
+# The `docs/build` status this posts is advisory, and must not be made a
+# required check. It is a commit status written from another repository, not an
+# Actions check run in this one, so it is never created at all for a PR that
+# touches no published docs: this workflow does not run, nothing dispatches,
+# and branch protection waits for a status that will never arrive. The
+# auto-skip that rescues path-filtered Actions jobs does not apply. Since
+# ci.yaml and python.yaml both `paths-ignore: docs/**`, the code and docs
+# halves are disjoint by design and every code PR would hang. The same gap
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-preview-trigger.yaml
```diff
@@ -0,0 +1,60 @@
...
+#
+# Requires the "estuary-docs-previews" GitHub App (PREVIEW_APP_ID repo
+# variable + PREVIEW_APP_PRIVATE_KEY secret, installed on both estuary/docs
+# and estuary/connectors). Until those exist, this job skips rather than
+# failing, so it's safe to merge ahead of the App being created.
+
+on:
+  pull_request:
+    paths:
+      - docs/**
```
aeluce: Should this be scoped specifically to the `docs/reference/Connectors/` path? There are some other internal references under `docs/` in this repo that wouldn't become actual docs pages.

.github/workflows/docs-preview-trigger.yaml
```diff
@@ -0,0 +1,60 @@
...
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - closed
```
danielnelson: Looking at the docs for this, the default values are [opened, reopened, synchronize], so this adds closed?  Why do we want to trigger on close?

jwhartley:
Parking this until the docs migration is done, alongside its companion estuary/docs#7. Leaving it open rather than closing it.

The receiving side turns out not to work as written: `pr-preview-action` publishes to a `gh-pages` branch, but `estuary/docs` deploys Pages via `actions/deploy-pages` and has no such branch, so a preview would never be served. Resolving that means picking a hosting model, and the cutover host is trending to Cloudflare Workers, which would replace the mechanism rather than fix it. Details in the comment on estuary/docs#7.

Nothing wrong with this side specifically. @aeluce's ask to scope the path filter to `docs/reference/Connectors/` and @danielnelson's question about triggering on `closed` are both still open and still valid; they're logged so they get picked up when this resumes.

aeluce (dismissed):

skord (changes requested):
Reviewed this against the receiving side in estuary/docs#7 (which I've reviewed and approved separately).

## Verified correct

**Cross-repo contract matches.** `event_type: connector-docs-preview` lines up with the receiver's `repository_dispatch: types:`, and both payload fields (`pr_number` as a number, `head_sha` as a string) are consumed on the other side. The receiver's `format()`-before-`fromJSON()` handling covers the number-vs-string difference between the dispatch payload and its `workflow_dispatch` fallback, so both entry points work.

**Path filter is complete.** I checked it against `docs/README.md`, which classifies the directory itself: `agents/`, `materialize/`, `feature_flags.md`, `inbound_networking.md`, and `oauth.md` are engineering docs, and only `reference/Connectors/` is published. The filter covers exactly the published set plus `redirects.yaml`. `oauth.md` is not in the filter and correctly should not be. This resolves @aeluce's comment fully.

**Injection-safe dispatch.** `PR_NUMBER` / `HEAD_SHA` go through `env:` rather than direct `${{ }}` interpolation into the `run:` body. The unquoted heredoc is fine given both values are GitHub-controlled.

**Conventions match.** `.yaml` extension, `ubuntu-24.04` (all 11 existing `runs-on` lines use it), and `permissions: {}` resolves the CodeQL finding correctly given nothing uses the ambient token. @danielnelson's `closed` question is resolved by dropping it.

## Requesting changes on one point

**Making `docs/build` a required check will block every non-docs PR.** The header comment argues for this ("Making `docs/build` a required check on this repo closes that"), and the reasoning about the unattended hourly deploy is sound, but the mechanism does not work as written.

`docs/build` is a commit status created by `createCommitStatus` from a different repository. It is not an Actions check run in this repo. When a PR does not touch `docs/reference/Connectors/**`, this workflow never runs, nothing dispatches, and the status is never created, so branch protection sits at "Expected, waiting for status to be reported" indefinitely. The auto-skip behaviour that rescues path-filtered Actions jobs from this does not apply to an externally-posted commit status.

That is the common case, not an edge case: `ci.yaml:12` and `python.yaml:8` both `paths-ignore: docs/**`, so the two halves are disjoint by design. Every code PR in the repo would hang.

The same gap applies on three narrower paths, all of which currently skip rather than post a status: fork PRs, `PREVIEW_APP_ID` unset (which is the state at merge time), and the receiver's status step being gated on `steps.app-token.outputs.token != ''` with `continue-on-error: true`.

Options, roughly in order of how well they fit what is already here:

1. Add a second job to this workflow with no `paths` filter that posts `docs/build` as `success` when the PR touches no published docs. Keeps one context always reported. Costs a runner-second per PR.
2. Drop the `paths:` filter, run the job always, and let it either dispatch or self-report a passing status. Simplest to reason about, but loses the "do not spend a build" property the third commit deliberately added.
3. Leave `docs/build` advisory and not required. Loses the protection the comment is arguing for.

None of this blocks the file as written, since the workflow is inert until the App exists. But the required-check step is written up as the payoff, so at minimum the header comment should say what has to happen first, and whoever picks this up from the migration tracker should not be able to flip the switch and wedge the repo. Happy to approve on either a fix or an amended comment.

## Minor

- **Document the App permissions this side needs.** The receiving side spells out "Commit statuses: Read and write"; this side only says "installed on both". `POST /repos/{owner}/{repo}/dispatches` needs **Contents: read and write** on estuary/docs. Missing App permissions return 404 rather than 403, so getting this wrong produces a misleading failure during setup, and the App does not exist yet.
- `owner: estuary` is hardcoded in the token step while the docs side uses `github.repository_owner`. Harmless, just asymmetric.
- A PR touching only `docs/redirects.yaml` builds correctly but gets an empty changed-pages list in the comment, since the receiver's diff is scoped to `docs/reference/Connectors/`. Cosmetic; the build check is the point there.
- Drafts spend a full site build on every push. Probably what you want for a docs PR, but `ready_for_review` is the lever if build minutes become a concern.

jwhartley:
You are right, and the header was arguing for something that cannot work. `docs/build` is a commit status posted from estuary/docs, not a check run here, so for a PR touching no published docs it is never created and branch protection waits forever. With `docs/**` in `paths-ignore` on both `ci.yaml` and `python.yaml` that is every code PR.

Taking option 3, and the reason it is now the right one rather than a concession: #5090 merged since your review and already provides the gate. `docs-build-check.yaml` runs the same docs build as a native Actions job in this repo, so it path-filters and auto-skips correctly, and it is the context safe to require. Adding a second always-running job here to keep `docs/build` reportable would be building the protection twice, by the more fragile of the two mechanisms.

So no code change. Pushed a header rewrite that says `docs/build` is advisory and must not be required, records why (including the three narrower paths you listed: forks, `PREVIEW_APP_ID` unset, and the receiver's token gate), and points at #5090 as the check to require instead.

One thing your review surfaced that I have documented rather than fixed: this workflow and #5090 now share the same `paths:` filter, so a published-docs PR builds the site twice, about three minutes each. Collapsing them means either the required check waiting on a cross-repo dispatch, which is the fragility above, or the preview being built by a job that cannot publish it. Two independent builds looked like the cheaper failure mode, but say if you would rather I made the preview conditional on something narrower.

Also rebased onto main, which had moved 37 commits.

.github/workflows/docs-preview-trigger.yaml
jwhartley: Yes, and it is scoped that way now. The filter is `docs/reference/Connectors/**` plus `docs/redirects.yaml`, which is the whole published set: `redirects.yaml` is not a page, but `docusaurus.config.js` reads it out of the submodule to build the connector redirect table, so a bad entry there fails the build. Everything else under `docs/` (`agents/`, `materialize/`, `feature_flags.md`, `inbound_networking.md`, `oauth.md`) is engineering notes that publish nothing and are correctly excluded.

.github/workflows/docs-preview-trigger.yaml
jwhartley: Right, and `closed` is dropped. It was left over from the earlier gh-pages design, where a preview was a published directory that had to be deleted when the PR closed. This version uploads a Cloudflare Workers version instead, and versions cannot be deleted, so there is nothing to tear down. Cloudflare ages old ones out on its own. The trigger is now the explicit default set, `[opened, reopened, synchronize]`, with a comment recording why `closed` is absent so it does not get added back.

skord (approved):
LGTM

PR #4762, opened 2026-07-06
source-genesys: fix RateLimiter argument error

### LINKED ISSUE #4761: source-genesys: conversations stream capture crashes on RateLimiter.update() signature mismatch
## 1. Priority
Med/Low - narrow connector (only 3 tenants total run `source-genesys`), but the affected tenant's capture is likely fully down (all 7 bindings), not just `conversations` (see Scope). Single task/tenant, and a clean stopgap is available (drop the `conversations` binding), which keeps this from being High.

## 2. Scope and prevalence
Theoretical: any tenant running the `source-genesys` `conversations` stream in incremental mode hits this on every sync, since the crashing call is unconditional in the job-status polling loop.

Observed: 3 tenants total run `source-genesys` (control-plane query on `connector_image_name`), of which 1 is a real production tenant actively hitting this (confirmed via `flowctl catalog status`); the other 2 are a personal test catalog and a possibly-inactive tenant, not checked further. This is a regression, not a long-standing bug: it was only introduced when a companion `estuary-cdk` change (see Root cause) landed on 2026-07-02, and the affected tenant's first failure followed within a day (2026-07-03T03:15:48Z), so prevalence should track how many tenants actively run the `conversations` stream, not the full `source-genesys` install base.

## 3. Trigger (testable)
Every iteration of the `conversations` stream's async-job status-polling loop (`_perform_conversation_job` in `source_genesys/api.py`) calls `RateLimiter.update()` with 2 positional args. The current `estuary_cdk.http.RateLimiter.update` signature requires 3 (`self, log, cur_delay, failed`), so this raises unconditionally on the very first status check of every job, before any data is ever yielded.

Minimal repro input (synthetic, no customer data): instantiate `estuary_cdk.http.RateLimiter()` and call `.update(1.0, True)`. This raises:
```
TypeError: RateLimiter.update() missing 1 required positional argument: 'failed'
```

## 4. Root cause (if known) + confidence
Confirmed (code read, `git log -L` on both sides, and the failure signature matches exactly) - `source-genesys/source_genesys/api.py:249` still calls the pre-regression 2-argument form of `RateLimiter.update()`.

`estuary-cdk` commit `d62abf39a` ("estuary-cdk: warn once when the rate limiter saturates at MAX_DELAY", 2026-07-02) changed `RateLimiter.update()`'s signature from `(self, cur_delay, failed)` to `(self, log, cur_delay, failed)`, and updated the CDK's own internal call site (`estuary_cdk/http.py:630`) to match. `source-genesys`'s hand-rolled call site at `api.py:249` (`rate_limiter.update(delay, True)`) was never updated. `log` is already in scope in the enclosing function (`_perform_conversation_job`, `api.py:199`), so the correct call is simply `rate_limiter.update(log, delay, True)`.

- Compared the reported error text to `estuary_cdk.http.RateLimiter.update`'s current signature -> the "missing 1 required positional argument: 'failed'" message is exactly what Python raises when a 3-arg method is called with 2 positional args.
- `git log -L 466,466:estuary-cdk/estuary_cdk/http.py` -> signature changed from `(self, cur_delay, failed)` to `(self, log, cur_delay, failed) -> None` in commit `d62abf39a`, 2026-07-02.
- `git log -L 249,249:source-genesys/source_genesys/api.py` -> call site last touched by the original connector commit (`a023adb36`, "source-genesys: new connector", 2024-11-12), never updated for the new signature.
- `grep -rn "RateLimiter()" .` across the connectors repo -> 4 connectors instantiate the CDK's `RateLimiter` directly: `source-shopify-native`, `source-gong`, `source-ringcentral`, `source-genesys`. Checked each: `source-shopify-native` never calls `.update()` itself (relies on the CDK's internal HTTP path, which already uses the correct 3-arg form); `source-gong` and `source-ringcentral` define their own independent rate-limiter classes unrelated to the CDK's `RateLimiter`. `source-genesys` is the only connector with a stale hand-rolled call.
- Confirmed active on the affected tenant via `flowctl catalog status <task>`: shardFailure alert firing since first_ts `2026-07-03T03:15:48.860103488Z`, still firing as of `2026-07-05T23:09:36.426929980Z` (9+ recent failures at last check). Error verbatim:
```
2026-07-03T03:15:48.860103488Z  runTransactions: readMessage: Task capture.conversations.incremental: RateLimiter.update() missing 1 required positional argument: 'failed'
```
- Checked whether this crash halts the capture's other 6 bindings. All bindings in an `estuary-cdk` Python capture run inside one `asyncio.TaskGroup` per shard (`estuary_cdk/capture/task.py`), so an unhandled exception in one binding's task cancels its siblings and crashes the whole shard on that attempt. Grepped `RateLimiter()` usage in `source_genesys/api.py` and confirmed it is instantiated exactly once, inside `_perform_conversation_job` - every other stream (`campaigns`, `messaging_campaigns`, `queue_members`, `queues`, `teams`, `users`) calls `http.request()` directly (the CDK's own HTTP path, unaffected), so only `conversations` can ever raise this specific error.
- Initially found log activity from other bindings (e.g. `queue_members` logging a schema line as late as `2026-07-05T21:37:13Z`) and read that as partial progress. Checked hourly `catalog_stats` (`docs_written_by_me`, `bytes_written_by_me`) for `users` and `queue_members` from `2026-07-03T00:00:00Z` onward and found zero bytes/docs written in every hourly bucket since the incident began. Given `conversations` crashes within seconds of each restart (one job submission + one status poll, ~3-7s) and `asyncio.TaskGroup` cancels all sibling tasks the instant one raises, the other bindings likely never survive long enough to reach a checkpoint commit either - the log line shows a task starting, not completing. Net effect (likely, not fully confirmed at the commit level): the entire capture is down, not just `conversations`; the earlier "other bindings degraded but progressing" read was an overstatement from thin evidence.

1. Customer-immediate stopgap (ops action, human-run, no code change): republish the affected capture spec with the `conversations` binding removed. This removes the exception source from the shared `asyncio.TaskGroup`, which should let the other 6 bindings actually run to completion instead of getting cancelled every few seconds; re-add `conversations` once fixed. No data loss for the other streams; `conversations` simply continues not receiving new data during that window, same as today.
2. Fix: change `source-genesys/source_genesys/api.py:249` from `rate_limiter.update(delay, True)` to `rate_limiter.update(log, delay, True)`, matching the current 3-arg signature.
3. Platform direction (optional): a signature change to a widely-instantiated `estuary-cdk` class isn't caught by anything except a manual grep for other call sites - Python's duck typing means this ships clean through CI. Worth a repo-wide grep for other hand-rolled callers as a step before merging future CDK signature changes to this class (or elsewhere), to avoid a repeat.

## 7. Definition of done
- Acceptance: `conversations` incremental sync completes at least one full job-status poll cycle without raising `TypeError`; `RateLimiter.update() missing 1 required positional argument` absent from logs over a subsequent 24h window; the affected tenant's `conversations` collection resumes receiving new documents.
- Regression test: exercise `_perform_conversation_job`'s polling loop through at least one non-terminal ("QUEUED"/"EXECUTING") state transition and assert no exception is raised.
- Constraints / do not break: preserve existing rate-limiting/backoff semantics for `source-genesys`; do not change `RateLimiter`'s public signature again without grepping all call sites repo-wide first.

## 8. References
- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1783295238799229

Alex-Bair:
**Regression?**

Yes, https://github.com/estuary/connectors/pull/4749. Only `source-genesys` was affected.

**Description:**

When I updated the `RateLimiter.update` function signature in the CDK in d62abf39a1bc9e48f14863cb094a1518fafda981, I neglected to update all locations where `RateLimiter.update` was called.

This fixes the positional arguments for all other `RateLimiter.update` call sites, which is only one spot in `source-genesys`.

### DIFF
--- a/source-genesys/source_genesys/api.py
+++ b/source-genesys/source_genesys/api.py
@@ -246,7 +246,7 @@ async def _perform_conversation_job(
             case "CANCELLED" | "EXPIRED":
                 raise RuntimeError(f"Conversations job status is {state}. Job request body was: ", body)
 
-        rate_limiter.update(delay, True)
+        rate_limiter.update(log, delay, True)
 
     # Paginate through results.
     url = f"{COMMON_API}.{domain}/api/v2/analytics/conversations/details/jobs/{job_id}/results"


jwhartley (approved):
LGTM

PR #4764, opened 2026-07-06
materialize-bigquery: add local goccy emulator variant (materialize-bigquery-emulator-goccy)

### LINKED ISSUE #4768: materialize-bigquery: advanced.endpoint for BigQuery-compatible servers
**Problem**

materialize-bigquery can only target Google's SaaS endpoint and hard-requires service-account credentials — there is no way to point it at a BigQuery-compatible server such as [goccy/bigquery-emulator](https://github.com/goccy/bigquery-emulator) (local dev, credential-free integration tests) or a proxy in front of the real service.

**Proposal**

Add an `advanced.endpoint` config field (full API base path, e.g. `http://host:9050/bigquery/v2/` for goccy), keeping SaaS behavior byte-for-byte unchanged. Implemented in #4764:

- `advanced.endpoint` field; credentials become optional when it is set, and are still validated and used when provided.
- A one-time runtime probe classifies the server as SaaS vs goccy — behavior never keys off the endpoint being set or credentials being absent.
- Emulator-gated behaviors: dialect tweaks (INT64, STRING-for-JSON, split ALTERs), load path via fake-gcs-server, and a recreate-and-copy workaround for goccy's no-op `ALTER TABLE`/`tables.patch`.
- Local docker-compose integration suite (`BIGQUERY_TEST_LOCAL_EMULATOR_GOCCY`, default on), env-gated SaaS suite (`BIGQUERY_TEST_SAAS`, default off), and a `materialize-bigquery-emulator-goccy` image variant.

**Status**

#4721 is the original proposal, reference-only: several of its changes apply unconditionally to the production SaaS path. #4764 implements the above behind the runtime probe, with SaaS SQL-generation snapshots byte-identical to `main`. The capture-side endpoint support #4721 also proposed (`source-bigquery-batch`, `source-ga4-bigquery`) is out of scope here and would need its own issue.

**Acceptance**

- SaaS suite green with zero snapshot diffs vs `main`.
- Emulator suite green with no GCP credentials.

jacobmarble:
Closes #4768

**Description:**

Adds a runtime-detected `materialize-bigquery-emulator-goccy` variant so materialize-bigquery can run against [goccy/bigquery-emulator](https://github.com/goccy/bigquery-emulator) + fake-gcs-server, with SaaS BigQuery behavior unchanged (SQL-generation snapshots byte-identical to main).

- New `advanced.endpoint` config field; credentials become optional when it's set. A one-time probe at endpoint setup classifies the server as SaaS vs emulator.
- Emulator-gated behaviors: INT64/STRING-for-JSON type mappings, split ALTER statements, native temp tables + GCS load jobs (no gzip), and a recreate-table workaround for goccy's silently no-op'd `ALTER TABLE`/`tables.patch`.
- Local integration suite (`TestIntegrationLocalEmulatorGoccy`) with its own docker-compose stack and snapshots.

**Workflow steps:**

Set `advanced.endpoint` to the emulator URL; SaaS usage is unchanged. Integration suites are env-gated: the emulator suite runs by default (`BIGQUERY_TEST_LOCAL_EMULATOR_GOCCY=0` disables), the SaaS suite is opt-in (`BIGQUERY_TEST_SAAS=1`).

None.

**Notes for reviewers:**

PR #4721 is reference-only; its unconditional production-path changes were deliberately not carried over. The goccy schema-evolution bug and detection-probe rationale are documented in `emulator_alter.go` / `emulator_detect.go` headers. Verified: emulator suite green locally (no GCP needed), full SaaS suite green against real BigQuery with zero snapshot diffs.


materialize-bigquery/bigquery.go
```diff
@@ -152,6 +160,12 @@ func (c config) effectiveBucketPath() string {
 
 func (c config) CredentialsClientOption() (option.ClientOption, error) {
 	if c.Credentials == nil {
+		// A custom endpoint with no credentials at all means an unauthenticated server, like an
+		// emulator. Credentials provided alongside an endpoint are still used, since the endpoint
+		// may be a proxy in front of the real authenticated service.
+		if c.CredentialsJSON == "" && c.Advanced.Endpoint != "" {
```
mdibaiee: `CredentialsJSON` is a deprecated field only kept for backwards-compatibility. It is not a secret and is thus persisted as plaintext and in general we must not revive it for another purpose. Let's re-use `Credentials.CredentialsJSON`

PR #4769, opened 2026-07-07
source-http-ingest: Update Flow dependencies

willdonnelly:
**Description:**

This is necessary so that builds of this connector pick up the "tolerate unknown JSON fields" change from Flow commit 184db0e, which in turn is necessary so that CI tests of https://github.com/estuary/flow/pull/3119 can pass since the Dekaf test uses `source-http-ingest:dev` and currently it errors out because that PR adds a new field which it doesn't recognize.

Part of https://github.com/estuary/flow/issues/2985


nicolaslazo (approved):
LGTM

PR #4770, opened 2026-07-07
materialize-databricks: runtime v2 scale-out

### LINKED ISSUE #4717: Materialization scale-out: Databricks

mdibaiee:
**Description:**

- Adds multi-shard support to materialize-databricks behind the `scale_out` feature flag
- checkpoint state is scoped per shard key-range and emitted as merge patches, and only the primary shard (`key_begin=0`) executes all shards' staged MERGE/COPY queries at `Acknowledge` — non-primary shards only stage files. To support this, `Transactor.Acknowledge` now receives the aggregated cross-shard state patches (always empty under v1; all other connectors ignore the parameter).

- The materialization integration test gains a `shard-count` knob (`RuntimeV2Config`) that runs it on the v2 runtime via `flowctl raw preview-next --fixture` — databricks always at 2 shards, postgres at 1 — reproducing the legacy snapshots byte-for-byte apart from v2's extra drain session (v2 runtime by default does not run a last Acknowledge when we finish a transaction. It runs the Acknowledge on the next txn's begin -- discussing with data-plane team to see if we can improve on this). Requires the multi-shard fixture support from estuary/flow#3154 (merged, in dev-next)

**Workflow steps:**

**Notes for reviewers:**


materialize-databricks/.snapshots/TestIntegration-materialize
```diff
@@ -1,12 +1,6 @@
...
...TRING COMMENT 'auto-generated projection of JSON at: /stringRelativeJsonPointerField with inferred types: [string]',\n  `stringTimeField` STRING COMMENT 'auto-generated projection of JSON at: /stringTimeField with inferred types: [string]',\n  `stringUint32Field` NUMERIC(38,0) COMMENT 'auto-generated projection of JSON at: /stringUint32Field with inferred types: [string]',\n  `stringUint64Field` NUMERIC(38,0) COMMENT 'auto-generated projection of JSON at: /stringUint64Field with inferred types: [string]',\n  `stringUriField` STRING COMMENT 'auto-generated projection of JSON at: /stringUriField with inferred types: [string]',\n  `stringUriReferenceField` STRING COMMENT 'auto-generated projection of JSON at: /stringUriReferenceField with inferred types: [string]',\n  `stringUriTemplateField` STRING COMMENT 'auto-generated projection of JSON at: /stringUriTemplateField with inferred types: [string]',\n  `stringUuidField` STRING COMMENT 'auto-generated projection of JSON at: /stringUuidField with inferred types: [string]'\n) COMMENT 'Generated for materialization acmeCo/tests/materialize-databricks of collection tests/data-types' TBLPROPERTIES ('delta.columnMapping.mode' = 'name');\n"]
```
mdibaiee, line 4: The new `flowctl raw preview-next` does not emit the Applied response for sharded runs. I think given that we do an Apply test in `TestIntegration-apply`, it is okay for this to be omitted here.

materialize-databricks/driver.go
```diff
@@ -461,29 +638,155 @@ func (d *transactor) Store(it *m.StoreIterator) (_ m.StartCommitFunc, err error)
...
+		}
 	}
 
-	for stateKey, item := range d.cp {
+	for rk, executed := range executedPeers {
+		var bucket = d.peerShardsCheckpoints[rk]
+		for _, stateKey := range executed {
+			if rk == legacyRangeKey {
+				clear[stateKey] = nil
+			} else {
```
mdibaiee, line 750: A lot of this can be removed after we have rolled out the new runtime v2 implementation to everyone.

go/materialize/transactor.go
```diff
@@ -62,18 +62,41 @@ type Transactor interface {
...
 	// the runtime, allowing the next pipelined transaction to begin to close.
-	Acknowledge(context.Context) (*pf.ConnectorState, error)
+	Acknowledge(ctx context.Context, statePatches []json.RawMessage) (*pf.ConnectorState, error)
 
 	// Destroy the Transactor, releasing any held resources.
 	Destroy()
 }
 
+// SplitStatePatches decodes a state_patches_json payload into its individual
+// RFC 7396 merge patches. The wire format is a JSON array whose elements are
```
jacobmarble, line 86: Neat, I didn't know about RFC 7396 before this.

materialize-databricks/driver.go
```diff
@@ -109,32 +112,134 @@ func newDatabricksDriver() *sql.Driver[config, tableConfig] {
...
+	templates             templates
+	materializationName   string
 }
 
 func (d *transactor) RecoverCheckpoint(_ context.Context, _ pf.MaterializationSpec, _ pf.RangeSpec) (m.RuntimeCheckpoint, error) {
 	return d.runtimeCheckpoint, nil
 }
 
 func (d *transactor) UnmarshalState(state json.RawMessage) error {
-	if err := pf.UnmarshalStrict(state, &d.cp); err != nil {
```
jacobmarble, line 130: You've relaxed "strict" JSON unmarshaling. Intentional?

materialize-databricks/driver.go
```diff
@@ -461,29 +635,146 @@ func (d *transactor) Store(it *m.StoreIterator) (_ m.StartCommitFunc, err error)
...
+	var patch, err = json.Marshal(rangeCheckpoints{d.rangeKey: d.cp})
+	if err != nil {
+		return nil, fmt.Errorf("creating checkpoint patch json: %w", err)
+	}
+
+	return &pf.ConnectorState{UpdatedJson: patch, MergePatch: true}, nil
 }
 
 // Acknowledge merges data from temporary table to main table
 // TODO: run these queries concurrently for improved performance
```
jacobmarble: Does this TODO still apply? Perhaps it applies to method `acknowledgeApply()` after this change?

jacobmarble (dismissed):

materialize-databricks/driver.go
mdibaiee: Intentional, and the per-item strictness is preserved. The old `pf.UnmarshalStrict(state, &d.cp)` decoded into `checkpoint` (a `map[string]*checkpointItem`), and `DisallowUnknownFields` is a no-op on a map — it only ever enforced strictness on each `checkpointItem` value. The state document is now heterogeneous (range-key buckets vs. legacy top-level stateKeys), so the top level has to be decoded as `map[string]json.RawMessage`, but every leaf still goes through `unmarshalStrict` (`DisallowUnknownFields`) via `parseCheckpointItem`/`parseRangeBucket`. So no field that was rejected before is silently accepted now — the relaxation is only at the container level, which never enforced anything.

jacobmarble (dismissed):

go/materialize/transactor.go
```diff
@@ -62,18 +62,41 @@ type Transactor interface {
...
+// each followed by a tab; tabs are JSON whitespace, so a standard decode
+// suffices. An empty payload means no patches.
```
danielnelson, line 88: This mostly is about the runtime protocol, but I feel like I'm missing something here.  What is the significance of the tab?  Why is this part of the protocol at all if its just standard JSON?  Is this some sort of extra promise for special parsers?

materialize-boilerplate/testutil/test_support.go
```diff
@@ -169,11 +178,19 @@ func RunMaterializationTestParallel[EC boilerplate.EndpointConfiger, FC boilerpl
...
+		skipUnlessRuntimeV2Flowctl(t)
+	}
+
 	ctx := context.Background()
 	tsSuffix := testItemIdentifier + fmt.Sprintf("%d", time.Now().Unix())
 
 	names, results := RunTestAllTasksParallel(t, sourcePath, func(t *testing.T, bundled []byte, taskName string, cfg EC) string {
+		if len(v2) > 0 {
+			return runMaterializationTestForTaskV2(t, ctx, newMaterializer, taskName, bundled, tsSuffix, makeResourceFn, actionDescSanitizers, v2[0])
+		}
```
danielnelson, line 185: In RunMaterializationTest we run the v2 test or the v1 test, but not both.  Here when we run the v2 test we also run the v1 test.  Should we have an if/else here?

materialize-boilerplate/testutil/test_support_v2.go
```diff
@@ -0,0 +1,214 @@
...
+	// Timeout is a backstop against a hung run; a healthy run ends by
+	// transaction count. Zero defaults to ten minutes.
+	Timeout time.Duration
+}
+
+// skipUnlessRuntimeV2Flowctl skips the test, rather than failing it, while
+// the flowctl release on PATH predates multi-shard fixture support in
+// `raw preview-next` (which v2 runs of any shard count depend on, for its
+// legacy-matching fixture document clocks).
+func skipUnlessRuntimeV2Flowctl(t *testing.T) {
```
danielnelson: Is there a reason we still want to support running tests with an older flowctl?  Perhaps could be useful to run with older flowctl so we can check for regressions?  Otherwise feels like some code that could go away, less code is better all else being equal, and it removes the chance that we think the test is running when its not.

materialize-boilerplate/testutil/test_support_v2.go
```diff
@@ -0,0 +1,214 @@
...
+	actionDescSanitizers []func(string) string,
+	v2 RuntimeV2Config,
+) string {
+	var snap strings.Builder
+
+	shards := max(v2.Shards, 1)
+	timeout := v2.Timeout
+	if timeout == 0 {
+		timeout = 10 * time.Minute
+	}
```
danielnelson, line 145: I wonder if we could incorporate `flag.Lookup("test.timeout")` here.  What happens to flowctl if the test timeout is hit, is it killed?  If so maybe this could be `test.timeout - 1m` to give some time to shutdown.

danielnelson (dismissed):
It is pretty if/else-y, I hope we can remove some of the code that allows us to revert back to the classic checkpoints soon so we can reduce the number of paths.  It seems like then we will be able to treat the connector as always in scale-out mode perhaps with a single shard.

materialize-boilerplate/testutil/test_support.go
mdibaiee: This is intentional, I want to have both v1 and v2 run for connectors that switch to v2 harness until a time when v1 is completely commissioned out

materialize-boilerplate/testutil/test_support_v2.go
mdibaiee: yeah this is vestigial from before the new preview-next being released, I'll clean it up

go/materialize/transactor.go
mdibaiee: https://github.com/estuary/flow/blob/74bfc1a608211056ea521b3ef5429cc656135376/crates/runtime-next/src/patches.rs#L99

Not necessarily for special parser, since \t is considered whitespace, normal JSON parses usually terminate on seeing the whitespace after a `{}`, so at least in the case of Go `json.Unmarshal`, it works fine with the new tab separation as well

I think the purpose is being able to have an append-only patch list growing in the runtime instead of having to re-parse and append to an existing array.

mdibaiee:
also ran the runtime v2 soak test on this and it worked all okay

jacobmarble (approved):

PR #4780, opened 2026-07-08
source-quickbooks: switch to "bring your own prod app" model

nicolaslazo:
**Description:**

The QuickBooks connector used to rely on an Estuary-hosted production application. While this worked initially, the API queries we run use a metered CorePlus endpoint that consumed our requests budget as adoption grew. Once the monthly budget was exhausted, access token refreshes would 429 until the month rolled over.

This commit switches to a new approach where users provide a client id, secret and refresh token for their own app. A single capture's API usage fits comfortably within the free tier.

This is a breaking change: all users will be required to apply for a new app before they can re-enable their tasks.

**Description:**

**Workflow steps:**

**Notes for reviewers:**

Tested with a local stack using a sandbox environment on an unpublished app. Documents are captured as expected.

### DIFF
--- a/source-quickbooks/source_quickbooks/models.py
+++ b/source-quickbooks/source_quickbooks/models.py
@@ -1,15 +1,15 @@
 from __future__ import annotations
 
 from datetime import UTC, datetime, timedelta
-from typing import TYPE_CHECKING, ClassVar
+from typing import ClassVar, override
 
-from estuary_cdk.capture.common import (
-    BaseDocument,
-    ResourceState,
-)
 from estuary_cdk.capture.common import (
     ConnectorState as GenericConnectorState,
 )
+from estuary_cdk.capture.common import (
+    ResourceState,
+)
+from estuary_cdk.capture.document import BaseDocument
 from estuary_cdk.flow import (
     OAuth2ClientCredentialsPlacement,
     OAuth2Spec,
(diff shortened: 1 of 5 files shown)


source-quickbooks/source_quickbooks/models.py
```diff
@@ -53,14 +53,29 @@ def default_start_date():
...
+    access_token: str = Field(
+        default="",
+        title="Access Token",
+        json_schema_extra={"secret": True, "x-hidden-field": True},
+    )
+    access_token_expires_at: datetime = Field(
+        default=EPOCH,
+        title="Access token expiration time.",
+        json_schema_extra={"x-hidden-field": True},
+    )
```
Alex-Bair, line 77: nit: My understanding is that immediately after a user creates a new capture, the connector will perform a token rotation & emit a `configUpdate` event containing the fresh set of tokens, including an actual `access_token` & its expiration. Could we describe that strategy in a comment around here? It's a little odd to see `access_token` and `access_token_expires_at` to have default values, but it makes sense to do that to force a rotation to happen the first time a capture receives an `Open` request.

source-quickbooks/source_quickbooks/resources.py
```diff
@@ -30,11 +30,40 @@
...
-        oauth_spec=OAUTH2_SPEC, credentials=config.credentials
-    )
+    credentials = config.credentials
+    http.token_source = TokenSource(oauth_spec=OAUTH2_SPEC, credentials=credentials)
+
+    # TokenSource never exchanges tokens on its own for
+    # RotatingOAuth2Credentials, so an absent or expiring access token must be
+    # exchanged here for the validation probe to have a usable token. Durable
+    # rotation happens at connector startup, which persists the updated tokens
+    # back to the endpoint config.
```
Alex-Bair: nit: Could we also mention here why it's ok to perform a token exchange in `validate_credentials` for `source-quickbooks`? It's intentional that the CDK's `TokenSource` never rotates tokens; if refresh tokens for the source system are strictly single use, a token rotation not followed by a `configUpdate` containing those rotated tokens could invalidate the refresh token persisted in the endpoint config & require the user to re-authenticate to un-break their capture. But that's not a concern with Quickbooks since refresh tokens it issues are not single use, & it'd be good to note the multi-use characteristic here so we remember later why these additional token rotations are ok.

Alex-Bair (dismissed):
LGTM % a couple optional documentation nits so we remember later that Quickbooks' refresh tokens can be used multiple times without invalidating them.

source-quickbooks/source_quickbooks/resources.py
nicolaslazo: Yeah, this is a weird system that they offer. I am more explicit about it now.

Alex-Bair (dismissed):
LGTM

Alex-Bair (approved):
LGTM

PR #4782, opened 2026-07-08
materialize-databricks: revert to direct COPY INTO for non-merge bindings

jacobmarble:
**Description:**

Reverts a FIXME in `materialize-databricks/driver.go` that forced `MERGE INTO` for every non-delta-updates binding, working around [databricks-sql-go#254](https://github.com/databricks/databricks-sql-go/issues/254). That issue was fixed upstream in driver v1.6.2 (Mar 2025); we've been on v1.10.0 since Mar 2026. Restores the original fast path: direct `COPY INTO` for delta-updates bindings and bindings with no existing keys to merge.

The migration integration test initially failed with `DELTA_FAILED_TO_MERGE_FIELDS` on fields whose type change has no migration support (NUMERIC(38,0)→LONG, STRING→DOUBLE). Per review discussion, that state is a test-harness artifact: Validate returns `INCOMPATIBLE` for those fields and the control plane resolves it by backfilling, but the test drives specs through `flowctl preview`, which bypasses that resolution and let the fields limp through Apply with a skewed column type. The fix here is per review suggestion: exclude those fields in the databricks migrate test spec. (History includes two superseded connector-side approaches — forced MERGE, cast-to-existing-type — and a reverted generic harness change that may return as a separate PR.)

**Workflow steps:**

No user-facing change. Standard bindings use `COPY INTO` instead of `MERGE INTO` when there's nothing to merge against.

None.

**Notes for reviewers:**

Net diff vs main: the driver.go FIXME revert, ten field exclusions in `testdata/migrate.flow.yaml`, and regenerated `materialize`/`migrate` snapshots. Full `TestIntegration` suite passes against the live integration warehouse with flowctl v0.6.10. CI is currently red fleet-wide due to #4790 (flowctl preview → runtime-next, fix: estuary/flow#3141), unrelated to this PR.


mdibaiee:
@jacobmarble the snapshots are outdated

mdibaiee:
The integration tests fail with
```
existing field type mismatch without migration support: existingType=DECIMAL field=intStr000ToInt000 mappedType=LONG
```

I did some more reading into this and this case has already been there for Databricks delta updates, we just don’t seem to do delta updates bindings for the migration integration test. We are just now exposed to this through standard updates tables with documents that do not require a merge… so it’s not a new bug/edge case introduced, it’s an existing one that we just uncovered through the tests.

It seems the reason this connector is different from others is that we explicitly map the types in the staged type to the DDL of the target table, which during a migration will be incorrect and fails, whereas other systems do an implicit COPY INTO that implicitly casts the type.

jacobmarble:
Integrated this diagnosis in 21de456fe: instead of forcing MERGE INTO for skewed bindings (previous commit), the staging casts are now rewritten to the column's *actual* type when the info schema shows an incompatible existing type — the staged `SELECT` casts to e.g. `NUMERIC(38,0)` instead of the spec's `LONG`, which is semantically what MERGE INTO's implicit write-casting did. Two details worth noting:

- The full type text is fetched via the Tables API for skewed columns only, because the info schema records just the bare type name and a bare `DECIMAL` cast means `DECIMAL(10,0)` — silent truncation.
- This also covers the pre-existing delta-updates hole you identified: delta bindings always used COPY INTO and had no MERGE path to fall back to, but they get the same cast override.

Validated with the full `TestIntegration` suite against the live warehouse (flowctl v0.6.10): `migrate` now runs the previously-failing COPY INTO shape with overridden casts and produces identical table data to main's MERGE-based behavior, no snapshot changes.

materialize-databricks/driver.go
```diff
@@ -223,6 +226,72 @@ type binding struct {
 	storeMergeBounds *sql.MergeBoundsBuilder
 }
 
+// overrideSkewedColumnCasts rewrites the DDL of columns whose existing
+// destination type is incompatible with their mapped type, so that staged
+// values are cast to the type the column actually has. This state arises when
+// Apply skips an unsupported column type migration ("existing field type
+// mismatch without migration support"): the staging queries would otherwise
```
mdibaiee: I can't really understand how we could have a case where Apply skips an unsupported column migration type, and we end up with incompatible types? if the type migration is skipped, shouldn't the types just stay as they were before?

mdibaiee: Okay, I read https://github.com/estuary/connectors/blob/main/materialize-boilerplate/materializer.go#L690-L709 again, and this is my understanding: During Apply, if we notice that a column's type has changed in the existing table, which is either done by an external process / user, or it is caused by a column migration that is still not fully completed, Apply does not bail out, mainly to support the case of an in-progress column migration. (see the commit message here: https://github.com/estuary/connectors/commit/796ff3d390726ca4299499675ea85ebed2874729)

Trying to state the full sequence of events that are described in [this comment](https://github.com/estuary/connectors/blame/main/materialize-boilerplate/materializer.go#L690-L709):

Connector is running with this collection spec:
```myField: { type: "string", "format": "integer" }```
Materialized as:
```myField: INTEGER```

Then we get a new Apply RPC that changes the spec:
```
myField: { type: ["string"] }
```
So it wants this column now:
```
myField: TEXT
```

So now Apply RPC sees a migratable type, and it tries to migrate the column using these statements:
```
ALTER TABLE `t` ADD COLUMN `myField_flowtmp1 TEXT`;
UPDATE `t` SET `myField_flowtmp1 = CAST(myField AS TEXT);
ALTER TABLE `t` DROP COLUMN `myField`;
ALTER TABLE `t` RENAME COLUMN `myField_flowtmp1` TO `myField`;
```

Now if another Apply is racing with this running Apply any point before the last two steps, and if this racing Apply has a different spec, this time to a column type that is not migratable from the original `INTEGER` type, e.g.:

```
myField: { type: ["string"], "contentEncoding": "base64" }
```
So now it wants:
```
myField: BINARY
```

The racing Apply RPC runs with this spec, and it sees that `myField: INTEGER` exists, and there is no `INTEGER -> BINARY` migration support. Instead of bailing out, this racing Apply will just log a warning saying it found an incompatible column in `myField` which could not be migrated and continues as if this Apply was successful. In the meantime the old Apply RPC is still trying to run the rename to make `myField: STRING`.

---------

Now to take a step back, this kind of sequence of events is only possible if Validate allows these type changes, which would entail a scenario where we have a task with spec A, and we allow it to transition to spec B (with a column migration), and we allow another transition from spec B to spec C, but the issue is that spec A could not transition to spec C with a column migration.

Essentially:

Column migration A -> B is possible
Column migration B -> C is possible
Column migration A -> C is not possible

Two racing Apply runs, from A -> B and B -> C running at the same time, can cause the second Apply from B -> C to still see the existing field as having type A, thus not migratable.

This is what I call the "real-world narrow edge case".

A real case of it in code is column migrations integer -> string and string -> binary are supported, but integer -> binary is not supported in databricks.

---------

Now the issue we are seeing in this case with the integration tests is that, for the integration tests, as part of the migration tests, we are running Validate, and Validate is telling us this migration is not supported, we get:
```
Type: pm.Response_Validated_Constraint_INCOMPATIBLE,
Reason: "Field 'intStr000ToInt000' is already being materialized as endpoint type 'DECIMAL' but endpoint type 'LONG' is required..."
```
In control plane this would normally cause the backfill counter to be incremented on this bindilg before Apply is called, but that doesn't happen in our integration tests, and that's why [this line](https://github.com/estuary/connectors/blame/main/materialize-boilerplate/materializer.go#L690-L709) gets triggered in Apply and just logs a warning and allows the task to proceed with Apply, even though it shouldn't. 

With all that said, I think the issue is really in the test harness, not in the connector(s) themselves: we should only exercise migrations for transitions that are supported by each connector, by first validating that they support it, and only Applying compatible changes, and just logging incompatible ones as part of the migration test snapshots.

I don't think connectors need to handle dealing with the real-world narrow edge case that I described above either, it just adds complexity without really solving anything in a maintainable manner.

In order to deal better with the real-world narrow edge case, if we really want to, we should just eliminate that narrow edge case, perhaps by supporting composable/associative migrations, e.g. if `x -> y` is possible and `y -> z` is possible, we should automatically generate the column migrations for `x -> z` by calling the templates associatively. If I understand the issue correctly, this should eliminate this scenario.

mdibaiee: @williamhbaker does my understanding of the real world narrow edge case make sense to you? and does the mitigation / possible solution I propose also make sense?

williamhbaker: Yeah that edge case description makes sense. I'm not entirely sure about the mitigation, I'd probably say just not to worry about it unless it is actually seen in the wild, and in that case the solution might just be to backfill the table. Or, maybe it would resolve itself naturally via the connector crashing and doing the `y -> z` migration by itself when it restarts.

This does seem like a test harness issue exactly as you described so the practical thing to do at this point would seem to be to fix that.

jacobmarble: Thank you, this is a very helpful analysis of the problem, and teaches me a bit about the control plane. I'll adjust the test harness to not worry about this particular real-world narrow edge case.

jacobmarble: Changing the test harness touches some other connectors, so I'll open a separate PR.

williamhbaker: Totally up to you ultimately on what makes sense, but I was thinking you could just exclude the field that doesn't support the migration in https://github.com/estuary/connectors/blob/main/materialize-databricks/testdata/migrate.flow.yaml#L21.

jacobmarble: Done in cb4193ff8 — went with this approach and reverted the generic harness change (de6574bc4): the ten unsupported-migration fields (`intStr*ToInt*` narrowing and `intStr*ToNum*` string→double) are now excluded in `migrate.flow.yaml` with a comment explaining why. Full integration suite passes against the live warehouse. Net diff vs main is now just the driver.go FIXME revert, the test spec exclusions, and regenerated snapshots.

jacobmarble: Well that does make this simpler, Will. I'll do that instead.

mdibaiee (approved):
Thanks

Can we also have a PR for the other connectors to skip their migrations they don't support?

The reason I liked doing Validate and doing the filter and snapshot ourselves to record the unsupported migrations was that it would be generic across the board, but I think agents can be good at detecting these scenarios and excluding them.

jacobmarble:
> Can we also have a PR for the other connectors to skip their migrations they don't support?
> 
> The reason I liked doing Validate and doing the filter and snapshot ourselves to record the unsupported migrations was that it would be generic across the board, but I think agents can be good at detecting these scenarios and excluding them.

Sure thing, I've opened an issue to follow up https://github.com/estuary/connectors/issues/4793

jacobmarble:
> > Can we also have a PR for the other connectors to skip their migrations they don't support?
> > The reason I liked doing Validate and doing the filter and snapshot ourselves to record the unsupported migrations was that it would be generic across the board, but I think agents can be good at detecting these scenarios and excluding them.
> 
> Sure thing, I've opened an issue to follow up #4793

And a PR https://github.com/estuary/connectors/pull/4794

PR #4794, opened 2026-07-10
materialize: exclude unsupported type migrations from migrate integration tests

### LINKED ISSUE #4793: Exclude unsupported type migrations from every connector's migration integration test
## Summary

Follow-up to #4782, requested by @mdibaiee in [his approving review](https://github.com/estuary/connectors/pull/4782#pullrequestreview-4674553392): apply the same treatment to the other materialize connectors — each connector's migration integration test should skip type transitions its dialect doesn't support.

## Background

The shared migrate fixture (`materialize-boilerplate/testdata/integration/collections.migrate-*.flow.yaml`) includes type transitions that not every dialect can migrate (e.g. NUMERIC(38,0)→LONG, STRING→DOUBLE for Databricks). For such fields, Validate returns an `INCOMPATIBLE` constraint, which the control plane resolves by backfilling the binding — but the test harness drives specs through `flowctl preview`, which bypasses that resolution. The result is Apply skipping the migration with a warning ("existing field type mismatch without migration support") and the test continuing with a column type the spec doesn't expect — a state the control plane never produces.

For most connectors this limps along via implicit casting in their store queries, but it is not a meaningful test of anything, and it hard-breaks connectors whose staging queries cast explicitly (how #4782 surfaced it, via `COPY INTO ... DELTA_FAILED_TO_MERGE_FIELDS`).

#4782 fixed this for Databricks only, with static exclusions in `materialize-databricks/testdata/migrate.flow.yaml` (see the comment there for the pattern).

## Scope

For each connector with a migrate integration test:

1. Determine which fixture transitions its dialect can't migrate (its `MigratableTypes` table, or just run the test and collect the "existing field type mismatch without migration support" warnings / Validate `INCOMPATIBLE` constraints).
2. Add those fields to the `fields.exclude` list in the connector's `testdata/migrate.flow.yaml`, with a comment noting why.
3. Regenerate the connector's migrate snapshot and run its full integration suite.

## Alternative

A generic harness-level version was built and then reverted in #4782 (`da7ee0981`, reverted by `de6574bc4`): the harness Validates the migrated collection between test phases, auto-excludes `INCOMPATIBLE` fields, and records them in the snapshot. Mahdi noted he liked that approach for being generic across the board; it could be resurrected here instead of maintaining per-connector static lists — the snapshot-regeneration sweep is the same either way.

jacobmarble:
**Description:**

Closes #4793

Implements #4793 (requested by @mdibaiee in his approving review of #4782): each materialize connector's migrate integration test now excludes type transitions its dialect can't migrate. For these fields Validate returns `INCOMPATIBLE`, which the control plane resolves by backfilling — but the test harness drives specs through `flowctl preview`, which bypasses that resolution and leaves Apply skipping the migration with a column type the spec doesn't expect.

Exclusions were discovered by running each connector's migrate test and collecting the `existing field type mismatch without migration support` warnings, then added to `fields.exclude` in `testdata/migrate.flow.yaml` with the same comment pattern as the Databricks fix in #4782. One commit per connector: postgres (5 fields), mysql/mariadb (14), sqlserver (20), spanner (5), iceberg (16, both spec files), bigquery (10), snowflake (6), redshift (3), motherduck (10), azure-fabric-warehouse (10).

mongodb, dynamodb, and clickhouse produced zero warnings — unchanged. s3-iceberg's migrate test is already disabled; databricks is covered by #4782.

**Workflow steps:**

No user-facing change; test-only.

None.

**Notes for reviewers:**

Every changed connector's full `TestIntegration` suite was run twice against live/local endpoints with `UPDATE_SNAPSHOTS=true`: first run regenerates the migrate snapshot, second run passes clean with zero mismatch warnings. Iceberg's `migrate.flow.yaml` (ICEBERG_TEST_ALL catalogs) got identical exclusions to `migrate-rest-local.flow.yaml` but only the default rest-local suite was run, which is what governs the committed snapshot.


mdibaiee (approved):
thanks!

PR #4797, opened 2026-07-13
materialize-s3-iceberg: parallelize appendFiles duplicate-file check

jacobmarble:
**Regression?**

Yes — #4347's iceberg-go duplicate-file check scans all manifests sequentially, one S3 GET at a time. On long-lived tables this exceeds the 5-minute appendFiles deadline and the shard crash-loops.

**Description:**

- Replace the built-in check with a connector-side `referencedFilePaths` that reads manifests concurrently, then call `AddFiles` with `ignoreDuplicates=true`.
- Matches pyiceberg's `check_duplicate_files` semantics (deleted entries ignored); duplicate hits stay retryable so the checkpoint fence resolves zombie double-appends.
- New `duplicate-append-regression` integration test.

**Notes for reviewers:**

Full suite passes locally, including Glue and snapshot parity; no snapshot changes.

### DIFF
--- a/materialize-s3-iceberg/catalog.go
+++ b/materialize-s3-iceberg/catalog.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/apache/iceberg-go"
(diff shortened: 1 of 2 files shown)


materialize-s3-iceberg/catalog.go
```diff
@@ -579,6 +609,65 @@ func (c *catalog) appendFiles(
...
+	}
+
+	var mu sync.Mutex
+	var referenced []string
+
+	group, groupCtx := errgroup.WithContext(ctx)
+	group.SetLimit(manifestReadConcurrency)
+	for _, m := range manifests {
+		group.Go(func() error {
+			if err := groupCtx.Err(); err != nil {
```
jacobmarble, line 648: FYI this is the essence of the PR - concurrently fetch manifests, which the sdk doesn't apparently do.

mdibaiee (approved):

PR #4803, opened 2026-07-14
materialize-s3-parquet/iceberg: return error instead of panicking on transient parquet sink write failure

### LINKED ISSUE #4801: materialize-s3-parquet/iceberg: shared parquet writer panics (not error) on transient S3 write failure
## 1. Priority
Low - no data loss, very low customer-facing impact (the shard auto-restarts within seconds; a customer would only notice a transient shard failure in their dashboard/logs, and a high-frequency cluster could cross the default `shardFailed` alert threshold, 3 failures per 8h, and trigger a Task Failed email). No manual workaround needed. Recurring across multiple tenants over 30+ days, and a low-risk fix is available (arrow-go already ships the non-panicking constructor), so worth fixing when convenient rather than urgently.

## 2. Scope and prevalence
Shared component: `go/writer/parquet.go`'s `NewParquetWriter`, used by both `materialize-s3-parquet` and `materialize-s3-iceberg`. Theoretical reach is any task on either connector that hits a transient write failure while uploading a file.

Observed prevalence (30-day Loki sweep, all 5 public data planes queried individually plus a private-plane sweep, literal string `failed to write magic number`): 6 distinct tenants, 13+ distinct tasks, 85 occurrences. Not a first occurrence - the earliest confirmed hit is 2026-06-13, three weeks before the arrow-go version bump in #4750 (2026-07-06), so this is not a new regression from that change.

## 3. Trigger (testable)
Any failure on the initial write to the sink `io.Writer` passed into `arrow-go/v18/parquet/file.NewParquetWriter` (writing the 4-byte Parquet magic header) causes a panic with the literal string `failed to write magic number`, instead of returning an error. In this connector, the sink is one end of an `io.Pipe` streaming to a cloud-storage upload; a transient network/write error on that pipe (e.g. `Broken pipe`) triggers it when the writer starts a NEW file, mid-materialization.

Minimal repro input (no customer data needed): construct `file.NewParquetWriter` (or the connector's `go/writer.NewParquetWriter` wrapper) with an `io.Writer` whose first `Write` call returns an error - observe the panic instead of a returned error.

## 4. Root cause + confidence
Confirmed (source diff) - arrow-go added a non-panicking alternative constructor, `file.NewParquetWriterWithError`, in the same version range that's currently pinned (diffed `v18.5.3-0.20260309043541` vs `v18.6.1-0.20260702190817` locally). The old `NewParquetWriter` is kept panicking intentionally, per its own updated doc comment: "preserved for backward compatibility with callers that string-match the panic value in a recover() block; new code should prefer `NewParquetWriterWithError`, which returns the failure as an error instead." This connector's `go/writer.NewParquetWriter` (go/writer/parquet.go) still calls the old panicking constructor.

## 5. Investigation steps / reproduction
- `flowctl catalog status` + `flowctl logs --since 5d` on an affected task -> 7 occurrences in 5 days, identical signature each time:
```
<ts>  i/o error writing to connector stdin: Broken pipe (os error 32)
<ts>  panic: failed to write magic number
        arrow-go/v18/parquet/file.NewParquetWriter (.../parquet/file/file_writer.go:84)
        go/writer.NewParquetWriter (parquet.go:~228)
        filesink.NewParquetWriter (format.go:~67)
```
- Confirmed the currently-pinned arrow-go version (`v18.6.1-0.20260702190817`) matches `origin/main`'s `go.mod` -> ruled out a stale local checkout before drawing conclusions.
- Downloaded and diffed the actual arrow-go module source for `v18.5.3-0.20260309043541` vs `v18.6.1-0.20260702190817` (`go mod download`, diffed the cached module directories) -> confirmed `NewParquetWriterWithError` was added in this range; confirmed no changes to JSON logical-type handling in the same diff (separate, unrelated customer question ruled out this way).
- Diffed connectors PR #4750 (the arrow-go version bump, merged 2026-07-06, fixes an unrelated memory leak, see #4736) -> confirmed it only touched `go/writer/parquet.go` (buffer/scratch-file handling) and `filesink/format.go`, not the constructor call site relevant here.
- `gh issue list --repo estuary/connectors --search "arrow.json OR JSON extension OR NewParquetWriter panic OR magic number"` -> no existing issue tracks this specific panic (best-effort search, not exhaustive).
- Fleet-wide Loki sweep, last 30 days, literal string `failed to write magic number`:
  - First pass batched all ~87 data planes under one query (2000-row cap) -> undercounted badly on public planes (a single public plane's combined multi-tenant log volume crowded out real matches within the row cap).
  - Re-queried each of the 5 public planes individually with a higher cap -> accurate per-plane counts.
  - Private planes: counts came from the original batched sweep, not independently re-verified per-plane - treat as a floor, not an exact number.
  - Result: 6 distinct tenants, 13+ tasks, 85 occurrences, earliest 2026-06-13 (three weeks before #4750) on one tenant's task, spread evenly through 2026-07-12 with no clustering around the version bump.

## 6. Suggested directions
1. Stopgap: none needed today - Flow's shard supervisor already auto-restarts the crashed connector within seconds; no ops or customer action required.
2. Data-preserving option: n/a - the panic happens before any file is committed, so no data loss occurs.
3. Platform/code direction: switch `go/writer.NewParquetWriter` (go/writer/parquet.go) to call `file.NewParquetWriterWithError` instead of `file.NewParquetWriter`, and propagate the returned error through the Store/startFile call path as a normal error instead of a panic.

## 7. Definition of done
- Acceptance: a Store call whose underlying upload writer fails its first Write returns a handled error visible in logs as a normal error (no panic/stack trace), and the shard fails cleanly or retries rather than crashing on an unrecovered panic.
- Regression test: construct the writer with an `io.Writer` whose first `Write` call errors; assert an error is returned, not a panic.
- Constraints: no change to the successful-write path or output file format/content.

## 8. References
- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1783998937710289
- Related, distinct root cause: estuary/connectors#4736 (arrow-go MinMax-stats memory leak, fixed by #4750)

mdibaiee:
## What

The shared parquet writer in `go/writer/parquet.go` called arrow-go's legacy `file.NewParquetWriter`, which **panics** with the literal string `failed to write magic number` when the initial write of the parquet magic header to the sink fails. In `materialize-s3-parquet` / `materialize-s3-iceberg` (and the other filesink parquet connectors), the sink is one end of an `io.Pipe` streaming to a cloud-storage upload, so a transient write failure (e.g. a broken pipe) while *starting a new file* crashed the shard with an unrecovered panic instead of surfacing a handleable error.

Flow auto-restarts the shard within seconds so there's no data loss, but each crash is a shard failure that can trip the `shardFailed` alert threshold on high-frequency tasks.

## Fix

Switch to arrow-go's non-panicking `file.NewParquetWriterWithError`, return the failure as a normal error, and propagate it through all callers:

- **`go/writer/parquet.go`** — `NewParquetWriter` now returns `(*ParquetWriter, error)`. Both the sink writer and the scratch-file writer (in `flushBuffer`) use the non-panicking constructor.
- **`filesink/format.go`** — forwards the error (covers `materialize-s3-parquet`, `materialize-gcs-parquet`, `materialize-azure-blob-parquet`).
- **`materialize-s3-iceberg/transactor.go`** — the `startFile` closure returns an error; on failure it calls `w.CloseWithError(err)` so the in-flight `PutStream` upload goroutine unblocks from the pipe and terminates rather than leaking.
- **`materialize-snowflake/bdec.go`** — propagates the error (its sink is an in-memory buffer, so this is just to satisfy the new signature).

The successful-write path and output file format/content are unchanged.

## Testing

- New regression test `TestParquetWriterSinkWriteErrorReturnsError`: constructs the writer with a sink whose first `Write` returns an error and asserts an error is returned rather than a panic.
- `go build ./...` and `go vet` pass across all affected packages; `go/writer` unit tests pass.

## Notes

- Root cause confirmed against the pinned arrow-go source (`v18.6.1-0.20260702190817`): `file.NewParquetWriter` now wraps `NewParquetWriterWithError` and re-panics for back-compat; its doc comment recommends new code use the error-returning variant.
- Not a regression from the arrow-go bump in #4750 — the earliest observed occurrence predates it.

Fixes #4801

### DIFF
--- a/filesink/format.go
+++ b/filesink/format.go
@@ -64,7 +64,7 @@ func NewParquetWriter(cfg ParquetConfig, b *pf.MaterializationSpec_Binding, w io
 	// compression.
 	opts = append(opts, writer.WithParquetCompression(writer.Snappy))
 
-	return writer.NewParquetWriter(w, sch, opts...), nil
+	return writer.NewParquetWriter(w, sch, opts...)
 }
 
 type CsvConfig struct {
--- a/go/writer/oomrepro_test.go
+++ b/go/writer/oomrepro_test.go
@@ -91,11 +91,14 @@ func TestOOMRepro(t *testing.T) {
 			stat["active_file"]/1024/1024, stat["inactive_file"]/1024/1024)
 	}
 
-	w := NewParquetWriter(oomReproDiscardCloser{}, sch,
+	w, err := NewParquetWriter(oomReproDiscardCloser{}, sch,
 		WithParquetRowGroupByteLimit(*oomReproRowGroupByteLimitMiB*1024*1024),
 		WithParquetRowGroupRowLimit(*oomReproRowGroupRowLimit),
 		WithParquetBufferSize(*oomReproBufferSizeMiB*1024*1024),
 	)
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	var totalBytes, maxRowsSeen int64
 	start := time.Now()
--- a/go/writer/pagecache_linux_test.go
+++ b/go/writer/pagecache_linux_test.go
@@ -130,7 +130,10 @@ func BenchmarkScratchPageCache(b *testing.B) {
 	b.ResetTimer()
 
 	for b.Loop() {
-		w := NewParquetWriter(discardCloser{}, sch)
+		w, err := NewParquetWriter(discardCloser{}, sch)
+		if err != nil {
+			b.Fatal(err)
+		}
 		inClose.Store(false)
 		for i := range numRows {
 			if err := w.Write([]any{int64(i), fmt.Sprintf("%08d-%s", i, payload)}); err != nil {
--- a/go/writer/parquet.go
+++ b/go/writer/parquet.go
@@ -201,7 +201,7 @@ func WithParquetMetadata(meta map[string]string) ParquetOption {
 	}
 }
 
-func NewParquetWriter(w io.WriteCloser, sch ParquetSchema, opts ...ParquetOption) *ParquetWriter {
+func NewParquetWriter(w io.WriteCloser, sch ParquetSchema, opts ...ParquetOption) (*ParquetWriter, error) {
 	cfg := parquetConfig{
 		compression:       Uncompressed,
 		bufferSize:        defaultBufferSize,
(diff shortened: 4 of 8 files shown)


dyaffe:
Would it make sense if we used this as opportunity to fork and pin the parquet writer that snowpipe streaming uses just to eliminate the possibility of mistakes?

jacobmarble (approved):

PR #4808, opened 2026-07-14
source-postgres: Default 2m statement_timeout

willdonnelly:
**Description:**

For historical reasons we defaulted to an unlimited statement timeout. This is probably not the right default behavior in the general case since it overrides any database-default timeouts _and_ an unlimited query timeout can cause issues in prod DBs.

After this change we will default to `2m` and if a user wants to let backfill queries run longer than that they are free to raise the timeout or set it to `0` for unlimited.

**Workflow steps:**

Users will now get a 2-minute timeout for all backfill queries by default. If the old behavior is desired, set `/advanced/statement_timeout` to `0`.

Documentation is updated in this PR. Do we need to sync docs changes over to the Flow repo these days or what? It looks like right now the docs exist in both locations.


Alex-Bair (approved):
LGTM

@jwhartley seems to be leading the charge around the docs move into this repo, so he likely has more intelligent input than I do on if the doc changes need made to the Flow repo too. I don't _think_ that we've actually cut over to serving documentation from this repo yet, and if it's going to be much longer before that cut over happens, we may want to keep both repos' documentation up to date.

Issue #4814, opened 2026-07-14
filesource: read parquet via byte-range requests instead of staging whole files to local disk


Alex-Bair:
# Problem

Parquet can't be decoded from a forward-only stream because its schema and layout metadata live in a footer at the end of the file. The parser's input is a pipe from the connector, so `Input::into_file()` copies the entire stream to a temp file before parsing begins (`crates/parser/src/input/mod.rs`, via `format/parquet.rs`). That means local disk required equals file size. Connector containers have no disk quota, so `/tmp` writes land on the host's shared container-storage volume. A large enough file kills the capture with the error `parsing failed: failed to read stream: No space left on device (os error 28)`.

We hit this in production. A capture crash-looped on a 72.76 GB parquet object against a 137 GB shared host volume. Every retry re-downloaded the full object, and the capture could never checkpoint past the file.

# Possible fix

Give the parser ranged access instead of a pipe. The connector passes a presigned GET URL through a new `--url` flag, and the parser implements the parquet `ChunkReader` over HTTP Range requests. It fetches the footer first, then row groups on demand. This keeps object-store SDKs and credentials out of the parser, and GCS/Azure signed URLs can slot in later. Peak resource usage drops from file size on disk to about one row group in memory.

See Slack threads https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1784054239764749 and https://estuaryworkspace.slack.com/archives/C03PF2FSM5H/p1784054845562169 for more context.

PR #4815, opened 2026-07-14
filesource: skip parquet files larger than 8GB

Alex-Bair:
**Regression?**

No.

**Description:**

Parquet files require random access and are staged to local disk in its entirety before parsing, so very large files can exhaust the shared scratch disk of the host. Skip parquet files larger than 8GB with a warning, as a temporary bandaid until the parser can read parquet without staging it to disk (see issue #4814).

Filesource connectors' documentation should be updated to reflect that parquet files larger than 8 GB are skipped.

### DIFF
--- a/filesource/filesource.go
+++ b/filesource/filesource.go
@@ -42,6 +42,16 @@ type Config interface {
 	PathRegex() string
 }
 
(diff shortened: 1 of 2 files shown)


skord (approved):
LGTM

PR #4819, opened 2026-07-15
materialize-s3-iceberg: self-heal manifest lists with null partitions

### LINKED ISSUE #4816: materialize-s3-iceberg: self-heal manifest lists with null `partitions` written by iceberg-go v0.6.0
## Background

iceberg-go v0.6.0 (adopted in #4347) had a manifest-list encoding regression, apache/iceberg-go#1309: when a fast-append commit inherited manifests from a snapshot written by a different producer (e.g. an external compaction/delete), a present-but-empty `partitions` field summary re-encoded as the **null** union branch. Strict readers reject the file:

```
ERROR: Wrong type in Avro file.
Field: partitions. Expected: 12. Got: 7
location: avro_utils.hpp:55   (Redshift Spectrum)
```

Athena does not error but can silently return NULLs for affected rows.

The upstream fix (apache/iceberg-go#1464, adopted in #4809) stops writing new nulls, but **deliberately preserves decoded null as null** (`ensurePartitionList` normalizes only present-but-nil). Because each commit re-encodes decoded inherited entries into a new manifest list, corruption written during the exposure window propagates forward on every commit and never heals on its own. One buggy-window commit typically nullified *all* inherited entries (>90% of the manifest list).

A fleet scan (2026-07-15) of tables written by the affected build found **86 corrupted tables across 9 active tasks / 5 tenants**, with more possible among tasks that couldn't be scanned externally (IAM-role-auth configs). Details are internal — see the incident Slack thread.

## Proposed fix: self-heal on Open

Every table this connector creates is unpartitioned by construction (`CreateTable` passes no partition spec — `catalog.go`), so for our tables a null `partitions` field in the current manifest list is unambiguous corruption: null and empty summaries are semantically identical, and normalizing cannot change query results.

On `Open`, once per table:

1. GET the current snapshot's manifest list (one small S3 read).
2. If any record has null `partitions`: decode with the file's **embedded writer schema**, flip null → empty array (change nothing else), re-encode with the same schema, and PUT back to the **same key**. The manifest list is referenced by exact path with no checksum, so no catalog commit is needed.
3. Otherwise, no-op.

Why this shape:

- **Race-free:** runs before transactions start, and the connector is the table's sole appender. (A concurrent external compaction just makes the repair moot — its Java-written manifest list is clean — and the check re-runs next Open.)
- **Self-extinguishing:** once one clean manifest list exists, subsequent commits inherit present-empty arrays and stay clean; the check degrades to one GET per table per Open, and the code can be removed after a fleet re-scan is clean.
- **Complete coverage:** heals role-auth and currently-disabled tasks without per-customer ops.
- Avoids `Transaction.RewriteManifests`, whose single-manifest bins pass through unchanged (`NoOpAlreadyOptimal`) and would silently skip the common one-bad-manifest case.

## Notes

- Only the current snapshot's manifest list needs repair; historical snapshots remain broken for time travel only.
- The invariant "null ⇒ corruption" holds only while all tables are unpartitioned — revisit if partitioned-table support lands.
- Regression test: write a manifest list containing null-`partitions` entries, run the Open-time repair, assert all entries decode as present arrays and the file's writer schema is unchanged.

jacobmarble:
**Regression?**

Yes — heals data corruption introduced by the iceberg-go v0.6.0 adoption in #4347 (apache/iceberg-go#1309). The write path was fixed in #4809, but manifest lists corrupted during the exposure window re-corrupt every subsequent commit and never heal on their own. Closes #4816.

**Description:**

Self-heal manifest lists whose `partitions` field summaries were encoded as the Avro null union branch, which strict readers (e.g. Redshift Spectrum) reject. On Open, for each bound table: GET the current snapshot's manifest list; if any entry has null `partitions`, flip it to an empty array and re-encode with the file's own embedded writer schema, codec, and metadata; PUT back to the same key (no catalog commit needed). Otherwise no-op. All tables this connector creates are unpartitioned, so null ⇒ corruption, and normalizing cannot change query results.

Commit 1 is a failing regression test that reproduces the corruption and its propagation; commit 2 is the fix.

**Workflow steps:**

None — the repair runs automatically at task startup and logs a warning per healed manifest list. Once one clean manifest list exists the check degrades to a single GET per table per Open, and the code can be removed after a fleet re-scan is clean.

None.

**Notes for reviewers:**

- The generic Avro map round-trip is deliberate: `ManifestFile.Partitions()` cannot distinguish null from present-but-empty, and iceberg-go's typed writer would substitute its own schema for the file's. The test asserts the writer schema is byte-identical after repair.
- Verified against both the Polaris REST and AWS Glue integration suites (all subtests green, no snapshot drift).

**Checklist:**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -79,6 +79,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	github.com/tidwall/gjson v1.17.3
 	github.com/tidwall/sjson v1.2.5
+	github.com/twmb/avro v1.7.2
 	github.com/wk8/go-ordered-map/v2 v2.1.8
 	go.gazette.dev/core v0.103.1-0.20260505142537-ce718ada37ee
 	go.mongodb.org/mongo-driver v1.17.7
@@ -258,7 +259,6 @@ require (
 	github.com/substrait-io/substrait-protobuf/go v0.85.0 // indirect
 	github.com/tidwall/match v1.1.1 // indirect
 	github.com/tidwall/pretty v1.2.1 // indirect
-	github.com/twmb/avro v1.7.2 // indirect
 	github.com/twmb/murmur3 v1.1.8 // indirect
 	github.com/twpayne/go-geom v1.6.1 // indirect
 	github.com/vmihailenco/msgpack/v5 v5.4.1 // indirect
--- a/materialize-s3-iceberg/CHANGELOG.md
+++ b/materialize-s3-iceberg/CHANGELOG.md
@@ -1,0 +1,11 @@
+# Changelog
+
+## 2026-07-15
+
+### Fixed
+- Tables affected by a manifest-list encoding defect in a previous connector
+  release are now repaired automatically when the materialization starts up.
+  The defect could cause external query engines to reject a table's current
+  snapshot (Redshift Spectrum: `Wrong type in Avro file. Field: partitions`)
+  or to silently return NULL values for affected rows (Athena). No action is
+  needed; the connector logs a warning for each table it repairs.
--- a/materialize-s3-iceberg/catalog.go
+++ b/materialize-s3-iceberg/catalog.go
@@ -332,16 +332,32 @@ func (c *catalog) populateInfoSchema(ctx context.Context, is *boilerplate.InfoSc
 	return nil
 }
 
-// tablePaths returns the registered storage path for each resource path in a
-// list, in the same order as the input.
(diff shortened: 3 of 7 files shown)


mdibaiee:
@jacobmarble how long do we plan on keeping this? I think it is best not to have this code on `main` unless we know there are lots of tasks affected. If only a few tasks are affected, I would suggest we approve, deploy the built image for those tasks to clean them up, and then back to latest version

What do you think?

mdibaiee (approved):

PR #4823, opened 2026-07-16
Obfuscation scripts for collection documents and catalog specs

mdibaiee:
Scripts to produce customer-safe copies of Estuary collection data and catalog specs (for repros, debugging, fixtures). Everything lives in `scripts/obfuscation/`.

## What
- **`obfuscate_jsonl.py`** — obfuscates **every** value in a collection-document JSONL, at every depth, **except `/_meta/uuid` and `/_meta/op`** (the collection key and `/_meta/source` are obfuscated too). Deterministic + shape-preserving (same value+salt → same output; strings keep length/char-class, numbers keep magnitude, RFC3339 stays valid). No catalog needed.
- **`strip_catalog.py` + `config_allowlist.txt`** — strips connector configs down to an **allow-list** of customer-agnostic knobs and removes everything else, recursively, across every connector/`local` config and binding `resource` in captures, materializations, and derivations.

## Reviewer notes (the part worth scrutiny)
- **`config_allowlist.txt` is the security surface.** It's **strip-by-default**: a config value survives only if its leaf key name is listed, so the output is safe-by-construction even for config fields nobody has classified yet — the risk is only ever "stripped a harmless knob", never a leak. The list was derived from a survey of every Go + Python connector config schema and deliberately excludes revealing/over-generic keys (`auth_type`, `catalog_type`, `compute_type`, `method`, `mode`, `is_sandbox`, `timezone`, `region`, `name`, `limit`, `days`, `start_date`, cursor/host/credential/schema/table fields). Please sanity-check entries against your connector knowledge; **re-review it as connectors add config fields.**
- Nested handling: an agnostic knob under an otherwise-stripped block (e.g. `advanced.sslmode`, `advanced.feature_flags`) is kept while identifying siblings (e.g. `advanced.skip_backfills`) are dropped.

## Note
An earlier revision of this PR explored obfuscating connector **logs** (a `go/logsanitize` sentinel-marker package + a log-obfuscation script). That direction was dropped — it's been fully reverted, so this PR now touches only `scripts/obfuscation/` and no connector/Go code. (The earlier "markers survive Gazette + flowctl" validation comment refers to that removed approach.)


mdibaiee:
## Validation: markers survive Gazette + flowctl round-trip

Verified end-to-end on a local Flow stack (Lima VM: full control plane + `local-flow-cluster` dataplane) that the `logsanitize` sentinels (U+E000 / U+E001) survive persistence and read-back, and that `scripts/obfuscation/obfuscate_logs.py` works on the retrieved logs.

**Setup:** added a temporary probe log to `source-hello-world` using the real helpers —
`log.WithFields{email: logsanitize.Quoted("jane.doe@example.com"), created_at: logsanitize.Quoted("2020-01-02T03:04:05Z")}.Info("...probe for user " + logsanitize.Quoted("Jane Doe"))` —
built it as a `:local` image, published a capture, then read the task's ops logs with `flowctl logs --task test/hello-world`.

**Round-trip integrity:** the read-back ops-log line preserved the markers byte-for-byte through connector stdout → reactor ingestion → Gazette journal → `flowctl logs` (JSON encoding, journal storage, and flowctl's render all keep the PUA code points):

```
{"fields":{"created_at":"⟦"2020-01-02T03:04:05Z"⟧","email":"⟦"jane.doe@example.com"⟧"},
 "level":"info","message":"logsanitize round-trip probe for user ⟦"Jane Doe"⟧",
 "ts":"2026-07-16T19:13:18Z", ...}      (⟦ ⟧ = U+E000 / U+E001)
```

**Obfuscation on the real logs (`obfuscate_logs.py`, default mode):**
- Age gate **passed with no override** — the logs' `ts` is after `MARKING_RELEASED`.
- `"2020-01-02T03:04:05Z"` → `"2025-08-06T22:18:16Z"` (valid RFC3339), `"jane.doe@example.com"` → `"pmpt.rvv@qyxkmcp.mir"` (length + character classes preserved), `"Jane Doe"` → `"Koxz Znk"`.
- Non-marked fields (`uuid`, `level`, `shard`, `ts`) untouched; **zero markers left** in the output.
- The datetime obfuscated to the same value as standalone tests (same salt) — deterministic cross-artifact consistency held through the real pipeline.

**Note for local-stack users on Apple Silicon:** the local dataplane runs connectors as **amd64 via Rosetta**. The data-plane link initially failed to build because `ghcr.io/estuary/derive-typescript:dev` had been pulled as arm64; re-pulling the amd64 variant fixed it. (This is why `build-local.sh` forces `--platform linux/amd64`.) Unrelated to this change.

jacobmarble (dismissed):

mdibaiee:
Update: the connector-**log** obfuscation approach (go/logsanitize markers + obfuscate_logs.py) has been dropped and fully reverted. This PR now covers only document + catalog-spec obfuscation under `scripts/obfuscation/`. The earlier validation comment above (markers surviving Gazette + flowctl) applies to that removed approach and is no longer relevant.

scripts/obfuscation/obfuscate_jsonl.py
```diff
@@ -0,0 +1,140 @@
...
+    chars = []
+    for ch in s:
+        if ch.islower():
+            chars.append(r.choice(string.ascii_lowercase))
+        elif ch.isupper():
+            chars.append(r.choice(string.ascii_uppercase))
+        elif ch.isdigit():
+            chars.append(r.choice(string.digits))
+        else:
+            chars.append(ch)  # punctuation / whitespace / separators kept
```
jacobmarble: CJKV, Hebrew, Arabic, Thai are all caseless, so they won't be obfuscated at all, correct?

scripts/obfuscation/config_allowlist.txt
jacobmarble, line 1: Perhaps this file needs a special mention in CLAUDE.md, to encourage reasonable new entries as needed.

jacobmarble: Perhaps mention that only scalar-valued keys can be allow-listed, and in order to allow-list an entire object, each key must be added here.

jacobmarble (dismissed):

scripts/obfuscation/obfuscate_jsonl.py
mdibaiee: Good catch — confirmed they were passing through unobfuscated. Fixed: any character whose Unicode category is a letter (L*), a non-decimal numeric (N*), or a non-ASCII symbol/emoji (S* above U+007F) now maps to a deterministic CJK ideograph, so every content char is obfuscated. ASCII punctuation/symbols, whitespace, and combining marks are kept as structure. Added `test_obfuscate_jsonl.py` covering Arabic, Hebrew, Mandarin, Japanese, Korean, Thai, Devanagari, Greek, Cyrillic, accented Latin, emoji (incl. ZWJ/flags), and symbols with per-character coverage assertions.

scripts/obfuscation/config_allowlist.txt
mdibaiee: Done — added a note to CLAUDE.md (under JSON Schema) telling contributors to add customer-agnostic config fields here, and noting that leaving one out only makes sanitized configs less complete, never unsafe.

mdibaiee: Added to the header: only scalar leaves (and lists of scalars) are kept, matched by their own key name; listing an object's key has no effect, so to keep a nested object's contents you add each scalar key inside it individually.

jacobmarble (approved):

PR #4824, opened 2026-07-16
source-dynamics-365-finance-and-operations: additional debug logging

Alex-Bair:
**Regression?**

No.

**Description:**

When troubleshooting captures that hit 404 errors when reading a `model.json` or `.csv` file that _should_ exist since we were able to list the timestamp folder previously, it'd be good to know if the list of timestamp folders is consistent across runs or if Azure is deleting timestamp folders as part of some data retention policy. This debug log will help us figure out if that's the case.


nicolaslazo (approved):
LGTM

PR #4825, opened 2026-07-16
source-postgres: Disable statement timeout for slot creation

willdonnelly:
**Regression?**

https://github.com/estuary/connectors/pull/4808

**Description:**

Replication slot creation blocks until a consistent restart_lsn is established, which means that a long-running transaction on the server can make it take arbitrarily long and if we run into the statement timeout during that blocking the creation fails.

Since there's nothing else we need to get back to while we're in the process of recreating the slot, we should just be as patient as we need to be for this query.

**Workflow steps:**

No user action required

**Checklist:**

### DIFF
--- a/source-postgres/CHANGELOG.md
+++ b/source-postgres/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 2026-07-16
+
+### Fixed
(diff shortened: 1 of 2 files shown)


Alex-Bair (approved):
LGTM

PR #4826, opened 2026-07-16
source-dynamics-365-finance-and-operations: skip timestamp folders missing `model.json`s

Alex-Bair:
**Regression?**

No.

**Description:**

We've observed D365 F&O capture connector tasks encounter older timestamp folders that do not have a model.json present. The model.json file contains the schema necessary for reading the entries in the various entities' headerless CSV files, so the connector has been crashing on this error.

After some research, it appears that when a Synapse Link export (i.e. what generates these timestamp folders) is aborted for some reason, no model.json is written to the timestamp folder & Synapse Link retries exporting the same data later. Meaning, we should treat 404s when attempting to read `model.json`s from older timestamp folders as signal to skip that folder since the export wasn't finalized & we _should_ see the data that Synapse Link attempted to export in a later timestamp folder. This PR updates the connector to do just that.

`fetch_changes` now gives a folder `SETTLE_DELAY` after it closes (approximated by its successor's creation time) for the model.json to appear. If it's still missing, the folder is treated as abandoned and skipped with a warning. Its changes are expected to be re-committed in a later folder. A folder that closed recently is left for the next sweep in case it's still finalizing.

This is safe per Microsoft's documented behavior:

- The retry/failure machinery can leave anomalous folders: "Due to the retry mechanism features, an extra empty timestamp folder might be created." https://learn.microsoft.com/en-us/power-apps/maker/data-platform/azure-synapse-incremental-updates

- On an unacknowledged commit Synapse Link re-commits the data, and consumers are expected to be resilient to the duplicate: "if Azure Synapse Link for Dataverse doesn't get an acknowledgment from the Azure data lake that the data has been committed ... Azure Synapse Link retries in those scenarios and commit the data again. The downstream consumption should be made resilient to this scenario by filtering data using SinkModifiedOn." The connector's last-write-wins reduce on /Id (ordered by versionnumber) provides that resiliency. https://learn.microsoft.com/en-us/power-apps/maker/data-platform/export-data-lake-faq

**Checklist:**

### DIFF
--- a/source-dynamics-365-finance-and-operations/CHANGELOG.md
+++ b/source-dynamics-365-finance-and-operations/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Changelog
+
+## 2026-07-16
+
+### Fixed
+- Captures no longer crash when an older timestamp folder is missing its `model.json`. Such folders come from Synapse Link exports that never finalized; they are now skipped once enough time has passed for finalization, and their data is picked up from the later folder Synapse Link re-commits it to.
--- a/source-dynamics-365-finance-and-operations/source_dynamics_365_finance_and_operations/api.py
+++ b/source-dynamics-365-finance-and-operations/source_dynamics_365_finance_and_operations/api.py
@@ -1,6 +1,6 @@
 import asyncio
 from bisect import bisect_right
-from datetime import datetime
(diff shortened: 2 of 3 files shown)


nicolaslazo (approved):
LGTM

PR #4827, opened 2026-07-16
sqlcapture: dump goroutine stacks during long-running streaming cycles

Alex-Bair:
**Regression?**

No.

**Description:**

We've observed a `source-mysql` capture stall for 12+ hours, logging out that replication streaming is taking an unexpectedly long time every so often with suspiciously zero data movement. In this situation, it's difficult to tell from the logs why the capture is apparently stalled. This PR hangs a full goroutine dump off the existing progress ticker on the first tick and every fifth tick thereafter. This should help us investigate stalls even if an operator isn't around to get a goroutine stack dump before the capture is restarted & un-stuck.

**Checklist:**

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -1,6 +1,7 @@
 package sqlcapture
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
LGTM

willdonnelly (approved):
LGTM

PR #4831, opened 2026-07-17
source-dynamics-365-finance-and-operations: use fallback per-table `model.json` when folder level `model.json` omits the table

Alex-Bair:
**Regression?**

No.

**Description:**

This PR's scope includes:
- Updating the CDK's `IncrementalCSVProcessor` to raise an exception when there's a column mismatch between a CSV's headers and actual rows.
- Falling back to use `{timestamp}/Microsoft.Athena.TrickleFeedService/{table}-model.json` for table metadata when a timestamp folder's `model.json` is present, lists entities, but omits a given one. I suspect these individual per-table `model.json`s are what feed into the folder level `model.json` - during testing, parsing CSVs with these per-table `model.json`s worked fine & produced rows with all the anticipated columns. I'm not sure why some entities can be missing from the folder level `model.json`, but I suspect it's an unusual situation where Synapse Link almost finished writing everything to the folder level `model.json` but was interrupted for some reason.

**Notes for reviewers:**

Tested with an account where the folder level `model.json` was missing three entities, causing the capture to fail. The capture now fallsback to the per-table `model.json`s, parses the CSV rows successfully for those entities, and moves on.

**Checklist:**

### DIFF
--- a/estuary-cdk/estuary_cdk/incremental_csv_processor.py
+++ b/estuary-cdk/estuary_cdk/incremental_csv_processor.py
@@ -68,6 +68,14 @@ class CSVProcessingError(Exception):
         super().__init__(message)
 
 
+# Sentinels for column-count validation. aiocsv's dict reader collects surplus
+# cells under `restkey` and pads short rows with `restval`. Using unique objects
+# for both lets us detect either mismatch without mistaking a genuinely empty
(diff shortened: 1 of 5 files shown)


nicolaslazo (approved):
LGTM 👍

PR #4836, opened 2026-07-17
materialize-clickhouse: support custom PARTITION BY expression

### LINKED ISSUE #4832: materialize-clickhouse: support custom PARTITION BY per table
ClickHouse MergeTree-family tables support a [custom partitioning key](https://clickhouse.com/docs/engines/table-engines/mergetree-family/custom-partitioning-key) (`PARTITION BY <expr>`). Customers need a way to set this per table through the connector; today all tables are created with ClickHouse's default single partition.

## Proposed UX

A new optional `partition_by` string field on the per-binding resource config. The user supplies just the expression — e.g. `toYYYYMM(flow_published_at)` or `region, toYYYYMM(ts)` — and the connector renders it as `PARTITION BY (<expr>)` in `CREATE TABLE`, for both standard (ReplacingMergeTree) and delta-updates (MergeTree) bindings. Empty/absent means no `PARTITION BY` clause, preserving current behavior.

## Semantics

- **Changing the value requires a backfill.** ClickHouse only accepts `PARTITION BY` at `CREATE TABLE` time — it can't be altered. Validate will return `INCOMPATIBLE` constraints when `partition_by` changes on an existing binding without a backfill (same pattern as materialize-spanner's key-distribution setting), and the backfill will drop + re-create the table (via `MustRecreateResource`) rather than truncate it.
- **Validation is a dry-run.** The expression is too open-ended to check statically, so Validate creates a scratch-named table with the binding's real column DDL and the user's expression, then drops it — surfacing server errors (bad syntax, missing columns, non-deterministic functions) at publish time instead of at Apply.
- The internal store staging tables are created `AS <target>`, so they inherit the partition key, and the commit path already moves each `_partition_id` individually — multi-partition transactions work without commit-path changes.

## Adjacent fix required

The store-table schema-drift check added in ac900a820 compares columns only, not the partition key. After a backfill re-creates the target with a different `PARTITION BY`, the persistent store table would keep the old key and every `ALTER TABLE ... MOVE PARTITION` would fail ("tables have different partition key") in a permanent loop. The drift check needs to also compare `system.tables.partition_key` so the store table is re-created on mismatch.

jacobmarble:
**Description:**

Adds an optional `partition_by` string to materialize-clickhouse's per-binding resource config, setting a [custom partitioning key](https://clickhouse.com/docs/engines/table-engines/mergetree-family/custom-partitioning-key) on created tables (e.g. `toYYYYMM(flow_published_at)`). Closes #4832.

- The expression is spliced as `PARTITION BY (<expr>)` into `CREATE TABLE`, for both standard and delta bindings; store staging tables inherit it via `CREATE TABLE ... AS`, and the existing per-partition MOVE loop commits multi-partition transactions unchanged.
- Since the expression is too open-ended to check statically, Validate dry-runs it: creates and drops a scratch table with the binding's real column DDL, surfacing server errors at publish time.
- ClickHouse only accepts PARTITION BY at table creation, so changing the value requires a backfill: Validate returns `INCOMPATIBLE` constraints until the binding is backfilled (materialize-spanner pattern), and `MustRecreateResource` makes the backfill drop+recreate instead of truncate.
- Shared change: `sql.TableShape` now carries the binding's raw `ResourceConfigJson`, so dialect templates can render resource-level options. Additive; no other connector affected.
- Adjacent fix: the store-table drift check from the pending-commit recovery fix (ac900a820) now also compares partition keys, since a backfill into a new partition key would otherwise crash-loop on `MOVE PARTITION`.

**Workflow steps:**

Set `partition_by` (advanced) on a binding's resource config. To change it later, backfill the binding; the table is dropped and re-created.

materialize-clickhouse connector docs should mention the new field.

**Notes for reviewers:**

- Existing sqlgen snapshots pass byte-identical with `partition_by` unset.
- New integration tests cover multi-partition commits, dry-run accept/reject with scratch cleanup, INCOMPATIBLE-on-change, backfill recreate via Apply, and store-table drift recovery.

**Checklist:**

### DIFF
--- a/materialize-clickhouse/client.go
+++ b/materialize-clickhouse/client.go
@@ -282,7 +282,24 @@ func (c *client) AlterTable(ctx context.Context, ta sql.TableAlter) (string, boi
 	}
 
 	if len(ta.ColumnTypeChanges) > 0 {
-		var steps, err = sql.StdColumnTypeMigrations(ctx, c.ep.Dialect, ta.Table, ta.ColumnTypeChanges, clickHouseMigrationSteps...)
+		// A column referenced by the table's PARTITION BY key cannot change
+		// type at all: ClickHouse forbids both MODIFY COLUMN (code 524) and
+		// DROP COLUMN (code 47) on partition-key columns, so the multi-step
+		// rename migration can never succeed. Fail before executing any step,
+		// so no temporary column is left behind and the error says what to do.
+		partitionKeyCols, err := c.partitionKeyColumns(ctx, ta.Path[0])
+		if err != nil {
+			return "", nil, err
+		}
+		for _, m := range ta.ColumnTypeChanges {
+			if partitionKeyCols[m.Field] {
+				return "", nil, fmt.Errorf(
+					"cannot change the type of column %q: it is referenced by the PARTITION BY key of table %s, which ClickHouse forbids altering; backfill the binding to re-create the table with the new column type",
+					m.Field, ta.Identifier)
+			}
+		}
+
+		steps, err := sql.StdColumnTypeMigrations(ctx, c.ep.Dialect, ta.Table, ta.ColumnTypeChanges, clickHouseMigrationSteps...)
 		if err != nil {
 			return "", nil, fmt.Errorf("rendering column migration steps: %w", err)
 		}
@@ -297,6 +314,29 @@ func (c *client) AlterTable(ctx context.Context, ta sql.TableAlter) (string, boi
 		}
 		return nil
 	}, nil
+}
+
+// partitionKeyColumns returns the set of column names referenced by the
+// table's PARTITION BY key.
+func (c *client) partitionKeyColumns(ctx context.Context, table string) (map[string]bool, error) {
+	rows, err := c.db.QueryContext(ctx, fmt.Sprintf(
+		"SELECT name FROM system.columns WHERE database = %s AND table = %s AND is_in_partition_key = 1",
+		c.ep.Dialect.Literal(c.ep.Config.Database), c.ep.Dialect.Literal(table),
+	))
+	if err != nil {
+		return nil, fmt.Errorf("querying partition key columns of %q: %w", table, err)
+	}
+	defer rows.Close()
+
+	var out = make(map[string]bool)
+	for rows.Next() {
+		var name string
+		if err := rows.Scan(&name); err != nil {
+			return nil, fmt.Errorf("scanning partition key column of %q: %w", table, err)
+		}
+		out[name] = true
+	}
+	return out, rows.Err()
 }
 
 func (c *client) TruncateTable(_ context.Context, path []string) (string, boilerplate.ActionApplyFn, error) {
--- a/materialize-clickhouse/client.go
+++ b/materialize-clickhouse/client.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	stdsql "database/sql"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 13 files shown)


materialize-clickhouse/validate.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return fmt.Errorf("rendering dry-run CREATE TABLE: %w", err)
+	}
+
+	_, createErr := db.ExecContext(ctx, createSQL)
+	if _, err := db.ExecContext(ctx, fmt.Sprintf("DROP TABLE IF EXISTS %s;", dialect.Identifier(scratch))); err != nil {
+		log.WithFields(log.Fields{"table": scratch, "error": err}).Warn("failed to drop partition_by dry-run table")
+	}
+	return createErr
+}
```
mdibaiee, line 134: If we don't have this dry-run, and instead we just attempt to create the target table with the expression and it fails, is it any different to failing it on the scratch table? so I guess the question is: what is the risk of failing on the target table instead of this scratch table?

mdibaiee: ah I see, this is done in Validate RPC...

I wonder if this is a paradigm shift to make temporary writes to the destination during Validate. I think we have avoided any writes to destinations during Validate -- doing some research to see if we have ever done similar temporary writes.

mdibaiee: Okay did a check, we do already have similar kind of "write and cleanup" pattern in some other connectors so 👍🏽

materialize-clickhouse/validate.go
```diff
@@ -0,0 +1,207 @@
...
+			continue
+		}
+		out.Values = append(out.Values, p.Field)
+	}
+	return out
+}
+
+// findLastBinding mirrors the unexported helper of materialize-boilerplate: a
+// previously disabled binding still constrains the current one, so inactive
+// bindings are consulted as well.
```
mdibaiee: can we instead export that helper from materialize-boilerplate?

materialize-clickhouse/client.go
```diff
@@ -317,8 +358,21 @@ func (c *client) ExecStatements(ctx context.Context, statements []string) error
...
+		return false, nil
+	}
+	var lastRC, newRC tableConfig
+	if err := json.Unmarshal(lastBinding.ResourceConfigJson, &lastRC); err != nil {
+		return false, fmt.Errorf("parsing last binding resource config: %w", err)
+	}
+	if err := json.Unmarshal(newBinding.ResourceConfigJson, &newRC); err != nil {
+		return false, fmt.Errorf("parsing new binding resource config: %w", err)
+	}
+	return strings.TrimSpace(lastRC.PartitionBy) != strings.TrimSpace(newRC.PartitionBy), nil
```
mdibaiee, line 375: Can this cause an automatic backfill of the table? If yes, is that a desirable outcome? often when a config update by a user requires a backfill, we require explicit backfill because the user is already changing the config, and they may not be aware of the backfill (even when we document it)

jacobmarble: Clickhouse does not permit changing the partitioning of an existing table - partitioning is immutable.

It is possible to bulk copy from one table to another with `INSERT INTO new SELECT * FROM old` but the operation is slow and non-atomic, so it needs application-side batch logic and retry logic. It is possible, but (1) the code would be rarely used in anger, so I worry that it would be fragile and (2) it isn't clear that backfill is actually slower than this.

The custom `Validate` logic yields INCOMPATIBLE constraints when `partition_by` is changed without a backfill bump. This blocks the publication until the user explicitly backfills the binding.

materialize-clickhouse/validate.go
jacobmarble: Done

jacobmarble:
I've applied `docs-check-skip` because the change touches materialize-sql and materialize-boilerplate without updating corresponding CHANGELOG files.

mdibaiee (approved):

mdibaiee (approved):

mdibaiee (approved):

mdibaiee (approved):

mdibaiee (approved):

PR #4839, opened 2026-07-19
source-sqs: initial connector

### LINKED ISSUE #4534: source-sqs: new connector
We need to build a capture connector to capture data from Amazon [Simple Queue Service](https://aws.amazon.com/sqs/) (SQS).

Primary requirement is that it's able to handle 50MM messages per hour (so ~14k messages per second).

See this [Slack thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1777047990660669) for additional context & details.

Alex-Bair:
**Regression?**

No.

**Description:**

This PR introduces a new capture connector `source-sqs` that captures messages from AWS SQS queues, both standard and FIFO. Messages are deleted from SQS after Flow durably commits them, via the capture protocol's explicit acknowledgements, so every failure mode degrades to an at-least-once duplicate rather than data loss. Duplicates collapse under the discovered collection keys, `/_meta/messageId` for standard queues and `(/_meta/messageGroupId, /_meta/sequenceNumber)` for FIFO, which also preserve per-group queue order.

A single shard run on my dev GCP VM sustains ~37M messages/hour measured against an actual AWS SQS instance, with ~50M/hour projected in-region. Integration tests run against LocalStack and are wired into CI.

The overview and the SQS behaviors that shaped the design are in `source-sqs/README.md`, internals and maintenance invariants in `ARCHITECTURE.md`, and measured results with sizing rationale in `BENCHMARKS.md`.

Documentation should be created for this connector. Documentation for the ATF `source-amazon-sqs` connector should be updated to reflect that it's deprecated.

**Notes for reviewers:**

I squeezed as much throughput out of a single shard as I thought was reasonable at this point. IMO efforts to improve throughput further should be tackled once we have a better understanding of the typical SQS use case - primarily, the average message size. We may end up exposing the various pipeline tuning parameters (ex: the number of receivers, the maximum number of unacked messages allowed per binding) so the connector can be tuned per use case, but I didn't want to expose those knobs pre-maturely since the additional optional complexity may confuse users.

The initial version of this connector is a bit more complex than I usually lean towards. Most of the complexity that I would usually defer is either necessary in order to potentially hit the 50M messages / hour throughput goal with a single task shard or are small tweaks that head off challenges I anticipate we'd quickly encounter once this connector gets much use (ex: detecting idle queues in order to direct more of the capture's attention towards queues containing messages, supporting FIFO queues, etc.).

With the introduction of this connector, the existing `source-amazon-sqs` connector is deprecated. Its documentation and associated rows in Supabase should be updated to reflect that it's deprecated.

**Checklist:**


nicolaslazo (approved):
Looks great! `ARCHITECTURE.md` helped tons

PR #4840, opened 2026-07-19
ci: fix docs-changelog-check SIGPIPE on large diffs

Alex-Bair:
**Regression?**

No.

**Description:**

The docs/CHANGELOG check piped `git diff` straight into `head -c 180000` under `set -o pipefail`. When a PR's connector diff exceeds the 180 KB cap, `head` closes the pipe once it has read enough, `git diff` dies with SIGPIPE (exit 141), and `pipefail` propagates that as a step failure - so the check crashed on exactly the large PRs it was meant to review (e.g. a new connector), before it could render any verdict.

Write the full diff to a file and truncate by reading that file instead. `head` reading a regular file never signals the producer, so `git diff` completes normally. The 180 KB cap on what the model sees is unchanged.

**Checklist:**

### DIFF
--- a/.github/workflows/docs-changelog-check.yaml
+++ b/.github/workflows/docs-changelog-check.yaml
@@ -90,11 +90,16 @@ jobs:
           } >> "$GITHUB_OUTPUT"
 
           # Bounded diff of connector code for the model, skipping files
-          # that are large or mechanical.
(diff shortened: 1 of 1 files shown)


jwhartley (approved):
LGTM thanks

PR #4842, opened 2026-07-20
estuary-cdk: render %-style log args via getMessage in LogFormatter

jwhartley:
## Problem

`LogFormatter.format` in `estuary-cdk/estuary_cdk/logger.py` builds the emitted `msg` from `record.msg` (the raw format template) instead of `record.getMessage()`. `record.args` are stashed into a separate `fields["args"]` and never applied to the human-readable message.

Result: any log call using lazy `%`-style formatting comes through with the literal placeholder. Seen in production from gunicorn's internal logger inside source-netsuite's `query_server`:

```
Booting worker with pid: %s
```

which surfaced verbatim in a customer's task-failure alert (wrapped as `completeRecovery: store.RestoreCheckpoint: Booting worker with pid: %s`).

## Scope

Affects every Python connector on estuary-cdk. The connectors' own code mostly uses f-strings (so `record.args` is empty and `record.msg` is already complete), which is why it went unnoticed. Any dependency that logs with `%`-args (gunicorn, much of the stdlib ecosystem) renders with raw placeholders.

## Fix

Use `record.getMessage()` for the emitted `msg`. It applies `record.args` when present and returns `record.msg` unchanged when args is empty, so f-string callers are unaffected (no regression). `fields["args"]` is left in place for structured logging.

## Note

estuary-cdk is a shared library with no per-connector `CHANGELOG.md`. If the docs-changelog-check flags this, it's a false positive for a shared-lib change — apply `docs-check-skip`.

### DIFF
--- a/estuary-cdk/estuary_cdk/logger.py
+++ b/estuary-cdk/estuary_cdk/logger.py
@@ -37,7 +37,7 @@ class LogFormatter(logging.Formatter):
         elif record.stack_info:
             fields["stack"] = self.formatStack(record.stack_info).splitlines()
 
-        return OpsLog(level=record.levelname, msg=record.msg, fields=fields).model_dump_json()
+        return OpsLog(level=record.levelname, msg=record.getMessage(), fields=fields).model_dump_json()
 
 
 # Boolean log field set to True on log lines we aggregate across the whole
--- a/estuary-cdk/estuary_cdk/logger.py
+++ b/estuary-cdk/estuary_cdk/logger.py
@@ -37,7 +37,11 @@ class LogFormatter(logging.Formatter):
         elif record.stack_info:
             fields["stack"] = self.formatStack(record.stack_info).splitlines()
 
-        return OpsLog(level=record.levelname, msg=record.getMessage(), fields=fields).model_dump_json()
+        try:
+            msg = record.getMessage()
+        except Exception:
+            msg = str(record.msg)  # fall back to raw template on malformed %-args
+        return OpsLog(level=record.levelname, msg=msg, fields=fields).model_dump_json()
 
 
 # Boolean log field set to True on log lines we aggregate across the whole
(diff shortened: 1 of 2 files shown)


estuary-cdk/estuary_cdk/logger.py
```diff
@@ -37,7 +37,7 @@ def format(self, record: logging.LogRecord) -> str:
         elif record.stack_info:
             fields["stack"] = self.formatStack(record.stack_info).splitlines()
 
-        return OpsLog(level=record.levelname, msg=record.msg, fields=fields).model_dump_json()
+        return OpsLog(level=record.levelname, msg=record.getMessage(), fields=fields).model_dump_json()
```
Alex-Bair: `record.msg` → `record.getMessage()` isn't a clean swap. In various connectors, we've been passing context `dict`s positionally (`log.info("msg", {...})`), so `getMessage()` now runs `msg % args` at ~246 call sites. Most are no-ops, but any `%` in the message throws - worst case an f-string interpolating data that contains one, e.g. salesforce `api.py:243` when an error echoes a SOQL `LIKE '%x%'`. It won't crash the connector, but logging swallows the error and drops the line.

The suggested fix below keeps the intended behavior introduced in this PR and is backwards compatible with the existing logs.
```suggestion
        try:
            msg = record.getMessage()
        except Exception:
            msg = str(record.msg)  # fall back to raw template on malformed %-args
        return OpsLog(level=record.levelname, msg=msg, fields=fields).model_dump_json()
```

jwhartley: Good catch, verified: source-salesforce-native/api.py:243 passes a context dict positionally with an f-string that can echo a SOQL LIKE '%x%', and "...%x%" % ({...},) raises TypeError, which logging then swallows and drops the line. Applied your suggestion and added a regression test (test_malformed_percent_args_fall_back_to_raw_template) covering the stray-% + positional-dict case. Thanks.

Alex-Bair (approved):
LGTM, thanks!

PR #4845, opened 2026-07-20
materialize-postgres: parameterize fence update to avoid pg_stat_statements bloat

### LINKED ISSUE #4841: materialize-postgres: fence update uses DO-block literal, floods pg_stat_statements
## 1. Priority
Low - materialization is healthy and there is no data loss. Raised above raw Low because a top-tier customer is actively tracking it.

## 2. Scope and prevalence
The `updateFence` template lives in `materialize-postgres`, shared by its variants (`materialize-amazon-aurora-postgres`, `-amazon-rds-postgres`, `-alloydb`, `-google-cloud-sql-postgres`, `-supabase-postgres`, `materialize-timescaledb`). Observed: 1 tenant so far - ~7k unique `pg_stat_statements` entries accumulated, approaching `pg_stat_statements.max` = 10k, with `LWLock:pg_stat_statements` waits.

## 3. Problem description
- Expected: repeated fence-update commits should not each register as a distinct `pg_stat_statements` entry; the view stays bounded under `track_utility = on`.
- Observed: on a Postgres with `pg_stat_statements` + `track_utility = on`, every fence-update commit registers as a new unique entry (~7k accumulated, approaching the 10k max, with `LWLock:pg_stat_statements` waits). The fence update is emitted as a `DO $$ ... SET checkpoint = <base64 literal> ... END $$` block (`materialize-postgres/sqlgen.go:386`); `pg_stat_statements` treats a DO block as an opaque utility statement and does not normalize the inlined literal, so each commit is distinct. (Checkpoint size, ~170k at 344 bindings, is expected for long path names and is unrelated.)

## 4. References
- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1784505203971889

mdibaiee:
**Regression?**

No.

**Description:**

Fixes #4841. The Postgres fence update was a `DO` block, which is a utility statement that `pg_stat_statements` can't normalize — so every commit's inlined checkpoint became a distinct entry. This switches it to an ordinary parameterized `UPDATE`, which collapses into a single normalized entry.

**Workflow steps:**

No user-facing workflow change. The materialization behaves identically; only the SQL issued for the checkpoint fence update changes.

None.

**Notes for reviewers:**

- The fenced-off check that `RAISE` performed inside the `DO` block now happens in Go via `RowsAffected`. This mirrors what `materialize-redshift` and `materialize-sql`'s `StdUpdateFence` already do — Postgres was the outlier.
- The checkpoint is still stored as base64 text (same as `installFence` writes), so it's backwards compatible with existing checkpoint rows.
- `TestIntegration` (including the fencing test) passes locally.

**Checklist:**

### DIFF
--- a/materialize-postgres/sqlgen.go
+++ b/materialize-postgres/sqlgen.go
@@ -436,12 +436,10 @@ limit 1
 ;
 {{ end }}
 
--- Templated parameterized update of the fence checkpoint. The bind parameters
--- ($1..$5) are supplied by fenceUpdateArgs. Emitting this as ordinary DML with
--- placeholders (rather than a DO block with inlined literals) lets Postgres'
--- pg_stat_statements collapse every commit into a single normalized entry; a
--- DO block is a utility statement whose body cannot be normalized, so each
--- unique checkpoint would otherwise create a distinct entry. The fenced-off
+-- Update the fence checkpoint. The bind parameters ($1..$5) are supplied by
+-- fenceUpdateArgs. This must stay parameterized rather than inlining the
+-- checkpoint as a literal, so that pg_stat_statements collapses every commit
+-- into a single normalized entry instead of one per checkpoint. The fenced-off
 -- check is performed in Go via the reported RowsAffected.
 
 {{ define "updateFence" }}
--- a/materialize-postgres/sqlgen.go
+++ b/materialize-postgres/sqlgen.go
@@ -436,20 +436,21 @@ limit 1
 ;
 {{ end }}
 
-{{ define "updateFence" }}
-DO $$
-BEGIN
-	UPDATE {{ Identifier $.TablePath }}
-		SET   checkpoint = {{ Literal (Base64Std $.Checkpoint) }}
-		WHERE materialization = {{ Literal $.Materialization.String }}
-		AND   key_begin = {{ $.KeyBegin }}
-		AND   key_end   = {{ $.KeyEnd }}
-		AND   fence     = {{ $.Fence }};
+-- Templated parameterized update of the fence checkpoint. The bind parameters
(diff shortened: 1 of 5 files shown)


materialize-postgres/sqlgen.go
```diff
@@ -436,20 +436,21 @@ limit 1
 ;
 {{ end }}
 
+-- Templated parameterized update of the fence checkpoint. The bind parameters
+-- ($1..$5) are supplied by fenceUpdateArgs. Emitting this as ordinary DML with
+-- placeholders (rather than a DO block with inlined literals) lets Postgres'
+-- pg_stat_statements collapse every commit into a single normalized entry; a
+-- DO block is a utility statement whose body cannot be normalized, so each
+-- unique checkpoint would otherwise create a distinct entry. The fenced-off
+-- check is performed in Go via the reported RowsAffected.
```
danielnelson: This comment talks about a DO block, but now that is gone, later on I think it will be confusing why we are going on about DO blocks.  There are certainly cases where it makes sense in a comment to talk about alternative approaches or previous bugs to avoid regressions, but I feel like here with the way it is worked its confusing unless you read the full history of the file.

This could be fixed by removing the section about DO blocks, or maybe just by clarifying why its being brought up at all.

danielnelson (dismissed):

materialize-postgres/sqlgen.go
mdibaiee: Good point — I removed the digression about DO blocks and kept just the rationale for why the statement stays parameterized (pg_stat_statements normalization) plus the Go-side RowsAffected check. Fixed in 7dfab2da1.

danielnelson (approved):

PR #4847, opened 2026-07-20
source-postgres: Use CREATE_REPLICATION_SLOT

willdonnelly:
**Description:**

Modifies the source-postgres replication slot (re)create logic so it uses the `CREATE_REPLICATION_SLOT` (and `DROP_REPLICATION_SLOT`) command over the logical replication connection rather than running a `SELECT pg_create_logical_replication_slot()` query over the query connection.

The motivation here is that the `SELECT` form is still a `SELECT` query and takes a database transaction snapshot which may cause undesired tuple retention when the create itself blocks for a long time due to a long-running transaction on the DB.

This also conveniently sidesteps the whole question of statement timeouts for this one query which we previously exempted.

**Checklist:**

### DIFF
--- a/source-postgres/CHANGELOG.md
+++ b/source-postgres/CHANGELOG.md
@@ -1,11 +1,18 @@
 # Changelog
 
-## 2026-07-16
+## 2026-07-20
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM

PR #4848, opened 2026-07-20
source-jira-native: skip 422 "This board has no columns with a mapped status" errors when snapshotting `sprints`

Alex-Bair:
**Regression?**

No.

**Description:**

Scrum boards whose column configuration maps no statuses can't compute sprints. For those boards, attempting to fetch their sprints causes Jira to respond with a 422 "This board has no columns with a mapped status" message. Instead of letting these 422s fail the capture, we should skip the board and continue.

**Checklist:**


nicolaslazo (approved):
👌

PR #4851, opened 2026-07-21
source-postgres: restrict table-detail catalog queries to requested schemas

jameswinegar:
## Problem

While setting up replication against a database with ~500k tables, the capture shard OOM-killed (SIGKILL ~10s after connect) on every startup, crash-looping indefinitely.

`DiscoverTableDetails` runs five catalog queries with no schema filtering — `getColumns`, `getPrimaryKeys`, `getSecondaryIndexes`, `getGeneratedColumns`, and `getColumnDescriptions` — then filters to the requested tables in memory. On this database the column query alone returns **3.2M rows** (confirmed via `pg_stat_activity`, stuck in `ClientWrite`), which exhausts connector memory before filtering ever happens. `discover_schemas` doesn't help since it only applies to `getTables`.

## Fix

Every caller of `DiscoverTableDetails` already passes an explicit list of requested tables. This change derives the distinct schema set from that list and pushes it into all five queries as an `AND <schema_col> = ANY($1)` predicate.

- Capture startup / Apply / Validate pass bound tables only → queries scoped to the bound schemas (in our case, 180 tables in one schema instead of 3.2M rows across 14k schemas).
- Full discovery passes all listed tables → all schemas present → results unchanged.
- Empty requested list → empty predicate → queries identical to before.

One incidental change: `queryColumnDescriptions` now passes through `fmt.Sprintf`, so its pre-existing SQL-level `format('"%s"."%s"', ...)` literals are escaped to `%%s`.

No config or schema changes; behavior is identical except for the reduced result sets.

## Impact

Measured on the affected production database (~500k tables, ~742k `pg_class` rows, ~6M `pg_attribute` rows), column discovery query (`queryDiscoverColumns`), warm cache:

| | Unfiltered (before) | Schema-filtered (after, `ANY('{public}')`, 180 tables) |
|---|---|---|
| Rows returned | 3,201,318 | 1,899 |
| Execution time | 2,812 ms | 34 ms |
| Shared buffers touched | 121,010 | 24,213 |
| Connector memory | All rows materialized → shard OOM (SIGKILL) at every capture startup | Negligible |

### DIFF
--- a/source-postgres/.snapshots/TestDiscoveryCrossSchemaIsolation
+++ b/source-postgres/.snapshots/TestDiscoveryCrossSchemaIsolation
@@ -1,0 +1,283 @@
+sql> CREATE SCHEMA IF NOT EXISTS otherschema
+sql> GRANT USAGE ON SCHEMA otherschema TO flow_capture
+sql> CREATE TABLE test.discoverycrossschemaisolation_568126 (k1 INTEGER PRIMARY KEY, foo TEXT)
+sql> CREATE TABLE otherschema.tbl568126 (
+		k2 TEXT PRIMARY KEY,
+		bar INTEGER,
+		gen INTEGER GENERATED ALWAYS AS (bar + 1) STORED
+	)
+sql> GRANT SELECT ON ALL TABLES IN SCHEMA otherschema TO flow_capture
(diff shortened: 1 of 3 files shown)


jameswinegar:
superseded by #4856

PR #4854, opened 2026-07-21
source-sqlserver: Support SMALLDATETIME as backfill keys

### LINKED ISSUE #4852: source-sqlserver: backfill fails on smalldatetime primary keys (conversion error on resume)
## Summary

Backfilling a SQL Server table whose primary key includes a `smalldatetime` column fails with:

```
error performing backfill: error scanning table "<schema>.<table>": mssql: Conversion failed when converting character string to smalldatetime data type.
```

The failure occurs when a keyed backfill resumes after its first chunk. This affects both `source-sqlserver` (CDC) and `source-sqlserver-ct` (change tracking), which share the same backfill key-encoding code (`go/capture/sqlserver`).

## Details

Keyed backfills page through a table by re-issuing a query with a `WHERE <key> > @cursor` predicate, where the cursor is the last row key from the previous chunk, serialized into the checkpoint and later bound back as a query parameter.

The key-encoding path (`EncodeKeyFDB` in `go/capture/sqlserver/datatypes/datatypes.go`) special-cases only the `datetime` type, serializing it to a SQL Server-parseable format. Every other `time.Time`-backed type — including `smalldatetime` — falls through to a default RFC3339-nano format that emits nine fractional-second digits and a `Z` timezone suffix (e.g. `2024-06-01T12:35:00.000000000Z`).

`smalldatetime` has only minute precision and accepts neither fractional seconds nor a timezone designator, so when this string is bound back into the resume predicate on the next chunk, SQL Server's implicit string-to-`smalldatetime` conversion fails and the backfill errors out.

Note the row-value output path handles `smalldatetime` correctly; only the primary-key/resume-cursor encoding is affected. Other datetime-family types that also fall through the default branch (e.g. `datetime2`, `datetimeoffset`) share this encoding pattern and may be affected as well.

## Steps to reproduce

1. In SQL Server, create a table with a primary key that includes a `smalldatetime` column, and insert enough rows to require more than one backfill chunk.
2. Capture the table with `source-sqlserver` or `source-sqlserver-ct`.
3. The first backfill chunk succeeds (no `WHERE` predicate); the second chunk fails with the conversion error above.

## Impact

Tables with a `smalldatetime` primary-key column cannot be backfilled. The capture cannot progress past the first chunk of the affected table.

Alex-Bair:
**Regression?**

No

**Description:**

`SMALLDATETIME` key columns were serialized into backfill resume keys as RFC3339 strings with nanosecond precision and a timezone suffix, which SQL Server cannot convert back to `SMALLDATETIME`. Any keyed backfill of such a table failed with "Conversion failed when converting character string to smalldatetime data type" when resuming after its first chunk.

Apply the timezone-free key encoding already used for `DATETIME` keys, and add a `SMALLDATETIME` case to the shared scan-key tests. This fixes both `source-sqlserver` and `source-sqlserver-ct` via the shared datatypes package. Captures already stuck on a bad persisted cursor will need their backfill restarted after upgrading.

**Checklist:**

### DIFF
--- a/go/capture/sqlserver/datatypes/datatypes.go
+++ b/go/capture/sqlserver/datatypes/datatypes.go
@@ -283,7 +283,7 @@ func EncodeKeyFDB(key, ktype any) (tuple.TupleElement, error) {
 	switch key := key.(type) {
 	case time.Time:
 		switch ktype {
-		case "datetime":
+		case "datetime", "smalldatetime":
 			return key.Format(DatetimeKeyEncoding), nil
 		default:
 			return key.Format(SortableRFC3339Nano), nil
--- a/go/capture/sqlserver/tests/datatypes.go
+++ b/go/capture/sqlserver/tests/datatypes.go
@@ -291,6 +291,11 @@ func testScanKeyTypes(t *testing.T, setup testSetupFunc) {
 			"('1991-08-31T12:34:54.333', 'Data 1')",
 			"('2000-01-01T01:01:01', 'Data 2')",
 		}},
+		{"SmallDateTime", "SMALLDATETIME", []string{
+			"('1991-08-31T12:34:00', 'Data 0')",
+			"('1991-08-31T12:35:00', 'Data 1')",
+			"('2000-01-01T01:01:00', 'Data 2')",
+		}},
 	} {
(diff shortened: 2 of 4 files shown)


willdonnelly (approved):
LGTM

PR #4856, opened 2026-07-21
source-postgres: Improved discovery filters

willdonnelly:
**Description:**

Makes source-postgres discovery operations more efficient when discovering a subset of tables from a massive catalog, and introduces the planned "Discovery Filters" config section as described in https://github.com/estuary/connectors/issues/4476 to make it easier for users to configure captures which only discover some of the available tables.

**Workflow steps:**

No user action is required if they're happy with the existing advanced config option for schema whitelisting. If desired users can instead use the new "Discovery Filters" config section. For now both sets of options are merged in the appropriate ways, eventually we will migrate and deprecate the legacy advanced-option properties.

Docs edits are in the PR, I'll see about syncing those over to the Flow repo.

**Checklist:**


Alex-Bair (approved):
LGTM, thanks!

Issue #4539, opened 2026-05-26
estuary-cdk: publish package for use in external connectors


Alex-Bair:
To make creating connectors easier for third-party developers, we want to make our CDK publicly available on a package index like PyPI. There are a few things we may want to consider doing before publishing on PyPI include:
- Updating the `README.md` with installation instructions, license, core concepts, standard architecture, and some reference connectors.
- Thinking through the public API surface that should be exposed from the CDK & ensuring we're exposing those.
- Adding a `CHANGELOG.md` and a `LICENSE`.
- Figuring out the publishing process to make it as easy as possible to bump the CDK version when functionality is added, fixed, or deprecated.

PR #4861, opened 2026-07-21
base-image: use debian trixie

danielnelson:
**Regression?**

No

**Description:**

Debian Trixie has been stable for around a year, so I think it's about time we move the images onto the new base.  I incremented the base-image version so the old bookworm base is still available.

**Workflow steps:**

New builds will use trixie.

None

**Notes for reviewers:**

Debian release info: https://www.debian.org/releases/

**Checklist:**

- [x] ~~If this change is user-visible, the affected connector's `CHANGELOG.md` and documentation have been updated (see [CONTRIBUTING.md](../CONTRIBUTING.md#changelog-entries)). The `Docs / CHANGELOG check` CI job reviews this automatically; apply the `docs-check-skip` label to dismiss a false positive.~~ Not user-visible.


jacobmarble (dismissed):
Looks pretty simple. How to test?

danielnelson:
I think we can rely on the existing tests, I'm not too concerned since we don't use much base image functionality.  Of course there could be an unknown, one thing I could do is split the PR up to roll out incrementally.

mdibaiee:
> I think we can rely on the existing tests, I'm not too concerned since we don't use much base image functionality.  Of course there could be an unknown, one thing I could do is split the PR up to roll out incrementally.

@danielnelson I think you can ask Claude to set up some tasks with these new base images (built with build-local.sh), preferably the ones with local docker compose setups (source-postgres and materialize-postgres) and verify a simple capture + materialization works on the new base-image. Perhaps also a python and a Rust connector. Claude is pretty good at running these things locally so might be an easy one to delegate. 

If you point it to the flow repo clone you have locally, it can use your mise setup and create a new VM or use an existing one for this

danielnelson:
I did an end-to-end test with Go source-postgres -> materialize-postgres, no issues found.

I also did an end-to-end test with the Kafka rust connectors.  No problems though I did end up removing the copying of some shared objects to the runtime container in the Dockerfile.  These are already contained by the base-image, though some of the soname's have changed.  Perhaps this goes back to a time where the base-image and the builder image didn't line up?

It looks like this PR does not affect any of the python connectors, unless you count materialize-iceberg.  They are using the base image in `python/Dockerfile`, which appears to be based on debian-bullseye. I suggest that separately it gets updated to use `python:3.11-slim-trixie`, so it has a stable Debian base image.  > Debian Bulleye is EOL, past LTS, and into [ELTS](https://wiki.debian.org/LTS/Extended), while Python 3.11 still has 1 year before EOL. cc @Alex-Bair 

The test for `materialize-kafka` failing is unrelated to this PR, and is fixed by #4888.

mdibaiee (approved):

willdonnelly (approved):
LGTM

Alex-Bair:
Thanks for the heads up, @danielnelson!

> They are using the base image in `python/Dockerfile`, which appears to be based on debian-bullseye.

We actually have _two_ Python libraries for developing capture connectors (`python` was the initial version that eventually birthed the `estuary-cdk` that we've been actively maintaining), and each have their own `Dockerfile`. `python/Dockerfile` is used by the former, and I'm working on removing the `python` directory altogether. `estuary-cdk` is using `estuary-cdk/common.Dockerfile`, which uses the `python:3.12-slim` tag, which [resolves to the same image](https://github.com/docker-library/official-images/blob/50262ab3acbdbf7153468cb17c9feb6a14bfad9c/library/python#L155-L158) as `python:3.12-slim-trixie`, so I think we're ok on that front.

re: Python versions - I plan to audit the Python connectors & bump them up to more recent Python versions where possible (along with other misc. housekeeping that'd be nice to address) as part of https://github.com/estuary/connectors/issues/4539.

PR #4862, opened 2026-07-21
docs: use :::deprecated admonition for deprecated connector notices

jwhartley:
## Summary
Swaps the "this connector is deprecated" `:::warning`/`:::danger` blocks on 16 deprecated connector pages to a custom `:::deprecated` admonition (renders the same as `:::warning`, just its own type).

Requires estuary/docs#11, which registers the `deprecated` admonition type. Without that PR merged first (or alongside), these pages would render with a raw/unstyled fallback.

Context: Kapa's search widget was surfacing these deprecated pages as generic filler results for unrelated queries (near-duplicate boilerplate clusters tightly in embedding space). Karl Jones proposed excluding pages via a CSS class once we had one to target — [Slack thread](https://estuaryworkspace.slack.com/archives/C08G1A3LPMM/p1784205687491909). This is that class (`theme-admonition-deprecated`, applied automatically by the paired docs-repo change).

## Test plan
- [x] Verified all 16 files have exactly one `:::warning`/`:::danger` block each (the deprecation notice), confirmed no unrelated warnings were touched
- [x] Verified locally: aggregated into a docs-repo build with the paired admonition-type change, HubSpot page renders correctly with the DEPRECATED admonition
- [ ] Merge alongside/after estuary/docs#11

### DIFF
--- a/docs/reference/Connectors/capture-connectors/HubSpot.md
+++ b/docs/reference/Connectors/capture-connectors/HubSpot.md
@@ -3,7 +3,7 @@
 
 This connector captures data from a Hubspot account.
 
-:::warning
+:::deprecated
 This connector is deprecated. It is recommended that you use the new native [HubSpot connector](/reference/Connectors/capture-connectors/HubSpot-real-time) instead.
 :::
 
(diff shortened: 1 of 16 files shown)


SeanWhelan (approved):
LGTM

PR #4863, opened 2026-07-22
materialize-iceberg: add table_identifier_case and field_name_case advanced options

jacobmarble:
**Description:**

Adds a new advanced config option `all_caps_table_names` to `materialize-iceberg`. When enabled, created table names are upper-cased. Mirrors the existing `lowercase_column_names` option but targets the table identifier. Off by default; no change to existing materializations.

**Workflow steps:**

Set `advanced.all_caps_table_names: true`. The table element of the resource path is upper-cased at the single chokepoint (`resource.Parameters()`), so creation, lookup, and apply all agree. Namespace is unchanged.

None.

**Notes for reviewers:**

- TDD: commit 1 failing test, commit 2 implementation.
- Caveat: InfoSchema existence matching is case-sensitive against catalog-returned names, so this relies on the catalog preserving case (Glue, Snowflake Polaris do) — same assumption the existing lowercase path makes.
- `materialize-iceberg` has no `CHANGELOG.md` (opt-in), so no entry; `docs-check-skip` may be needed if the CI check flags it.

**Checklist:**

- [x] If this change is user-visible, the affected connector's `CHANGELOG.md` and documentation have been updated.


mdibaiee (dismissed):

jacobmarble:
**Direction change:** this PR's feature is now an enum instead of a boolean.

The advanced option `all_caps_identifiers` (née `all_caps_table_names`) is replaced by:

```yaml
advanced:
  identifier_case: lowercase | uppercase | preserve   # default: lowercase
```

- `lowercase` (default) — fold namespace and table names to lower case; identical to the connector's historical behavior, and what an absent/empty value means.
- `uppercase` — fold namespace and table names to upper case. This is the Snowflake-managed-Iceberg (Polaris/Horizon) compatibility mode: Snowflake's catalog is case-sensitive and unquoted Snowflake identifiers resolve upper-case, so upper-cased Iceberg identifiers are queryable unquoted.
- `preserve` — keep names as written in the spec (character sanitization still applies).

Rationale: one extensible knob instead of accreting `all_caps_*`/`preserve_case` booleans with conflicting combinations. A single enum also matches precedent elsewhere in the repo where case behavior is configurable (e.g. materialize-redshift's `enable_case_sensitive_identifier` handling).

The interim boolean shipped only in dev tags (`ec2deaf`, `5c329b2`) applied to one materialization; that spec will need `identifier_case: uppercase` when it moves off its pinned tag.

mdibaiee:
@jacobmarble does this apply to column names as well? "identifier" usually includes column name identifiers as well, and if they want to query without quoting, I think it applies to column names as well?

If so, then we would have two feature flags that cover column name casing. Maybe we can deprecate the old one and only have the new one going forward?

jacobmarble:
> @jacobmarble does this apply to column names as well? "identifier" usually includes column name identifiers as well, and if they want to query without quoting, I think it applies to column names as well?
> 
> If so, then we would have two feature flags that cover column name casing. Maybe we can deprecate the old one and only have the new one going forward?

This change only modifies the case of namespaces and tables. I agree that "identifier" could also be interpreted as column name, index name, view name, others. The alternative `path_case` relies on Estuary jargon, and would be even more confusing to users. The Iceberg spec uses the term "table identifier" aka `TableIdentifier` precisely to mean `namespace + table name`.

Merging the two configuration options creates a conflict in the default state of each -- before this change, column names default to `preserve` behavior, while namespace and table names default to `lowercase` behavior.

FYI the Iceberg spec favors "fields" not "columns".

### Plan

- Rename `identifier_case` to `table_identifier_case`, keeping the default as `lowercase`, and be clear about what that means in the English description.
- Add `field_name_case` using the same enum values, defaulting to `preserve`.
- Deprecate `lowercase_column_names` and fail validation if both `field_name_case` and `lowercase_column_names` are set and conflict.

materialize-iceberg/.snapshots/TestSpec
```diff
@@ -392,7 +392,29 @@
           "lowercase_column_names": {
             "type": "boolean",
             "title": "Lowercase Column Names",
-            "description": "Create all columns with lowercase names. This is necessary for compatibility with some systems such as querying S3 Table Buckets with Athena."
+            "description": "Deprecated: set field_name_case to 'lowercase' instead. Create all columns with lowercase names."
```
mdibaiee: When we deprecate fields such as this, we stop exposing them in JSONSchema so that new users don't see it in their config anymore.

mdibaiee:
@jacobmarble we haven't done this before in materialization team, but there is a new cool feature `configUpdate` which allows us to automatically migrate the config of tasks from the connector itself.

I wonder if we can use that to migrate people using `lowercase_column_names: true` to `field_name_case: lowercase` by automatically setting the latter on their tasks if the former exists, and once we confirm they both exist and are equal in value (in another run of the connector), we then unset `lowercase_column_names`, lending a clean state for the customer

jacobmarble:
> @jacobmarble we haven't done this before in materialization team, but there is a new cool feature `configUpdate` which allows us to automatically migrate the config of tasks from the connector itself.
> 
> I wonder if we can use that to migrate people using `lowercase_column_names: true` to `field_name_case: lowercase` by automatically setting the latter on their tasks if the former exists, and once we confirm they both exist and are equal in value (in another run of the connector), we then unset `lowercase_column_names`, lending a clean state for the customer

I would like to try this, but getting the change merged for a customer is higher priority at the moment IMO. How about in a follow-up PR?

mdibaiee (approved):

PR #4871, opened 2026-07-22
source-sqlserver-ct: Detect changed tables with a single query

willdonnelly:
**Description:**

Previously each poll issued a CHANGETABLE read against every active table. On a capture with hundreds of low-traffic bindings that meant hundreds of empty polling queries issued serially.

After this change we'll run one batched detection query (a UNION ALL of an EXISTS-over-CHANGETABLE per binding) to find exactly which tables have new changes and then read only those ones.

A local benchmark (200 tables, 199 idle) showed this to be ~3x faster. Real-world deployments will typically have higher per-query costs (since network latency is nonzero, the database is under load, etc), we should thus expect more dramatic improvements in production.

**Workflow steps:**

No user action required.

No documentation updates required.

**Checklist:**

### DIFF
--- a/source-sqlserver-ct/CHANGELOG.md
+++ b/source-sqlserver-ct/CHANGELOG.md
@@ -1,5 +1,8 @@
 # source-sqlserver-ct
 
-## v1, 2026-01-20
+## 2026-07-22
 
-- Connector created.
+### Changed
+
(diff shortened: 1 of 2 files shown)


Alex-Bair (approved):
LGTM

PR #4874, opened 2026-07-22
source-sage-intacct: handle CURRENCY and PERCENT datatypes

### LINKED ISSUE #4873: source-sage-intacct: CURRENCY (and PERCENT/SEQUENCE) datatypes raise "Unhandled datatype" on populated values
The Sage Intacct connector fails when it encounters a populated `CURRENCY`-typed field. The capture dies during `runTransactions` with:

```
Value error, Unhandled datatype CURRENCY for BASIC: 6.50 [type=value_error]
```

## Root cause

In `source-sage-intacct/source_sage_intacct/sage.py`, `SageRecord._normalize_values` explicitly raises for `CURRENCY | SEQUENCE | PERCENT`:

```python
case "CURRENCY" | "SEQUENCE" | "PERCENT":
    # TODO(whb): Update this once we see some values and know
    # what they look like.
    raise ValueError(
        f"Unhandled datatype {datatype} for {field_name}: {val}"
    )
```

These datatypes were left as `# TODO(whb)` placeholders pending real sample values. The `ITEM` object has a `CURRENCY` field (`BASIC`) that is now populated in production (e.g. `6.50`), triggering the raise.

## Also affected

- `sourced_schema()` currently drops `CURRENCY/SEQUENCE/PERCENT` fields (`continue`) instead of emitting a schema type.
- `DATATYPE_MAP` already maps `CURRENCY -> str`.

## Suggested fix

Handle `CURRENCY` (and likely `PERCENT`) the same way as `DECIMAL`:
- In `_normalize_values`: `val = str(val)` rather than raising.
- In `sourced_schema()`: emit `{"type": "string", "format": "number"}` instead of `continue`.

`SEQUENCE` shape should be confirmed against a real value before removing its raise.

## Impact

Any Sage Intacct capture that includes an object with a populated CURRENCY field fails on `runTransactions`. There is currently no config-level workaround.

JSap0914:
**Regression?**

Not a regression from a specific PR — `CURRENCY`/`SEQUENCE`/`PERCENT` were left as `# TODO(whb)` placeholders that raise. This surfaced now that a production object (`ITEM.BASIC`) has a populated `CURRENCY` value.

**Description:**

Fixes #4873. The Sage Intacct connector fails during `runTransactions` when it encounters a populated `CURRENCY`-typed field:

```
Value error, Unhandled datatype CURRENCY for BASIC: 6.50
```

`SageRecord._normalize_values` explicitly raised for `CURRENCY | SEQUENCE | PERCENT`, and `sourced_schema()` dropped those fields.

`CURRENCY` and `PERCENT` are numeric values serialized as strings, so this handles them exactly like `DECIMAL`:

- `_normalize_values`: `val = str(val)` instead of raising.
- `sourced_schema()`: emit `{"type": "string", "format": "number"}` instead of `continue`.

`SEQUENCE` is intentionally left raising, per the issue's guidance to confirm its real value shape before emitting an unverified type. It's split out into its own `case` so it fails loudly rather than silently.

**Workflow steps:**

No config change. Captures that include an object with a populated `CURRENCY` or `PERCENT` field now succeed, and those fields appear in the captured document schema as numeric strings.

None. Behavior is described in the connector's `CHANGELOG.md` (added here).

**Notes for reviewers:**

- `DATATYPE_MAP` already mapped all three types to `str`; only the normalization/schema branches raised, so the field type is unchanged.
- Added `source-sage-intacct/tests/test_datatypes.py` covering: `CURRENCY`/`PERCENT` normalize to strings, `sourced_schema()` emits `{"type":"string","format":"number"}` for them, and `SEQUENCE` still raises (guarding the deferred scope).
- `pytest source-sage-intacct` → 28 passed.

**Checklist:**

- [x] If this change is user-visible, the affected connector's `CHANGELOG.md` and documentation have been updated. Added `source-sage-intacct/CHANGELOG.md`.

### DIFF
--- a/source-sage-intacct/CHANGELOG.md
+++ b/source-sage-intacct/CHANGELOG.md
@@ -1,0 +1,9 @@
+# Changelog
+
+## 2026-07-22
+
+### Fixed
+- Captures no longer fail on populated `CURRENCY` and `PERCENT` fields. These
+  values are now emitted as numeric strings (like `DECIMAL`) instead of raising
+  `Unhandled datatype ... for ...`, and their fields are included in the
+  captured document schema.
--- a/source-sage-intacct/source_sage_intacct/sage.py
+++ b/source-sage-intacct/source_sage_intacct/sage.py
@@ -34,9 +34,9 @@ DATATYPE_MAP: dict[str, Any] = {
     "TIMESTAMP": AwareDatetime,
     "ENUM": str,
     "DECIMAL": str,
-    # TODO(whb): Update the remaining types once we see some values and know
-    # what they look like. `str` is just a placeholder, and may or may not be
-    # appropriate.
+    # CURRENCY and PERCENT are numeric values serialized as strings, handled
+    # like DECIMAL. SEQUENCE's shape is still unconfirmed; `str` is a
+    # placeholder and it is not yet normalized or emitted in schemas.
     "CURRENCY": str,
     "PERCENT": str,
     "SEQUENCE": str,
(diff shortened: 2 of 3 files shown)


Alex-Bair:
Superseded by https://github.com/estuary/connectors/pull/4877.

PR #4875, opened 2026-07-22
source-salesforce-native: paginate backfills by `Id` instead of date windows

Alex-Bair:
**Regression?**

No.

**Description:**

Backfills in `source-salesforce-native` have been querying records in chunks of date windows. For organizations with large amounts of data & uneven distributions of data among date windows, it has been tedious finding the best `advanced.window_size` setting for all bindings' backfills to complete in a reasonable amount of time.

Instead of continuing to use date windows, this PR updates the connector to use objects' `Id` field as a page cursor in queries like `SELECT Id, Field1, Field2 FROM Object WHERE Id > :lastId ORDER BY Id LIMIT N`. This eliminates the need to chunk backfill ranges up by date windows completely and instead paginate through the records we care about by the always present `Id` field. This should both make it easier for backfills to complete without requiring tinkering with capture tasks' specs _and_ potentially help bulk jobs complete faster since `Id` _should_ be indexed on Salesforce's side.

Per-API strategy is now:
  - Bulk API backfills run one query job per 500,000-record page, checkpointing an `Id` cursor every 25,000 records within a job.
  - REST API backfills (objects with compound fields, or when falling back from the Bulk API) fetch 2,000-record pages. Each page's boundary `Id` is located upfront with a cheap probe query so the field-chunked queries share an identical closed `Id` range and merge deterministically.

**Workflow steps:**

All new backfills (including scheduled formula field refreshes) will use `Id` as a page cursor and use the new `Id` centric queries. Ongoing backfills will continue using date window queries until they complete.

**Notes for reviewers:**

Tested with `flowctl preview` and dropping the backfill limit low enough to require multiple bulk jobs to fetch all records. Confirmed all records are captured by the backfill as anticipated.

Note: We may need to add another advanced config option to allow users to change the `BULK_ID_PAGE_LIMIT` if the 500,000 default does not work well for their use case. But I'd like to see how well the connector performs with the 500,000 limit before introducing another setting in the endpoint config.

**Checklist:**

### DIFF
--- a/source-salesforce-native/CHANGELOG.md
+++ b/source-salesforce-native/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+
+## 2026-07-22
+
+### Changed
+- Backfills of incremental streams now paginate by the object's `Id` field instead of
+  date windows over the cursor field. Backfill performance no longer depends on how
+  records are distributed over time, and the `advanced.window_size` setting no longer
+  needs tuning for large objects (it still applies to backfills already in progress
+  and to incremental catch-up sweeps).
+- Backfills that were already in progress before this change continue from their
+  saved position using the previous date window strategy; no re-backfill is needed.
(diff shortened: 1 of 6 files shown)


nicolaslazo (approved):
👌

PR #4876, opened 2026-07-22
materialize-clickhouse: log table name in error messages

### LINKED ISSUE #4870: materialize-clickhouse: include target table name in `code: 252` (too-many-partitions) store errors
## TL;DR

When a ClickHouse materialization hits `code: 252, message: Too many partitions for single INSERT block (more than 100)`, the error that Flow surfaces (in `catalog status`, the `shard_failed` alert, and the task logs) is **completely anonymous** — it does not say **which binding / target table** caused it. Since the whole point of the error is that a *specific* table's `PARTITION BY` is too fine-grained for the batch being written, not naming that table forces the operator into a long, credentialed investigation against the customer's ClickHouse instance just to answer "which table?". We should attach the target table name (and ideally the offending partition count) to the error before it propagates.

This is purely a **diagnosability / logging** improvement — no data-plane behavior change.

## Background / concrete incident that motivated this

While investigating `AltopEnergy/MarketData/materialize-clickhouse` (a real prod materialization with **214 bindings**), the task was crash-looping. The only diagnostic signal anywhere in the Flow control plane was this, repeated identically across every failure:

```
runTransactions: txnStartCommit: app.FinalizeTxn: transactor.Store:
flushing store batch: code: 252, message: Too many partitions for single
INSERT block (more than 100). The limit is controlled by
'max_partitions_per_insert_block' setting. ...
```

Note what is **absent**: the binding name, the ClickHouse table name, the partition count that tripped the limit, and the partition-key expression. With 214 bindings, that error is nearly useless on its own.

### What it took to answer "which table?" without this feature

Every step below was necessary *only* because the table name wasn't in the error:

1. Pulled the materialization spec and confirmed none of the 214 bindings set partitioning in Flow — so the `PARTITION BY` lives in customer-created ClickHouse DDL (endpoint had `allow_existing_tables_for_new_bindings`).
2. Grepped 6h of task logs (1871 lines, 32 `code: 252` occurrences). Confirmed the error is raised at the **aggregate store-batch flush** and the connector dies (`i/o error writing to connector stdin` → `connector failed`) without ever logging the in-flight table.
3. Decrypted the connector's sops/GCP-KMS-encrypted password to connect to the customer's ClickHouse directly.
4. Discovered `system.parts` is a **dead end** for this: the failing partitions are rejected *before* commit, so `... GROUP BY table HAVING count()>100` returns **zero rows** even while the task fails every insert. The identifying state never persists server-side — it exists *only* at the moment the connector issues the INSERT.
5. Had to reconstruct the culprit indirectly: enumerate every `toYYYYMM(...)` table, reason that only a monthly key spanning >100 distinct months (>8.3 years) in a single backfill block can exceed the limit, and narrow to the deep-history monthly tables (e.g. `marketdata_monthlydartmccpjm`, history back to 201301 ≈ 156 months).

That is a lot of privileged, manual work to recover one string the connector already had in hand at the moment of failure.

## Current behavior

- ClickHouse returns exception `code: 252` on the `INSERT` into the per-transaction staging table (`flow_temp_store_0_<resource>`) or on the final merge INSERT into the target.
- The connector wraps and returns this as a generic `transactor.Store: flushing store batch: ...` error.
- The Flow runtime reports it as a shard failure with the raw ClickHouse message and **no binding/table attribution**.
- `system.parts` cannot be used post-hoc to find the table, because the rejected partitions never commit. The information is irretrievable after the fact.

## Desired behavior

When the connector catches a ClickHouse error during store (especially, but not limited to, `code: 252`), enrich the error with context it uniquely holds at that point:

- **Binding / collection name** and the **resolved ClickHouse table name** (e.g. `MarketData.marketdata_monthlydartmccpjm`).
- The **staging vs. final** phase (which INSERT failed).
- If cheaply available, the **partition-key expression** and the **number of distinct partitions in the failing block** (the single most useful number — turns "something is over 100" into "this table put 156 months into one block").

Example of a good enriched message:

```
transactor.Store: flushing store batch for binding
"AltopEnergy/MarketData/marketdata/monthlydartmccpjm" -> table
`MarketData.marketdata_monthlydartmccpjm` (PARTITION BY toYYYYMM(DeliveryMonth)):
INSERT into staging table spanned 156 distinct partitions in a single block;
ClickHouse code: 252, Too many partitions for single INSERT block (more than 100). ...
```

Even the minimal version — appending `binding=<name> table=<db>.<table>` — would eliminate ~90% of the investigation above.

## Implementation notes / pointers

> Hypotheses to verify against the actual connector source, not confirmed line numbers. Connector is Go, built on Estuary's `materialize-sql` framework.

1. **Find where the store batch is flushed.** Grep the `materialize-clickhouse` module for the literal `"flushing store batch"` (the anchor in the observed error). That wrapper is almost certainly iterating bindings or flushing a combined batch.
2. **The connector already knows the table per binding.** `materialize-sql` carries per-binding metadata (resource path / table identifier). Make the flush loop attribute the error to the specific binding it was inserting when ClickHouse returned the error — wrap the per-table `INSERT`/`exec` error with binding + table *at the call site*, rather than letting a bare driver error bubble up from an aggregate flush.
3. **Watch for the "combined batch" case.** If the connector pipelines multiple tables' inserts into one flush and only sees a single error, it needs to either (a) track which table's statement was in flight, or (b) at minimum enrich with the set of tables in the failing batch. Per-statement attribution (a) strongly preferred.
4. **Optional: surface the partition count.** ClickHouse's 252 message doesn't include the actual count. The connector could compute `count(distinct <partition_expr>)` over the block only when handling a 252 (avoid overhead on the happy path). Nice-to-have; the table name is the must-have.
5. **Don't regress the happy path.** Gate any added computation to the error path only.
6. **Consider doing this generically.** `code: 252` is the motivating case, but the same anonymity affects *any* ClickHouse store error (schema mismatches, type-conversion failures, quota errors). A general "wrap store errors with binding + table" improvement helps across the board.

## Acceptance criteria

- [ ] On a `code: 252` store failure, the error text propagated to Flow (`catalog status` `lastFailure`, the `shard_failed` alert, and task logs) names the **binding and target ClickHouse table** responsible.
- [ ] Attribution is correct on a multi-binding task (the Altop case has 214).
- [ ] (Stretch) The error includes the number of distinct partitions in the failing block and the table's partition-key expression.
- [ ] Happy-path store performance unaffected (no per-transaction distinct counting except on the error path).
- [ ] A regression test covers a store error being enriched with the offending table name (unit test around the flush/error-wrapping logic with a faked driver error; no live ClickHouse needed).

## Related context

The immediate operator remedy for `code: 252` on ClickHouse **Cloud** is *not* a settings change: `max_partitions_per_insert_block` is a session/user-level setting pinned by a `CONST` constraint on Cloud's managed `default` profile, and that constraint cannot be relaxed via `ALTER USER` or `ALTER SETTINGS PROFILE default` (both return `452 SETTING_CONSTRAINT_VIOLATION`), nor by a child profile (constraints only tighten downward). The real fix is coarsening the offending table's `PARTITION BY` (e.g. `toYYYYMM` → `toYear`). This request does **not** change that remedy — it just makes it possible to find the table that needs the DDL change in seconds instead of via a credentialed forensic session.

---
*Filed from an operator investigation on 2026-07-22.*

jacobmarble:
**Description:**

Closes #4870. ClickHouse load/store errors (notably `code: 252, Too many partitions for single INSERT block`) surfaced with no indication of which binding/table caused them, forcing a manual forensic investigation on multi-binding tasks. All per-binding error messages in the `Load` and `Store` paths now include the target table identifier, e.g.:

```
transactor.Store: flushing store batch for `db`.`my_table`: code: 252, ...
```

**Workflow steps:**

No behavior change; store/load failures in task logs, `catalog status`, and `shard_failed` alerts now name the affected table.

None — error-message wording only. `CHANGELOG.md` updated.

**Notes for reviewers:**

Message-string-only change (11 error sites); the binding was already in scope at each site. Existing integration suite passes.

**Checklist:**

### DIFF
--- a/materialize-clickhouse/CHANGELOG.md
+++ b/materialize-clickhouse/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2026-07-22
+
+### Changed
+- Errors that occur while loading or storing data now name the affected table, for example ``flushing store batch for `db`.`my_table`: code: 252, ...``. Previously these errors did not identify which table caused them, which made diagnosis difficult on materializations with many bindings.
+
 ## 2026-07-17
 
 ### Added
--- a/materialize-clickhouse/driver.go
+++ b/materialize-clickhouse/driver.go
@@ -517,18 +517,18 @@ func (t *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage) 
 			chBatch, err = t.load.conn.PrepareBatch(ctx, t.bindings[lastBinding].load.insertSQL)
 			return err
 		}); err != nil {
-			return fmt.Errorf("preparing load batch: %w", err)
+			return fmt.Errorf("preparing load batch for %s: %w", t.bindings[lastBinding].target.Identifier, err)
 		}
 		defer chBatch.Close()
 		for _, record := range batch {
 			if err = chBatch.Append(record...); err != nil {
-				return fmt.Errorf("appending load batch: %w", err)
+				return fmt.Errorf("appending load batch for %s: %w", t.bindings[lastBinding].target.Identifier, err)
 			}
 		}
 		batch = batch[:0]
 		batchBytes = 0
 		if err = chBatch.Send(); err != nil {
-			return fmt.Errorf("flushing load batch: %w", err)
+			return fmt.Errorf("flushing load batch for %s: %w", t.bindings[lastBinding].target.Identifier, err)
 		}
 		return nil
 	}
@@ -556,7 +556,7 @@ func (t *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage) 
 				// The load table exists (ensured at session start); truncate
 				// clears any keys left over from a previous round.
 				if err = t.load.conn.Exec(ctx, b.load.truncateSQL); err != nil {
-					return fmt.Errorf("truncating load stage table: %w", err)
+					return fmt.Errorf("truncating load stage table for %s: %w", b.target.Identifier, err)
 				}
 				activeBindings[it.Binding] = 0
 			}
@@ -565,7 +565,7 @@ func (t *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage) 
 		b := t.bindings[it.Binding]
 		converted, err := b.target.ConvertKey(it.Key)
 		if err != nil {
-			return fmt.Errorf("converting load key: %w", err)
+			return fmt.Errorf("converting load key for %s: %w", b.target.Identifier, err)
 		}
 		batch = append(batch, converted)
 		batchBytes += len(it.PackedKey)
@@ -596,7 +596,7 @@ func (t *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage) 
 		b := t.bindings[i]
 		var counted uint64
 		if err = t.load.conn.QueryRow(ctx, b.load.countKeysSQL).Scan(&counted); err != nil {
-			return fmt.Errorf("counting load table keys: %w", err)
+			return fmt.Errorf("counting load table keys for %s: %w", b.target.Identifier, err)
 		}
 		if int64(counted) != inserted {
 			return fmt.Errorf(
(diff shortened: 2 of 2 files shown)


danielnelson (approved):
Maybe use `%q`?

jacobmarble:
> Maybe use `%q`?

I've considered it, but `b.target.Identifier` is already dialect-quoted (e.g. \`MyData\`.\`md_foo\`), so %q would wrap that in escaped double quotes on top of the backticks — " + backticks + " — which is noisy.

PR #4877, opened 2026-07-22
source-sage-intacct: handle CURRENCY and PERCENT datatypes

### LINKED ISSUE #4873: source-sage-intacct: CURRENCY (and PERCENT/SEQUENCE) datatypes raise "Unhandled datatype" on populated values

Alex-Bair:
**Regression?**

No.

**Description:**

Populated `CURRENCY`-typed fields (e.g. `ITEM.BASIC = "6.50"`) crashed captures with "Unhandled datatype CURRENCY" since `_normalize_values` raised for `CURRENCY`/`SEQUENCE`/`PERCENT` as TODO placeholders. Sage's JSON returnformat delivers these values as numeric strings, so treat `CURRENCY` and `PERCENT` exactly like `DECIMAL`: pass values through as strings and emit `{"type": "string", "format": "number"}` in the sourced schema. `SEQUENCE` still raises until we've seen a real value.

Fixes #4873

**Checklist:**

### DIFF
--- a/source-sage-intacct/CHANGELOG.md
+++ b/source-sage-intacct/CHANGELOG.md
@@ -1,0 +1,8 @@
+# Changelog
+
+## 2026-07-22
+
+### Fixed
+- Populated `CURRENCY` and `PERCENT` fields no longer fail the capture with an
+  "Unhandled datatype" error. Their values are now captured as number-formatted
+  strings, the same as `DECIMAL` fields.
--- a/source-sage-intacct/source_sage_intacct/sage.py
+++ b/source-sage-intacct/source_sage_intacct/sage.py
@@ -34,11 +34,11 @@ DATATYPE_MAP: dict[str, Any] = {
     "TIMESTAMP": AwareDatetime,
     "ENUM": str,
     "DECIMAL": str,
-    # TODO(whb): Update the remaining types once we see some values and know
-    # what they look like. `str` is just a placeholder, and may or may not be
-    # appropriate.
     "CURRENCY": str,
     "PERCENT": str,
+    # TODO(bair): Update SEQUENCE once we see some values and know what they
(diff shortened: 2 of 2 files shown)


dyaffe (approved):
LGTM

PR #4878, opened 2026-07-22
source-zuora: use AQuA API & paginate backfills by `Id`

Alex-Bair:
**Regression?**

No.

**Description:**

Shortly after releasing `source-zuora`, we observed that the [`/v1/object/export`](https://developer.zuora.com/v1-api-reference/older-api/exports/object_postexport) endpoint didn't support fields that the `/v1/describe` endpoint says are exportable. There are likely many other incompatibilities & quirks with the legacy `/v1/object/export` endpoint & Export API, and I'd like to completely move the connector off of it & onto an actually support API.

This PR's scope includes:
- Using the [AQuA API](https://docs.zuora.com/en/zuora-platform/data/aggregate-query-api-aqua/aqua-api-introduction) for exporting data instead of the legacy `/v1/object/export` endpoint. The query structure and general export manager control flow remained the same; I mostly had to update models to reflect AQuA API response shapes & the spots the connector referenced fields on those model instances.
- Paginating backfills by `Id` rather than using date window chunks to improve backfill speed. This was inspired by the same improvement for `source-salesforce-native` in https://github.com/estuary/connectors/pull/4875. 
  - Note: Unlike with `source-salesforce-native`, we don't need to support the date window chunk backfill path after this PR is merged since there are no capture tasks that have successfully made any backfill progress. 

**Notes for reviewers:**

This is untested against the actual Zuora API, but the AQuA API docs are more thorough than the previous `/v1/object/export` docs, so I'm cautiously optimistic that this docs-only based development will work at least reasonably well & actually be able to capture some data.

**Checklist:**

### DIFF
--- a/docs/reference/Connectors/capture-connectors/zuora.md
+++ b/docs/reference/Connectors/capture-connectors/zuora.md
@@ -1,7 +1,7 @@
 # Zuora
 
 This connector captures data from [Zuora](https://www.zuora.com/) objects into Estuary Flow collections.
-It authenticates with Zuora's [OAuth 2.0](https://developer.zuora.com/api-references/api/overview/#section/Authentication) client credentials flow and reads data through Zuora's [Data Query / Export API](https://developer.zuora.com/api-references/api/operation/POST_DataQuery/).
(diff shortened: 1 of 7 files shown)


nicolaslazo (approved):
Looks good!

PR #4879, opened 2026-07-23
source-google-ads: upgrade Google Ads API v21 -> v25

### LINKED ISSUE #4597: source-google-ads: upgrade off of API version v21
`source-google-ads` is currently using v21 of the Google Ads API, which will be sunset in August 2026. We'll need to bump the API version used by the connector to a more recent one (ideally the most recent if possible) and communicate any breaking changes to customers.

For some reason, I _thought_ there were breaking changes in versions after v21, so it'd be good to investigate this a little while before v21 is sunset so customers have plenty of warning about any changes.

Alex-Bair:
**Regression?**

No.

**Description:**

Closes #4597 

v21 sunsets ~Aug 5, 2026. This PR moves `source-google-ads` to v25, the latest API version, to maximize how long we can go without needing to bump the API version again. Bumps the `google-ads` library pin from ^28.0.0 to ^31.2.0 (the minimum version that supports v25) and updates all v21 imports, doc URLs, and the API_VERSION constant.

Every breaking change between v21 and v25 was verified against Google's proto diff tool. Seven changes touch fields the connector queries, and the schemas are updated accordingly:

| Ver | Change | Old -> New | Streams |
|-----|--------|------------|---------|
| v22 | rename | `metrics.average_cpv` -> `metrics.trueview_average_cpv` | `account_performance_report`, `ad_group_ad_report`, `display_keyword_performance_report`, `display_topics_performance_report`, `user_location_report` |
| v22 | rename | `metrics.video_view_rate` -> `metrics.video_trueview_view_rate` | (same 5) |
| v22 | rename | `metrics.video_views` -> `metrics.video_trueview_views` | (same 5) + `campaigns` |
| v23 | rename + type | `campaign.start_date` -> `campaign.start_date_time` | `campaigns` |
| v23 | rename + type | `campaign.end_date` -> `campaign.end_date_time` | `campaigns` |
| v23 | removed | `ad_group_ad.ad.call_ad.*` (14 fields; `CallAdInfo` deleted) | `ad_group_ads` |
| v24 | removed | `campaign.video_brand_safety_suitability` | `campaigns` |

Notes:
- The v22 metric renames are semantic narrowings, not pure renames: the `trueview_*` variants cover only TrueView formats, so historical values may shift.
- `campaign.start_date_time` / `end_date_time` values now include a time component.
- The removed call ad fields have no replacement (Call ads are deprecated).
- To recover the brand-safety signal removed from campaign, added `customer.video_brand_safety_suitability` to the `accounts` and `service_accounts` streams.

Everything else the connector queries is proto-confirmed unchanged. v25 itself introduces no changes affecting the connector.

Discover and spec snapshot changes are expected due to the removal & addition of various fields.

**Notes for reviewers:**

We should wait to merge this until it's a bit closer to the sunset date of 05AUG2026. Coordination with solutions on the rollout date happened in this [Slack thread](https://estuaryworkspace.slack.com/archives/C07SEA0G40N/p1784771476667569), and we'll target merging on 03AUG2026.

**Checklist:**


Alex-Bair:
> ### Docs / CHANGELOG check
> **Documentation update likely needed:**
> 
> * `source-google-ads` — API upgrade to v25 renames/removes several schema fields (metrics, campaign dates, call ads) which changes captured document shapes and warrants documentation updates.
> 
> Update the relevant docs in this PR, or link the docs PR if the docs for this connector still live in `estuary/flow`.
> 
> If this is a false positive or the updates are tracked elsewhere, apply the `docs-check-skip` label to dismiss the check (it will re-run and pass).

Schemas for streams are not specified in the connector docs. Schema changes impacting streams are documented in the new `CHANGELOG.md`.

source-google-ads/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
+
+## 2026-07-22
+
+### Changed
+- Upgraded to Google Ads API v25 from v21.
+- Renamed video metrics on `campaigns` and the performance-report streams: `metrics.average_cpv` → `metrics.trueview_average_cpv`, `metrics.video_view_rate` → `metrics.video_trueview_view_rate`, `metrics.video_views` → `metrics.video_trueview_views`. These now cover TrueView formats only, so values may differ for non-TrueView video.
+- Renamed `campaign.start_date` → `campaign.start_date_time` and `campaign.end_date` → `campaign.end_date_time` on the `campaigns` stream; values now include a time component.
```
nicolaslazo, line 8: I just went through my locally regenerated `acmeCo` schemas and noticed `campaign.start_date_time` and `campaign.end_date_time` are still marked as `"format": "date"`, which should cause validation errors now that these timestamps include time components. 

If I'm not mistaken, the fix would be to add these to `NON_DATE_FORMATTED_FIELDS`.

nicolaslazo (changes requested):
Thanks for tackling this one! API version bumps are delicate work. I'm mostly concerned about a schema annotation that could break captures at runtime.

This is something to maybe fix when we don't have a deadline looming over us, but IMO we should redesign tests so these field renames/deletions make snapshot tests go red. CI checks just passed even though a number of `acmeCo` files keep the old field names.

And a small nit on the documentation side: while the text itself doesn't need updating, URLs point to deprecated API versions. Only flagging it because it's something easy for Claude to figure out in the background.

source-google-ads/CHANGELOG.md
Alex-Bair: Good catch, thanks for catching that! I frequently forget to update those schema files. I've updated them and added `campaign.start_date_time` and `campaign.end_date_time` to the `NON_DATE_FORMATTED_FIELDS` list. They are no longer marked with `format: date`.

nicolaslazo (approved):
Looks great, thanks!

Alex-Bair:
Thanks for the review, I appreciate it! And to respond to your earlier comments:

> This is something to maybe fix when we don't have a deadline looming over us, but IMO we should redesign tests so these field renames/deletions make snapshot tests go red. CI checks just passed even though a number of acmeCo files keep the old field names.

Our discover snapshot does currently cover field renames & deletions - I had to update `snapshots__discover__capture.stdout.json` with those updates in order to get CI to pass. Those `.schema.yaml` files under `acmeCo` aren't actually used unless there's a capture snapshot test, and this connector doesn't have a capture snapshot test. I'd say we either add one, or if we can't for whatever reason, remove the `acmeCo` directory and all the bindings in `test.flow.yaml`.

> And a small nit on the documentation side: while the text itself doesn't need updating, URLs point to deprecated API versions. Only flagging it because it's something easy for Claude to figure out in the background.

Good point, this is done.

PR #4880, opened 2026-07-23
source-{python}: remove purely proof of concept connectors

Alex-Bair:
**Regression?**

No.

**Description:**

The `source-pokemon` and `source-yahoo-finance` connectors are early proof-of-concept connectors for our `python` library - our initial attempt at a CDK before the `estuary-cdk`. We've never published these two connectors for general use, and we keep getting dependabot updates for them, so I'd like to remove them from the repo .

**Notes for reviewers:**

For reference, here's the [PR](https://github.com/estuary/connectors/pull/970) that introduced these two connectors.

I'd also like to remove the `python` library eventually since the `estuary-cdk` is the library we've been building out & using to extract data from SaaS APIs, but `python` contains a shim that we use to run a small handful of singer taps like `source-shopify` so we'd need to sort that out first before removing our `python` library.

**Checklist:**


nicolaslazo (approved):
LGTM. A reference to `source-pokemon` remains in `python/README.md` but, like you mentioned, we'll be deleting that subdirectory soon.

Alex-Bair:
> A reference to source-pokemon remains in python/README.md but, like you mentioned, we'll be deleting that subdirectory soon.

Removing the `python` subdirectory is certainly a long-term goal, but I don't think we'll realistically be able to do it any time soon since we we have captures using the `source-shopify` and `source-criteo` connectors that are built with it.

PR #4882, opened 2026-07-23
source-sqlserver: Always uppercase discovery queries

willdonnelly:
**Description:**

This is just a bit of code health / cruft cleanup work. As noted on the removed-in-this-PR `uppercase_discovery_queries` feature flag's comment, we believe that it's always safe to use uppercase identifiers for INFORMATION_SCHEMA tables but at the time we just needed to make the change available quickly to users running into issues with Turkish collations (where `i` capitalizes to `İ`) so we played it safe and added a flag.

But having two variants of each discovery query makes any other changes to the discovery logic more painful than they need to be and in fact `source-sqlserver-ct` dispensed with that and uses the uppercase forms consistently and has had zero issues, so we should really just clean up this whole bit of technical debt by making them unconditionally uppercase and removing the flag.

**Workflow steps:**

No user action is required. Most source-sqlserver captures will begin using a differently-capitalized but identically-behaved and more technically correct variant of their discovery queries. A small handful which previously had the `uppercase_discovery_queries` feature flag set will continue to use those forms, but the feature flag itself will now be meaningless and ignored.

None, this flag and behavior were undocumented.

**Checklist:**

### DIFF
--- a/source-sqlserver/CHANGELOG.md
+++ b/source-sqlserver/CHANGELOG.md
@@ -1,5 +1,14 @@
 # source-sqlserver
 
+## 2026-07-23
+
+### Changed
+- Discovery queries now always reference the `INFORMATION_SCHEMA` catalog using
+  its canonical uppercase names. This matches the stored catalog names in every
+  collation, so discovery works under case-mapping collations such as
+  Turkish/Azeri without any additional configuration. The internal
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM

PR #4887, opened 2026-07-23
materialize-s3-iceberg: add variant_columns option

### LINKED ISSUE #4466: Iceberg: support variant column type for object, array, and multi-type fields
Let's add a `materialize-iceberg` task-level option to use the Iceberg `variant` column type for object, array, and multi-type fields.

### Parquet Variant

The Parquet `variant` binary encoding originated in Apache Spark's VARIANT type and was standardized in the Apache Parquet format spec (V1) in 2024. It provides the on-disk representation that Iceberg v3 adopts directly.

### Iceberg v3

[Iceberg v3 adds the `variant` semi-structured type](https://iceberg.apache.org/spec/#semi-structured-types), using the [Parquet Variant V1](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) binary encoding (`metadata` + `value` fields). Variant supports richer primitives than JSON (`date`, `timestamp`, `timestamptz`, `binary`, `decimal`), nested arrays and objects, and disallows bucket/identity partitioning.

### Flow

Flow documents are JSON, schematized via JSON Schema. Today, `materialize-iceberg` maps any field that isn't pinned to a single primitive (open `type: object`, `oneOf`/`anyOf`, multi-type, unconstrained) to a JSON-as-text column. Variant would preserve the semi-structured shape natively and avoid bespoke string parsing downstream.

### Reader compatibility

Variant requires Iceberg v3 readers. Validate writes with a **DuckDB-based unit test** covering nested objects, arrays, mixed-type fields, and nulls.

### Schema evolution

No type migration is supported.

jacobmarble:
**Description:**

Closes #4466. New advanced option `variant_columns` (default off) for `materialize-s3-iceberg`: object, array, and multi-type fields — and the root document — materialize as Iceberg v3 `variant` columns (unshredded Parquet Variant V1) instead of JSON strings. Collection keys keep their string mapping; per-field `castToString` still forces a string column.

- `go/writer` gains a variant parquet column kind (JSON text → Variant V1 `metadata`/`value` group, field IDs embedded, no binary column stats).
- New tables are created as format v3; a user-supplied `format-version=2` property fails loudly.
- Enabling on an existing materialization upgrades tables v2→v3 and converts affected columns in place (same name, new field ID) in one atomic commit; prior rows read null going forward, full history stays readable via time travel. Disabling migrates variant→string the same way.

**Workflow steps:**

Set `advanced.variant_columns: true` on the materialization. Existing bindings convert on the next apply; backfill a binding to re-encode history as variant.

`docs/reference/Connectors/materialization-connectors/amazon-s3-iceberg.md` updated in this PR (new config row + a "variant columns" section).

**Notes for reviewers:**

- Two TDD pairs: failing tests land before each implementation commit (`go/writer`, then the connector).
- iceberg-go's `UpdateSchema.AddColumn` rejects `VariantType`, so variant schema updates stage string placeholders and commit the rewritten successor schema via `CommitTable` with schema-ID + UUID assertions (`commitSchemaWithVariants`).
- Writes are verified by an independent reader: pinned dockerized DuckDB 1.5.4 (`read_parquet` in `go/writer`, catalog attach in the connector integration tests). All parquet snapshot reads now go through that pinned image; existing snapshots are byte-unchanged.
- Name-mapping scrub on migration means time travel reads pre-flip rows fully populated only for field-ID-embedded files (same trade-off as the `nanosecond_timestamps` migration); pinned in `runVariantMigrateTest`.

**Checklist:**


go/writer/parquet.go
```diff
@@ -577,6 +601,56 @@ func writeColumn[T parquetValue](
...
+
+		jsonVal, err := getJsonVal(row[colIdx])
+		if err != nil {
+			return fmt.Errorf("getting JSON value: %w (type %T)", err, row[colIdx])
+		}
+
+		encoded, err := variant.ParseJSONBytes(jsonVal, false)
+		if err != nil {
+			return fmt.Errorf("encoding JSON value as variant: %w", err)
+		}
```
danielnelson: This feels kind of expensive since we end up marshaling to JSON and back again.  Should be pretty good if the data is already `json.RawMessage` but not so good on other inputs.

It's also not clear to me if the types are the same as they would be going through the non-variant paths, such as if a type like: [{type: string, format: date-time}, {type: integer}], will this be a [timestamp, int] or a [string, int]?  There is a lot of logic in the various getTypeVal functions that doesn't seem replicated here.

go/writer/test_support.go
```diff
@@ -18,6 +21,41 @@ import (
...
+// that build on this writer. DuckDB 1.5.3 is the earliest release that reads
+// variant columns.
+const DuckDBDockerImage = "duckdb/duckdb:1.5.4"
+
+// duckdbDockerQueryFile runs query against the parquet file f using a pinned
+// dockerized DuckDB, substituting the container-side path of f for %s in
+// query, and returns the result rows as JSON. Unlike duckdbReadFile's
+// in-process DuckDB, this is an independent reader whose version isn't tied
+// to the duckdb-go dependency pinned in go.mod.
+func duckdbDockerQueryFile(t *testing.T, f string, query string) string {
```
danielnelson: If we keep this, since this is only used in tests, put it in a `_test.go` file so it doesn't get compiled in.  I'm not sure to what degree the rest of this file can be put into one of these but...some other PR we should do that too.

go/writer/test_support.go
```diff
@@ -18,6 +21,41 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// DuckDBDockerImage pins the DuckDB used in tests as an independent reader of
+// parquet variant encoding and Iceberg v3 tables, here and in the connectors
+// that build on this writer. DuckDB 1.5.3 is the earliest release that reads
+// variant columns.
+const DuckDBDockerImage = "duckdb/duckdb:1.5.4"
```
danielnelson: Should we bump the go.mod to this version, or the latest version, in a separate PR?  I think it might be worth it to avoid the need for docker.

go/writer/parquet_test.go
```diff
@@ -152,6 +152,114 @@ func TestParquetWriterTimestampNanos(t *testing.T) {
...
+		cc, err := rgMeta.ColumnChunk(5)
+		require.NoError(t, err)
+		statsSet, err := cc.StatsSet()
+		require.NoError(t, err)
+		require.True(t, statsSet, "statistics must remain enabled for non-variant columns")
+	}
+
+	// Read the file back with an independent parquet variant reader to prove
+	// the encoding is accepted outside of arrow-go.
+	cupaloy.SnapshotT(t, duckdbDockerQueryFile(t, sink.Name(),
```
danielnelson: I think this test should be skipped on testing.Short() because it requires a local docker.

materialize-s3-iceberg/testdata/config-glue-variant.yaml
```diff
@@ -0,0 +1,36 @@
+# AWS Glue integration-test config.
+#
+# Disabled by default: TestIntegrationGlue only runs with -s3iceberg.test-glue
+# (or S3_ICEBERG_TEST_GLUE set) and needs real AWS credentials with Glue Data
+# Catalog + S3 access. Replace the REPLACE_ME placeholders below and sops-encrypt
+# this file before enabling the test. The connector creates the `namespace`
```
danielnelson: Is this comment out of date?  I don't see any REPLACE_ME placeholders or this test flag.

go/writer/parquet.go
jacobmarble: Good catch on both — I dug into each.

**On types:** they do match. Your example `[{type: string, format: date-time}, {type: integer}]` never reaches the `getTypeVal` logic on *either* path. As soon as a projection has a second non-null type, `ProjectionToParquetSchemaElement` breaks out to the JSON-shaped fallback and drops all format annotations, so that field is a JSON *string* column today and a variant holding a JSON *string* with the option on. So it's `[string, int]` both ways — no divergence, and nothing from `getTypeVal` that needs replicating. The one case that does keep its type is string-encoded numbers, which `AsFormattedNumeric` catches before that loop and which stay on their numeric column.

**On cost:** you're right that the round trip was real, so I've added a fast path. Objects, arrays, and the root document arrive as `json.RawMessage` and were already zero-copy into `ParseJSONBytes`, but a multi-type field's *scalars* were being marshaled to JSON just to be parsed straight back. Those now go to a `variant.Builder` directly:

| | before | after |
|---|---|---|
| string | 633 ns, 9 allocs | 127 ns, 3 allocs |
| int64 | 704 ns, 9 allocs | 141 ns, 4 allocs |
| float64 | 2345 ns, 23 allocs | 361 ns, 4 allocs |
| bool | 573 ns, 7 allocs | 127 ns, 3 allocs |

I added test coverage for native Go scalar inputs first, and the snapshot is byte-identical before and after. That test also turned out to be worth having — `Builder.Append` rejects `uint64` outright, so a `uint64` too large for the variant integer types explicitly falls back to the JSON path.

go/writer/test_support.go
jacobmarble: Done — and it turned out we didn't keep it at all. Bumping duckdb-go (your next comment) let me drop the dockerized reader entirely and use the in-process one, so `duckdbDockerQueryFile` and `DuckDBDockerImage` are both gone. The replacement helper lives in `parquet_test.go`.

go/writer/test_support.go
jacobmarble: Agreed, and done — separate PR is #5052, bumping duckdb-go to v2.10505.0 (DuckDB 1.5.5). This branch is rebased on it and no longer uses docker anywhere: `go/writer` and both readers in `materialize-s3-iceberg` are all in-process now.

go/writer/parquet_test.go
jacobmarble: This one resolved itself with the duckdb-go bump — the test now uses the in-process reader, so there's no docker requirement left to skip for. I left it running unconditionally, alongside the other tests here that use in-process DuckDB.

materialize-s3-iceberg/testdata/config-glue-variant.yaml
jacobmarble: Yes, fixed in two places.

go/writer/parquet.go
danielnelson: Ahh, that makes sense about the types, but now for maybe a harder question, should the types diverge?  Previously we were forced into going with JSON types but now we can utilize the parquet types too.  Maybe it ought to be [timestamp, int]?

jacobmarble: I'm leaning "no".

We only have format annotations for top-level projections, not for values nested inside an object, an array, or `flow_document`. So we'd encode `stringDateTimeField` as a variant timestamp at the top level but leave the identical value inside `flow_document` as a string; that's the same logical data, but two different variant types depending on where you read it.

Related: `flow_document` currently round-trips the document faithfully (the test data has `2023-01-15T10:30:45z`, lowercase z, and it comes back exactly that way). Typing it would normalize and lose the original spelling. For example, in the snapshot data, a dedicated `timestamptz` column converts to `2023-01-15 10:30:45+00` but inside a `flow_document` we get `2023-01-15T10:30:45z`. And it isn't only spelling that goes: a variant timestamp stores an instant, so the UTC offset goes with it — `2023-01-15T10:30:45+02:30` and `2023-01-15T08:00:45Z` would be indistinguishable once stored, and the reader picks how to render them.

Also, a multi-type field is where the schema is least pinned down, so it's the worst place to start trusting a `format` annotation enough to parse on it — a value that passes through fine today would start failing.

Fields with an unambiguous type aren't affected either way: `stringDateTimeField` is still a real `timestamptz` column with variant enabled.

danielnelson: It is typical that the values in the `flow_document` are different from the target columns, this would be the case for a non-variant `{type: string, format:date-time}` field too.  On a field with format date-time but not actually parseable as a timestamp, my understanding is that this shouldn't be possible. If the field had a non-datetime string then schema inference would be widened or it would fail to validate.

I was kind of leaning towards the variants being how they would be as standalone fields, but it would be quite a bit more involved, but I'll leave it up to you.

danielnelson (dismissed):

go/writer/parquet.go
jacobmarble: Agreed, I've implemented the suggestion.

A date-time string in a variant column is now stored as a variant timestamp, date as a date, and base64 as binary, using the same resolution the standalone column gets. Formats we lower to strings anyway (uuid, time, duration) stay strings, and nanosecond_timestamps carries through.

Only top-level projections have format annotations, so a date-time nested inside an object or flow_document is still a string — there's a test pinning that.

Tested queryability with DuckDB.

danielnelson (approved):

PR #4889, opened 2026-07-24
source-iterate: allow extra fields on API response models

Alex-Bair:
**Regression?**

No.

**Description:**

We've observed captures failing because Iterate added a `project` field to the response when fetching survey responses. Based on the [API docs](https://iteratehq.com/api-docs#get-responses), the `project` should already be present in the documents we emit, so it's safe to just ignore `project` in the response wrapper. I'm opting to ignore all other, non-essential fields in response wrappers as well to avoid similar task failures in the future.

**Checklist:**

### DIFF
--- a/source-iterate/source_iterate/models.py
+++ b/source-iterate/source_iterate/models.py
@@ -29,18 +29,18 @@ class FullRefreshResource(BaseDocument, extra="allow"):
     pass
 
 
-class SurveysResponse(BaseModel, extra="forbid"):
+class SurveysResponse(BaseModel, extra="ignore"):
     results: list[FullRefreshResource]
 
 
-class SurveyResponsesResponse(BaseModel, extra="allow"):
-    class Results(BaseModel, extra="forbid"):
+class SurveyResponsesResponse(BaseModel, extra="ignore"):
+    class Results(BaseModel, extra="allow"):
         count: int
         list: list[FullRefreshResource]
 
     results: Results
 
-    class Links(BaseModel, extra="forbid"):
+    class Links(BaseModel, extra="ignore"):
         next: str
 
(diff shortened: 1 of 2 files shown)


source-iterate/source_iterate/models.py
```diff
@@ -29,18 +29,18 @@ class FullRefreshResource(BaseDocument, extra="allow"):
...
 class SurveyResponsesResponse(BaseModel, extra="allow"):
-    class Results(BaseModel, extra="forbid"):
+    class Results(BaseModel, extra="allow"):
         count: int
         list: list[FullRefreshResource]
 
     results: Results
 
-    class Links(BaseModel, extra="forbid"):
+    class Links(BaseModel, extra="allow"):
```
nicolaslazo:
```suggestion
class SurveysResponse(BaseModel, extra="ignore"):
    results: list[FullRefreshResource]


class SurveyResponsesResponse(BaseModel, extra="ignore"):
    class Results(BaseModel, extra="allow"):
        count: int
        list: list[FullRefreshResource]

    results: Results

    class Links(BaseModel, extra="ignore"):
```

If we won't use these fields I'd suggest we skip them entirely... the performance and memory improvements are negligible, but we'd be more declarative about our intentions

nicolaslazo (dismissed):
Makes sense! I just left a small nit

source-iterate/source_iterate/models.py
nicolaslazo: Wow GitHub really messed up that diff

Alex-Bair: Good point, thanks! I've made those updates.

nicolaslazo (approved):
🙌

PR #4894, opened 2026-07-24
source-braintree-native: retry 422 responses indicating timeouts

### LINKED ISSUE #4892: source-braintree-native: disputes advanced_search 422 <reason>timeout</reason> crashes shard instead of retrying
## Summary

During a large `disputes` backfill, `source-braintree-native` fails the shard when Braintree's `disputes/advanced_search` endpoint returns **HTTP 422** with a body of `<unprocessable-entity><reason>timeout</reason></unprocessable-entity>`. This is a transient server-side search timeout, but the CDK classifies 422 as a non-retryable client error, so it raises `HTTPError` and crashes the binding task.

## Observed behavior

Under sustained load, Braintree returns two distinct errors:

- `504 Gateway time-out` (Cloudflare HTML) — retried correctly by the CDK, non-fatal.
- `422 <unprocessable-entity><reason>timeout</reason></unprocessable-entity>` on `disputes/advanced_search` — **not** retried, fails the shard.

Representative traceback:

```
File ".../source_braintree_native/api/disputes.py", line 111, in fetch_disputes_between
    async for dispute in process_completed_fetches(
File ".../source_braintree_native/api/common.py", line 36, in process_completed_fetches
    result = await coro
File ".../source_braintree_native/api/disputes.py", line 39, in _fetch_dispute_page
    await http.request(log, url, "POST", params, body, headers=HEADERS)
File ".../estuary_cdk/http.py", line 663, in _request_stream
    raise HTTPError(
estuary_cdk.http.HTTPError: Encountered HTTP error status 422 which cannot be retried.
URL: https://api.braintreegateway.com/.../disputes/advanced_search?page=N
Response: <unprocessable-entity><reason>timeout</reason></unprocessable-entity>
```

Fails from both `backfill_disputes` and `fetch_disputes` paths, across many page numbers.

## Secondary symptom (not the root cause)

When one page's 422 fails the binding, the TaskGroup tears down and closes the shared aiohttp session while other `_fetch_dispute_page` tasks (spawned eagerly by `asyncio.as_completed` in `process_completed_fetches`) are still in flight. Those surface as noisy, non-fatal `Task exception was never retrieved ... RuntimeError('Session is closed')` warnings, which can mislead diagnosis.

## Why it happens

Disputes fan out all pages up front through `process_completed_fetches` (semaphore of `SEMAPHORE_LIMIT = 20`), and the dispute search window has a hard 24h floor (the search field is a date, not a date-time, so it can't be narrowed further, and disputes don't bisect the window the way transactions do). Under that concurrency Braintree's search backend times out on individual page queries and returns the 422.

## Proposed fixes

1. **Treat the 422 `<reason>timeout</reason>` as retryable with backoff** — pass a custom `should_retry` to `http.request` in `disputes.py` that inspects the body for the timeout reason. This mirrors the 4xx-matching approach used in other connectors (e.g. source-facebook-marketing / source-apple-app-store) where a documented 4xx represents a retryable/too-much-work condition. Primary fix.
2. **Make `process_completed_fetches` clean up on early exit** — cancel outstanding tasks and `await asyncio.gather(..., return_exceptions=True)` in a `finally`, so a single page failure doesn't leave orphaned "Session is closed" tasks. Removes the misleading warnings.
3. **Consider lowering disputes concurrency and/or making `SEMAPHORE_LIMIT` configurable**, since the dispute window can't shrink below 24h. Open question pending Braintree's guidance on hard caps / recommended backoff.

Alex-Bair:
**Regression?**

No.

**Description:**

Closes #4892 

We've observed the `disputes` & other streams crash on large captures when Braintree responds with a 422 status code with a body containing `<reason>timeout</reason>`. The `estuary-cdk` HTTP layer treats all 4xx as non-retryable, so these fail the task when retrying the request may succeed.

This PR's scope includes:
1. Retry transient 422 search timeouts across all search requests, with exponential backoff, matching only 422s whose body carries `<reason>timeout</reason>` and re-raising everything else. Exhausted retries still propagate, restarting the task from its last cursor.
2. Cancel outstanding fetches on error in `process_completed_fetches`, so sibling tasks don't outlive the generator and emit `Session is closed` warnings.

**Notes for reviewers:**

Until we have a better understanding of what's causing the 422s on Braintree's side (the number of concurrent queries? the sheer number of requests?), I'd rather not adjust the connector's concurrency yet.

The 422 retry behavior is _probably_ better suited to be handled at the CDK HTTP client layer, but the `should_retry` hook that we'd use for that only applies to 5xx responses. I'd rather not make a CDK-wide changes to the `should_retry` behavior on a Friday afternoon, so I stuck with the less ideal but still workable connector specific implementation.

**Checklist:**

### DIFF
--- a/source-braintree-native/CHANGELOG.md
+++ b/source-braintree-native/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2026-07-29
+
+### Fixed
+- Transient Braintree search timeouts (`HTTP 422` responses with a `timeout` reason) are now retried instead of failing the capture.
+
 ## 2026-07-27
 
 ### Added
--- a/source-braintree-native/source_braintree_native/api/common.py
+++ b/source-braintree-native/source_braintree_native/api/common.py
@@ -69,11 +69,22 @@ async def request_with_search_timeout_retries(
 async def process_completed_fetches(
     fetch_coroutines: list[Awaitable[list[dict[str, Any]]]],
 ) -> AsyncGenerator[dict[str, Any], None]:
-    """Helper to process fetching multiple pages of resources and yield individual resources."""
-    for coro in asyncio.as_completed(fetch_coroutines):
-        result = await coro
-        for resource in result:
-            yield resource
+    """Helper to process fetching multiple pages of resources and yield individual resources.
+
+    If one fetch raises or the consumer stops iterating early, the remaining in-flight
+    fetches are cancelled so they don't outlive this generator and emit "Session is closed"
+    warnings when they touch the torn-down HTTP session.
+    """
+    tasks = [asyncio.ensure_future(coro) for coro in fetch_coroutines]
+    try:
+        for task in asyncio.as_completed(tasks):
+            result = await task
+            for resource in result:
+                yield resource
+    finally:
+        for task in tasks:
+            task.cancel()
+        await asyncio.gather(*tasks, return_exceptions=True)
 
 
 def braintree_xml_to_dict(xml_data):
--- a/source-braintree-native/source_braintree_native/api/common.py
+++ b/source-braintree-native/source_braintree_native/api/common.py
@@ -1,4 +1,5 @@
 import asyncio
+import random
 from braintree import (
     BraintreeGateway,
     AddOnGateway,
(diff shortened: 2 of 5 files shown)


Alex-Bair:
Note: priority-wise, we should merge https://github.com/estuary/connectors/pull/4908 first since the number of concurrent `disputes` searches appear to be main issue. Then once that's merged, I'll rebase these changes and we can merge this after it's approved.

nicolaslazo (dismissed):
LGTM

nicolaslazo (approved):
👍

PR #4895, opened 2026-07-24
source-mailchimp-native: new connector

### LINKED ISSUE #4308: source-mailchimp-native: New connector
### System Name

Mailchimp

### Type

Capture

### Details

Native connector that replicates the current one but...works.

nicolaslazo:
**Regression?**

**Description:**

Adds a new capture connector for [Mailchimp](https://mailchimp.com/), the email and SMS marketing platform.

Closes https://github.com/estuary/connectors/issues/4308

**Workflow steps:**

**Notes for reviewers:**

Tested using a local deployment.

**Checklist:**


.claude/skills/create-webhook-connector/SKILL.md
nicolaslazo, line 1: This skill is mostly vestigial and I intend to merge it into `create-capture-connector` soon, leaving only the webhook-specific knowledge.

source-mailchimp-native/source_mailchimp_native/resources.py
```diff
@@ -0,0 +1,168 @@
...
+    Campaign,
+    EndpointConfig,
+    MailchimpEntity,
+    MailchimpList,
+)
+
+DEFAULT_SCHEDULE = "0 0 * * *"
+
+# Standard tombstone for snapshot resources (CDK convention)
+TOMBSTONE = BaseDocument(_meta=BaseDocument.Meta(op="d"))
```
Alex-Bair: nit: If a connector is using the CDK's `BaseDocument` model for tombstones, it's not necessary to provide them when instantiating a resource - the CDK automatically uses a `BaseDocument` tombstone if none were provided.
https://github.com/estuary/connectors/blob/2d033a55084ca5f6960d643bc1fb6479387c3f16/estuary-cdk/estuary_cdk/capture/common.py#L834-L839

source-mailchimp-native/source_mailchimp_native/resources.py
```diff
@@ -0,0 +1,168 @@
...
+            tombstone=TOMBSTONE,
+        )
+
+    return [
+        SnapshotResource[BaseDocument, ResourceConfigWithSchedule](
+            name=model.NAME,
+            model=model,
+            open=functools.partial(open, model.NAME),
+            initial_config=ResourceConfigWithSchedule(
+                name=model.NAME, interval=timedelta(hours=1), schedule=""
```
Alex-Bair: nit: The default schedule is `""`, so we don't need to specify one here.

```suggestion
                name=model.NAME, interval=timedelta(hours=1)
```

source-mailchimp-native/source_mailchimp_native/resources.py
```diff
@@ -0,0 +1,168 @@
...
+        initial_state=ResourceState(
+            inc=ResourceState.Incremental(cursor=cutoff),
+            backfill=ResourceState.Backfill(cutoff=cutoff, next_page=None),
+        ),
+        initial_config=ResourceConfigWithSchedule(
+            name=Campaign.NAME,
+            interval=timedelta(minutes=5),
+            schedule=(
+                DEFAULT_SCHEDULE if Campaign.NAME in SCHEDULED_BACKFILL_STREAMS else ""
+            ),
```
Alex-Bair: nit: Since the `campaigns` function only creates a single resource, could we simplify the schedule to always be the `DEFAULT_SCHEDULE`?

```suggestion
            schedule=DEFAULT_SCHEDULE,
```

source-mailchimp-native/source_mailchimp_native/resources.py
```diff
@@ -0,0 +1,419 @@
...
+        )
+
+    return MailchimpResource(
+        name=EmailActivity.NAME,
+        key=["/_meta/campaign_id", "/_meta/email_id", "/action", "/timestamp"],
+        model=EmailActivity,
+        open=open,
+        initial_state=ResourceState(
+            # cutoff − 1s, per the note in _patch_missing_subtask_states.
+            inc=ResourceState.Incremental(cursor=cutoff - timedelta(seconds=1)),
```
Alex-Bair: nit: IMO, instead of referring to a note elsewhere, it'd be clearer to restate that reason for subtracting a second here in a comment or using a clearly named constant like `cutoff - API_LAG` (or whatever name is appropriate).

source-mailchimp-native/source_mailchimp_native/resources.py
```diff
@@ -0,0 +1,419 @@
...
+            backfill=ResourceState.Backfill(cutoff=cutoff, next_page=None),
+        ),
+        initial_config=ResourceConfigWithSchedule(
+            name=EmailActivity.NAME,
+            interval=timedelta(minutes=20),
+            schedule="",
```
Alex-Bair: nit: same here, schedule defaults to `""` if not provided.

source-mailchimp-native/source_mailchimp_native/models.py
```diff
@@ -0,0 +1,561 @@
...
+    @model_validator(mode="before")
+    @classmethod
+    def _inject_parent_ids(cls, data: Any, info: ValidationInfo) -> Any:
+        if isinstance(data, dict) and isinstance(
+            info.context, ParentIdValidationContext
+        ):
+            meta = dict(data.get("_meta") or {})
+            meta.update(info.context.fields)
+            data["_meta"] = meta
+        return data
```
Alex-Bair, line 211: I think we've done this "inject some relevant fields into `_meta`" in other CDK-based connectors too, right? Eventually (i.e. not in the PR), it'd be nice if we could extract some of this into the CDK so adding fields to `_meta` is more ergonomic. Maybe something like:
```python
  # estuary-cdk/estuary_cdk/capture/document.py
  @dataclass(frozen=True)
  class MetaFields:
      """Validation context stamping connector-supplied fields into `_meta`."""
      fields: dict[str, Any]

  # We shouldn't put the model validator on BaseDocument in order to avoid
  # forcing all connectors using BaseDocument to take the performance hit
  # it incurs. So, we could create a new class (name TBD) for this specific use case.
  class BaseDocumentWithAdditionalMetaFields(BaseDocument):
      """BaseDocument whose `_meta` carries connector-supplied scope fields."""

      @model_validator(mode="before")
      @classmethod
      def _inject_meta_fields(cls, data: Any, info: ValidationInfo) -> Any:
          if isinstance(data, dict) and isinstance(info.context, MetaFields):
              data["_meta"] = {**(data.get("_meta") or {}), **info.context.fields}
          return data
  ```
then usage would be like:
```python
validation_context=MetaFields({"segment_id": segment_id})
```

Again, this isn't something we need to tackle right now, but we've seen enough connectors need to add fields into `_meta` that I feel we have enough examples to abstract some of this common boilerplate into the CDK.

source-mailchimp-native/source_mailchimp_native/models.py
```diff
@@ -0,0 +1,561 @@
...
+
+    def iter_events(self) -> Iterator["EmailActivity"]:
+        """Flatten this object into per-event documents: the member envelope is
+        stamped into `_meta`."""
+
+        envelope = self.model_dump(exclude={"activity"})
+        for event in self.activity:
+            yield EmailActivity.model_validate(
+                {**event.model_dump(exclude_unset=True), "_meta": envelope}
+            )
```
Alex-Bair: Pydantic validation has a performance cost, and it'd be nice to avoid the additional model validation here in `iter_events` if possible since `email_activity` is likely to be a high volume stream. Could we restructure so `activity[]` parses directly into the shape we need? For example, instead of `activity: list[EmailActivityEvent]`, could we define it as `activity: list[EmailActivity]` then inject the fields we need into `_meta`?
```python
class MemberEmailActivity(BaseModel, extra="allow"):
    campaign_id: str
    email_id: str
    activity: list[EmailActivity]  # ← was list[EmailActivityEvent]

    @model_validator(mode="before")
    @classmethod
    def _inject_event_meta(cls, data: Any) -> Any:
        # inject {campaign_id, email_id} onto each raw activity entry
```

Additionally, I think the current code injects other fields into `_meta` that we may not want to include, like `_links`. The restructure above avoids that by just injecting the envelope fields we specify instead of all of them.

source-mailchimp-native/source_mailchimp_native/api/shared.py
```diff
@@ -0,0 +1,511 @@
...
+)
+
+# Mailchimp's base URL is data-center-dependent. The `{dc}` placeholder (e.g. "us6")
+# must be resolved from the configured credentials at runtime via `resolve_base_url`
+# before any request is made.
+API = "https://{dc}.api.mailchimp.com/3.0"
+OAUTH_METADATA_URL = "https://login.mailchimp.com/oauth2/metadata"
+
+
+async def resolve_base_url(log: Logger, http: HTTPSession, credentials: object) -> str:
```
Alex-Bair: nit: Could we narrow the typing on `credentials` here? Perhaps it could be `credentials: OAuth2Credentials | ApiKey`?

source-mailchimp-native/source_mailchimp_native/api/shared.py
```diff
@@ -0,0 +1,511 @@
...
+    """Resolve the account's data-center-specific base URL.
+
+    API keys carry their data center as a suffix (`<key>-us6`); OAuth tokens
+    require asking Mailchimp's metadata endpoint, which reports the account's
+    `api_endpoint` (e.g. "https://us6.api.mailchimp.com").
+    """
+    if isinstance(credentials, ApiKey):
+        _, _, dc = credentials.password.rpartition("-")
+        if not dc:
```
Alex-Bair: Is the `if not dc` condition ever truthy? From the Python docs on [`str.rpartition`](https://docs.python.org/3.12/library/stdtypes.html#str.rpartition) (emphasis mine):
>Split the string at the last occurrence of sep, and return a 3-tuple containing the part before the separator, the separator itself, and the part after the separator. **If the separator is not found, return a 3-tuple containing two empty strings, followed by the string itself**.

So for a key with no separator `"-"` at all, then the entire password is stuffed into `dc` and `if not dc` evaluates to `False`.

Checking if the separator is present would cover the case where there's no `"-datacenter"` suffix:
```suggestion
        _, sep, dc = credentials.password.rpartition("-")
        if not sep or not dc:
```

source-mailchimp-native/source_mailchimp_native/api/shared.py
```diff
@@ -0,0 +1,511 @@
...
+        async for item in snapshot_collection(
+            http,
+            base_url,
+            path,
+            items_key,
+            IdOnly,
+            log,
+            params={"fields": f"{items_key}.id"},
+        )
+    ]
```
Alex-Bair: Should we sort the list of parent ids so it's in a stable order before returning it? Or does the API always return records in a stable order?

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
```diff
@@ -0,0 +1,301 @@
...
+            for doc in member.iter_events():
+                if doc.timestamp > horizon:
+                    continue
+                if not cache.should_yield(
+                    EmailActivity.NAME, doc.cache_key, doc.timestamp
+                ):
+                    continue
+
+                yield doc
+                emitted += 1
```
Alex-Bair: nit: The equivalent behavior could be done without the `not`, which I find is just a little easier to follow:
```suggestion
                if cache.should_yield(
                    EmailActivity.NAME, doc.cache_key, doc.timestamp
                ):
                    yield doc
                    emitted += 1
```

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
```diff
@@ -0,0 +1,301 @@
...
+
+    evicted = cache.cleanup(EmailActivity.NAME, horizon - LOOKBACK)
```
Alex-Bair: Is it possible to emit a duplicate still if we're moving the log cursor forward to the `horizon` only when we emitted a document but entries older than `horizon - LOOKBACK` are evicted from the cache on each `fetch_email_activity` invocation?

Consider this scenario. With `LOOKBACK=10m`, `CONSISTENCY_LAG=5m` and the default `interval: PT20M`, take one event at 12:00 with the cursor at 11:45, after which the account goes quiet:
  | poll  | `horizon` | `since` (`cursor − LOOKBACK`) | re-read finds | in cache? | `emitted` | cursor after | `cleanup` evicts ≤ |
  | ----- | --------- | ----------------------------- | ------------- | --------- | --------- | ------------ | ------------------ |
  | 12:10 | 12:05     | 11:35                         | 12:00 event   | no        | 1         | 12:05    | 11:55 - kept       |
  | 12:30 | 12:25     | 11:55                         | 12:00 event   | yes       | 0     | 12:05 (held) | 12:15 - evicted|
  | 12:50 | 12:45     | 11:55                         | 12:00 event   | no    | 1         | 12:45    | 12:35              |
  | 13:10 | 13:05     | 12:35                         | nothing       | —         | 0         | 12:45 (held) | 12:55              |
  | 13:30 | 13:25     | 12:35                         | nothing       | —         | 0         | 12:45 (held) | 13:15              |

`cleanup` runs unconditionally against the current `horizon`, but the cursor only moves when something was emitted. So the 12:30 poll evicts an entry that the 12:50 poll still re-reads, and the event is emitted a second time. It settles after that - once the cursor reaches 12:45 the event falls outside `since` and can't be re-read.

If we only evict entries older than `since` (i.e. the absolute lowest bound for this `fetch_email_activity` invocation), then we avoid emitting duplicates in scenarios like the one described above.

```suggestion
    evicted = cache.cleanup(EmailActivity.NAME, since)
```

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+@dataclass(frozen=True)
+class _EmailActivityPageCursor:
+    """Decoded backfill `PageCursor` ("send_time|campaign_id"): the
+    (send_time, campaign_id) watermark of the last fully-drained campaign in
+    the gate-selected set."""
+
+    send_time: datetime
+    campaign_id: str
```
Alex-Bair: Would adding the `offset` we should use in the next `backfill_email_activity` pass & page through a given campaign to `_EmailActivityPageCursor` help us make durable progress through campaigns with large amounts of email activity? I seems like we could checkpoint in the middle of fetching email activity of a single campaign rather than only in between campaigns.

source-mailchimp-native/source_mailchimp_native/resources.py
```diff
@@ -0,0 +1,419 @@
...
+            state,
+            task,
+            fetch_snapshot=snapshot_fetchers[resource_name],
+            tombstone=TOMBSTONE,
+        )
+
+    return [
+        SnapshotResource[BaseDocument, ResourceConfigWithSchedule](
+            name=model.NAME,
+            model=model,
```
Alex-Bair: nit: When using `SnapshotResource`, you don't have to provide a `model` during instantiation. It defaults to using `BaseDocument`, which looks to be sufficient here since `/_meta/row_id` is the only primary key component.

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
```diff
@@ -0,0 +1,301 @@
...
+
+    if emitted:
+        yield horizon
+
+
+@dataclass(frozen=True)
+class _EmailActivityPageCursor:
+    """Decoded backfill `PageCursor` ("send_time|campaign_id"): the
+    (send_time, campaign_id) watermark of the last fully-drained campaign in
+    the gate-selected set."""
```
Alex-Bair: `PageCursor`s can be `dict`s, and using a structured `dict` would be a bit clearer than encoding information into a `str`. Could we do that instead? You'll need to use [these helpers](https://github.com/estuary/connectors/blob/e5eca53ba567da6258b9b656251737e7fc7cdcf7/estuary-cdk/estuary_cdk/capture/common.py#L55-L116) in the CDK when creating the `dict` cursor.

I try to avoid using `dict` page cursors as much as possible since it can be challenging to figure out what's actually _in_ that `dict` & what it represents, but there are rare situations where a backfill strategy requires a more complex `PageCursor` than what we normally use. I'd suggest continue having a dataclass describing the shape of the cursor `dict` so the general shape & meaning of fields are well documented.

[`source-monday`](https://github.com/estuary/connectors/blob/e5eca53ba567da6258b9b656251737e7fc7cdcf7/source-monday/source_monday/models.py#L555-L594) is currently our only CDK connector using a `dict` `PageCursor`, and it could serve as a reference for how to use them since there are a few non-obvious things you need to be aware of to avoid subtle bugs (ex: each emitted cursor `dict` needs to carry a marker so the CDK can determine whether an emitted `dict` is a document or a `PageCursor`, to remove fields from the persisted `dict` cursor they need to be `null`ed out per JSON merge patch semantics).

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
```diff
@@ -0,0 +1,301 @@
...
+                http,
+                base_url,
+                EmailActivity.PATH_TEMPLATE.format(campaign_id=campaign.id),
+                EmailActivity.ITEMS_KEY,
+                MemberEmailActivity,
+                {
+                    "count": MAX_PAGE_SIZE,
+                    "offset": offset,
+                    EmailActivity.SINCE_PARAM: start_date.isoformat(),
+                },
```
Alex-Bair: It looks like the [`/reports/{campaign-id}/email-activity` endpoint](https://mailchimp.com/developer/marketing/api/email-activity-reports/) supports an `exclude_fields` query parameter that claims to exclude certain fields from the API response. Would that be helpful to exclude the `_links` field? It sounds like that field can contain large values that we don't use anywhere?

source-mailchimp-native/source_mailchimp_native/models.py
```diff
@@ -0,0 +1,561 @@
...
+        def _empty_to_none(cls, value: object) -> object:
+            return None if value == "" else value
+
+    class Bounces(BaseModel):
+        # Count only — `/reports` carries no bounce timestamps, which is why
+        # the settle-window gate clause exists (see `fetch_email_activity`).
+        soft_bounces: int
+
+    id: str
+    send_time: AwareDatetime
```
Alex-Bair, line 452: Does `send_time` need an `_empty_to_none` field validator too? From one of the [bruno files](https://github.com/estuary/connectors/blob/ea3cc84d60884f59eab3edc5b8006453a4ee0acf/source-mailchimp-native/bruno/Campaigns/status%20filter.yml#L59-L63), it sounds like `send_time` can be an empty string for drafts?

source-mailchimp-native/source_mailchimp_native/resources.py
Alex-Bair: Additionally to explain why this matters, the schema of the `model` becomes the write schema in a collection spec. Any fields required on the write schema must be present on all documents written to a collection, **including tombstones**. So, let's say you have a full refresh stream whose key is `/_meta/row_id` and is saying there's a required `id` field on the write schema. Regular documents that contain `id` enter the collection fine since they comply with the write schema. But once a tombstone is emitted that's missing an `id` field, the runtime fails the capture with a JSON schema violation that says "Hey, this document (the tombstone) didn't match the write schema you told me about!".

That's why `SnapshotResource` defaults to using `BaseDocument` for both the `model` (to generate the write schema used in the collection) and the tombstone. It avoids tripping JSON schema violations for the garden path use case.

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
```diff
@@ -0,0 +1,301 @@
...
+
+    emitted = 0
+
+    email_campaigns = await _fetch_email_campaign_stubs(http, base_url, log)
+    selected = [
+        campaign
+        for campaign in email_campaigns
+        if campaign.has_activity_after(lookback_floor)
+        or campaign.has_soft_bounces_settling(settle_floor)
+    ]
```
Alex-Bair, line 112: Could you please remind me about why we don't select for hard bounces at all? I understand that after an email is sent, hard bounces typically occur very quickly, but are all emails sent right on a campaign's `send_time`? Or is it possible for emails to be sent out a little later, in which case hard bounces could come in minutes after a campaign's `send_time`?

Alex-Bair: And another question whose answer may help - is it possible for a campaign to only have hard bounces (no opens, no clicks, no soft bounces)? If so, I don't think we'd capture any of the activity for that campaign, assuming the hard bounce only situation is possible.

Alex-Bair (commented):
I haven't reviewed the Claude skills yet (although it sounds like with the removal of over 80% of Claude Code's system prompt recently, it'd probably be worthwhile to go back over the skills with the `claude doctor` command you mentioned in Slack). But I've reviewed the `source-mailchimp-native` code and had a few comments & questions.

source-mailchimp-native/source_mailchimp_native/models.py
nicolaslazo: Hard agree. We should strive to keep the top level namespace unaltered, so we need to settle an ergonomic alternative.

source-mailchimp-native/source_mailchimp_native/models.py
nicolaslazo: That's true, `send_time` can be `""`, but we won't encounter that problem in our use case. `Campaigns` does not model the field, and draft campaigns are ignored when we sweep the `/reports` endpoint for campaigns with email activity.

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
nicolaslazo: Good catch, I forgot that hard bounces didn't update any timestamps either.

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
nicolaslazo: That'll make documents significantly smaller. I've excluded the fields server-side and added new Bruno request to document the behaviour

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
nicolaslazo: I didn't know those helpers existed! I updated our Claude rules to use it — hopefully it doesn't come up too often, but now it's indexed and discoverable.

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
nicolaslazo: We discussed this on our 1:1, but I found no evidence that email ordering is stable across time and I'd rather we didn't run that risk until it proves impractical.

source-mailchimp-native/source_mailchimp_native/api/email_activity.py
nicolaslazo: You have a good point here, and while artificially manipulating the time bounds on both ends was more cache-efficient, it was hard to wrap one's head around it. I removed the `CONSISTENCY_LAG` mechanism, now we yield the last emitted document's timestamp instead.

source-mailchimp-native/source_mailchimp_native/api/shared.py
nicolaslazo: Good point, I didn't realise snapshot hashes are order-dependent and reordered docs would cause document re-emissions.

Alex-Bair: Yep! That's why whenever we try to parallelize API calls to speed up snapshot streams, we use our custom `buffer_ordered` function like here in `source-looker` to maintain ordering instead of `asyncio.as_completed` or some standard library helper.
https://github.com/estuary/connectors/blob/20a3139db572293272b02ed8150644b5a932a36d/source-looker/source_looker/api.py#L165-L167

Alex-Bair (approved):
LGTM.

I'm not familiar with the new stacking GitHub feature you've applied to this PR, but I'm assuming the other two PRs in the "stack" need approved before any of them land on `main`?

nicolaslazo:
> LGTM.
> 
> I'm not familiar with the new stacking GitHub feature you've applied to this PR, but I'm assuming the other two PRs in the "stack" need approved before any of them land on `main`?

Not necessarily, they can be merged individually as long as one does not depend on the other. Here, I'm only using stacks to illustrate the merge sequence (and because I was curious about how stacks worked).

PR #4898, opened 2026-07-26
source-ashby: rely on CDK defaults for SnapshotResource model and tombstone

Alex-Bair:
**Regression?**

No.

**Description:**

We've observed `source-ashby` captures failing when a tombstone is written to collections for snapshot bindings. The failures are due to additional required fields specified in the collection's write schema (which is determined by the `model` provided during `Resource`/`SnapshotResource` instantiation) that are not in the emitted tombstones.

This PR fixes this by using the CDK defaults for `SnapshotResource`s- `BaseDocument` for the model as well as for the tombstone. This makes the write schema align with the fields specified in deletion tombstones & address the errors we observed.

Discover snapshot changes are expected due to the removal of additional fields from snapshot bindings' schemas.

**Checklist:**

### DIFF
--- a/source-ashby/CHANGELOG.md
+++ b/source-ashby/CHANGELOG.md
@@ -1,0 +1,8 @@
+# Changelog
+
+## 2026-07-27
+
+### Changed
+- The `archive_reasons`, `interview_stages`, `job_postings`, and `sources` collections are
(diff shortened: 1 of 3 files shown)


nicolaslazo (approved):
Good catch! I wonder if there's a way we can make document tombstones final at the type level...

PR #4899, opened 2026-07-26
source-looker: skip explores & child resources that 404

Alex-Bair:
**Regression?**

No.

**Description:**

Snapshots of `lookml_model_explores` fail outright when any single explore returns a 404. The nav listing from `GET /lookml_models` advertises explores that `GET /lookml_models/:model/explores/:explore` refuses to serve: abstract explores (`extension: required`), explores in a model that no longer compiles, and explores the credentials can't see. Looker returns an identical, non-diagnostic 404 for all of these, and `LookmlModelNavExplore` carries no field to filter on beforehand, so the only option is to handle the 404 when it happens.

`snapshot_child_resources` has the same gap. `required_can_permission` is a pre-filter heuristic and doesn't prevent all possible reasons for a 404, like a parent deleted between listing it and requesting its children.

Both now log a warning identifying the skipped resource and continue.

Note: Looker also serves an HTML "unavailable" page under a 404 during maintenance, so only its JSON error envelope that contains the string "documentation_url" counts as a missing resource.

**Checklist:**

### DIFF
--- a/source-looker/CHANGELOG.md
+++ b/source-looker/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+
+## 2026-07-26
+
+### Fixed
+- The `lookml_model_explores` binding no longer fails when a LookML model advertises an
+  explore that Looker won't serve, such as an abstract explore or one in a model that no
+  longer compiles. These explores are now skipped with a warning.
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
LGTM

PR #4901, opened 2026-07-27
docs: agent context map and issue-tracker conventions

mdibaiee:
Documents the repo for agent consumption, in two parts.

## `CONTEXT-MAP.md` — captures vs materializations

The repo holds two domains that share a build system and little else. Splitting them means an agent working a task reads the relevant slice of ~100 connector directories instead of guessing across all of them.

The boundary is a real one, not filing convenience: the two sides implement different protocols, have different failure modes, and overload the same words — *binding*, *checkpoint*, *backfill* — with different meanings. It also happens to be a clean language cut, with Rust the only thing straddling it (`source-kafka`, `source-http-ingest`, `bson-transcoder` vs `materialize-kafka`).

Per-context glossaries are *reserved* under `docs/contexts/` but deliberately **not written yet**. They get seeded when a term actually needs resolving; authoring vocabulary up front is how docs rot. `docs/agents/domain.md` tells agents to proceed silently when they're absent rather than nag about it.

ADR scoping follows the same split: `docs/adr/` for decisions that must hold on both sides of the protocol, `docs/contexts/<context>/adr/` for one-sided ones.

## `docs/agents/` — conventions

- **`issue-tracker.md`** — issues live in GitHub Issues via `gh`, with a note about the org's SAML enforcement (worth having written down; it silently 403s every API call until the token is authorized) and a convention to prefix issue titles with the connector name so the queue stays readable.
- **`domain.md`** — how to consume the context map and scope ADRs.

`CLAUDE.md` gains a short `## Agent skills` section pointing at both.

Issue **labelling is deliberately not covered** — we haven't decided on a vocabulary, and writing one down before agreeing on it just creates a doc to unlearn.

## Note on overlap

`CLAUDE.md` already has `## Layout` and `## Development Paths` sections listing some of the same paths as `CONTEXT-MAP.md`. That duplication is real and left alone here — deduping it is a judgement call about which file should own the path inventory, and worth deciding separately from landing this.


docs/agents/domain.md
```diff
@@ -0,0 +1,50 @@
...
+
+How the engineering skills should consume this repo's domain documentation when exploring the codebase.
+
+This repo is **multi-context**: captures and materializations are separate domains with separate vocabulary, separate protocols, and separate failure modes. A term like "binding", "checkpoint", or "backfill" does not mean the same thing on both sides.
+
+## Before exploring, read these
+
+- **`CONTEXT-MAP.md`** at the repo root — it names each context and points at that context's `CONTEXT.md`. Read the `CONTEXT.md` for every context your topic touches, and only those.
+- **`docs/adr/`** — system-wide decisions (things spanning both captures and materializations, or the repo as a whole).
+- **`docs/contexts/<context>/adr/`** — decisions scoped to one context. Read the ADRs that touch the area you're about to work in.
```
Alex-Bair, line 11: Are we committing to documenting architectural decisions with ADRs? If there's been discussion & alignment on using ADRs, I missed that. I assume we'd want to talk about the ADR process first with the teams and probably provide a template, an example of what a good ADR looks like. and guidelines for what warrants an ADR?

Alex-Bair (approved):
LGTM % a question about the ADR part. Did I miss a conversation somewhere that discussed implementing them?

docs/agents/domain.md
mdibaiee: @Alex-Bair these were set up by [this skill](https://github.com/mattpocock/skills) that Johnny mentioned earlier.

I don't think of this as any sort of committment, but I think the skills have tools for writing ADR documents with the help of Claude if necessary, and it provides the template and the structure as part of the skills.

I am mostly interested in the new contexts that we can develop for materializations and captures to help us in our development, and the `grilling` skill that Johnny mentioned, and based on my first trial it has been pretty good.

Alex-Bair: Got it, cool, sounds good! I haven't played around with those skills yet, but they sound promising.

PR #4903, opened 2026-07-27
source-intercom-native: ignore 404s when fetching child resources

Alex-Bair:
**Regression?**

No.

**Description:**

The `company_segments` stream only tolerated a 404 from `/companies/{id}/segments` when the response body contained 'Company Not Found', but Intercom sometimes returns a bare nginx 404 page for a deleted company instead & caused the capture to fail. This PR fixes this by treating any 404 from that endpoint as a deleted company and skip its segments.

The `conversation_parts` stream races the same way - a conversation can be deleted between when the search endpoint returns it and when its parts are requested - so I also applied the same handling to `/conversations/{id}`.

**Checklist:**

### DIFF
--- a/source-intercom-native/source_intercom_native/api.py
+++ b/source-intercom-native/source_intercom_native/api.py
@@ -436,9 +436,25 @@ async def _fetch_parts(
     async with parts_semaphore:
         url = f"{API}/conversations/{conversation_id}"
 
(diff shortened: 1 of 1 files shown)


nicolaslazo (approved):
LGTM

PR #4904, opened 2026-07-27
materialize-snowflake: list TIMESTAMP_NTZ (normalize to UTC) first

mdibaiee:
Reorders the `timestamp_type` enum in the Snowflake materialization config so that `TIMESTAMP_NTZ (normalize to UTC)` is the first option presented in the UI.

`TIMESTAMP_NTZ` is Snowflake's default `TIMESTAMP_TYPE_MAPPING`, so "normalize to UTC" is the correct choice for most accounts. The other NTZ variant exists mainly to preserve the behavior of existing materializations, and previously sat above it purely because the list was alphabetical.

Presentation only — the stored config values are unchanged, enum order has no effect on validation, and no descriptions were touched. The only other change is the regenerated `TestSpecification` snapshot.

## Testing

`go build`, `go vet`, and `go test ./materialize-snowflake/` (unit + snapshot) all pass.


jwhartley (approved):
LGTM

PR #4908, opened 2026-07-27
source-braintree-native: configurable concurrency per binding

Alex-Bair:
**Regression?**

No.

**Description:**

The connector fanned out search page/batch fetches at a fixed concurrency of 20 with no way to tune it. We observed that for a high-volume merchant, this drove a large number of concurrent, long-running (>50s) `disputes/advanced_search` requests. `disputes` must offset-paginate the full search on every page since there's no ids endpoint, and the window can't go below 24h, so these searches are costly on Braintree's side. Braintree's SRE flagged the resulting volume of long-running requests as harmful to endpoint availability.

This PR addresses this by exposing fetch concurrency as a per-binding resource config field so we can dial back the expensive streams without slowing the cheaper ones. It defaults to 5 for `disputes`, 20 for the other search-based streams, and 1 for full-refresh snapshots since they don't fetch concurrently. The field is optional with a default so existing bindings keep working and stay at their current concurrency until re-discovered or set explicitly.

**Checklist:**

### DIFF
--- a/source-braintree-native/source_braintree_native/models.py
+++ b/source-braintree-native/source_braintree_native/models.py
@@ -71,6 +71,25 @@ class EndpointConfig(BaseModel):
         json_schema_extra={"advanced": True},
     )
 
+# Default maximum number of concurrent requests used to fetch a stream. Disputes default
+# lower than other streams: each disputes advanced_search re-runs the full search server-side
+# (there is no ids endpoint to page over cheaply), and Braintree has flagged a high volume of
+# these long-running requests as harmful to endpoint availability. Full-refresh snapshots don't
+# fetch pages concurrently, so they use 1.
+DEFAULT_CONCURRENCY = 20
+DEFAULT_DISPUTES_CONCURRENCY = 5
+DEFAULT_SNAPSHOT_CONCURRENCY = 1
+
+
+class ResourceConfig(ResourceConfigWithSchedule):
+    concurrency: Annotated[int, Field(
+        title="Concurrency",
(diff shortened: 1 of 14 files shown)


source-braintree-native/source_braintree_native/models.py
```diff
@@ -71,6 +71,25 @@ class Advanced(BaseModel):
         json_schema_extra={"advanced": True},
     )
 
+# Default maximum number of concurrent requests used to fetch a stream. Disputes default
+# lower than other streams: each disputes advanced_search re-runs the full search server-side
+# (there is no ids endpoint to page over cheaply), and Braintree has flagged a high volume of
+# these long-running requests as harmful to endpoint availability. Full-refresh snapshots don't
+# fetch pages concurrently, so they use 1.
```
aeluce, line 78: Out of curiosity, what happens if someone manually bumps up a snapshot stream's concurrency?

Alex-Bair: Nothing happens. All streams share the same resource config shape, so even though snapshot streams don't use the `concurrency` field at all, they have to have it since the incremental streams do use it.

Ideally, there wouldn't even be a `concurrency` field present on resource configs for snapshot streams. But I _think_ various component (protocol, UI, etc.) have been built assuming there's a single JSON schema to represent all resource configs per connector, and I haven't found a way to finagle the JSON schema in a way to support different resource configs that also renders nicely in the UI.

nicolaslazo (approved):
LGTM!

PR #4935, opened 2026-07-28
source-asana: fix typo for portfolios_memberships query parameter

Alex-Bair:
**Regression?**

No.

**Description:**

The key used for the `stream_slices` lookup in `PortfolioMemberships.request_params` needs to be the same as the `slice_field` defined 7 lines above it. They weren't aligned, so the lookup has been failing & crashing the connector. This PR fixes that typo.

**Checklist:**


midianyoseph (approved):
LGTM

PR #4941, opened 2026-07-29
Mark nonsensitive config fields across connectors

mdibaiee:
**Description:**

Adds the `nonsensitive: true` JSON-schema annotation to endpoint and resource config locations that are safe to modify outside the SOPS MAC.

Flow's runtime uses this annotation to decide whether a config's plaintext `sops.overlay` may write a location ([`crates/unseal/src/overlay.rs`](https://github.com/estuary/flow/blob/master/crates/unseal/src/overlay.rs)), which lets non-secret fields be changed without decrypting and re-encrypting the config. It also gives us one place to express "this config value is safe to expose", intended to replace the hand-maintained `scripts/obfuscation/config_allowlist.txt`.

The annotation is the sole protection for overlay writes, so a location earns it only if it satisfies **both**:

- **Integrity** — rewriting it cannot let someone unable to forge the SOPS MAC obtain or redirect a protected secret. Excludes everything used to establish a connection (hosts, endpoints, URIs, accounts, buckets, ARNs and roles, tunnels) and anything weakening how a credential is transmitted or authenticated (`sslmode`, `tls`, SASL `mechanism`).
- **Confidentiality** — it reveals nothing about the customer. Excludes databases, schemas, tables, datasets, namespaces, warehouses, project IDs, topics, prefixes and paths, plus customer-authored text such as `additional_table_create_sql` and the dbt trigger `cause`.

What remains is behaviour, format, scheduling, sizing and retention knobs, plus variant discriminators, which are `const`-constrained to their branch and expose nothing within it.

**Workflow steps:**

No user-facing change. The annotation only takes effect when a config carries a `sops.overlay`, and no behaviour depends on it yet.

None.

**Notes for reviewers:**

- `OneOfSchema` builds its discriminator property itself rather than reading the Go field, so the annotation is applied there. That covers `auth_type`, `catalog_type`, `compute_type` and `stagingBucketType` for **every** connector, captures included — please sanity-check that blanket call.
- the file parser options (`format`, `encoding`, `delimiter`, `quote`, `escape`, `defaultOffset`, `contentType`) come from `crates/parser` in flow repository and cannot be annotated from this repo.

**Checklist:**


filesink/store_s3.go
```diff
@@ -38,7 +38,7 @@ type AccessKeyCredentials struct {
 }
 
 type CredentialsConfig struct {
-	AuthType AuthType `json:"auth_type"`
+	AuthType AuthType `json:"auth_type" jsonschema_extras:"nonsensitive=true"`
```
willdonnelly: Seeing anything at all related to authentication and credentials marked as nonsensitive feels sketchy even if it's just the type field. Is this motivated by one of the config migrations you're hoping to run? If not I'd recommend erring on the side of not doing this, but you might have more context on this one.

willdonnelly (commented):
Thanks for tackling this! Overall this looks like a great set of conservative annotations.

Things I would recommend removing `nonsensitive` from:
- In `source-sqlserver` the `change_table_cleanup` and `capture_instance_management` options are _probably_ not security relevant, but I'm not 100% confident saying that there's no security implications there and TBH even if there aren't those are very "weighty" options which we might want to require credentials for just to make it harder to turn them on by accident.

Things to consider adding in the future. None of this is necessarily asking you to do it now, the review box was just the most convenient place to put these notes as I read through the diffs:
- Source Tag properties are probably nonsensitive.
- Default timezones added to timezone-less datetime columns to make RFC3339 timestamps are probably nonsensitive.
- Discovery options like schema whitelist, table names, etc are probably nonsensitive. In a very abstract sense these could be called "security relevant" but in the specific context of Estuary I don't think they are because in general a user can modify the _bindings_ of a capture arbitrarily without running afoul of SOPS validation or having to re-enter their credentials, and discovery options like that are basically just a shorthand for having us discover and use specific bindings automatically.

filesink/store_s3.go
mdibaiee: I initially thought it would be useful for migrating from old credentials format to new `credentials`, but in those cases we must encrypt for the credentials themselves anyway, so it is no use. I removed it 👍🏽

danielnelson (dismissed):

jacobmarble (dismissed):

Alex-Bair (approved):
LGTM, thank you!

willdonnelly:
I would like to register some concern about this PR now that I've looked more closely. Previously I skipped the commit message and just looked at the code changes and thought that it seemed like a reasonable set of annotations, however:

> It also gives us one place to express "this config value is safe to expose", intended to replace the hand-maintained scripts/obfuscation/config_allowlist.txt.
> The annotation is the sole protection for overlay writes, so a location earns it only if it satisfies both:
>
>    Integrity — rewriting it cannot let someone unable to forge the SOPS MAC obtain or redirect a protected secret. Excludes everything used to establish a connection (hosts, endpoints, URIs, accounts, buckets, ARNs and roles, tunnels) and anything weakening how a credential is transmitted or authenticated (sslmode, tls, SASL mechanism).
>    Confidentiality — it reveals nothing about the customer. Excludes databases, schemas, tables, datasets, namespaces, warehouses, project IDs, topics, prefixes and paths, plus customer-authored text such as additional_table_create_sql and the dbt trigger cause.

I disagree with all of this. The concept of whether a config value is safe to expose is not the same thing as whether it's safe to allow overriding them via config overlays, and we should not conflate them in this way.

Relatedly, I previously shrugged about but didn't push back on the addition of `nonsensitive` annotations on resource config schemas since it didn't really make any difference, but strictly speaking they make no sense in the context of config overrides because binding configuration isn't subject to SOPS MAC validation in the first place. I really think you need to introduce another annotation to express the semantics you desire here. **_(edit: Never mind, I just checked again and I see that the resource config changes were already removed)_**

mdibaiee:
@willdonnelly that PR description is out of date, the obfuscation scripts have not changed based on this yet, and I agree with your sentiment in general

willdonnelly:
Okay, that's good to hear. Like I said, the code changes themselves seemed reasonable when I reviewed it, I just happened to reread the description yesterday when the config-encryption issues came up and was concerned about the conflation of the two concepts.

PR #4954, opened 2026-07-30
source-sqlserver{,ct}: bind datetime backfill resume keys as datetime parameters

Alex-Bair:
**Regression?**

No

**Description:**

Backfill chunks resume with `WHERE (k1 > @p1) OR (k1 = @p1 AND k2 > @p2)` when there's more than a single key column. Datetime resume keys were bound as strings - `EncodeKeyFDB` formats them with `time.Format` so they sort correctly as FDB tuples - leaving SQL Server comparing a datetime column against an `nvarchar` parameter. That implicit conversion step combined with the disjunctive predicate stopped the optimizer from deriving a seek range, so every chunk scanned from the start of the index and re-read everything that had already been backfilled.

This PR fixes this by parsing the value back into a `time.Time` before binding it so the comparison is datetime-to-datetime. This lets the optimizer decide to seek instead of scan. I tested this out locally, and on a 4M-row table with a two-column key, the chunk at row 3.9M went from reading 3,950,000 rows to 50,000 rows (which was the chunk size).

I also added a few more test cases for `testScanKeyTypes` in `sqlserver/tests/datatypes.go` so we have snapshots covering the various datetime types that someone might use as table keys.

**Checklist:**

### DIFF
--- a/go/capture/sqlserver/tests/datatypes.go
+++ b/go/capture/sqlserver/tests/datatypes.go
@@ -23,6 +23,7 @@ func TestDatatypes(t *testing.T, setup testSetupFunc) {
 	t.Run("MiscTypes", func(t *testing.T) { testMiscTypes(t, setup) })
 	t.Run("NotNullTypes", func(t *testing.T) { testNotNullTypes(t, setup) })
 	t.Run("ScanKeyTypes", func(t *testing.T) { testScanKeyTypes(t, setup) })
+	t.Run("CompositeScanKey", func(t *testing.T) { testCompositeScanKey(t, setup) })
 	t.Run("OversizedFields", func(t *testing.T) { testOversizedFields(t, setup) })
 	t.Run("BitNotNullDeletion", func(t *testing.T) { testBitNotNullDeletion(t, setup) })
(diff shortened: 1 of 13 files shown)


go/capture/sqlserver/tests/datatypes.go
```diff
@@ -248,17 +248,18 @@ func testScanKeyTypes(t *testing.T, setup testSetupFunc) {
 	for _, testCase := range []struct {
 		Name       string
 		ColumnType string
+		SecondKey  string   // When set, the table gets a two-column primary key (k, k2)
```
willdonnelly: I'm not a fan of making this test more complicated than it already is and would recommend just writing a new test exercising the composite-key-where-one-is-smalldatetime scenario.

willdonnelly (dismissed):
LGTM % one nit

go/capture/sqlserver/tests/datatypes.go
Alex-Bair: Got it, I'll do that!

willdonnelly (approved):
LGTM, thanks!

PR #4955, opened 2026-07-30
source-sqlserver{,ct}: discovery query cache use improvements

### LINKED ISSUE #4946: source-sqlserver: chunked discovery predicate defeats the plan cache, exhausting compile memory at high database counts
## 1. Priority

High - ongoing. A customer's production SQL Server instance is being driven to ~99% CPU for ~10 minutes at a time by connector-issued discovery queries, with blocking chains 386 sessions deep and individual waits over 5 minutes. Most recent episode 2026-07-29 ~15:13-15:27 US Eastern. The captures are running, so this is active impact on the customer's own production database rather than a stalled pipeline. The only mitigation confirmed to stop it is disabling the connector's SQL login, which takes all ~1,350 captures down. No config flag gates the affected code path.

## 2. Scope and prevalence

Theoretical blast radius: every `source-sqlserver`, `source-sqlserver-ct`, and `source-postgres` capture. The chunked/filtered discovery change landed for SQL Server in `ade7654` and for Postgres in `0f23a7bc`, both merged 2026-07-21..23. Compile memory (`RESOURCE_SEMAPHORE_QUERY_COMPILE`) is an instance-wide gateway, so any deployment with many captures against databases on one shared server is exposed.

Observed prevalence: one tenant, first occurrence, recurring since 2026-07-27 and still reproducing. ~1,350 single-tenant databases, ~168 bindings each, one SQL Server instance. Not yet observed on Postgres.

The trigger scales with databases-per-instance, not table count. Below the compile-memory threshold the symptom is invisible; above it the instance saturates. Priority is High on severity and absence of workaround, not on breadth.

## 3. Trigger (testable)

N captures, each with enough bindings that `DiscoverTableDetails` chunks (>100 tables), all targeting databases on one SQL Server instance. Every capture startup plus every `rediscoverInterval` (15 min) issues five discovery queries per chunk, each carrying up to 200 parameters in an OR-chain. Query text varies by chunk size and parameter types vary by table-name length, so no two are plan-cache-equivalent. Compile memory saturates, sessions queue on `RESOURCE_SEMAPHORE_QUERY_COMPILE`, and every other session needing the same compile queues behind an exclusive `OBJECT.COMPILE` lock.

Minimal repro input (synthetic, NO customer data): a capture with 101+ bindings across databases whose table names differ in length, e.g.

```
dbo.a           -> @pN nvarchar(1)
dbo.abcdefghij  -> @pN nvarchar(10)
```

Two databases with the same logical schema but different table-name lengths produce different parameter signatures for the same logical query, and therefore separate plan cache entries. Repeat across enough databases to exhaust compile memory.

## 4. Root cause + confidence

**Confirmed** (code ref + lock evidence) for the mechanism. `ade7654` ("source-{sqlserver,sqlserver-ct}: Chunked discovery queries") replaced five static, parameterless discovery queries with `fmt.Sprintf`-built queries carrying a `tableIDsPredicate` OR-chain.

Before (`source-sqlserver/discovery.go`, pre-`ade7654`) - constant string, one cached plan per database:

```sql
SELECT KCU.TABLE_SCHEMA, KCU.TABLE_NAME, KCU.COLUMN_NAME, KCU.ORDINAL_POSITION
  FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE KCU
  JOIN INFORMATION_SCHEMA.TABLE_CONSTRAINTS TCS ...
  WHERE TCS.CONSTRAINT_TYPE = 'PRIMARY KEY'
```

After:

```go
const discoveryChunkSize = 100
func tableIDsPredicate(schemaCol, tableCol string, tables []sqlcapture.TableID, startArg int) (string, []any)
func queryDiscoverPrimaryKeys(opts discoveryOptions, predicate string) string
```

**Confirmed by measurement** for the plan-cache attribution. go-mssqldb sizes each `nvarchar` parameter to the actual string length, so the parameter signature is a function of table-name lengths. Observed in captured `sql_text`: `(@p1 nvarchar(3),@p2 nvarchar(23),...)` on one session and `(@p1 nvarchar(3),@p2 nvarchar(31),...)` on another, same logical query.

`sys.dm_exec_query_stats` grouped by compile minute shows `plans_compiled` equal to `total_executions` on every row for the discovery query family - every plan compiled, executed once, never reused:

```
query_family  compile_minute           plans_compiled  total_executions  total_cpu_ms
discovery     2026-07-29 16:57:00.000  79              79                1151
cdc_poll      2026-07-29 16:56:00.000  3               3                 41
discovery     2026-07-29 16:56:00.000  73              73                15287
discovery     2026-07-29 16:55:00.000  3               3                 38
```

Sampled 16:55-16:57 US Eastern, i.e. steady state, well after the 15:13-15:27 episode. Cache reuse is zero even when no storm is in progress. Because every plan is single-use it is also a prime eviction candidate, so these counts are a floor rather than a total.

Note `total_worker_time` is execution CPU, not compile CPU - `dm_exec_query_stats` has no compile-time column, so compile cost is additional and unmeasured here. At 16:56 the discovery family burned 15,287 ms of execution CPU against 41 ms for `cdc_poll`, roughly 209 ms per execution. The OR-chain predicate is therefore expensive to *run* as well as to compile, which argues for the table-valued-parameter rewrite over parameter padding alone. Per-execution cost varies widely across the sampled minutes (14.6 ms at 16:57 vs 209 ms at 16:56), presumably because the five discovery queries differ in cost and chunk sizes vary, so do not over-read a single minute.

### Contributing factors (not the root cause, but they set the conditions)

- `sqlcapture/capture.go`: `rediscoverInterval = 15 * time.Minute`, hardcoded, ungated by any feature flag. The full five-query suite re-runs unconditionally with no change-detection pre-check. Cost scales with bindings x captures; value scales with rate of schema change.
- `rediscoverAfter` is a zero-valued `time.Time`, so the loop's first iteration fires immediately after the `Run()`-entry call. **Every capture startup performs two full discovery passes back to back.** Relevant if captures restart-loop under the storm.
- `emit_sourced_schemas` does not gate the queries - `DiscoverTableDetails` runs before the flag is consulted (`discovery.go:194`), so disabling it does not reduce load.
- The affected spec sets `polling_interval: 600s` (default 500ms). See section 5: this produces a separate, continuous ~10 minute CPU oscillation independent of the compile storm.

Customer-side via `sp_WhoIsActive @get_plans = 1, @get_locks = 1, @find_block_leaders = 1`. Raw `collection_time` `2026-07-29 14:23:45.227` (SQL Server local, US Central = 15:23:45 US Eastern). 665 well-formed sessions across 554 databases.

Lock resource types across the 479 sessions carrying lock data:

```
419  OBJECT.COMPILE       mode=X      status=WAIT
419  DATABASE.PLANGUIDE   mode=S      status=GRANT
 42  KEY                  mode=X      status=GRANT
 40  OBJECT               mode=IX     status=GRANT
 78  PAGE                 mode=IX/IU  status=GRANT
  1  METADATA.SCHEMA      mode=Sch-S  status=GRANT
```

The `KEY`/`OBJECT`/`PAGE` locks belong to the customer application. **No connector session holds a data lock.**

Lock XML, root block leader session 749:

```xml
<Lock resource_type="OBJECT.COMPILE" request_mode="X" request_status="WAIT" request_count="1" />
```

Root block leaders, all on compile memory, all `writes = 0` and `open_tran_count = 0`:

```
749   RESOURCE_SEMAPHORE_QUERY_COMPILE 378ms   blocking 386   elapsed 00:05:20
1297  RESOURCE_SEMAPHORE_QUERY_COMPILE 107ms   blocking  62   elapsed 00:00:23
672   RESOURCE_SEMAPHORE_QUERY_COMPILE  99ms   blocking   2   elapsed 00:00:01
319   RESOURCE_SEMAPHORE_QUERY_COMPILE 149ms   blocking   2   elapsed 00:00:01
```

Session 749's statement is the discovery query with **200 parameters** (100 tables = `discoveryChunkSize`). The pre-`ade7654` code cannot emit a parameterized discovery query at all, so the deployed build is post-change. The spec pins the floating tag `ghcr.io/estuary/source-sqlserver:v0`.

Also in that snapshot: 449 sessions waiting `LCK_M_X`, 24 on `RESOURCE_SEMAPHORE_QUERY_COMPILE`, 517 `Flow CDC Connector` sessions. All connector sessions `writes = 0`. Total connector CPU ~11.4 CPU-seconds - the cost is compilation, not data movement.

**Caveat on session 749.** Its elapsed time at collection was 5m20s, so it started ~15:18 ET, but the CPU plateau began ~15:13 ET. 749 was not the originator; it is the deepest-rooted session still present when the snapshot was taken. Whatever first exhausted compile memory had already finished or died. "Root block leader" in a point-in-time snapshot means head-of-chain now, not cause.

**Negative control (customer-run).** `capture_instance_management` was set to `false` at 2026-07-29 10:55 US Eastern. The snapshot above is from 4.5 hours later with the identical signature. The customer independently reports that every session involved is running the `INFORMATION_SCHEMA.KEY_COLUMN_USAGE` discovery query. Capture-instance management is ruled out by experiment.

**Plan cache measurement.** Run against the affected instance, 16:55-16:57 US Eastern:

```sql
SELECT
    CASE
        WHEN st.text LIKE '%KEY_COLUMN_USAGE%'
          OR st.text LIKE '%is_computed = 1%'
          OR st.text LIKE '%idx.is_unique = 1%' THEN 'discovery'
        WHEN st.text LIKE '%fn_cdc_get_all_changes%' THEN 'cdc_poll'
        ELSE 'other'
    END AS query_family,
    DATEADD(MINUTE, DATEDIFF(MINUTE, 0, qs.creation_time), 0) AS compile_minute,
    COUNT(*)                       AS plans_compiled,
    SUM(qs.execution_count)        AS total_executions,
    SUM(qs.total_worker_time)/1000 AS total_cpu_ms
FROM sys.dm_exec_query_stats qs
CROSS APPLY sys.dm_exec_sql_text(qs.sql_handle) st
WHERE st.text LIKE '%KEY_COLUMN_USAGE%'
   OR st.text LIKE '%is_computed = 1%'
   OR st.text LIKE '%idx.is_unique = 1%'
   OR st.text LIKE '%fn_cdc_get_all_changes%'
GROUP BY ..., DATEDIFF(MINUTE, 0, qs.creation_time)
ORDER BY compile_minute DESC, query_family;
```

Result is in section 4. `plans_compiled = total_executions` on every row.

**CPU chart, 2026-07-29 14:50-16:50 ET (customer-supplied, Azure `Percentage CPU (Avg)`).**

- One episode: ~99% plateau from ~15:13 to ~15:27 ET.
- Steady state outside the episode: 30-70% oscillation with peaks at approximately :02, :13, :24, :34, :44 - **a ~10 minute period**, matching `polling_interval: 600s`.
- Peak phase is **not wall-clock aligned** (not :00/:10/:20). Connector timers are relative to each capture's start; SQL Agent jobs and the customer's own 5-minute CDC capture controller would be wall-clock anchored. This rules the customer's orchestration layer out as the driver of the 10 minute beat.
- The plateau begins on a 10 minute beat (~15:13), i.e. it looks like a poll wave that failed to drain within its cycle rather than an unrelated event.

Period values are read off the chart image to +/-1 minute and the series contains gaps (dashed segments). Export the underlying series before relying on exact values.

Connector source read at `84a154a` to establish the write surface: `read_only` defaults true (no watermark writes), `change_table_cleanup` false in the spec, no `sp_cdc_enable_table`/`disable_table` in any snapshot. Also ruled out: data-lock contention on `cdc.*` or on any shared application table, since no connector session holds a data lock in either snapshot.

1. Customer-immediate stopgap (ops, human-run, no code change): disabling the connector's SQL login or pausing the captures stops it, at the cost of taking ~1,350 captures down. `capture_instance_management: false` is worth keeping (removes a 30s `cdc.ddl_history` / `cdc.change_tables` ticker) but demonstrably does not stop this.
2. Data-preserving options, no code change: lower `discoveryChunkSize` (smaller predicates compile faster even without cache reuse). Separately, reconsider `polling_interval: 600s` - counterintuitively, a *shorter* interval should reduce peak CPU by smoothing 10-minute burst scans into a trickle, and it addresses the 30-70% baseline rather than the episodes.
3. Platform / code direction:
   - Replace the `tableIDsPredicate` OR-chain with a table-valued parameter or a `VALUES` join, so both query text and plan stabilise regardless of chunk size. Preferred: the measurement in section 4 shows the predicate is expensive to execute (~209 ms per run at 16:56), not only to compile, so this addresses both cost channels.
   - Alternatively, pad discovery parameters to a fixed width (`nvarchar(128)`, the `sysname` bound) so the parameter signature stops varying with table-name length. Smaller change and should restore cache reuse, but leaves the per-execution cost of the 100-way OR in place.
   - Add a change-detection pre-check before the five-query suite (e.g. `MAX(modify_date)` over `sys.objects`/`sys.columns`) and short-circuit when unchanged. Preserves pending-stream and `TableStateMissing` recovery behaviour exactly.
   - Make `rediscoverInterval` configurable, or scale it by binding count.
   - Initialise `rediscoverAfter = time.Now().Add(rediscoverInterval)` so startup does one discovery pass rather than two.

Any code fix is drafted for human review - not self-merged or published, changes affect live tenants.

## 7. Definition of done

- Acceptance: with N captures against many databases on one instance, `sys.dm_os_wait_stats` shows no sustained `RESOURCE_SEMAPHORE_QUERY_COMPILE` accumulation and `sys.dm_tran_locks` shows no `OBJECT.COMPILE` waits attributable to connector sessions. The section 5 plan-cache query shows `total_executions` substantially greater than `plans_compiled` for the discovery family, against the measured pre-fix baseline of exactly 1:1 (section 4), with a bounded number of distinct cached entries per database.
- Regression test: two databases with identical logical schemas but table names of differing lengths must produce byte-identical discovery query text and identical parameter type signatures.
- Constraints / do not break: keep the `ade7654` benefit (discovery of very large catalogs must not time out or exhaust memory); do not reintroduce unbounded single-query discovery; pending-binding activation and `TableStateMissing` recovery must continue to work without a capture restart; no change to discovered schema output; no forced backfill or re-version of existing captures.

## 8. References

- Slack thread: `<#integrations permalink - to be backfilled>`
- Regression commit: `ade765443d3e05fc841eed0ea681d7367bb7d796` (source-{sqlserver,sqlserver-ct}: Chunked discovery queries)
- Same change for Postgres: `0f23a7bc5a5610c5d60845e34dac204932fa019f` (source-postgres: Chunked discovery queries)
- Parent refactor: #4481
- PR #2727 (sqlcapture: Add `emit_sourced_schemas` feature flag) - notes SourcedSchema rides on preexisting periodic rediscovery

Alex-Bair:
**Regression?**

No.

**Description:**

After rolling out the chunked discovery query behavior as part of https://github.com/estuary/connectors/pull/4872, a user running 1,000+ captures against single-tenant databases on one SQL Server instance saw it driven to sustained ~99% CPU with blocking chains hundreds of sessions deep. They identified that the server was spending all of its CPU compiling query plans for our discovery queries.

Chunked discovery (ade7654) filters each query by a disjunction of (schema, table) equalities. The statement text is identical across runs, but sp_executesql keys its cached plan on the statement text plus the parameter declaration, and `go-mssqldb` sizes every string parameter to the length of the value it carries. So the declaration is a function of the customer's table names, and could be something like:
```
  (@p1 nvarchar(3),@p2 nvarchar(27),@p3 nvarchar(3),@p4 nvarchar(16),...)
```
Compounding it, `bindingTableIDs` iterates a map, so the list is reshuffled on every rediscovery pass. The net effect was there was essentially one compiled plan per execution and none were ever reused.

This PR improves query plan cache re-use for chunked discovery queries by:
- `sqlcapture`: sorting `bindingTableIDs` output.
- Binding identifiers as `mssqldb.NVarCharMax` in our chunked discovery queries. This makes the declaration constant regardless of name length.
- Padding the trailing chunk. A list that doesn't divide evenly into `discoveryChunkSize` ends with a short chunk whose text has fewer terms, so it caches its own plan. Repeating the final table makes every chunk's text identical - X OR X is X, so results are unchanged. This padding is only applied when the list actually spans more than one chunk.

**Notes for reviewers:**

See this [Slack thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1785361977655109) for more context around the escalation that motivated this change.

**Checklist:**

### DIFF
--- a/source-sqlserver-ct/CHANGELOG.md
+++ b/source-sqlserver-ct/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## 2026-07-30
 
+### Fixed
+- Discovery queries are now shaped so that SQL Server reuses a single cached
+  query plan across repeated discovery passes, rather than compiling a new plan
+  for every chunk of tables every time. On instances hosting many databases the
+  previous behavior could exhaust query compile memory and drive the server to
+  sustained high CPU, blocking unrelated queries.
+
 ### Changed
 - Backfill resume keys for `DATETIME2`, `SMALLDATETIME`, `DATETIMEOFFSET` and `DATE`
   primary-key columns are now bound as typed datetime parameters rather than as
--- a/source-sqlserver-ct/discovery.go
+++ b/source-sqlserver-ct/discovery.go
@@ -11,6 +11,7 @@ import (
 	"github.com/estuary/connectors/go/tableglob"
 	"github.com/estuary/connectors/sqlcapture"
 	"github.com/invopop/jsonschema"
+	mssqldb "github.com/microsoft/go-mssqldb"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -97,12 +98,23 @@ func matchesAnyPattern(patterns []*tableglob.Pattern, t sqlcapture.TableID) bool
 // This disjunction-of-equalities form plans as per-table index seeks on the
 // system catalogs, but only so long as discoveryChunkSize is kept smallish.
 // Currently we use 100 and that's probably in the sweet spot.
+//
+// Arguments are bound as nvarchar(max) rather than as plain Go strings, so that
+// the parameter type signature stays identical regardless of how long the
+// identifiers are. go-mssqldb otherwise sizes each string parameter to the value
+// it happens to carry (makeStrParam sets ti.Size = len(res.buffer)), and that
+// declaration forms part of SQL Server's plan cache key, so binding them plainly
+// would declare as '(@p1 nvarchar(3), @p2 nvarchar(23), ...)' and every distinct
+// combination of identifier lengths would compile a plan of its own. Since each
+// of those plans is only ever used once it would also be a prime eviction
+// candidate. A constant declaration lets one cached plan serve every chunk
+// holding the same number of tables.
 func tableIDsPredicate(schemaCol, tableCol string, tables []sqlcapture.TableID, startArg int) (string, []any) {
 	var terms = make([]string, len(tables))
 	var args = make([]any, 0, len(tables)*2)
 	for i, t := range tables {
 		terms[i] = fmt.Sprintf("(%s = @p%d AND %s = @p%d)", schemaCol, startArg+2*i, tableCol, startArg+2*i+1)
-		args = append(args, t.Schema, t.Table)
+		args = append(args, mssqldb.NVarCharMax(t.Schema), mssqldb.NVarCharMax(t.Table))
 	}
 	return "(" + strings.Join(terms, " OR ") + ")", args
 }
--- a/source-sqlserver-ct/discovery.go
+++ b/source-sqlserver-ct/discovery.go
@@ -85,6 +85,30 @@ func matchesAnyPattern(patterns []*tableglob.Pattern, t sqlcapture.TableID) bool
 	return false
 }
 
+// padChunk repeats the final table until the chunk holds exactly
+// discoveryChunkSize entries, so that a trailing partial chunk generates the
+// same query text as a full one and shares its cached plan.
+//
+// Repeating a table is fine. The predicate is a disjunction of equalities,
+// so the extra terms match rows which already matched and the result set is
+// unchanged.
+//
+// Callers should only do this when the table list actually spans more than one
+// chunk. A capture whose bindings all fit in a single chunk already issues just
+// one query shape, so padding it out would add redundant predicate terms
+// without consolidating anything.
+func padChunk(chunk []sqlcapture.TableID) []sqlcapture.TableID {
+	if len(chunk) == 0 || len(chunk) >= discoveryChunkSize {
+		return chunk
+	}
+	var padded = make([]sqlcapture.TableID, discoveryChunkSize)
+	copy(padded, chunk)
+	for i := len(chunk); i < discoveryChunkSize; i++ {
+		padded[i] = chunk[len(chunk)-1]
+	}
+	return padded
+}
+
 // tableIDsPredicate builds a WHERE-clause predicate matching any of the
 // specified tables, using the given column expressions for the schema and
 // table names. The returned clause has the form
@@ -108,7 +132,7 @@ func matchesAnyPattern(patterns []*tableglob.Pattern, t sqlcapture.TableID) bool
 // combination of identifier lengths would compile a plan of its own. Since each
 // of those plans is only ever used once it would also be a prime eviction
 // candidate. A constant declaration lets one cached plan serve every chunk
-// holding the same number of tables.
+// holding the same number of tables. padChunk handles the trailing partial one.
 func tableIDsPredicate(schemaCol, tableCol string, tables []sqlcapture.TableID, startArg int) (string, []any) {
 	var terms = make([]string, len(tables))
 	var args = make([]any, 0, len(tables)*2)
@@ -141,10 +165,18 @@ func (db *sqlserverDatabase) ListTables(ctx context.Context) ([]sqlcapture.Table
 // about the specified tables. The per-table metadata queries are chunked into
 // batches so that discovery scales to databases with very large catalogs.
 func (db *sqlserverDatabase) DiscoverTableDetails(ctx context.Context, requested []sqlcapture.TableID) (map[sqlcapture.StreamID]*sqlcapture.DiscoveryInfo, error) {
+	// When the list spans several chunks the trailing partial chunk is padded so
+	// that every chunk yields the same query text.
+	var padPartialChunk = len(requested) > discoveryChunkSize
+
 	var tableMap = make(map[sqlcapture.StreamID]*sqlcapture.DiscoveryInfo, len(requested))
 	for i := 0; i < len(requested); i += discoveryChunkSize {
 		var end = min(i+discoveryChunkSize, len(requested))
-		if err := db.extendTableDetails(ctx, tableMap, requested[i:end]); err != nil {
+		var chunk = requested[i:end]
+		if padPartialChunk {
+			chunk = padChunk(chunk)
+		}
+		if err := db.extendTableDetails(ctx, tableMap, chunk); err != nil {
 			return nil, err
 		}
 	}
--- a/source-sqlserver/CHANGELOG.md
+++ b/source-sqlserver/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## 2026-07-30
 
+### Fixed
+- Discovery queries are now shaped so that SQL Server reuses a single cached
+  query plan across repeated discovery passes, rather than compiling a new plan
+  for every chunk of tables every time. On instances hosting many databases the
+  previous behavior could exhaust query compile memory and drive the server to
+  sustained high CPU, blocking unrelated queries.
+
 ### Changed
 - Backfill resume keys for `DATETIME2`, `SMALLDATETIME`, `DATETIMEOFFSET` and `DATE`
   primary-key columns are now bound as typed datetime parameters rather than as
--- a/source-sqlserver/discovery.go
+++ b/source-sqlserver/discovery.go
@@ -86,6 +86,30 @@ func matchesAnyPattern(patterns []*tableglob.Pattern, t sqlcapture.TableID) bool
 	return false
 }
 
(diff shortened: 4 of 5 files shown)


willdonnelly (approved):
LGTM

PR #4957, opened 2026-07-30
source-klaviyo-native: normalize space-separated datetimes to RFC3339

### LINKED ISSUE #4765: source-klaviyo-native: normalize space-separated datetimes to RFC3339
### Problem

Klaviyo formats some datetimes space-separated (`2026-06-23 00:15:23.918411+00:00`) instead of RFC3339. Once schema inference tags a field `format: date-time`, the value violates that contract and materializations break with `cannot parse ... as "T"`. This is the native version of #4726, fixed for legacy `source-klaviyo` in #4742.

Pydantic doesn't save us here: it only normalizes typed datetime fields, but the affected values live in free-form `attributes` under `extra="allow"`, so they pass through verbatim. Known offenders: `Profiles` consent timestamps (`attributes.properties.$consent_timestamp`, `attributes.subscriptions.*.consent_timestamp`), the `events.datetime` cursor (normalized in legacy, not in native), and any other free-form datetime attribute including full-refresh streams like `Metrics`.

  ### Possible Fix

  Add a `@model_validator(mode="before")` on `BaseStream` that recursively rewrites strings matching the observed space-separated shape and leaves everything else alone, reusing the anchored regex from #4742 (`^YYYY-MM-DD HH:MM:SS[.ffffff]+HH:MM$`) so free text like `"May 5, 2021"` isn't mangled.

  Related: #4726, #4742.

nicolaslazo:
**Description:**

Klaviyo emits some datetime fields (consent timestamps, the events datetime cursor, free-form custom properties) space-separated rather than RFC3339 (e.g. `2026-06-23 00:15:23.918411+00:00`). Once schema inference tags such a field `format: date-time`, the value violates the collection's contract and materializations fail with `cannot parse ... as "T"`.

This adds a `mode="before"` model validator on `BaseStream` that recursively rewrites strings matching the anchored space-separated shape, ported from the legacy source-klaviyo fix (#4742) but hooked on the shared base model so full refresh, incremental, and events streams are all covered. Free-text values that merely contain spaces (e.g. `"May 5, 2021"`) are left alone — the anchored regex only matches the full space-separated datetime shape.

Closes #4765.

**Workflow steps:**

No user action required. Existing captures pick up the normalization on the next connector update; affected fields flow through as RFC3339 from then on.

None — no config, binding, or workflow changes; the normalization is internal to document parsing.

**Notes for reviewers:**

Tested with unit tests (`tests/test_models.py`): matching strings at the top level and nested in dicts/lists are rewritten, free-text strings with spaces and near-miss shapes are untouched.

**Checklist:**

### DIFF
--- a/source-klaviyo-native/source_klaviyo_native/models.py
+++ b/source-klaviyo-native/source_klaviyo_native/models.py
@@ -1,3 +1,4 @@
+import re
 from datetime import UTC, datetime, timedelta
 from enum import StrEnum
 from typing import Annotated, Any, ClassVar, Generic, Literal, TypeVar
@@ -22,14 +23,12 @@ def default_start_date():
 
 class ApiKey(AccessToken):
     credentials_title: Literal["API Key"] = Field(
-        default="API Key",
-        json_schema_extra={"type": "string", "order": 0}
+        default="API Key", json_schema_extra={"type": "string", "order": 0}
     )
     access_token: str = Field(
         title="API Key",
         json_schema_extra={"secret": True, "order": 1},
     )
-
 
 
 class EndpointConfig(BaseModel):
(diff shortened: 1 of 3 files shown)


nicolaslazo (commented):
Sorry about the formatting noise — I'll make the `bugfix-capture-connector` workflow commit fixes to preexisting violations as a separate commit. Formatting changes introduce churn in git diffs, polluting LLM contexts when they try to format their own edits.

The referenced skill is deliberately left out of this PR, as it depends on yet unmerged work in #4895

source-klaviyo-native/source_klaviyo_native/models.py
```diff
@@ -84,13 +88,59 @@ class CampaignType(StrEnum):
...
+# RFC3339 contract the value violates, breaking downstream consumers. Rather
+# than enumerate the known offenders (Klaviyo custom properties are free-form,
+# so new ones surface unpredictably), we normalize any string that matches the
+# space-separated shape and leave everything else untouched. The anchored regex
+# is what keeps free-text values (e.g. "May 5, 2021") safe from mangling.
+#
+# Matches the observed Klaviyo format "YYYY-MM-DD HH:MM:SS[.ffffff]+HH:MM".
+_SPACE_SEPARATED_DATETIME_RE = re.compile(
+    r"^(\d{4}-\d{2}-\d{2}) (\d{2}:\d{2}:\d{2}(?:\.\d+)?)([+-]\d{2}:\d{2})\Z"
+)
```
Alex-Bair, line 103: note: I noticed this is very slightly different from the `_SPACE_SEPARATED_DATETIME_RE` in `source-klaviyo`:
https://github.com/estuary/connectors/blob/07c0f3539adff26df877381b855d8be9ac0083a7/source-klaviyo/source_klaviyo/streams.py#L38-L40
The one you landed on with a final `Z` instead of `$` sounds like the one we'd want to use since it doesn't match when there's a trailing `\n`. As a follow up after this PR, could you please align `source-klaviyo`'s `_SPACE_SEPARATED_DATETIME_RE` with the one you landed on here so the behavior between the two connectors are the same?

Alex-Bair (approved):
LGTM, thanks!

PR #4958, opened 2026-07-30
claude: add create-capture-connector skillset

nicolaslazo:
Claude tooling changes split out of #4895, stacked on top of it:

- `claude: make regenerate-flow-discovery into a Haiku agent`
- `claude: add create-capture-connector skillset`
- `claude: child-entities skill: sort drained parent IDs`


CLAUDE.md
Alex-Bair, line 1: It looks like some of the changes Mahdi made in https://github.com/estuary/connectors/pull/4901 got wiped here. Was that intentional? I'm guessing not & we should preserve those changes.

.claude/shared/provider-api-consent.md
```diff
@@ -0,0 +1,66 @@
...
+- **All required endpoints > 20 req/hr:** run `flowctl preview`, `pytest` (including snapshot tests that drive `flowctl preview` underneath), and capture-running commands as often as needed, without further consent.
+- **Any required endpoint ≤ 20 req/hr:** ask before each run, and apply `API-COST-SAVER-DISABLE`.
+
+Verified limits live in `<connector>/bruno/opencollection.yml`'s `docs:` block under `## API constraints (account-wide)`. Check there before re-deriving anything; record newly-established limits there so the next session doesn't repeat the work.
+
+---
+
+### `API-DISCOVER-FREE` · checkable-from-diff
+
+`flowctl raw discover` is always free. It exercises only the connector's discovery path and burns no provider API budget. No consent needed, regardless of the `API-BUDGET-20RPH` verdict.
```
Alex-Bair: note: `flowctl raw discover` may use some of the API budget if connectors make API calls to discover what resources the source system supports. Examples of those are `source-salesforce-native` and `source-zuora`. We intentionally try to limit how many calls are made during discovery, so the cost is usually pretty minimal. But it's not zero like this line implies.

estuary-cdk/CLAUDE.md
Alex-Bair, line 1: Did all the content here in `estuary-cdk/CLAUDE.md` get moved somewhere else? I saw bits and pieces in some of the skills, but I don't remember seeing the connector layout diagram anywhere. It's possible I missed it, though.

Alex-Bair (commented):
LGTM % not removing the changes to the top `CLAUDE.md` that Mahdi made in order to make Claude work better with the `grilling` skill Johnny mentioned in Slack a number of days ago.

Also, I saw that when you merged https://github.com/estuary/connectors/pull/4895, the files changed in that PR also show up under the "Files changed" tab for this one now? I'm not sure how I feel about that; I'd rather only see the changes made in _this_ slice of the stack rather than this slice + the approved and merged slices before it. Maybe there's a way to do that in the GitHub UI I couldn't find, or maybe waiting to merge slices until the entire stack is approved would be best practice to ease the burden on the reviewer?

Alex-Bair:
Oh, I see there's a "Rebase stack" button now. Maybe doing that would make it so the `source-mailchimp-native` files don't show up under the "Files changed" tab.

.claude/shared/provider-api-consent.md
nicolaslazo: Good point. I've broadened the API quota rule to cover `raw discover` as well

CLAUDE.md
nicolaslazo: It wasn't! I don't like how stacks and rebases work with Linux's basic `git` client

estuary-cdk/CLAUDE.md
nicolaslazo: `claude doctor`'s analysis stated that `estuary-cdk/CLAUDE.md` was never read during standard operations, so I tried to move everything to more commonly referenced files. The decision to remove the layout diagram came from Anthropic's guidelines for 5 family models — frontier models are better at inferring project structures.

Alex-Bair (approved):
LGTM

PR #4960, opened 2026-07-30
source-iterable-native: emit sourced schemas for all streams

Alex-Bair:
**Regression?**

No.

**Description:**

Every binding now reports a sourced schema naming `_meta` and its key fields. The emission of a sourced schema causes the control plane to raise the inferred-schema complexity limit from 1,000 to 10,000 locations. It turns out that Iterable's events can contain over 1,000+ locations, so a higher complexity limit is necessary to avoid pruning locations from the inferred schema.

**Checklist:**

### DIFF
--- a/source-iterable-native/CHANGELOG.md
+++ b/source-iterable-native/CHANGELOG.md
@@ -1,0 +1,7 @@
+# Changelog
+
+## 2026-08-05
+
+### Fixed
+- Bindings now report a source-defined schema for their key fields, which raises the
+  number of fields a collection's inferred schema can track from 1,000 to 10,000.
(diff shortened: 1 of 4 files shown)


nicolaslazo (commented):
LGTM!

Alex-Bair:
I'm planning on revising this PR a smidge after https://github.com/estuary/connectors/pull/4978 lands. Instead of re-declaring the `_meta` portion of the sourced schemas in `source-iterable-native`, I'd like to take advantage of the `BaseDocument.Meta.sourced_schema()` helper introduced in that other PR.

nicolaslazo (dismissed):

Alex-Bair:
https://github.com/estuary/connectors/pull/4978 is merged, and I've updated this PR based on the changes introduced in it. Not much really changed beyond relying on the CDK to add all the standard defaults we always want (`_meta` property, `required` lists, bound keywords, etc.).

nicolaslazo (approved):
LGTM

PR #4961, opened 2026-07-30
claude: add bugfix-capture-connector skill

nicolaslazo:
**Description:**

Adds a `bugfix-capture-connector` skill.

Supporting change:

- **New `CONDUCT-CLEAN-ROOM` shared rule** (`session-conduct.md`, registered in
  `rules-index.md`): run the connector's baseline suite in the background at
  session start and gate the first code edit on it. Pre-existing drift lands
  first as its own commit — stale snapshots as `<connector>: update tests`,
  formatter output as `<connector>: formatting` — so the real change diffs
  against a green baseline instead of mixing fix with cleanup.

None.

**Notes for reviewers:**

None.

**Checklist:**

- [x] No connector `CHANGELOG.md` or documentation update needed — this PR
  touches only `.claude/` agent tooling and changes no user-visible connector
  behavior. The `Docs / CHANGELOG check` job may need the `docs-check-skip`
  label applied as a false positive.

### DIFF
--- a/.claude/shared/rules-index.md
+++ b/.claude/shared/rules-index.md
@@ -79,6 +79,7 @@ Invoke before writing or changing any `fetch_*` / `backfill_*` function.
 | `CONDUCT-STATE-CONVENTIONS` | conduct | Say when you follow a convention, and when you deviate |
 | `CONDUCT-CITE-SOURCE` | diff | Copied patterns cite `file:line` |
 | `CONDUCT-GREP-BEFORE-CITE` | conduct | Grep for remembered symbols before recommending them |
+| `CONDUCT-CLEAN-ROOM` | conduct | Baseline suite in background at session start; gate the first edit on it; pre-existing drift commits first — stale snapshots as `<connector-name>: update tests`; format files before editing them, committing as `<connector-name>: formatting` |
 | `CONDUCT-VERIFY-FIRST` | conduct | Live behavior over docs over recall |
 
 ### Situational skills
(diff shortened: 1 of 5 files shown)


.claude/shared/rules-index.md
```diff
@@ -78,6 +78,7 @@ Invoke before writing or changing any `fetch_*` / `backfill_*` function.
 | `CONDUCT-STATE-CONVENTIONS` | conduct | Say when you follow a convention, and when you deviate |
 | `CONDUCT-CITE-SOURCE` | diff | Copied patterns cite `file:line` |
 | `CONDUCT-GREP-BEFORE-CITE` | conduct | Grep for remembered symbols before recommending them |
+| `CONDUCT-CLEAN-ROOM` | conduct | Baseline suite in background at session start; gate the first edit on it; pre-existing drift commits first — stale snapshots as `<connector-name>: update tests`; format files before editing them, committing as `<connector-name>: formatting` |
```
Alex-Bair: Does using less direct language affect how closely Claude follows these rules or when it recognizes they apply? I ask because `CONDUCT-ESTABLISH-BASELINE` more clearly communicates what this rule is for. I recognize I'm _extremely_ biased against `CONDUCT-CLEAN-ROOM` because "clean room" had a very specific meaning in my prior experiences. But the logical connection between "clean room" -> "establish a baseline" is one I didn't make immediately & had to read further to make.

We've noticed that good comments & naming help both humans _and_ agents, and I guess my concern is that this human didn't find the `CONDUCT-CLEAN-ROOM` name to be clear so I'm assuming Claude could also be confused about it too.

.claude/skills/changelog/SKILL.md
```diff
@@ -22,12 +22,14 @@ for the convention.
    ```bash
    git diff --name-only main...HEAD
    ```
-   Group changed files by top-level directory. Keep only connector
-   directories that have an existing `CHANGELOG.md` at their root —
-   ignore directories without one (per opt-in convention).
+   Group changed files by top-level directory. Every affected connector
+   gets an entry — if one has no `CHANGELOG.md` at its root yet, create
+   it as part of applying the draft (per CONTRIBUTING.md, each connector
+   has its own changelog).
```
Alex-Bair: Previously, adding a `CHANGELOG.md` was opt-in for connectors that didn't previously have it. Was it intentional to remove that wording?

Alex-Bair (commented):
Had a couple questions.

Alex-Bair (approved):

PR #4963, opened 2026-07-31
source-sqlserver: more misc. improvements

Alex-Bair:
**Regression?**

No.

**Description:**

This PR's scope includes various changes focused on reducing the CPU load `source-sqlserver` captures place on users' servers, either by reducing unneeded queries (ex: checking diagnostics every 5 minutes when there's a 10 minute polling interval) or allowing queries to be more easily cached by stabilizing the components that are used to construct the query plan cache key.

### `sqlcapture`
  - scale the long-streaming diagnostics threshold with the polling interval.

### `source-sqlserver`
- sort capture instances in the max-LSN polling query.
- bind the change table cleanup watermark as a parameter.
- parameterize the database name in the CDC-enabled check.

For a single capture, these changes probably won't have a big impact, but they should help reduce CPU load at least a little bit when users have 1,000+ captures pointing at the same server.

**Notes for reviewers:**

See [this Slack thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1785361977655109) and #4946 for context on the motivation for these changes.

Since the plan cache is server wide, I'm hoping even these little improvements help reduce eviction of our plans for the other queries the connector makes, like the chunked discovery queries.

**Checklist:**

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -251,6 +251,22 @@ func ParseRediscoveryInterval(interval string) time.Duration {
 	return parsed
 }
 
+// streamingDiagnosticsThreshold returns how long a single streaming cycle must run before it's
+// considered to have gone on for an unexpectedly long time and replication diagnostics are
+// collected.
+//
+// Diagnostics are more expensive than the progress logging they're interleaved with, since
+// for some connectors they query server-wide metadata, so they mustn't fire on cycles which are
+// slow due to their configuration. A polling connector can't complete a cycle any faster than one
+// polling period, so the threshold is that period plus a fixed grace interval rather than a
+// fixed duration on its own.
+func (c *Capture) streamingDiagnosticsThreshold() time.Duration {
+	if interval := c.Database.ReplicationPollingInterval(); interval > 0 {
+		return streamingProgressInterval + interval
+	}
+	return streamingProgressInterval
+}
+
 // Run is the top level entry point of the capture process.
 func (c *Capture) Run(ctx context.Context) (err error) {
 	// Fetch detailed schema info for the tables that this capture's bindings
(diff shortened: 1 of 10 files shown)


sqlcapture/capture.go
```diff
@@ -666,23 +682,31 @@ func (c *Capture) streamToFence(ctx context.Context, replStream ReplicationStrea
...
+				if time.Since(cycleStarted) < diagnosticsAfter {
+					continue
+				}
 				log.Warn("replication streaming has been ongoing for an unexpectedly long amount of time, running replication diagnostics")
-				// Dump goroutine stacks on the first tick and every fifth tick thereafter
-				// to help with investigations into potentially stalled captures.
-				if progressTicks%5 == 0 {
+				// Dump goroutine stacks on the first diagnostics run and every fifth run
+				// thereafter to help with investigations into potentially stalled captures.
+				if diagnosticRuns%5 == 0 {
```
willdonnelly: It feels like we could probably combine the diagnostics and stack dumping behaviors again, in both cases it seems useful to have a shorter timeout for the first instance followed by less frequent repeats after that. Heck, for years we had it just running diagnostics once and then not repeating within a streaming cycle and that was enough _most_ of the time.

willdonnelly (dismissed):
LGTM % one comment

sqlcapture/capture.go
Alex-Bair: Good point, that makes sense. 👍  I've put the diagnostics and stack dumping on the same intervals.

Alex-Bair: (Rebasing onto `main` shows a lot of noise in the "Compare" button diff since some beefy PRs were merged this morning. Here's a [link](https://github.com/estuary/connectors/compare/a834591ff82c60da42821320652c2b9096ead5e2..a25dab940292ffb5cb18c5cb39b62b852ff34d26#diff-2f6b748465d6ef38afae2fe58f56a57943f6e90a2105b63e5701402b50e87845) to more easily see the changes I made to `sqlcapture/capture.go` in the most recent force-push in response to this comment).

willdonnelly (approved):
LGTM, thanks for these!

PR #4974, opened 2026-07-31
source-dynamics-365-finance-and-operations: read CSVs written across a column addition

Alex-Bair:
**Regression?**

No.

**Description:**

We observed a capture failing with an error stating `CSV row N has X columns but Y were expected`. It turns out that it's possible for Synapse Link to write headerless CSVs with variable row widths, and that error is indicating that rows in a CSV file don't all have the same number of columns as the headers that were provided to the `IncrementalCSVProcessor`.

What appears to be happening is that when a column is added to a table, Synapse Link starts writing it at the end of every row. And if that happens while a CSV is actively being written to, the file ends up with narrow rows before the column addition & wide rows after (ex: The first few rows have 117 columns -> D365 adds a new column -> the remaining rows have 118 columns). A timestamp folder's `model.json` is written at the end of the interval & describes the wider shape, and the connector was (intentionally) choking on the narrow rows.

To fix this, the connector now treats narrow rows as a prefix of the total column set, where the rows are bound sequentially to bound sequentially to the field names in the order `model.json` lists the table's attributes, stopping when the row runs out of cells. The columns a narrow row lacks are left absent from the document rather than set to null. This works because Synapse Link appears to only ever appends a new column to the end of a row, after `IsDelete` - which makes a narrow row exactly the declared schema truncated on the right, so every cell it does have still lines up with the correct field name. Based on my investigation of a source system that contains headerless, variable width CSVs, this appears to be safe handling.

- Adding a new `IncrementalCSVRowProcessor` to the CDK that incrementally reads rows from a CSV without binding them to any field names. This means the connector needs to do the field name binding itself, but it also allows more flexibility for handling these hopefully rare headerless, variable width CSV situations.
- Updating `source-dynamics-365-finance-and-operations` to use the new `IncrementalCSVRowProcessor` and bind field names to rows. This allows the connector to handle the column addition scenario described above that resulted in a variable width CSV & no longer choke on it.
- Adding a lot of guardrails and checks for the class of problems that are no longer checked for automatically by `IncrementalCSVProcessor`. Stuff like rows that are too narrow, rows that are too wide, decreasing row widths in a file, new columns no longer being append only. We want to know if/when this happens so we can investigate further and determine how to handle whenever an odd CSV situation happens again.

**Notes for reviewers:**

Best reviewed commit-by-commit. The essence of the changes are "use a subset of the `model.json` schema to parse narrow rows and add a lot of guardrails to detect when that's probably not safe to do so" but spread out over a few commits to hopefully make them more digestible.

Tested with `flowctl preview`. Confirmed the connector now successfully processes the variable width CSV that it previously crashed on.

**Checklist:**

### DIFF
--- a/estuary-cdk/estuary_cdk/incremental_csv_processor.py
+++ b/estuary-cdk/estuary_cdk/incremental_csv_processor.py
@@ -76,6 +76,24 @@ _SURPLUS_COLUMNS_KEY = object()
 _MISSING_COLUMN = object()
 
 
+def _dialect_kwargs(config: CSVConfig) -> dict[str, Any]:
+    """Build the csv dialect kwargs shared by both processors."""
+    kwargs: dict[str, Any] = {
+        'delimiter': config.delimiter,
+        'quotechar': config.quotechar,
+        'doublequote': config.doublequote,
+        'skipinitialspace': config.skipinitialspace,
+        'lineterminator': config.lineterminator,
+        'quoting': config.quoting,
+        'strict': config.strict,
+    }
+
+    if config.escapechar is not None:
+        kwargs['escapechar'] = config.escapechar
+
+    return kwargs
+
+
 class _AsyncByteReader(aiocsv.protocols.WithAsyncRead):
     """Internal class that handles incremental decoding for aiocsv."""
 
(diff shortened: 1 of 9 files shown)


nicolaslazo (approved):
LGTM

Issue #4970, opened 2026-07-31
materialize-snowflake: a streaming v2 backfill races the outgoing generation and silently duplicates rows


jacobmarble:
## Parent

- estuary/connectors#4324 — Snowflake Snowpipe Streaming: Switch to SDK

## What to build

Bumping a binding's backfill on a live `snowpipe_streaming_v2` task duplicates rows. This
is the only exactly-once violation found in a 12-hour chaos run, and it is **silent** — no
error, no refusal, nothing in the logs.

`Apply` truncates the table for the new generation while the *outgoing* generation's shards
are still appending to their old channel, which still points at the same table. Those rows
land after the truncate, survive it, and are then re-materialized by the new generation.

Confirmed by arithmetic rather than inference:

```
rows before the duplicate window    : 2,057,375
outgoing generation's final counter : 2,057,375     <- identical
duplicates: 20,000, each document exactly twice, contiguous from that point
total_rows 62,440,172 - distinct_docs 62,420,172 = 20,000  (= sum of counters)
```

Ordering the turnover eliminates it, validated twice — 0 duplicates from both a healthy
and a wedged task, against 20,000 when the shards were live during `Apply`. Same root
cause as the join-then-split wedge: nothing ever closes a channel. There, a *retired*
channel keeps its token; here, an *abandoned* channel keeps writing.

Deliver a backfill bump on a running task that produces zero duplicates, with no operator
ceremony required.

## Acceptance criteria

- [x] A regression test for the racing turnover lands before the fix, and is confirmed
      failing.
- [x] The outgoing generation's channels can no longer append to the table once the new
      generation has truncated it.
- [x] The conservation check holds across a turnover on a live task: total rows minus
      distinct documents is zero.
- [x] The outgoing generation cannot reopen the channel it was dropped from, and the
      conservation check is run with its shards actively storing rather than merely alive
      (see the drop-semantics comment below).
- [ ] ~~Until the fix ships, the validated operator recipe is documented: publish
      `shards.disable: true`, confirm every shard has lost its assignment, *then* publish
      the backfill bump. Idle took 26s on a failed task and 14 minutes on a healthy one.~~
      Dropped: this write path is unreleased, so there is no fleet to carry through an
      interim.

## Blocked by

- estuary/connectors#4965 — the retirement mechanism this fix is built on.

---

*Found by a 12-hour autonomous chaos test of the `snowpipe_streaming_v2` write path on
[...]

PR #4828, opened 2026-07-17
materialize-snowflake: Snowpipe Streaming v2 via Python SDK sidecar

jacobmarble:
**Description:**

Fixes https://github.com/estuary/connectors/issues/4324

Adds an opt-in write path for Snowflake's high-performance Snowpipe Streaming architecture, behind the `snowpipe_streaming_v2` feature flag. There is no Go SDK, so the official Python SDK (`snowpipe-streaming`, PyO3 over Snowflake's Rust core) runs in a supervised sidecar process, spoken to over NDJSON RPC on a unix domain socket (TCP-over-localhost fallback). Existing write paths (Go-native BDEC streaming, classic Snowpipe, COPY/MERGE) are untouched.

- **Requires the v2 runtime.** The path resumes from Snowflake's committed offset token, which is only correct if an interrupted transaction is replayed identically. Setting the flag on a task without the `enable-runtime-v2` shard flag is refused in `Apply` (so it fails at publish) and again in `Open` (so turning the runtime flag back off bricks the task rather than silently changing delivery semantics). Setting both streaming flags is a config-validation error. No escape hatch for either.
- **Direct append, no staging.** Rows are appended to the channel during `Store` — each row is transmitted to Snowflake exactly once, and nothing is written to the internal stage. Buffered per binding and appended at 8 MB / 10,000 rows (plus a 128 MB cap on total buffered bytes, so a wide materialization's worst-case resident memory isn't a multiple of the binding count), with at most one in-flight append per channel so ordering holds while `Store` keeps buffering. Back-pressure is implicit: a saturated pipe blocks the append, which blocks `Store`.
- **Offset tokens are a monotonic per-channel document counter**, never reset across transactions. The driver checkpoint holds, per binding, the channel name, the counter, and the shard's key range — durable state rather than pending work, so `Acknowledge` deliberately does not drain it. At `Open`, Snowflake's committed token `C` is reconciled against the checkpoint counter `K`: equal or absent is the steady state, `C > K` skips what Snowflake already holds (guarded by an unchanged key range), and `C < K` — which the commit ordering below narrows to Snowflake having lost committed data — is refused. Backfilling the binding is the remedy for either refusal.
[...]

PR #4975, opened 2026-07-31
materialize-snowflake: read the streaming v2 turnover decision from the configuration in hand

jacobmarble:
**Regression?**

Yes — of the not-yet-merged turnover work on #4828. That branch decided whether a
streaming v2 backfill must drop its table from the *replaced* specification's
endpoint configuration, read out of `req.LastMaterialization.ConfigJson`. On a
live task that document is not the endpoint configuration this connector is handed
for itself: it arrives as the control plane stores it, the connector image
alongside a configuration whose secrets are still sops ciphertext. It parsed into
an empty configuration, the feature flag read as unset, and the backfill truncated
the table instead of re-creating it — the very thing #4970 is about.

**Description:**

The decision is now read from the configuration in hand, which is the one that
generation is running unless the same publication also changes the write path. The
replaced specification's own configuration is still consulted, in either shape it
may arrive in, but only to widen the answer — a document that cannot be read never
narrows it.

**Workflow steps:**

Unchanged for users. Backfilling a `snowpipe_streaming_v2` binding drops and
re-creates its table, as #4828's changelog entry already describes.

None.

**Notes for reviewers:**

Found by deploying #4828's CI image (`:c4b0916`) to a live 8-shard task and
watching the backfill truncate rather than drop. This branch is based on #4828's
tip `ccf059d17` and carries only the one fix, so it can be folded into #4828
instead of merged on its own.

**Checklist:**


jacobmarble:
Folded into #4828 as cfcd20818 — this branch existed only so CI would build an image to test the fix on a live task. Verified there: the backfill now drops and re-creates the table (`created_on` moves, `flow_generation` marker written) instead of truncating it.

PR #4977, opened 2026-07-31
source-postgres: fix read-only captures stalling when captured tables are quiet

Alex-Bair:
**Regression?**

No.

**Description:**

A fence is set from `pg_current_wal_flush_lsn()` - a server-wide position - but can only be satisfied by a commit to a table we capture. Write-mode captures manufacture one via the watermark write; `read_only_capture` skips it. When a source's captured tables are quieter than the instance, the fence is never reached and the capture blocks in `StreamToFence` indefinitely.

The checkpoint sits behind that fence, so the runtime can't end the session to rotate IAM credentials either. They expire, the pool dies, and when the fence finally clears the capture crashes on its next query. I'm pretty sure that's the sequence of events that led to the `SQLSTATE 28000` error reported in this [Slack thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1785522487203939).

Observed in production: fence target 2.25 GiB of WAL ahead of the capture's position, stalls of 60 minutes to 5 hours.

To fix this, we're now checking keepalive messages to see how far the server has decoded. Primary keepalive messages are queued behind everything it sent from below that position, so one at or past the fence proves no published change before the fence is still in flight. The connector was already parsing these messages and discarding the position. Now it uses it, and an idle fence clears in seconds.

**Checklist:**

### DIFF
--- a/go/capture/blackbox/capture.go
+++ b/go/capture/blackbox/capture.go
@@ -581,8 +581,17 @@ func (tc *TranscriptCapture) DiscoverFull(description string) []json.RawMessage 
 }
 
 func (tc *TranscriptCapture) Run(description string, sessions int) []byte {
+	return tc.RunWithContext(context.Background(), description, sessions)
+}
+
+// RunWithContext is Run with a caller-provided context bounding how long the capture may
+// run. When the context expires the capture process is killed and the resulting error is
+// recorded in the transcript, so a capture which stops making progress shows up as an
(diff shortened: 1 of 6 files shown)


Alex-Bair:
Closing this for now - the immediate need for this have been alleviated, and I'm not actively working on this problem. If we need to / decide to address this issue later, we can open a different PR then.

PR #4978, opened 2026-08-01
estuary-cdk: apply defaults to sourced schemas

Alex-Bair:
**Regression?**

No.

**Description:**

`SourcedSchema`s need to be maximally restrictive, and these earlier sourced schema implementations in various connectors were lax around whether `_meta` was required. The `minLength`, `maxLength`, and other keywords indicating the bounds of a field should be included as well, and they often weren't.

This PR's scope includes:
- adding a method to the CDK's `BaseDocument.Meta` class so generating the `_meta` portion of the sourced schemas is easier.
- automatically adding `_meta` information to sourced schemas for CDK based captures when it was omitted.
- automatically adding bounds for string and numeric fields in sourced schemas for CDK based captures when they're omitted.
- automatically marking every declared property as required.
- automatically setting `type: object` and `additionalProperties: False` on the root and on every location whose properties are declared. The runtime rejects a sourced schema that isn't closed, so this is a hard requirement.

The above automatic additions do not override fields that are already set, so captures can continue to customize the `_meta` and bounds in sourced schemas as needed. The two exceptions are the ones with no valid alternative: the root is
always an object, and an object whose properties are declared is always closed. 

Marking every field as required is likewise not opt-out. If a document is captured without one of those fields, schema inference will handle removing it from the `required` array.

Note: A connector that wants a nested object left open to spend inferred schema complexity points elsewhere can still do so by declaring it as `type: object` with no properties.

**Checklist:**

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/document.py
+++ b/estuary-cdk/estuary_cdk/capture/document.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Generic, Literal, TypeVar
+from typing import Any, Generic, Literal, TypeVar
 from pydantic import BaseModel, Field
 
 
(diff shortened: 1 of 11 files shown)


nicolaslazo (dismissed):
Looks good! But it seems like there are some merge conflicts to resolve. I'll keep an eye out and re-approve

Alex-Bair:
Sorry about that, thanks for catching that! It was a conflict in one of the `CHANGLOG.md` files, but it's fixed now. I rebased this branch back onto `main`, so there shouldn't be merge conflicts now.

nicolaslazo (dismissed):
🙌

Alex-Bair:
> source-facebook-marketing-native/CHANGELOG.md — Removing 'required' from sourced schemas changes captured document validation behavior but doesn't affect documented config/setup.

That's the neat thing - we didn't remove `required` from the actual sourced schemas! It is added in automatically at the CDK level instead of requiring each individual connector to include it themselves. I'll apply the `docs-check-skip` label as suggested.

Alex-Bair:
I caught a bug prior to merging: if a connector defined a `_meta` property in its sourced schema that's passed to the CDK, then the CDK didn't make sure all the standard subfields (`op`, `row_id`, `uuid`) were included in under `_meta` in the sourced schema it ultimately emits. I updated the PR to handle this situation now. Here's the [diff](https://github.com/estuary/connectors/compare/b0e608bc3840abb19e64d05438ca35efa71d4bb1..378c199cfb4f3252d7d8a00b89539686de20b4df) for what I just pushed.

nicolaslazo (approved):

PR #4979, opened 2026-08-02
source-criteo: rewrite with the `estuary-cdk`

Alex-Bair:
**Regression?**

No.

**Description:**

This PR rewrites `source-criteo` as a native connector using the `estuary-cdk`. The native implementation is significantly different, so the version has been bumped from `v1` to `v2`.

The existing `source-criteo` connector built with a shim from the legacy `python` library around the Singer `tap-criteo` has rotted & no longer works. There are no enabled captures for `source-criteo`, and its usage of the `python` library is the last thing stopping me from removing the legacy `python` library altogether. Instead of ripping out the `source-criteo` connector and saying "we don't support capturing from Criteo any longer", I opted to rewrite it with our CDK so we still have a `source-criteo` connector. I opted to do this since the `create-capture-connector` skill from https://github.com/estuary/connectors/pull/4958 lowered the time investment for implementing a reasonable native connector.

The connector's documentation should be updated to reflect the native implementation instead of the `tap-criteo` implementation.

**Notes for reviewers:**

I didn't have any Criteo credentials for use during development, but the main motivation of this PR is to enable the removal of the legacy `python` library, _not_ to build a perfectly working `source-criteo` connector. To that end, I think it's reasonable to not invest much more effort into fine tuning the connector until there's an actual use case for a `source-criteo` capture.

Supabase will need updated with a new `connector_tags` row for `v2` of `source-criteo`.

**Checklist:**


source-criteo/source_criteo/api.py
```diff
@@ -0,0 +1,505 @@
...
+    # check reports any `errors` Criteo returned.
+    probe_day = floor_to_day(datetime.now(tz=UTC)) - TICK
+
+    async with aclosing(
+        _request_report_window(
+            http, report, document_model, advertiser_ids, probe_day, probe_day, log
+        )
+    ) as rows:
+        async for _ in rows:
+            break
```
nicolaslazo: Can we guarantee errors/warnings never yield any rows? In the case they do, those problems would never raise, as `aclose` kills the generator before we call `_raise_for_problems`.

```python
    async for row in processor:
        rows += 1
        yield row  # `async for _ in rows` yields here
                   # `break` breaks here, the generator is closed
    envelope = processor.get_remainder()   # this line is never executed
    _raise_for_problems(envelope.errors, f"report {report.name}")
    _log_problems(log, envelope.warnings, f"report {report.name}")
```

source-criteo/source_criteo/api.py
```diff
@@ -0,0 +1,505 @@
...
+from typing import Any, AsyncGenerator
+
+# `alru_cache` keeps the portfolio lookup off the wire for repeat callers and
+# collapses concurrent misses onto a single shared future, so simultaneous
+# binding tasks don't stampede the endpoint.
+from async_lru import alru_cache
+from estuary_cdk.capture.common import LogCursor, PageCursor
+from estuary_cdk.http import HTTPSession
+from estuary_cdk.incremental_json_processor import IncrementalJsonProcessor
+from pydantic import ValidationError as PydanticValidationError
```
nicolaslazo: This import is not referenced. Maybe it was meant to be used for endpoint config errors?

source-criteo/source_criteo/models.py
```diff
@@ -0,0 +1,413 @@
...
+class EndpointConfig(BaseModel):
+    credentials: OAuth2Credentials = Field(
+        title="Authentication",
+        description="Criteo API client credentials.",
+    )
+    start_date: AwareDatetime = Field(
+        default_factory=default_start_date,
+        title="Start Date",
+        description="UTC date and time in the format YYYY-MM-DDTHH:MM:SSZ. Report data before this date will not be replicated. If left blank, the start date will be set to 30 days before the present.",
+    )
```
nicolaslazo, line 294: Does Criteo gracefully handle `start_date`s behind the 2-year data retention period? If not, we should probably automatically clamp this and raise a warning

source-criteo/source_criteo/models.py
```diff
@@ -0,0 +1,413 @@
...
+
+    Criteo returns one flat object per row, keyed by the requested dimension and
+    metric names. Only the dimensions are declared — including the grain, which
+    leads them: they form the collection key, so they must be present and
+    non-null on every document. Metrics vary in type across the ~80 Criteo offers
+    and are left to schema inference.
+    """
+    fields: dict[str, Any] = {
+        dimension: (str, ...) for dimension in report.all_dimensions
+    }
```
nicolaslazo, line 417: Nit: I understand dimensions can be numerical ids too

```suggestion
    fields: dict[str, Any] = {
        dimension: (str | int, ...) for dimension in report.all_dimensions
    }
```

source-criteo/source_criteo/api.py
```diff
@@ -0,0 +1,505 @@
...
+        ):
+            yield row
+
+        window_start = window_end + TICK
+
+        # Checkpoint between windows so a long sweep doesn't restart from the
+        # beginning. Cursors stay below `now` so the final one is always greater.
+        if cursor < window_start < now:
+            cursor = window_start
+            yield cursor
```
nicolaslazo, line 528: Should we be returning here or is it safe to continue? I don't see why we can't, but it strays from convention so asking jic

nicolaslazo (commented):
Left a few concerns, some less serious than others

source-criteo/source_criteo/models.py
Alex-Bair: Good catch. I don't think Criteo does handle requests for reports earlier than 2 years - we've actually seen errors before when requesting reports for data older than 2 years, so we should definitely clamp the start date we use like you suggested. Fixed.

source-criteo/source_criteo/api.py
Alex-Bair: Thanks for catching that, it's been removed. It was probably leftover from when there was one of the validation functions in `api.py` before I moved it closer to where it was used in `resources.py`.

source-criteo/source_criteo/api.py
Alex-Bair: Nope, thanks for catching this. It always trips me up when APIs return errors alongside data (or just respond with 200 OK and the body contains an error). This is fixed now by `pass`-ing through the entire response instead of breaking out early.

source-criteo/source_criteo/models.py
Alex-Bair: Hmm, I'm not 100% sure if the Criteo API returns certain dimensions as actual numbers or as strings that are formatted as numbers. Their docs aren't explicitly clear about which is the case, and it seems like it may vary depending on how you request the report, although after digging through the docs I'd say the odds are 65% they're always strings.

Regardless of what Criteo returns, our platform enforces that [key locations must be either nullable strings, nullable integers, or nullable booleans](https://github.com/estuary/flow/blob/ae3495ec790a7ab1fe0a5ecf74bd943b3abf1375/crates/json/src/schema/types.rs#L267-L295) - they can't be a union of those types like `str | int`. Meaning, since a report's dimensions are its keys, those dimensions must be either strings or integers. Realistically, if Criteo does return some dimensions as numbers like `75.43`, we'll likely need to convert those to strings before emitting documents. But I'd rather do that only if needed. We'll quickly find out if that's the case once this capture gets some real use, and we can fix if needed then.

source-criteo/source_criteo/api.py
Alex-Bair: In this case, it's safe to continue here instead of returning before yielding `now`, and it's actually preferable due to how this connector is using the lookback window on every single `fetch_report` invocation. The `first_day` is `floor(log_cursor) - lookback_days`, so if we returned after `yield cursor`, we'd rewind the next invocation to 30 days behind the cursor we just emitted.

If the capture stays relatively caught up to the present, that's not really an issue since there are essentially no intermediate cursors between `log_cursor` and `now` to emit. But if the capture falls behind for whatever reason and `fetch_report` is catching up on many days, there's now intermediate cursors between `log_cursor` and `now` that are yielded, and we'd fetch a lookback window for each window  processed as we catch up to the present. It's not the end of the world if that happens - we'd fetch some duplicate data that gets de-duped elsewhere - and we'd end up doing that if the capture crashes after `yield cursor` but before `yield now`. But we can avoid the data duplication in most situations by ensuring the final checkpoint in a `fetch_report` invocation is always `now`.

Alex-Bair:
Thanks for the review & comments! I've addressed most as suggested & explained my reasoning for those I didn't take action on. PTAL when you have some spare time.

nicolaslazo (approved):
LGTM, thanks!

PR #4980, opened 2026-08-02
source-brevo: rewrite with the `estuary-cdk`

Alex-Bair:
**Regression?**

No.

This PR rewrites `source-brevo` as a native connector using the `estuary-cdk`. The native implementation is significantly different, so the version has been bumped from `v1` to `v2`.

The existing, imported `source-brevo` connector doesn't work well, has no active captures associated to it, and has been a dependency management nightmare due to using a newer version of the Airbyte CDK than any other imported connect in this repo (ex: the dependencies for this connector are restricted enough that sorting it out _without_ rewriting it as a native connector is more difficult than just rewriting it). Instead of ripping out the `source-brevo` connector and saying "we don't support capturing from Brevo any longer", I opted to rewrite it with our CDK so we still have a `source-brevo` connector. I opted to do this since the `create-capture-connector` skill from https://github.com/estuary/connectors/pull/4958 lowered the time investment for implementing a reasonable native connector.

The connector's documentation should be updated to reflect the native implementation instead of the imported Airbyte `source-brevo` implementation.

**Notes for reviewers:**

The main motivation of this PR is to address tech debt and make dependency management for the `estuary-cdk` easier, _not_ to build a perfectly working `source-brevo` connector. To that end, I think it's reasonable to not invest much more effort into fine tuning the connector until there's an actual use case for a `source-brevo` capture.

Supabase will need updated with a new `connector_tags` row for `v2` of `source-brevo`.

**Checklist:**


nicolaslazo (approved):
Looks good, great to see the new skillset working for you too

PR #4981, opened 2026-08-03
materialize: make connector drivers importable, so a harness can read a resource back

mdibaiee:
**Description:**

Lets a program outside a materialization connector construct its `Materializer`, so
estuary/flow's `materialize-consistency` suite can read a destination back to check what
landed and drop the resources it created. `SnapshotTestResource` and `DeleteResource` already
exist on `boilerplate.Materializer` and the integration tests use them; what was missing was
any way to reach them, since every connector was `package main`.

- Every materialization connector with integration tests becomes `package connector` with
  `func main` under `cmd/connector`, as `materialize-iceberg` already was, and exports
  `NewDriver` (plus a package-level `NewMaterializer` where its `Connector` is not a
  `*sql.Driver` already carrying one).
- `materialize-boilerplate/testutil` gains `SnapshotResource`, `DropResource`, `OpenResource`
  and `SweepTestResources`/`SweepTestTasks`, usable without a `*testing.T`. The first two
  replace logic that `snapshotTestTable`, the fencing test and the drain test each had
  inlined; the existing cleanup helpers now wrap the sweeps.
- `tests/materialize/testctl` is the program the harness invokes, with the 17 connectors whose
  destination can be read back.

No connector's production interface or protocol surface changes.

**Workflow steps:**

```console
go run ./tests/materialize/testctl -connector materialize-databricks \
    -config config.yaml -resource resource.yaml -mode snapshot
{"balanceDelta":-28,"id":0,"oracle/balance":-266,"seq":22, ...}
```

One JSON object per line, keyed by column name. `-mode drop` removes the named resource;
`-mode sweep` removes what earlier test runs left behind, `-dry-run` lists it first. Configs
are file paths (JSON or YAML), not inline values, since an endpoint config carries credentials
and an argv is readable by anyone who can list processes.

None. `tests/materialize/testctl/README.md` documents the tool.

**Notes for reviewers:**

- The Dockerfile and testdata changes are required: `go build -o ./connector
  ./materialize-x/...` cannot write two packages to one file, and `go run .` no longer names a
  main package.
- Snapshot rows are objects of columns rather than the documents that produced them — a
  standard binding need not carry a root document — and are unordered, since nothing here can
  know which column defines a meaningful order.
- Sweeping selects resources carrying `_flow_test_<unix>`; the timestamp is what lets it leave
  a concurrent run's resources alone.
- Verified: all 21 importable connectors vet clean and pass short tests with
  `UPDATE_SNAPSHOTS=true` and no snapshot changes, and every new `cmd/connector` answers a Spec
  RPC. The full `materialize-databricks` integration suite passes. Against the real warehouse
  the tool snapshotted 275,581 rows as NDJSON, dropped that resource, and a dry-run sweep over
  207 tables correctly spared an in-flight test's tables.
- CI is green for 30 of 32 materialize connectors. `materialize-azure-fabric-warehouse` (Fabric
  mid-system-update) and `materialize-s3-iceberg` (`apache/polaris:latest` now rejects
  namespace locations outside the catalog base) also fail on `main`.

**Checklist:**

- [x] Not user-visible — no connector behaviour, configuration or protocol change — so no
  `CHANGELOG.md` or documentation update applies.


materialize-postgres/client.go
```diff
@@ -1,4 +1,4 @@
-package main
+package connector
```
danielnelson, line 1: Optional.  Could be worth using the suggested package alias instead of always using `connector`, I believe this will set the import name so we don't HAVE to alias them manually, though I still like to for readability.

The `goimports` tool, maybe others, automatically use this to update imports with this value.  So if you use `postgres.NewDriver` it will (in theory) automatically add:
```go
import (
	postgres "github.com/estuary/connectors/materialize-postgres"
)
```
Sometimes it gets goofy when it has multiple choices.

tests/materialize/testctl/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+A connector can be driven once two things are true. Its package must be
+importable — `package connector` with `func main` under `cmd/connector` — and it
+must expose the Materializer: either `NewDriver`, whose `*sql.Driver` carries
+`NewMaterializer`, or a package-level `NewMaterializer`. Then add it to the
+`connectors` map in `main.go`.
+
+Connectors materializing into something that cannot be read back are left out
+even though they are importable: `materialize-eventbridge` and `materialize-sns`
+publish to an event bus, `materialize-hubspot` writes through an API, and
```
danielnelson, line 92: The main issue for supporting HubSpot is that it requires a custom collection format; we can't just write to the shared set of fields. Other than that, the interactive step required to auth the app with a test-account definitely is annoying.  Reading back isn't a big problem, though loading isn't required for normal operation I added a version of it for the integration test.

danielnelson (approved):

tests/materialize/testctl/README.md
danielnelson: An additional issue is that HubSpot doesn't support anything other than LWW since it works in delta update mode.

PR #4985, opened 2026-08-03
source-zuora: capture foreign keys by selecting related-object ids

Alex-Bair:
**Regression?**

No.

**Description:**

Exports selected only the fields `/describe` lists under `<fields>`, which omitted every foreign key on every object: an `InvoiceItem` with no `InvoiceId`, an `Invoice` with no `AccountId`. Export ZOQL exposes foreign keys only as joins on the related object, and `/describe` lists those relationships in their own section rather than among `<fields>`. Each joinable relationship now contributes one `<Name>.Id` to the query and nothing else since the related entities are captured as their own bindings.

Joined columns land as `<Name>Id`. That means a column can no longer be named by stripping its `<Object>.` prefix unconditionally: every joined `<Related>.Id` would strip to `Id` and overwrite the document's own key. A prefix is now stripped only when it names the object being exported.

Skipped are self-references, relationships the object already exposes as a scalar `<Name>Id` field, and `SubscriptionStatusHistory`, which joins only under a mandatory `WHERE` bound on its effective-date range that would constrain the parent query and drop parent rows.

**Checklist:**

### DIFF
--- a/source-zuora/CHANGELOG.md
+++ b/source-zuora/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 2026-08-04
+
+### Added
+- Exports now select each object's related-object ids, so documents carry the
+  foreign keys that let them be joined to the objects they belong to
+  (`InvoiceItem.InvoiceId`, `RatePlan.SubscriptionId`, `Invoice.AccountId`, ...).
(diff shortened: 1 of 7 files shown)


nicolaslazo (approved):
LGTM

PR #4994, opened 2026-08-04
source-braintree-native: release fetched pages as they're emitted

Alex-Bair:
**Regression?**

Yes, https://github.com/estuary/connectors/pull/4908. Specifically, commit 20a3139db572293272b02ed8150644b5a932a36d.

**Description:**

`process_completed_fetches` retained every page it fetched until the whole fan-out finished, so a backfill window held all of its records in memory at once instead of a bounded working set. Backfill windows are bisected to just under Braintree's search limit, so a whole fan-out is a _lot_ of data, and tasks that are backfilling have been OOMing as a result.

A completed `asyncio.Task` holds onto its return value, so the list of tasks introduced to make cancellation possible was enough to pin every page, including pages already yielded. This PR switches the connector to use `asyncio.wait` instead of `asyncio.as_completed`. `asyncio.wait` reports the still-pending tasks each round and lets each task be dropped once its page has been drained.

**Checklist:**


source-braintree-native/source_braintree_native/api/common.py
```diff
@@ -74,17 +74,37 @@ async def process_completed_fetches(
...
+                completed, pending = await asyncio.wait(
+                    pending, return_when=asyncio.FIRST_COMPLETED
+                )
+
+            task = completed.pop()
+            for resource in task.result():
                 yield resource
+
+            # Release the page before awaiting the next fetch.
+            del task
```
nicolaslazo, line 102: Please correct me if I'm misinterpreting things, but I'm not sure this would be the definitive solution to our memory issue. While deleting tasks would make memory usage drop off faster, meaning we'd have more space for pages yet to come, this line:

```python
pending = {asyncio.ensure_future(coro) for coro in fetch_coroutines}
```

Would eagerly schedule every coro to start at the same time. In the worst case scenario where all requests return simultaneously, we'd still be holding all pages in memory at the same time.

A more deterministic approach would be making every fetch coroutine acquire a shared semaphore before making any API requests, and only releasing it after it has been consumed. That would make it safe to return to the `asyncio.as_completed` loop, with memory usage bounded to `semaphore_count * page_size` in the worst case.

nicolaslazo (changes requested):
Left a worst case scenario concern

source-braintree-native/source_braintree_native/api/common.py
nicolaslazo: It is my impression that [`buffer_ordered`](https://github.com/estuary/connectors/blob/main/estuary-cdk/estuary_cdk/buffer_ordered.py) addresses this exact page lifetimes issue, even though we don't particularly care about which pages get yielded first

Alex-Bair: That's correct, we definitely can't go fetch every single page at once. There's already a semaphore per stream to bound how many requests are in flight at one time. The semaphores used to be hard coded to 20, but I recently made them configurable in this [commit](https://github.com/estuary/connectors/commit/aa77b4d3517d32118a460f4bb030ae1633b0495f). Those semaphores make sure that, even though all coroutines are scheduled to start at the same time, only `concurrency` of them actually make API requests at a time.

I moved away from the `asyncio.as_completed` loop because we don't have access to the `Task`s in order to cancel them - `asyncio.as_completed` creates `Task`s internally and only yields awaitables that resolve to results, so we don't have a handle on the underlying `Task`s to cancel them when a request fails & prevent a ton of confusing `Session is closed` log noise since there are outstanding `Task`s still making requests after the `aiohttp.ClientSession` has shutdown.

nicolaslazo (approved):

source-braintree-native/source_braintree_native/api/common.py
nicolaslazo: Ah! Sorry, I forgot we merged that recently. Now I understand the actual scenario we're addressing

Alex-Bair: Yep yep, no problem. I should have worded it a bit better so the actual problem we're trying to solve here was clearer.

And on `buffer_ordered`, you're right that it would give us a hard guarantee on page lifetime, but it would do so at the cost of some throughput. With `buffer_ordered`, since output is ordered, a worker can't pick up another page while it's still holding one the output loop hasn't reached yet. So if a single page is slow, we fetch at most `concurrency - 1` more pages and then the stream stops fetching entirely until that page finishes.

Since we've seen some requests hit timeouts that we now retry, I'd anticipate those timeouts would slow things down further if we used `buffer_ordered`. We've put in quite a bit of effort to squeeze as much throughput out of `source-braintree-native` as is reasonable, so I'd rather we solve this problem without accepting the throughput hit it would incur.

PR #5004, opened 2026-08-05
source-(various): Shutdown logging when there are zero bindings

### LINKED ISSUE #5001: source-mongodb: capture with no enabled bindings is accepted and exits silently
### What happens

A `source-mongodb` capture with no enabled bindings (either no bindings at all, or every binding `disable: true`) validates, publishes, and runs. Each session starts the connector, which returns immediately without logging why.

`Pull` handles the case explicitly, but returns with no log:

```go
allBindings := append(changeStreamBindings, batchBindings...)   // pull.go:157
...
if err := c.output.Ready(false); err != nil { return err }      // pull.go:189
...
if len(allBindings) == 0 {
	// No bindings to capture.
	return nil                                                  // pull.go:201-203
}
```

`Ready` is sent before that check, so the task emits `Starting capture` / `Capture started` and then goes quiet. `driver.Validate` (`main.go:398`) iterates `req.Bindings` with no cardinality check, so an empty set is trivially valid.

### Observed effect

Task logs read like a healthy capture that stops saying anything:

```
06:33:17  started connector container   {image: "ghcr.io/estuary/source-mongodb:v2"}
06:33:18  Starting capture
06:33:20  Capture started
06:38:20  initialized catalog task term     <- next session, 5 minutes later
```

Nothing indicates the connector finished because there was nothing to do. Working that out means going and reading the spec's binding list, and the repeating 5-minute cycle looks like a restart loop rather than a no-op.

Under runtime v2 it is also costly: the shard session is held for the remainder of the poll interval after the connector exits, so a capture in this state accrues roughly 24 task-hours/day while moving zero bytes. That metering behaviour is a separate matter and is being tracked on its own; noting it only as motivation for surfacing this state rather than leaving it silent.

### Suggested

1. Fail or warn when no bindings are enabled. Rejecting in `Validate` is the strongest signal, but it would also block re-publishing a capture that someone has deliberately disabled in full, so a warn-level log in `Pull` may be the safer default. Either seems better than silence.
2. Log on the exit path, so the quiet is explained, e.g. `no enabled bindings, nothing to capture; shutting down`.

### Possibly broader than this connector

The platform accepts publishing a capture with zero enabled bindings in the first place. Whether that should be rejected centrally is a `flow` validation question rather than a connector one, so the better fix may live there. Other capture connectors likely have the same silent-exit shape, and this was only noticed in `source-mongodb`.

willdonnelly:
**Description:**

Adds informational logging to several connectors (source-mongodb, source-snowflake, and source-*-batch) when there are zero enabled bindings. Each of these connectors would previously shut down in this case (MongoDB intentionally as a special case, Snowflake because it always shuts down after polling each binding, and the batch SQL connectors because if you wait on an empty worker group it returns immediately), but now we'll actually be saying _why_ they're shutting down.

Among other things, this fixes https://github.com/estuary/connectors/issues/5001

### DIFF
--- a/source-bigquery-batch/driver.go
+++ b/source-bigquery-batch/driver.go
@@ -304,6 +304,14 @@ func (s *captureState) Validate() error {
 }
 
 func (c *capture) Run(ctx context.Context) error {
+	// With no enabled bindings there are no polling workers, so we'd fall through the
+	// empty worker group and exit silently. This connector normally runs forever, so
+	// report the reason.
+	if len(c.Bindings) == 0 {
+		log.Info("capture has no enabled bindings, shutting down")
+		return nil
+	}
+
 	// Always discover and output SourcedSchemas once at startup.
 	if err := c.emitSourcedSchemas(ctx); err != nil {
 		return err
--- a/source-db2-batch/driver.go
+++ b/source-db2-batch/driver.go
@@ -377,6 +377,14 @@ func (s *captureState) Validate() error {
 }
 
 func (c *capture) Run(ctx context.Context) error {
+	// With no enabled bindings there are no polling workers, so we'd fall through the
+	// empty worker group and exit silently. This connector normally runs forever, so
+	// report the reason.
+	if len(c.Bindings) == 0 {
+		log.Info("capture has no enabled bindings, shutting down")
+		return nil
+	}
+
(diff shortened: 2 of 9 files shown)


Alex-Bair (approved):
LGTM

PR #5005, opened 2026-08-05
source-mongodb: Fix dropped collection type validation in Pull

willdonnelly:
**Description:**

The collection type check in Pull didn't actually call `validate()` and instead was checking a stale `err` value from earlier which is provably nil (since the earlier error branch wasn't taken). This meant that error check was a no-op and presumably meant bad things for the handling of unrecognized collection types (if any of those exist?).

Fixed by copy-pasting the presumed-correct logic from the discovery path which calls `validate()` on the type and checks that error.

**Checklist:**

### DIFF
--- a/source-mongodb/CHANGELOG.md
+++ b/source-mongodb/CHANGELOG.md
@@ -1,9 +1,8 @@
-# source-mongodb
+# Changelog
 
-## v2, 2024-04-10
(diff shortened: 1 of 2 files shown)


Alex-Bair (approved):
LGTM, thanks!

PR #5006, opened 2026-08-05
source-azure-blob-storage: Add secret annotation to ConnectionString

willdonnelly:
**Description:**

The connection string embeds the storage account's full AccountKey, so it should be annotated as secret.

I believe this is trivially backwards-compatible, since runtime decryption collapses `"foobar"` and `"foobar_sops"` values together and isn't too picky about which form was present in the encrypted config or whether that lines up with the schema annotations.

**Workflow steps:**

The annotation will take effect going forward, the next time a task spec is published. Any currently-impacted specs should probably be republished.

**Checklist:**


nicolaslazo (approved):
LGTM

PR #5007, opened 2026-08-05
source-klaviyo: normalize space-separated datetimes in every stream

nicolaslazo:
**Description:**

Followup to https://github.com/estuary/connectors/pull/4742.

The earlier fix added the normalizer but called it from a single override on `Profiles`, so only `profiles` and `global_exclusions` were covered. Move the call onto the base `KlaviyoStream.map_record`, which every stream's `parse_response` routes through, and drop the now-redundant `Profiles` override. This mirrors the same fix on `source-klaviyo-native`, which hooks the shared base model for the same reason.

`Events.parse_response` overrides the base method and normalized its cursor by hand with an unconditional `str.replace(" ", "T")`, leaving its free-form `event_properties` alone. It now calls `map_record` instead. To keep that path's coverage, the UTC offset is optional in the regex; the old replace did not require one.

Two smaller corrections ride along, both in the ported normalizer:

- Anchor with `\Z` rather than `$`, so a value ending in a newline is no longer silently stripped of it.
- Normalize before the cursor is copied to the top level, so the copy carries the RFC3339 form. `SemiIncrementalKlaviyoStream` compares cursors as strings against state written as `isoformat()`, where a space sorts below the `T`.

**Workflow steps:**

**Notes for reviewers:**

**Checklist:**

### DIFF
--- a/source-klaviyo/CHANGELOG.md
+++ b/source-klaviyo/CHANGELOG.md
@@ -1,0 +1,8 @@
+# Changelog
+
+## 2026-08-05
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM

PR #5008, opened 2026-08-05
estuary-cdk: log when no bindings are enabled

Alex-Bair:
**Regression?**

No.

**Description:**

We've too often (i.e. more than once) spent time investigating why a capture is not moving data when the root cause is simply there are no bindings enabled. This log / connector status should make that situation more obvious, either when looking in the logs or looking at the Connector Status in the dashboard.

**Checklist:**

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
+++ b/estuary-cdk/estuary_cdk/capture/base_capture_connector.py
@@ -248,7 +248,12 @@ class BaseCaptureConnector(
                         self._transactor,
                         catalog_task_name=open.capture.name,
                     )
-                    log.event.status("Capture started")
+
+                    message = "Capture started."
+                    if len(open.capture.bindings) < 1:
+                        message += " No bindings are enabled."
+                    log.event.status(message)
+
                     await capture(task)
             except* TerminateTaskGroup:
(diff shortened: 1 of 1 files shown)


estuary-cdk/estuary_cdk/capture/base_capture_connector.py
```diff
@@ -248,7 +248,12 @@ async def enforce_shutdown(tg: asyncio.TaskGroup) -> None:
                         self._transactor,
                         catalog_task_name=open.capture.name,
                     )
-                    log.event.status("Capture started")
+
+                    message = "Capture started."
+                    if len(open.capture.bindings) < 1:
+                        message += " No bindings are enabled."
```
willdonnelly, line 254: Putting it in the startup connectorStatus message is a really good idea, I should rip this off and put it in the Go source-boilerplate "Capture started" logging as well.

willdonnelly (approved):
LGTM

estuary-cdk/estuary_cdk/capture/base_capture_connector.py
Alex-Bair: That's a good idea. I added an [issue](https://github.com/estuary/connectors/issues/5020) to our board so we remember to do that when we've got time.

Issue #5013, opened 2026-08-06
materialize-snowflake: unescaped table name in SHOW TABLES LIKE query can match wrong table


jwhartley:
## 1. Priority
Low - narrow trigger condition, no known customer occurrence, found via code audit while investigating #5012.

## 2. Scope and prevalence
None observed. Found during code review, not reported by any customer or seen in production logs. Swept `materialize-snowflake`, `source-snowflake`, and `materialize-sql` for the same pattern (an interpolated identifier used as a SQL `LIKE`/`STARTS WITH` argument without escaping `_`/`%`): this is the only occurrence. `snowflake.go:609` (`SHOW PIPES LIKE`) already escapes underscores in an interpolated pipe name before use (`strings.ReplaceAll(name, "_", "\\_")`), and `snowflake.go:1204` only substitutes a non-table hex range key into an otherwise-escaped literal pattern, so this file otherwise follows the escaping discipline that line 1355 skips.

## 3. Problem description
- Expected: `queryCurrentClusterBy` returns the clustering state of exactly the target table.
- Observed: `queryCurrentClusterBy` (`materialize-snowflake/snowflake.go:1355`) builds `SHOW TABLES LIKE '<table>' IN SCHEMA <schema>` by interpolating the raw table name directly into a SQL `LIKE` pattern, where `_` is Snowflake's single-character wildcard. It reads only the first row returned (`rows.Next()`) without verifying the returned name matches the target, so a same-schema table whose name differs only in an underscore position can match and supply its `cluster_by` state instead of the intended table's.
- Steps to reproduce: in one schema, create two tables differing only by an underscore vs. any other character in the same position (e.g. `orders_v1` and `ordersxv1`). Give only `ordersxv1` a `CLUSTER BY`. Call `queryCurrentClusterBy` for `orders_v1` - the `LIKE` pattern can match both names, and reading only the first row can return `ordersxv1`'s clustering state for the `orders_v1` query.

## 4. References
- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1785994466373499
- Related: #5012 (unrelated capability request, found while investigating it; both touch the same clustering-sync code path in `materialize-snowflake`, worth picking up together if #5012 is scheduled)

Issue #5012, opened 2026-08-06
materialize-snowflake: support expression-based clustering keys (not just literal columns)


jwhartley:
## 1. Priority
Low - feature request, single customer ask, existing workaround available (manual `ALTER TABLE ... CLUSTER BY` in Snowflake, though it does not survive a materialization restart today).

## 2. Scope and prevalence
One customer request so far (<tenant>, support ticket). No prior customer asks found for this specific capability via internal knowledge-source search.

## 3. Problem description
- Expected: the `clustering` resource config field on materialize-snowflake bindings accepts a Snowflake clustering key expression (e.g. `TRUNC(key_column, -5)`, `DATE_TRUNC('day', ts)`), matching Snowflake's own guidance for clustering high-cardinality columns.
- Observed: `ClusteringColumns()` splits the field on commas and resolves each entry as an exact, case-sensitive match against a materialized field name via `clusteringKeyExpr()`. No expressions or functions are supported today. There is no config-time validation; an unresolvable value fails either during `CreateTable` at publish (new table/binding) or on the next materialization restart (existing table).
- Customer's stated use case: cluster on a truncated/rounded form of a high-cardinality column to limit cardinality, which is the pattern Snowflake's own docs recommend for clustering key selection.

## 4. References
- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1785994466373499
- Snowflake docs, clustering key selection strategies (recommends expressions like `TRUNC()` for high-cardinality columns, max 3-4 columns/expressions per key): https://docs.snowflake.com/en/user-guide/tables-clustering-keys#strategies-for-selecting-clustering-keys
- Related prior art: #4379
- Docs gap tracked separately: `clustering` resource config field's current (literal-column-only) behavior is undocumented on docs.estuary.dev; a docs PR covering today's behavior is in progress independent of this feature decision

jwhartley:
Companion issue: #5013, found while investigating this one. Both touch the same clustering-sync code path in materialize-snowflake, worth picking up together if this ships.

Issue #4735, opened 2026-07-01
mysql: add IAM authentication support


jwhartley:
**Current Behavior**

The mysql connectors support username/password authentication only.

**Desired Behavior**

Add IAM authentication to the source-mysql and materialize-mysql connectors, ideally across all supported cloud providers, similar to the postgres connector (which uses the shared `go/auth/iam` package and swaps in a token at connect time via `rds/auth.BuildAuthToken`).

- [ ] MySQL
  - [x] source-mysql
    - [x] Azure
    - [x] GCP
    - [x] AWS IAM
  - [ ] materialize-mysql
    - [ ] Azure
    - [ ] GCP
    - [ ] AWS IAM

**Context / Priority**

Requested by a customer who is mandated to rotate all DB passwords within about a week and would prefer to migrate their MySQL captures to IAM rather than rotate passwords. AWS IAM is the immediate need (they run custom/BYOC data planes on AWS). Not a bug blocking existing users. Triage estimate is roughly one day of focused work.

Part of #3432.

nicolaslazo:
Capture side is done, `materialize-mysql` is left. Reopening

PR #5014, opened 2026-08-06
docs(materialize-snowflake): document clustering resource config

jwhartley:
**Description:**

Documents the `clustering` resource config property on materialize-snowflake bindings. This property has existed since April 2026 (per-binding Snowflake Automatic Clustering) but was never documented on docs.estuary.dev.

**Workflow steps:**

No behavior change, docs only. Adds the `/clustering` row to the bindings properties table and a new "Clustering keys" subsection under Performance considerations covering: literal-column-only support (no SQL expressions), that the key is applied/re-applied on every restart/republish/backfill (not just table creation), and that leaving the field blank causes the connector to drop any existing clustering key on the table, including one applied manually in Snowflake.

- `docs/reference/Connectors/materialization-connectors/Snowflake.md` (this PR)

**Notes for reviewers:**

Came out of a support ticket (Flock Freight, Pylon #4735) where the customer found this field with no docs. Verified against `materialize-snowflake/snowflake.go` at HEAD. Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1785994466373499

A related feature request (expression-based clustering keys, e.g. `TRUNC()`) is tracked separately in estuary/connectors#5012, and an unrelated bug found while reading this code path is in estuary/connectors#5013. This PR documents today's behavior only; will update if #5012 ships.

**Checklist:**

- [x] Docs-only change, no CHANGELOG entry needed.

### DIFF
--- a/docs/reference/Connectors/materialization-connectors/Snowflake.md
+++ b/docs/reference/Connectors/materialization-connectors/Snowflake.md
@@ -161,6 +161,7 @@ Use the below properties to configure a Snowflake materialization, which will di
 | **`/table`** | Table | Table name | string | Required |
 | `/schema` | Alternative Schema | Alternative schema for this table | string |  |
 | `/delta_updates` | Delta updates | Whether to use standard or [delta updates](#delta-updates) | boolean |  |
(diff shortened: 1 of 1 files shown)


aeluce (approved):
Are we getting pretty close to cutover or do you want to make this live via flow repo as well?

jwhartley:
we should be getting close to cutover yes

Issue #20, opened 2021-08-17
Kafka Source Connector


saterus:
We'd like to be able to consume messages from Kafka topics/partitions as an Airbyte-compatible connector.

williamhbaker:
Closing this since it is done!

PR #5015, opened 2026-08-06
docs: reconcile OpenSearch and materialize-hubspot pages missed by the first pass

jwhartley:
## Context

While spiking a Cloudflare Workers deploy target for the docs migration (#20 on estuary/docs), Mike Danko diffed the flow/site and staging sitemaps and found 7 production URLs missing from this repo's `docs/reference/Connectors/`. Checked each one against git history:

- 5 of the 7 (`amazon-sqs-native`, `smartsheet`, `zuora`, `amazon-sns`, `google-spanner`) already exist on `main` here, added between 2026-07-05 and 2026-07-22. The hourly staging sync (`estuary/docs` `deploy.yml`, confirmed via a live run today) correctly tracks `main`'s tip, so those should already be serving on staging without a 404 — the sitemap diff that flagged them was most likely comparing against a stale build.
- The other 2 landed on flow/site after the first reconciliation pass's 2026-07-20 cutoff and were genuinely never migrated. This PR closes that gap.

## Changes

- `materialization-connectors/hubspot.md` — new, the HubSpot materialization connector (distinct from the HubSpot capture connectors already migrated).
- `materialization-connectors/Elasticsearch.md` → `materialization-connectors/Elasticsearch/Elasticsearch.md` — relocated to match flow/site's restructure into a category folder, content refreshed from flow/site (a couple of small typo fixes), links kept in the existing absolute-path style already used elsewhere in this repo (e.g. the Salesforce docs) since that's what resolves correctly once aggregated into `estuary/docs`.
- `materialization-connectors/Elasticsearch/opensearch.md` — new, the OpenSearch materialization connector.
- `materialization-connectors/README.md` — added HubSpot and OpenSearch entries, updated the Elasticsearch link path.

## Verification

Built the full `estuary/docs` site locally against this branch (real `docusaurus build`, not just a syntax check):
- All three URLs resolve at their intended paths, `.../materialization-connectors/Elasticsearch/` (unchanged, no doubled segment from the folder move), `.../Elasticsearch/opensearch/`, and `.../materialization-connectors/hubspot/`.
- No broken links introduced (one pre-existing, unrelated broken-anchor warning on a reduction-strategies page).
- README-derived category index links to all three correctly.

### DIFF
--- a/docs/reference/Connectors/materialization-connectors/Elasticsearch/Elasticsearch.md
+++ b/docs/reference/Connectors/materialization-connectors/Elasticsearch/Elasticsearch.md
@@ -17,7 +17,7 @@ To use this connector, you'll need:
 - At least one Estuary collection
 
 :::tip
-If you haven't yet captured your data from its external source, start at the beginning of the [guide to create a dataflow](../../../guides/create-dataflow.md). You'll be referred back to this connector-specific documentation at the appropriate steps.
+If you haven't yet captured your data from its external source, start at the beginning of the [guide to create a dataflow](/guides/create-dataflow). You'll be referred back to this connector-specific documentation at the appropriate steps.
 :::
 
 ## Configuration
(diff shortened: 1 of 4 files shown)


aeluce (approved):
LGTM.

There have definitely been some other changes since then that will need to be reconciled as well (ex. I recently reworked webhook capture pages a bit in the flow repo and haven't migrated those changes here). Is there a specific cutoff on when we should stop merging connector updates to flow?

jwhartley:
> There have definitely been some other changes since then that will need to be reconciled as well (ex. I recently reworked webhook capture pages a bit in the flow repo and haven't migrated those changes here). Is there a specific cutoff on when we should stop merging connector updates to flow?

@aeluce yes once it's all ready I'll schedule migration, notify the team, and we can probably bake in a transition period where if you commit to the old place I'll yell at you and/or gracefully migrate it

PR #5022, opened 2026-08-06
materialize-clickhouse: use shared binding buffer for load batches

danielnelson:
Previously we would send a batch whenever the load binding changed. Load documents are received in timestamp order, unlike Store documents which are ordered by binding.  Because of this when a task had many interleaved bindings being updated, the binding changeover was too frequent and often containing just several items.

With this change documents across all binding are buffered together until the original limits of 100k docs or 10MiB is reached.  Then they are grouped by binding and, for simplicity, submitted one binding at a time.  In most use cases, this should result in better/larger sized batches without exceeding memory limits.

**Workflow steps:**

No changes

None

**Checklist:**

### DIFF
--- a/materialize-clickhouse/driver.go
+++ b/materialize-clickhouse/driver.go
@@ -1,12 +1,14 @@
 package connector
 
 import (
+	"cmp"
 	"context"
 	"crypto/tls"
 	_ "embed"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	chproto "github.com/ClickHouse/ch-go/proto"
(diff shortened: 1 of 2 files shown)


materialize-clickhouse/driver.go
```diff
@@ -579,9 +580,6 @@ func (t *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage)
 
 	for it.Next() {
 		if it.Binding != lastBinding {
-			if err = flushLastBinding(); err != nil {
-				return err
-			}
 			lastBinding = it.Binding
```
jgraettinger: nit: is `lastBinding` now vestigial / can be dropped?

materialize-clickhouse/driver.go
```diff
@@ -518,44 +522,41 @@ func (t *transactor) Load(it *m.LoadIterator, loaded func(int, json.RawMessage)
...
-	batch := make([][]any, 0, maxBatchSize)
+	batch := make([]batchItem, 0, maxBatchSize)
 	batchBytes := 0
 
-	flushLastBinding := func() error {
-		if len(batch) == 0 || lastBinding < 0 {
-			return nil
+	flushBindings := func() error {
+		// Group batch records by Binding.
+		bindingBatch := make(map[int][][]any, len(t.bindings))
```
jgraettinger: nit: this is a lot of allocation overhead as compared to just sorting `batch` on `Binding` and then walking its contiguous runs

jgraettinger (dismissed):

materialize-clickhouse/driver.go
danielnelson: It is still being used to gate the check of `activeBindings` to see if the binding is new.  But we aren't getting much for it because the binding can change so frequently, I think I'll remove.

jgraettinger (approved):

PR #5024, opened 2026-08-06
source-boilerplate: Log when no bindings are enabled

### LINKED ISSUE #5020: source-boilerplate: update connectorStatus to indicate when no bindings are enabled
See https://github.com/estuary/connectors/pull/5008#discussion_r3723020866.

[Originating Slack thread](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1785915450532239).

willdonnelly:
**Description:**

Going along with https://github.com/estuary/connectors/pull/5004, which added shutdown messages to various captures which shut down when there are zero bindings, and https://github.com/estuary/connectors/pull/5008 which added startup-time logging to the CDK, this adds similar startup logging to Go captures using `source-boilerplate`.

Since the way ready signals are emitted is a bit indirect here, this required adding a binding-count field to the `PullOutput` struct which the `Ready()` method consults, but I think that's fine.

### DIFF
--- a/source-boilerplate/boilerplate.go
+++ b/source-boilerplate/boilerplate.go
@@ -167,7 +167,10 @@ func (s *ConnectorServer) Capture(stream pc.Connector_CaptureServer) error {
 			}
 		case request.Open != nil:
 			log.WithField("eventType", "connectorStatus").Info("Starting capture")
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM, thanks!

PR #5029, opened 2026-08-07
materialize-bigquery: report NULL flow_document in load errors

### LINKED ISSUE #5028: materialize-bigquery: NULL flow_document load error reports the wrong value's type
## What happens

`bindingDocument.Load` in `materialize-bigquery/binding.go` asserts that the second value returned by the load query is a string, and on failure formats the type of `v[0]` rather than `v[1]`:

```go
if document, ok := v[1].(string); !ok {
    return fmt.Errorf("value[1] wrong type %T expecting string", v[0])
}
```

`v[0]` is the binding index, which was already asserted to be `int64` a few lines above, and the function returns early if that assertion fails. So by the time this line runs, `v[0]` is always `int64`.

The message is therefore a constant. Every occurrence, whatever the real cause, reads:

```
load row read: value[1] wrong type int64 expecting string
```

It carries no information about what the document value actually was.

## Why it matters

This is the error surfaced when a standard-updates binding is attached to a pre-existing table that holds rows the materialization did not write. The load query inner-joins on the merge key, so those rows match and come back with `flow_document` NULL. A SQL NULL arrives as an untyped `nil`, the string assertion fails, and the task fails.

With the correct index the message would read `value[1] wrong type <nil> expecting string`, and `<nil>` points directly at a NULL column value. Instead it reports `int64`, which sends the operator looking for a type mismatch: an INT64 column where a STRING was expected, a bad projection, schema drift. All dead ends. The one diagnostic that identifies the cause is the one the misprint suppresses.

The wrap at `transactor.go:322` adds `load row read:` and nothing else, so this reaches task logs verbatim.

## Reproduction

`Load` is directly testable, no BigQuery instance needed:

```go
var bd bindingDocument
err := bd.Load([]bigquery.Value{int64(3), nil}, nil)
// got:  value[1] wrong type int64 expecting string
// want: something identifying the value as NULL
```

Passing a genuinely wrong non-nil type shows the misprint just as clearly:

```go
err := bd.Load([]bigquery.Value{int64(3), 3.5}, nil)
// got:  value[1] wrong type int64 expecting string
// want: value[1] wrong type float64 expecting string
```

## Expected

Two changes:

1. Format `v[1]` in the type-mismatch message, so the reported type is the document's.
2. Check for `nil` before the assertion and return a message that says the value is NULL and names `Exclude Flow Document` (`no_flow_document`) as the remedy, since NULL is the dominant cause in practice and is otherwise expensive to diagnose.

Also worth considering, but not included in the linked PR: the same class of failure surfaces differently on other destinations. Postgres and Snowflake pass the nil through to the runtime, which fails at `crates/runtime/src/materialize/protocol.rs` with `couldn't parse loaded document as JSON`. That message does not mention NULL or `flow_document` either, so the diagnosis cost is similar, and a shared fix at the boilerplate level may be preferable to per-connector guards.

jwhartley:
Closes #5028

## The bug

`bindingDocument.Load` asserts `v[1].(string)` but formatted `v[0]` on failure:

```go
if document, ok := v[1].(string); !ok {
    return fmt.Errorf("value[1] wrong type %T expecting string", v[0])
```

`v[0]` is the binding index, already asserted to be `int64` a few lines above with an early return. So it was always `int64`, and the message was the constant `value[1] wrong type int64 expecting string` whatever the document value actually was.

## Why not just fix the index

The common way to hit this is a NULL `flow_document`: the column is added as nullable when a binding adopts a table that already held rows, nothing backfills it for those rows, and the load query's inner join returns them. Fixing only the index gives `wrong type <nil>`, which is honest but still needs decoding. So NULL gets its own branch, as an `errNullFlowDocument` sentinel naming the cause and the `Exclude Flow Document` option.

## Verification

```
go test ./materialize-bigquery/ -run TestBindingDocumentLoad
```

Two of the five cases fail against `main`: `document is NULL`, and `document wrong type`, which passes a `float64` and so pins the `%T` fix.

Also checked against live BigQuery through the connector's own read path, with the Storage Read API engaged (`IsAccelerated() == true`, which needs `Job.Read` rather than `Query.Read`): a NULL document returns the sentinel, a valid document passes through verbatim, a `float64` reports `float64`.

`go vet` is clean, and cannot catch this class of bug: `%T` accepts any argument, so the printf analyzer sees a well-formed call.

## Two things the message deliberately does not say

Both were in earlier drafts and both are wrong, so flagging them in case they look like omissions:

- **It does not tell you to recreate the binding.** The load template is selected from the endpoint config alone, per transaction (`transactor.go:270-273`), so `Exclude Flow Document` takes effect on an existing binding's next transaction.
- **It does not present a missing value in a required root-level column as a precondition.** Materializing those is enforced by the connector already (`materialize-sql/driver.go:235-238`), and a reconstructed null is discarded by the default LastWriteWins reduction rather than validated (`crates/doc/src/combine/memtable.rs:421-425`, `reduce/strategy.rs:282-284`). It only matters for collections that reduce the root document, so it lives in a code comment rather than in front of operators.

## Out of scope, worth knowing

- **A separate bug in `loadQueryNoFlowDocument`.** It aliases each field by `$col.Field` unquoted (`sqlgen.go:318`) where the rest of the file uses `$col.Identifier`. Confirmed against BigQuery: `AS array` fails with `Unexpected keyword ARRAY`, and the committed snapshot also contains `AS key!binary` and `AS nonAsciiκόσμε`, both syntax errors. Quoting fixes those; it does not fix the second half, which is that a renamed root-level projection reconstructs under the projection name rather than the schema property name. `materialize-mysql` has the same template. Happy to file separately.
- Postgres and Snowflake pass the nil through and the runtime fails with `couldn't parse loaded document as JSON`. `bindingDocument` is the only `ValueLoader` here and `grep "wrong type %T"` hits only this file, so the misprint is not copy-pasted, but the missing NULL diagnostic is effectively repo-wide. A boilerplate-level guard may be the better long-term fix.
- A JSON `null` inside a JSON-typed column is not a SQL NULL and still passes through. Not covered.
- No `VERSION` bump: error text only, and both branches return a non-nil error wrapped identically at `transactor.go`, so no caller control flow changes.

### DIFF
--- a/materialize-bigquery/binding.go
+++ b/materialize-bigquery/binding.go
@@ -33,7 +33,7 @@ type binding struct {
 // its reconstruction yields an explicit null for any root-level column that is
 // NULL, which only matters for collections that reduce the root document rather
 // than replacing it.
-var errNullFlowDocument = errors.New("value[1] is NULL: this row has no flow_document value, " +
+var errNullFlowDocument = errors.New("row has no flow_document value, " +
 	"typically because the binding adopted a table that already held data: see https://go.estuary.dev/matff. " +
 	"The 'Exclude Flow Document' advanced option makes loads reconstruct the document " +
 	"from the table's own columns instead")
--- a/materialize-bigquery/binding.go
+++ b/materialize-bigquery/binding.go
@@ -2,6 +2,7 @@ package connector
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"cloud.google.com/go/bigquery"
@@ -21,6 +22,22 @@ type binding struct {
 	storeMergeBounds *sql.MergeBoundsBuilder
 }
 
+// errNullFlowDocument is returned when a load query reads a row whose
+// flow_document column is NULL. The column is added as nullable when a binding
+// adopts a table that already held rows, and nothing backfills it for them.
+//
+// A sentinel so this stays distinguishable from a type mismatch without matching
+// on message text. Two caveats not worth putting in front of operators: enabling
+// no_flow_document takes effect on the next transaction of an existing binding
+// (transactor.go selects the load template from the endpoint config alone), and
+// its reconstruction yields an explicit null for any root-level column that is
+// NULL, which only matters for collections that reduce the root document rather
+// than replacing it.
+var errNullFlowDocument = errors.New("value[1] is NULL: this row has no flow_document value, " +
+	"typically because the binding adopted a table that already held data: see https://go.estuary.dev/matff. " +
+	"The 'Exclude Flow Document' advanced option makes loads reconstruct the document " +
+	"from the table's own columns instead")
+
 // bindingDocument is used by the load operation to fetch binding flow_document values
 type bindingDocument struct {
 	Binding  int
@@ -38,8 +55,10 @@ func (bd *bindingDocument) Load(v []bigquery.Value, s bigquery.Schema) error {
 	} else {
 		bd.Binding = int(binding)
 	}
-	if document, ok := v[1].(string); !ok {
-		return fmt.Errorf("value[1] wrong type %T expecting string", v[0])
+	if v[1] == nil {
+		return errNullFlowDocument
+	} else if document, ok := v[1].(string); !ok {
+		return fmt.Errorf("value[1] wrong type %T expecting string", v[1])
 	} else {
 		bd.Document = json.RawMessage(document)
 	}
--- a/materialize-bigquery/CHANGELOG.md
+++ b/materialize-bigquery/CHANGELOG.md
@@ -1,5 +1,14 @@
 # materialize-bigquery
 
+## 2026-08-07
+
+### Fixed
+- A load query that reads a NULL `flow_document` now says so, and names the
+  `Exclude Flow Document` option that addresses it. It previously reported
+  `value[1] wrong type int64 expecting string`, giving the type of the binding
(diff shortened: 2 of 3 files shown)


materialize-bigquery/binding.go
```diff
@@ -21,6 +22,22 @@ type binding struct {
...
+// adopts a table that already held rows, and nothing backfills it for them.
+//
+// A sentinel so this stays distinguishable from a type mismatch without matching
+// on message text. Two caveats not worth putting in front of operators: enabling
+// no_flow_document takes effect on the next transaction of an existing binding
+// (transactor.go selects the load template from the endpoint config alone), and
+// its reconstruction yields an explicit null for any root-level column that is
+// NULL, which only matters for collections that reduce the root document rather
+// than replacing it.
+var errNullFlowDocument = errors.New("value[1] is NULL: this row has no flow_document value, " +
```
danielnelson: As a reader of the logs, I won't know what `value[1]` is unless I go look up the code.  I would drop before the `:` and start with `row has no flow_document value`.

danielnelson (dismissed):

materialize-bigquery/binding.go
jwhartley: Dropped the prefix, the message now opens with `row has no flow_document value`. Sentinel identity is what callers match, so nothing else changed.

danielnelson (approved):

PR #5037, opened 2026-08-10
materialize-motherduck: use rename migration path for DuckLake destinations

### LINKED ISSUE #5002: materialize-motherduck: destination rejects in-place column migration, task fails on every recovery
## 1. Priority
Medium - single tenant, full materialization pipeline down twice (4 days, then 9 hours). Customer applied a manual workaround (dropping the affected columns) both times; no preventive workaround exists.

## 2. Scope and prevalence
Two occurrences observed on one tenant's `materialize-motherduck` task, materializing into a MotherDuck-hosted DuckLake destination: first incident 2026-07-24 to 2026-07-28 (4 days), second 2026-08-03 06:11-15:04 UTC (9 hours, 46 shard failures at roughly 11-12 minute intervals). Same task, two different columns, two different rendered migrations.

## 3. Problem description
- Expected: an in-place column type migration the connector classifies as compatible either succeeds against the destination, or its failure is handled so the task recovers.
- Observed: the connector renders an `ALTER TABLE ... ALTER COLUMN <col> TYPE <type> USING <expr>` statement, and the destination rejects it with a catalog error. The task then fails identically on every recovery attempt with no self-heal path.

  Incident 1 (2026-07-24, customer-reported, outside our log retention window):
  ```
  ALTER TABLE <schema>.<table> ALTER COLUMN "<col>" TYPE JSON USING to_json("<col>")
  ```

  Incident 2 (2026-08-03, verified in task logs):
  ```
  ALTER TABLE <schema>.<table> ALTER COLUMN "<col>" TYPE VARCHAR USING CAST("<col>" AS VARCHAR)
  ```
  Destination response (verbatim):
  ```
  Cannot change type of column <col> from HUGEINT to VARCHAR - only widening type promotions are allowed
  ```
  Failure as it surfaces to the task:
  ```
  completeRecovery: store.RestoreCheckpoint: executing apply actions: Catalog Error: Cannot change type of column <col> from HUGEINT to VARCHAR - only widening type promotions are allowed
  ```
  The failure occurs during checkpoint restore, so the whole task fails before any binding reads, not only the binding with the affected column.

- Steps to reproduce: materialize a MongoDB-sourced field via `materialize-motherduck` into a MotherDuck-hosted DuckLake destination. Have schema inference change the field's inferred type (e.g. integer to string) so the connector renders an in-place `ALTER COLUMN ... TYPE ... USING ...` against the existing column. The destination rejects the DDL and the task fails on every subsequent recovery attempt.

## 4. References
- Related prior art: https://estuary-dev.slack.com/archives/C06R0CVD0KZ/p1781412230695139 (different connector, `materialize-amazon-aurora-postgres`, same trigger pattern)
- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1785920329896319

jacobmarble:
Closes #5002 

**Regression?**

Yes — #4603 (`ee76eb453`) switched this connector's column type migrations to a single
in-place `ALTER TABLE ... ALTER COLUMN ... TYPE ... USING`.

**Description:**

A DuckLake destination rejects that statement for anything but a lossless widening
promotion (`Cannot change type of column ... only widening type promotions are
allowed`), because DuckLake tracks columns by field id and never rewrites existing
data files. The change is classified compatible, so it is applied during shard
recovery — the task then fails on every restart with no path to self-heal.

`NewEndpoint` now determines the destination format before building the dialect, and
the dialect withholds `sql.WithDirectCast()` from its migration specs when the
destination is DuckLake, leaving those migrations on the multi-step rename path that
both formats support. Classic MotherDuck keeps the in-place cast. `client.go` is
unchanged — the specs alone decide which path `sql.StdColumnTypeMigrations` renders.

The format is queried, since both formats are served by the same `md:` endpoint and
the configuration carries no signal of which is which:
`select type from md_information_schema.databases where name = ?` → `DUCKLAKE` vs
`DEFAULT`. To keep that cheap, the bare connection is split out of `config.db` into
`config.open`, so detection skips the extension and staging-secret setup.

Also adds a `ducklake` integration variant, since CI previously exercised classic
MotherDuck only and could not see this failure mode at all.

**Workflow steps:**

No user-facing change beyond the fix. A task materializing into a DuckLake database
now survives a field type change instead of crash-looping.

None — no config or workflow surface changes. `docs-check-skip` applied.

**Notes for reviewers:**

- `TestIntegration` is now `{fence, motherduck/{materialize,apply,migrate},
  ducklake/{materialize,apply,migrate}}`, variants in parallel against separate
  databases (`ci_testing`, `ci_testing_ducklake`).
- `fence` runs against the classic config only: its goldens live in the
  `materialize-sql/.snapshots` set shared by every SQL materialization connector and
  keyed by test name, so a per-variant run would fork connector-specific copies.
  DuckLake checkpoint DDL/DML is still covered via `ducklake/materialize`.
- The two migrate snapshots show the fork: 89 `ALTER COLUMN ... TYPE` for
  motherduck, 89 add/copy/drop/rename sequences for ducklake, converging on
  byte-identical final column state. `TestIntegration-motherduck-migrate` is
  unchanged, confirming classic behavior is preserved.
- Only `DirectCast` varies by format. The `WithCastSQL` overrides stay in both, since
  the rename path's `UPDATE ... SET tmp = <cast>` consumes the same expression the
  direct cast does — and the default `CAST(x AS type)` would be wrong for all four
  (RFC3339 datetimes, base64 <-> blob, `to_json` of non-JSON text).
- `ErrNoRows` from the format query (a database attached from outside the account)
  falls back to the rename path, which works on either format.
- `MigratableTypes` targets are identical for both formats, so validation and
  compatible-vs-incompatible classification are unaffected.

**Checklist:**


materialize-motherduck/testdata/config.ducklake.s3.yaml
```diff
@@ -0,0 +1,23 @@
+database: ci_testing_ducklake
```
danielnelson, line 1: The difference when it comes to setup from regular Motherduck is that this database was pre-created with `TYPE DUCKLAKE`?

danielnelson (dismissed):

materialize-motherduck/testdata/config.ducklake.s3.yaml
jacobmarble: Correct, this database was created manually with `TYPE DUCKLAKE`.

jacobmarble: I've added a check to the integration tests to ensure that the expected database type is set.

danielnelson (approved):

PR #5039, opened 2026-08-10
source-google-ads: read geo_target_constant once per sync

Alex-Bair:
**Regression?**

No.

**Description:**

`geo_target_constant` is a global table with no customer dimension, but it was sliced per customer like every other full refresh stream. A capture with N customer IDs re-read the same ~270k rows N times every sync, at N times the API quota and collection volume. Emit a single slice instead so `geo_target_constant` is only read a single time per sync.

**Checklist:**

### DIFF
--- a/source-google-ads/CHANGELOG.md
+++ b/source-google-ads/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2026-08-10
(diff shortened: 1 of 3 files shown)


willdonnelly (approved):
LGTM

PR #5040, opened 2026-08-10
ci: Test Go connector helper packages

willdonnelly:
**Description:**

We have a bunch of helper packages which various connectors depend on, and in theory those helper packages have tests, but in practice those tests weren't getting run in CI. See for example the `sqlcapture` test case bitrot also fixed in this PR.

It looks like at present all of the other tests were still passing, but to avoid future breakage not getting noticed it seems like it would be a good idea for connector CI testing to include the unit tests for the various helper packages they rely on.

**Checklist:**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -212,6 +212,27 @@ jobs:
         run: |
           bash source-mongodb/docker-compose-init.sh
 
+      # Run the tests of every in-repo helper package which this connector
+      # transitively depends on. This list is computed on the fly so that we
+      # don't have to keep an explicit list up-to-date. Likewise instead of
+      # having an explicit `contains()` clause governing Go connectors we'll
+      # just identify them as connectors whose subdirectory contains Go files.
+      - name: Test Go helper packages for ${{ matrix.connector }}
+        if: hashFiles(format('{0}/*.go', matrix.connector)) != ''
+        env:
+          GIT_LFS_SKIP_SMUDGE: 1 # See 'Database Tests' step comment
+        run: |
+          set -euo pipefail
+          DEPS=$(go list -deps -test \
+            -f '{{if and (not .ForTest) .Module (eq .Module.Path "github.com/estuary/connectors")}}{{.ImportPath}}{{end}}' \
+            './${{ matrix.connector }}/...')
+          PKGS=$(echo "$DEPS" \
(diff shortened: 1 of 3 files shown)


source-mongodb/Dockerfile
```diff
@@ -27,6 +27,10 @@ COPY source-mongodb     ./source-mongodb
 ARG TEST_DATABASE=yes
 ENV TEST_DATABASE=$TEST_DATABASE
 COPY --from=rust-builder /builder/bson-transcoder/target/release/bson-transcoder /usr/bin/bson-transcoder
+RUN go test -short -v \
+    ./go/materialize/... \
```
Alex-Bair: note for myself: It felt a little odd seeing tests from `./go/materialize` being run for the capture connector, but `source-mongodb/change_stream.go` does indeed use the `RunAsyncOperation` function from `./go/materialize`, so it makes sense to run those tests. I wonder if we should move the functionality `source-mongodb` needs from `./go/materialize` into a more generic namespace?

Alex-Bair (dismissed):
LGTM.

Running tests for helper packages anywhere in CI is definitely better than what we're doing right now, but I'm wondering if manually keeping which helper packages are run for which connectors in sync will be a pain? (It already looks like there's some gaps - a number of filesource connectors like `source-gcs`, `source-dropbox`, `source-azure-blob-storage` aren't running the `go/schema-gen` tests, no materializations are running tests for `go/` helper packages, etc.). Maybe we could run all helpers' tests once before building any connectors, determine which helpers failed, then run `go list -deps` for each connector to determine if any of the helpers it depends on had failing tests & that connector's CI job should fail?

willdonnelly:
>Maybe we could run all helpers' tests once before building any connectors, determine which helpers failed, then run go list -deps for each connector to determine if any of the helpers it depends on had failing tests & that connector's CI job should fail?

That's an interesting idea. My concern was that any sort of "just run everything all the time" gate would cause collateral damage where a broken helper package used by a single connector could render us unable to publish fixes to a completely unrelated connector (and since CI push-on-green is our usual path for pushing out urgent fixes anything that could block that for unrelated reasons is Very Bad in my opinion). But I quite like the idea of trying to make it automatic based on what a given connector actually depends on rather than using explicit lists, I'll take a stab at that.

Alex-Bair (approved):
LGTM, thanks!!

PR #5054, opened 2026-08-12
ci: retry transient failures when fetching Flow binaries

Alex-Bair:
**Regression?**

No

**Description:**

GitHub's release download endpoint returns 503 under concurrent load. A recent connectors CI run lost 10 of 67 jobs to it, each rejected ~100ms into the step before any bytes transferred -- the shape of edge throttling rather than an outage, with 67 jobs pulling the same 243MB asset within seconds of each other.

This commit adds `--retry 5` to all five downloads, which covers 503 along with 429, the other 5xx codes, and timeouts.

The legacy package download also has to stop piping into tar. `--retry` restarts the transfer from the beginning, so retrying mid-stream would feed tar a second copy of the header and corrupt the extract rather than recover from it. Download it to a file first, then extract. The file lands in `flow-bin/`, which this script already wipes on entry and which is gitignored, so a failed run leaves nothing behind.

**Checklist:**

### DIFF
--- a/fetch-flow.sh
+++ b/fetch-flow.sh
@@ -9,12 +9,16 @@ mkdir -p flow-bin &&
 rm -f flow-bin/* &&
 cd flow-bin &&
 
-# Get the older package for etcd and gazette.
-curl -L --proto '=https' --tlsv1.2 -sSf "https://github.com/estuary/flow/releases/download/dev/flow-x86-linux.tar.gz" | tar -zx &&
+# Get the older package for etcd and gazette. Downloaded to a file rather than
+# piped into tar: GitHub's release endpoint returns transient 503s under load,
+# and `--retry` restarts the transfer, which would corrupt an in-progress extract.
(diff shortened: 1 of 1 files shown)


danielnelson (approved):
I wonder if it would help if we uploaded an artifact in build_base_image and download it in the matrix jobs?

willdonnelly:
If the issue is simultaneous requests would adding a small random jitter potentially help here as well?

Alex-Bair:
> I wonder if it would help if we uploaded an artifact in build_base_image and download it in the matrix jobs?

That's a good idea, I think that will help. That'd minimize how many times we're making requests & downloading the binaries from GH. I'll work on doing that in a separate PR.

> If the issue is simultaneous requests would adding a small random jitter potentially help here as well?

I think a small jitter would help too, but once the artifact approach Daniel mentioned is implemented, we'll be making a lot fewer requests & I'm optimistic that'll reduce how often we receive 503 & friends from GH enough that a small jitter won't be needed.

PR #5056, opened 2026-08-12
estuary-cdk: don't crash when an HTTP error body isn't UTF-8

### LINKED ISSUE #5050: estuary-cdk: HTTP error handler raises UnicodeDecodeError when an error body isn't UTF-8
### What we saw

```
File "/opt/estuary-cdk/estuary_cdk/http.py", line 225, in request_stream
  return await self._request_stream(
File "/opt/estuary-cdk/estuary_cdk/http.py", line 664, in _request_stream
  f"Encountered HTTP error status {resp.status} which cannot be retried.\nURL: ...\nResponse:\n{body.decode('utf-8')}",
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 1: invalid start byte
```

All CDK frames in the full traceback (`task.py:409`, `capture/common.py:925`, `capture/common.py:1057`, `http.py:225`, `http.py:664`) match current `main` line-for-line, so this doesn't look version-specific.

### What seems to be going on

It looks like the failure is in the error-reporting path rather than the request itself — the `decode` sits inside the argument being passed to `HTTPError`, so it appears to run before that exception is ever constructed. A similar `decode` shape appears at lines 653 and 658 as well as 664; 653 looks like it's on the retry path, which may mean a retryable 5xx can take a task down.

Two consequences we observed:

- The real HTTP status never reaches the logs. We could not tell what the upstream actually objected to.
- Because the typed `HTTPError` is never constructed, connector-side `except HTTPError` / `e.code` handling appears to be bypassed entirely. The same connector handles other 4xx responses from this API cleanly, so this seems specific to whatever was different about this response body.

`0x8b` at position 1 is consistent with gzip magic (`1f 8b`), and the customer's own connector carries a decompress helper for this API, which suggests it returns compressed payloads that aren't always labelled as such. We have not captured the actual bytes, so that part is inference.

### Impact

This is in the shared CDK, so the exposure isn't limited to this connector — any Python connector talking to an upstream that can return a non-UTF-8 error body looks affected. Failure mode is a hard task crash with the diagnostic information destroyed.

### What we don't know

- Whether the body was genuinely gzip, and if so why it arrived undecompressed.
- Whether this has been silently killing other connectors — it wouldn't be obvious in logs, since the useful part of the error is what gets lost.

Alex-Bair:
**Regression?**

No.

**Description:**

The 4xx and 5xx paths in `estuary-cdk/estuary_cdk/http.py` decoded bodies with a strict `body.decode("utf-8")` inside the argument to `HTTPError`, so a non-UTF-8 body raised `UnicodeDecodeError` before the exception existed. The status never reached the logs, and `except HTTPError` handling was bypassed. The same decode sits in the 5xx "will retry" log line, so a retryable 503 with a binary body killed the task instead of backing off.

This PR fixes that by escaping undecodable bytes rather than raising on them, and errors report `Content-Type` so a body that arrived compressed without saying so is identifiable.

Fixes #5050.

**Checklist:**

### DIFF
--- a/estuary-cdk/estuary_cdk/http.py
+++ b/estuary-cdk/estuary_cdk/http.py
@@ -82,6 +82,17 @@ class HTTPError(RuntimeError):
         self.message = message
 
 
+def format_error_body(body: bytes) -> str:
+    """
+    Decode a response body for an error message or log field.
+
+    Error bodies aren't guaranteed to be UTF-8, and raising here would destroy
+    the error being reported, so undecodable bytes are escaped rather than
+    dropped or replaced.
+    """
+    return body.decode("utf-8", errors="backslashreplace")
+
+
 class HTTPSession(abc.ABC):
     """
     HTTPSession is an abstract base class for an HTTP client implementation.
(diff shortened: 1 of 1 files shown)


nicolaslazo (approved):
LGTM. 

I suspect this is not the final iteration of this solution, as `aiohttp` is supposed to decompress response bodies regardless of status code — `source-smartsheet` might be reporting `Content-Type: gzip` under certain circumstances. I'll monitor our captures to see where things are going off track

Alex-Bair:
Yeah, I suspect the `Content-Type` response header in some circumstances doesn't indicate the body is compressed even though it is, and `aiohttp` ends up receiving no signal telling it to decompress as a result.

PR #5057, opened 2026-08-13
source-mongodb: add `skipBackfills` option

nicolaslazo:
**Regression?**

**Description:**

Adds the skipBackfills option present in other database connectors.

Colons are used as database/collection delimiters despite being valid characters to follow the standard established with `/advanced/excludeCollections`. `*:*` is used as a magic value to specify that all backfills should be skipped.

**Workflow steps:**

**Notes for reviewers:**

**Checklist:**

### DIFF
--- a/source-mongodb/main_test.go
+++ b/source-mongodb/main_test.go
@@ -138,10 +138,10 @@ func TestCaptureMixedIdTypeBackfillResumption(t *testing.T) {
 		oid, err := primitive.ObjectIDFromHex(fmt.Sprintf("%024x", idx))
 		require.NoError(t, err)
 		for _, id := range []any{
-			idx,                                         // number
-			fmt.Sprintf("str-%04d", idx),                // string
+			idx,                          // number
+			fmt.Sprintf("str-%04d", idx), // string
 			primitive.Binary{Subtype: 0x04, Data: uuid}, // binData (UUID)
-			oid,                                         // objectId
+			oid, // objectId
 		} {
 			_, err := col.InsertOne(ctx, bson.M{"_id": id, "value": idx})
 			require.NoError(t, err)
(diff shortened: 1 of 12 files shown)


source-mongodb/main_test.go
```diff
@@ -474,6 +474,110 @@ func TestCaptureExclusiveCollectionFilter(t *testing.T) {
...
+	ctx := context.Background()
+	client, cfg := testClient(t)
+
+	cleanup := func() {
+		require.NoError(t, client.Database(database).Drop(ctx))
+	}
+	cleanup()
+	t.Cleanup(cleanup)
+
+	cfg.Advanced.SkipBackfills = skipBackfillsId(database, skippedCol)
```
Alex-Bair: nit: I suggest writing `cfg.Advanced.SkipBackfills` as a literal string in `TestCaptureSkipBackfills` and `TestCaptureSkipBackfillsMidBackfill` rather than building it with `skipBackfillsId`. As written, the test asks the connector to generate the config string and then checks that the connector agrees with itself, so a change to the namespace format would leave both tests green even though it would change the behavior for existing captures.

The literal values in `config_test.go` do already pin the format, so this isn't a coverage gap. It'd just be nice for the integration tests to be independently meaningful, and to read as the config string a user would actually type.

Alex-Bair: Additionally, if you do end up removing the use of the `skipBackfillsId` in `main_test.go`, you may not need that helper any longer & can inline the formatting it does in the only place it's called in the `shouldBackfill` helper.

Alex-Bair (dismissed):
LGTM % an optional nit around tests.

Alex-Bair (approved):
LGTM!

PR #5062, opened 2026-08-13
source-{python}: use `SnapshotResource` and regenerate YAML schema files

Alex-Bair:
**Regression?**

No.

**Description:**

We've stumbled over the "tombstone doesn't comply with the collection's write schema" pitfall recently, and I'm working on updating the CDK to hopefully prevent that from happening in the future. In order to make those CDK level changes though, all of our connectors need updated to use write schema compliant tombstones. This PR does all that ground work to fix existing connectors' tombstones mostly by simply using the `SnapshotResource` convenience class that ✨ just works ✨ for the majority of full refresh stream use cases.

I also regenerated the discovered YAML schema files we use for capture snapshot tests as well.

Snapshot changes are all expected due to the removal of non-key fields from the snapshot resources' schemas. Most changes are those non-key fields disappearing from the discovery snapshot. In the `source-smartsheet` case, regeneration of the YAML file caused the ordering of documents to be changed due to the removal of the `id` field from the key.

**Notes for reviewers**

This PR doesn't change the key for any collections. It just fixes the possible tombstone triggered JSON schema violations for the remaining CDK based connectors.

**Checklist:**

### DIFF
--- a/source-ada/source_ada/resources.py
+++ b/source-ada/source_ada/resources.py
@@ -4,7 +4,7 @@ from logging import Logger
 
 from estuary_cdk.flow import CaptureBinding, ValidationError
 from estuary_cdk.capture import Task
-from estuary_cdk.capture.common import BaseDocument, Resource, open_binding
+from estuary_cdk.capture.common import BaseDocument, Resource, SnapshotResource, open_binding
 from estuary_cdk.http import HTTPMixin, TokenSource, HTTPError
 
 from .models import (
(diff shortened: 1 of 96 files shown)


nicolaslazo:
I was thinking we could let `SnapshotResource` handle the default fields while we're at it... I don't like pushing to other people's branches without their permission, but in this case a sequence of fixup commits is easier to manage than 60+ edit suggestion comments. Let me know what you think!

Alex-Bair:
Yeah, those look good to me! Thanks!

nicolaslazo (approved):
🚀

PR #5063, opened 2026-08-13
estuary-cdk: add validation around `SnapshotResource`s and tombstones

### LINKED ISSUE #5051: estuary-cdk: snapshot-binding deletion records fail schema validation when the collection key isn't `/_meta/row_id`
### What we saw

```
captured collection acmeCo/foo/bar/reports (binding 2)
document failed validation against its collection JSON Schema

Caused by:
  detail: "Missing required property: id"
  document: {"_meta": {"op": "d", "row_id": 304, "uuid": "..."}}
```

The `reports` collection is keyed on `/id`, with `id` in `required`.

Seen in the same task as #5050.

### What seems to be going on

That document doesn't look like connector output — it has the shape the CDK appears to construct for snapshot deletions, and we couldn't find anything else that would emit `_meta.op = "d"` alongside a `_meta.row_id`. `SnapshotResource` in `capture/common.py` states that snapshot bindings need to be keyed on `/_meta/row_id`, which would make a `_meta`-only document valid — but this collection isn't keyed that way, so it seems the generated record can't satisfy its own schema.

What stands out is that nothing rejected this at publish time. There appears to be a key assertion around `common.py:937`, but as far as we can tell it only runs on the multi-subtask path; a single snapshot binding with an incompatible key seems to validate cleanly and then fail in production.

The timing supports that reading. The task ran from 20:13 without complaint and only started failing at 00:38. Deletion records appear to be emitted only when a pass returns fewer rows than the previous one, so the problem seems to stay invisible until the source count first shrinks — which could be hours or days after a bad configuration ships.

### Impact

Any snapshot binding whose collection isn't keyed on `/_meta/row_id` looks exposed. Deletions never apply, and the shard fails rather than degrading. The delayed onset makes it hard to connect back to the spec that caused it.

We also don't know how the customer ended up in this state — whether the key requirement is discoverable enough, or whether tooling led them there.

### What we don't know

- Whether other bindings/connectors are in the same state and simply haven't hit a shrinking pass yet. We're checking one more binding on this task.
- Whether the requirement is documented anywhere the customer would have seen.

Alex-Bair:
**Regression?**

No.

**Description:**

This PR adds additional validation around tombstones and snapshot resources to the CDK to catch common mistakes. These new validations include:
- `SnapshotResource.__post_init__` assert its own key includes `/_meta/row_id`.
- `resolve_bindings` reports a snapshot binding's collection key that doesn't contain `/_meta/row_id` as a `ValidationError`.
- `open_binding` asserts the tombstone carries every key field the CDK does not fill itself.

Closes #5051.

**Notes for reviewers:**

I'd like to also validate that tombstones satisfy the schema generated by a resource's `model`, but that's currently tricky / touches a lot of code since resources currently don't know what the tombstones are. I plan to add that validation in a future PR to keep this stack's scope to a reasonable level.

**Checklist:**

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/estuary-cdk/estuary_cdk/capture/common.py
+++ b/estuary-cdk/estuary_cdk/capture/common.py
@@ -782,6 +782,26 @@ def open(
     return (response.Opened(explicitAcknowledgements=requires_explicit_acks), _run)
 
 
+def _assert_tombstone_covers_key(tombstone: BaseDocument, key: list[str]) -> None:
+    """Assert every collection-key field other than the CDK-managed
+    /_meta/row_id is set on the tombstone.
+
+    A tombstone is sparse by design, but the CDK only fills in /_meta/row_id, so
+    every other key field has to be pre-set by the caller."""
+    dump = tombstone.model_dump(by_alias=True, exclude_unset=True)
+    for key_pointer in key:
+        if key_pointer == "/_meta/row_id":
+            continue
+        node: Any = dump
+        for field in key_pointer.strip("/").split("/"):
+            assert isinstance(node, dict) and field in node, (
+                f"A snapshot tombstone must have every collection key field "
+                f"other than /_meta/row_id pre-set, but {key_pointer} is unset. "
+                f"Key: {key}, tombstone: {dump}"
+            )
+            node = node[field]
+
+
 def _tombstone_discriminator_values(
     tombstone: BaseDocument, key: list[str]
 ) -> tuple[object, ...]:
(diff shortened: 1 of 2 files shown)


nicolaslazo (dismissed):
LGTM. Can't wait to not step on this footgun anymore

Alex-Bair:
It looks like merging #5062 (the first PR in this stack) automatically triggered a rebase on this PR & dismissed your approval... It's a little annoying merging the first PR caused your approval to get dismissed on this one. I'll look into how I can avoid that on future stacked PRs.

nicolaslazo (approved):
👌

PR #5065, opened 2026-08-13
source-twilio: stop filtering re-fetched Messages records against the…

jonwihl:
… date_sent cursor

The Messages stream's lookback window re-reads a trailing DateSent range to capture in-place updates, since the Twilio API cannot filter on date_updated. However, read_records then filtered records against the persisted cursor (max date_sent from the prior sweep) before yielding: records with both date_sent and date_updated older than that cursor were silently dropped. A re-fetched record typically satisfies exactly that condition, so the updates the lookback window exists to capture (e.g. price populated shortly after send) never reached the collection, and any message missed on first read was dropped permanently on every re-fetch.

Observed in production as ~25% of incrementally-captured messages retaining price: null, plus a smaller set of messages absent entirely.

Changes:
- Yield every record returned within the queried DateSent window. Re-emitting previously seen documents is safe since collections reduce on the key.
- Track the cursor as a running max against the persisted state instead of buffering and sorting the whole slice, so records stream as they are paged.
- Tolerate null date_sent (queued/failed messages): yield the record without advancing the cursor rather than crashing the sweep, which only checkpoints after every stream completes.

**Regression?**

**Description:**

**Workflow steps:**

**Notes for reviewers:**

**Checklist:**

### DIFF
--- a/source-twilio/source_twilio/streams.py
+++ b/source-twilio/source_twilio/streams.py
@@ -626,18 +626,33 @@ class Messages(IncrementalTwilioStreamWithLookbackWindow, TwilioNestedStream):
         # date_sent and date_updated older than the max date_sent observed in the previous sweep, and
         # filtering on the cursor silently discards exactly the updates the lookback window exists to
         # capture. Re-emitting previously seen documents is safe since collections reduce on the key.
+        #
+        # The API returns messages in reverse chronological order, and IncrementalTwilioStream
+        # checkpoints intra-slice every `state_checkpoint_interval` records. The cursor therefore
+        # must not advance while records are still streaming: the first record seen carries the
+        # newest date_sent in the slice, and an intra-slice checkpoint taken after it would persist
+        # a cursor beyond the older records still being processed, permanently skipping them if the
+        # connector crashes or restarts mid-slice. Instead, track the max date_sent locally and only
+        # advance the cursor once the slice has been read to completion. Checkpoints taken mid-slice
+        # carry the prior cursor, so a restart re-reads the slice from the beginning, which is safe
+        # (at-least-once) since collections reduce on the key.
+        max_cursor_value = None
         for record in super(IncrementalTwilioStream, self).read_records(sync_mode, cursor_field, stream_slice, stream_state):
             # A message that has not been sent yet (or failed to send) can have a null date_sent.
-            # Normalize timestamps when present, yield the record regardless, and only advance the
-            # cursor from non-null date_sent values so a single anomalous record can't fail the
-            # entire sweep (this connector only checkpoints after reading every stream).
+            # Normalize timestamps when present, yield the record regardless, and only consider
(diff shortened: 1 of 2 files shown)


source-twilio/source_twilio/streams.py
```diff
@@ -616,24 +616,28 @@ def read_records(
...
-            if record[self.cursor_field] >= initial_cursor:
-                self._cursor_value = record[self.cursor_field]
-                yield record
-            # Otherwise if it's an update to a record we've seen before, yield it.
-            elif record["date_updated"] >= initial_cursor:
-                yield record
+            # A message that has not been sent yet (or failed to send) can have a null date_sent.
+            # Normalize timestamps when present, yield the record regardless, and only advance the
+            # cursor from non-null date_sent values so a single anomalous record can't fail the
+            # entire sweep (this connector only checkpoints after reading every stream).
```
Alex-Bair: > (this connector only checkpoints after reading every stream)

This isn't correct. The `Messages` stream checkpoints every 1,000 records per the `state_checkpoint_interval` set on `IncrementalTwilioStream`:
https://github.com/estuary/connectors/blob/b7fd0e7b4f64b3a8aff627025a86bb369726cae1/source-twilio/source_twilio/streams.py#L143

source-twilio/source_twilio/streams.py
```diff
@@ -616,24 +616,28 @@ def read_records(
...
+            if record.get(self.cursor_field):
+                record[self.cursor_field] = pendulum.parse(record[self.cursor_field], strict=False).to_iso8601_string()
+                if self._cursor_value is None or record[self.cursor_field] > self._cursor_value:
+                    self._cursor_value = record[self.cursor_field]
```
Alex-Bair: Double check this, but I'm 95% sure the Twilio API returns records in reverse chronological order, i.e. the newest records first. That means that the first record we see is always going to have the highest possible cursor value, and combined with the `state_checkpoint_interval` triggering intra-slice checkpoints, we'll end up checkpointing a very recent cursor before we finished processing the older records. That why the current code sorts all records first before iterating over them in chronological order & setting the cursor during that iteration.

I'd suggest either still sorting them in memory like we're currently doing, or only setting the cursor to the max cursor seen after processing the entire slice. The former wouldn't require many code changes, but the latter would be better memory-wise.

Alex-Bair (commented):
Had some feedback that should be addressed since the proposed logic could cause the connector to miss data & skip the cursor forward if it crashes / gets restarted in the middle of processing `Messages`.

For awareness, these checkpoint mechanics and semantics are one of the reasons we _highly_ prefer working on native connectors instead of imported ones like `source-twilio`; we have a lot more flexibility and control over what gets checkpointed & when those checkpoints happen when working on native CDK based connectors.

Alex-Bair (approved):
LGTM

PR #5068, opened 2026-08-15
source-*-batch: fail fast when INSERTs fail during tests

Alex-Bair:
**Regression?**

No.

**Description:**

A [CI job](https://github.com/estuary/connectors/actions/runs/31748444778/job/94608596049) for `source-redshift-batch` ran for 6 hours before GitHub's default job timeout cancelled it. The job normally takes 7–9 minutes.

For the spinning job, the last 18,121 log lines repeat the same three lines about once per second:
```
  driver.go:474: executing query ( query="SELECT * FROM "public"."basiccapture_884124" WHERE ..." )
  main_test.go:312: capture terminated with error: ... relation "public.basiccapture_884124" does not exist (SQLSTATE 42P01)
  main_test.go:312: running test capture with checkpoint: {...}
```

The `source-redshift-batch` CI job gets stuck for hours like this because:
1. Two CI jobs shared one table. `uniqueTableID` derives the table name from `t.Name()` alone, so every run on every branch uses `public.basiccapture_884124` on the shared cloud Redshift cluster. Run `31748444363` started its `source-redshift-batch` job 4 seconds after ours and used the same table: it ran `DROP`+`CREATE` at 22:08:13 (our next insert failed at 22:08:14 with
`could not open relation with OID 25980427`), then its cleanup dropped the table at 22:09:06 (our spin began at 22:09:09).
2. `require` was called from a worker goroutine. The inserter calls `executeControlQuery` → `require.NoError` → `t.FailNow()` → `runtime.Goexit()`. Per the [`testing` docs](https://pkg.go.dev/testing#B.FailNow), `FailNow` from a non-test goroutine terminates only that goroutine and does not stop the test. The one failed insert killed the inserter before it reached `insertsDone.Store(true)`, so the flag could never be set.

This PR fixes that by having the worker goroutine report its outcome over a channel instead of asserting directly. The test goroutine reads the channel and calls `require.NoError`, so `FailNow` runs where it can actually abort the subtest & let it fail fast rather than wait 6 hours for GitHub to abort the job.

The other batch connectors (`source-postgres-batch`, `source-sqlserver-batch`, `source-oracle-batch`) were also calling `require` in a worker goroutine, so I fixed those as well for symmetry even though they don't run tests against a shared database instance.

**Checklist:**

### DIFF
--- a/source-db2-batch/capture_test.go
+++ b/source-db2-batch/capture_test.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"fmt"
 	"regexp"
-	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/bradleyjkemp/cupaloy"
 	st "github.com/estuary/connectors/source-boilerplate/testing"
 	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
 )
 
 // TestSimpleCapture exercises the simplest use-case of a capture first doing an initial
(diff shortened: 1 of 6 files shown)


nicolaslazo (dismissed):
Looks good!

I'd make two small fixes to be thorough:
- `source-db2-batch` uses the same atomic bool implementation. We should probably apply the same fix there
- This problem can be partially attributed to the fact that `source-redshift-batch` sets `-timeout=0`, contrary to every other batch connector we have. Will stated in [this thread](https://estuaryworkspace.slack.com/archives/C08LJ60MDBQ/p1787160764252139) that there is no reason to keep it this way — if tests take 7-9 minutes to run, 30m would be plenty enough. This measure should help if tests hang for any other reason in the future.

Alex-Bair:
Thanks for the review! I've applied the same fix to `source-db2-batch`. I've also changed `source-redshift-batch`'s tests to have an hour timeout instead of disabling it. I went _super_ generous here so it's clear just by glancing at the test run time if something is hanging vs. tests are failing for a non-timeout reason; I can imagine the test suite taking longer if we add more tests some day & the standard run time increasing a bit beyond 9 minutes, but I have a hard time believing that we'd ever add so many non-parallel tests that the entire test suite would take an hour to run normally.

willdonnelly (approved):
LGTM

PR #5070, opened 2026-08-17
Reconcile connector docs against flow/site ahead of the docs cutover

jwhartley:
Final pre-cutover reconciliation of `docs/reference/Connectors/` against `estuary/flow` `site/docs/reference/Connectors/`, which is still the canonical source for production `docs.estuary.dev` until the DNS swap.

Measured at `flow@cd8d2713c85` / `connectors@93f4f94fa`: 235 pages on the flow side, 236 here, **40 files differed**. Drift runs in both directions, so this is not a "source wins" replay. Every file was adjudicated individually and the residual diff is now exactly the set of files where this repo is deliberately ahead.

## Replayed from flow (25 files)

| What | Files |
|---|---|
| `/advanced/rediscovery_interval` properties row | 16 SQL capture connectors (MariaDB x2, MySQL x3, OracleDB, PostgreSQL x5, SQL Server x4, AlloyDB) |
| `##### Discovery Filters` section | MySQL, amazon-rds-mysql, google-cloud-sql-mysql |
| Mailchimp deprecation notice and `(Deprecated)` title suffix | `mailchimp.md` |
| Missing capture README index entries | Smartsheet (page existed but was orphaned from the index), `mailchimp-native`, and the four webhook connectors |
| Rewritten webhook pages | `datadog-ingest`, `intercom-ingest`, `jira-ingest`, `twilio-ingest` |
| Dead or wrong third-party links | `neon-postgres`, `bauplan`, `sentry`, `notion`, `pinecone`, `BigQuery` |
| Bare URLs that autolink in tables | `apache-kafka` (capture and materialization) |
| Brand wording, "Flow collections" to "Estuary collections" | `brevo`, `zuora` |
| Missing `description:` frontmatter | `smartsheet`, `zuora` |
| Snowflake service-user script (`type = service`) | `capture-connectors/snowflake.md`, `materialization-connectors/Snowflake.md` |

Each third-party link was checked rather than assumed. `www.bauplan.io` is NXDOMAIN, `docs.neon.tech` does not resolve, `docs.sentry.io/api/projects/list-your-projects/` 404s, and `developers.notion.com/reference/retrieve-a-comment` 404s. The flow-side URL returns 200 in every one of those cases.

## Left alone, this repo is ahead (15 files)

`MongoDB` x3 (`skipBackfills`), `SQLServer` x4 (`discoveryFilters` rows), `criteo.md` (native CDK rewrite), `google-ads.md` (v25 API), `materialization-connectors/README.md` and `apache-iceberg.md` and `dremio.md` (Dremio), `Snowflake.md` (clustering keys), `shopify.md` and the capture README's Shopify labelling.

Three files needed a genuine merge rather than a direction call, because they drift both ways inside one file:

- **SQL Server x4** gain the `rediscovery_interval` row while keeping the `discoveryFilters` rows this repo added in `8b9febb67`.
- **`materialization-connectors/Snowflake.md`** gains `type = service` and flow's SQL indentation while keeping the clustering-keys section and `/clustering` binding property.
- **`capture-connectors/README.md`** gains the webhook, Smartsheet and `mailchimp-native` entries and the `Mailchimp (deprecated)` label, while keeping this repo's `Shopify` / `Shopify (deprecated)` split.

## Structural change: webhook pages move into `http-ingest/`

The five webhook pages move from flat files into `capture-connectors/http-ingest/`, matching flow/site.

**URLs do not change**, verified against a real `docusaurus build` rather than reasoned about. The four sub-pages carry `slug:` frontmatter pinning them at `/reference/Connectors/capture-connectors/<name>/`, and `http-ingest.md` collapses to the folder route because its filename matches its parent. All five resolve at exactly the paths they resolve at today, and no nested child routes leak.

Two reasons to adopt the layout now rather than leave it:

1. The sidebar is autogenerated from the directory tree in both repos, so production currently renders these five as a group. Staying flat would silently flatten that grouping at cutover.
2. The move puts these pages at the same nesting depth as flow's, which makes flow's `../../connector-images/` and `../http-file.md` paths correct here verbatim. The four page rewrites then port with no hand-edited paths, which is the error-prone part of a content replay.

## Two calls worth a second opinion

**`mailchimp.md` uses `:::deprecated`, not flow's `:::warning`.** This matches `shopify.md` in this repo and gives the page the `theme-admonition-deprecated` class Kapa needs to exclude deprecated pages from search. Confirmed present in the build output. `amazon-sqs.md` and `zendesk-support.md` still need the same swap and are tracked separately.

**`qualtrics.md` takes flow's three vendor links, but the direction is genuinely undetermined.** `api.qualtrics.com` returns HTTP 404 server-side for every deep path, including slugs I invented, so status codes prove nothing about either repo's links. Taking flow's version preserves what production serves today rather than swapping three links during the cutover itself. The weak tiebreaker is that this repo's version pointed two different table rows at the same URL, which reads like a lossy edit. Worth a browser spot-check by someone with Qualtrics docs access.

## Verification

Full `npm run build` in `estuary/docs` against this branch, with `onBrokenLinks: 'throw'` and `onBrokenMarkdownLinks: 'throw'`, passes. 341 pages built. Slug sets before and after were diffed: the only additions are the four webhook slugs, and nothing was removed.

## Still due before the DNS swap

**A final re-diff.** The diff grows on its own while cutover waits: 37 files on 2026-08-12, 40 on 08-14, and 40 again today with a partly different composition. This pass is accurate as of `flow@cd8d2713c85` and will drift again.


docs/reference/Connectors/capture-connectors/smartsheet.md
```diff
@@ -1,3 +1,7 @@
+---
+description: Use the Smartsheet connector to capture sheets, reports, and their rows into Estuary Flow. Authenticates with a Smartsheet API access token and captures sheets incrementally and reports as full-refresh snapshots.
```
aeluce, line 2:
```suggestion
description: Use the Smartsheet connector to capture sheets, reports, and their rows into Estuary. Authenticates with a Smartsheet API access token and captures sheets incrementally and reports as full-refresh snapshots.
```

This _is_ what's currently in docs so updating the language isn't a blocker but ideally this should just be "Estuary."

docs/reference/Connectors/capture-connectors/smartsheet.md
```diff
@@ -1,3 +1,7 @@
...
+---
+
 # Smartsheet
 
 This connector captures data from [Smartsheet](https://www.smartsheet.com/) sheets and reports into Estuary Flow collections.
```
aeluce, line 7:
```suggestion
This connector captures data from [Smartsheet](https://www.smartsheet.com/) sheets and reports into Estuary collections.
```

aeluce (approved):

aeluce:
Also: the Qualtrics link changes are from https://github.com/estuary/flow/pull/3231. You've updated them to the correct version here.

PR #5072, opened 2026-08-17
materialize-s3-iceberg: document variant column value typing

jacobmarble:
**Description:**

#4887 added the `variant_columns` option, which stores object, array, and multi-type fields as Iceberg `variant` columns instead of JSON strings. That PR also settled how the values *inside* a variant column are typed, but the docs never mentioned it.

A variant can hold a typed value, not just text. So when a field carries a format annotation, we store a real timestamp, date, or binary inside the variant rather than a string — the same type the value would get if it were a column of its own. This PR documents that in the connector's changelog and docs page. No code changes.

**Workflow steps:**

Using the option doesn't change. The docs now tell you:

- `{format: date-time}` becomes a variant timestamp (nanosecond precision if `nanosecond_timestamps` is on), `{format: date}` a variant date, and a base64 string variant binary
- `uuid`, `time`, and `duration` stay strings, since a variant has no matching type — they're strings as standalone columns too
- a value that fails to parse is stored as a string rather than failing the materialization
- only top-level fields carry format annotations, so anything nested inside an object, an array, or the root document stays a string

**Notes for reviewers:**

- I wrote a new dated changelog entry instead of extending #4887's. Both shipped in the same release, so no user ever saw the old behavior — happy to fold it in if you'd prefer one entry.
- This connector's docs live in this repo, which the docs site pulls in as a submodule. An older copy of this page still sits in `estuary/flow`, but it stopped being the source in July, so I left it alone.

**Checklist:**

- [x] If there are user-facing behavior changes, documentation has been updated (see [documentation conventions](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing)), or the docs PR is linked here: this is the docs PR

### DIFF
--- a/materialize-s3-iceberg/CHANGELOG.md
+++ b/materialize-s3-iceberg/CHANGELOG.md
@@ -1,21 +1,6 @@
 # Changelog
 
-## 2026-07-23
-
-### Changed
-- Schema changes now first commit any transaction that was staged but not yet
-  fully applied to the tables they affect, instead of leaving it to be applied
-  afterwards. This prevents failures where staged data built against the
-  previous table schema could no longer be applied after a column was added,
-  made nullable, or had its type migrated.
-
-### Fixed
-- With the `retain_existing_data_on_backfill` feature flag enabled, backfilling
-  a binding no longer risks losing the rows of a transaction that was committed
-  but not yet fully applied to the destination: the pending transaction is now
-  applied before the backfill takes effect.
-
-## 2026-07-22
+## 2026-08-17
 
 ### Added
 - New advanced option `variant_columns` materializes object, array, and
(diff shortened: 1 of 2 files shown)


materialize-s3-iceberg/CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
...
+  the type implied by its format annotation instead of always being stored as
+  a variant string, matching how the same value is typed as a column of its
+  own: `{format: date-time}` becomes a variant timestamp (nanosecond precision
+  when `nanosecond_timestamps` is also enabled), `{format: date}` a variant
+  date, and `{contentEncoding: base64}` variant binary. Formats with no variant
+  equivalent — `uuid`, `time`, and `duration` — are stored as strings, as they
+  are in a column of their own, and a value that fails to parse as its
+  annotated type is stored as a string. Only top-level fields carry format
+  annotations, so values nested inside an object, an array, or the root
+  document are stored as strings.
```
danielnelson: This is the first entry for variant handling, but the way its written makes it sound like it already existed and its behavior is being modified.

danielnelson (commented):
The new entry should be on current date to match the merge date.

danielnelson (approved):

PR #5074, opened 2026-08-17
materialize-databricks: coalesce all shards' merge queries into one MERGE

mdibaiee:
**Description:**

Measurements show that running N merge queries back to back on Databricks takes roughly N times as long as a single merge. On a scaled-out task the primary shard committed each table with one MERGE per shard, so this changes Acknowledge to commit each table with a **single MERGE (or COPY INTO) query covering every shard's staged files**.

How it works:

- **Store** records structured staging metadata in each checkpoint entry — the staged file names, the rendered merge-bound literals of the observed key range, and whether a MERGE is required. Checkpoint entries no longer carry pre-rendered queries; `Queries` joins `Query` as a deprecated field parsed only from checkpoints written by prior versions.
- **One checkpoint format for all tasks**: every task now emits the range-scoped shape `{rangeKey: {stateKey: entry}}` as a merge patch — a single-shard task's entries nest under its full key range (`00000000-ffffffff`). Verified that a task which has run on runtime v1 without the feature flag gracefully switches to this new connector version without losing data or errors.
- **Acknowledge** groups pending entries by state key across the primary's own, peer-range, and legacy buckets, and renders one set of commit queries over the combined file set at execution time. Per-key bounds are unioned with `LEAST`/`GREATEST`, cast to the column's type since those functions compare in their operands' type (comparing timestamp literals as strings can pick the wrong extremum). A coalesced MERGE source never carries duplicate keys because range buckets cover disjoint slices of the key hash space.

To be able to delegate rendering of queries to Acknowledge is possible thanks to the Apply RPC changes which drain pending commits, and we do this by running the pending Acknowledge using `LastMaterialization` and `LastVersion` so that the Acknowledge drainage is done on the pre-Apply schema of the target table. Previously we rendered the queries and committed in Acknowledged to avoid mismatch between the target table schemas that are available during a recovery Acknowledge, and the table schema for which the files were staged.

Downgrade note: older connector versions parse checkpoint entries strictly and refuse the new format, so a task must be drained (idle at an acknowledged commit) before pinning an older image. This is called out in the changelog.

**Workflow steps:**

No user-facing workflow changes. The `scale_out` feature flag is removed — multi-shard operation (runtime v2) no longer requires opting in.

**Notes for reviewers:**

- Unit tests cover coalesced MERGE/COPY rendering, bound unioning (`TestCombineBounds`), recovery coalescing and its group-level tolerance, old-version entries, file batching, and the unified range-scoped state routing/clearing shapes.
- The 2-shard integration test (`flowctl raw preview-next --fixture --shards 2`) exercises the coalesced MERGE against a real warehouse; the `apply`, `apply-drain`, and `migrate` integration tests run single-shard and exercise the nested checkpoint format.

**Checklist:**


mdibaiee:
## Consistency suite: 14 of 14 applicable scenarios pass

[Suite](https://github.com/estuary/flow/pull/3294) run against `1b06d363` — real Flow runtime, live Databricks, `scale_out` enabled, class `postCommitApply`.

| Scenario | Documents | Perturbation |
|---|---|---|
| `baseline` | 3,401 | none |
| `crash-at-flush` | 5,621 | crash at `Flush` |
| `crash-between-commits` | 5,997 | crash after `Acknowledged` |
| `crash-mid-store` | 5,984 | crash mid-`Store` |
| `at-least-once-never-loses` | 5,696 | crash at `StartedCommit`, held to no-duplicates |
| `destination-ahead-of-checkpoint` | 5,505 | destination ahead of the checkpoint |
| `recovery-reconciles-with-destination` | 5,668 | recovery must reconcile |
| `split-during-store` | 3,909 | split mid-`Store` — coalesced MERGE, bound unioning |
| `split-during-commit` | 4,664 | crash at `StartCommit`, then split |
| `split-after-commit-before-apply` | 4,466 | crash at `Acknowledge`, then split — per-entry recovery fallback |
| `join-after-split` | 10,123 | split then join — survivor commits the absorbed range |
| `crash-in-split-leader` | 6,126 | split, crash the log-holding shard |
| `crash-in-split-non-leader` | 6,501 | split, crash a stateless shard |
| `split-lands-on-prepared-transaction` | 3,611 | membership change on a prepared transaction |

~77,000 documents. Each run compares the destination against an expectation read from the collection itself: no loss, no duplicates, conservation, oracle agreement, per-document cardinality on the delta bindings.

`zombie-at-start-commit` is skipped — it needs a class that fences at `Open`.

No violations, no wedges, no environmental failures.

materialize-databricks/CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+  `scale_out` flag still present in a task's configuration is ignored.
+- A task scaled out to multiple shards now commits each table with a single
+  `MERGE` (or `COPY INTO`) query covering every shard's staged files, instead
+  of running one query per shard back to back. Single-shard tasks are
+  unaffected. The connector checkpoint now uses the range-scoped format for
+  all tasks — a single-shard task's entries nest under its full key range —
+  and records the staged files and key bounds instead of pre-rendered
+  queries. Note for downgrades: versions predating this change refuse to
+  parse the new format — drain the task (let it idle at an acknowledged
+  commit) before pinning an older image.
```
danielnelson, line 17: For downgrades, the way I have heard to drain a task is to set `notAfter` on a binding and wait until all data to that point has been transferred (I've never actually done this).  How do you idle at an acknowledged commit?

Not something that needs answered now, but should we even support downgrade?  It feels like this would add a lot of cases to handle in general.

materialize-databricks/driver.go
```diff
@@ -689,18 +660,10 @@ func (d *transactor) Store(it *m.StoreIterator) (_ m.StartCommitFunc, err error)
...
-		}
-
-		return &pf.ConnectorState{UpdatedJson: checkpointJSON}, nil
-	}
-
 	// Emit only this shard's range bucket as a merge patch: range keys are
 	// disjoint across shards, so concurrent patches never clobber when the
-	// runtime consolidates connector state.
+	// runtime consolidates connector state. A single-shard task uses the same
+	// format, with one bucket covering the full key range.
```
danielnelson: Why mention that it's the same format?  Document the current code and not the diff.

materialize-databricks/driver.go
```diff
@@ -827,65 +788,148 @@ func (d *transactor) acknowledgeApply(ctx context.Context, db *stdsql.DB, should
...
-		if len(path) == 0 {
+// executeStateKeyItems commits one state key's entries. Entries with structured staging
+// metadata coalesce into a single set of queries; those written by older connector versions
+// have only their pre-rendered queries, so they run individually.
+//
+// Acknowledge fans-in by the runtime: the runtime waits for all Acknowledged responses from all
+// shards before a new transaction begins, and files are already cleaned up once all queries have
+// run. Thus a missing file means the whole query has already run, and during recovery that is
+// acceptable: we may have run the query but not have been able to clear the checkpoint.
+func (d *transactor) executeStateKeyItems(ctx context.Context, db *stdsql.DB, b *binding, items []*checkpointItem) error {
```
danielnelson: I don't love this function's name.

danielnelson: I'm not sure about this function name, it feels odd to refer to it as operating on a StateKey vs Binding.  Maybe `commitBinding`?

danielnelson (dismissed):

materialize-databricks/driver.go
mdibaiee: I think the distinction is that a binding can have multiple statekeys, e.g. we may have `"binding.v1"` as a state key in the checkpoint, and then later we will have `"binding.v2"` and it is possible (I think? maybe the runtime always clears the old versions of a binding's statekey? I am not sure) to have both of these in the checkpoint (when a backfill is triggered) but we only apply the latest statekey of a binding

But I suppose as it relates to naming we could name this more related to a binding

materialize-databricks/CHANGELOG.md
mdibaiee: It is just in case of an incident where we need to downgrade, otherwise downgrading is not a supported path in general

materialize-databricks/driver.go
mdibaiee: Renamed to `commitBindingCheckpointItems`

danielnelson (dismissed):

danielnelson (approved):

PR #5076, opened 2026-08-17
source-facebook-marketing-native: split insight jobs by field when ad level jobs fail

Alex-Bair:
**Description:**

We've observed async insight jobs consistently fail even when the job has been split to the lowest level (`ad`) for a single ad on a single day. This fails the capture & prevents backfills from completing. We originally deferred implementing logic to split `ad` level jobs into jobs for individual fields, but that appears to be what's needed to get past these occasional roadblocks where Facebook cannot / will not complete a job that includes a certain field on a specific day.

This PR's scope includes:

- **Log Facebook's error code and subcode on failed insights jobs.** The subcode picks between opposite remedies (2/1504038 = too large, 2/1504044 = reads as transient) but only the user-facing prose was logged, and it's identical for both. Also parses `error_message`, names entities in job descriptions, and pairs the job description with its `report_run_id`. These details will help us troubleshoot failing jobs in the future.

- **Split jobs by field when they cannot be split by entity.** Halves the non-PK fields until each part succeeds, carrying primary keys into every part, then merges by primary key. A field that fails alone is dropped after a retry. Measured against the ad behind the escalation: 47 of 48 fields recoverable, with the `actions` field alone as unfetchable. If Facebook refuses every field, the capture stops rather than advancing past data it never obtained.

**Checklist:**

### DIFF
--- a/source-facebook-marketing-native/source_facebook_marketing_native/insights/manager.py
+++ b/source-facebook-marketing-native/source_facebook_marketing_native/insights/manager.py
@@ -86,14 +86,23 @@ class FacebookInsightsJobManager:
         Examples:
             - "ACCOUNT job [depth=0]"
             - "campaigns job (2 entities) [depth=1, from ACCOUNT]"
-            - "adsets job (1 entities) [depth=2, from campaigns]"
+            - "adsets job (1 entities, ids=[123]) [depth=2, from campaigns]"
+
+        Small jobs name their entities, since those are the ones that fail
+        unsplittably and a bare count doesn't say which entity to go look at.
         """
         if job.scope == JobScope.ACCOUNT:
             return f"ACCOUNT job [depth={job.depth}]"
 
         count = len(job.entity_ids) if job.entity_ids else 0
+        ids_info = (
+            f", ids={job.entity_ids}" if job.entity_ids and count <= 3 else ""
+        )
         parent_info = f", from {job.parent_scope.value}" if job.parent_scope else ""
-        return f"{job.scope.value} job ({count} entities) [depth={job.depth}{parent_info}]"
+        return (
+            f"{job.scope.value} job ({count} entities{ids_info}) "
+            f"[depth={job.depth}{parent_info}]"
+        )
 
     def _build_filter(self, job: InsightsJob) -> InsightsFilter | None:
         """Build InsightsFilter from job scope and entity IDs."""
@@ -145,6 +154,7 @@ class FacebookInsightsJobManager:
                     action_breakdowns=model.action_breakdowns,
                     action_attribution_windows=model.action_attribution_windows,
                     filtering=filtering,
+                    job_desc=job_desc,
                 )
 
                 await self._wait_for_completion(log, job_id, job_desc)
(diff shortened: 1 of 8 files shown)


source-facebook-marketing-native/source_facebook_marketing_native/insights/manager.py
```diff
@@ -173,37 +179,263 @@ async def _execute_job(
...
+            list(model.fields), pk_fields, unfetched, depth=0,
+        )
+
+        if len(unfetched) == len(splittable):
+            raise CannotSplitFurtherError(
+                f"{job_desc}: Facebook would not report on any of its "
+                f"{len(splittable)} fields, even one at a time. Last error: {error}"
+            )
+
+        records = self._merge_by_primary_key(parts, pk_fields)
```
nicolaslazo: I suspect we are not deriving primary keys properly, and that might lead to a number of unrelated records being merged together.

In the definition for `_primary_key_fields`:

```python
    def _primary_key_fields(
        self, model: type[FacebookInsightsResource]
    ) -> list[str]:
        """
        Request fields that identify a row.
        """
        pointers = [key.lstrip("/") for key in model.primary_keys]
        assert all("/" not in p for p in pointers), (
            f"{model.name} has a nested primary key; the field-split join key "
            f"assumes top-level pointers: {model.primary_keys}"
        )
        return [p for p in pointers if p in model.fields]
```

We are filtering on the condition that `p in model.fields`. That does not apply in cases like `AdsInsightsAgeAndGender`, to give an example:

```python
class AdsInsightsAgeAndGender(AdsInsights):
    name: ClassVar[str] = ResourceName.ADS_INSIGHTS_AGE_AND_GENDER
    breakdowns: ClassVar[list[Breakdown]] = [Breakdown.AGE, Breakdown.GENDER]
    primary_keys: ClassVar[list[str]] = sorted(
        AdsInsights.primary_keys + ["/age", "/gender"]
    )
    age: str
    gender: str
```

`/age` and `/gender` are implicitly included in the response but are not defined in the document's `fields` list, so they would be ignored at merge time. 

This bug would affect every instance where we're adding fields to the pkey list that are not defined in the `fields` list.

nicolaslazo (changes requested):

source-facebook-marketing-native/source_facebook_marketing_native/insights/manager.py
Alex-Bair: Thanks for catching that! I've pushed a fix for that bug & to better distinguish between "key fields needed to join records back together" vs. "key fields requested in the `fields` query param":
- `_join_key` returns all key fields used to uniquely identify a row.
- `_requested_key_fields` returns only the key fields we have to explicitly request from Facebook via the `fields` query param.
- `_verify_join_key` makes sure each chunk we receive back contains all key fields from `_join_key` so we can stitch them back together correctly. If Facebook omits a key field for whatever reason, an exception is raised so we can look into it.
- `FieldSplitPart` dataclass couples a chunk's partial records with the fields it asked for.

nicolaslazo (approved):
Thanks for addressing that!

PR #5085, opened 2026-08-18
source-shopify-native: add `returnShippingFees` and `fulfillmentLineItem` to order_returns

### LINKED ISSUE #5081: source-shopify-native: add `returnShippingFees` and `fulfillmentLineItem` to the `order_returns` stream
## Problem

The `order_returns` stream doesn't capture [`Return.returnShippingFees`](https://shopify.dev/docs/api/admin-graphql/latest/objects/Return#field-Return.fields.returnShippingFees) or [`ReturnLineItem.fulfillmentLineItem`](https://shopify.dev/docs/api/admin-graphql/latest/objects/ReturnLineItem#field-ReturnLineItem.fields.fulfillmentLineItem). A customer needs the shipping fee charged on a return (their returns provider doesn't expose it), and the fulfillment line item to tie each return line item back to its order line item.

Minimum field set:

- `returnShippingFees { id amountSet { shopMoney { amount currencyCode } } }`
- `fulfillmentLineItem { id lineItem { id } }`

## Possible Solution

Query-only change in `graphql/orders/returns.py` — `order_returns` is already on the bulk path and neither field is a connection, so no state migration and no schema work.

One gotcha: `returnLineItems` is a connection over the [`ReturnLineItemType`](https://shopify.dev/docs/api/admin-graphql/latest/interfaces/ReturnLineItemType) interface and `fulfillmentLineItem` only exists on the concrete `ReturnLineItem`, so it needs an `... on ReturnLineItem { }` fragment.

Worth confirming against a real bulk job that both fields inline into their parent JSONL rows — `process_result` raises on any line it can't identify by gid prefix.

SeanWhelan:
**Description:**

Adds `Return.returnShippingFees` and `ReturnLineItem.fulfillmentLineItem` to the `order_returns` stream's bulk query. A customer needs the shipping fee charged on a return (their returns provider doesn't expose it), and the fulfillment line item to tie each return line item back to its order line item.

**Workflow steps:**

Query-only change in `graphql/orders/returns.py` — no config, schema, or state changes. `order_returns` was already on the bulk path, and both new fields are non-connection fields, so they inline directly into their parent JSONL rows and `process_result` didn't need any changes.

`returnLineItems` is a connection over the `ReturnLineItemType` interface, and `fulfillmentLineItem` only exists on the concrete `ReturnLineItem` (not the sibling `UnverifiedReturnLineItem`), so it's requested via an `... on ReturnLineItem { }` inline fragment. `returnShippingFees.amountSet` reuses this connector's existing shared `_MoneyBagFields` fragment, matching every other `MoneyBag`-typed field in the connector, rather than a narrower one-off shape.

**Checklist:**

Closes #5081

### DIFF
--- a/source-shopify-native/source_shopify_native/graphql/orders/returns.py
+++ b/source-shopify-native/source_shopify_native/graphql/orders/returns.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from logging import Logger
 from typing import AsyncGenerator, Any
 
+from ..common import money_bag_fragment
 from ...models import (
     ConditionalField,
     ShopifyGraphQLResource,
@@ -30,6 +31,12 @@ class OrderReturns(ShopifyGraphQLResource):
                     note
                     reason
                 }
+                returnShippingFees {
+                    id
+                    amountSet {
+                        ..._MoneyBagFields
+                    }
+                }
                 # {{ staffMember }}
                 returnLineItems {
                     edges {
@@ -44,6 +51,14 @@ class OrderReturns(ShopifyGraphQLResource):
                                 deleted
                                 handle
                                 name
+                            }
+                            ... on ReturnLineItem {
+                                fulfillmentLineItem {
+                                    id
+                                    lineItem {
+                                        id
+                                    }
+                                }
                             }
                         }
                     }
@@ -70,6 +85,7 @@ class OrderReturns(ShopifyGraphQLResource):
         }
     }
     """
+    FRAGMENTS = [money_bag_fragment]
     CONDITIONAL_FIELDS = [
         # staffMember requires the read_users scope, which is only grantable on
         # Shopify Plus / Advanced stores or finance embedded apps.
--- a/source-shopify-native/CHANGELOG.md
+++ b/source-shopify-native/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 2026-08-18
+
+### Added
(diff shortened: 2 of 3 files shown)


source-shopify-native/source_shopify_native/graphql/orders/returns.py
```diff
@@ -45,6 +52,14 @@ class OrderReturns(ShopifyGraphQLResource):
...
                                 name
                             }
+                            ... on ReturnLineItem {
+                                fulfillmentLineItem {
+                                    id
+                                    lineItem {
+                                        id
+                                    }
+                                }
+                            }
```
Alex-Bair, line 62: note for myself: we don't need to make the `fulfillmentLineItem` field a conditional one since the `read_orders` scope & friends it requires is already necessary for the top level `orders` query for this stream. However, this stream currently doesn't require a `read_orders` or similar scope to ensure the `orders` query succeeds, and we should eventually fix that by updating the connector to be able to express an AND of `anyOf`s in a stream's qualifying scopes (ex: `OrderReturns` requires both one of `read_returns` or `read_marketplace_returns` AND one of `read_orders` or `read_all_orders` or `read_marketplace_orders`).

We likely haven't been bitten by this yet because `read_orders` is a very common scope for any Shopify integration.

Alex-Bair (approved):
LGTM, thanks!

PR #5086, opened 2026-08-18
sqlcapture: Add 'Additional Backfill Filter' advanced option to resource configs

willdonnelly:
**Description:**

Adds an Advanced Options section to the resource spec with a new 'additional_backfill_filter' property, a SQL expression which is ANDed into the WHERE clause of every backfill query for that binding. This allows a power user (hopefully guided by support) to use application knowledge to make large re-backfills more efficient.

On PostgreSQL it composes with the existing XMIN filtering option.

Currently validation requires a re-backfill of the binding whenever the option is set or changed (clearing is exempt so you can remove it afterwards). I'm not certain whether this is desirable but it seemed like the safer approach to start with.

Because precise backfills drop replication events for rows beyond the backfill cursor on the assumption that they'll be read later on, and a backfill filter can violate that assumption, we force unfiltered backfill mode when a filter is set.

(Arguably we should be doing that any time an incremental backfill is performed, since deletions also violate this assumption. Precise backfills are kind of an impressive trick but practically speaking they're a giant headache.)

**Workflow steps:**

1. In the Estuary dashboard, edit the capture and open the configuration of the relevant binding.
2. Expand **Advanced Options** and fill out the "Additional Backfill Filter" field.
3. Trigger a backfill of that binding, making sure "Incremental backfill" is selected in the backfill drop-down, and publish.

   Selecting a [Dataflow Reset](#dataflow-reset) instead would drop the filter-excluded rows from destination tables entirely, so only choose that if a filtered subset is the intended end state.

   A publication which sets or changes the filter must also trigger a backfill of that binding, and will otherwise fail validation.
4. After the backfill completes, clear the field and publish again. Clearing it does not require another backfill. This prevents a stale filter from being applied to future backfills by mistake.

**Notes for reviewers:**

This is the first and only instance of an "Advanced Options" sub-object in a binding's resource config, but we've got a couple of other options (like "Backfill Mode") which arguably ought to be in there as well and it seemed like a situation where eventually we just have to bite the bullet and add the category with a single option in it so that it's there when we go to add another advanced toggle.

The UI's handling of a sub-group on a resource config hasn't really been exercised before since AFAICT we don't have any of those anywhere, but it's using the same rendering as the endpoint configs and seems to work fine in practice.

Collapsed:
2026-08-18-151925_3840x2160_scrot

Expanded:
2026-08-18-151933_3840x2160_scrot

**Checklist:**

### DIFF
--- a/sqlcapture/main.go
+++ b/sqlcapture/main.go
@@ -27,6 +27,8 @@ type Resource struct {
 
 	Priority int `json:"priority,omitempty" jsonschema:"title=Backfill Priority,description=An optional integer priority for this binding. The highest priority binding(s) will be backfilled completely before any others. The default priority is zero. Negative priorities are allowed and will cause a binding to be backfilled after others."`
 
+	Advanced *AdvancedResourceOptions `json:"advanced,omitempty" jsonschema:"title=Advanced Options,description=Options for advanced users. You should not typically need to modify these." jsonschema_extras:"advanced=true"`
+
 	// PrimaryKey allows the user to override the "scan key" columns which will be used
 	// to perform backfill queries and merge replicated changes. If left unset we default
 	// to the collection's key, which is basically always what the user wants, so we omit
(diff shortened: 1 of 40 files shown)


sqlcapture/main.go
```diff
@@ -37,6 +39,20 @@ type Resource struct {
...
+	AdditionalBackfillFilter string `json:"additional_backfill_filter,omitempty" jsonschema:"title=Additional Backfill Filter,description=Optional filter clause which will be applied to all backfill queries for this binding. Contact Estuary support for assistance before using this option."`
+}
+
+// BackfillFilter returns the user-configured backfill filter expression, or an
+// empty string when no filter is set.
+func (r Resource) BackfillFilter() string {
+	if r.Advanced == nil {
+		return ""
+	}
+	return r.Advanced.AdditionalBackfillFilter
```
Alex-Bair, line 53: nit: would it be worth it to `strings.TrimSpace` the user provided `AdditionalBackfillFilter` to avoid running into bugs where users accidentally leave the field full of whitespace and we try using a where clause like `WHERE ctid > $1 AND ctid <= $2 AND (   )` in our queries? And probably also avoid some potential pitfalls during validation if a user clears out the field after a backfill finishes but leaves a stray space?

Alex-Bair (dismissed):
LGTM % a small nit that may save us some troubleshooting headache later.

sqlcapture/main.go
willdonnelly: That's a good idea, I think the cleanest solution is probably a validation check forbidding an all-whitespace value (since otherwise the user might try to publish it and get told they have to backfill by the existing "you changed it and it's not gone so you need to backfill now" logic and blindly do that.

Alex-Bair (approved):
LGTM!

Issue #5082, opened 2026-08-18
materialize-databricks: staged files are uploaded to the Unity Catalog volume uncompressed


jwhartley:
## 1. Priority

Low - no task is failing. The effect is staged-file size, and therefore network transfer and destination ingest, across all `materialize-databricks` tasks.

## 2. Scope and prevalence

- All `materialize-databricks` tasks.
- `materialize-databricks` is the only production caller of `writer.WithJsonDisableCompression()` in this repo; the only other reference is `go/writer/json_test.go`.
- A BYOC deployment reported material NAT data-processing cost attributable to these uploads during a backfill.

## 3. Problem description

- Expected: staged files are compressed before upload, as they are for the other warehouse materializations that use `go/writer` (compression is on by default, `go/writer/json.go:63-71`).

- Observed: `materialize-databricks/staged_file.go:236` constructs its JSON writer with compression explicitly disabled:

  ```go
  f.writer = writer.NewJsonWriter(f.buf, f.fields, writer.WithJsonDisableCompression())
  ```

  Files are named `<uuid>.json` (`staged_file.go:232`) and PUT uncompressed to the Unity Catalog volume via `PUT '<local>' INTO '<volume path>' OVERWRITE` (`staged_file.go:203`). Target file size is 128MB of uncompressed JSON (`staged_file.go:19`).

- Vendor behaviour, for reference: Databricks documents that pre-compressed files are read directly, listing `gzip` among the supported codecs for text-based formats including JSON (https://learn.microsoft.com/azure/databricks/query/formats/#data-formats-with-special-considerations). The connector reads staged files back via `FROM json.\`<file>\`` (`sqlgen.go:200`, `:234`, `:300`) and `COPY INTO ... FILEFORMAT = JSON` (`sqlgen.go:274`).

- Origin: compression has never been enabled on this connector. `7ba41c088` (2023-11-07) introduced the current staged-file path with compression off, as `sql.NewCountingEncoder(w, false, f.fields)`; the implementation it replaced also wrote plain `.json` files. Neither the commit message nor any code comment records a reason. The reason was stated later in Slack (2025-01-31):

  > The reason we don't do compression for databricks is that at the time the connector was built, databricks couldn't read compressed json files. Don't know if that's still the case

  Same rationale given earlier (2024-09-01): "they aren't compressed (databricks can't read compressed JSON)".

- Steps to reproduce: run any `materialize-databricks` task and list the staged files under the volume's `flow_temp_tables` path during a transaction.

[...]

PR #5087, opened 2026-08-18
materialize-databricks: Optionally compress staged files

jacobmarble:
**Description:**

Helps #5082.

The Databricks materialization stages data as JSON files on a Unity Catalog volume, then tells Databricks to read them back. It has always written those files uncompressed. In the past, Databricks couldn't read compressed JSON — that's no longer true, so we can reduce network egress costs by enabling it. (One customer saw a large NAT processing bill during a backfill).

This adds an advanced option, `compress_staged_files`, that gzips the staged files. It's **off by default**, so nothing changes for existing tasks unless someone turns it on. When it's on, files get a `.json.gz` name, which is how Databricks knows to decompress them on read.

**Workflow steps:**

Set `advanced.compress_staged_files: true` in the endpoint config. Files staged during each transaction are then gzipped before upload. The trade is less data over the wire for more CPU in the connector.

**Notes for reviewers:**

- Compression is off by default. Once proven, we can decide to hard-code-on or default-on.
- One thing to be aware of before that flip: the 128MB staged-file size limit counts bytes written to disk, so with compression on a file holds roughly 8x more rows. This is probably safe, but safer to try it with some real workloads first.

**Checklist:**

- [ ] If there are user-facing behavior changes, documentation has been updated (see [documentation conventions](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing)), or the docs PR is linked here: n/a


docs/reference/Connectors/materialization-connectors/databricks.md
```diff
@@ -65,6 +65,7 @@ Use the below properties to configure a Databricks materialization, which will d
 | **`/schema_name`**                       | Schema Name  | Default schema to materialize to                                                                                                  | string                                                                                                             | `default` schema is used |
 | `/hardDelete` | Hard Delete | If enabled, items deleted in the source will also be deleted from the destination. By default, deletions are tracked via `_meta/op` (soft delete). | boolean | `false` |
 | `/advanced/no_flow_document` | Exclude Flow Document | When enabled, the root document will not be required for standard updates. See [excluding flow_document with standard updates](/guides/customize-materialization-fields/#excluding-flow_document-with-standard-updates) for details. | boolean | `false` |
+| `/advanced/compress_staged_files` | Compress Staged Files | When enabled, files staged to the Unity Catalog volume are compressed with gzip. This reduces the amount of data uploaded to Databricks, at the cost of additional CPU usage. | boolean | `false` |
```
danielnelson: I wonder if they will add additional supported compression methods, like zstd, in the future.  Do you think we should add the option in a way would support different compression methods?

danielnelson (dismissed):

docs/reference/Connectors/materialization-connectors/databricks.md
mdibaiee: I agree, materialize-kafka has an option for compression methods, we could use a similar approach 

https://github.com/estuary/connectors/blob/main/materialize-kafka/src/configuration.rs

jacobmarble: I ran with the assumption that people wouldn't particularly care because the practical difference is between uncompressed and compressed, and because these are ephemeral staging files, not permanent artifacts.

If this connector were storing persistent JSON objects, then I'd agree with more configuration, but KISS is more compelling here IMO.

mdibaiee: I see, since this is an internal detail of how we stage files and land the data, the customers won't really have to ever read the compressed files anyway...

In that case... why do we even have this as an option? Why not always compress? is there a reason we would not want to compress?

jacobmarble: I don't see a documented list of supported codecs from Databricks. They nod at Spark, and we could certainly test empirically.

The JSON code in `go/writer` only supports uncompressed and gzip. We can certainly add more codecs, but I'm not sure the complexity is worth it because only one customer is concerned about network costs, and that customer isn't concerned about which codec we use.

mdibaiee: @jacobmarble the idea was not to add additional codecs / compression options support just now, but to use an enum format for the config so that it is either `none/nil` or `gzip` for example (similar to the json writer), and later, if it is worth it, we can add additional options to the existing config rather having to switch from a boolean flag to an enum at a later stage

mdibaiee: but given that this is an internal only set of files, I would again ask why should we even have this as an option rather than always using compression whenever it is available? is there any tradeoff or any other reason a customer might want to opt out of compression for staged files?

jacobmarble: > In that case... why do we even have this as an option? Why not always compress? is there a reason we would not want to compress?

Agreed, the ideal route is to hard-code with compression.

I'm a little worried about [the 128MB limit](https://github.com/estuary/connectors/blob/e0e2dcc4aee6bcb0441576d68334149be54e57b8/materialize-databricks/staged_file.go#L19) which counts bytes *after* compression. That means 128 MB files will contain roughly 1GB of JSON data. This will likely not be a problem, but I'd prefer to have a fast kill switch when this is released, and not have to wait on the revert-release cycle.

Having said that, I realize now that default-on is the right way to write this PR.

mdibaiee: we can probably test the 128GB limit after compression case locally with a generated transaction that is large enough to be sure

mdibaiee: I bet Claude can drive such a test based on the consistency suite, the traces will help us know how many transactions ran and at which size, etc.

jacobmarble: > Having said that, I realize now that default-on is the right way to write this PR.

... and still plan to hard-code-on some time after releasing.

jacobmarble: > I bet Claude can drive such a test based on the consistency suite, the traces will help us know how many transactions ran and at which size, etc.

Perusing...
😆

jacobmarble: Ran integration tests with two larger payloads, both compressed to 128MB:
- 340MB (2.7x), query from staged data: 11s
- 590MB (4.7x), query from staged data: 13s
  - increased repetitive data to get higher compression ratio

No issues encountered.

jacobmarble: Also running the consistency suite with compression on.

jacobmarble: Consistency suite is done, run against the built connector with compression on (subject class `postCommitApply`): **16/16 passed** in ~16 min.

That covers `crash_at_flush`, `crash_mid_store`, `crash_between_commits`, `recovery_reconciles_with_destination`, `destination_ahead_of_checkpoint`, and the split/join set (`split_during_store`, `split_during_commit`, `split_after_commit_before_apply`, `split_lands_on_prepared_transaction`, `join_after_split`, both `crash_in_split_*`).

Two non-failures worth naming: `zombie_at_start_commit` reported not-applicable (it can only be upheld by `remoteAuthoritative`), and the `exempt: [monotonicity]` lines are the suite's pre-existing exemptions from `scenarios.rs` — one of which was already calibrated against materialize-databricks. The set-based checks that carry the exactly-once claim (no-loss, no-duplicates, conservation, oracle agreement) are not exempt, and passed.

mdibaiee: @jacobmarble if we introduce a publicly available option like the new disable_compression, it is not as easy to just remove it since customers may start using it by the time we decide to take it away.

Other than keeping for the case of an incident / this causing trouble, is there another reason you would like to keep it? If this ends up causing an incident we can revert the pull-request (and that probably is faster than having to set the option to true on the broken tasks).

jacobmarble: I don't have any other reason to keep the config option. Compression is now hard-coded-on.

mdibaiee (approved):

PR #5038, opened 2026-08-10
sqlcapture: Support linked_collections

willdonnelly:
**Description:**

Connector support for the `linked_collections` mechanism added to the runtime in https://github.com/estuary/flow/pull/3303. We don't actually refer to collection specs in very many places so really all we need to do is a Flow version bump and using new accessors in a couple of places.

Also adds a test in `source-postgres` which exercises the feature and demonstrates that it produces the expected results into a single collection, although strictly speaking the test doesn't show that the `linked_collections` field is actually used and I had to verify that bit myself out-of-band, but if you assume that the `shards.flags.indirect-specs = true` setting does what it's supposed to the test still has a little value.

**Checklist:**

- [X] If this change is user-visible, the affected connector's `CHANGELOG.md` and documentation have been updated (see [CONTRIBUTING.md](../CONTRIBUTING.md#changelog-entries)). The `Docs / CHANGELOG check` CI job reviews this automatically; apply the `docs-check-skip` label to dismiss a false positive.

PR #5088, opened 2026-08-19
sqlcapture: performance improvements for captures with many bindings

Alex-Bair:
**Description:**

After the `linked_collections` mechanism support was added in https://github.com/estuary/connectors/pull/5038, a user set up a capture with 40,500+ bindings and noticed backfill throughput was significantly slower than anticipated. Captures have never needed to work with that many bindings before, so it's little surprise that there are strategies sprinkled throughout the code that are slow at that scale.

This PR is a coterie of performance improvements for SQL captures so they're more performant in the new `linked_collections` world. The improvements are described in each commit along with benchmarks, but a brief summary of the changes in `sqlcapture/capture.go` is below:
- Avoid allocating in the sort comparator by comparing the `Schema`/`Table` directly instead of using `StreamID.String()`.
- Stop building a list in the main loop guard. Instead of using `if BindingsCurrentlyBackfilling() != nil` which allocated and sorted every backfilling binding just to see if any existed, we perform an exit early scan.
- Use an `iter.Seq` for the binding state predicate, requiring callers to compose with `slices.SortedFunc`, `slices.AppendSeq`, or returning early as necessary.
- Make the removed binding check no longer O(N^2) by using an existence map instead of scanning every binding inside a loop over every state entry.

On my machine, I measured the overall improvements as:
|   |  before | after |
| -- | -- | -- |
| per backfill chunk | 591.80 ms, 225 MB, 11.8M allocs | 15.73 ms, 655 KB, 2 allocs |
| startup binding reconciliation | 33.6 s | 10.5 ms |

Motivating Slack thread: [link](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1787100817430129)

**Notes for reviewers:**

~~Additionally, I updated a stale comment that indicated a loop could be removed a couple days after a late 2024 migration. I _almost_ removed it, but Claude figured out that the loop is still necessary so `source-mysql` can detect primary key updates.~~

There may be additional performance improvements we need to make as captures start exercising the `linked_collections` mechanism and increasing the number of bindings

**Checklist:**

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"iter"
 	"math"
 	"math/rand"
 	"os"
@@ -61,17 +62,29 @@ func (ps *PersistentState) Validate() error {
 	return nil
 }
 
-// BindingsInState returns all the bindings having a particular persisted state, in sorted order for
-// reproducibility.
-func (c *Capture) BindingsInState(modes ...BackfillMode) []*Binding {
-	var bindings []*Binding
-	for _, binding := range c.Bindings {
-		if slices.Contains(modes, BackfillMode(c.State.Streams[binding.StateKey].Mode)) {
-			bindings = append(bindings, binding)
+// bindingsInState yields every binding whose persisted state is one of the given modes, in
+// map iteration order. It exists so that the state predicate lives in exactly one place:
+// callers compose it with slices.SortedFunc when they need a reproducible order, with
+// slices.AppendSeq when the order doesn't matter, or range over it with an early return
+// to test for existence without building a list at all.
+func (c *Capture) bindingsInState(modes ...BackfillMode) iter.Seq[*Binding] {
+	return func(yield func(*Binding) bool) {
+		for _, binding := range c.Bindings {
+			if slices.Contains(modes, BackfillMode(c.State.Streams[binding.StateKey].Mode)) {
+				if !yield(binding) {
+					return
+				}
+			}
 		}
 	}
-	slices.SortFunc(bindings, func(a, b *Binding) int { return a.StreamID.Compare(b.StreamID) })
-	return bindings
+}
+
+// BindingsInState returns all the bindings having a particular persisted state in sorted order for
+// reproducibility.
+func (c *Capture) BindingsInState(modes ...BackfillMode) []*Binding {
+	return slices.SortedFunc(c.bindingsInState(modes...), func(a, b *Binding) int {
+		return a.StreamID.Compare(b.StreamID)
+	})
 }
 
 // BindingsCurrentlyActive returns all the bindings currently being captured.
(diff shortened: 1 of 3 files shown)


sqlcapture/capture.go
```diff
@@ -61,17 +62,35 @@ func (ps *PersistentState) Validate() error {
...
+			}
 		}
 	}
-	slices.SortFunc(bindings, func(a, b *Binding) int { return strings.Compare(a.StreamID.String(), b.StreamID.String()) })
-	return bindings
+}
+
+// compareBindingsBySchemaAndTable orders bindings by schema and then by table name.
+func compareBindingsBySchemaAndTable(a, b *Binding) int {
+	if n := strings.Compare(a.StreamID.Schema, b.StreamID.Schema); n != 0 {
```
willdonnelly: This comparison itself seems like it ought to live in a method `(a StreamID) Compare(b StreamID) int` or something along those lines.

sqlcapture/capture.go
```diff
@@ -452,10 +485,19 @@ func (c *Capture) reconcileStateWithBindings(_ context.Context) error {
 		}
 	}
 
-	// This is part of a migration dated 2024-12-12 and can be removed within a few days. The
-	// only purpose here is to make sure that preexisting only-changes bindings have an
-	// initialized StateKey value, the same as activatePendingStreams has been modified to
-	// produce going forward.
+	// Make sure that only-changes bindings which are already Active have initialized
```
willdonnelly: I'm not at all certain this is correct. I don't follow the reasoning given and the claims appear to jump from "a keyless backfill" to "cannot detect primary key updates" without justifying how this scenario actually hangs together.

_(edit: well, I guess there is the scenario where somebody explicitly sets a binding to keyless mode even though when the collection itself still has a key. I overlooked that at first because it's less common and because nobody would ever do that in a MySQL capture specifically because the keyless backfill mode is significantly less efficient (rather than more as in Postgres). Still not entirely following the reasoning beyond that though)_

That said there totally _could_ be some subtle reason this code is still needed, my concern is just that changing the comment could be misleading if the reasoning as stated here is wrong. If you've worked through it and this makes sense to you then feel free to ignore this note.

willdonnelly (dismissed):
LGTM % two nits

sqlcapture/capture.go
willdonnelly: Okay, I've been pondering this a little bit more and I think this is a sort of Claude-myopia where it's answering the question "does this block of code right here cause any observable behavior changes" with an example of a scenario where removing it would cause the connector's behavior to differ, but not asking whether that's a reasonable or important change.

What I think is actually going on here is that the behavior around PK-update changes and "Without Primary Key" bindings is just always incorrect, and this bit of migration code happens to _fix_ that incorrectness if you subsequently switch the binding to "Only Changes" mode, but only if you do that.

Alex-Bair: Yeah, I trusted Claude too much on this one. I'm sorry about that. The scenario it was concerned about was purely the `source-mysql` case where we discover a table with a primary key but the user explicitly picks the "Without Primary Key" mode & switches to "Only Changes" at some later point. Like you suggest, that's not a reasonable scenario. I'll drop this comment update so it's not confusing later.

Alex-Bair: The commit updating the comment has been dropped.

sqlcapture/capture.go
Alex-Bair: Good call, that does make a lot of sense. I added a `Compare` method and used it here instead. I also updated the only other spot where we're comparing `StreamID`s in `source-sqlserver-ct` so the pattern for doing so is consistent throughout the codebase.

willdonnelly (approved):
LGTM, thanks!

PR #5093, opened 2026-08-20
source-facebook-marketing: remove deprecated connector

Alex-Bair:
**Description:**

`source-facebook-marketing` has been deprecated for 9+ months now, and all active tasks have been moved over to us the native connector instead. All disabled tasks still referencing the deprecated `source-facebook-marketing` are disabled and have been for almost a year, so they're likely abandonded. Even if someone were to try to re-enable them, I suspect re-enabling will fail due to the tasks' recovery logs aging out. The only path forward for any of those users who want to capture Facebook Marketing data is to create a new capture with the native connector.

All that to say, it's safe to remove the `source-facebook-marketing` code from this repo since we don't need to maintain it. I'm leaving the associated package alone since there are various references to it throughout Supabase, and I'm unsure we can safely remove the image from ghrc.io without borking something in the control plane.

**Checklist:**

- [x] If there are user-facing behavior changes, documentation has been updated (see [documentation conventions](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing)), or the docs PR is linked here: [insert link] - (I removed the docs for the `source-facebook-marketing` connector)


nicolaslazo (approved):
LGTM

PR #5096, opened 2026-08-20
ci: don't run Go & Rust connectors' CI workflow when only Python connectors are edited

Alex-Bair:
**Description:**

We've been running the `ci.yaml` workflow for Go & Rust connectors on _any_ changes in the repo. As the repo has grown to include Claude skills, docs, Python CDK & connectors, we've been running the `ci.yaml` workflow when no Go or Rust connector code has changed. This PR makes it so we don't run the Go & Rust connectors' `ci.yaml` workflow when a PR only contains unrelated changes.

~~Additionally, I reduced some duplication in the various `paths` and `paths-ignore` portions by using [YAML anchors](https://yaml.org/spec/1.2.2/#692-node-anchors).~~

I'm hoping that by removing extraneous CI runs, we run into fewer issues with concurrent CI runs fighting with each other for common resources.

**Notes for reviewers:**

~~I chose to use a `paths-ignore` for `ci.yaml` instead of an allow list since we're already in the habit of updating these workflow files any time we add a new Python connector, so updating one more spot shouldn't be a massive burden.~~

I recognize that there's more precise ways to structure CI runs (i.e. only run a connector's workflow when code affecting it changes ex: don't exercise the `source-postgres` tests when only `source-sqs` code changes), but that's a pretty hefty change I'd like to punt on. Don't let perfect be the enemy of better and all that.

**Checklist:**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,11 +7,37 @@ concurrency:
 on:
   push:
     branches: [main]
+    paths-ignore: &non_code_paths
+      - "**.md"
+      - "docs/**"
+      - ".claude/**"
+      - ".github/workflows/python.yaml"
+      - ".github/workflows/estuary-cdk.yaml"
+
   pull_request:
     branches: [main]
+    paths-ignore: *non_code_paths
 
 jobs:
+  # Which connectors did this change touch? This job classifies every changed
+  # path, and the build jobs below gate on its answer.
+  #
+  # It can't be a `paths` filter because dynamically telling a Python connector
+  # from a Go or Rust one means looking for a pyproject.toml, which needs the
+  # checked out tree & nothing is checked out when `paths` filters are evaluated.
+  classify_changes:
+    runs-on: ubuntu-24.04
+    outputs:
+      run: ${{ steps.classify.outputs.go_rust }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - id: classify
+        uses: ./.github/actions/classify-changes
+
   fetch_flow:
+    needs: classify_changes
+    if: needs.classify_changes.outputs.run == 'true'
     runs-on: ubuntu-24.04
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -7,8 +7,83 @@ concurrency:
 on:
   push:
     branches: [main]
+    # Don't run this workflow when Python connectors, docs, or other unrelated
+    # documentation or tools are changed.
+    paths-ignore: &non_go_or_rust_paths
```
willdonnelly: Duplicating this whole big list seems less than ideal, especially considering that it's a blacklist which we'll inevitably forget to keep in sync as new Python connectors are added. Is there any way to deduplicate the big-list-of-Python-connectors part out of both this and the python.yaml `paths` setup so that the two sets stay in sync automatically?

willdonnelly (dismissed):
LGTM % one complaint you probably already considered

.github/workflows/ci.yaml
Alex-Bair: Not super simply, unfortunately. I couldn't find a way to extract the list out of both files without the scope ballooning into more work than I initially wanted to spend time on. And if I were to spend that time on it, it'd be nice to figure out how to not have to maintain these lists of Python connectors at all instead of consolidating them into one spot.

I did end up sinking some more time into dynamically figuring out which workflow should run based off which files are changed in the PR & pushed a new `classify-changes` action that does that for us. It's a little bit of extra complexity, but I don't think it's _too_ terrible, and it'll be nice to not have to maintain a list of Python connectors at all.

(We're still maintaining a separate list of Python connectors in `python.yaml` that includes their versions and usage rates, but I'm hoping we can extract that out of there & into the connectors themselves at some point in the near-ish future).

willdonnelly (approved):
LGTM

PR #5098, opened 2026-08-21
materialize-sql: carry a binding's state key onto tables resolved for Apply

jacobmarble:
**Description:**

Every binding of a materialization has a *state key*: the name the connector
stores that binding's checkpoint state under. It also identifies which
generation of the binding we are on, so it changes when the binding is
backfilled.

`getTable` builds the `Table` value that the shared SQL code hands to a
connector. Two paths call it: `NewTransactor`, for the tables a task writes to,
and Apply, for the tables it creates or alters. Only `NewTransactor` was
attaching the state key afterwards, so a table resolved for Apply arrived with
an empty one.

This moves the assignment inside `getTable`, so both paths get it.

**Workflow steps:**

Nothing changes for users. No connector reads the state key during Apply today,
so this only makes the field available to one that wants to.

**Notes for reviewers:**

- Split out of a larger Snowflake branch to keep that PR focused. A follow-up
  needs the state key at Apply time, to record in each table which generation of
  the binding created it.
- The assignment lives in `getTable` rather than in each caller, so a third
  caller cannot forget it.
- The new test goes through `CreateResource`, which is the Apply path, and
  checks the `TableCreate` handed to the client. It fails on `main`: the state
  key arrives empty.

**Checklist:**

- [x] If fixing a regression, I have linked the PR that introduced the regression: n/a
- [x] If there are user-visible changes, `CHANGELOG.md` has been updated: n/a, shared library only, no user-visible change
- [x] If there are user-facing behavior changes, documentation has been updated: n/a

### DIFF
--- a/materialize-sql/driver.go
+++ b/materialize-sql/driver.go
@@ -454,9 +454,6 @@ func getTable[EC boilerplate.EndpointConfiger, RC boilerplate.Resourcer[RC, EC]]
 	if err != nil {
 		return Table{}, err
 	}
-	// Carried here rather than by each caller, so that a table resolved for an
-	// Apply action names the binding's generation as one resolved for a
-	// transactor does.
 	table.StateKey = binding.StateKey
 
 	return table, nil
--- a/materialize-sql/driver.go
+++ b/materialize-sql/driver.go
@@ -384,7 +384,6 @@ func (s *sqlMaterialization[EC, RC]) NewTransactor(
 		if err != nil {
 			return nil, fmt.Errorf("getting table for binding %d: %w", binding.Index, err)
 		}
-		table.StateKey = binding.StateKey
 		tables = append(tables, table)
 	}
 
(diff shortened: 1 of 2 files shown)


materialize-sql/driver.go
```diff
@@ -451,7 +450,16 @@ func getTable[EC boilerplate.EndpointConfiger, RC boilerplate.Resourcer[RC, EC]]
...
 	}
 	tableShape := BuildTableShape(materializationName, &binding.MaterializationSpec_Binding, binding.Index, path, delta)
-	return ResolveTable(tableShape, endpoint.Dialect)
+	table, err := ResolveTable(tableShape, endpoint.Dialect)
+	if err != nil {
+		return Table{}, err
+	}
+	// Carried here rather than by each caller, so that a table resolved for an
+	// Apply action names the binding's generation as one resolved for a
+	// transactor does.
```
mdibaiee:
```suggestion
```

mdibaiee (dismissed):
LGTM % extra comment

mdibaiee (approved):

PR #5099, opened 2026-08-21
materialize-boilerplate: detect runtime v2, and test a refused materialization

jacobmarble:
**Description:**

Two additions to the shared materialization library. Both are split out of a
larger Snowflake branch, to keep that PR focused.

1. `IsMaterializationSpecRuntimeV2` tells a connector whether the v2
   materialization runtime is driving it. A task runs on the v2 runtime when its
   shard template carries the `enable-runtime-v2` feature flag, so the helper
   reads that label off the spec the runtime hands the connector. The flag name
   is exported too, because a connector that refuses to run without v2 must name
   the flag in its error message — setting it is what the operator has to do.

   The shard template is the right place to read. Every shard of the task is a
   copy of the template with only runtime labels added on top, so a shard's
   feature flags always match the template's.

2. `RunMaterializationRefusalTest` is a test helper for the opposite of the
   usual integration test. It drives each task of a fixture through
   `flowctl preview` and requires the connector to *refuse* it, checking that
   the error message contains the strings you pass. Use it for a fixture whose
   whole point is proving that some configuration is rejected.

   The refusal happens during Apply, before the connector touches the
   destination, so nothing is created and there is nothing to clean up.

**Workflow steps:**

Nothing changes for users. Both are new library functions with no existing
callers.

**Notes for reviewers:**

- The refusal helper checks the message by substring instead of snapshotting it.
  `flowctl`'s failure output carries timestamps and temporary paths, which a
  snapshot cannot hold steady.
- It also logs the whole output, so a test that fails for the wrong reason still
  shows you why.
- Building the `flowctl` command moved into a small shared function, so the
  success and failure helpers set up the environment the same way.

**Checklist:**

- [x] If fixing a regression, I have linked the PR that introduced the regression: n/a
- [x] If there are user-visible changes, `CHANGELOG.md` has been updated: n/a, shared library only, no user-visible change
- [x] If there are user-facing behavior changes, documentation has been updated: n/a

### DIFF
--- a/materialize-boilerplate/runtime.go
+++ b/materialize-boilerplate/runtime.go
@@ -1,0 +1,30 @@
+package boilerplate
+
+import (
+	"github.com/estuary/flow/go/labels"
+	pf "github.com/estuary/flow/go/protocols/flow"
+)
+
+// RuntimeV2FlagName is the name of the task shard flag that puts a task on the
+// v2 runtime. Connectors that refuse to run outside the v2 runtime name it in
+// their error messages, since setting it is the operator's remedy.
+const RuntimeV2FlagName = "enable-runtime-v2"
+
+const runtimeV2Flag = labels.FlagPrefix + RuntimeV2FlagName
+
+// IsMaterializationSpecRuntimeV2 reports whether the task's shard template
+// carries the runtime-v2 feature flag, meaning the connector is being driven by
+// the v2 runtime. Pass either pm.Request_Open.Materialization or
+// pm.Request_Apply.Materialization, which are the specs the runtime hands to a
+// connector.
+//
+// The template is authoritative for the individual shard the connector is
+// running under: each shard spec is a clone of the template with only runtime
+// labels (key range, r-clock range, split source/target, cordon) layered on, so
+// a shard's feature flags are always exactly the template's. The reactor
+// selects its v1 or v2 code path from this same label.
+func IsMaterializationSpecRuntimeV2(spec *pf.MaterializationSpec) bool {
+	return spec != nil &&
+		spec.ShardTemplate != nil &&
+		spec.ShardTemplate.LabelSet.ValueOf(runtimeV2Flag) == "true"
+}
--- a/materialize-boilerplate/runtime.go
+++ b/materialize-boilerplate/runtime.go
@@ -6,23 +6,14 @@ import (
 )
 
 // RuntimeV2FlagName is the name of the task shard flag that puts a task on the
-// v2 runtime. Connectors that refuse to run outside the v2 runtime name it in
-// their error messages, since setting it is the operator's remedy.
+// v2 runtime.
 const RuntimeV2FlagName = "enable-runtime-v2"
 
 const runtimeV2Flag = labels.FlagPrefix + RuntimeV2FlagName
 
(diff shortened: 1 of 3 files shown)


materialize-boilerplate/testutil/test_support.go
```diff
@@ -1216,6 +1262,21 @@ func RunFlowctl(t *testing.T, args ...string) []byte {
...
+func RunFlowctlExpectErr(t *testing.T, args ...string) string {
+	t.Helper()
+
+	output, err := flowctlCommand(args...).CombinedOutput()
+	t.Log(string(output))
+	require.Error(t, err, "flowctl was expected to fail")
+
+	return string(output)
+}
+
```
mdibaiee: This seems really weird to have a function to expect a specific error from flowctl when it is only used once. We could just do `strings.Contains(string(RunFlowctl(...)), "error"` instead of this wrapper

materialize-boilerplate/runtime.go
```diff
@@ -0,0 +1,30 @@
+package boilerplate
+
+import (
+	"github.com/estuary/flow/go/labels"
+	pf "github.com/estuary/flow/go/protocols/flow"
+)
+
+// RuntimeV2FlagName is the name of the task shard flag that puts a task on the
+// v2 runtime. Connectors that refuse to run outside the v2 runtime name it in
+// their error messages, since setting it is the operator's remedy.
```
mdibaiee:
```suggestion
// RuntimeV2FlagName is the name of the task shard flag that puts a task on the
// v2 runtime.
```

materialize-boilerplate/runtime.go
```diff
@@ -0,0 +1,30 @@
...
+const RuntimeV2FlagName = "enable-runtime-v2"
+
+const runtimeV2Flag = labels.FlagPrefix + RuntimeV2FlagName
+
+// IsMaterializationSpecRuntimeV2 reports whether the task's shard template
+// carries the runtime-v2 feature flag, meaning the connector is being driven by
+// the v2 runtime. Pass either pm.Request_Open.Materialization or
+// pm.Request_Apply.Materialization, which are the specs the runtime hands to a
+// connector.
```
mdibaiee:
```suggestion
// the v2 runtime.
```

materialize-boilerplate/runtime.go
```diff
@@ -0,0 +1,30 @@
...
+//
+// The template is authoritative for the individual shard the connector is
+// running under: each shard spec is a clone of the template with only runtime
+// labels (key range, r-clock range, split source/target, cordon) layered on, so
+// a shard's feature flags are always exactly the template's. The reactor
+// selects its v1 or v2 code path from this same label.
```
mdibaiee:
```suggestion
```

mdibaiee (commented):
I find Claude's comments to be not only not-human-readable and time-consuming to read, but also they cause issues down the line because they make claims and assertions that are not evergreen, and end up confusing agents down the line, since they become stale pretty quickly.

I think it is important for us to keep these comments evergreen and human-readable. Too verbose comments can be harmful both for us and for agents.

mdibaiee (approved):

PR #4335, opened 2026-04-29
ci: fix source-kinesis docker-build hang and materialize-bigquery 404 flake

jacobmarble:
**Description:**

Two CI jobs have been red on every recent build. This PR fixes both. Closes #4334.

- **`source-kinesis`** has been hanging the docker-build's `RUN go test` step for the full 10-minute test alarm because `TestDiscover` issues `DeleteStream` against an unreachable LocalStack inside the buildx network. Two fixes: (1) gate the integration tests behind `testing.Short()` and pass `-short` in the Dockerfile (matching `source-mongodb` / `source-mysql-batch` / `materialize-bigquery`), (2) restore the bounded `http.Client` (Dial 10s, ResponseHeader 20s, Timeout 30s) that was removed in `3cd0f0a54`, so a network-dead endpoint fails per-attempt in seconds rather than minutes.
- **`materialize-bigquery`** `TestIntegration/apply` and `/migrate` have been failing intermittently with `Not found: Table … was not found in location us-east1` — BigQuery's metadata layer is eventually consistent, so a `getMetadata` shortly after a `CREATE`/`ALTER` can return 404 even though the table exists. Wrapped `table.Metadata(...)` in `PopulateInfoSchema` with a small bounded retry (5 × 2s, 404-only).

**Workflow steps:**

`./build-local.sh source-kinesis` now completes in ~1 minute (verified locally; previously 13 minutes ending in a panic). `go test -short -v ./source-kinesis/...` skips the 5 integration tests cleanly. `go test -v ./source-kinesis/...` (without `-short`, with `TEST_DATABASE=yes` and a reachable LocalStack/AWS) runs them as before — no behavioral change for the integration test harness path. The bigquery retry is invisible on the happy path; on a 404 it logs a debug message and waits 2s before retrying.

**Documentation links affected:**

None.

**Notes for reviewers:**

- Two commits, one per concern:
  - `source-kinesis: skip integration tests in -short mode and restore HTTP timeouts`
  - `materialize-bigquery: retry table metadata reads on 404`
- The kinesis change keeps `MaxAttempts=20` (production retry budget, intentional per `3cd0f0a54`); the actual fix for the test-time hang is the `-short` skip. The HTTP timeouts are belt-and-suspenders for production network blips.
- Issue #4334 contains the full evidence table (47 runs surveyed) and goroutine stack traces that point at `aws-sdk-go-v2/internal/sdk.sleepWithContext` for the kinesis hang.
[...]

PR #5100, opened 2026-08-21
source-kinesis: get ARN from `ListStreams` response instead of `DescribeStreamSummary`

Alex-Bair:
**Description:**

We've observed auto-discovery and validation consistently failing for a Kinesis capture with:
```
  listing streams: operation error Kinesis: DescribeStreamSummary,
  exceeded maximum number of attempts, 5
```

`listStreams` is describing every stream in the AWS account five at a time to determine their ARNs. But, `DescribeStreamSummary` is capped at 20 TPS per account per region. The connector's bursty `DescribeStreamSummary` calls can easily exhaust the 20 TPS budget. When a stream exhausts its retries, it fails the listing & causes the capture to fail with the error above.

This PR fixes this by taking ARNs from the `ListStreams` response instead of making `DescribeStreamSummary` calls for each stream. The `ListStreams` response already carries a `StreamSummaries` entry per stream that contains its ARN, so when it's present, we take the ARN from there. However, `StreamSummaries` isn't a required response field, so any stream arriving without one is still described individually via `DescribeStreamSummary`. We also now rate-limite our `DescribeStreamSummary` calls to remain below the AWS enforced budget.

While I was in the connector, I made a few additional changes:
- **Restored `MaxAttempts` to 20.** The reduction to 5 in #4335 is no longer what prevents the CI hang it was made for. The same PR gated those tests behind `testing.Short()` and passes `-short` in the Dockerfile. The lower budget applied to every AWS call, including the `GetRecords` retries it had been raised for originally.
- **Fixed three latent bugs in the `listStreams` function:** nil dereferences of `HasMoreStreams` and of the described `StreamARN`, and a pagination loop that would re-request the same page forever if `HasMoreStreams` was set without a `NextToken`.
- **Wait for asynchronous stream deletion in tests** before recreating same-named streams, which fails against real AWS with `ResourceInUseException`.
- **Give captures longer to shut down against real AWS**, since the test harness counts `pruneShards`' checkpoint as "data received" and collapses the watchdog to 100ms before any record is read.

Behavior and changes were verified against a real AWS Kinesis instance. Confirmed the ARN taken from the `ListStreams` response is the same as the one we had been getting from `DescribeStreamSummary`.

**Checklist:**

### DIFF
--- a/source-kinesis/capture_test.go
+++ b/source-kinesis/capture_test.go
@@ -572,6 +572,14 @@ func buildGlueHeader(t *testing.T, schemaVersionId string) []byte {
 }
 
 func advanceCapture(t testing.TB, cs *st.CaptureSpec) {
+	// 100ms is enough for the LocalStack emulator, but not for real Kinesis.
+	// The harness counts the checkpoint that pruneShards always emits as data
+	// received, so the shutdown countdown starts before any records have been
+	// read.
+	if useRealAWS() {
+		advanceCaptureWithTimeout(t, cs, 10*time.Second)
+		return
+	}
 	advanceCaptureWithTimeout(t, cs, 100*time.Millisecond)
 }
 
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
LGTM

PR #5106, opened 2026-08-21
estuary-cdk: undo chained content codings left on response bodies

nicolaslazo:
**Description:**

aiohttp auto-decompresses a response only when Content-Encoding holds a single coding it knows; a chained value like `gzip,gzip` matches nothing in its parser and the body arrives still compressed. Parse the codings aiohttp skipped and undo them ourselves, both on the streaming success path and on error bodies (which fall back to the raw bytes, with a warning, rather than fail while reporting the HTTP error).

Edge-case handling deliberately mirrors aiohttp's `DeflateBuffer`.

**Workflow steps:**

**Notes for reviewers:**

**Checklist:**

### DIFF
--- a/estuary-cdk/estuary_cdk/content_encoding.py
+++ b/estuary-cdk/estuary_cdk/content_encoding.py
@@ -1,0 +1,235 @@
+"""
+Undo the HTTP content codings that aiohttp leaves on a response body.
+
+aiohttp auto-decompresses only when `Content-Encoding` holds a single coding. A
+chained value like `gzip,gzip` matches nothing in its parser, so the body
+arrives with every layer still on it.
+"""
+
+from collections.abc import AsyncGenerator, AsyncIterable
+
+from aiohttp import ClientRequest, hdrs
+from aiohttp.compression_utils import (
+    HAS_BROTLI,
+    HAS_ZSTD,
+    BrotliDecompressor,
+    DecompressionBaseHandler,
+    ZLibDecompressor,
+    ZSTDDecompressor,
+)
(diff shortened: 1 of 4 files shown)


Alex-Bair (commented):
I've thought about this some more, and I'm not convinced we should make a CDK-wide change to decompress an error message we're seeing for a single connector. We've gotten this far without needing this double-decompression behavior, and I'm wary of implementing something that ideally would be handled/configured in `aiohttp` instead based off a single case where it'd be useful for diagnostics.

Some examples of concerns I have with the changes in the PR to help explain why I'd rather we leave decompression behavior & handling the associated complexity to `aiohttp`:
- There's no ceiling on expansion. The implementation in this PR caps how many layers are decompressed but not how big the output gets. `aiohttp` handles this already by stopping at 32 MiB.
- We're depending on `aiohttp` internals that may change underneath us.
- Unrecognized encodings now fail the request, where `aiohttp` disregards them & allows requests to succeed.

Have we tried telling the Smartsheet API not to encode responses at all with an [`Accept-Encoding: identity`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding#identity) request header? If Smartsheet honors it, the error body should come back readable as-is. That'd avoid the double decompression problem entirely. If we want Smartsheet to still compress responses, we could gate the `Accept-Encoding: identity` behind a toggle in the config & only enable it when captures are consistently hitting errors that need investigating.

I'd recommend investigating that route first before making CDK-wide changes for a single instance where this double-decompression behavior is needed.

PR #5108, opened 2026-08-23
fix(facebook-marketing-native): retain reduced page size

timkeeeeeen:
Meta can return a recoverable server error asking clients to reduce the requested page size. The connector already retries with a smaller limit, but subsequent paginated requests restart at the configured limit and repeat the same failed request.\n\nCache the smallest successful limit per resource URL for the lifetime of the client and reuse it on later requests. This keeps the existing retry behavior while avoiding repeated oversized requests.\n\nValidation: source-facebook-marketing-native unit tests pass (15 tests).

### DIFF
--- a/source-facebook-marketing-native/source_facebook_marketing_native/client.py
+++ b/source-facebook-marketing-native/source_facebook_marketing_native/client.py
@@ -220,6 +220,7 @@ class FacebookAPIClient:
         self.base_url = BASE_URL
         self.log = log
         self.permission_manager = PermissionManager(http, BASE_URL, log)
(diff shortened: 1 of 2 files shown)


timkeeeeeen:
This is a focused two-file retry fix with a regression test. The fork-origin workflows have not started; a maintainer workflow approval/review would be appreciated. The connector unit suite passes locally (15 tests).

timkeeeeeen:
@Alex-Bair @JustinASmith — could one of you approve the fork workflows and review this focused Facebook Marketing retry fix when convenient? It addresses repeated oversized-page retries observed in a live capture and includes a regression test.

PR #5115, opened 2026-08-24
sqlcapture: track dirty stream states in an explicit set

Alex-Bair:
**Description:**

After merging the improvements in https://github.com/estuary/connectors/pull/5088, it was observed that `emitState` is another performance bottleneck for a capture with 40,500 enabled bindings since its speed also scales based off of how many bindings are being captured.

`emitState` decided what belonged in a checkpoint by scanning all of `State.Streams` for entries with the dirty flag set. Since a checkpoint is emitted for every database transaction commit, that made the per-commit cost scale with a capture's binding count rather than with the number of tables which actually changed.

To avoid scaling `emitState`'s speed when there are many bindings being captured, the dirty state keys are now kept in a set on the `Capture` and `emitState` walks that instead of `State.Streams`, which makes it proportional to the number of tables which changed and flat in binding count. Benchmarking the change showed the following improvement for `emitState`:

|    bindings   |   before  |    after |
| - | - | - |
|     1,000    |   7.45 us  |  0.41 us |
|    10,000   |   77.80 us  |  0.32 us |
|    40,500   |  324.00 us |   0.38 us |
|    81,000  |  1248.00 us  |  0.36 us |

The emitted checkpoints don't change. An empty dirty set now yields a nil `Streams` map rather than an empty one, which `omitempty` omits identically.

Motivating Slack thread: [link](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1787538563575619?thread_ts=1787351843.925139&cid=C03Q2NRFKDL)

**Checklist:**

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -153,9 +153,6 @@ type TableState struct {
 	Metadata json.RawMessage `json:"metadata,omitempty"`
 	// BackfilledCount is a counter of the number of rows backfilled.
 	BackfilledCount int `json:"backfilled"`
-	// dirty is set whenever the table state changes, and cleared whenever
-	// a state update is emitted. It should never be serialized itself.
-	dirty bool
 }
 
 const (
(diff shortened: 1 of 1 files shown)


willdonnelly (approved):
LGTM, thanks for handling this

PR #5119, opened 2026-08-24
sqlcapture: Reduce main loop overhead on large captures

willdonnelly:
**Description:**

I'm not entirely sure whether this change is worthwhile. The headline number here is that at 50k bindings, a capture with an infinitely fast database and output sink previously could only run 39 backfill chunk queries per second (~26ms overhead per chunk) and with these changes that increases to 307 (~3ms overhead per chunk).

This is objectively an improvement and we're no longer doing a bunch of redundant work reconstructing the set of tables currently backfilling each time. But on the other hand, it's ~26ms per backfill chunk and ideally each backfill chunk should take, like, 100 times that long at minimum and this difference is essentially noise.

On the third hand, 26ms per chunk of overhead means that in any backfill of 50k bindings, there will be 22 minutes wasted entirely on a bunch of redundant computation inside our connector. If each table in the source DB is individually small, this could be a non-negligible fraction of the overall backfill time.

Anyway, this commit makes two changes:

- The "Backfilling Tables" status update now runs at most once each ten seconds. Before, the full progress computation ran after every chunk.
- A new backfillingStreams set records the streams that are currently in a backfill mode, so we can just use that instead of recomputing it over and over.

Some of the helper functions seemed kind of pointless after these changes, so `BindingsCurrentlyBackfilling` (sorted version) has been removed, as is the `bindingsInState` iterator (no longer used for anything distinct from BindingsInState).

Benchmark results:

**Main Loop: Backfill**

| bindings | baseline | final | speedup | allocs/iter |
|---|---|---|---|---|
| 100 | 25,872 chunks/s | 57,937 | 2.2x | 31KB → 9.9KB |
| 1,000 | 3,048 | 18,114 | 5.9x | 317KB → 47KB |
| 10,000 | 252 | 1,995 | 7.9x | 3.5MB → 417KB |
| 50,000 | 38.9 | 307 | 7.9x | 17.5MB → 2.1MB |
| 100,000 | 15.5 | 119 | 7.7x | 35.2MB → 4.3MB |

**Main Loop: Streaming**

| workload | bindings | baseline | final | speedup |
|---|---|---|---|---|
| 1,000 commits × 1 row | 100 | 2.35M | 2.27M | ~1x |
| | 10,000 | 695k | 2.03M | 2.9x |
| | 100,000 | 60.6k | 1.58M | 26x |
| 100 commits × 100 rows | 100 | 16.2M | 15.3M | ~1x |
| | 100,000 | 574k | 4.60M | 8.0x |
| 10 commits × 10,000 rows | 100 | 18.5M | 17.6M | ~1x |
| | 100,000 | 2.86M | 4.73M | 1.7x |

At 100,000 bindings, the framework overhead for one backfill chunk went from 64ms to 8.4ms, and the bookkeeping cost of a streaming cycle grows less sharply as binding count increases.

**Workflow steps:**

No action required.

**Checklist:**

### DIFF
--- a/sqlcapture/capture.go
+++ b/sqlcapture/capture.go
@@ -196,6 +196,11 @@ type Capture struct {
 	Output   *boilerplate.PullOutput // The encoder to which records and state updates are written
 	Database Database                // The database-specific interface which is operated by the generic Capture logic
 
+	replStream          ReplicationStream           // The replication stream from which the main capture loop reads change events
+	lastDiscovery       map[StreamID]*DiscoveryInfo // The most recent table discovery results, refreshed on every rediscovery
+	rediscoverAfter     time.Time                   // When the next table rediscovery should occur
+	periodicChecksAfter time.Time                   // When the next periodic database checks should occur
+
 	lastStatus string // Last connectorStatus update. Used to suppress exact duplicates to reduce log noise.
 
 	// dirtyStreams is the set of stream state keys changed since the last checkpoint.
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM, thanks! These looks like they're generally good improvements we should make / we'll need to make at some point anyways, so I'd vote to merge them.

PR #5120, opened 2026-08-24
tests/benchmark: run materialize arms with more than one shard

jacobmarble:
**Description:**

Lets the materialize benchmark run a task with more than one shard, so we can measure how a connector scales.

Six changes:

1. **`--shards N`.** Above one shard the harness switches from `flowctl preview` to `flowctl raw preview-next`, because only the latter can drive several shards. It emits no `connectorState` lines, so timing now comes from the runtime's own per-transaction commit log instead.
2. **Runtime picked by flag, not shard count.** A spec carrying `enable-runtime-v2=true` now runs on `preview-next` at any shard count. Before, a single-shard run set that flag but ran the V1 runtime — and materialize-snowflake's streaming v2 gate believes the flag, so it took its v2 path under V1. That combination cannot happen in production, so we were measuring something unreal.
3. **Three new knobs:** `--shard-flag`, `--table-suffix` (two arms of one comparison can share a scenario without sharing a table), and `--shard-log-level`.
4. **A timestamp fix.** Some log lines have whole-second timestamps with no fraction. The parser padded them to 26 characters, which put zeros after the `Z`, and `strptime` raised a `ValueError` that killed the whole results block. A run could finish after several hours and write no numbers at all. Every streaming v2 run hits this.
5. **`--boot-disk-size SIZE`.** `create` hardcoded a 100 GB boot disk, which a 200 GiB run exhausted at round 25 of 40 with `writing fixture log block: No space left on device`. estuary/flow#3408 has since bounded the fixture log to 512 MiB plus the transaction in flight, so 100 GB is now ample and this is just a knob for transactions much larger than these. Default is unchanged at 100 GB.
6. **A longer SSH wait.** `wait_for_ssh` gave IAP 120 s, which was short enough that `create` failed on every VM and left a running, unprovisioned instance behind. Now 300 s.

Also adds delta-updates scenarios at 1 KB and 1 MB documents, and bumps the VM setup to install `python3-venv` (the Snowpipe sidecar needs it).

**Workflow steps:**

Unchanged by default. To run multi-shard:

```
./tests/benchmark/materialize/run.sh \
  --connector materialize-snowflake \
  --scenario tests/benchmark/materialize/scenarios/50gb-delta-1mb.yaml \
  --shard-flag enable-runtime-v2=true --shards 4
```

**Notes for reviewers:**

- The last two commits are `gcp-vm.sh` fixes found while running the benchmark on #4828; like the first commit they stand alone.
- Multi-shard needs a flowctl built from flow `master`, not a release: estuary/flow#3391 (the 65,536-documents-per-shard panic), #3408 (unbounded fixture memory and disk) and #3410 (a dead fixture feeder hanging the run instead of failing it) have all merged, but the newest release, v0.6.13, predates all three.
- The first commit fixes a bug already on `main`, so it stands alone if you'd rather take it separately.
- Verified end to end: a 4-shard run against Snowflake completes with all rounds measured and four channels opened. Full 50 GiB results are on #4828.

**Checklist:**

- [x] If there are user-visible changes, `CHANGELOG.md` has been updated — n/a, test tooling only
- [x] If there are user-facing behavior changes, documentation has been updated — n/a, no connector behavior changes


tests/benchmark/gcp-vm.sh
```diff
@@ -22,6 +22,9 @@
 #
 # Create flags:
 #   --machine-type TYPE   Machine type (default: e2-standard-2)
+#   --boot-disk-size SIZE Boot disk size (default: 100GB). `preview-next` writes
+#                         the whole fixture to an on-disk log, so a scenario
+#                         needs a disk larger than the bytes it moves.
```
mdibaiee: Is this still true? we cannot stream the fixture to v2?

mdibaiee (dismissed):
LGTM % one question

tests/benchmark/gcp-vm.sh
jacobmarble: Still true, though the comment described it badly — I reworded it in 6394eb8f2.

It does stream: the fixture goes through the same on-disk shuffle log as the live V2 path. The problem is that the fixture writer doesn't honor the log's *disk* backpressure gate. `LogActor` (`crates/shuffle/src/log/actor.rs:103`) gates every append on `!block.is_full() && !disk_back_pressure`, where `disk_back_pressure` trips once sealed segments pass `shuffle_disk_limit_bytes` (128 MiB default) and releases at half that. Since segments are unlinked by the consumer, the live writer ends up paced by consumer progress. The fixture calls `Writer::append_block` directly (`flowctl/src/raw/preview_next/fixture.rs:683,709`) and consults neither gate, so the generator runs arbitrarily far ahead of the connector and sealed segments pile up.

estuary/flow#3391 restored the `is_full()` equivalent — that's the 65,536-entry `block::encode` assert — but not the backlog gate, so the disk growth is unchanged. Measured at roughly 4 GiB of resident log per 5 GiB round; a 200 GiB scenario wanted ~160 GiB of disk, and a 100 GB boot disk died at round 25/40 with `writing fixture log block: No space left on device`. V1 `preview` has no shuffle log at all — the connector backpressures the generator through the FIFO — which is why this is new.

The real fix is a disk-backlog gate in the fixture writer, which I suspect would also fix the unbounded RSS (21.7 GB on the 1 KB-document arms). That's flow-side, so this flag is the stopgap.

mdibaiee: Is this something we are going to fix on Flow side? I saw a bunch of PRs on Flow but not sure if this is also included

jacobmarble: You're right, I'm getting ahead of myself. I'll update this PR after https://github.com/estuary/flow/issues/3399 is closed, then ping you directly.

jacobmarble: Fixed on the Flow side: estuary/flow#3408 bounds the fixture log to 512 MiB plus the in-flight transaction (and peak RSS to ~500 MB); estuary/flow#3410 makes a dead feeder fail instead of hang.

So I reworded the `--boot-disk-size` comment in 24986f73f — 100 GB is ample now — and the reviewer note points at a `master` build, since v0.6.13 predates all three PRs.

tests/benchmark/gcp-vm.sh
```diff
@@ -38,6 +42,7 @@ USER_LABEL="$(echo "${USER:-unknown}" | tr '[:upper:]' '[:lower:]' | sed 's/[^a-
 VM_NAME="bench-${USER_LABEL}"
 ZONE="us-central1-a"
 MACHINE_TYPE="c3-standard-4-lssd"
+BOOT_DISK_SIZE="100GB"
```
mdibaiee, line 45: Was the default 100GB prior to this PR as well ? Seems like a lot now that we have bounded disk usage

tests/benchmark/materialize/run.sh
```diff
@@ -366,13 +440,64 @@ def boundaries_from_log(path):
     return apply_s, bounds, len(rounds)
 
 
+def boundaries_from_log_v2(path):
```
danielnelson: I always get a large `MB/s` reading for the first transaction:
```
Tx  Collection    Docs  Doc Size   Total   Wall    MB/s  Fresh  Overlaps
────────────────────────────────────────────────────────────────────────
#0  bench/medium   500      1 MB  500 MB   0.2s  2676.6   500c  none    
#1  bench/medium   500      1 MB  500 MB  11.5s    43.4   300c  200u@tx0
```

tests/benchmark/gcp-vm.sh
```diff
@@ -179,14 +186,14 @@ cmd_create() {
...
 
-  info "creating VM: $VM_NAME (zone=$ZONE, type=$MACHINE_TYPE)"
+  info "creating VM: $VM_NAME (zone=$ZONE, type=$MACHINE_TYPE, disk=$BOOT_DISK_SIZE)"
   gcloud compute instances create "$VM_NAME" \
     --zone="$ZONE" \
     --project="$PROJECT" \
     --machine-type="$MACHINE_TYPE" \
     --image-family=ubuntu-2404-lts-amd64 \
     --image-project=ubuntu-os-cloud \
-    --boot-disk-size=100GB \
```
jacobmarble, line 189: *

tests/benchmark/gcp-vm.sh
jacobmarble: Yes, 100GB was the hard-coded value before.

mdibaiee (dismissed):

tests/benchmark/materialize/run.sh
jacobmarble: I've confirmed this is a real off-by-one error.

jacobmarble: @danielnelson would you try again? I've just committed a fix.

danielnelson: I'm still slightly suspicious of my first round results, they always seem quite a bit faster, but they are plausible now.  With an extra transaction added in to the medium scenario:
```
Tx  Collection    Docs  Doc Size   Total   Wall   MB/s  Fresh  Overlaps
───────────────────────────────────────────────────────────────────────
#0  bench/medium   500      1 MB  500 MB   4.4s  114.2   500c  none    
#1  bench/medium   500      1 MB  500 MB  11.3s   44.4   500c  none    
#2  bench/medium   500      1 MB  500 MB  13.3s   37.5   300c  200u@tx0
```

danielnelson (approved):

PR #5121, opened 2026-08-25
materialize-bigquery: add support for scale out

### LINKED ISSUE #4716: Materialization scale-out: BigQuery
Implement runtime v2's materialization scaleout for bigquery

danielnelson:
**Description:**

Add support for scaling via shard splitting.

All shards take part in loading and staging updates.  The checkpoint format is changed to hold the information needed to expand the query template, instead of the query itself.  Each checkpoint now maps from `{stateKey}` -> `{rangeKey}` -> `checkpointItem`.

During the Acknowledge phase, a query for each binding is generated and ran on the primary shard.  Doing this prevents errors due to concurrent table updates.

closes #4716

#### Testing

I ran `tests/soak` from the flow repo, and checked the results for issues.  Since this connector does not update atomically, you must check only between Acknowledge operations.

The `ci:consistency` checks pass:
```
Nextest run ID 44757471-78b7-4b15-bbc8-014781f6007e with nextest profile: consistency
    ...
────────────
    Summary [ 849.590s] 17 tests run: 17 passed (15 slow), 0 skipped
```

I also injected a failure into Acknowledge and then upgraded the connector before recovery, to ensure that the connector can be upgraded to the new checkpoint format.

#### Performance

The numbers below are all using `--shard-flag enable-runtime-v2=true`.  Without this flag the baseline is faster but of course the multi-shard variants can't be ran.  I used a modified flowctl to avoid OOM errors during the small-docs test.

##### small-docs.yaml
| arm | shards | throughput | wall |
|---|---|---|---|
| OLD | 1 | 3.9 MB/s | 25m50s |
| NEW | 1 | 3.9 MB/s | 26m4s |
| NEW | 2 | 6.9 MB/s | 14m44s |
| NEW | 4 | 9.8 MB/s | 10m24s |
| NEW | 8 | 10.0 MB/s | 10m13s |

##### medium-docs.yaml
| arm | shards | throughput | wall |
|---|---|---|---|
| OLD | 1 | 40.2 MB/s | 24.8s |
| NEW | 1 | 41.3 MB/s | 24.2s |
| NEW | 2 | 60.5 MB/s | 16.5s |
| NEW | 4 | 82.6 MB/s | 12.1s |
| NEW | 8 | 92.9 MB/s | 10.8s |

##### large-docs.yaml
| arm | shards | throughput | wall |
|---|---|---|---|
| OLD | 1 | 60.7 MB/s | 2m45s |
| NEW | 1 | 64.6 MB/s | 2m35s |
| NEW | 2 | 77.6 MB/s | 2m9s |
| NEW | 4 | 102.0 MB/s | 1m38s |
| NEW | 8 | 144.7 MB/s | 1m9s |
| NEW | 16 | 151.7 MB/s | 1m6s |

**Workflow steps:**

No configuration change needed to the connector, the new behavior is used on all instances.  In order to take advantage of multiple shards you need to `flowctl raw split-shards`.

**Notes for reviewers:**

**Checklist:**


materialize-bigquery/transactor.go
```diff
@@ -85,6 +129,16 @@ func prepareNewTransactor(
 			return nil, fmt.Errorf("creating GCS bucket: %w", err)
 		}
 
+		var keyBegin, keyEnd uint32 = 0, math.MaxUint32
```
mdibaiee, line 129: Do we need this default? can `open.Range` be null? seems unexpected 🤔

materialize-bigquery/transactor.go
```diff
@@ -85,6 +129,16 @@ func prepareNewTransactor(
...
+		if open.Range != nil {
+			keyBegin, keyEnd = open.Range.KeyBegin, open.Range.KeyEnd
+		}
+
+		primaryShard := true
+		if boilerplate.IsMaterializationSpecRuntimeV2(open.Materialization) {
+			primaryShard = keyBegin == 0
+		}
```
mdibaiee, line 137: Is this condition necessary? can we just have `primaryShard = keyBegin == 0`? on runtime v1 it is the same definition

materialize-bigquery/transactor.go
```diff
@@ -409,21 +464,99 @@ func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
...
 }
 
+func isJSONNull(data json.RawMessage) bool {
+	return bytes.Equal(bytes.TrimSpace(data), []byte("null"))
+}
+
+// mergeStatePatches merges the RFC 7396 JSON Merge Patch in patches into the
+// target checkpoint, and returns the merged checkpoint.
+func mergeStatePatches(target checkpoint, patches []json.RawMessage) (checkpoint, error) {
+	checkpointJSON, err := json.Marshal(target)
```
mdibaiee: I believe `json.Marshal` won't ever return `"null"` unless there is a nil pointer

materialize-bigquery/transactor.go
```diff
@@ -409,21 +464,99 @@ func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, patch := range patches {
+		// The jsonpatch library returns an error if the target is "null", so
+		// we set it back to {} to better match RFC 7396.
+		if isJSONNull(checkpointJSON) {
```
mdibaiee: so this is not necessary I think (I might have to revisit my own Databricks code for this)

materialize-bigquery/transactor.go
```diff
@@ -409,21 +464,99 @@ func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return merged, nil
+}
+
+// groupCheckpointsByBinding yields the checkpoint keys that belong to each
+// individual binding.
+func groupCheckpointsByBinding(cp checkpoint) iter.Seq[[]CheckpointKey] {
```
mdibaiee: this iterator idea is nice!

mdibaiee (commented):
Thanks! a few small nits, feel free to skip

materialize-bigquery/transactor.go
mdibaiee: https://go.dev/play/p/ei0f88U8EvQ?v=goprev

materialize-bigquery/transactor.go
danielnelson: Technically this is correct, but I keep the check because:
```
jsonpatch.MergePatch(`{"foo": "bar"}`, `null`) -> `null`
```

So if `checkpointJSON` became null in the loop it would crash on the next patch.  But we can only get a null patch if we use `MergePatch:false` checkpoints, which would be a different problem, so this should never hit in practice.

In this situation the library behavior does not match the RFC, which allows the target to be null, and I thought it would be nice to make this function behave more as expected from a general merge patch function.

P.S. In `encoding/json/v2`, the stdlib uses the merge patch same semantics, so I think in the future we no longer need a library for this, from the package docs:
> In general, unmarshaling follows merge semantics (similar to RFC 7396) where the decoded Go value replaces the destination value for any JSON kind other than an object. For JSON objects, the input object is merged into the destination value where matching object members recursively apply merge semantics.

materialize-bigquery/transactor.go
danielnelson: >	Range *flow.RangeSpec `protobuf:"bytes,3,opt,name=range,proto3" json:"range,omitempty"`

It's a pointer and omitempty so I think we should keep the check, even if the runtime always sets it.  One thing I would love to do in the future is parse the `pm.Request_Open` into a struct and pass that into connectors.  Then we can have stronger guarantees than the PB gives us on the state of the struct.

materialize-bigquery/transactor.go
danielnelson: With V1, my understanding is that each shard would have it's own checkpoint, they wouldn't be provided as statePatches to Acknowledge.  So they would need to all work as if they were a primary.  However, it doesn't seem like this would actually work very well in practice, I expect that there would be a lot of concurrency modification errors in Acknowledge.

It may be better to just block it on v1:
```
		primaryShard := keyBegin == 0
		if !primaryShard && !boilerplate.IsMaterializationSpecRuntimeV2(open.Materialization) {
			return nil, fmt.Errorf("shard scale out not supported on runtime v1")
		}
```

mdibaiee: It is effectively not possible to run v1 runtime tasks multi-shard. If it feels safer, we can have this hard error and then we can remove it once the switch to v2 is done

materialize-bigquery/transactor.go
```diff
@@ -23,15 +29,47 @@ import (
...
+type CheckpointKey string
+
+func NewCheckpointKey(stateKey string, rangeKey string) CheckpointKey {
+	return CheckpointKey(stateKey + ":" + rangeKey)
+}
+
+// Return the StateKey portion of the key which identifies only the binding.
+func (k CheckpointKey) StateKey() string {
+	return strings.SplitN(string(k), ":", 2)[0]
+}
```
mdibaiee: I don't have a strong preference between this format and the nested object format used by databricks. For illustration this is the difference:

Nested (Databricks):
```
{ "myBinding.v1": { "00000-888888": {....}, "88888-fffffff": {....} } }
```

Colons (BigQuery):
```
{ "myBinding.v1:00000-888888": {....}, "myBinding.v1:88888-fffffff": {....} }
```

But I definitely want us to have one uniform format across connectors. What is your reasoning for preferring the colons format over nested format?

@jacobmarble any thoughts on these two?

materialize-bigquery/transactor.go
mdibaiee: I believe by using the nested checkpoint format, we don't need this anymore, because the checkpoint is already a map keyed on the binding stateKey, so that each binding's stateKey is a list of its ranges

mdibaiee (commented):
Overall LGTM, one point about checkpoint format that we need to settle

materialize-bigquery/transactor.go
danielnelson: I used this format because it bridges well across the old and new checkpoint styles.  There were fewer situations where I needed to have separate code for each version.

materialize-bigquery/transactor.go
```diff
@@ -409,21 +459,99 @@ func (t *transactor) Store(it *m.StoreIterator) (m.StartCommitFunc, error) {
...
+			if shouldProcess(b.target.StateKey) {
+				cpKey := NewCheckpointKey(b.target.StateKey, t.rangeKey)
+				delete(t.cp, cpKey)
+			}
+		}
+		return nil, nil
+	}
+
+	var err error
+	t.cp, err = mergeStatePatches(t.cp, statePatches)
```
jacobmarble: Is it safe if `t.cp` get the zero value when `err` is returned? Probably yes, but just checking.

jacobmarble (dismissed):
LGTM. Why do the benchmark throughput values not scale linearly with shards?

materialize-bigquery/transactor.go
mdibaiee: If we look into the future where we don't have to maintain the old checkpoint format (all tasks have migrated after 20 days), which one would be easier for us to keep do you think?

danielnelson: Probably `map[string]map[string]checkpointItem` is better long term.  However, I'm not sure we can drop this format until the oldest task is created after the change over.  Disabled checkpoints and prior backfills stay in the recovery log indefinitely.  Perhaps we need to add some sort of checkpoint cleanup?

danielnelson:
> Why do the benchmark throughput values not scale linearly with shards?

I assume that this is because of the fixed Acknowledge cost.  Load and Store gets faster but Acknowledge stays the same.  It would be nice to add timing metrics for each Transactor step to verify this.

materialize-bigquery/transactor.go
danielnelson: I believe it is safe because an error is fatal and the checkpoint is never read again, but I don't love this style because it makes you think about it.  I'll switch this to an assignment after the error is checked.

materialize-bigquery/transactor.go
jacobmarble: Nested field (Snowflake streaming v2):                                                                                                                                      

    { "myBinding.v1": { "StreamV2": { "myMat_00000000-88888888_myBinding_v1": {....}, "myMat_88888888-ffffffff_myBinding_v1": {....} } } }

It would be difficult to merge this with either proposal mentioned above. The range key includes other information so that it can be directly used as a Snowflake channel name without conflict.

jacobmarble: For Snowflake in particular, the `StreamV2` layer is a pattern already baked in: https://github.com/estuary/connectors/blob/c374b393b96d99d09f2cb168dee8d09032af964a/materialize-snowflake/snowflake.go#L618-L627

Then we add this field:

```golang
	StreamV2      map[string]*streamV2Item `json:",omitempty"` // keyed by channel name
````

jacobmarble: I can change the key of the Snowpipe v2 checkpoint map to nested/Databricks style without trouble. The middle `StreamV2` layer is stuck, but at least the first and third layers will resonate with another connector.

jacobmarble (approved):

mdibaiee (approved):

PR #5126, opened 2026-08-25
tests: drive capture snapshot tests with `flowctl raw preview-next`

jgraettinger:
**Description:**

Moves all 23 Python capture snapshot tests from `flowctl preview` (V1 runtime) to `flowctl raw preview-next` (V2 runtime). This is the capture-side counterpart to the materialization migration already underway in `materialize-boilerplate/testutil/test_support_v2.go`.

The two connectors whose capture test is currently commented out — `source-twilio` and `source-google-ads` — are updated in place, so they don't come back stale if re-enabled.

Also updates the prose in the capture skills, agents, and shared rules that tell contributors and agents to run this command, plus the docstring in `source-brevo`'s cursor unit tests. `README.md`'s `flowctl preview` references are deliberately left alone — those are entirely about materialization fixtures, which is a separate migration.

**Workflow steps:**

No change in how the tests are invoked — `pytest` still runs them the same way. The command they shell out to underneath is now `flowctl raw preview-next`, which runs the task on the V2 runtime.

**Notes for reviewers:**

The flag surface these tests use is compatible between the two commands. Diffing both `Preview` clap structs on flow master: `--sessions` keeps identical semantics in V2.

Output is equivalent too: both runtimes against `source-hello-world` with a throwaway spec:

| Shape | v1 `preview` | v2 `raw preview-next` |
|---|---|---|
| `--sessions 1 --delay 5s`, 1 binding | exit 0, 25 docs | exit 0, 25 docs — byte-identical modulo timestamps |
| `--sessions 1 --delay 4s`, 3 bindings | exit 0, 36 docs | exit 0, 36 docs, same stream order |
| `--sessions 3 --delay 2s` (the `--sessions 27` shape `source-zendesk-support` uses) | exit 0, 30 docs | exit 0, 30 docs |

The multi-binding case is the one that matters most, since most of these tests snapshot "first document per stream in first-appearance order": both runtimes emit bindings in **spec order** (not collection-name order) and key-sort within a collection. So existing snapshots should hold without regeneration — which is what CI of this PR is really testing.

**Checklist:**

- [ ] If there are user-visible changes, `CHANGELOG.md` has been updated — n/a, test-harness only
- [ ] If there are user-facing behavior changes, documentation has been updated — n/a


Alex-Bair (approved):
LGTM

PR #5131, opened 2026-08-26
source-sqlserver: add IAM auth support

nicolaslazo:
**Description:**

Existing configurations keep their top-level `password`, which is now hidden from the schema and folded into the union as a `UserPassword` credential by `normalizeCredentials`, so everything downstream of config loading reads exactly one credentials shape.

**Workflow steps:**

**Notes for reviewers:**

**Checklist:**

### DIFF
--- a/source-sqlserver/main.go
+++ b/source-sqlserver/main.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go-v2/feature/rds/auth"
+	"github.com/estuary/connectors/go/auth/iam"
 	"github.com/estuary/connectors/go/capture/sqlserver/datatypes"
 	"github.com/estuary/connectors/go/common"
 	cerrors "github.com/estuary/connectors/go/connector-errors"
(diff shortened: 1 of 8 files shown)


source-sqlserver/main.go
```diff
@@ -189,15 +266,44 @@ func (c *Config) ToURI() string {
...
 	params.Add("database", c.Database)
+
+	var userInfo *url.Userinfo
+	switch c.Credentials.AuthType {
+	case AWSIAM:
+		userInfo = url.User(c.User)
+	default:
+		userInfo = url.UserPassword(c.User, c.Credentials.Password)
+	}
+
```
Alex-Bair, line 279: nit: could we use explicit `case`s for each anticipated `AuthType` rather than falling back to username & password authentication? I usually prefer this type of exhaustiveness checking whenever we have a small set of possible `case`s; it's a little clearer in the code, plus it'll help if we add another `AuthType` later but forget to update this switch/case with it; we'd get a clearer error like "This AuthType isn't supported" or something rather than some username & password related error.

For an example, here's the same type of switch/case from `source-postgres`:
https://github.com/estuary/connectors/blob/70d77a396a2f03af2ffa8ff32a74dc8571482f19/source-postgres/main.go#L346-L372

Alex-Bair (dismissed):
LGTM % one small nit.

Alex-Bair:
Also, it looks like the CI check failed, so that will need fixed before merging.

nicolaslazo:
> Also, it looks like the CI check failed, so that will need fixed before merging.

That same issue exists in `main`:

```
time="2026-08-26T17:44:20Z" level=error msg="servePrimary failed" err="completeRecovery: app.NewStore: creating V2 task service: FLOW_DATA_PLANE_FQDN not set\n\nCaused by:\n    environment variable not found" shard="/gazette/consumers/*runtime.FlowConsumer/items/capture/tests/source-sqlserver/from-source/ffffffffffffffff/00000000-00000000"
```

[Source here](https://github.com/estuary/connectors/actions/runs/32994737558/job/98261520970). I'll try to investigate in the morning

Alex-Bair:
Ah, good catch, I just saw CI failed & didn't dig any deeper. This is a v2 runtime issue. I asked the data plane team to take a look in [Slack](https://estuaryworkspace.slack.com/archives/C03Q9A3GFEE/p1787797729296369).

Until that's fixed, I _think_ moving the tests back onto the v1 runtime should unblock CI jobs. We can do that by disabling the v2 runtime in https://github.com/estuary/connectors/blob/main/tests/template.flow.yaml. Doing that should look something like:
```yaml
  captures:
    tests/${CONNECTOR}/from-source:
      shards:
        flags:
          enable-runtime-v2: "false"
      endpoint:
        connector:
          image: "${CONNECTOR_IMAGE}"
          config: ${CONNECTOR_CONFIG}
```

Alex-Bair (approved):
LGTM!

PR #5134, opened 2026-08-26
go/capture: drive Go capture tests with `flowctl raw preview-next`

jgraettinger:
**Description:**

Switches the Go capture blackbox harness from `flowctl preview` to `flowctl raw
preview-next`, so Go captures run against the V2 runtime. All of the code change
is in `go/capture/blackbox/capture.go`; every Go capture suite picks it up.

Two things are needed beyond swapping the command:

- **Lockstep transactions.** V2 runs the production transaction-close policy, so
  a capture emitting N checkpoints yields anywhere from 1 to N transactions.
  Batching corrupts snapshots two ways: documents in one transaction are emitted
  in collection-key order rather than emission order, and two changes to the same
  key reduce to one document. `withLockstepTransactions` rewrites the catalog on
  the way out, injecting `shards.maxTxnDuration: 1ns` into every capture, which
  admits exactly one checkpoint sequence per transaction — matching what V1's
  harness did by construction. (Not `0`, which livelocks the leader;
  `minTxnDuration` stays unset because the build clamps `max` up to `min`, and
  `--delay` is incompatible for the same reason.)
- **Timeout moved into the harness.** `--timeout=30s` is no longer passed.
  V1 read it as an idle bound; V2 reads it as wall-clock from process start and
  stops *gracefully*, so flowctl would exit 0 having silently dropped whatever it
  hadn't reached. The bound is now a context deadline in `RunWithContext`
  (default 2m, reachable only by a genuine hang) and overrunning it is an error.

Snapshot changes are limited to 15 lines across 12 files, all of them the
harness's own error prefix (`flowctl preview failed:` →
`flowctl raw preview-next failed:`), isolated in 25cb4d57a. No test expectations
change. Remaining commit updates prose in four READMEs that told contributors to
run `flowctl preview` directly.

New `TestLockstepTransactions` covers the catalog rewrite — nothing else asserts
it applies, and every snapshot in the repo depends on it.

**Notes for reviewers:**

Only the truncation coupling fails loudly. If `maxTxnDuration` doesn't take
effect, ordering and reduction regressions regenerate into green, meaningless
tests — a batched `TestReplicationUpdates` would silently drop the insert and
assert only the final state. So this is verified by diffing against committed
V1-era snapshots, not by refreshing them; `UPDATE_SNAPSHOTS` was never set.

- `source-postgres`: 3/3 full-suite passes with snapshots frozen, zero failures.
  All 108 transcript snapshots byte-identical.
- `source-sqlserver`, `source-sqlserver-ct`, `source-ga4-bigquery`: **unverified
  locally.** A failure showing added, removed or reordered fixture records means
  `maxTxnDuration` isn't taking effect there — stop and investigate, don't
  refresh.

Built against flowctl `v0.6.13-48-g964f4b438dc`, which includes `d61eeef3798`
(estuary/flow#3356) — that commit changed error text on both paths, so baselines
taken before it are stale.

Deciding which tests *should* exercise V2's batching is out of scope. Follow-on:
add a per-test opt-*out* back to the production close policy (with a positive
`--sessions` count invalid when opted out) and move tests off lockstep one at a
time.

**Checklist:**

- [ ] If there are user-visible changes, `CHANGELOG.md` has been updated — n/a, test-harness only
- [ ] If there are user-facing behavior changes, documentation has been updated — n/a, no connector behavior changes

### DIFF
--- a/go/capture/blackbox/capture.go
+++ b/go/capture/blackbox/capture.go
@@ -13,6 +13,7 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"time"
 
 	jsonpatch "github.com/evanphx/json-patch/v5"
 	"github.com/tidwall/gjson"
@@ -27,6 +28,10 @@ type Capture struct {
 	Checkpoint      json.RawMessage // Persistent checkpoint state between captures
 	DiscoveryFilter *regexp.Regexp  // Filter for discovered bindings (nil = no filtering)
 	Logger          func(...any)    // Log function (defaults to stderr, set to t.Log in tests)
+
+	// Timeout bounds a single capture run, and is enforced by the harness rather
+	// than by flowctl. Exceeding it is an error. Zero means the default.
+	Timeout time.Duration
 
 	// Env holds environment variables applied to each flowctl invocation, and so inherited
 	// by the connector process flowctl spawns. Settings which alter connector behavior have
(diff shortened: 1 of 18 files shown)


go/capture/blackbox/capture.go
```diff
@@ -170,14 +177,71 @@ func (c *Capture) writeCatalog() (catalogFile, catalogDir string, err error) {
...
-	if err := os.WriteFile(catalogFile, c.Catalog, 0644); err != nil {
+	if err := os.WriteFile(catalogFile, catalog, 0644); err != nil {
 		os.RemoveAll(catalogDir)
 		return "", "", fmt.Errorf("writing catalog: %w", err)
 	}
 	return catalogFile, catalogDir, nil
 }
 
+// withLockstepTransactions returns the catalog with every capture pinned to one
+// connector checkpoint sequence per runtime transaction.
```
jacobmarble:
```suggestion
// withLockstepTransactions returns the catalog with every capture pinned to one
// connector checkpoint sequence per runtime transaction by simply setting
// `maxTxnDuration` to 1ns.
```

go/capture/blackbox/capture.go
```diff
@@ -170,14 +177,71 @@ func (c *Capture) writeCatalog() (catalogFile, catalogDir string, err error) {
...
+// The V2 runtime's production close policy decides transaction boundaries by
+// whether the connector's next message happens to be buffered at the instant the
+// leader evaluates, so a capture which emits N checkpoints yields anywhere from 1
+// to N transactions. Two things follow, and both corrupt a snapshot: documents
+// batched into one transaction are emitted in collection-key order rather than
+// connector-emission order, and two changes to the same key within a transaction
+// reduce to one document. V1's harness avoided this by construction, which is why
+// existing snapshots record emission order and per-change documents.
+//
+// Collapsing the transaction window restores that. maxTxnDuration bounds how long
```
jacobmarble:
```suggestion
// Reducing the transaction window to 1ns restores that. maxTxnDuration bounds how long
```

go/capture/blackbox/capture.go
```diff
@@ -170,14 +177,71 @@ func (c *Capture) writeCatalog() (catalogFile, catalogDir string, err error) {
...
+// a transaction may extend, and the capture leader only evaluates the close policy
+// at checkpoint-sequence boundaries -- so any value below one leader iteration
+// admits exactly one sequence and then rotates. It cannot be zero: a closed
+// transaction has an open age of zero, which would make the extend arm
+// unreachable and livelock the leader.
+//
+// Correspondingly, minTxnDuration is left unset. The build clamps maxTxnDuration
+// up to minTxnDuration when the latter is larger, so setting both would silently
+// undo this. Passing --delay to preview-next does the same thing, and so is
+// incompatible with the harness.
```
jacobmarble: This strategy feels needlessly complicated. If we know that the 1ns value is just a trigger, then why not explicitly set both `maxTxnDuration` and `minTxnDuration` here?

The last bit "and so is incompatible with the harness" is also a big distraction. Does it mean that `withLockstepTransactions` doesn't have any effect when `--delay` is used?

jgraettinger:
Thanks. It now sets minTxnDuration as you suggest, and I generally took a hacksaw to new comment prose. PTAL

willdonnelly (approved):
LGTM!

PR #5136, opened 2026-08-27
source-mongodb: don't checkpoint a mid-split resume token

Alex-Bair:
**Description:**

### Problem

We observed a capture failing with `Error: Fragment("received fragment 2 without first fragment")`. That's indicating that the connector started up, used the checkpointed resume token, then received the second fragment of a change event. The only way I can see that happening is if the connector checkpointed a resume token in the middle of split change event.

Some background:

With pre-images enabled the connector appends `$changeStreamSplitLargeEvent`, so MongoDB delivers change events over 16MB as fragments. The transcoder reassembles them in process memory: fragments `1..N-1` return a `skip`, only fragment `N` returns a document.

`processBatch` checkpoints a post-batch resume token when a batch emitted no documents, with no regard for split state. That is the normal path for a fragment, not a rare race: a split event exceeds 16MB and a `getMore` reply is capped at 16MB, so fragments can never share a cursor batch; a batch of only fragments emits nothing, which is exactly the fallback's trigger; and the once-a-minute throttle means fragment 1's batch usually wins.

MongoDB resumes a fragment's token at the *next* fragment. So on the next restart, fragment 2 arrives into a fresh, empty accumulator and the transcoder exits. The restart reads the same token and fails identically, forever. **The capture cannot recover on its own**; rolling back to
runtime v1 changed nothing, because the bad token is in persisted connector state.

### Solution

Enforce one invariant: a resume token is only persisted for an event that is not a non-final fragment of a split event.

| Path | Token used | Status |
|---|---|---|
| `processBatch`, `emittedCount > 0` | last **emitted** document's token | Already safe — the transcoder only returns a document for a completed split or a non-split event. Unchanged. |
| `processBatch`, PBRT fallback | last event in the batch | Was unsafe. Now gated. |
| `handleIdleBatch` | post-batch resume token | Was unsafe when the cursor stopped mid-split. Now gated. |

1. **`isMidSplitEvent`** reads `splitEvent` off the raw BSON, which the pipeline already projects — no transcoder protocol change. It fails closed: suppressing a checkpoint costs a delayed resume token, checkpointing a fragment position is unrecoverable.
2. **Gate the two unsafe sites** on a `midSplit` flag, updated for every event consumed since it tracks where the cursor is, not what was captured.
3. **Don't finish catch-up mid-split.** Load-bearing, not a nicety: with (2) in place the last checkpoint is *before* the split, so re-opening the stream during backfill replays fragment 1 into an accumulator already holding it. (2) alone would trade one crash for another. Also adds a 5-minute warning, since suppressed checkpointing would otherwise let a stalled stream look healthy while its resume token froze.

**Checklist:**

### DIFF
--- a/source-mongodb/change_stream.go
+++ b/source-mongodb/change_stream.go
@@ -577,3 +577,26 @@ func extractTimestamp(token bson.Raw) (primitive.Timestamp, error) {
 
 	return primitive.Timestamp{T: timestampSeconds, I: increment}, nil
 }
+
+// isMidSplitEvent reports whether a change event is a non-final fragment of an
+// event split by $changeStreamSplitLargeEvent. A resume token for such an event
+// must never be checkpointed: resuming from a fragment's token delivers the
+// *next* fragment, and the fragments already consumed live only in the
+// transcoder's memory, which does not survive a restart.
+//
+// An event carrying a splitEvent which cannot be read is reported as mid-split.
+// The two outcomes are not symmetric: suppressing a checkpoint costs nothing
+// more than a delayed resume token, while checkpointing a fragment position
+// wedges the capture in a way it cannot recover from on its own.
+func isMidSplitEvent(raw bson.Raw) bool {
+	if _, err := raw.LookupErr("splitEvent"); err != nil {
+		return false
+	}
+
+	fragment, fragmentOk := raw.Lookup("splitEvent", "fragment").Int32OK()
+	of, ofOk := raw.Lookup("splitEvent", "of").Int32OK()
+	if !fragmentOk || !ofOk {
+		return true
+	}
+	return fragment < of
+}
--- a/source-mongodb/change_stream.go
+++ b/source-mongodb/change_stream.go
@@ -29,6 +29,10 @@ type changeStream struct {
 	ms                 *mongo.ChangeStream
 	db                 string
 	lastPbrtCheckpoint time.Time
+	// midSplit is true when the last event consumed from this stream was a
+	// non-final fragment of a split event, meaning the stream's current
+	// position is not a safe place to checkpoint.
+	midSplit bool
 }
 
 type streamEvent struct {
@@ -370,9 +374,10 @@ type processedEvent struct {
 }
 
 // handleIdleBatch processes an empty batch (idle signal), emitting a PBRT checkpoint
-// if enough time has passed. Returns the opTime extracted from the resume token.
+// if enough time has passed and the stream is not parked mid-split. Returns the
+// opTime extracted from the resume token.
 func (c *capture) handleIdleBatch(s *changeStream, token bson.Raw) (primitive.Timestamp, error) {
-	if time.Since(s.lastPbrtCheckpoint) > pbrtCheckpointInterval {
+	if !s.midSplit && time.Since(s.lastPbrtCheckpoint) > pbrtCheckpointInterval {
 		if err := c.emitPbrtCheckpoint(s, token); err != nil {
 			return primitive.Timestamp{}, err
 		}
@@ -421,6 +426,8 @@ func (c *capture) processBatch(
 
 	for i, resp := range responses {
 		ev := batch.events[i]
+
+		s.midSplit = isMidSplitEvent(ev.raw)
 
 		c.mu.Lock()
 		c.processedStreamEvents++
@@ -502,7 +509,7 @@ func (c *capture) processBatch(
 			c.lastEventClusterTime[s.db] = lastOpTime
 		}
 		c.mu.Unlock()
-	} else if time.Since(s.lastPbrtCheckpoint) > pbrtCheckpointInterval {
+	} else if !s.midSplit && time.Since(s.lastPbrtCheckpoint) > pbrtCheckpointInterval {
 		// No documents emitted, but we can still emit a PBRT checkpoint
 		lastEvent := batch.events[len(batch.events)-1]
 		if err := c.emitPbrtCheckpoint(s, lastEvent.resumeToken); err != nil {
--- a/source-mongodb/change_stream.go
+++ b/source-mongodb/change_stream.go
@@ -31,7 +31,9 @@ type changeStream struct {
 	lastPbrtCheckpoint time.Time
 	// midSplit is true when the last event consumed from this stream was a
 	// non-final fragment of a split event, meaning the stream's current
-	// position is not a safe place to checkpoint.
+	// position is not a safe place to checkpoint. Checkpoints are suppressed
+	// for as long as it is set. The fragments of one event are delivered back
+	// to back, so that is only ever briefly.
 	midSplit bool
 }
 
@@ -202,7 +204,7 @@ func (c *capture) streamChanges(
 					return fmt.Errorf("processing batch for %q: %w", s.db, err)
 				}
 
-				if coordinator.gotCaughtUp(s.db, opTime) && catchup {
+				if !s.midSplit && coordinator.gotCaughtUp(s.db, opTime) && catchup {
 					cancelProducer()
 					<-producerDone
 					return nil
--- a/source-mongodb/change_stream.go
+++ b/source-mongodb/change_stream.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
+	"sync"
 	"time"
 
 	m "github.com/estuary/connectors/go/materialize"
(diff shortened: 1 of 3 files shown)


jwhartley (dismissed):
Approved: The core fix looks right to me: never persist a resume token for a non-final fragment, and the two gated sites are the only paths that can write one. 

Two things I would raise:

**1. The transcoder accumulator is shared across database streams.** One transcoder per capture, but one stream goroutine per database, and the accumulator is a single process-global variable. `transcodeBatch`'s mutex makes each batch atomic, but the accumulator's lifetime spans batches, since fragments cannot share a cursor batch. So if database A is parked mid-split and database B sends a non-split event, the transcoder hits `received non-split event while accumulating fragments` and exits.

This is pre-existing rather than something you introduced, and after this PR the crash is non-poisoning, since the last checkpoint is pre-split and fragment 1 replays cleanly. So I do not think it blocks merging. But it does mean the fix does not fully hold for a multi-database capture with pre-images, and multi-database is possible: the endpoint `database` field is an optional comma-separated list, and omitting it discovers every database in the deployment. Worth a follow-up issue.

Minor, related: `TranscoderResponse.CompletedSplitEvent` is parsed off the wire and never read by any Go code. The transcoder already knows when a split completed. Was there a reason to re-derive split state from raw BSON rather than extend the existing signal? Re-deriving seems defensible, since the protocol has no "still accumulating" state, but the dead field is confusing next to the new predicate.

**2. An indefinite mid-split park hangs rather than fails.** Suppressing checkpoints is correct, but the only mitigation for a split that never completes is a repeating warning. Two consequences: the resume token stops advancing and moves toward oplog expiry, which ends in an unresumable token and a full backfill; and because `!s.midSplit` short-circuits the `gotCaughtUp` call, the coordinator never learns the stream reached the watermark, so backfill waiters stay blocked.

Failing fast would be strictly better here. A restart resets both the accumulator and the cursor, and the pre-split token replays cleanly, so the connector self-heals. Right now it chooses to hang quietly, which is harder to notice than a crash. Would you consider escalating to an error after some multiple of the warning interval?

Alex-Bair:
> The transcoder accumulator is shared across database streams

As you said, this is pre-existing and not currently an issue for the current escalation. I'm not going to expand the PR's scope to address this since this is currently an after hours change I'm not going to be around to monitor when it's deployed. Please create an issue to track this if you feel like this is something that needs addressed.

> TranscoderResponse.CompletedSplitEvent is parsed off the wire and never read by any Go code.

`CompletedSplitEvent` is also pre-existing & even though it looks harmless to remove it, I prefer to minimize how many changes are made in an after hours PR.

>An indefinite mid-split park hangs rather than fails.

This has been addressed. We'll fail after waiting 15 minutes now.

I also added recovery for captures that have persisted a mid-split resume token & are crash looping. Once they're running on these changes, they'll "rewind" the change stream to resume reading from the first fragment of the change event instead of the 2nd, 3rd, etc. fragment. That should let captures recover gracefully without missing data. That handling _should_ only be needed for the current captures stuck like this, and the handling shouldn't be needed once those captures recover since we'll prevent mid-split resume tokens from being checkpointed in the first place.

source-mongodb/change_stream.go
```diff
@@ -23,12 +26,58 @@ var (
...
+	// stream which resumed partway through a split event.
+	reopenAt func(ctx context.Context, at primitive.Timestamp) (*mongo.ChangeStream, error)
+	// accumulating is true when a first fragment has been handed to the
+	// transcoder and the split event it began has not yet been completed.
+	//
+	// This describes what the transcoder is holding, where midSplit describes
+	// where the cursor is. They differ for fragments which are dropped because
+	// their first fragment was never seen: those advance the cursor without
+	// giving the transcoder anything to accumulate.
+	accumulating bool
```
mdibaiee: Is this a real scenario? It is possible for us to "bypass" a document, but the cursor can advance?

source-mongodb/change_stream.go
```diff
@@ -23,12 +26,58 @@ var (
...
+
+	// midSplitStallLimit is how many reporting intervals a stream may stay
+	// parked mid-split before the capture fails. Hanging is worse than failing
+	// here: a restart resets both the transcoder's accumulator and the cursor,
+	// and the last checkpoint precedes the split event, so the event replays
+	// from its first fragment and the capture heals itself. Waiting instead
+	// leaves the resume token sitting still until it falls out of the oplog,
+	// and blocks any backfill waiting on this stream to report itself caught
+	// up.
+	midSplitStallLimit = 3
```
mdibaiee: Is this also a known issue where we can stall mid split?

Alex-Bair: Before the changes in this PR, no since we weren't considering whether we were mid-split before checkpointing. But this PR changed that & we're now intentionally not checkpointing mid-split & waiting for MongoDB to send the remaining fragments. The motivation around all the mid-split stall handling was more defensive rather than reactive to something we've observed. While I'd think MongoDB would send the remaining fragments pretty quickly and we wouldn't end up waiting very long for them, I was trying to guard against network partitions or misbehaving MongoDB servers that would cause us to not receive the remaining fragments in a timely manner.

source-mongodb/change_stream.go
Alex-Bair: It's possible if the following conditions are met:
1. The capture is restarted with a mid-split resume token, so it doesn't receive the first fragment of a change event on start up.
2. The oplog no longer reaches that change event, so we can't rewind the stream to replay it from its first fragment. MongoDB returns a `ChangeStreamHistoryLost` error and we skip it.

If we do skip ahead, we log a warning naming the collection we skipped & advance the cursor past the event.

I don't anticipate this "rewind the change stream and skip if the oplog no longer reaches" handling will be needed long-term since the other commits in this PR should prevent us from checkpointing a mid-split resume token in the first place. The handling is only to let the one existing capture that has already checkpointed a mid-split resume token recover without requiring backfills.

source-mongodb/change_stream.go
```diff
@@ -403,9 +616,22 @@ func (c *capture) processBatch(
 	var lastResumeToken bson.Raw
 	var emittedCount int
 
-	rawDocs := make([]bson.Raw, len(batch.events))
+	// A capture which resumed from a checkpoint taken partway through a split
+	// event receives that event's remaining fragments with no way to reassemble
+	// them: the earlier fragments only ever existed in the memory of a previous
+	// transcoder process. Drop them rather than handing them over, which would
+	// fail the transcoder and, because the checkpoint is never able to advance
+	// past them, every restart after it.
```
mdibaiee: Is this expected? I would imagine we want to loudly error and have the customer backfill rathr than just drop and continue, leading to data loss

source-mongodb/change_stream.go
mdibaiee: is there anything special about mid split events that makes us want to monitor them for a stall, but not other events? it just seems strange to me that we only monitor mid-split events for a stall, while they are just regular events

source-mongodb/change_stream.go
mdibaiee: I am surprised that we want to skip ahead in cases where we have not captured a fragmented event properly... that leads to data loss 🤔
isn't it better to crash and have the customer backfill / somehow find the oplog record that is appropriate and rewind to that?

Alex-Bair: Yeah, you're right. I'm updating to crash instead of skipping if we can't rewind successfully.

source-mongodb/change_stream.go
Alex-Bair: Mid-split events are the only ones where we intentionally suppress checkpoints, so I figured we'd want to bound how long we're intentionally stopping the capture from checkpointing. I can remove the stall monitoring completely if you'd like? Then we'd just wait indefinitely for MongoDB to send us the remaining fragments, which sounds like it'd probably be fine.

Alex-Bair: Yeah, I'm removing it. I'll have that removed in the next force push.

Alex-Bair: The mid-split stall monitoring has been removed.

source-mongodb/change_stream.go
Alex-Bair: It's been updated to crash instead of skipping now if we can't rewind successfully. So, there's not a scenario where we would skip a change event and advance the cursor any longer.

source-mongodb/change_stream.go
Alex-Bair: Same concern as the one addressed in this [thread](https://github.com/estuary/connectors/pull/5136#discussion_r3872413226). We no longer skip & instead error out.

Alex-Bair:
Thanks for the comments, @mdibaiee! I believe I've addressed them, but please let me know if I missed anything. Could you please take another look when you get a chance? 🙇

nicolaslazo (approved):

mdibaiee (approved):

PR #5139, opened 2026-08-27
source-shopify-native: add `name` to orders line items

SeanWhelan:
**Description:**

Adds `LineItem.name` to the `orders` bulk query, so each line item carries its title as it appears on the order. A customer has orders whose line items were entered without a `sku`, where `name` is the only field identifying what sold.

**Workflow steps:**

Query-only change in `graphql/orders/orders.py`. `name` is a scalar, so it inlines into the line item's JSONL row and `process_result` needed no changes. No access scope beyond the `read_orders` the stream already requires. Verified with the capture snapshot test against the test store.

Existing captures won't backfill - `name` appears on orders whose `updatedAt` advances after deploy.

**Checklist:**

### DIFF
--- a/source-shopify-native/CHANGELOG.md
+++ b/source-shopify-native/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 2026-08-27
+
+### Added
+- `orders` line items now include `name`, the line item's title as it appears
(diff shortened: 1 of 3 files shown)


Alex-Bair (approved):
LGTM, thanks!

PR #5142, opened 2026-08-27
tests: migrate integration tests off the flowctl-go temp-data-plane

jgraettinger:
**Description:**

Migrates `tests/run.sh` off the `flowctl-go` temp-data-plane (V1 runtime) onto `flowctl raw preview-next` (V2), ahead of the V1 harness being removed.

The old harness activated a catalog, materialized into a SQLite sidecar, and polled it until a projected `id`/`canary` table matched a checked-in `expected.txt`. All of that is gone. `run.sh` now runs the capture directly under `flowctl raw preview-next --sessions 1 --delay 30s` — one transaction batching everything produced inside a 30s window — and diffs its stdout, full documents with `_meta`, against a checked-in `tests/<connector>/snapshot.jsonl`. The assertion widens from two projected columns to the whole document.

Two tiers:

- **Full capture** (`source-gcs`, `source-kinesis`, `source-s3`, `source-sftp`) — these have no other end-to-end coverage. They run the smoke test below first, so a broken image fails in a second rather than after a capture.
- **Image smoke test** (`source-postgres`, `source-mysql`, `source-sqlserver`, `source-dynamodb`, `source-kafka`) — thorough suites of their own, duplicated by `tests/` at much higher cost. Replaced with `tests/smoke.sh`, which fetches the spec from the built image: the one thing those suites can't check, since they never touch it.

`source-kafka` lands as its own commit. Its Rust suite covers seven topic encodings plus capture resumption, but had **never run in CI**, because the Dockerfile only does `COPY source-kafka/src ./src`. This wires it in and migrates it to `raw preview-next`. Its catalog names the binary cargo already built rather than `cargo run`, whose nested cargo contends with the outer `cargo test` for the build lock. Waits in the suite are bounded, as are the job and step via `timeout-minutes`.

Also drops `testsupport/build_catalog.go` (zero callers) and `template.flow.yaml`'s `enable-runtime-v2: "false"` pin, which existed only to keep captures working under the temp data plane this PR deletes. `go/capture/blackbox` is being migrated separately.

A last commit aligns the Rust connectors' Dockerfiles on the release profile: the dummy-src layer primed dev while the build asked for release, compiling every dependency twice. `source-http-ingest`'s image build drops from 682 `Compiling` lines to 313, over 308 crates. **It also changes `materialize-kafka` and `source-http-ingest`**, whose Dockerfiles are identical to source-kafka's.

**Workflow steps:**

Unchanged; `materialize-sqlite` and `flowctl-go` are no longer needed.

```bash
UPDATE_SNAPSHOTS=true CONNECTOR=source-s3 VERSION=local ./tests/run.sh
```

**Notes for reviewers:**

Two pre-existing bugs are fixed in their own commits: `tests/source-sftp/setup.sh` ran its readiness check from the wrong directory and the failure was swallowed by `source … || bail`, so that test has been racing container startup; and two `tests/source-s3` config templates are dead.

- **Resource names are de-randomized** for `source-s3` and `source-kinesis`, partition keys included so Kinesis' shard assignment repeats. `source-gcs` can't be — globally-unique bucket in a shared project — so it keeps a random suffix and `run.sh` rewrites `/_meta/file` before comparing.
- **Always-true healthchecks are replaced.** Seven compose files declared `healthcheck: test: "true"`, a fixed sleep unrelated to readiness. Real probes now, with `up --wait` bounded by `--wait-timeout`.

**Checklist:**

- [x] If there are user-visible changes, `CHANGELOG.md` has been updated — n/a, test-only, no connector behavior change
- [x] If there are user-facing behavior changes, documentation has been updated — n/a; `tests/README.md` and the affected connector READMEs are updated


source-sqlserver/docker-compose.yaml
```diff
@@ -16,8 +16,19 @@ services:
       ACCEPT_EULA: "Y"
       MSSQL_AGENT_ENABLED: "true"
     healthcheck:
-      test: "true"
-      interval: 30s
+      # The entrypoint starts sqlservr in the background and then runs the
```
mdibaiee, line 19:
```suggestion
      # The entrypoint starts sqlserver in the background and then runs the
```

mdibaiee (approved):
This looks good to me but is a capture team concern so might want to wait for them

Alex-Bair (approved):
LGTM, thank you for this!

source-sqlserver/docker-compose.yaml
jgraettinger: `sqlservr` is the literal name of the executable

PR #5145, opened 2026-08-28
source-kafka: per-binding backfill mode to start at latest offset

### LINKED ISSUE #5143: source-kafka: per-binding backfill mode to start at latest offset
## Problem Statement

When you create a `source-kafka` capture, it always reads every message still retained in the topic. There is no way to say "start from now, ignore the history."

For a topic with meaningful retention this is a barrier to adoption. A customer bringing a production Azure Event Hub into Estuary has 100GB+ of retained product analytics data. They only want events from the point the capture is published forward. Today their options are to accept a 100GB+ read they do not want (with the storage and processing cost and the time that implies), or to destroy retained data by shortening the event hub's retention before publishing, or to stand up a brand new event hub with no history and re-point their producers at it.

This is a gap specific to us, not to Kafka. The `auto.offset.reset` consumer property is the standard Kafka answer, and Azure Event Hubs honors it over the Kafka protocol. Users reasonably arrive expecting it, look for it in the connector's configuration, and do not find it.

The workarounds are all bad in a way that is worth stating plainly:

- Setting `auto.offset.reset` in any client config has no effect, because the connector assigns partitions explicitly rather than using consumer-group subscription, and librdkafka only consults that property for unset or invalid offsets under a subscription.
- Pre-seeding the broker-side consumer group offset has no effect, because the connector never commits offsets to the broker. It tracks them in Flow state.
- Incrementing the binding's `backfill` counter makes things strictly worse. A new backfill counter produces a new state key, so the connector finds no saved state and restarts from the earliest offset.
- `notBefore` cannot help. It is not expressible on a capture binding at all, and only filters what downstream consumers read. The data would still be captured, stored, and billed.

Every apparent escape hatch is closed, so the connector has to grow the option.

## Solution

Add a per-binding backfill mode to `source-kafka` that lets you choose where a binding starts when it has no saved state.

The default is unchanged: start at the earliest retained offset and read the topic's history. Selecting "Only Changes" makes the binding start at the end of each partition instead, so the capture picks up only messages produced after it starts.

Users already know this control. The SQL CDC connectors (Postgres, MySQL, SQL Server) expose a per-binding `mode` field whose values include "Only Changes", with exactly this meaning: skip the existing contents, take changes from now. `source-kafka` reuses that field name and that value so the concept transfers, the UI renders the familiar dropdown, and the docs can describe one idea rather than two.

The setting only applies to a binding with no saved state. A binding that has already run keeps resuming from its stored offsets, and the mode is ignored. This makes the option safe to add to a running capture and impossible to use as an accidental data-skipping switch on an established binding.

## User Stories

1. As a data engineer onboarding a high-retention Kafka topic, I want to capture only new messages, so that I do not pay to move terabytes of history I have no use for.
2. As a data engineer with a production Azure Event Hub holding 100GB+ of retained data, I want to publish a capture without triggering a full read of that retention, so that I can go live without a large one-off cost and delay.
3. As a data engineer, I want to choose the start position when I create the binding, so that I do not have to modify my source system to control what Estuary reads.
4. As a data engineer, I do not want to shorten my event hub's retention to control where the capture starts, so that I do not permanently destroy data other consumers depend on.
5. As a data engineer, I do not want to create a parallel topic with no history and re-point my producers, so that I avoid a production change unrelated to the problem I am solving.
6. As a Kafka user, I want the option named in terms I already know from other Estuary captures, so that I do not have to learn a connector-specific concept.
7. As an existing `source-kafka` user, I want my captures to behave exactly as before after this change ships, so that I am not surprised by a silent change in what gets captured.
8. As a user who sets the mode and then restarts the task, I want the binding to resume from where it left off rather than jumping to the end again, so that I do not lose messages across a restart.
9. As a user who sets the mode on a binding that has already been running, I want the setting to be ignored, so that I cannot accidentally skip unprocessed data.
10. As a user who wants to re-read a topic's history later, I want to be able to switch the mode back and trigger a backfill, so that the choice is not permanent.
11. As a user adding a new topic to an existing capture, I want the new binding to honor the mode I chose, so that adding a topic does not pull in that topic's full history unexpectedly.
12. As a user reading the connector's documentation, I want the start-position behavior described, so that I can decide before I publish rather than after.
13. As a user of the Estuary UI, I want the mode to appear as a labelled dropdown on the binding, so that I can find it without reading the connector source.
14. As a user, I want the default clearly indicated in the UI, so that I understand what happens if I leave it alone.
15. As a user who selects "Only Changes" on a topic that is currently idle, I want the capture to sit healthy and wait for new messages, so that I do not mistake an empty capture for a broken one.
16. As a user with a multi-partition topic, I want every partition to start at its own end offset, so that I do not get a partial history from some partitions and none from others.
17. As a user whose topic gains partitions after the capture starts, I want the new partitions handled predictably, so that I understand whether their existing content is read.
18. As a support engineer, I want the connector to log the chosen start position per partition at startup, so that I can confirm from logs what a capture actually did.
19. As a support engineer, I want an invalid mode value rejected at validation time with a clear message, so that a typo fails at publish rather than producing a silent full backfill.
20. As a support engineer, I want to explain this feature by pointing at the same concept used for the SQL captures, so that the answer is consistent across connectors.
21. As a solutions engineer sizing a customer's onboarding, I want to tell a customer with a large retained topic that they can start from now, so that retention size stops being a blocker to adoption.
22. As a developer of this connector, I want the new behavior covered by tests that run against a real broker, so that a regression is caught before release.
23. As a developer, I want the spec snapshot to record the new configuration property, so that unintended schema changes are visible in review.

## Implementation Decisions

**Naming, borrowed from the SQL captures.** The field is `mode` on the binding's resource configuration, and the non-default value is the string `Only Changes`. This matches the `BackfillMode` field the SQL CDC connectors already expose, where the same string carries the same meaning. Do not invent `auto.offset.reset`, `startingOffset`, or `skipBackfill` for this. The empty string is the default and means "read from the beginning", mirroring the SQL captures' `BackfillModeAutomatic`.

**Only two values.** The SQL captures expose five modes because table backfills have genuine variants around chunking, key handling, and watermarks. A Kafka partition has exactly two meaningful start positions. Expose only the empty default and `Only Changes`. Reject any other value during validation with an error naming the accepted values, rather than accepting it and silently falling through to a full read.

**Per-binding, not per-endpoint.** The setting lives on the resource configuration, alongside `topic`, because that is where every other capture puts it and because a user may reasonably want history for one topic and not another. This also means the binding-level UI already has a place to render it.

**The behavior change is confined to the start-position decision.** The connector builds a topic-partition assignment list at startup, choosing per partition between a saved offset and a default. Only the default branch changes: instead of always resolving to the beginning of the partition, it resolves to either the beginning or the end according to the binding's mode. The saved-offset branch is untouched, which is what makes the setting inert for a binding that has already run.

**Use the Kafka client's native end-of-partition offset.** The rdkafka `Offset` enum already provides both `Beginning` and `End`, and the existing explicit partition assignment accepts either. No new dependency, no broker round trip, and no change to how offsets are stored or checkpointed.

**No state format change.** Saved state remains a map of binding state key to per-partition offsets. The mode is configuration, never persisted state, so there is no migration and no new state version.

**Discovery sets the default explicitly.** The discover path constructs the resource configuration for each topic. It should emit the default mode so that discovered bindings are explicit about their start position rather than relying on an absent field.

**Schema authoring.** The endpoint configuration schema for this connector is hand-written rather than derived, but the resource configuration is derived from the type. Follow the repo's JSON schema guidelines for the new property: a Title Case `title`, a `description` written in second person and ending with a period, an explicit `type` alongside the `enum`, and a stated default. The property is not sensitive and is not an object, so it needs neither the `secret` nor the `advanced` annotation.

**Logging.** Log the resolved start position per topic-partition at assignment time. This is the only way to confirm after the fact what a capture did on its first run, and support has no other artifact to check.

**Interaction with the backfill counter is documented, not engineered.** Incrementing a binding's backfill counter produces a new state key, so the binding is treated as fresh and honors the mode again. This is the correct behavior and gives users a supported way to re-apply the setting, but it is surprising enough to belong in the field's description and the connector docs.

## Testing Decisions

A good test here asserts what the connector emits, not how it decided. The observable behavior is the set of documents produced and the resulting connector state. Tests should drive the connector the way the runtime does and snapshot its output, never reach into the offset-selection logic directly.

**Prior art.** The existing tests in this connector's integration suite establish the pattern: a shared setup that creates topics and produces a fixed number of messages across a fixed number of partitions against the docker-compose Kafka and schema registry, then a test that runs the connector through `flowctl preview` and snapshots stdout with `insta`. There is already a resume test that passes a hand-written initial state via `--initial-state` and asserts the connector picks up from those offsets. The new tests extend this suite; no new test module or harness is introduced beyond the concurrency change described below.

**Harness change: driving the capture concurrently.** The existing capture tests invoke `flowctl preview` with `.output()`, which runs to completion and collects stdout. That shape cannot express this feature's central assertion, because a capture starting at the end of the partition has nothing to read from the pre-produced fixture and would produce an empty snapshot whether or not the feature works. The test must instead start the preview subprocess, produce messages after it is running, and assert that only those messages arrive. This requires spawning the subprocess with piped stdout and reading from it while the test produces, rather than blocking on completion. Keep this change additive: the existing tests continue to use the simpler blocking form, and the concurrent driver is a helper used by the new tests.

Tests to add:

1. **Only Changes skips history and captures new messages.** With the standard fixture already produced, start a capture whose bindings are set to `Only Changes`. Once it is running, produce a distinct, identifiable set of new messages. Assert that the captured documents are exactly the new messages, that none of the fixture messages appear, and that the final state reflects offsets past the fixture. This is the test that actually proves the feature.
2. **Saved state takes precedence over the mode.** Run a capture with `Only Changes` set on the bindings but with an initial state supplying partition offsets in the middle of the fixture, in the manner of the existing resume test. Assert the connector resumes from the supplied offsets and reads the remaining fixture messages, demonstrating the mode is ignored once state exists. This can use the existing blocking harness.
3. **Default behavior is unchanged.** The existing capture test already covers this and must continue to pass untouched. Treat any change to its snapshot as a regression rather than an expected update.
4. **Spec snapshot records the new property.** The existing spec test snapshots the full connector spec. Its snapshot will change to include the new resource property; review that diff deliberately, confirming the title, description, type, enum values, and default all match the guidelines.
5. **Invalid mode is rejected.** Assert that a binding configured with an unrecognized mode value fails validation with a message naming the accepted values, rather than silently defaulting.

Determinism matters in this suite. The existing resume test filters intermediate connector-state lines because messages across topics may be polled in any order while the final state is consistent. Apply the same treatment to the new tests, and make the post-start messages distinguishable from the fixture so ordering across partitions does not make the assertion flaky.

## Out of Scope

**Starting from a specific timestamp.** This is the natural next request and deserves its own spec. It is a materially bigger change: it needs a broker round trip to resolve one timestamp into per-partition offsets, a timeout and failure path for that lookup, and a defined fallback for partitions whose timestamp is past the end (where the correct answer is the end, not the beginning). It also depends on the broker implementing the Kafka offset-for-timestamp lookup, which has not been verified against Azure Event Hubs. None of that is required to unblock the case that prompted this.

**Starting from explicit per-partition offsets.** Technically about the same amount of code as this spec, because the assignment already accepts an explicit offset per partition, but a much worse interface. It obliges the user to hand-maintain a partition-to-offset map that is wrong as soon as partitions are added, has no sensible default, and fails silently when a number is stale or off by one. Do not build it unless a concrete request arrives that timestamp cannot serve.

**Changing how offsets are stored or committed.** The connector will continue to track offsets in Flow state and will continue not to commit them to the broker. Broker-side consumer group offsets remain irrelevant to this connector's behavior.

**Retroactively changing existing captures.** No migration, no backfill of the new field onto published specs, and no change in behavior for any binding with saved state.

**Azure Managed Identity and federated credentials for `source-kafka`.** Raised by the same customer in the same conversation and genuinely wanted, but a separate piece of work. The connector's credentials currently cover username/password and AWS MSK IAM, with no Entra path. It belongs with the existing cross-connector IAM authentication effort, not here.

**Other Kafka-protocol connectors.** This spec covers `source-kafka` only.

## Further Notes

The reason this change is small is worth recording, because it will not be obvious to whoever picks it up. The connector already stores per-partition offsets itself and re-derives its partition assignment on every startup. A start position is therefore a free parameter at assignment time, and all three variants (earliest, latest, explicit) reuse a single code path. A connector that relied on consumer-group subscription and broker-side offsets would have to negotiate this with the broker instead, and would be a much larger change.

Worth stating in the customer-facing documentation, because users will try these first and be confused when they do nothing: setting `auto.offset.reset` has no effect on this connector, and neither does setting the consumer group's offset on the broker.

The `group.id` the connector sets exists to satisfy the client library. Automatic commit is disabled and partitions are assigned explicitly, so the consumer group is not used for coordination or offset storage. This is a common source of confusion when debugging with Kafka admin tooling, which will show the group with no committed offsets.

jwhartley:
Closes #5143.

## What changed

A `mode` field on the resource config chooses where a binding starts. Left empty it behaves exactly as before, starting at the earliest retained message. Set to `Only Changes` each partition starts at its end offset, so the capture picks up only messages produced after it starts. The field name and value match the SQL CDC captures.

Under `Only Changes` the connector resolves each partition's end offset once with `fetch_watermarks`, assigns it as an explicit number, and checkpoints it before polling. The saved-offset branch is untouched, which is what makes the setting inert for a binding that has already run.

## Why

Onboarding a high-retention topic meant reading history the user had no use for, and every apparent escape hatch is closed: `auto.offset.reset` is only consulted under a consumer-group subscription and this connector assigns partitions explicitly, broker-side group offsets are irrelevant because offsets live in task state, and incrementing the backfill counter makes it worse by producing a new state key that restarts from the earliest offset. #5143 has the full reasoning.

## Why the round trip, and a spec correction

The first version used `Offset::End` and no broker round trip, per the spec's Implementation Decisions. Review found that this loses a message.

`Offset::End` is resolved by the broker at fetch time, on every startup, and the connector checkpointed only after receiving a message. So a binding that stopped before its first message had saved nothing, and the next run resolved the end of the partition again, past anything that arrived in between. No error, no gap in state. An idle topic is exactly where this bites, and that is the case story 15 asks to work.

The same constraint made the spec's Logging requirement unsatisfiable, since a resolved position is a number and getting one needs the round trip. Resolving the offset once fixes both. The spec is amended in [this comment](https://github.com/estuary/connectors/issues/5143#issuecomment-5452028295); happy to argue the trade if you would rather keep the lazy offset.

One `fetch_watermarks` call per affected partition, bounded at 30s, and only for partitions that both lack a saved offset and are set to `Only Changes`. The default path is unchanged.

## How it was verified

Two clean back-to-back `cargo test` runs, both green and nothing skipped: 23 unit tests and 8 integration tests. The suite now runs on the V2 runtime (`flowctl raw preview-next`) after the migration on main, and this is the first branch where CI runs it.

New integration tests cover: skipping retained history, an idle capture that checkpoints a position and resumes from it without losing the message produced in the gap, the empty-partition boundary where the stored offset is -1, and a saved state overriding the mode. Plus unit tests for accepting, defaulting and rejecting a mode value.

The V2 migration needed two harness changes: `catalog()` generalized to `catalog_for(path)` so the only-changes fixtures also get `cargo run` rewritten to the built binary, and `spawn_bounded`, a non-blocking `bounded_command` for the test that produces messages while the capture runs.


jwhartley:
One thing I considered and rejected, recorded so it does not come up fresh in review.

`deserialize_mode` is hand-written where a derived `Deserialize` plus a separate validation step would also work. `materialize-kafka` takes that second approach: `do_validate` deserializes and then calls `config.validate()` (materialize-kafka/src/validate.rs:18).

I kept the custom deserializer because `source-kafka` already validates this way: `validate_json_only_true` (source-kafka/src/configuration.rs:82) rejects a bad value with a custom serde error, same file, same reason. `source-kafka` has no `validate()` method, so matching `materialize-kafka` would mean adding one, and would make this connector internally inconsistent to make it externally consistent.

Either approach fails a typo at publish, since deserialization is what `do_validate` runs. The difference is only the message: serde's derived `unknown variant` renders the empty default as bare backticks and never names the accepted values.

Open question, not a blocker: the two Kafka connectors validate differently. Worth converging, or is per-connector consistency the right granularity?

jwhartley:
Rebased onto main to pick up the `flowctl raw preview-next` migration, which also turned this Rust suite on in CI for the first time. The conflict was confined to `tests/test.rs`.

docs/reference/Connectors/capture-connectors/apache-kafka.md
```diff
@@ -143,6 +143,34 @@ See [connectors](../../../concepts/connectors.md#using-connectors) to learn more
...
+The mode applies only to a partition with no saved offset, so:
+
+- A binding that has already run resumes from its stored offsets and ignores the
+  mode. You can add the setting to a running capture without skipping
+  unprocessed data.
+- To apply the mode to a binding that has already run, increment that binding's
+  backfill counter. This gives the binding a new state key, so the connector
+  treats it as new.
+- A partition added after the capture starts follows the mode, so under
+  `Only Changes` it starts at its end offset rather than reading what it holds.
```
Alex-Bair, line 164: Is this the behavior users will expect? My understanding is that in our platform, an "Only Changes" backfill mode means "capture everything from this point in time onward", and that includes partitions added after the capture starts up for the first time. Meaning, if a capture starts up & a binding already has pre-existing state, the capture should read any new partitions from their beginnings rather than their ends.

docs/reference/Connectors/capture-connectors/apache-kafka.md
```diff
@@ -143,6 +143,34 @@ See [connectors](../../../concepts/connectors.md#using-connectors) to learn more
...
+
+Two Kafka settings that look like they should control this do not affect this
+connector. `auto.offset.reset` is only consulted for an unset or invalid offset
+under a consumer-group subscription, and this connector assigns partitions
+explicitly. Setting the consumer group's offset on the broker also does nothing,
+because the connector tracks offsets in Estuary's task state and never commits
+them to the broker. The `group.id` it sets exists only to satisfy the Kafka
+client library, which is why Kafka admin tooling shows that group with no
+committed offsets.
```
Alex-Bair, line 173: I think this paragraph should be removed from the user facing docs. It describes how the connector interfaces with Kafka, not how the user interfaces with the connector.

source-kafka/src/configuration.rs
```diff
@@ -474,14 +474,91 @@ impl EndpointConfig {
...
+        ))),
+    }
+}
+
 #[derive(Serialize, Deserialize, JsonSchema)]
 pub struct Resource {
     #[schemars(title = "Topic", description = "Kafka topic to capture messages from.")]
     pub topic: String,
+
+    #[serde(default, deserialize_with = "deserialize_mode")]
```
Alex-Bair, line 513: nit: Instead of using an empty string `""` to signify "No backfill mode is specified, so use `Automatic`", can we omit the field when it's the default, like the SQL captures' `mode,omitempty`? I'm not super familiar with how to do that in Rust. Maybe this would work?
```suggestion
    #[serde(
        default,
        deserialize_with = "deserialize_mode",
        skip_serializing_if = "BackfillMode::is_automatic"
    )]
```
then you'd have to add an `is_automatic` implementation for `BackfillMode`:
```rust
impl BackfillMode {
    /// Whether serialization skips the field, so a discovered binding omits the
    /// default the way the SQL captures' `mode,omitempty` does.
    fn is_automatic(&self) -> bool {
        *self == BackfillMode::Automatic
    }
}
```

Alex-Bair (commented):
Had a few questions & comments. The main blocking one is the behavior for a binding with pre-existing state that's configured with an "Only Changes" backfill mode. I think that any new partitions that appear after a binding has already run once should be captured from their beginnings, not their ends. That'd be more in line with how "Only Changes" backfill modes work for other captures. It'd be nice to have a test covering that situation too so the behavior is clearly demonstrated & covered by a test.

jwhartley:
Less work for eng to complete, closing.

PR #5148, opened 2026-08-28
Finish the flowctl preview -> raw preview-next migration

jgraettinger:
Completes the move off `flowctl preview` (V1 runtime) onto `flowctl raw preview-next` (V2), picking up where the capture-side migration left off. No `flowctl preview` invocations remain in the tree.

This is groundwork: `flowctl preview` is being reimplemented on top of `preview-next`, so the two become the same command, both backed by V2. Anything still on `preview` would have quietly turned into a V2 run wearing a V1 label, with its snapshot delta surfacing later as a mystery CI failure instead of being reviewed here.

Four commits, best read in order:

1. **`.actual` sidecars** — on a snapshot mismatch the harness writes the run's value beside the snapshot, and CI uploads it. Cloud-credentialed snapshots can only be regenerated from CI, and the job log can't do it: secret redaction mangles the connector state they embed.
2. **Checkpoint-token sanitizer** — V2 advances the checkpoint per transaction, over a randomized per-run task name, so connectors that hash it (`materialize-s3-iceberg`, `materialize-snowflake`) produce snapshot values unique to the run.
3. **The migration itself** — every materialization onto V2, the harness collapsed onto its single remaining path, snapshots regenerated.
4. **The benchmark harness** — drops its V1 comparison arm, which was about to become a second V2 arm.

No destination table data or schema moves in any snapshot. Every regenerated line is one of flowctl's `--output-apply` / `--output-state` lines or the runtime's own document metadata; commit 3's message enumerates the four deltas.

Verified locally against real Docker infrastructure for every connector whose endpoint is reachable. The cloud-credentialed snapshots are the bytes CI's `.actual` sidecars handed back.

Unrelated and pre-existing on main: `materialize-azure-fabric-warehouse`'s endpoint returns "Couldn't complete the operation due to a system update", failing every subtest including ones this PR doesn't touch.


jacobmarble (commented):
LGTM

jacobmarble (approved):

PR #5153, opened 2026-08-28
materialize-snowflake: snowpipe streaming high-performance

### LINKED ISSUE #4324: Snowflake Snowpipe Streaming: Switch to SDK
## Problem Statement

Snowflake is retiring the Snowpipe Streaming *classic* architecture that
`materialize-snowflake` reimplements in Go today. Snowflake plans a formal
deprecation announcement in mid-2026 followed by an 18-month sunset, putting
end-of-life around the end of 2027. Users materializing to Snowflake with
delta-updates bindings need the replacement high-performance architecture —
server-side pipes, throughput-based billing, automatic compute scaling — and the
only supported client is Snowflake's Python SDK over a closed-source Rust core,
whose behavior we do not trust ourselves to reproduce against the documented
REST API.

Two constraints shape the design beyond the SDK itself.

First, the v1 runtime gives no idempotent-replay guarantee, so any v1-compatible
design must hold a durable, container-independent copy of a transaction's rows
in order to re-apply them after a crash — which means transmitting every row to
Snowflake twice. Users would pay double egress and double Snowflake-side storage
churn indefinitely, for a compatibility property that is itself being retired.

Second, Snowflake meters ingest throughput per channel, and a channel is bound
to the process that opened it. A naive one-channel-per-shard design caps each
shard at the single-channel ceiling, and — worse — ties channel identity to
shard topology, so that a shard split or join lands mid-flight rows in channels
whose committed offset tokens no successor can safely interpret.

## Solution

Restrict the new write path to tasks running **runtime v2**, and append rows
directly to Snowflake channels during `Store`. Runtime v2 replays an
interrupted transaction identically, so the connector needs no durable copy of
its own: recovery is a comparison between each channel's committed offset token
and a document counter the connector keeps in its driver checkpoint, skipping
rows Snowflake already holds. There is no spill, no stage, and each row crosses
the wire once.

Each shard subdivides its key range into **four channels per binding**, one per
equal subrange, and routes each document to a channel by the same packed-key
hash the runtime routes documents to shards by. A channel's contents therefore
depend only on the data, never on the shard count: an aligned split or join
hands successor shards *whole channels*, tokens and all, and each successor
converges to its own four-channel layout at a safe transaction boundary. Four
channels also give each shard four times the per-channel throughput ceiling.

Because a channel is bound to its table, the table itself becomes part of the
recovery story: a backfill re-creates the table rather than emptying it, and
each table carries a **generation marker** in its comment naming the
materialization and state key it belongs to, so a replaced generation's shards
are refused rather than allowed to append into a table that is no longer
theirs.

The path is opt-in behind the `snowpipe_streaming_v2` feature flag, requires
key-pair (JWT) authentication, serves delta-updates bindings only, and refuses
to run — loudly, at apply time and at startup — on a task that is not on
runtime v2.

### Runtime v2 detection

- `materialize-boilerplate` reports whether a materialization spec runs under
  runtime v2 by testing the `enable-runtime-v2` flag label on the spec's shard
  template (`IsMaterializationSpecRuntimeV2`). It accepts the spec carried by
  either `Open` or `Apply`.
- This is sound because the reactor selects its own v1/v2 code path from that
  same label, and the build folds a task's `shards.flags` into the shard
  template's label set. The connector inspects the same source of truth the
  runtime uses.
- A nil spec or an absent/`"false"` label reads as **not** v2 — fail-safe by
  construction: `flowctl preview` and local runs read as v1.
- Deliberately not used: a new protocol field (requires a flow change we are
  not making), a runtime-supplied environment variable (does not exist), or
  inference from `state_patches_json` (legitimately empty exactly when its
  absence would be read as v1, losing data in that direction).

### Eligibility and gating

- A binding uses the new path when **all** of these conditions are met:
  - the task is on runtime v2
  - the `snowpipe_streaming_v2` feature flag is set
  - the binding is delta-updates
  - the endpoint authenticates with a key pair (JWT)
- Setting both `snowpipe_streaming` and `snowpipe_streaming_v2` is a
  configuration error rejected during config validation, as is setting the v2
  flag without JWT credentials. Silently choosing between two write paths is
  worse than making the operator choose.
- The connector ships as a **gated driver** wrapping the SQL driver.
  - `Validate` **warns** when the endpoint asks for the v2 path and the last
    published spec's shard template lacks the runtime flag. It cannot refuse:
    the Validate request carries no shard template of the task being
    published, so a publish that adds both flags together is indistinguishable
    from one that forgets the runtime flag.
  - `Apply` **refuses** when the spec in hand asks for the path without the
    runtime flag — publish-time feedback, before anything in Snowflake is
    created or altered.
  - `Open` (transactor construction) refuses again, as the backstop for a
    runtime flag turned off later without touching the endpoint configuration.
- The gate lives in the driver wrapper rather than a Materializer method
  because the runtime is a property of the task, and Apply and Open are the
  only RPCs that carry the task's spec.

### The sidecar

- Snowflake ships no Go SDK for the high-performance architecture, so the
  connector supervises a small Python sidecar package that wraps the
  `snowpipe-streaming` SDK. Connector and sidecar speak NDJSON request/response
  RPC over a unix socket: `configure`, `open_channel`, `append` (rows framed as
  a raw pre-encoded JSON array payload), `wait_commit`, `channel_status`,
  and `shutdown`, each with an operation-appropriate timeout (`wait_commit`
  much longer than the rest, since it is bounded by Snowflake's commit
  latency).
- Append payloads pass through the sidecar **opaque**: the Go side encodes each
  row directly into the batch's JSON array, and the sidecar hands the bytes to
  the SDK without re-parsing them (`msgspec.Raw`). A guard test verifies the
  installed SDK actually accepts raw rows, and the Docker build runs it with
  `SNOWPIPE_SDK_GUARD=required` so an SDK the test cannot import is a build
  failure rather than a silent skip.
- The sidecar is **crash-only**. The connector never restarts it; any sidecar
  error is fatal to the transactor session, the runtime restarts the
  connector, and recovery replays through the offset tokens. The supervisor
  handles the ready-line handshake with a timeout, escalates stop through
  SIGTERM to SIGKILL on the process group, keeps a bounded ring buffer of
  recent stderr for error reports, and relays the sidecar's structured ops-log
  lines into the connector's own logging.
- The Docker image builds the sidecar's virtualenv in a dedicated stage on the
  same base image as the runtime stage, so interpreter and site-packages match
  exactly; a throwaway venv holds pytest so test dependencies never ship.

### Row encoding

- The connector writes each converted document straight into the batch buffer
  as a JSON object keyed by column name, with per-binding precomputed
  quoted-name prefixes, instead of building a `map[string]any` and calling
  `encoding/json`. Nil columns are omitted. Floats that JSON cannot represent
  (NaN, ±Inf) fail the row with the column named. Pre-encoded JSON values are
  scanned for line breaks — which would break NDJSON framing — and compacted
  only when one is present, so the common case is a straight copy.
- The direct writer is held byte-equivalent (up to JSON number/whitespace
  identity) to the map-and-marshal reference by test, and benchmarked against
  it; the batch then needs no second per-row scan to frame.

### Channels, routing, and naming

- `streamV2ChannelsPerShard = 4`. Each shard cuts its inclusive key range
  `[keyBegin, keyEnd]` into four equal subranges — its **target layout** — and
  opens one channel per subrange per binding. A shard range that does not
  divide evenly is refused (the runtime's midpoint splits always divide
  evenly; anything else is not a topology this design recognizes).
- Documents route to subranges by `PackedKeyHash_HH64` of the packed key — the
  same hash the runtime shards by — so a channel's contents are a function of
  the data alone.
- A channel is named `<materialization>_<keyBegin>-<keyEnd>_<stateKey>` (with
  the name components sanitized and hash-suffixed as the classic path does).
  Carrying both subrange bounds means subdivision depths never collide; carrying
  the **state key** means a backfill rotates a binding's channels and its
  checkpoint items together — both start empty — which is what makes a
  backfill the escape from every refusal in this design.
- The offset token is `counter@keyBegin-keyEnd`: a monotonic per-channel
  document counter that never resets, plus the subrange the documents were
  appended under. The range is in the token because the token is the only
  record written at the same instant as the rows it accounts for. A token
  whose range is not the channel's own was written by something else, and
  nothing it counts may be skipped.

### Checkpoint shape

- The driver checkpoint holds **one item per channel, keyed by the channel's
  key-hash subrange** (e.g. `00000000-7fffffff`), under the binding's state
  key: channel name, counter, and the subrange. On runtime v2 every shard
  merge-patches its connector state into one task-global document, so two
  shards of one binding write the same JSON path; keyed by state key alone,
  whichever sibling reduced last would overwrite the other's counter. Channels
  are per-subrange, so sibling shards' keys are disjoint.
- These items are **durable per-binding state, not pending work**. The
  checkpoint-clearing pass of `Acknowledge` — which the Apply-time
  pending-transaction drain also relies on — leaves them in place and reports
  no state update on their account.

### Batching, pacing, and the hot path

- Rows buffer per channel in memory. A batch is appended at **10 000 rows or
  8 MiB**, whichever first; `flush` appends the remainder, so the counter the
  checkpoint records matches exactly what was appended. At most one in-flight
  append per channel preserves ordering while `Store` keeps buffering.
- A task-global ceiling of **128 MiB** of buffered documents across all
  bindings triggers an append of every non-empty buffer, however short of a
  full batch, bounding worst-case resident memory independently of binding
  count.
- Appends are paced to **20 MB/s uncompressed per channel**, the documented
  per-channel ceiling. The sidecar's existing `RECEIVER_SATURATED` exponential
  backoff is load-bearing — it throttles `Store` directly — and is logged at
  info, since saturation is an operating condition rather than a defect.
- All knobs are package `var`s so tests can cut batches cheaply; nothing reads
  them from the endpoint configuration. Because recovery resumes from the
  committed token regardless of how the interrupted attempt cut its batches,
  batching is a pure performance knob and never checkpoint state.
- No time-based flush: the transaction lifecycle already bounds staleness, and
  a timer would race the `Store` stream for the buffer.

### Commit and `Acknowledge`

- The wait for Snowflake to commit each channel's final token happens **inside
  checkpoint production** (`flush`), not in `Acknowledge`. The runtime's
  checkpoint is durable before `Acknowledge` runs, so a commit failing there
  could no longer fail the transaction whose counter it belongs to — the task
  would resume from a counter Snowflake never committed, which the next `Open`
  could only refuse. Failing during checkpoint production leaves the
  transaction to be replayed, and the documents Snowflake did commit are
  skipped by that replay's reconciliation.
- `Acknowledge` is where **channel topology may change** (see below): the
  runtime has durably committed the transaction whose checkpoint the last
  flush produced, and the next `Store` phase has not begun.

### Recovery reconciliation

At `Open`, per channel, with `C` = Snowflake's committed token counter and `K`
= the checkpoint item's counter:

| Condition | Meaning | Action |
|---|---|---|
| `C` absent, `K == 0` | fresh channel | append from 1 |
| `C` absent, `K > 0` | the channel is gone from under a live item | **refuse** |
| token range ≠ channel subrange | a foreign writer's token | **refuse** |
| `C == K` | clean boundary — the steady state | append from `K+1` |
| `C > K` | appends committed that no checkpoint saw | verify the token's range, then skip indices `≤ C` |
| `C < K` | the channel lost committed data | **refuse** |

- `C` absent with `K > 0` refuses because the connector cannot bound what was
  lost: the state is produced both by Snowflake losing the channel and by the
  table being dropped underneath a still-running generation (a backfill's
  re-create, another task, manual DDL). Nothing this connector does retires a
  shard's *own* live channel, so the state always means something outside its
  control.
- The `C > K` guard reads the key range **off the token**, never off the
  checkpoint item: the item records the range of the last transaction that
  committed, while the replay needs the range that appended the documents
  standing beyond it — which by definition no checkpoint saw.
- Because indices are absolute, a single skip threshold captured at `Open`
  suffices for the session; later transactions exceed `C` naturally, with no
  "am I recovering" bookkeeping.

### Shard splits, joins, and convergence

- An aligned split hands each child two of the parent's four channels, whole;
  a join hands the parent all eight of its children's. The inheriting shard
  opens each inherited channel, reconciles its token as above, and continues it
  — including through the rows of an interrupted transaction, so a topology
  change during a backfill neither wedges the task nor duplicates rows. An
  inherited channel that took no rows is adopted the same way.
- An inherited layout works indefinitely at the wrong channel count. At the
  first transaction boundary where every channel is settled, the shard
  **converges in two phases**: it first records its own target layout's
  channels in the checkpoint, and only after that record is durable does it
  retire the inherited channels (deleting their items via merge-patch nulls)
  and route to its own. Either phase can be interrupted and repeats safely.
- A shard refuses when its inherited items do not tile its key range, when an
  item's channel should exist in Snowflake and does not, or when a channel
  carries a foreign token — the same principle throughout: refuse rather than
  guess, with a backfill as the escape.

### Tables, generations, and backfills

- A streaming table's comment carries a **generation marker**
  (`flow_generation: ` + materialization + state key), embedded without
  disturbing the operator-visible comment text.
- `CreateTable` refuses a streaming binding whose resolved binding carries no
  state key — *before* the CREATE, so no unmarked table is left behind — and
  stamps the marker immediately after creation. Every generation of a
  streaming table passes through `CreateTable`, because a backfill of a
  streaming binding **re-creates the table rather than emptying it**
  (`MustRecreateResource`): a truncate leaves the outgoing generation's
  channels valid and pointed at the emptied table, so rows they append next
  would survive and be silently re-materialized. Dropping the table drops
  every channel bound to it, including channels of shards that committed
  nothing and so appear in no checkpoint; the outgoing generation is then
  refused at reconciliation rather than allowed to reopen one.
- `MustRecreateResource` consults the incoming configuration, and widens — never
  narrows — its answer from the *replaced* specification's configuration, since
  a publication may turn the write path off in the same breath as it bumps a
  backfill counter. The replaced spec arrives in whatever shape the control
  plane stored it (endpoint spec wrapper, sops-ciphertext secrets), so its
  config is parsed tolerantly and an unreadable one is an error rather than a
  guessed "no".
- Before appending, the manager checks the table's recorded generation against
  its own and refuses a mismatch: a table recording another task's — or another
  generation's — identity is not this binding's to stream into.
- `Apply` **adopts** a table that an earlier write path created: a table whose
  comment records no generation gets stamped with the binding's current one
  (matching the table by its stored, case-sensitive name even when the
  resource names it differently), so turning the flag on does not force a
  backfill. Consequences documented for users: grants do not survive a
  backfill's re-create, and `retain_existing_data_on_backfill` cannot work on
  this path.

### Switching write paths

- **Off the path is refused.** A binding whose checkpoint still holds streaming
  v2 items may not run on another write path: `Apply` refuses the publication
  (reading the last connector state), and transactor startup refuses again as
  the backstop. The remedy is a backfill, which rotates state key, channels,
  and table together.
- **Onto the path drains first.** Work the previous path staged and did not
  finish is drained through the machinery that staged it: staged-file queries
  and pipe files drain through the transactor's ordinary Apply-time drain, and
  classic-streaming blobs re-open a channel of the previous path solely to
  finish them. A switch that cannot drain pending blobs is refused with the
  count.

### Semantics accepted

Exactly-once for committed transactions; **at-least-once at the abort
boundary**. Rows become visible before the Flow transaction commits, and rows
appended by a transaction that never commits persist. This is inherent to
direct append against a channel and is documented in the connector README,
changelog, and user-facing documentation.

## Testing Decisions

A good test here exercises externally observable behavior — what lands in
Snowflake, what error a user sees, what the connector refuses to do — and never
asserts on the shape of internal state. Snapshot tests are preferred where
output is stable, per repo convention.

### Seams

Four, in descending order of preference; only the topology fake is new, and it
reuses the supervisor's process seam.

1. **The streaming-v2 manager against live Snowflake** — the load-bearing seam:
   happy path, recovery skip (kill the sidecar mid-transaction, replay the
   identical sequence, assert the exact final row count), generation recording
   and adoption, and the full datatype sweep as cupaloy snapshots. Test state
   keys carry a per-run nonce, because Snowflake keeps serving a dropped
   table's channels for a time and a reused key can meet a stale committed
   token.
2. **The topology suite against a protocol-fake sidecar** whose committed
   tokens persist to disk across manager instances — steady-state routing,
   split and join inheritance, split-then-join-back, crash windows on either
   side of the two-phase convergence, empty-channel inheritance, and the
   misaligned-split / lost-channel / foreign-token refusals.
3. **The transactor, client, and gate unit seams** — checkpoint shape (one
   item per channel), commit-precedes-checkpoint ordering, `Acknowledge`
   preserving counters, backfill-recreates-table decisions across
   configuration shapes, generation comment embed/parse/conflict, create-table
   state-key refusal, reconciliation truth table, both directions of a
   write-path switch including the pending-work drain and refusals, and the
   runtime gate itself: configuration validation of the flag pair and JWT
   requirement, `Validate`'s missing-runtime warning, and `Apply`/`Open`
   refusals across flag, template, and spec shapes.
4. **The sidecar suites** — the supervisor against a stdlib-only fake with
   fault injection (ready timeout, exit-before-ready, crash mid-RPC, hang,
   signal escalation, process-group kill, log relay), and the Python pytest
   suite with the SDK stubbed except for the raw-rows guard, which runs against
   the installed SDK under `SNOWPIPE_SDK_GUARD=required` in the image build.

Routing needs no golden vectors of its own: the packed-key hash is pinned to
the runtime's implementation by the golden-vector test in `go/keyhash`, which
this path routes through.

The materialization integration harness is deliberately not a seam. It drives
tasks through `flowctl raw preview-next`, which runs the **v2** runtime — the
very condition the gate exists to detect the absence of — so a
refuses-outside-v2 fixture is not expressible there, and `flowctl preview` is
being reimplemented on preview-next, so a fixture built on it would become a v2
run wearing a v1 label. The gate is covered at the unit seam, and end-to-end
coverage of the refusal is deferred until a harness can genuinely present a
non-v2 runtime, rather than simulated.

### Benchmarks

- A row-encoding benchmark holds the direct writer against the map-and-marshal
  reference it replaces.
- 200 GB delta-updates scenarios at 1 KB and 1 MB document sizes join the
  materialization benchmark suite, sized for four-shard arms, to measure the
  path at the scale the four-channel design is for.

## Out of Scope

- **Any change to flow.** No new protocol field, no capability negotiation, no
  runtime work. Runtime detection is done entirely from the shard template.
- **Making the new path the default.** It stays behind the feature flag;
  flipping the default is a later decision, once the deprecation timeline is
  announced and runtime v2 is broadly deployed.
- **Retiring the classic BDEC path.** It remains the only streaming option for
  delta-updates bindings on runtime v1, which is most tasks today.
- **Migrating existing checkpoints.** The feature flag has never been released,
  so no legacy checkpoint format is supported and no migration code is
  written.
- **Coordinated multi-shard apply.** Per-shard state patches merged into the
  task's driver checkpoint are used; coordinating apply across shards is
  future work.
- **Standard-updates bindings** over the new path: the high-performance
  architecture is insert-only.
- **Non-JWT authentication**: the SDK requires key-pair credentials.
- **User configuration of channel count, batch sizes, or pacing.** All are
  fixed knobs; a configuration surface would trade recovery simplicity for
  tuning nobody has asked for.

## Further Notes

- The four-channels-per-shard scheme is the fixed-layout half of the
  "Idempotent Shards with Document Counters" strategy in flow#2581, brought
  inside a single connector: channel identity is a function of the key-hash
  space rather than the shard topology, which is precisely what lets topology
  changes inherit channels whole. The counter guard deviates from that
  design's implicit same-split replay assumption by checking ranges explicitly
  and refusing, because the runtime does not tell connectors about shard
  splits.
- Determinism of the replayed document sequence is a property of flow, verified
  rather than assumed: idempotent replay is one-shot and never extends, the
  combiner sorts stably and preserves arrival order, its spill thresholds are
  compile-time constants, and the shuffle log is durable and read in LSN
  order. For delta-updates bindings the combiner still applies associative
  reduction, so *documents appended* is not *documents read* — the counters
  count what is appended.
- Every refusal is terminal by design, with no escape hatch in the connector.
  The operator's remedy is always the same ordinary operation — bump the
  binding's backfill counter — and the connector offers nothing that trades
  correctness for liveness.
- Branch shape: the history builds bottom-up so each commit compiles and
  carries its own tests — sidecar package, supervisor and
  RPC client, row encoding, routing math, channel manager, topology suite,
  feature flag, connector wiring, runtime gate, live-Snowflake suite, switch
  suite, benchmarks, README, changelog. User-facing documentation lands as a
  separate PR against flow.

jacobmarble:
**Description:**

Closes #4324

Implements a client for [Snowflake Snowpipe High-Performance architecture](https://docs.snowflake.com/en/user-guide/snowpipe-streaming/data-load-snowpipe-streaming-overview).

The Go connector supervises a small Python sidecar that wraps Snowflake's SDK because Snowflake does not provide a Golang SDK for Snowpipe HP.

**Workflow steps:**

Enable the `snowpipe_streaming_v2` feature flag in the endpoint's advanced configuration. The endpoint must use key-pair (JWT) authentication, and the task must be on Estuary runtime v2 (`enable-runtime-v2` shard flag).

**Notes for reviewers:**

The commits are ordered for review and build on each other: sidecar → supervisor/RPC → row encoding → routing → channel manager → connector wiring → runtime gate, with test suites (topology, live Snowflake, write-path switches) following the code they pin. Reading commit-by-commit should work well.

**Checklist:**


jacobmarble:
Superseded: this branch's history now lives on jgm-snowpipe-streaming-v2 (same commits, 027a69911).

PR #5154, opened 2026-08-28
docs: source-netsuite add metadata read advanced option

Alex-Bair:
**Description:**

Documentation updates for https://github.com/estuary/source-netsuite/pull/246.

**Notes for reviewers:**

I'm not really sure how we want to handle changelogs for connectors the don't reside in this repo like `source-netsuite`, and I'm open to whatever ideas folks have. It seems to me that the purpose of the `CHANGELOG.md` is to surface user-visible changes, but users can't view private repos like `source-netsuite` so the changelog shouldn't live there. That's the extent of my opinion on the matter.

**Checklist:**


aeluce:
Good point re: handling changelogs. It's easy to overlook the few miscellaneous connectors that don't live in this repo.

Part of @danthelion's idea behind the changelog was to eventually surface the release notes through docs. TBH, I don't have a clear picture of what that would look like yet. Perhaps once that's decided, we could add the changelog directly to the Netsuite docs using the final intended format, or keep the changelog in a new folder under `/connectors/docs` that we port over and process similar to the redirects file.

cc @jwhartley as well in case you have a preference.

aeluce (approved):

PR #5155, opened 2026-08-28
source-airtable-native: inject empty values for fields airtable doesn't include

Alex-Bair:
**Description:**

Airtable omits an emptied-out field from its API response entirely instead of sending `null`. Under this connector's `merge` reduction strategy (needed so scheduled formula-field refreshes can update just their own fields without erasing the rest of the record), an omitted key means "no change" - so a field that was ever set and then genuinely cleared in Airtable stayed frozen at its last value forever. This showed up as non-cleared out values for documents materialized in users' destinations.

Records are now validated against the set of fields each fetch actually requested, and any still missing afterward are explicitly filled in - with the type's own blank value ("", [], false) where that's the type's only possible empty state, or null otherwise. That turns the ambiguous omission into an explicit clear, so `merge` correctly wipes the field instead of preserving it. A formula field caught mid-computation-error is left alone rather than filled, since that's a transient glitch, not a real clear.

Motivating Slack thread: [link](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1787690565839629)

**Checklist:**

### DIFF
--- a/source-airtable-native/CHANGELOG.md
+++ b/source-airtable-native/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Changelog
+
+## 2026-08-31
+
+### Fixed
+- Fields cleared in Airtable (e.g. all options removed from a multi-select) are now correctly captured as cleared, instead of permanently retaining their last non-empty value.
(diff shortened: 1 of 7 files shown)


nicolaslazo (approved):
LGTM

PR #5156, opened 2026-08-29
source-stripe-native: stop leaking a logger per connected-account subtask

### LINKED ISSUE #3887: source-stripe-native: OOM'ing when capturing from many connected accounts
The Stripe native capture connector is OOM'ing for captures that have many connected accounts and have `capture_connected_accounts` set to `True`. The root cause is unknown currently, but is suspected to be a memory leak within the `priority_capture.py` logic.

Alex-Bair:
**Description:**

There's been a slow memory leak present in the `priority_capture.py` code for `source-stripe-native`, and Claude helped me finally figure out what's causing it.

`Task.spawn_child` derives the child's logger from the name suffix via `Logger.getChild`, and Python's logging registry retains every distinct logger name for the life of the process. The priority queue named each subtask "backfill.<acct>" / "incremental.<acct>" under one of 40 rotating worker `Task`s, so every rotation registered a new ~0.6 KB logger and the connector climbed into its 1 GiB limit with tens of thousands of accounts.

This PR fixes the memory leak by naming the children "backfill" / "incremental" so the registry is bounded by `bindings x workers x 2`, and log a subtask failure with its `subtask_id` before re-raising so the account stays attributable.

Closes https://github.com/estuary/connectors/issues/3887.

**Checklist:**

### DIFF
--- a/source-stripe-native/CHANGELOG.md
+++ b/source-stripe-native/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 2026-08-31
+
+### Fixed
+- Memory no longer grows without bound while capturing from connected accounts. Every
+  per-account subtask registered a permanently-retained logger, leaking roughly 0.6 KB
+  per subtask until the connector exceeded its memory limit. Subtask names and the log
+  `source` field no longer embed the account id; it is carried in the `subtask_id` log
+  field instead.
+
 ## 2026-08-20
 
 ### Changed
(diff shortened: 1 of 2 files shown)


nicolaslazo (approved):
LGTM. I'll try to think if we have any clear task lifecycle signals that could let us know if it's safe to deallocate a logger at the CDK level

Alex-Bair:
Considering CDK based captures are intended to run for extended periods of time, potentially with many bindings & with configurable `interval`s, I'm not sure the CDK is able to reliably determine when it's safe to deallocate a logger. I'm not sure we should do anything at the CDK level about this yet since it was an isolated instance in a single connector, but if we did want to do something, something like "monitor how many loggers are instantiated by a connector, log a warning if it exceeds N or constantly increases across a connector's lifecycle"; I'd hope that would be enough for us to figure out that we shouldn't be creating so many loggers.

PR #5157, opened 2026-08-30
materialize-redshift: post-commit apply with the shard-zero coordinator

### LINKED ISSUE #4719: Materialization scale-out: Redshift
transition to post-commit apply

### LINKED ISSUE #5110: materialize-redshift: convert to post-commit apply (Recovery Log with Idempotent Apply)
## Problem Statement

A user materializing into Redshift cannot scale their materialization out. The
task runs on one shard however much data it has to move.

The reason is the pattern the connector uses. materialize-redshift is a
"Destination is Authoritative" connector, and this class of materializations
cannot support scale-out.

The same design also puts the entire destination commit on the critical path
of every transaction. Each transaction takes a table-level lock on the
checkpoints table, holds it across every binding's COPY, MERGE and DELETE, and
ends with a fence UPDATE. The runtime cannot commit its recovery log until all
of that returns, so nothing overlaps the destination work and the
per-transaction floor grows with the number of bindings.

Users experience this as a Redshift materialization that cannot be made faster
by giving it more shards, and that spends a fixed cost per transaction on
bookkeeping no matter how little data moved.

## Solution

Convert materialize-redshift to the "Recovery Log with Idempotent Apply"
pattern with the shard-zero-executes coordinator that materialize-databricks
already uses, so that the task scales out on the v2 runtime.

After the change, `Store` continues to stage documents to S3 and records what
it staged in the connector state. `StartCommit` returns that state and nothing
else, so the runtime commits its recovery log immediately. The destination
commit moves to `Acknowledge`, where it no longer blocks the runtime and can
overlap the next transaction's staging.

Exactly-once delivery is preserved for every binding type, including
delta-updates bindings, which have no key and therefore cannot be made
idempotent by merging. The checkpoints table the connector already creates for
every Redshift task stops holding a fence and a runtime checkpoint and starts
holding applied-transaction tokens, written inside the same transaction as the
data they vouch for. A recovered entry whose token is already committed is
skipped rather than re-applied.

Existing tasks cross over without replaying or losing a transaction, without a
backfill, and without any new Redshift privilege.

Under multiple shards every shard stages its own files to S3, and only the
primary shard — the one whose key range begins at zero — applies. It commits
every shard's staged files for a table with a single COPY and MERGE, so the
destination does one apply per table per transaction however many shards
stage into it. Benchmarking phase 1 showed why this matters: with each shard
applying its own files, Redshift ran the four shards' COPY/MERGE transactions
one after another, and the apply cost grew linearly with the shard count.

## Implementation Decisions

**Pattern and lifecycle**

- materialize-redshift moves from "Destination is Authoritative" to "Recovery
  Log with Idempotent Apply". `RecoverCheckpoint` stops returning a
  destination checkpoint once a task has crossed over, and the destination
  commit moves out of the function returned by `Store` and into `Acknowledge`.
- The endpoint keeps declaring its checkpoints meta-table, because that is
  what creates the row the tokens live in; it is only the fence that goes.
  `InstallFence` stays and  becomes a no-op returning the fence it was given.
  Setting `MetaCheckpoints` to nil instead would stop `Setup` creating the table,
  and a new task's first apply would then insert into a table that does not exist.
- `Store` keeps staging to S3 exactly as it does now, and additionally records
  one entry per binding in the connector state, keyed by the binding's state
  key and then by the shard's key range. `StartCommit` emits that state as a
  merge patch and returns immediately. Every task uses this shape, grouped by
  state key so the primary can read a binding's pending work with one lookup;
  a single-shard task's entries nest under its full key range.
- The apply is one Redshift transaction per `Acknowledge` covering every
  binding with pending work, preserving today's cross-binding atomicity and
  today's per-binding statement order (deletes before stores).
- The connector state is parsed when the transactor is built, from
  `open.StateJson`, and not only in `UnmarshalState`. The boilerplate calls
  `RecoverCheckpoint` as soon as `NewTransactor` returns, while the runtime
  calls `UnmarshalState` later, so a connector that parses its state only in
  `UnmarshalState` reaches the crossover decision believing nothing is staged
  — and hands back the old checkpoint for a task that has staged work, which
  is the double-apply this design exists to prevent.
- `Acknowledge` returns a nil state update when it processed nothing, which is
  what tells the Apply RPC's drain loop it has converged.

**Exactly-once mechanism**

Glossary:
- Manifest: A manifest is a small JSON file you put in S3 that lists exactly which objects a COPY should load. You point COPY at the manifest instead of at the data. Manifest key is the object path of the manifest file in S3.

- The checkpoints table that every Redshift task already has is repurposed as
  the applied-token store. It is not a new object, and no new grant or table
  ownership is required. Its `fence` column is NOT NULL and so stays, written
  as zero and read by nothing; the tokens replace what the `checkpoint` column
  held. The row is upserted: an UPDATE, falling back to an INSERT when the
  materialization has no row yet. The UPDATE matching more than one row is an
  error rather than a success: nothing the connector writes can produce two
  rows for one key range, so a second one means bookkeeping it does not
  understand.
- A staged transaction's token is an id minted when its first row is staged,
  carried in its entry. `Store` writes only data files, under a prefix unique
  to the transaction; the manifests Redshift copies from are written by the
  apply, over every shard's files for a table, and deleted afterwards.
- The apply transaction's final statement writes the tokens of everything it
  just applied, into the primary shard's row. The payload is scoped by the key
  range of the shard that staged each transaction, so the primary's row
  accounts for its peers' work as well as its own, keyed by state key and then
  by range for the same reason the connector state is:

  ```
  { "<stateKey>": { "<keyBegin>-<keyEnd>": "<manifestKey>" } }
  ```

- Because the token is written in the same transaction as the COPY, MERGE and
  DELETE, a committed token proves the data committed and a rolled-back apply
  leaves no token. On recovery the connector reads the tokens, skips any
  pending entry whose token is present, and applies the rest.
- A recovered entry keeps whichever path it was staged for. The token settles
  whether its data committed, for keyed and delta-updates bindings alike.

**Scale-out (shard-zero coordinator)**

- The primary shard is the one whose key range begins at zero. Only it
  applies. Non-primary shards stage to S3, emit their `StartCommit` patch, and
  do nothing in `Acknowledge`: they recover no state and hold no bookkeeping,
  because the primary replays the entire consolidated state document.
- The primary learns of its peers' staged transactions from the aggregated
  `StartedCommit` patches the runtime hands `Acknowledge` (`statePatches`),
  folding every range but its own into its pending set; its own contribution is
  already there. A `null` patch is a full state replacement, which no shard
  emits, and is an error.
- One apply per table per transaction: for each state key the primary unions
  every range's store files into one COPY manifest and every range's delete
  files into another, then runs the same statements as a single shard would —
  one COPY into the staging table and one MERGE (or one direct COPY when no
  entry needs a merge), and one COPY plus DELETE for the deletions. Ranges
  cover disjoint slices of the key hash space, so the merged source never
  carries duplicate keys.
- The manifests are written under a fresh prefix for every apply, single-shard
  tasks included, and deleted afterwards whether or not the apply committed.
  The entries' files are deleted once their tokens are committed.
- Tokens are written per `(range, state key)`. An entry already carried by the
  primary's row is left out of the combined manifest rather than re-copied, so
  a primary restart mid-apply re-applies exactly the entries that did not
  commit.
- Entries staged under a key range that no longer exists — after a change of
  shard count — stay in the state until the primary applies them, which it
  does for any range whose state key has a live binding.
- The Apply RPC's drain runs the transactor over the task's full key range and
  so as the primary.

**Crossover for existing tasks**

- Nothing extra marks the crossover: the row is its own marker. Today's
  connector keeps a fence and a base64 runtime checkpoint in it, and the new one
  keeps applied-transaction tokens. A leading `{` tells the two apart, since
  base64 cannot produce one.
- At `Open` the connector returns the row's runtime checkpoint only when the row
  is still in the old format *and* the connector state carries no staged
  entries. The old format alone means no apply has committed, but the runtime
  may already have committed a transaction whose entries are staged, and
  returning the old checkpoint then would rewind the runtime and apply that
  transaction twice.
- Today's connector emits no connector state at all, so an upgraded task starts
  with an empty state and the hand-off is exact: the row can legitimately be one
  committed transaction ahead of the recovery log, which is what today's model
  relies on.
- The first apply rewrites the row into the new format within its own
  transaction. Until that commits, either the state is empty and the row is
  authoritative, or the state has entries and the recovery log is. A transaction
  staging no entries leaves the state empty and is harmless to rewind across,
  having moved no data.
- A materialization with no row is new rather than crossing over: it gets no
  checkpoint at `Open`, and its first apply inserts the row.
- The connector ships as a single unconditional code path with no feature flag,
  since turning one off after the row has been rewritten would resume from a
  checkpoint that stopped advancing. Per-task rollout control comes from image
  pinning, and the changelog records that downgrading requires a drain and would
  need a backfill.

**Concurrency control**

- The fence nonce, its startup installation, and the fence check are removed.
  Recovery-log fencing already rejects a zombie's state update before the log
  commit in this pattern, and the token makes a duplicate apply a no-op.
- The table-level lock on the checkpoints table is kept, because the
  serialization errors it prevents between materializations sharing a metadata
  schema are unrelated to fencing. It moves from the start of the transaction
  to immediately before the token write, shrinking the contention window from
  the whole apply to one small statement. All applies then take that single
  lock in the same order, so they cannot cycle.

**Staged-transaction entries**

- An entry carries its id, the keys of the data files it staged for stores and
  for deletes (kept apart so that the primary can combine shards' store files
  and delete files separately, and so that a later session can clean up every
  object knowing only the entry), whether a MERGE is required, and the columns
  that must be widened to VARCHAR(MAX). Whether the binding has deletions is the
  presence of a delete manifest rather than a flag of its own. It does not
  carry pre-rendered SQL.
- Deleting an applied transaction's objects chunks them into the 1000 per
  call that S3 allows, rather than refusing at staging time to process a
  transaction with more than 1000 files. The old fixed manifest key made that
  cap easy to enforce in one place, but it is an arbitrary limit on how large
  a transaction may be.
- Every field of an entry is written even when empty. Entries travel as
  RFC 7396 merge patches, which merge field by field, so an omitted field
  inherits whatever the previous transaction's entry held for it — and a stale
  delete manifest names keys to delete from the target table. Spelling every
  field out makes each patch replace the entry outright.
- The `StartCommit` patch carries only the entries this transaction staged,
  not everything pending. Entries whose state key no longer has a binding are
  never applied and never cleared, so emitting the whole pending set would
  re-send them on every transaction for the life of the task.
- The COPY, MERGE and DELETE statements are rendered in `Acknowledge` from the
  entry plus the resolved table, rather than at startup. Correctness across a
  schema change rests on the Apply-drains-pending-work contract already in the
  shared materialization boilerplate, which commits a binding's staged work
  through `Acknowledge` before altering its resource.
- The prefixes that the COPY error handler searches `sys_load_error_detail`
  with come from the files the failed COPY listed, not from a live staged
  file. A recovered entry was staged by a previous session under a different
  prefix, and a combined manifest spans several shards' prefixes.
- `copyFromS3Params.Columns` goes: the COPY template never reads it, and
  rendering per transaction makes that plain.
- VARCHAR widening stays out of the apply transaction, because Redshift
  forbids ALTER in a transaction block, and runs before the COPY. It is driven
  by the entry's widen list, so the shard that applies can widen for a column
  another shard discovered. Widening is monotonic and idempotent, so being
  outside the transaction costs nothing. Staging-table widths continue to come
  from the observed table shape, upgraded to VARCHAR(MAX) only for listed
  columns, which preserves the sizing that avoids declaring every string
  column at maximum width.

**Shape**

- The checkpoints row is reached through a `tokenStore` object constructed
  with the references its work needs — the quoted table identifier, and the
  materialization and key range that identify the row — with small methods
  hung off it: a read, a write, and the parse that tells a token payload from
  a row in the old format. Prefer this over free functions that each retake
  the same four arguments. The write takes the table lock itself, so the lock
  cannot drift away from the statement it exists to serialize.
- `Acknowledge` and the apply share one view of the bindings that have work
  staged for them from any shard — each binding with its pending entries, every
  entry carrying the key range that staged it — computed once rather than
  looked up again at each call site.
  The rule that entries with no live binding are left pending then lives in
  the iterator instead of being restated at each call site.
- A binding accumulates its entry as a `*stagedTransaction` over the course of
  the Store phase, started by the first row and taken by `StartCommit`, rather
  than as loose per-transaction fields on the binding that have to be reset
  one by one. There is then one thing to reset between transactions, and no
  way for half of an entry to survive into the next one.
- Code that is inline today stays inline. Moving the apply out of `StartCommit`
  and into `Acknowledge` is a move: the per-binding body wants to read as the
  same statements in a new place, so that a reviewer can see nothing else
  changed with them. Do not take the opportunity to lift its parts out into
  helpers — the widening, the staging-table metadata, the COPY rendering, or
  the per-binding apply itself — where they are used once.

**Unchanged**

- Endpoint and resource configuration, including the absence of any table
  comment or additional-SQL field.
- The load path's behaviour: staging load keys, the temporary load table, and
  the unioned load query all work as they do now. Its COPY is rendered per
  transaction rather than at startup, because manifest keys are no longer
  fixed, but nothing about what it does changes.
- Table creation, table comments, column comments, and the column migration
  behavior.
- Redshift privileges required of the connector's user.

## Testing Decisions

A good test here asserts what an operator or a downstream query could
observe: the rows in the destination table, the connector state the runtime
persisted, and the checkpoint the connector handed back at `Open`. It does not
assert which statements were issued, in what order, or what the transactor
holds in memory. The exception is the pure state-handling logic, which is
externally observable only as a state document and is worth testing directly
at that boundary.

Four seams, all of them existing entry points. No new seam is introduced.

1. **The materialize, apply, and migrate integration tests** (highest seam).
   These already drive the connector through a real runtime against a real
   Redshift cluster and snapshot the destination contents. The materialize test
   runs at two shards, as materialize-databricks' does, so that the primary
   applies a peer's staged files against a real warehouse on every run; the
   apply and migrate tests run single-shard. They are the primary proof that
   moving the apply into `Acknowledge` and merging shards' files changed no
   observable behavior.

2. **The shared Apply-drain test helper** (existing seam, new Redshift call).
   materialize-databricks' call is the prior art: a `seedPending` callback
   writes a pending entry into connector state and a `verifyDrained` callback
   asserts the staged rows were committed before the schema update ran.
   Redshift now depends on that contract, so it gets the same coverage, seeded
   with the structured entry shape rather than databricks' pre-rendered
   queries.

3. **The slot the fencing test occupies today**, renamed for Redshift to an
   idempotency test, because what it verifies is no longer fencing. The shared
   `RunFencingTest` helper stays where it is for the six connectors that still
   have fences; Redshift stops calling it. The Redshift test builds the
   endpoint and manipulates the checkpoints table directly, which is the same
   injection point the fencing test uses, and covers the two properties a
   higher seam cannot reach because it cannot interrupt an apply:
   - a pending entry whose token is already committed is skipped, leaving the
     destination row count unchanged, exercised against a delta-updates
     binding: with no MERGE promotion the token is the only thing standing
     between a re-applied entry and duplicated rows, and a keyed binding's
     MERGE would hide that. Assert both directions — applied when no token is
     present, skipped when it is — so the test cannot pass by doing nothing;
   - a row still in the old format, with an empty connector state, yields that
     row's checkpoint at `Open`, and once the first apply has rewritten the
     row the connector returns nothing;
   - a row still in the old format whose connector state carries entries
     yields nothing, which is the case that would otherwise rewind the runtime
     and apply a transaction twice.

   The crossover case is Redshift-specific one-time migration behavior that a
   general-purpose suite would not model, so it belongs here permanently. The
   double-apply case is a generic exactly-once invariant and is the interim
   guard until the consistency test suite specified in #4956 exists; its
   crash-and-replay and delta-updates duplication scenarios are the durable
   home for that verification against a real runtime with real interruption,
   and this local case can be retired in favour of them.

4. **A state unit test file** (existing pattern, new file). Entry parsing,
   including the empty state a task crossing over opens with and the rejection
   of an entry carrying fields this connector does not understand; the choice
   of token; the discrimination of a token payload from a row still in the old
   format; that one entry's merge patch replaces the previous entry outright
   rather than inheriting its fields; and the coordinator's routing — the
   primary recovers every range and folds peers' patches, a non-primary shard
   recovers and acknowledges nothing, and pending work groups every range's
   entries per binding while leaving entries with no live binding pending. No
   warehouse, no network.

The SQL generation snapshots cover the template changes.

The materialize snapshot changes only in that the two-shard harness does not
capture connector state; no materialized value changes.

**Performance verification.** Measured with the benchmark harness on a
`c3-standard-4-lssd` in us-east4-a, the same VM type and zone as the recorded
baseline, one run at a time, at one, two and four shards, against the
pre-change connector run on the same VM and runtime as the control. The
recorded baseline was measured with v1 `flowctl preview`, and the same old
connector runs about 1.5x slower under `preview-next`, so the control arm
rather than the recorded number is the comparison. Single-shard throughput is
not expected to move much: the apply only overlaps the next transaction's
load-key staging, because the load iterator waits for the acknowledgement
before the load query runs. The gain is in the shard count.

Two harness properties matter for reading the numbers. The timed window ends
at the last commit, and a post-commit apply connector commits transaction N
during the `Acknowledge` that N+1 waits on, so the scenarios gain a tiny
trailing transaction that pulls the last apply inside the window. And the
connector's integration-test config carries `allow_existing_tables_for_new_bindings`,
which makes the connector disable the runtime's load optimization; the
benchmark uses a copy of the config without it.

## Out of Scope

- Row-level verification of multi-shard runs by the benchmark harness, which
  drops the destination table when a run ends. The two-shard integration test
  is what checks the merged apply's contents.
- Per-binding apply parallelism. Bindings still commit serially within one
  transaction, as they do today. Whether Redshift rewards concurrent MERGEs on
  separate connections is a question for measurement, and answering it would
  also mean re-keying the token rows per binding.
- Parallelising the load path's per-binding COPY statements. It is a real
  performance lever and it is independent of this change.
- Dropping the checkpoints-table lock in favour of retrying on serialization
  failures. The token makes such a retry safe, and the conflict signatures are
  identifiable, but it trades a short wait for a full redo of the apply and
  should be decided with measurements.
- Any change to the load path, to table creation, or to column migration.
- Any new Redshift privilege, role, or ownership requirement.

## Further Notes

**Why the checkpoints table rather than no destination metadata at all.** The
appealing version of this change leaves no destination-side bookkeeping,
because Databricks and Snowflake need none: their COPY skips files already
recorded in the table's load metadata, so re-applying a recovered entry over
the same file paths is inherently idempotent. Redshift's COPY has no
equivalent, so an append has no identity and delta-updates bindings cannot be
made idempotent by any rewriting of the apply. Every substitute was tested
against a live Redshift Serverless cluster and rejected:

- The load-history system view keys on the manifest URI verbatim and is
  readable by a plain user with no grants, but it is query-level rather than
  commit-level: a COPY inside a transaction that then rolls back is still
  recorded as completed, with an identical loaded-row count. Using it alone
  would skip an entry that never committed, turning a duplication risk into a
  loss risk.
- The commit-level load view is permission denied on Serverless even for a
  superuser, consistent with the note already in the connector about another
  view in that family being available only on provisioned clusters.
- Redshift's auto-copy job, which does track loaded files, refuses key-based
  credentials and requires an IAM role. The connector's configuration carries
  an access key and secret, so adopting it would be a breaking configuration
  change plus per-customer setup in AWS, and its file tracking lives in system
  catalogs rather than in the transaction.
- A table comment is genuinely transactional, so it rolls back with a failed
  transaction, reads back cheaply, holds far more than enough text, and
  carries nothing a user authored, since the connector generates it. But it is
  owner-only, and on Redshift, unlike PostgreSQL, ALTER TABLE succeeds for a
  non-owner holding full grants. The connector therefore requires no ownership
  today, and using the comment would introduce that requirement and break
  tasks pointed at tables the connector did not create.
- Joining the load history's transaction id to the transaction-history view
  does yield a genuinely commit-authoritative answer, distinguishing a
  committed load from a rolled-back one. It was rejected on cost rather than
  soundness: the view is superuser-only unless the connector's user is granted
  a monitoring role, which is a new prerequisite for every existing task and
  exposes cluster-wide monitoring data; both views reach roughly a week back,
  so an entry pending longer resolves to nothing; and load rows with no
  transaction row do occur in practice, though only for temporary-table loads.

**Scale-out makes the append problem larger, not smaller.** Under the
coordinator model, shard zero applies work that other shards staged, so a
coordinator restart re-applies peers' entries. The token mechanism holds for
that case because the payload is scoped by the key range that staged each
transaction, and an entry whose token the primary's row already carries is
left out of the combined manifest.

**Companion reading.** The scale-out strategy taxonomy is estuary/flow#2581,
which places Redshift under "Authoritative Remote Store" and identifies
post-commit apply with a coordinator as the strategy to move it to. The
databricks precedents are #4770, which introduced range-scoped state and the
shard-zero-executes coordinator, and #5074, which moved entries from
pre-rendered queries to structured metadata so that one MERGE could cover
every shard's files. The pattern contract itself is in the repository's
materialization documentation.

### LINKED ISSUE #5166: materialize-redshift: transaction staging more than 999 files fails at commit and cannot make progress
## 1. Priority
Low - one tenant, and bounding transaction size via `maxTxnDuration` appears to avoid the condition. That workaround has been applied on the reporting task but has not yet been observed to complete a commit.

## 2. Problem description
- Expected: a materialization transaction of any size either commits or fails in a way that lets the task make forward progress.
- Observed: a transaction that stages more than 999 data files fails at `StartCommit` with the error below. The task restarts, re-reads from the same point, stages the same volume, and fails identically. No data is ever committed, so the destination table stays empty while the task reports as running between failures.

```
runTransactions: txnStartCommit: store.StartCommit: transactor.Store: final binding flushing staged files for collection "<collection>": flushing store file: cannot process transaction having more than 1000 files: had 2395 files
```

Observed file counts across four attempts: 1667, 2310, 2395, 1670. Raw timestamps: `2026-08-31T19:33:51.243839065Z`, `2026-09-01T03:52:04.756982026Z`, `2026-09-01T12:04:58.273639708Z`, `2026-09-02T00:04:00.737166350Z`.

The error is raised in `materialize-redshift/staged_file.go` at the `len(toDelete) > 1000` check in `flush`.

- Steps to reproduce:
  1. Backfill a collection into Redshift at a read rate high enough that one read window yields more than roughly 45M documents. Each staged file is flushed at 250 MB of compressed output (`go/writer/json.go`, `DefaultJsonFileSizeLimit`); on the reporting task this worked out to about 45,500 documents per file.
  2. The read phase ends at `maxTxnDuration` (20 minutes by default for materializations). The store phase that follows is not bounded by that setting.
  3. The connector stages the full read volume, exceeds 999 files, and fails at commit with the error above.

Observed on the reporting task: the read phase ended exactly on the 20 minute default (`finished reading load requests`, `took: 20m0.606377074s`) having read 108,851,244 documents. The store phase that followed ran `7h50m59s` and staged 2,395 files before failing. Other attempts: 76,094,655 documents / 5h48m store; 105,168,734 / 7h56m; 76,096,955 / 5h36m.

## 3. Scope and prevalence
One tenant, one task, one binding. First observed 2026-08-31. Requires one read window to yield more than roughly 45M documents, which in practice needs a large collection backfilling at high read rate. The reporting task had no `maxTxnDuration` set and so used the 20 minute materialization default. No prior report of this error and no existing issue.

The reporting task has since been set to `maxTxnDuration: 3m` to reduce the per-transaction read volume. The first read window under that setting ended at `took: 3m0.024863511s` having read 17,403,019 documents, which at the observed rate of about 45,500 documents per file projects to roughly 400 staged files. Whether the transaction commits has not yet been observed at time of filing.

`materialize-boilerplate`'s shared `StagedFiles` takes `fileSizeLimit` as a parameter and has no equivalent file-count check; `materialize-redshift` uses its own `stagedFile` implementation.

## 4. References
- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1788327196176669
- Related prior art: #4893

mdibaiee:
**Description:**

Converts materialize-redshift from "Destination is Authoritative" to "Recovery Log with Idempotent Apply" with the shard-zero coordinator, per #5110.

- `Store` stages to S3 and records one entry per binding in the connector state, keyed by the shard's key range then by state key. `StartCommit` returns that state and nothing else, so the runtime commits its recovery log immediately; the destination commit moves to `Acknowledge`.
- The checkpoints table stops holding a fence and runtime checkpoint and records the manifest key of each applied transaction, written in the same Redshift transaction as the data, so a recovered entry is applied exactly once — delta-updates bindings included. Existing tasks cross over without a backfill.
- Only the shard whose key range begins at zero applies. It folds its peers' `StartedCommit` patches into its pending set and commits every shard's staged files for a table with a single COPY and MERGE per transaction, so the destination cost does not grow with the shard count.

Closes #5110. Closes #4719.

**Workflow steps:**

No configuration changes. Tasks scale out on the v2 runtime; on v1 nothing changes for the user. Downgrading a task to an earlier image requires draining it first and would need a backfill (see the changelog).

**Notes for reviewers:**

Benchmark with `tests/benchmark/materialize` (`small-docs`: 2 × 25M × 128 B documents; `large-docs`: 2 × 5,000 × 1 MB), serial runs against Redshift Serverless in us-east-2, runtime v2. OLD is the connector at `40f21fe8b`; NEW is this branch. The recorded baselines (v1 `preview`) are not comparable across runtimes, so OLD was re-run under the same harness.

| VM | arm | shards | large-docs wall | small-docs wall | small-docs docs/s |
|---|---|---|---|---|---|
| c3-standard-4-lssd | OLD | 1 | 146.5 s | 1577 s | 31.7k |
| c3-standard-4-lssd | NEW | 1 | 140.2 s | 1553 s | 32.2k |
| c3-standard-4-lssd | NEW | 2 | 131.5 s | 1103 s | 45.3k |
| c3-standard-4-lssd | NEW | 4 | 99.1 s | 1036 s | 48.3k |
| c3-standard-22-lssd | NEW | 1 | 97.6 s | 1116 s | 44.8k |
| c3-standard-22-lssd | NEW | 2 | 80.9 s | 718 s | 69.6k |
| c3-standard-22-lssd | NEW | 4 | 70.9 s | 505 s | 99.1k |

Single-shard throughput is unchanged; the Redshift apply per transaction stays flat (~26–35 s for 25M rows) as shards are added, where per-shard applies would serialize on the table. What remains at 4 shards is on the runtime side (estuary/flow#3444).

Trailer variants of the scenarios (a tiny closing transaction) put the last transaction's apply inside the timed window, and the bench config drops `allow_existing_tables_for_new_bindings` so the runtime's load optimization stays on; both are noted in the benchmark README.

**Checklist:**


mdibaiee:
Ran the materialization consistency suite ([flow#3294](https://github.com/estuary/flow/pull/3294), `e6989d3cecd`) against this branch (`be953564`), declared class `postCommitApply`. **17/17 passed**, 1266s wall at 4 threads.

| Scenario | Result |
|---|---|
| baseline | PASS 261s |
| crash-between-commits | PASS 298s |
| crash-mid-store | PASS 270s |
| crash-at-flush | PASS 277s |
| recovery-applies-committed-work | PASS 265s |
| recovery-reconciles-with-destination | PASS 237s |
| destination-ahead-of-checkpoint | PASS 213s |
| at-least-once-never-loses | PASS 292s |
| split-during-store | PASS 264s |
| split-during-commit | PASS 282s |
| split-after-commit-before-apply | PASS 272s |
| split-lands-on-prepared-transaction | PASS 218s |
| join-after-split | PASS 673s |
| crash-in-split-leader | PASS 243s |
| crash-in-split-non-leader | PASS 253s |
| zombie-at-start-commit | not-applicable (`RemoteAuthoritative` only) |
| every-scenario-is-reached-by-a-test | PASS |

15 scenarios verified, 1 not-applicable. The only exemptions were 7 × `[monotonicity]`, expected for a named remote subject; the set-based exactly-once checks were not exempt anywhere, and no `RUNTIME GAP OBSERVED`.

materialize-redshift/driver.go
```diff
@@ -281,9 +285,11 @@ func NewDriver() *sql.Driver[config, tableConfig] {
...
 
 			return &sql.Endpoint[config]{
-				Config:              cfg,
-				Dialect:             dialect,
-				SerPolicy:           boilerplate.SerPolicyStd,
+				Config:    cfg,
+				Dialect:   dialect,
+				SerPolicy: boilerplate.SerPolicyStd,
+				// Creates the row that holds applied-transaction tokens; the
+				// fence itself is unused.
```
jacobmarble, line 292: Would it be OK to delete this comment?
```suggestion
```

materialize-redshift/staged_file.go
```diff
@@ -22,10 +23,126 @@ const (
...
 )
 
-// stagedFile is a wrapper around an s3 upload manager & client for streaming file uploads to s3.
-// The same stagedFile should not be used concurrently across multiple goroutines, but multiple
-// concurrent processes can create their own stagedFile and use them. stagedFile acts as a single
-// sink for writes during a transaction, but will automatically split files into multiple parts.
+// s3Store is the S3 client and bucket shared by every stagedFile of a
+// transactor, and the S3 operations the transactor runs outside of any one
+// stagedFile: writing COPY manifests and deleting objects once they are no
+// longer needed.
```
jacobmarble: Can you rewrite this comment? Feels like it's important, but I can't parse after the comma.

materialize-redshift/staged_file.go
```diff
@@ -22,10 +23,126 @@ const (
...
+}
+
+// putManifest writes a COPY manifest at manifestKey listing the given data
+// files, each a full object key.
+func (s *s3Store) putManifest(ctx context.Context, manifestKey string, files []string) error {
+	manifest := copyManifest{Entries: make([]manifestEntry, 0, len(files))}
+	for _, key := range files {
+		manifest.Entries = append(manifest.Entries, manifestEntry{
+			URL:       s.objectURI(key),
+			Mandatory: true, // Always true
```
jacobmarble:
```suggestion
			Mandatory: true,
```

materialize-redshift/staged_file.go
```diff
@@ -22,10 +23,126 @@ const (
...
+
+type manifestEntry struct {
+	URL string `json:"url"`
+	// The "mandatory" flag means the copy job will fail if the file can't be found. This should
+	// always be set to `true`. See https://docs.aws.amazon.com/redshift/latest/dg/loading-data-files-using-manifest.html
+	Mandatory bool `json:"mandatory"`
+}
+
+// putManifest writes a COPY manifest at manifestKey listing the given data
+// files, each a full object key.
```
jacobmarble: "full object key" made me wonder if this was the s3:// URI, but it isn't. Maybe just say "object key".

materialize-redshift/state.go
```diff
@@ -0,0 +1,208 @@
...
+	ID          string   `json:"id"`
+	StoreFiles  []string `json:"storeFiles"`
+	DeleteFiles []string `json:"deleteFiles"`
+	MustMerge   bool     `json:"mustMerge"`
+	Widen       []string `json:"widen"`
+}
+
+// objects is every S3 object the transaction staged.
+func (e *stateItem) objects() []string {
+	return append(append([]string{}, e.StoreFiles...), e.DeleteFiles...)
```
jacobmarble:
```suggestion
	return slices.Concat(e.StoreFiles, e.DeleteFiles)
```

materialize-redshift/driver.go
```diff
@@ -305,18 +311,31 @@ func NewDriver() *sql.Driver[config, tableConfig] {
...
+	bindings                       []*binding
+	be                             *m.BindingEvents
+	cfg                            config
+	store                          *s3Store
+	caseSensitiveIdentifierEnabled bool
+
+	rangeKey string
+	// The shard whose key range begins at 0 applies every shard's staged
+	// work; the others only stage.
+	primary          bool
```
jacobmarble, line 324: This could also be a function that checks `rangeKey` for `^000000-`

materialize-redshift/driver.go
```diff
@@ -378,22 +430,9 @@ type binding struct {
...
-	cleanupFiles []func(context.Context)
-	// If any deletions must be done for the Store data processed by this binding.
-	hasDeletes bool
-	// If any non-deletion Store requests were received for this binding.
-	hasStores bool
-	// If a merge operation must be done, or if a more efficient COPY INTO can
-	// be used to store the data for the transaction.
-	mustMerge bool
+
+	// The transaction being staged, if any.
```
jacobmarble, line 458:
```suggestion
```

jacobmarble (dismissed):

materialize-redshift/.snapshots/TestIntegration-materialize
```diff
@@ -1,13 +1,6 @@
...
-["connectorState",{}]
```
danielnelson, line 5: It is unfortunate that we needed to remove the `--output-apply --output-state` from the snapshot in order to run with shards=2.

danielnelson (dismissed):

materialize-redshift/driver.go
mdibaiee: I think it is an important comment because normally post-commit apply connectors do not have MetaCheckpoints table, so here we are explaining why this post-commit apply connector has it.

materialize-redshift/driver.go
mdibaiee: it doesn't change throughout a run of a process so we calculate once when instantiating transactor

materialize-redshift/staged_file.go
mdibaiee: the object key doesn't have `s3://` but the object URI would... I intended "full object key" to mean that the key includes the bucket path and prefix, etc. -- will rename

materialize-redshift/staged_file.go
mdibaiee:
```suggestion
// s3Store is the S3 client configured with the user bucket.
```

materialize-redshift/staged_file.go
mdibaiee:
```suggestion
// files, each an object key.
```

jacobmarble (approved):

PR #5158, opened 2026-08-31
source-klaviyo-native: increase eventual consistency lag to 7 days

Alex-Bair:
**Description:**

This PR's scope includes:
- Raising the `events` lookback horizon from 4 days to 7 days. We've observed Klaviyo display eventually consistent behavior beyond 4 days. We've confirmed that 17 days later the missed data was available in the API, but I suspect it was available earlier than 17 days (that was just when the incident was raised and the investigation began) and I'd like to avoid bumping the horizon to 17 days until we have more definitive proof that kind of lag is necessary. 
  - As an aside, I also saw that Klaviyo uses [5 days as a default attribution window for some events](https://help.klaviyo.com/hc/en-us/articles/1260804504250#h_01JC68GA5F1TWGXGE0NCETVXJZ), so bumping the horizon to 7 days will allow us to capture late arriving attributions for events too.
- Fix a race condition bug in the `events` backfill that could cause data loss, even though we haven't observed this race condition cause data loss (yet). The shutdown monitor signaled completion when there were no active workers and both the work queue and subdivision queue were empty. But, it's possible for that to be true briefly when a worker has submitted work to be subdivided; it would mark itself inactive, and both queues would be empty while the subdivision worker was subdividing the work. Now, the shutdown monitor / completion watcher uses the work queue's `.join()` method to make sure all work has been completed before signaling completion. This hasn't been observed to cause data loss, it's mostly a hardening effort that I wanted do while I was looking at the connector.

**Checklist:**

### DIFF
--- a/source-klaviyo-native/CHANGELOG.md
+++ b/source-klaviyo-native/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2026-08-31
+
+### Changed
+- Raised the events lookback horizon from 4 days to 7 days. On upgrade, existing tasks' lookback subtask idles for ~3 days while `now - horizon` catches back up to its cursor; this is expected and no data is skipped.
+
 ## 2026-07-30
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


nicolaslazo (approved):
LGTM

PR #5159, opened 2026-08-31
source-dynamodb: add idleBackoffMax option to slow polling of idle shards

### LINKED ISSUE #5135: source-dynamodb: back off GetRecords polling on shards returning no records
## 1. Priority
Low - one customer, no deadline.

## 2. Scope and prevalence
Affects any `source-dynamodb` capture tailing open stream shards on low-write tables. Cost scales with open shard count times poll rate, independent of data volume.

Reported by one customer, 2026-08-26: 10 bindings, ~44 open shards. They report ~$30/month in DynamoDB StreamRequests charges against ~$6/month for all reads and writes on the same tables combined.

Independently computed from the shard count: 44 shards at 1 request/second is ~114M `GetRecords` calls per 30 days, ~$22 after the 2.5M/month free tier for non-Lambda consumers. The customer's reported figure is consistent with this.

A related effect was observed internally on 2025-03-15, when the same once-per-second loop across many `source-dynamodb` tasks generated enough DNS traffic to destabilize reactor nodes.

## 3. What this enables
As a DynamoDB capture user, I want the connector to stop making billed `GetRecords` calls at a fixed rate against shards that have no new data, so that my DynamoDB Streams cost tracks my write volume rather than my shard count.

1. There is no way to reduce `GetRecords` call volume on an idle capture. The 1/second limit on open shards is a constant in code, and the connector's advanced options expose only `backfillSegments`, `scanLimit`, and `endpoint`.
2. Cost cannot be traded against latency. A user who tolerates minutes of source-side delay pays the same as one who needs sub-second freshness.

Affects users whose tables are low-write relative to their shard count. Users with continuously active shards see no change, because a poll that returns records is not idle.

## 4. Requested behaviour
`GetRecords` polling slows down on a shard whose polls return no records, and returns to the current rate as soon as records are returned.

The customer stated two acceptable shapes: a configurable interval applied only to empty responses, or a capped backoff on consecutive empty polls. Either satisfies the requirement.

Their stated tolerance: they materialize to BigQuery every 30 minutes, so up to 30 seconds of added source-side delay is invisible end to end.

My recommendation: a progressive backoff on consecutive empty polls, capped at 1 minute, rather than a config field. It needs no user input and cannot be misconfigured into silent replication lag.

## 5. Current behaviour
Open shards are polled at a fixed 1 request/second whether or not the previous poll returned records. Closed shards are read at full speed and are not affected.

DynamoDB Streams bills per `GetRecords` request regardless of payload size, and requests made by non-Lambda consumers are charged past the first 2,500,000 per month per region ([AWS: Evaluate your DynamoDB streams usage](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CostOptimization_StreamsUsage.html)). An empty response therefore costs the same as a full one.

Steps to confirm: run a `source-dynamodb` capture against tables with open shards and no writes, then read the stream's `SuccessfulRequestLatency` CloudWatch metric, `SampleCount` statistic, which counts `GetRecords` calls per stream. The count tracks open shard count times elapsed seconds, with no writes occurring.

## 6. References
- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1787795629330659
- AWS ships an equivalent control for its own Streams consumer: the DynamoDB Streams Kinesis Adapter's `idleTimeBetweenReadsInMillis`, the wait before re-polling a shard whose previous `GetRecords` returned no records, default 1000ms. Documented at [Evaluate your DynamoDB streams usage](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CostOptimization_StreamsUsage.html).

Alex-Bair:
**Description:**

DynamoDB Streams bills per `GetRecords` request regardless of payload, and open shards are polled at a fixed 1/s whether or not anything comes back.

The new advanced `idleBackoffMax` option caps a growing wait applied once an open shard has returned nothing for 60 consecutive reads. Each further empty read doubles the wait from 2s up to the cap, any read returning records resets it, and the wait is interruptible by shard-topology restarts. Closed shards are always read at full speed. At the default of 1s no wait is applied and polling is unchanged. At 1m an idle shard costs about 60 requests an hour instead of 3,600.

The grace period exists because an empty read of an open shard does not mean it is caught up: stepping through an empty portion of a shard can take several reads. Holding 1/s for the first minute of empties is intended to keep traversal through empty portions fast.

Closes #5135

Motivating Slack thread: [link](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1787795629330659)

**Notes for reviewers:**

The new `idleBackoffMax` option allows users to trade latency for cost. I'm not 100% sure `'1s'` should be the default max backoff value, but it preserves existing behavior and is in line with most of our technical captures' methodology of "keep latency as low as possible by default", so I'm alright with it for now.

**Checklist:**

### DIFF
--- a/docs/reference/Connectors/capture-connectors/amazon-dynamodb.md
+++ b/docs/reference/Connectors/capture-connectors/amazon-dynamodb.md
@@ -152,6 +152,7 @@ You configure connectors either in the Estuary web app, or by directly editing t
 | **`/credentials`**                   | Credentials             | Credentials for authentication.                                                                               | [Credentials](#credentials) | Required |
 | `advanced/backfillSegments`          | Backfill Table Segments | Number of segments to use for backfill table scans. Has no effect if changed after the backfill has started.  | integer |                  |
 | `advanced/endpoint`                  | AWS Endpoint            | The AWS endpoint URI to connect to. Use if you're capturing from a compatible API that isn't provided by AWS. | string  |                  |
+| `advanced/idleBackoffMax`            | Idle Poll Backoff Maximum | Maximum wait between polls of an open stream shard whose reads keep returning no records. Polling returns to once per second as soon as records appear. Accepts Go duration strings between `1s` and `10m`. | string  | `1s`             |
 | `advanced/scanLimit`                 | Scan Limit              | Limit the number of items to evaluate for each table backfill scan request.                                   | integer |                  |
 
 #### Credentials
@@ -224,6 +225,10 @@ If your DynamoDB tables don't appear when configuring the capture:
 - Verify DynamoDB Streams is enabled on the table
 - Verify the stream view type is set to "New and old images"
 - Verify your IAM policy includes `ListTables` permission on `table/*`
+
+### High DynamoDB Streams request charges on low-traffic tables
+
+DynamoDB bills every `GetRecords` request the capture makes, even when no records are returned, and open stream shards are polled once per second by default. On tables with many shards and few writes, these polls can make up most of the stream's cost. Set `advanced/idleBackoffMax` (for example, `1m`) to let polling slow down on shards that keep returning no records. Polling returns to once per second as soon as records appear, so new data is delayed by at most the configured value.
 
 ### "dynamodb:ListTables... AccessDeniedException"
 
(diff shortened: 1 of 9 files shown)


nicolaslazo (approved):
LGTM!

PR #5160, opened 2026-08-31
materialize-snowflake: Snowpipe Streaming high-performance

### LINKED ISSUE #4324: Snowflake Snowpipe Streaming: Switch to SDK

jacobmarble:
**Description:**

Closes #4324

Adds [Snowpipe Streaming "high performance"](https://docs.snowflake.com/en/user-guide/snowpipe-streaming/data-load-snowpipe-streaming-overview) to `materialize-snowflake`.

## V2 Runtime required

The mode of this path is the new "Idempotent Shards with Document Counters". This feature depends on the idempotent, deterministic document order guaranteed by the V2 Runtime.

Rows are appended to the channel during `Store()` - each row is transmitted to Snowflake exactly once, and nothing is written to the internal stage. On recovery, only rows beyond a channel's committed offset token are re-transmitted.

## SDK

There is no Go SDK, so the official Python SDK runs in a supervised sidecar process, spoken to over NDJSON RPC on a unix domain socket.

## Snowpipe Channels

The Snowpipe service receives data via "channels", a resource tied to exactly one Snowflake table. This feature maintains four Snowflake channels per Estuary shard, each channel assigned to keyspace aligning with potential future shards, eg the shard that handles keys `00~ff` assigns documents to four Snowpipe channels, each handling keys `00~3f`, `40~7f`, `80~bf`, `c0~ff`.

We do this for two reasons:

### Throughput limits

Snowflake strictly limits the throughput of each channel to 20MB (20,000,000 bytes, not 20MiB) per second. By parallelizing the work over multiple channels, a single shard can deliver much higher throughput.

### Recovery after shard count change

A connector instance assigned `00~ff` exits gracelessly at the moment its shard is split. Child instances check the Estuary and Snowpipe state of their ranges, `00~7f` and `80~ff`. Each inherits two channels formerly handled by the parent and, if the split interrupted a transaction, skips over completed work to close the transactions in the parent channels, then closes the parent channels, and opens four new channels from a quiescent state.

Offset tokens are a monotonic per-channel document counter. The driver checkpoint holds per channel (four per shard per binding), the channel name, the counter, and the channel's key range. Where `C` is the Snowflake counter, and `K` is the Estuary counter, recovery logic follows this table:

| condition | meaning | action |
| --- | --- | --- |
| `C == K` | quiescent state, Snowflake and Estuary agree | proceed normally or change channel topology |
| `K == 0` and `C` absent | no rows have been written yet | proceed as though `C == K` |
| `C > K` | interrupted transaction | skip work through offset `C` and continue the transaction |
| `C < K`, or `C` absent and `K > 0` | invalid state; Snowflake is *behind* Estuary, or the channel is gone | refuse loudly; backfill is required to recover |
| `C > 0` and `K` absent, checkpoint holds other items for the binding | invalid state; an item is recorded for every channel before anything is appended to it, so this token was written by something else | refuse loudly; backfill is required to recover |
| `C > 0` and `K` absent, checkpoint holds nothing for the binding | the channel's interrupted first transaction | skip work through offset `C` and continue the transaction |

### Synthetic benchmark

Benchmarked against the classic (BDEC) streaming path in a matched configuration: one VM per arm, delta updates, single binding, 5 GiB transactions. Single-shard arms moved 50 GiB, four-shard arms 200 GiB.

| Arm | Machine | MB/s | MiB/s | Docs/s | vs BDEC | vs single shard |
|---|---|---|---|---|---|---|
| BDEC 1KB | c3-highmem-8 | 29.31 | 27.95 | 28,622 | | |
| v2 1 shard 1KB | c3-highmem-8 | 57.70 | 55.02 | 56,344 | 1.97x | |
| v2 4 shard 1KB | c3-highmem-8 | 69.36 | 66.15 | 67,734 | 2.37x | 1.20x |
| v2 4 shard 1KB | c3-highmem-22 | 83.69 | 79.81 | 81,730 | 2.86x | 1.45x |
| BDEC 1MB | c3-highmem-8 | 51.69 | 49.30 | 49 | | |
| v2 1 shard 1MB | c3-highmem-8 | 60.63 | 57.82 | 58 | 1.17x | |
| v2 4 shard 1MB | c3-highmem-8 | 153.38 | 146.27 | 146 | 2.97x | 2.53x |
| v2 4 shard 1MB | c3-highmem-22 | 178.14 | 169.89 | 170 | 3.45x | 2.94x |

Both ratio columns compare within the same document size, against the `c3-highmem-8` rows.

### Production backfill

Both paths run as production tasks over one live collection, no backfill bound, same
connector build, each configured the way its path ships.

| Arm | Docs | Bytes | Avg doc | Duration | Docs/s | MB/s |
|---|---|---|---|---|---|---|
| BDEC, 1 shard, runtime v1 | 37,266,700 | 21.00 GiB | 605 B | 45.7 min | 13,601 | 8.23 |
| v2, 2 shards, runtime v2 | 37,613,116 | 21.23 GiB | 606 B | **18.3 min** | **34,256** | **20.76** |

Inputs are near-identical, so the 2.50x duration ratio is the comparison. Rates are
end-to-end, charging every commit and stall against them.

Peak rates do not compare across paths: BDEC sat inside a transaction for 97% of wall
clock and held a flat 13,669-14,473 docs/s throughout, while v2 appended for only 25%,
reaching 130,773 docs/s over one 125s transaction and awaiting commit for the rest.

Shard counts differ because multi-shard BDEC is unsupported. Runtime v2 reports task-wide
stats under shard `00000000`.

**Workflow steps:**

Enable the `snowpipe_streaming_v2` feature flag in the endpoint's advanced configuration. The endpoint must use key-pair (JWT) authentication, and the task must be on Estuary runtime v2 (`enable-runtime-v2` shard flag).

**Notes for reviewers:**

**Checklist:**


materialize-snowflake/sidecar/snowpipe_sidecar/__main__.py
```diff
@@ -0,0 +1,56 @@
...
+    signal.signal(signal.SIGTERM, lambda *_: os._exit(0))
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
+
+    # The auth token arrives on stdin so it appears in no argv or environ.
+    auth_token = sys.stdin.readline().strip()
+
+    if args.uds:
+        server = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+        server.bind(args.uds)
+        ready = {"ready": True, "transport": "uds"}
```
mdibaiee: Do we actually use UDS ever, or only TCP? for the sake of simpliity / minimising diff I would propose we keep only one that we use

materialize-snowflake/sidecar/snowpipe_sidecar/channels.py
```diff
@@ -0,0 +1,224 @@
...
+# decode below scans the payload to find each row's extent without building it.
+# tests/test_sdk_raw_rows.py is the guard on that route through the SDK.
+_ROWS_DECODER = msgspec.json.Decoder(List[Raw])
+
+# A row must be a JSON object: the decode validates each row's syntax, but the
+# SDK would forward anything else — a number, a string, an array — to Snowflake
+# unchecked, and Snowflake counts what it cannot store as a rejected row rather
+# than failing the append. Reading the first byte is enough to tell: a decoded
+# Raw begins at its value, the whitespace between array elements excluded.
+_OPEN_BRACE = 0x7B
```
mdibaiee: This comment reads hard... I couldn't fully grok it... I _think_ it means this:

```suggestion
# A row must be a JSON object, but the SDK forwards everything
# to Snowflake unchecked, and Snowflake counts invalid rows
#  as rejected row rather than failing the append. We check the
# first byte of each row to be equal to an opening curly brace {
_OPEN_BRACE = 0x7B
```

materialize-snowflake/sidecar/snowpipe_sidecar/channels.py
```diff
@@ -0,0 +1,224 @@
...
+
+# RECEIVER_SATURATED is the SDK's flow-control signal: its internal buffer is
+# full and the caller should back off, not fail. The total wait stays under the
```
mdibaiee: to keep the vocabulary consistent with code
```suggestion
# full and the caller should back off, not fail. The max_wait stays under the
```

materialize-snowflake/sidecar/snowpipe_sidecar/channels.py
```diff
@@ -0,0 +1,224 @@
...
+    except msgspec.DecodeError as err:
+        raise SidecarError("invalid_rows", f"append payload is not a JSON array of rows: {err}") from err
+
+    if len(rows) != row_count:
+        raise SidecarError(
+            "invalid_rows",
+            f"append declared {row_count} row(s) but its payload holds {len(rows)}",
+        )
+    for i, row in enumerate(rows):
+        if memoryview(row)[0] != _OPEN_BRACE:
```
mdibaiee, line 83: TIL `memoryview`

> A memoryview is a Python object that lets you access an object’s underlying buffer (like bytes or bytearray) without copying the data.

materialize-snowflake/sidecar/snowpipe_sidecar/ops_logging.py
```diff
@@ -0,0 +1,55 @@
+"""Structured logging in the Flow ops-log shape, one JSON object per stderr line.
+
+Mirrors estuary-cdk's logger format ({"level", "msg", "fields": {...}}) so the
```
mdibaiee: I don't know if this is even worth it given that this file is small, but can we reuse estuary-cdk's logger implementation? just so that we always stay in sync with their format in case the format is updated in the future, we won't have to update two places

materialize-snowflake/CHANGELOG.md
```diff
@@ -1,5 +1,81 @@
...
+- Each binding on this path writes through four channels per shard, one per
+  equal subrange of the shard's key-hash range, with each document routed by
+  the same key hash the runtime routes documents to shards by. Snowflake meters
+  throughput per channel, so a shard's ceiling is 80 MB/s. Because a channel's
+  contents depend only on the data, a shard split or join hands each surviving
+  shard whole channels, and the shard continues them where their committed
+  offset tokens left off — including the rows of a transaction the topology
+  change interrupted — before converging back to four channels of its own
+  range at the next settled transaction boundary.
+- Rows on this path become visible in the destination table slightly before the
```
mdibaiee: This seems like a "change" not an "addition", I think the changelog can be made shorter and may be grouped differently.

materialize-snowflake/client.go
```diff
@@ -188,13 +195,38 @@ func (c *client) PopulateInfoSchema(ctx context.Context, is *boilerplate.InfoSch
 var errInsufficientPrivileges = regexp.MustCompile(`Insufficient privileges to operate on schema '([^']+)'`)
 
 func (c *client) CreateTable(ctx context.Context, tc sql.TableCreate) error {
+	// The generation this table belongs to is recorded in its comment. A missing
+	// state key parses as no generation at all — which streamV2GenerationConflict
+	// reads as permission for any generation to append, leaving the backfill race
+	// it guards unchecked. Refused ahead of the CREATE, so such a binding leaves
+	// no table, instead of a table whose generation cannot be parsed from the
+	// table comment.
```
mdibaiee: I don't quite grok this... what is the backfill race? "it guards unchecked" seems one of those claudey grammars

materialize-snowflake/client.go
```diff
@@ -353,7 +385,82 @@ func (c *client) InstallFence(ctx context.Context, checkpoints sql.Table, fence
...
+	// generation may not then do is reopen one, which reconcileStreamV2Channel
+	// refuses.
+
+	// Decide whether the generation being replaced could hold channels on the
+	// table: a delta-updates binding, on an endpoint configured for the
+	// streaming path.
+	var outgoingDeltaUpdates = lastBinding == nil || lastBinding.DeltaUpdates
+	if !outgoingDeltaUpdates {
+		return false, nil
+	}
```
mdibaiee: The comment and the naming of the variable is confusing... if I understand correctly we are saying that if a binding is not delta updates, we do not need to recreate? If that's it, I think I would find this more readable, and it doesn't need a comment either:


```suggestion
	if lastBinding != nil && !lastBinding.DeltaUpdates {
		return false, nil
	}
```

materialize-snowflake/client.go
```diff
@@ -353,7 +385,82 @@ func (c *client) InstallFence(ctx context.Context, checkpoints sql.Table, fence
...
+
+	// The endpoint configuration is read from the one in hand, which is the
+	// configuration that generation was running unless this same publication also
+	// changes the write path.
+	if streamsV2(&c.cfg, true, c.streamingV2Enabled) {
+		return true, nil
+	} else if req.LastMaterialization == nil {
+		return false, nil
+	}
```
mdibaiee: the comment here is also quite ambiguous to me, what do we mean "the one in hand"? I guess it means the configuration that the task is currently running with?

this "generation" term just has me confused a lot... what do we mean by a generation?

materialize-snowflake/client.go
```diff
@@ -353,7 +385,82 @@ func (c *client) InstallFence(ctx context.Context, checkpoints sql.Table, fence
...
+	// A publication may turn the write path off in the same breath as it bumps a
+	// backfill counter, and the generation it replaces is streaming either way. So
+	// the replaced specification's own configuration is consulted too — but only
+	// ever to widen this answer, never to narrow it, because a request carries that
+	// configuration in whatever shape the control plane stored it rather than the
+	// shape this connector is handed for itself.
+	outgoingCfg, ok := endpointConfigOf(req.LastMaterialization.ConfigJson)
+	if !ok {
+		return false, fmt.Errorf(
+			"cannot read the endpoint configuration of the specification being replaced, so whether its generation streams into the table cannot be established: backfilling this binding requires knowing whether the table must be re-created rather than emptied",
```
mdibaiee: How can a user resolve this error once they hit it? Is it resolvable?

mdibaiee (commented):
One round of comments

materialize-snowflake/sidecar/snowpipe_sidecar/__main__.py
jacobmarble: I'll drop it.

materialize-snowflake/sidecar/snowpipe_sidecar/ops_logging.py
jacobmarble: You're probably right. It adds bloat to this connector's Python venv, but your point is valid.

materialize-snowflake/snowflake.go
```diff
@@ -180,10 +180,12 @@ func NewDriver() *sql.Driver[config, tableConfig] {
...
 			return &sql.Endpoint[config]{
-				Config:              cfg,
-				Dialect:             dialect,
-				SerPolicy:           serPolicy,
-				NewClient:           newClient,
+				Config:    cfg,
+				Dialect:   dialect,
+				SerPolicy: serPolicy,
+				NewClient: func(ctx context.Context, materializationName string, ep *sql.Endpoint[config]) (sql.Client, error) {
+					return newClient(ctx, materializationName, ep, featureFlags[flagSnowpipeStreamingV2])
```
mdibaiee: Other clients usually infer the feature flags again in `newClient` with something like:

```
flags := common.ParseFeatureFlags(ep.cfg.Advanced.FeatureFlags, featureFlagDefaults)
```

I don't have a strong opinion either way, but just noting there is prior code that works differently

materialize-snowflake/snowflake.go
```diff
@@ -277,6 +281,15 @@ func (d *transactor) UnmarshalState(state json.RawMessage) error {
 	return nil
 }
 
+// priorStreamV2 returns the streaming v2 state the checkpoint records for a
+// binding, or nil if it records none.
+func (d *transactor) priorStreamV2(stateKey string) map[string]*streamV2Item {
```
mdibaiee, line 284: what does "prior" mean here? it sounds to me like this is something legacy / old?

materialize-snowflake/snowflake.go
```diff
@@ -386,16 +419,60 @@ type binding struct {
...
+	}
+
 	var b = new(binding)
 	b.target = target
 	b.nullFieldsToStrip = target.NullableFieldsToStrip()
 	b.load.mergeBounds = sql.NewMergeBoundsBuilder(target.Keys, d.ep.Dialect.Literal)
 	b.store.mergeBounds = sql.NewMergeBoundsBuilder(target.Keys, d.ep.Dialect.Literal)
 
+	if streamingV2 {
+		// Work that the previous write path staged and did not finish is drained by
```
mdibaiee: I'm a bit confused about what "previous write path" means here, is it the pre-runtime-v2-snowpipe-streaming-SDK write path? or something else?

in the interest of having the comment be evergreen as well, we might want to be explicit

materialize-snowflake/client.go
jacobmarble: The logic is nuanced here.

On backfill, the connector has to decide whether to `TRUNCATE` or `DROP`+`CREATE`. The condition you have highlighted can be correctly resolved quietly by simply deciding to `DROP`+`CREATE`.

mdibaiee (commented):
I am a bit confused by the namings of things, there is d.streamV2, and there is d.streamManager, and there are terms like "retire" and "route" and "gate" that seem to be novel? I couldn't find them in the README either... are there analogous concepts we already have so that we could use the same name for these, or are they genuinely novel? if so let's clarify what they mean please

materialize-snowflake/client.go
jacobmarble: I've improved comments that mention "generation" or removed the term where possible.

materialize-snowflake/snowflake.go
jacobmarble: Good one, let's use the existing tools. 👍

materialize-snowflake/client.go
```diff
@@ -188,13 +197,40 @@ func (c *client) PopulateInfoSchema(ctx context.Context, is *boilerplate.InfoSch
...
+	// We record the task name and the binding's state key (streamV2RecordedStateKey)
+	// in the Snowflake table comment. A binding without a state key would leave
+	// that record empty, and an empty record lets any task or any state key
+	// append to the table. The check therefore runs before the CREATE, so the
+	// error below creates no table at all.
+	var recordStateKey = streamsV2(&c.cfg, tc.Table.DeltaUpdates, c.streamingV2Enabled)
+	if recordStateKey && tc.StateKey == "" {
+		return fmt.Errorf("cannot record the state key of table %s: the binding resolved for it carries no state key", tc.Identifier)
+	}
+
```
mdibaiee: I think it is not possible for a binding to have no statekey. It is a given by the runtime that each binding must carry a stateKey. Is there a scenario where it can be missing?

materialize-snowflake/client.go
```diff
@@ -353,7 +389,71 @@ func (c *client) InstallFence(ctx context.Context, checkpoints sql.Table, fence
 }
 
 func (c *client) MustRecreateResource(req *pm.Request_Apply, lastBinding, newBinding *pf.MaterializationSpec_Binding) (bool, error) {
-	return false, nil
+	// A backfill of a streaming v2 binding must drop and re-create the table
+	// rather than truncate it. Apply runs while the shards of the old spec are
+	// still running, and their Snowpipe channels are bound to this table. A
```
mdibaiee: If I recall correctly, Apply does not run in parallel to a running shard.

The snowpipe channels bound to the table will still be there, but no materialization shard will be running

materialize-snowflake/client.go
```diff
@@ -353,7 +389,71 @@ func (c *client) InstallFence(ctx context.Context, checkpoints sql.Table, fence
...
+	// truncate leaves those channels open, so rows they append after it survive
+	// and are then materialized a second time by the backfill. Dropping the
+	// table closes every channel on it, including channels no checkpoint names
+	// because their shard committed nothing yet.
```
mdibaiee: If channel naming is deterministic, could we instead clean up channels of a binding rather than having to re-create the table?

There are reasons for this and the main one is that sometimes users bring their own table (it is a supported use case) and sometimes users want to make modifications on top of the tables or build dependent objects on their tables (e.g. a view) -- this is the reason we do truncations instead of drop-and-recreate.

materialize-snowflake/client.go
mdibaiee: I checked and the runtime does indeed wait for shards to finish their work, and once no shards are running, it runs Apply RPC by the leader, and waits for Apply to finish before starting shards again.

This changes the code, and I believe persisting binding stateKey, etc. in the table comment becomes unnecessary, and perhaps other changes too.

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+// configuration names snowpipe_streaming itself and does not name
+// snowpipe_streaming_v2. The default value of snowpipe_streaming does not count,
+// because leaving the path duplicates documents and must be asked for.
+func streamV2Downgrade(cfg *config, deltaUpdates bool) bool {
+	if cfg.Credentials == nil || cfg.Credentials.AuthType != snowflake_auth.JWT || !deltaUpdates {
+		return false
+	}
+	var configured = common.ParseFeatureFlags(cfg.Advanced.FeatureFlags, nil)
+	return configured[flagSnowpipeStreaming] && !configured[flagSnowpipeStreamingV2]
+}
```
mdibaiee, line 260: I think this may need a better name because it doesn't really tell us that v2 is "leaving", it just tells us v1 is explicitly enabled and v2 is not, right?

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+// bdec manager for that drain, so this is reached only where the registration is
+// impossible: a table whose columns that path does not support, or one it may not
+// stream into at all.
+//
+// The blobs are the only record of documents which Snowflake does not hold yet, so
+// they are neither dropped nor left to fail every transaction that tries to drain
+// them.
+func streamV2CannotDrainPendingBlobs(table string, blobs int, cause error) error {
+	var err = fmt.Errorf(
+		"this binding is moving onto the snowpipe_streaming_v2 write path while the task's checkpoint still records %d Snowpipe Streaming blob(s) that the snowpipe_streaming write path staged into %s and did not finish. Those blobs are the only record of documents which Snowflake does not hold yet, and only the snowpipe_streaming path can finish them. Restore that path for one transaction, which finishes them, before you move the binding onto snowpipe_streaming_v2, or backfill the binding, which discards them and materializes those documents again",
```
mdibaiee:
```suggestion
		"this binding is moving onto the snowpipe_streaming_v2 write path while the task's checkpoint still records %d Snowpipe Streaming blob(s) that the snowpipe_streaming write path staged into %s and did not finish and only the snowpipe_streaming path can finish them. Restore that path for one transaction before you move the binding onto snowpipe_streaming_v2, or backfill the binding, which discards them and materializes those documents again",
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+// A shard holds several channels per binding. In steady state they are the shard's
+// target layout — its key range cut into streamV2ChannelsPerShard equal key ranges —
+// and after a topology change they are the channels inherited from the shards this
+// one replaced, until a rebalance converges them.
+type streamV2Channel struct {
+	name string
+	// r is the key range of the key-hash space this channel covers. Documents route
+	// here when their packed-key hash falls inside it. The key range is embedded in
+	// the channel's name and in every offset token it appends under.
+	r streamV2Range
```
mdibaiee: for human readability
```suggestion
	range streamV2Range
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	)
+}
+
+// streamV2Channel is one channel of a binding: the key range of the key-hash space it
+// covers, the document counter it advances, and the batch it is building.
+//
+// A shard holds several channels per binding. In steady state they are the shard's
+// target layout — its key range cut into streamV2ChannelsPerShard equal key ranges —
+// and after a topology change they are the channels inherited from the shards this
+// one replaced, until a rebalance converges them.
```
mdibaiee:
```suggestion
// streamV2Channel is one channel of a binding: the key range it
// covers, the document counter it advances, and the batch it is building.
```

The comment about the sharding holding multiple channels could belong to the shard if it is deemed necessary:

```
// A shard holds several channels per binding. In steady state they are the shard's
// target layout and after a topology change they are the channels inherited
// from the shards this one replaced, until a rebalance converges them.
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+type streamV2Channel struct {
+	name string
+	// r is the key range of the key-hash space this channel covers. Documents route
+	// here when their packed-key hash falls inside it. The key range is embedded in
+	// the channel's name and in every offset token it appends under.
```
mdibaiee:
```suggestion
	// range is the key range this channel covers. The key range is embedded in
	// the channel's name and in every offset token it appends to.
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	// routed here.
+	counter int64
+	// recorded is the counter value that the last checkpoint item reported. It is how
+	// flush finds a channel that stored nothing this transaction.
+	recorded int64
+	// skip is the document index that Snowflake already held when this session opened
+	// the channel. An interrupted attempt of the transaction now replayed appended the
+	// documents at or below that index. The manager counts those documents again, but
+	// does not append them again. Document indices are absolute, so this one threshold
+	// holds for the whole session.
```
mdibaiee, line 369:
```suggestion
	// the channel. The manager counts those documents again, but
	// does not append them again. Document indices are absolute, so this one threshold
	// holds for the whole session.
```

mdibaiee: I am a bit unclear about the use of the word "session" here, what is meant by a session?

I think it doesn't mean "a run of the materialization process"
I also think it doesn't mean "one cycle of Load + Store + Ack" (one transaction)

Document indices are held for the lifetime of a binding (a backfilled binding is then considered new)... so the word session is confusing to me

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	skip int64
+	// committed is the highest index that Snowflake durably holds. It starts at the
+	// committed offset token when this session opened the channel, and then takes each
+	// counter this session committed. waitCommit must not wait on a counter at or below
+	// it. The wait is for the committed token to *equal* the requested token, and a
+	// channel already past that token never reports it again. Such a wait runs to its
+	// timeout. A replay of documents that Snowflake already holds produces exactly such
+	// a counter.
```
mdibaiee, line 377: This is hard to understand for me, could you update it please?

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	committed int64
+
+	// buf is the payload of the batch under construction: the framed JSON array that
+	// the append carries. bufferRow writes it row by row as Store stores each document.
+	// This way bufferRow assembles the bytes of a batch once, and not once per row and
+	// again per batch. bufRows counts the rows it holds, and bufFirst is the index of
+	// the first row.
```
mdibaiee:
```suggestion
	// buf is the payload of the batch under construction
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	buf      []byte
+	bufRows  int
+	bufFirst int64
+	// bufHint is the size of the batch handed off before this one, and it sizes the next
+	// buffer. The manager cannot reuse the buffer itself. Ownership passes to the
+	// append. That append still writes the buffer to the sidecar while bufferRow builds
+	// the next batch.
```
mdibaiee:
```suggestion
	// bufHint is the size of the batch handed off before this one, and it sizes the next
	// buffer.
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	bufHint int
+
+	// pipe carries one append at a time. Store can buffer more rows while a batch is on
+	// the wire, and the batches keep their order.
+	pipe appendPipe
+
+	// limiter meters the uncompressed bytes this channel hands to Snowflake. One limiter
+	// per channel, because Snowflake meters each channel independently. This is why
+	// several channels per shard raise the shard's ceiling: four channels are four
+	// independently metered pipes, and the shard may legitimately drive all of them.
```
mdibaiee:
```suggestion
	// limiter meters the uncompressed bytes this channel hands to Snowflake.
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+func (c *streamV2Channel) offsetToken(counter int64) string {
+	return streamV2Token(counter, c.r)
+}
+
+type streamV2Binding struct {
+	database string
+	schema   string
+	table    string
+	// stateKey is the state key of the binding that this shard materializes. The
+	// table it appends to must record the same state key.
```
mdibaiee:
```suggestion
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	stateKey string
+	// columns holds the row-object prefix of each Snowflake column, in the same order
+	// as the output of Table.ConvertAll. The append writes row objects by name, and the
+	// SDK ingests them.
```
mdibaiee, line 409: This is probably Snowflake terminology, but what is a "row-object prefix" of a Snowflake column?

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	columns []rowColumn
+	// prior is the streaming v2 state that the driver checkpoint recorded for this
+	// binding. The map key is the channel's key range, and the map covers every shard
+	// of the task. It is nil when the checkpoint recorded no state.
+	prior map[string]*streamV2Item
+	// opened reports whether this session classified the checkpoint's channels,
+	// opened its own, and reconciled each of them.
```
mdibaiee, line 416: What does "classified" mean here?

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	opened bool
+	// dropped holds the item keys of the channels this shard has dropped in
+	// Snowflake. The next checkpoint deletes their items. Every checkpoint of the
+	// session reports them again, not only the first, so that a transaction that
+	// never commits does not lose the deletion.
+	dropped []string
+
+	// channels is the active layout: the channels rows route to, sorted by key range
+	// begin. The layout always covers the shard's key range with no gaps or
+	// overlaps, so every document the runtime delivers routes to exactly one of them.
```
mdibaiee:
```suggestion
	// channels is the active layout: the channels rows route to, sorted by
	// keyRangeBegin. The layout always covers the shard's key range with no gaps or
	// overlaps, so every document the runtime delivers routes to exactly one of them.
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	channels []*streamV2Channel
+	// targets is the shard's target layout: its key range cut into
+	// streamV2ChannelsPerShard equal key ranges. The active layout converges to it.
+	targets []streamV2Range
+	// declared reports that the last flush wrote the target layout's items into the
+	// checkpoint. Acknowledge runs after the runtime made that checkpoint durable,
+	// which is what makes the switch it performs crash-safe: a shard that dies
+	// after the switch's drops restarts to a checkpoint that names the channels it
+	// was converging to.
```
mdibaiee:
```suggestion
	// declared reports that the last flush wrote the target layout's items into the
	// checkpoint. Acknowledge runs after the runtime made that checkpoint durable,
	// which is what makes the switch it performs crash-safe.
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1524 @@
...
+	var err = c.pipe.submit(func() error {
+		if err := c.limiter.WaitN(ctx, len(payload)); err != nil {
+			return fmt.Errorf("pacing channel %q: %w", c.name, err)
+		}
+		client, err := m.ensureStarted(ctx)
+		if err != nil {
+			return err
+		}
+		return client.Append(ctx, c.name, c.offsetToken(first), c.offsetToken(last), payload, rows)
+	})
```
mdibaiee, line 1394: I think we can refactor away `c.pipe.submit`, it doesn't really do much to deserve a separate structure, it is essentially just an `errgroup`:

```
g, gctx := errgroup.WithContext(ctx)
g.SetLimit(1)

g.Go(func() error {
        if err := c.limiter.WaitN(ctx, len(payload)); err != nil {
			return fmt.Errorf("pacing channel %q: %w", c.name, err)
		}
		client, err := m.ensureStarted(ctx)
		if err != nil {
			return err
		}
		return client.Append(ctx, c.name, c.offsetToken(first), c.offsetToken(last), payload, rows)
})

// later, instead of c.pipe.Wait:
if err := g.Wait(); err != nil { ... }
```

materialize-snowflake/client.go
```diff
@@ -352,6 +359,10 @@ func (c *client) InstallFence(ctx context.Context, checkpoints sql.Table, fence
 	return sql.Fence{}, nil
 }
 
+// MustRecreateResource permits every backfill of a Snowflake table to truncate it.
+// A streaming v2 channel is bound to its own name and epoch, not to the table, and
+// the runtime this write path requires holds every shard of the task idle for the
+// whole of Apply, so no shard appends into a table a backfill is truncating.
```
mdibaiee:
```suggestion
```

mdibaiee (commented):
Some more comments, continuing...

materialize-snowflake/streamv2.go
```diff
@@ -666,33 +714,50 @@ func (m *streamV2Manager) ensureOpened(ctx context.Context, b *streamV2Binding)
...
+	// collide with nothing a rebalance abandoned and a sweep has yet to drop.
+	// Channels outside this shard's range are a sibling's, and their names, at
+	// whatever epoch, never collide with this shard's, so a sibling's convergence
+	// does not push this shard's epoch.
+	//
+	// The epoch is read from the checkpoint alone, never from the channels standing
+	// on the pipe: an interrupted first transaction leaves a committed channel with
+	// no checkpoint item, and that channel must be reopened and its committed
+	// documents skipped, not stepped over by a fresh epoch that would materialize
+	// them twice.
```
mdibaiee, line 690: This is a mouthful and gets confusing quickly with the Claude grammar

materialize-snowflake/snowflake.go
```diff
@@ -879,6 +1021,19 @@ func (d *transactor) Acknowledge(ctx context.Context, statePatches []json.RawMes
...
+		if len(item.StreamV2) > 0 {
+			// There is nothing to apply for a streaming v2 binding: its rows were
+			// appended and committed to Snowflake before this transaction's
+			// checkpoint was produced, since a commit awaited here could no longer
+			// fail the transaction whose counter it belongs to.
+			//
+			// The counter is durable per-binding state rather than pending work —
+			// it must outlive this transaction to be reconciled against Snowflake's
+			// committed offset token at the next Open — so this state key is
+			// deliberately not drained.
```
mdibaiee, line 1013:
```suggestion
			// There is nothing to apply for a streaming v2 binding: its rows were
			// appended and committed to Snowflake during Store, before this
			// transaction's checkpoint was produced
```

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1719 @@
...
+		return "", false
+	}
+	return groups[1], true
+}
+
+// streamV2ParseBackfilledChannel reads the key range of a channel this
+// materialization derived for a state key other than the one given, and reports
+// whether the name is such a channel. A backfill rotates a binding's state key and
+// never restores one, so a channel this task named under any other state key belongs
+// to a binding that no longer exists.
```
mdibaiee, line 160: I feel like what this function does is probably simpler than this comment tries to explain... I tried a few times but couldn't grok it 🤔

materialize-snowflake/streamv2.go
jacobmarble: `range` is a Golang keyword so I'll use `keyRange`.

materialize-snowflake/streamv2.go
jacobmarble: The term "session" here does refer to a "connector session". Correcting this prose has led me to drop the `skip` field altogether because its presence contributes to this confusion. `skip` is set once on initialization, and only needed when the first transaction is handled.

Instead, we can depend only on `committed` which is initialized to the same value (the document offset according to Snowflake).

Some of this terminology is difficult for me to track because two systems are tracking the same concepts in their own ways (channels, offsets, documents, indexes).

materialize-snowflake/streamv2.go
jacobmarble: No, this just Claude. Similarly we have `type rowColumn struct`.

materialize-snowflake/streamv2.go
jacobmarble: I'm still thinking on this one. The errgroup would have to be a field on the `streamV2Manager` struct, so not quite as simple as you laid out, but an improvement for sure.

materialize-snowflake/config.go
```diff
@@ -229,5 +241,57 @@ func (c config) Validate() error {
...
+// v2 write path by explicit choice. Only the flags the configuration sets are
+// consulted, because flagSnowpipeStreaming is enabled by default and leaving the
+// path duplicates documents, so it must be asked for.
+func (c config) isStreamsDowngradeV2ToV1(deltaUpdates bool) bool {
+	if c.Credentials == nil || c.Credentials.AuthType != snowflake_auth.JWT || !deltaUpdates {
+		return false
+	}
+	var configured = common.ParseFeatureFlags(c.Advanced.FeatureFlags, nil)
+	return configured[flagSnowpipeStreaming] && !configured[flagSnowpipeStreamingV2]
+}
```
mdibaiee: Thinking about this and https://github.com/estuary/connectors/pull/4646

With https://github.com/estuary/connectors/pull/4646 we will set every feature flag explicitly to its default value after the first run of the connector... with that I think it won't work as-is?

materialize-snowflake/snowflake.go
```diff
@@ -388,15 +417,67 @@ type binding struct {
...
+		// drain, and its rows still go to streaming v2.
+		if prior := d.cp[target.StateKey]; prior != nil && len(prior.StreamBlobs) > 0 {
+			var loc = d.ep.Dialect.TableLocator(target.Path)
+			const remedy = "Restore the snowpipe_streaming write path for one transaction before moving the binding onto snowpipe_streaming_v2, or backfill the binding, which discards those blobs and materializes their documents again"
+			if d.snowpipeStreaming == nil {
+				return fmt.Errorf(
+					"the task's checkpoint records %d Snowpipe Streaming blob(s) staged into %s that only the snowpipe_streaming write path can finish, and that path is not available. %s",
+					len(prior.StreamBlobs), target.Identifier, remedy,
+				)
+			} else if err := d.snowpipeStreaming.addBinding(ctx, loc.TableSchema, d.ep.Identifier(loc.TableName), target); err != nil {
```
mdibaiee: Are both of these condiitons really possible?

```
streamingV2 = deltaUpdates && c.Credentials != nil && c.Credentials.AuthType == snowflake_auth.JWT &&
		boilerplate.ParseFlags(c)[flagSnowpipeStreamingV2]
```

We also have:

```
if cfg.Credentials.AuthType == snowflake_auth.JWT {
		var accountName string
		if err := db.QueryRowContext(ctx, "SELECT CURRENT_ACCOUNT()").Scan(&accountName); err != nil {
			return nil, fmt.Errorf("fetching current account name: %w", err)
		} else if sm, err = newStreamManager(&cfg, open.Materialization.TaskName(), accountName, open.Range.KeyBegin); err != nil {
			return nil, fmt.Errorf("newStreamManager: %w", err)
		}
```

So it seems to be that if `streamingV2` is true, which implies `cfg.Credentials.AuthType == snowflake_auth.JWT`, we must have `snowpipeStreaming != nil`

materialize-snowflake/config.go
mdibaiee: Conceptually as well this function does not really tell us anything about a downgrade, but rather that there is an explicit `snowpipe_streaming` feature flag set, without a `snowpipe_streaming_v2` feature flag set... which I think actually already exists for some of the older tasks which were eventually switched to `snowpipe_streaming` and so they have this explicit config

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1346 @@
...
+	var channelNames = cp.channelNames()
+	if len(channelNames) == 0 {
+		return nil
+	}
+
+	return fmt.Errorf(
+		"this binding materialized into %s through the snowpipe_streaming_v2 write path, which this task's specification no longer selects for it, and the task's checkpoint still records its channel(s) %s. Leaving this path discards that record, and returning later would skip the documents those channels already hold. Restore the snowpipe_streaming_v2 write path — it needs the feature flag, delta updates, and key-pair authentication — or backfill this binding",
+		table, strings.Join(channelNames, ", "),
+	)
+}
```
mdibaiee: The name of the function made me expect it to do more. As I read it, this function just asserts that there are channels still remaining on the checkpoint item. Whether these remaining  channels are "orphans" requires additional conditions that are checked by the caller of this function. The error message however is one that talks about orphaned channels... it's just confusing as-is I think

materialize-snowflake/snowflake.go
```diff
@@ -312,19 +324,35 @@ func newTransactor(
 			return nil, fmt.Errorf("NewPipeClient: %w", err)
 		}
 
+		sv2 = newStreamV2Manager(ctx, &cfg, open.Materialization.TaskName(), accountName, open.Range)
+		// The manager holds no SQL connection of its own, so the listing runs on the
+		// transactor's.
+		sv2.listChannels = func(ctx context.Context, database, schema, table string) ([]string, error) {
+			return streamV2ListChannels(ctx, db, ep.Dialect, database, schema, table)
+		}
+		sv2.dropStreamingChannel = sm.dropChannel
```
mdibaiee: This is really strange. Why not lend the connection to sv2 and have its own definitions use that connection? we even later end up having checks `if sv.listChannels == nil { return nil }`

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1346 @@
...
+				"channel %q covers %s, which crosses the boundary of this shard's range %s: the shard was split off a boundary its channels do not subdivide along, so the rows that channel holds cannot be attributed to either side. Restore the task's shard key ranges to the topology which appended them, or backfill this binding",
+				sv2ChannelCheckpointItem.ChannelName, keyRange, shard,
+			)
+		}
+	}
+	slices.SortFunc(nested, func(a, b streamV2Range) int {
+		return cmp.Or(cmp.Compare(a.keyBegin, b.keyBegin), cmp.Compare(a.keyEnd, b.keyEnd))
+	})
+
+	// priorDeclaresTargets reports that the prior checkpoint holds an item for every
```
mdibaiee: The code below here can be simplified a lot.

`flush` waits for every channel to commit before it returns the checkpoint that carries a declaration, and no Store runs between that flush and the Acknowledge that cuts over. So: if the durable checkpoint contains the declaration, every channel the declaration overlaps is idle (committed == routed).

A nested item is live iff no nested item with a higher epoch overlaps its range. Everything else is abandoned.

----

A lot of the checks below are unnecessary: they are not possible with the current code. One fault of Claude is that it often tries to maintain backwards compatibility with its own commits in an unmerged PR, and this code seems like one of those where it is trying to maintain backwards compatibility with the previous implementation. With the current implementation, there are invariants that mean we do not need all these checks

materialize-snowflake/snowflake.go
mdibaiee: I suppose this is designed this way to allow for testing, but I think this is too much change in our actual code. I'm sure tests can be written some other way to the same effect

materialize-snowflake/config.go
jacobmarble: I've refactored the feature configuration. There is no longer an invalid state (`!snowpipe_streaming && snowpipe_streaming_v2`), rather a chain of conditions enables `snowpipe_streaming_v2`. The chain of conditions is exactly the way that `snowpipe_streaming` is enabled, plus the `snowpipe_streaming_v2` flag.

jacobmarble:
I've refactored several code areas, and trimmed lots of comments. The SLOC count is below 12,000 now, but still very high. However the code is IMHO much easier to read in many ways.

materialize-snowflake/config.go
```diff
@@ -231,3 +242,12 @@ func (c config) Validate() error {
 
 	return validHost(c.Host)
 }
+
+func (c config) isStreamsV2(deltaUpdates bool) bool {
+	return c.isStreamsV1(deltaUpdates) && boilerplate.ParseFlags(c)[flagSnowpipeStreamingV2]
+}
```
mdibaiee, line 248: beautiful 🤌

materialize-snowflake/snowflake.go
```diff
@@ -387,16 +404,52 @@ type binding struct {
...
+				"table": b.target.Identifier,
+				"blobs": len(priorCheckpointItem.StreamBlobs),
+			}).Info("opened a snowpipe_streaming channel to finish the blobs that path staged")
+		}
+
+		d.snowpipeStreamingV2.addBinding(d.cfg.Database, loc.TableSchema, d.ep.Identifier(loc.TableName), b.target, priorCheckpointItem.StreamV2)
+		b.streamingV2 = true
+		d.bindings = append(d.bindings, b)
+		return nil
+	}
```
mdibaiee, line 436: Eye candy compared to the previous implementation 🤩

materialize-snowflake/streamv2.go
```diff
@@ -0,0 +1,1041 @@
...
+			}
+			dest[nameAt] = &channelName
+			if err := rows.Scan(dest...); err != nil {
+				return nil, fmt.Errorf("scanning the channel listing in %s: %w", in, err)
+			}
+			channelNames = append(channelNames, channelName)
+		}
+		if err := rows.Err(); err != nil {
+			return nil, fmt.Errorf("iterating the channel listing in %s: %w", in, err)
+		}
```
mdibaiee, line 244: I think this logic can be simplified to something like this:
```
type row struct {
    Name string `db:"name"`
}
var rows []row
if err := c.xdb.SelectContext(ctx, &rows, fmt.Sprintf("SHOW CHANNELS IN TABLE %s", in)); err != nil {
    return nil, fmt.Errorf("listing channels in %s: %w", in, err)
}
```

We already use `xdb` for `SHOW` queries precisely because it is able to get specific columns without us having to do gymnastics with columns, etc.

mdibaiee (commented):
Satisfying to read now! A few files remain, will review one final round those files

materialize-snowflake/streamv2.go
jacobmarble: Good one. Fixed.

materialize-snowflake/STREAMV2_RPC.md
```diff
@@ -0,0 +1,125 @@
...
+      |-- append header + row payload --------->|
+      |<------------------------- ok {appended} |
+      |-- wait_commit {token} ----------------->|
+      |<-------------------------------- status |
+      |                                         |
+      |-- shutdown ---------------------------->|
+      |<----------------------- ok, then exit 0 |
+```
+
+The connector spawns the sidecar and writes a random auth token to its stdin,
```
mdibaiee: What is the auth token protecting?

materialize-snowflake/STREAMV2_RPC.md
```diff
@@ -0,0 +1,125 @@
...
+where no argv or environ listing can see it. The sidecar binds a unix domain
+socket (or, started with `--tcp`, an ephemeral localhost port) and prints one
+line to stdout:
+
+```json
+{"ready": true, "transport": "uds"}
+{"ready": true, "transport": "tcp", "port": 54321}
```
mdibaiee: I think this is no longer valid
```suggestion
```

materialize-snowflake/STREAMV2_RPC.md
```diff
@@ -0,0 +1,125 @@
...
+      |-- wait_commit {token} ----------------->|
+      |<-------------------------------- status |
+      |                                         |
+      |-- shutdown ---------------------------->|
+      |<----------------------- ok, then exit 0 |
+```
+
+The connector spawns the sidecar and writes a random auth token to its stdin,
+where no argv or environ listing can see it. The sidecar binds a unix domain
+socket (or, started with `--tcp`, an ephemeral localhost port) and prints one
```
mdibaiee:
```suggestion
socket and prints one
```

mdibaiee (approved):
LGTM! A few nit comments below

I would like to ask @danielnelson for a round of review as well

materialize-snowflake/config.go
```diff
@@ -231,3 +242,12 @@ func (c config) Validate() error {
 
 	return validHost(c.Host)
 }
+
+func (c config) isStreamsV2(deltaUpdates bool) bool {
+	return c.isStreamsV1(deltaUpdates) && boilerplate.ParseFlags(c)[flagSnowpipeStreamingV2]
```
danielnelson: A user sets both `snowpipe_streaming` and `snowpipe_streaming_v2`?

materialize-snowflake/snowflake.go
```diff
@@ -312,20 +316,31 @@ func newTransactor(
...
+		snowpipeStreaming:   sm,
+		snowpipeStreamingV2: sv2,
+		pipeClient:          pipeClient,
+		_range:              open.Range,
+		version:             open.Version,
+		be:                  be,
+		cp:                  make(checkpoint),
+	}
+
+	// Streaming v2 needs its part of the checkpoint state sooner than usual.
```
danielnelson: This comment feels a bit random, suggest removing.

materialize-snowflake/snowflake.go
```diff
@@ -312,20 +316,31 @@ func newTransactor(
...
+	if len(open.StateJson) > 0 {
+		if err := d.UnmarshalState(open.StateJson); err != nil {
+			return nil, fmt.Errorf("unmarshalling connector state: %w", err)
+		}
 	}
+	priorCheckpoint := maps.Clone(d.cp)
```
danielnelson: What makes this "prior" vs just checkpoint?

materialize-snowflake/snowflake.go
```diff
@@ -411,11 +464,10 @@ func (d *transactor) addBinding(ctx context.Context, target sql.Table, streaming
...
 			}
 			log.WithError(err).WithField("table", b.target.Path).Info("not using Snowpipe Streaming for table")
-		} else {
-			b.streaming = true
-			d.bindings = append(d.bindings, b)
-			return nil
 		}
+		b.streaming = true
+		d.bindings = append(d.bindings, b)
+		return nil
```
danielnelson: Can you describe why this change is needed?

materialize-snowflake/sidecar/tests/test_sdk_raw_rows.py
```diff
@@ -0,0 +1,137 @@
...
+no credentials are needed — and fail loudly on a version that closes the route
+rather than letting it silently change what reaches Snowflake.
+
+Run with SNOWPIPE_SDK_GUARD=required to turn a missing SDK into a failure, which
+is what keeps the image build from passing over the guard it exists to enforce.
+"""
+
+import inspect
+import os
+from typing import List
```
danielnelson: `typing.List` is deprecated in favor of `list`, though I'm not familiar with how msgspec is using it.

materialize-snowflake/sidecar/snowpipe_sidecar/server.py
```diff
@@ -0,0 +1,178 @@
...
+Failure policy is crash-only: any unhandled exception logs a traceback and
+exits non-zero, letting the Go supervisor surface the failure.
+"""
+
+import json
+import logging
+import os
+import queue
+import threading
+from typing import Any, Callable, Dict, Optional
```
danielnelson: Deprecated: `typing.Dict` -> `dict` and for `typing.Optional` use `Xyz | None`.

materialize-snowflake/sidecar/snowpipe_sidecar/channels.py
```diff
@@ -0,0 +1,223 @@
...
+The high-performance SDK scopes a client to a single pipe (one auto-created
+default pipe per table), so a client is created per (database, schema, table)
+and channels are tracked by their RPC channel name.
+"""
+
+import logging
+import threading
+import time
+import uuid
+from typing import Any, Callable, Dict, List, Optional, Tuple
```
danielnelson: Tuple also deprecated in addition to Dict, List, Optional.

materialize-snowflake/sidecar/pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+# The SDK is pinned exactly: the append path hands its channels one msgspec.Raw
+# per row, which is an internal of the SDK rather than something it documents —
+# see tests/test_sdk_raw_rows.py. msgspec is the SDK's own dependency, which
+# pins it exactly in turn, and is listed here because this package imports it
+# directly to frame those rows.
+snowpipe-streaming = "1.6.2"
+msgspec = "*"
+
+[tool.poetry.group.dev.dependencies]
+pytest = "^8.3.4"
```
danielnelson: I suggest adding a type checker too, I've had good luck with these finding issues.  I'm a fan of mypy since I think its the most standard/official checker but there are several good options these days.

danielnelson (approved):

PR #5164, opened 2026-09-01
source-shopify-native: capture media on product variants

### LINKED ISSUE #5163: source-shopify-native: capture media on product variants
## Request

Capture the `media` field on product variants. `media` is already captured at the product level by the `product_media` stream; this is the equivalent at variant level, on `product_variants`.

## Scope

Each variant in `product_variants` gains a `media` list, carrying `id`, `alt` and `mediaContentType`, plus `image` (`url`, `width`, `height`) for image media.

## Notes

Two constraints shape the implementation and are worth recording:

- `products → variants → media` is two nested connections, which is exactly Shopify's bulk-operation ceiling (the root connection isn't counted). Adding a connection beneath `media` later would push the query over the limit and break the stream.
- `Product.media` takes a `query` argument, which is how `product_media` filters to `media_type:IMAGE`. `ProductVariant.media` takes only pagination arguments, so videos and 3D models come through with no `image` object. `mediaContentType` is captured so consumers can tell those from an image.

Bulk JSONL flattens connections onto their own lines, so variant media arrives carrying the variant's id in `__parentId`. The existing parser raised on any line that was neither a product nor a variant, so the reassembly had to change too.

## Verification

Verified end to end against a live capture. Two variants of **Selling Plans Ski Wax** (`gid://shopify/Product/8086708027437`) were seeded with different images and a third left with none. Each media node arrives with `__parentId` set to its variant rather than the product, attaches to the correct variant, and the third variant stays empty — so media is neither falling through to a first match nor leaking across siblings within a product.

The capture snapshot now prefers a `product_variants` document whose variants carry media, mirroring the existing `order_returns` preference. Without that it lands on the Gift Card and records all-empty arrays.

SeanWhelan:
**Description:**

Adds `ProductVariant.media` to the `product_variants` bulk query, so each variant carries its attached media — `id`, `alt`, `mediaContentType`, and `image` for image media. A customer needs at variant level what `product_media` already gives at product level.

**Workflow steps:**

Query change in `graphql/products/variants.py` plus its `process_result`. `media` is a connection, so bulk JSONL puts each node on its own line carrying the variant's id in `__parentId`; media attaches by that id rather than positionally, since Shopify only guarantees a child follows its parent somewhere. `products → variants → media` sits at Shopify's two-connection nesting ceiling, so nothing can nest beneath `media`.

No access scope beyond the `read_products` the stream already requires. Verified with the capture snapshot test, after seeding two Ski Wax variants with media and a third without. `test_snapshots.py` now prefers a product whose variants carry media, which accounts for most of the snapshot diff.

Existing captures won't backfill — media appears on products whose `updatedAt` advances after deploy.

**Checklist:**

Closes #5163

### DIFF
--- a/source-shopify-native/source_shopify_native/graphql/products/variants.py
+++ b/source-shopify-native/source_shopify_native/graphql/products/variants.py
@@ -31,6 +31,22 @@ class ProductVariants(ShopifyGraphQLResource):
                 image {
                     id
                 }
+                media {
+                    edges {
+                        node {
+                            id
+                            alt
+                            mediaContentType
+                            ... on MediaImage {
+                                image {
+                                    url
+                                    width
+                                    height
+                                }
+                            }
+                        }
+                    }
+                }
                 selectedOptions {
                     name
                     value
@@ -69,12 +85,17 @@ class ProductVariants(ShopifyGraphQLResource):
         log: Logger, lines: AsyncGenerator[bytes, None]
     ) -> AsyncGenerator[dict, None]:
         VARIANTS_KEY = "variants"
+        MEDIA_KEY = "media"
         current_product = None
-        current_variant = None
+        # Variants of the current product, keyed by id, so media lines can be attached to the
+        # variant named by their __parentId. Shopify only guarantees a child appears somewhere
+        # after its parent, not immediately after, so media can't be attached positionally.
+        current_variants_by_id: dict[str, dict[str, Any]] = {}
 
         async for line in lines:
             record: dict[str, Any] = json.loads(line)
             id: str = record.get("id", "")
+            parent_id: str = record.get("__parentId", "")
 
             if "gid://shopify/Product/" in id:
                 if current_product:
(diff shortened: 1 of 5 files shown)


source-shopify-native/source_shopify_native/graphql/products/variants.py
```diff
@@ -83,24 +104,33 @@ async def process_result(
...
+                record[MEDIA_KEY] = []
+                current_product[VARIANTS_KEY].append(record)
+                current_variants_by_id[id] = record
+
+            elif parent_id in current_variants_by_id:
+                # Any other line parented to one of this product's variants is one of that
+                # variant's media nodes. Media is matched on __parentId rather than on a gid
+                # prefix because ProductVariant.media takes no media_type filter, so a node can
+                # be a MediaImage, Video, ExternalVideo or Model3d.
+                current_variants_by_id[parent_id][MEDIA_KEY].append(record)
```
Alex-Bair, line 133: Instead of using a catch-all here that classifies any line with a current variant parent as media, could we gate it on the media gid prefixes instead? That way, if another connection is added later and we forget to update `process_result` to handle it, we won't accidently add it to the `media` field and it'll fall through to the "Unidentified line in JSONL response" error.

For example, near the top of the file we can list out the various possible media gid prefixes like:
```python
  MEDIA_GID_PREFIXES = (
      "gid://shopify/MediaImage/",
      "gid://shopify/Video/",
      "gid://shopify/ExternalVideo/",
      "gid://shopify/Model3d/",
  )
```

then instead of `elif parent_id in current_variants_by_id` we only stitch the line into `media` if it starts with one of those prefixes:
```python
              elif id.startswith(MEDIA_GID_PREFIXES):
                  if parent_id not in current_variants_by_id:
                      log.error(
                          "Media's parent is not a variant of the current product. Check if the JSONL response from Shopify is not ordered correctly.",
                          {
                              "media.id": id,
                              "media.__parentId": parent_id,
                              "current_product.id": current_product.get("id") if current_product else None,
                          },
                      )
                      raise RuntimeError()

                  current_variants_by_id[parent_id][MEDIA_KEY].append(record)

```

nicolaslazo:
I'm evaluating whether we should normalise these "variant media" objects into their own separate stream and let consumers join them downstream.

Do you know if updating a media object bumps the product's `updatedAt` timestamp? If not, then we might miss media updates until the product itself is updated for an unrelated reason.

Alex-Bair (commented):
It looks like @nicolaslazo is reviewing this, but I had one hardening ask as I skimmed through.

Some additional context around why this isn't necessarily a quick ✅: Shopify places [restrictions](https://shopify.dev/docs/api/usage/bulk-operations/queries#operation-restrictions) on bulk queries. The ones we need to keep in mind are:
> Maximum of five total connections in the query
> Maximum of two levels deep for nested connections.

The changes in this PR stay below that budget, so as-is it's fine, but we need to be cognizant that the `product_variants` stream is using up 3 of the 5 total connections after this change & moving forward we only have 2 more connections left in the budget.

PR #5168, opened 2026-09-02
source-gong: add live integration test against the Gong sandbox

SeanWhelan:
Wires `source-gong`'s snapshot tests to a real Gong developer instance (Collective sandbox, region `us02-16381`) and adds the Bruno collection used to seed it. No connector code changes.

## What's here

- `config.yaml` — sops-encrypted credentials for the sandbox.
- `test.flow.yaml` — points at `config.yaml`, bindings from live discovery.
- `acmeCo/` — generated collections and schemas for all four streams.
- `test_capture` — runs `flowctl preview` against the instance and snapshots one representative document per stream, chosen deterministically so the snapshot doesn't churn on document ordering. Contact fields (`emailAddress`, `phoneNumber`, `personalMeetingUrls`, …) are redacted, since this snapshot is committed to a public repo.
- `bruno/` — read probes for each stream, plus a `Seeding/` folder that creates two calls and uploads their media, so the account can be repopulated from a clean checkout.

The capture snapshot covers `calls`, `scorecard_definitions` and `users`.

## Seeding the sandbox

`bru run Seeding --env Gong --sandbox developer`, then wait ~4 minutes for Gong to analyse the media. Things that cost real time to discover, all documented in the collection alongside the observations behind them:

- Gong discards uploaded calls whose media it finds no speech in, and calls under 60 seconds (measured after trimming blank audio). Neither is reported through the API — the upload returns `201` and the call simply never appears. Only the UI says why.
- Media must be unique per call across the whole account, not just within a run. Reusing a file returns a bare `400 "Bad request"`.
- A call with no media attached is invisible to every read endpoint. Its `callId` is recoverable by re-posting the same `clientUniqueId`, which names the existing call in the duplicate error.
- Seed data is attributed to a dedicated non-human team member (`primary_user_id`), never a real account. The environment file ships a loud placeholder so an unconfigured run fails instead of writing to a real user.
- Empty *filtered* lists return `404`; empty *unfiltered* lists return `200`. Both 404 strings were checked against `_NO_RESULTS_404_PATTERN`.
- The two scorecard streams have no create endpoints and are UI-only to seed; the runbook is in the collection root.

## Not covered

`scorecards` is absent from the capture snapshot because the connector cannot currently read that stream at all. Three layered defects, each masking the ones beneath it — filed as a separate issue rather than fixed here, since one of them is a data-loss bug in shared fetch code and deserves its own review.


source-gong/config.yaml
```diff
@@ -0,0 +1,15 @@
...
+    access_key_sops: ENC[AES256_GCM,data:rqGDi7szCifNy3gX+u26ZA4vh0V+61cKB1kALMhQmZg=,iv:c3UBa6ZIIxc93cn6Z8lW0aTiHFhLIFGmHhDZ+8yQW4s=,tag:QMibcA9PopHYLjMSYgKpYw==,type:str]
+    access_key_secret_sops: ENC[AES256_GCM,data:KfiHiyM1+6/K87u5zeP0PjyM1z/dzAZeSrL2vVLeU3alVNoKBvkAkMuJs75kSL5YKQWJpuH22KsUEJxfLs8NGKQt76VKniMLsQEW4t7zoSX8X35TFTidONAPu80JM2qoXXztpJ+bGF6VPvUSyi6A0rcEfxqLjfKoVX54K2ErAeKBqjVv18pm+8cJVpSnAEqLRpstIRQUdIU=,iv:0rWV+12DE0DHk1k3A/M4H+Lq+8i/HVx3H5xpoCZKlrg=,tag:lR00S6y4bBSs8oJor5ApOg==,type:str]
+region: us02-16381
+start_date: "2026-08-01T00:00:00Z"
+sops:
+    encrypted_suffix: _sops
+    gcp_kms:
+        - created_at: "2026-09-01T15:02:03Z"
+          enc: CiUAW8BC2IxOqplmR4xnTlrrgOxqXvQXGpB1QIdsM45A+WYADugtEkkAXnXMaENXpyoP5bzIWSZ7ffMu6oWUWuZySwG/LGk9rjtjR/w3Sf/BLrGpXMdgBN4/O10kTZYNbgIEFNlRV30RSFZaDlopWyaD
+          resource_id: projects/helpful-kingdom-273219/locations/us-central1/keyRings/dev/cryptoKeys/CI-estuary-flow
```
Alex-Bair, line 12: The connectors' CI doesn't have access to this key ring, and that's why the CI check is failing for `source-gong`. The `projects/estuary-theatre/locations/us-central1/keyRings/connector-keyring/cryptoKeys/connector-repository` key ring needs to be used for encrypting `config.yaml`s that live in this repo instead; our CI is able to use that one. Here's an example `config.yaml` that does that, if it's helpful:
https://github.com/estuary/connectors/blob/002353c38936148a13caf050b5c8d072b91936a3/source-intercom-native/config.yaml#L10

SeanWhelan: Thanks for catching that @Alex-Bair, resolved!

Alex-Bair (approved):
LGTM, thanks for getting a Gong sandbox account set up!

PR #5179, opened 2026-09-02
source-snowflake: Scope dynamic table listing to the capture database

willdonnelly:
**Description:**

The dynamic table listing used `SHOW DYNAMIC TABLES IN ACCOUNT` and then keyed every result under the configured capture database, discarding the database name each row actually came from. A dynamic table in some other database of the same account, with the same schema and table name as a captured table, was therefore mistaken for that table. When the impostor had a full refresh mode the real table was filtered out of discovery, and Flow's auto-discover then removed the binding as though the table had been dropped.

The account-wide scope dates from before snowflakeObject carried a database component, and the cfg.Database stamping was added mechanically when that component was introduced. Every other discovery query is already scoped to the capture database, so this brings the dynamic table listing in line with them and adds the same sanity check the other queries have against results from unexpected databases.

**Workflow steps:**

No user action required.

**Checklist:**

- [x] If fixing a regression, I have linked the PR that introduced the regression: N/A

### DIFF
--- a/source-snowflake/CHANGELOG.md
+++ b/source-snowflake/CHANGELOG.md
@@ -1,5 +1,12 @@
 # source-snowflake
 
+## 2026-09-02
+
+### Fixed
+- Dynamic table discovery is now scoped to the capture database. Previously a
+  same-named dynamic table in another database could cause a captured table to
+  disappear from discovery.
+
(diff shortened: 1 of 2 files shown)


Alex-Bair (approved):
LGTM

PR #5180, opened 2026-09-02
source-hubspot-native: limit search chunks to below 10,000 results

Alex-Bair:
**Description:**

### Problem
The delayed streams for CRM objects read changes through HubSpot's Search API, sorted ascending by last-modified time. HubSpot does not consistently honor that sort. Records come back out of order, sometimes by seconds and sometimes by close to an hour, even in windows that are already an hour old.

The connector relies on HubSpot's sorting to determine when a checkpoint is safe to emit. A search can serve at most 10,000 results, so a window with more records than that was cut into chunks at the 10,000 offset cap, and the resume cursor for the next chunk was the newest timestamp seen so far. That is only correct if every record below that timestamp has already been returned, which is exactly what the sort was supposed to guarantee. Because a misordered record could mean records below the cursor were about to be skipped, the delayed stream treated any out-of-order result as fatal and crashed the task. A record HubSpot persistently misorders turns that into a crash loop for the whole capture.

### Solution
This PR addresses that problem by making it so searches never reach the 10,000 cap anymore and the connector no longer needs HubSpot's sorting to be correct. A window that holds more than 10,000 records is split by time into chunks that each hold fewer than 10,000, and every chunk is read completely. Within a completely read chunk, the order of the records is irrelevant. There is nothing left unread that the sort could have hidden, and the chunk's end time is a checkpoint that stands on its own. The out-of-order exception is gone, along with the client-side filters that dropped records whose reported timestamp fell outside the window they were returned in. HubSpot's own filter now decides which records belong to a window, and every record it returns is captured.

As for how a window is read now, the connector relies on the `total` HubSpot includes on the first page of a search to determine how many results are in the search's window. If it is at most 10,000, the window is read to the end like before this PR. If it is larger than 10,000, the connector peeks at the last page the cap allows, at offset 9,800. The earliest timestamp on that page bounds a chunk of at most 9,800 records. A fresh search bounded there confirms the chunk's total is under the cap before it is read in full, and the next chunk resumes one millisecond after where this one ended. The peek only sizes the chunk. Completeness comes from the verified total and the full read, so a misordered or stale timestamp on the peek page can only make a chunk smaller than necessary, never incomplete. When the peek offers no usable timestamp the window is halved instead, and a single millisecond holding more than 10,000 records is drained with the existing id-paged cycle path. Each chunk's end is checkpointed as it completes, so large windows still make incremental progress.

Motivating Slack thread: [link](https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1788361643803779?thread_ts=1788348729.468039&cid=C03Q2NRFKDL)

**Notes for reviewers:**

Best reviewed commit-by-commit. The change is conceptually simple - keep searches below the 10,000 cap so we don't have to worry about HubSpot's loose definition of "sorted" between separate searches - but a fair amount of code had to be touched to do that, and I tried to break those changes into atomic commits to they're easier to review.

**Checklist:**

### DIFF
--- a/source-hubspot-native/CHANGELOG.md
+++ b/source-hubspot-native/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 2026-09-02
 ### Fixed
+- CRM object bindings no longer fail when the HubSpot Search API returns records out
+  of order. Search windows holding more than 10,000 records are split by time into
+  chunks HubSpot can return completely, so result order is never relied on. Records
+  HubSpot returns for a window are now always captured, even when their reported
+  last-modified time falls outside it.
 - The `form_submissions` binding could permanently skip submissions that arrived on a
   form while a sweep was still checking other forms. Each sweep now only emits
   submissions made at least five minutes before it started and checkpoints the newest
--- a/source-hubspot-native/source_hubspot_native/api/companies.py
+++ b/source-hubspot-native/source_hubspot_native/api/companies.py
@@ -77,5 +77,5 @@ def fetch_delayed_companies(
         )
 
     return fetch_chunked_changes_with_associations(
-        Names.companies, Company, do_fetch, log, http, with_history, since, until
+        Names.companies, Company, do_fetch, log, http, with_history
     )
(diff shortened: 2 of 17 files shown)


Alex-Bair:
The `source-hubspot-native` CI check failure is due to HubSpot removing a field from their responses. https://github.com/estuary/connectors/pull/5181 updates the capture snapshot so the CI check passes.

nicolaslazo (approved):
LGTM

PR #5185, opened 2026-09-03
materialize-redshift: reland post-commit apply (#5157) with a runtime v1 fix

mdibaiee:
**Description:**

Relands #5157 — materialize-redshift's conversion to "Recovery Log with Idempotent Apply" with the shard-zero coordinator and a downgrade-compatible checkpoint mirror — and fixes a defect in it that only shows on runtime v1.

#5157 merged as `2c98ff379..73f4674b0` and was then reverted commit-by-commit (`d2637c670..ea6af8f58`). The first commit here reverts those 17 reverts as one, restoring the tree exactly as merged at `73f4674b0`. The second commit is the fix.

**The fix.** Runtime v1 never populates `Acknowledge`'s `statePatches`; only runtime-next does. #5157 had the primary learn its own just-staged entries from that echo. On v1 an in-session `Acknowledge` therefore found nothing pending and applied nothing — and worse, since `connectorState` keeps one entry per `(stateKey, rangeKey)` and the runtime merge-reduces `StartCommit` patches, each transaction's patch overwrote the previous one's staged files: when the session finally rotated, only the last transaction's data landed and everything before it was lost. The legacy checkpoint mirror also never advanced past crossover, so a downgrade would have replayed everything since. `StartCommit` now records the primary's own entries directly, and `mergePeerStatePatches` skips its own range so runtime-next's echo of them doesn't trip the collision check. Works on both runtimes; v1 is the default for materializations (`useRuntimeV2` requires the `enable-runtime-v2` shard label).

**Workflow steps:**

No user-visible change. Upgrading adds the `checkpoint_tokens` column on Open; downgrading to a pre-#5157 image resumes from the mirrored checkpoint. See the `lastRuntimeCheckpoint` comment in `driver.go` for the one documented downgrade caveat (a task that crashed and recovered must commit one more transaction before being downgraded).

**Notes for reviewers:**

- Review the fix commit on its own; the reapply commit is a pure revert-of-reverts (`git diff 73f4674b0 <reapply>` is empty).
- Verification so far, all on the #5157 tree plus this fix or on #5157 alone: full `TestIntegration` on this exact tree, green on both runtime-next (`flowctl raw preview-next`) and runtime v1 (`FLOW_TEST_RUNTIME=v1`), with the v1 snapshot byte-identical to the golden; flow's `materialize-consistency` suite 17/17 on #5157; a hand-driven upgrade→downgrade test against live Redshift, which found the documented caveat above.
- Every one of those runs v2 semantics or crosses session boundaries, which is why the v1 defect wasn't caught. The remaining commits add a runtime v1 pass to `testutil`/CI: `FLOW_TEST_RUNTIME=v1` runs legacy `flowctl preview` as a single session with no drain and requires every binding to hold data afterwards (a generic liveness check, run for every materialization connector), then a drained run that must match the connector's runtime-next golden with its output-only lines removed (both runtimes must leave the same destination; runs for every connector) and is snapshotted as a table-only `<test>-runtime-v1` golden where the connector has one (redshift's is byte-identical to its runtime-next golden) — pinned to a tagged `v0.6.13` flowctl (`flowctl-v1`, via `FLOW_TEST_FLOWCTL`) since legacy `preview` is slated to move onto runtime-next, while the runtime-next pass keeps tracking `dev-next`. Verified to fail on this tree with the fix commit reverted and pass with it (see comments).

**Checklist:**

- [x] If fixing a regression, I have linked the PR that introduced the regression: #5157 (the defect was introduced within it)
- [ ] If there are user-visible changes, `CHANGELOG.md` has been updated — none beyond what #5157 carried
- [ ] If there are user-facing behavior changes, documentation has been updated — none


mdibaiee:
Full `TestIntegration` (runtime-next, `flowctl raw preview-next`) on this exact tree: **all green** — `materialize` 158s, `apply` 284s, `apply-drain` 83s, `migrate` 287s, `idempotency` (both subtests) 71s; 882s total.

A runtime **v1** pass via legacy `flowctl preview` is running now; results and the CI step that makes it permanent follow in this PR.

mdibaiee:
Runtime **v1** pass (`FLOW_TEST_RUNTIME=v1`, legacy `flowctl preview --sessions=-1,0`, single-shard): **all green** — `materialize` 164s, `apply` 295s, `apply-drain` 85s, `migrate` 280s, `idempotency` 72s. Ran with `UPDATE_SNAPSHOTS=true` and the golden did not change: the v1 destination snapshot is byte-identical to the runtime-next one.

The third commit makes this permanent: the `FLOW_TEST_RUNTIME` toggle in `testutil`, and a CI step that repeats the connector tests on v1 for an allowlist starting with `materialize-redshift`. `materialize-databricks` is the other `Shards: 2` connector whose golden would compare cleanly; it should be added once it's been run on v1 once.

mdibaiee:
Fourth commit pins the v1 pass to a fixed flowctl: `fetch-flow.sh` also downloads tagged `v0.6.13` (its `preview` is still the legacy `runtime` crate — `--sessions/--fixture/--output-*`, no `--shards`) as `flowctl-v1`, and the v1 CI step drives tests with it via `FLOW_TEST_FLOWCTL`. The runtime-next pass keeps tracking `dev-next`. Rationale: legacy `preview` is slated to be rewired onto runtime-next, which would otherwise turn the v1 step into a second v2 run.

Verified end-to-end locally through the pinned binary (`FLOW_TEST_FLOWCTL=…/flowctl-v1 FLOW_TEST_RUNTIME=v1`): full `TestIntegration` **all green**, golden unchanged under `UPDATE_SNAPSHOTS=true`.

mdibaiee:
Fifth commit turns the v1 pass into a **generic liveness check** rather than a golden comparison. \`FLOW_TEST_RUNTIME=v1\` runs legacy \`flowctl preview --sessions=-1\` — one session, **no drain** — and requires every binding to be non-empty afterwards. The drained variant I had before masked the bug (its recovery Acknowledge applied everything at the end, so the snapshot matched); this one observes what a single v1 session actually applied. No golden, so it now runs for **every** materialization connector in CI, not an allowlist.

Verified it discriminates, on the pinned v1 flowctl:
- this branch (with the fix): **PASS** (153s)
- same tree with only the fix commit reverted: **FAIL** — \`runtime v1: [simple_standard_…] is empty after a single session; the connector applied nothing it acknowledged within the session (does it only learn its own staged work from Acknowledge's statePatches, which runtime v1 never sends?)\`

mdibaiee:
**Severity correction for the fix commit.** On runtime v1 the pre-fix connector didn't just apply late — it **lost data**. \`connectorState\` holds one \`stateItem\` per \`(stateKey, rangeKey)\` and the runtime reduces \`StartCommit\` patches by merge, so with no in-session Acknowledge clearing the entry, each transaction's patch overwrote the previous one's staged files. A drained v1 run without the fix landed only the last transaction's rows (every \`01:00:xx\` row from earlier transactions missing) — snapshot diff on this branch with the fix commit reverted. The description's "applied nothing within a v1 session" understates it; a task running the reverted #5157 on v1 for more than one transaction per session dropped everything but the last.

mdibaiee:
Sixth commit adds the runtime v1 **golden**: after the liveness check, the pass drops the tables, runs the whole fixture drained through legacy \`flowctl preview --sessions=-1,0\`, and snapshots the destination against \`.snapshots/<test>-runtime-v1\` (table-only — the apply/state output embeds per-run values and differs by runtime). Redshift's new golden is byte-identical to its runtime-next one. A connector without a v1 golden logs and enforces liveness alone, so the step stays runnable everywhere until owners generate theirs with \`UPDATE_SNAPSHOTS=true\`.

Verified both ways on the pinned v1 flowctl:
- this branch, plain compare (no \`UPDATE_SNAPSHOTS\`): **PASS** (258s), no \`.actual\` written
- fix commit reverted, liveness stubbed so the run reaches the compare: **FAIL** — 12 rows present in the golden missing from the destination (every earlier-transaction row; only the last transaction's rows landed), \`.actual\` written for the diff

So the two checks are complementary: liveness catches "nothing applied in-session" and is generic; the golden catches "the wrong things applied" and needs one generated file per connector.

mdibaiee:
Seventh commit cross-checks the two runtimes against each other: the v1 pass's drained run must now equal the connector's **runtime-next** golden with the runtime-next-only lines stripped (\`["applied.actionDescription", …]\` and \`["connectorState", …]\` — verified to be the only stdout-line prefixes in any materialization golden). Runs for every connector regardless of whether it has a \`-runtime-v1\` golden yet, so every connector's v1 destination content is held to its runtime-next content from day one. The \`-runtime-v1\` golden stays as the pass's own record.

Redshift on the pinned v1 flowctl: liveness + \`-runtime-v1\` golden + cross-check all **green** (260s). (One prior attempt hit the shared cluster's known intermittent connection drop during Apply's concurrent \`CREATE TABLE\`; unrelated, reran clean.)

mdibaiee:
Runtime v1 goldens generated and committed for **10 connectors**, each on the pinned v1 flowctl with liveness + cross-check against the runtime-next golden passing: redshift, bigquery, databricks, bigtable, clickhouse, dynamodb, mongodb, spanner, snowflake, motherduck (both variants).

Doing this shook out two harness bugs, both fixed in \`60ac47971\`: the drained run now runs as a **fresh task** rather than dropping the liveness run's tables (a destination-authoritative connector resumed from the liveness run's checkpoint stored in the destination and applied only the fixture's tail — postgres/sqlserver/spanner; and Snowflake upper-cases identifiers so the name-based sweep never dropped its tables, stacking the drained rows on the liveness rows), and a connector whose \`SnapshotTestResource\` is unimplemented (elasticsearch, iceberg) now logs instead of failing liveness.

Not yet: **postgres, mysql, sqlserver, elasticsearch, eventbridge, iceberg, s3-iceberg** — all blocked on the local Docker disk being full (containers die at init with ENOSPC), will follow once that's cleared. **hubspot** skips its integration test without \`-interactive\` (CI too). **azure-fabric-warehouse** is the known Azure login outage.

mdibaiee:
Runtime v1 goldens are now committed for **every materialization connector whose integration test can run here** — 17 connectors, 24 golden files (elasticsearch has an opensearch variant, motherduck two variants, s3-iceberg six tasks incl. Glue). Each was generated on the pinned v1 flowctl with both the liveness check and the cross-check against the connector's runtime-next golden passing.

| Connector | v1 pass | Notes |
|---|---|---|
| redshift | ✅ | byte-identical to runtime-next golden |
| bigquery, databricks, bigtable, clickhouse, dynamodb, mongodb | ✅ | |
| spanner, postgres, sqlserver | ✅ | after the fresh-task fix — destination-authoritative, resumed from the liveness run's checkpoint before |
| snowflake | ✅ | after the fresh-task fix — upper-cased identifiers defeated the name sweep before |
| motherduck (×2), iceberg, s3-iceberg (×6) | ✅ | |
| elasticsearch (×2), iceberg | ✅ | `SnapshotTestResource` unimplemented → liveness logged/skipped, cross-check still enforced |
| eventbridge | ✅ | after the fresh-materializer fix — its test double drains SQS once and caches |
| mysql | ✅ | needed ~7.7 GB free in the Docker VM for SingleStore |
| hubspot | — | integration test skips without `-interactive` (CI too) |
| azure-fabric-warehouse | — | known Azure login outage (`Couldn't complete the operation due to a system update`), red on main as well |

Three harness fixes fell out of the fan-out and are in this PR (`60ac47971`, `bb1faa2c8`): the drained run uses a fresh task, a fresh materializer, and tolerates an unimplemented `SnapshotTestResource`. No connector showed a genuine v1 defect beyond redshift's. Note `b40c4fa2c`'s message lists mysql prematurely; its golden landed in the following commit.

mdibaiee:
CI triage for the run at `2cb647439` (4 failed jobs):

- **materialize-kafka** — my bug: the new runtime v1 step ran `go test ./...` in a Rust connector (`matched no packages`). Fixed in `da5a1272e` with the same `Cargo.toml` guard `tests/materialize/run.sh` uses.
- **source-sqlserver** — `TestBasics/SimpleCapture`, pre-existing flake: it also failed on `main` run 33795523460 and passes on the runs around it. Untouched by this PR.
- **materialize-azure-fabric-warehouse** — the known Azure login outage, red on `main` too.
- **materialize-motherduck** — failed in the *runtime-next* `ducklake/apply` snapshot (`requiredString` nullability true→false), not in the v1 pass. Nothing in this PR touches the apply path and `main` is green on it 4/4 recently, so I'm treating it as an open question until the rerun triggered by `da5a1272e` reports; if it reproduces I'll dig in.

danielnelson (dismissed):

mdibaiee:
Rerun at `da5a1272e` (run 33888917444): **68/69 green**. materialize-kafka passes with the `Cargo.toml` guard; materialize-motherduck passes, so its `ducklake/apply` failure was a flake; source-sqlserver passes. The only red job is materialize-azure-fabric-warehouse, the known Azure login outage that is red on `main` as well.

danielnelson (approved):

PR #5189, opened 2026-09-03
source-mysql: Feature-flagged support for MariaDB system-versioned tables

### LINKED ISSUE #4139: Support for Temporal tables in Mariadb
We offered building out support for temporal tables in MariaDB for a customer and they ended up signing.  Will D and I spoke about it previously and he has context.

willdonnelly:
**Description:**

Implements support (currently feature-flagged but intended to be enabled-by-default for new captures at a future date) for system-versioned tables in MariaDB.

This fixes https://github.com/estuary/connectors/issues/4139

**Workflow steps:**

Users will currently need to set the `system_versioned_tables` feature flag manually. At a future date (after we can do configUpdate migrations more easily, I imagine) this should become the default. At that time we should also introduce proper documentation mentioning that it's a thing we support.

**Checklist:**


source-mysql/mariadb_temporal_test.go
Alex-Bair, line 1: Should we add a test case to cover precise mode backfill behavior when there are replication events for the row at a backfill chunk boundary? I _think_ there's a data loss bug because the `row_end` key column is encoded with different FDB tuple types on the backfill and replication paths (backfill gives `[]byte`, replication gives `string`), and looking at how they're [compared under the precise backfill mode](https://github.com/estuary/connectors/blob/d726c76b642616c864127c857d8bfb6fdd3e7d02/sqlcapture/capture.go#L922-L927), it looks like we could end up skipping replication events.

Here's a test to better illustrate the case I'm thinking of:
```go
// TestMariaDBTemporalTablesBoundaryUpdate updates a row while it sits exactly at the
// backfill chunk boundary of a system-versioned table.
//
// A precise backfill drops replication events whose row key sorts after the backfill
// cursor, on the assumption that a later chunk will scan that row. The key of a
// system-versioned table ends with row_end, and the backfill and replication paths
// serialize timestamps as different FDB tuple types, so the boundary row's events
// always sort after the cursor. The resumed scan starts strictly after the cursor and
// never revisits the row, so the update is lost.
func TestMariaDBTemporalTablesBoundaryUpdate(t *testing.T) {
	if os.Getenv("TEST_MARIADB") != "yes" {
		t.Skip("Skipping MariaDB specific test, set TEST_MARIADB=yes to enable")
	}
	var tb, ctx = mysqlTestBackend(t), context.Background()
	tb.config.Advanced.FeatureFlags = "system_versioned_tables"

	var uniqueID = uniqueTableID(t)
	var table = tb.CreateTable(ctx, t, uniqueID, "(id INT PRIMARY KEY, data TEXT) WITH SYSTEM VERSIONING")
	tb.Query(ctx, t, fmt.Sprintf("INSERT INTO %s VALUES (1,'init'),(2,'init'),(3,'init'),(4,'init');", table))

	var cs = tb.CaptureSpec(ctx, t, regexp.MustCompile(uniqueID))
	cs.Validator = &st.OrderedCaptureValidator{}
	cs.EndpointSpec.(*Config).Advanced.BackfillChunkSize = 2

	// Stop after the first chunk, which scans ids 1 and 2 and leaves the backfill
	// cursor on the current row version of id=2.
	sqlcapture.TestShutdownAfterBackfill = true
	sqlcapture.TestShutdownAfterCaughtUp = true
	t.Cleanup(func() {
		sqlcapture.TestShutdownAfterBackfill = false
		sqlcapture.TestShutdownAfterCaughtUp = false
	})
	cs.Capture(ctx, t, nil)
	require.Contains(t, string(cs.Checkpoint), `"backfilled":2`)

	// Update the boundary row, plus an already-scanned row as a control. Neither is
	// scanned again, so both updates have to arrive via replication.
	tb.Query(ctx, t, fmt.Sprintf("UPDATE %s SET data = 'boundary-update' WHERE id = 2;", table))
	tb.Query(ctx, t, fmt.Sprintf("UPDATE %s SET data = 'interior-update' WHERE id = 1;", table))

	// Run the rest of the backfill to completion.
	sqlcapture.TestShutdownAfterBackfill = false
	cs.Capture(ctx, t, nil)
	require.Empty(t, cs.Errors)

	var summary = cs.Summary()
	require.Contains(t, summary, "interior-update", "control: an update to an already-scanned row must be captured")
	require.Contains(t, summary, "boundary-update", "an update to the row at the chunk boundary must be captured")
}
```

Alex-Bair: It looks like this is pre-existing behavior that could occur for any `TIMESTAMP`/`DATETIME` key columns, which I'm hesitant to guess is relatively uncommon for non-system-versioned tables. But based on my very rudimentary understanding of system-versioned tables, this sounds like a situation that could happen more frequently since `row_end` will be a `TIMESTAMP` key column.

nicolaslazo (dismissed):
LGTM

source-mysql/mariadb_temporal_test.go
willdonnelly: If we're serializing the same key value as distinct FDB tuples via replication and backfills that's a bug full stop, so before any other considerations we ought to either fix that by making the serialization consistent or by declaring that column type unable to be reliably predicted.

I need to think more about how backfill ordering interacts with `FOR SYSTEM_TIME ALL` though, thinking this through is making me question whether it's correct right now.

willdonnelly: Okay so yeah, this is just a bug impacting any temporal keys and should be simple enough to fix. And as far as I can tell the backfill ordering behavior is correct and I was just overthinking this, the `row_end` timestamp is just part of the ordering clause and the `WHERE` constraints the same as any other multi-column key involving a timestamp.

Alex-Bair (approved):
LGTM, thanks!

PR #5190, opened 2026-09-04
materialize-iceberg: use idempotency token to avoid job duplication

danielnelson:
**Description:**

We want to avoid duplicate jobs created due to unexpected connector restarts.  They waste resources and can cause the new job to fail due to concurrency errors.

Save an idempotency token that can be used by the merge job.  If the connector is interrupted it can adopt the job during the next run avoiding a duplicate job.

We don't rely on the job running once for correctness, it will be correct even with multiple runs.  In some cases we need to start a new job, such as if the job itself has changed, or if the job was cancelled or failed.

This is kind of a extension of the work on https://github.com/estuary/connectors/pull/5152.

I recommend looking at the commits individually, since there is a refactor commit.

**Workflow steps:**

No workflow changes.

**Checklist:**


materialize-iceberg/emr.go
```diff
@@ -90,6 +90,42 @@ func (e *emrClient) ensureSecret(ctx context.Context, wantCred string) error {
...
+	start, err := e.c.StartJobRun(ctx, startInput)
+	if err == nil {
+		return start, nil
+	}
+
+	// AccessDenied likely means the execution role lacks
+	// `emr-serverless:TagResource`. Disable tagging for the lifetime of
+	// this client and retry without tags so the materialization can keep
+	// running on existing IAM policies.
+	if startInput.Tags != nil && isAccessDeniedErr(err) {
```
jacobmarble, line 103: Is this a common condition? Seems like a leftover from functional testing.

materialize-iceberg/emr.go
```diff
@@ -90,6 +90,42 @@ func (e *emrClient) ensureSecret(ctx context.Context, wantCred string) error {
...
+		return start, nil
+	}
+
+	// When re-using a ClientToken, the job configuration must exactly match.
+	// We always ensure this is the case during normal operation.
+	//
+	// However should there be code change that modifies how the job is
+	// created, we don't want the task to get recovery loop.  Starting a new
+	// job with a different ClientToken will duplicate the work, but this is
+	// better than not being able to make progress.
```
jacobmarble: "there should be code change" -- wat? I suggest dropping the second paragraph entirely, it seems not be "load-bearing". ;-)

jacobmarble (dismissed):

materialize-iceberg/emr.go
danielnelson: Hitting this is pretty common.  Originally the `emr-serverless:TagResource` was not needed, later some optional tracking was added that needed it and so it was made an optional permission, since adding new permission requirements takes a lot of coordination.

It isn't new to this change, however perhaps one thing to call out is that I no longer re-roll the `ClientToken` at this point, only if it then fails with a ConflictError.

jacobmarble (approved):

PR #5203, opened 2026-09-09
source-salesforce-native: infer window_type for untagged window_size configs

nicolaslazo:
**Description:**

`advanced.window_size` is a discriminated union tagged by `window_type`, and it is also optional (it carries a `default_factory`). The UI only fills a union's hidden discriminator for required fields, so a window size set while first configuring a capture is saved as bare `{"interval": "PT12H"}` or `{"days": 30}`, and discover fails:

```
Unable to extract tag using discriminator 'window_type' [type=union_tag_not_found, input_value={'interval': 'PT12H'}, input_type=dict]
```

Both variants fail the same way. This has been broken since #4682 turned the plain integer field into a union.

The before-validator that already normalises pre-union bare-integer configs now also infers `window_type` from an untagged payload's value key. Each variant's value key (`interval`, `days`) doubles as its tag, so whichever key is present selects the variant.

**Notes for reviewers:**

Tested using a local deployment and unit tests. Both are now green.

**Checklist:**

- [x] If fixing a regression, I have linked the PR that introduced the regression: #4682
- [x] If there are user-facing behavior changes, documentation has been updated (see [documentation conventions](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing)), or the docs PR is linked here: `docs/reference/Connectors/capture-connectors/Salesforce/salesforce-native.md` is updated in this PR


Alex-Bair (approved):
LGTM.

Like I mentioned in the Slack thread, there's a UI component at play here that warrants someone from the control plane team to take a look (mostly, not including the `window_type` discriminator when users don't click between the different "Days" and "Interval" tabs).

Also while we're touching this, I wanted to mention I'm not sure how valuable this advanced window size field is any more. We introduced it to tune backfills when we were paginating backfills with date windows, but we're no longer doing that. The advanced window size field is only used in incremental queries now, and I'm not sure how useful it is in those? We shouldn't take action on this now, but something to keep in mind for the future.

PR #5206, opened 2026-09-09
source-boilerplate: Enlarge the stdout pipe and shrink the write buffer

willdonnelly:
**Description:**

We buffer up to 1MiB of output before flushing it to stdout, but the pipe on the other side of stdout only holds 64KiB by default. So the runtime sits idle while we fill our buffer, and then we sit blocked in write() while the runtime slowly drains it. The two never actually run in parallel, and the ping-pong between those states costs us throughput.

The fix is to make the pipe bigger than our buffer. We ask the kernel for a 1MiB pipe at startup and shrink our write buffer to 64kB since it's only there to amortize syscall overhead on small documents. That way there's always data in the pipe for the runtime to read while we're filling the next buffer, and a flush only blocks when there's real backpressure.

The pipe resizing only works on Linux for obvious reasons, but shouldn't cause problems on other platforms.

**Workflow steps:**

No action required.

**Notes for reviewers:**

This is the quick-and-dirty fix for the idle/backpressure ping-pong we've seen in production. I believe it should help to prevent stalls _during_ a backfill cycle but as noted by Johnny on Slack the 1MiB pipe buffer will not necessarily be enough decoupling to prevent stalls while waiting for a backfill query to begin returning data, depending on target throughput and database RTT and query execution latency.

We may want to either raise the pipe buffer size even higher (which AFAIK requires adjusting some kernel parameter to allow) or put additional concurrent ring-buffering inside the connector (probably as a replacement for the existing bufio.Writer buffering) to address that, but that's a larger piece of work.

**Checklist:**

### DIFF
--- a/source-boilerplate/boilerplate.go
+++ b/source-boilerplate/boilerplate.go
@@ -31,7 +31,15 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
-const outputWriteBuffer = 1 * 1024 * 1024 // 1MiB output buffer to amortize write syscall overhead
+const (
+	// requestedOutputPipeSize is the kernel pipe capacity we ask for on stdout.
+	requestedOutputPipeSize = 1 * 1024 * 1024
+
+	// outputWriteBuffer amortizes the cost of a write syscall across multiple documents.
+	// It needs to be smaller than the output pipe so that flushes only block due to
+	// runtime backpressure.
+	outputWriteBuffer = 64 * 1024
+)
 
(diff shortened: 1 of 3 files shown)


jgraettinger (dismissed):
LGTM

Alex-Bair (approved):
LGTM

Alex-Bair:
For discoverability purposes, this is the Slack thread mentioned in the PR description: https://estuaryworkspace.slack.com/archives/C03Q9A3GFEE/p1788904442616009

PR #5208, opened 2026-09-09
Docs: Add Impact capture reference

aeluce:
**Description:**

We didn't have a connector reference page in our docs for the Impact capture connector. This PR adds a basic reference with supported data streams, properties, and sample spec.

Docs-only change; no code updates.

**Notes for reviewers:**

It looks like the connector only supports `Brand` accounts, though the API Catalog property includes `Partner` and `Agency` options. Please correct me if my understanding is incorrect on that count, or let me know if there are other gotchas about the connector that would be good to note.


Alex-Bair (approved):
LGTM, thank you!

re: only supporting `Brand` accounts although we expose `Parter` and `Agency` options, it looks to me like your understanding is correct. I'm not sure why we're expose those options if we don't support them. We should probably remove them to avoid confusion.

PR #5209, opened 2026-09-09
source-zendesk-support-native: add eventual consistency lag to incremental streams

### LINKED ISSUE #5205: source-zendesk-support-native: incremental cursor pagination drops same-second records
## Summary

`fetch_incremental_cursor_paginated_resources` permanently drops any record whose cursor timestamp equals the checkpoint written by the preceding sweep. A customer running the image fixed by PR #5010 lost records three weeks after that fix shipped. All three members of `INCREMENTAL_CURSOR_PAGINATED_RESOURCES` share the function: `ticket_metric_events` (cursor `time`), `ticket_skips` and `ticket_activities` (cursor `updated_at`).

## What's happening

[`fetch_incremental_cursor_paginated_resources`](https://github.com/estuary/connectors/blob/main/source-zendesk-support-native/source_zendesk_support_native/api.py#L263-L307) checkpoints `last_seen_dt` as the maximum cursor value observed in the sweep, then admits only records strictly greater than the stored cursor: `if resource_dt > log_cursor`.

Zendesk stamps these timestamps at one-second resolution, and a single second routinely holds several records. If a sweep reads past second *T* and checkpoints `last_seen_dt = T`, and Zendesk later makes visible another record also stamped *T*, the next sweep requests `start_time=T`, receives that record, and discards it because `T > T` is false. The cursor only moves forward, so there is no later opportunity to pick it up.

PR #5010 was a one-line change on the mid-sweep checkpoint condition, `response.resources[0]` to `resource`, so that the decision was read from each resource rather than the first of the page. It did not touch the `resource_dt > log_cursor` gate or the checkpointed value, both of which predate it and are still on `main`.

The sibling incremental families do not lose the boundary second. The time-based export families re-request the previous sweep's `end_time` and skip records at or below it, accepting a duplicate to keep the boundary covered. `satisfaction_ratings` holds each sweep's end 61 seconds behind real time. `incremental_cursor_paginated_resources` does neither: it advances strictly past the checkpointed second.

## A reading that needs confirming

The customer's binding runs at `interval: PT1H`, and they independently report that the missing records cluster at the top of each hour. That is what this mechanism predicts, since the checkpointed second lands wherever the sweep ends. It also implies the observable loss rate tracks sweep frequency rather than data volume. Worth confirming with someone who knows this path before anyone acts on it.

## Impact

Low and steady rather than dramatic, which is what distinguishes it from #5010 and its whole dropped pages. Confirmed at one tenant, in pairs of records sharing a single second, on 2026-08-27. Any capture of these three streams has been losing boundary records for as long as it has been running, and the loss is silent — nothing in the logs or metrics marks it. [`backfill_incremental_cursor_paginated_resources`](https://github.com/estuary/connectors/blob/main/source-zendesk-support-native/source_zendesk_support_native/api.py#L313-L353) applies no `> log_cursor` filter, so already-lost records are still reachable from the source.

Note that #3911 records this connector under Fixed Cursor Lag at 61 seconds, citing the Zendesk time-parameter restriction. That holds for the other incremental families and not for these three streams, so the entry is worth qualifying.

Alex-Bair:
**Description:**

We've observed the Zendesk's API exhibit eventually consistent behavior: a record can become queryable seconds or minutes after its own timestamp, and after we've already seen records with later timestamps. Since incremental cursors are high-water marks, anything surfacing below one was filtered out and lost for good.

Incremental streams now ignore records younger than `INCREMENTAL_LAG`, and the backfill/incremental cutoffs are held back the same amount so a new binding's backfill doesn't drain into the same unsettled window.

**Notes for reviewers:**

I didn't implement eventual consistency handling for our streams based off of Zendesk's cursor-based incremental export endpoints (AKA our `users`, `tickets`, etc. streams). These are more difficult to implement a lag for - the cursor we persist is the next page cursor from the API response, not a timestamp derived from individual records, so we can't incrementally process a response's results any longer if we have to hold back emitting and checkpointing them until we know they're all below the eventual consistency horizon. Zendesk explicitly states these incremental cursor export endpoints are for capturing new changes, so optimistically I'm hoping they aren't as susceptible to eventually consistent behavior. Plus we haven't seen these endpoints exhibit eventually consistent behavior yet, so I'm opting to not add eventual consistency handling for those streams yet due to the performance & complexity costs it would entail.

**Checklist:**

### DIFF
--- a/source-zendesk-support-native/CHANGELOG.md
+++ b/source-zendesk-support-native/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2026-09-10
+
+### Changed
+- Incremental streams now trail roughly 5 minutes behind the present to combat the Zendesk Support API's eventually consistent behavior. Records still arrive on each binding's normal polling interval, with up to 5 minutes of additional delay.
+
 ## 2026-08-05
 
 ### Fixed
--- a/source-zendesk-support-native/source_zendesk_support_native/api.py
+++ b/source-zendesk-support-native/source_zendesk_support_native/api.py
@@ -45,6 +45,13 @@ MAX_SATISFACTION_RATINGS_WINDOW_SIZE = timedelta(days=30)
 MIN_PAGE_SIZE = 1
 # Zendesk errors out if a start or end time parameter is 60 seconds or less in the past. 
 TIME_PARAMETER_DELAY = timedelta(seconds=61)
+# Zendesk's API is eventually consistent: a record can become queryable seconds or
+# minutes after its timestamp, after we've already seen records with later timestamps.
+# Since incremental cursors are high-water marks, anything that surfaces below one
+# would be filtered out and lost forever. To avoid that, we never advance a cursor
+# over records younger than INCREMENTAL_LAG, giving late arrivals time to appear
+# before the cursor moves past their timestamp.
+INCREMENTAL_LAG = timedelta(minutes=5)
 
 DATETIME_STRING_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
 INCREMENTAL_TIME_EXPORT_REQ_PER_MIN_LIMIT = 10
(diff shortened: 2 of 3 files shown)


nicolaslazo (approved):
LGTM

Issue #4620, opened 2026-06-04
source-hubspot{-native}: migrate off deprecated v1 OAuth API


Alex-Bair:
HubSpot is deprecating the v1 OAuth API on February 16, 2027 ([announcement](https://developers.hubspot.com/changelog/v1-oauth-api-deprecation), [migration guide](https://developers.hubspot.com/docs/api-reference/legacy/authentication/oauth-tokens/v1/migration-guide)). Both HubSpot connectors call deprecated endpoints; after the sunset, OAuth-authenticated captures will fail at the first token refresh.

Alex-Bair:
Closed by https://github.com/estuary/connectors/pull/5213 and https://github.com/estuary/connectors/pull/5246.

PR #5213, opened 2026-09-10
source-hubspot-native-migrate-oauth

midianyoseph:
Refs #4620.

HubSpot sunsets the v1 OAuth API on **February 16, 2027** ([announcement](https://developers.hubspot.com/changelog/v1-oauth-api-deprecation), [migration guide](https://developers.hubspot.com/docs/api-reference/legacy/authentication/oauth-tokens/v1/migration-guide)). This points `source-hubspot-native` at the replacement token endpoint.

`https://api.hubapi.com/oauth/v1/token` → `https://api.hubapi.com/oauth/2026-09/token`

The replacement is a *date-based* version, not `v3` — HubSpot's changelog is titled "OAuth v3" but every path is dated. `2026-09` is the version the migration guide names, and it's what `materialize-hubspot` already uses (`materialize-hubspot/client.go:90`), so this keeps our HubSpot connectors on one version.

### Why the one-line change covers both OAuth flows

`accessTokenUrlTemplate` has two consumers:

- **Initial `authorization_code` exchange** — the control plane, not the connector. `supabase/functions/oauth/access-token.ts` reads `oauth2_spec.accessTokenUrlTemplate` off the `connectors` table, which is populated from the connector's spec response. So this only takes effect once the connector is published.
- **Ongoing `refresh_token` exchange** — the connector at runtime, via `estuary-cdk/estuary_cdk/http.py:441`.

The 2026-09 token endpoint accepts the same form-encoded body and returns the same fields, so `accessTokenBody`, `accessTokenHeaders`, and `accessTokenResponseMap` are unchanged.

### DIFF
--- a/source-hubspot-native/CHANGELOG.md
+++ b/source-hubspot-native/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 2026-09-10
+### Changed
+- OAuth token exchanges now use HubSpot's date-based `2026-03` OAuth API instead of the
+  deprecated v1 API, which HubSpot sunsets on February 16, 2027. Without this, OAuth
+  captures would fail at their first access token refresh after the sunset. Existing
+  captures keep working and do not need to re-authorize; refresh tokens remain valid.
+
 ## 2026-09-02
 ### Fixed
 - CRM object bindings no longer fail when the HubSpot Search API returns records out
--- a/source-hubspot-native/source_hubspot_native/models.py
+++ b/source-hubspot-native/source_hubspot_native/models.py
@@ -73,7 +73,7 @@ OAUTH2_SPEC = OAuth2Spec(
         + r"&redirect_uri={{#urlencode}}{{{ redirect_uri }}}{{/urlencode}}"
         r"&response_type=code&state={{#urlencode}}{{{ state }}}{{/urlencode}}"
     ),
-    accessTokenUrlTemplate="https://api.hubapi.com/oauth/v1/token",
+    accessTokenUrlTemplate="https://api.hubapi.com/oauth/2026-03/token",
     accessTokenHeaders={"content-type": "application/x-www-form-urlencoded"},
     accessTokenBody=(
         "grant_type=authorization_code"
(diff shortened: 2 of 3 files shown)


source-hubspot-native/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 2026-09-10
+### Changed
+- OAuth token exchanges now use HubSpot's date-based `2026-03` OAuth API instead of the
+  deprecated v1 API, which HubSpot sunsets on February 16, 2027. Without this, OAuth
+  captures would fail at their first access token refresh after the sunset. Existing
+  captures keep working and do not need to re-authorize; refresh tokens remain valid.
```
Alex-Bair, line 8: nit: Claude loves to include comments like "this horrible thing would have happened if we didn't do XYZ!". I don't think we necessarily need to include that for this specific changelog entry, mostly because we're making this change well ahead of the 16FEB2027 sunset date. No need to change it for this one, but something to keep in mind in the future.

source-hubspot-native/source_hubspot_native/models.py
```diff
@@ -73,7 +73,7 @@
         + r"&redirect_uri={{#urlencode}}{{{ redirect_uri }}}{{/urlencode}}"
         r"&response_type=code&state={{#urlencode}}{{{ state }}}{{/urlencode}}"
     ),
-    accessTokenUrlTemplate="https://api.hubapi.com/oauth/v1/token",
+    accessTokenUrlTemplate="https://api.hubapi.com/oauth/2026-03/token",
```
Alex-Bair, line 76: nit: Would we be able to use the `2026-09` version instead? When bumping API versions we use, we usually try to go to the most recent, stable version that has no relevant breaking changes. That makes it so we have to bump versions less frequently, like if/when HubSpot sunsets the `2026-03` API version at some point. Based off the [docs](https://developers.hubspot.com/docs/api-reference/latest/authentication/manage-oauth-tokens) & a couple API requests I made, it looks like the access token response for the `2026-09` version is the same as `v1` and `2026-03`, so at a glance it looks like we could go all the way to `2026-09`.

Alex-Bair (dismissed):
LGTM, thanks! I had a few small nits, but nothing blocking.

source-hubspot-native/source_hubspot_native/models.py
midianyoseph: Yes makes sense, done!

Alex-Bair (approved):
LGTM, thank you! Please make sure to use the "Squash and merge" option when merging this PR so a single commit lands on `main` without the following commits that address PR comments.

PR #5215, opened 2026-09-10
Don't cancel in progress builds of main

danielnelson:
We only trigger a connector spec update for connectors that have changed in this build, so its possible a connector won't be updated if the build of main is skipped.

Builds of main should be serialized due to the concurrency group, so with this change it should do the builds of main one at a time.

**Checklist:**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,7 +2,9 @@ name: CI
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  # Don't cancel in progress builds of main, since that can result in connector
+  # specs not being updated.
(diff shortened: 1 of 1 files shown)


jacobmarble (approved):

Alex-Bair (commented):
LGTM.

re: the motivating problem of our merges to `main` trampling on each other - I think what would really help all of us is more granular CI. Like right now, making a change in `materialize-snowflake` causes `source-postgres`' CI to run all the way through. I think it'd be great if we could make it so only `source-postgres` changes (and any dependencies it relies on) trigger `source-postgres`' CI to run. That'd require a lot of time to wire up though.

danielnelson:
Agreed, I have thought about trying that change out before.  It is a bit to setup initially but I think the maintenance isn't that bad, and we have a lot of this information in the Dockerfiles so we may not need to determine it all from scratch.

PR #5217, opened 2026-09-11
source-sage-intacct: skip deletions whose AUDITHISTORY OBJECTKEY is not a RECORDNO

### LINKED ISSUE #5207: source-sage-intacct: deletions subtask crashes the connector on a non-numeric AUDITHISTORY OBJECTKEY
## 1. Priority
Medium - a single unparseable deletion record crash-loops the entire capture task, not just the
affected binding. A workaround exists (remove the binding) but it costs the whole binding, and
the deletions cursor does not advance, so re-enabling reproduces the crash.

## 2. Scope and prevalence
Theoretical blast radius: `DeletionRecord.parse_object_key` is shared by every `source-sage-intacct`
deletions subtask, so any Sage object whose `AUDITHISTORY.OBJECTKEY` is a text business key rather
than a numeric `RECORDNO` can trigger it. Sage objects with user-defined IDs (VENDOR, CUSTOMER,
EMPLOYEE, ITEM, PROJECT and similar) are the candidates. Amplification: the failure is raised out
of `_fetch_records_generic`, exits the connector with status 1 and fails the shard, so all bindings
on the task stop, not just the one that produced the record.

Observed prevalence: one tenant, one binding (VENDOR), first occurrence support has seen. Trigger
is a real delete of a text-keyed object in Sage, which is uncommon per tenant but not rare in
aggregate. 48 shard failures over 9.4h before the binding was removed.

## 3. Trigger (testable)
An `AUDITHISTORY` row with `ACCESSMODE = D` for an object whose `OBJECTKEY` is neither a
stringified positive integer nor `<int>--REC`. `DeletionRecord.parse_object_key` raises, pydantic
wraps it as a `ValidationError`, and the connector exits.

Minimal repro input (synthetic, NO customer data): feed `DeletionRecord` a record shaped like

    {"OBJECTKEY": "V10001", "ACCESSTIME": "09/09/2026 04:59:18", "ID": "1234567"}

Any non-numeric, non-`--REC` `OBJECTKEY` reproduces it. Reference values that must keep working:
`"12345"` and `"12345--REC"`.

## 4. Root cause (if known) + confidence
Confirmed for the validator, likely for the source of the value.

Confirmed: `source-sage-intacct/source_sage_intacct/models.py`, `DeletionRecord.parse_object_key`
handles exactly two `OBJECTKEY` shapes - a stringified positive int, and `<int>--REC` - and raises
`ValueError` on anything else. `DeletionRecord` declares
`RECORDNO: int = Field(alias="OBJECTKEY", serialization_alias="RECORDNO")`, so `OBJECTKEY` is
being used directly as the document key.

Likely: Sage populates `AUDITHISTORY.OBJECTKEY` with the object's user-facing ID (for VENDOR, the
`VENDORID`) rather than its `RECORDNO`, producing a third shape the validator does not model. Not
confirmed against Sage documentation or the tenant's data.

Consequence for the fix: because `RECORDNO` is the collection key, loosening the validator alone is
not sufficient. A text `OBJECTKEY` has to be resolved to a `RECORDNO` (or the deletion cannot be
keyed to the row it deletes).

1. Read the task log export covering `2026-09-07T16:58:33.018295115Z` to
   `2026-09-09T14:36:18.571168668Z`. Two clean daily runs on 09-07 and 09-08; first failure on
   09-09. Error, customer identifier scrubbed:

```
2026-09-09T04:59:19.331384567Z  Task capture.VENDOR.incremental.realtime_deletions: 1 validation error for DeletionRecord
2026-09-09T04:59:19.331384567Z  OBJECTKEY
2026-09-09T04:59:19.331384567Z    Value error, OBJECTKEY must be an integer or in format 'RECORDNO--REC', got <vendor-id> [type=value_error, input_value='<vendor-id>', input_type=str]
```

2. Traced the raising frame:

```
2026-09-09T04:59:18.126675955Z    File "/opt/source-sage-intacct/source_sage_intacct/api.py", line 61, in _fetch_records_generic
2026-09-09T04:59:18.126675955Z      rec = cls.model_validate(rec.model_dump())
2026-09-09T04:59:18.126675955Z  pydantic_core._pydantic_core.ValidationError: 1 validation error for DeletionRecord
```

3. Checked propagation -> the error leaves the subtask and terminates the process:

```
2026-09-09T04:59:19.529467821Z  connector failed   (module connector_init::rpc, status "exit status: 1")
2026-09-09T04:59:19.534133787Z  shard failed       (eventType shardFailure)
2026-09-09T04:59:19.534133787Z  runTransactions: readMessage: Capture error (expected Captured, SourcedSchema, Checkpoint, BackfillBegin, or BackfillComplete) from connector
```

4. Counted occurrences -> 48 `shard failed` between `2026-09-09T04:59:19.534133787Z` and
   `2026-09-09T14:22:58.324885922Z`, restarting roughly every 12 minutes. 240 validation-error
   lines, all carrying one single distinct `OBJECTKEY` value.

5. Checked whether it is object-specific -> only `capture.VENDOR.incremental.realtime_deletions`
   (69) and `capture.VENDOR.incremental.lookback_deletions` (75) appear. The other three bindings'
   deletions subtasks resume each restart on the same cursor with no error.

6. Checked whether a publish triggered it -> no. First failure occurred under the same build that
   had run clean for two days; the build only changed afterwards, at
   `2026-09-09T13:44:21.972139205Z` and again at `2026-09-09T14:35:52.064664726Z`.

7. Checked the deletions cursor -> unchanged at `2026-07-21T20:05:27.314727Z` on every restart, so
   the failing audit row is re-read on each attempt and no forward progress is possible.

8. Read `parse_object_key` in `models.py` and `get_deletions_since_request` in `sage.py` on
   `estuary/connectors` main -> confirms the two accepted shapes and the `AUDITHISTORY` query.

9. Observed the workaround -> at `2026-09-09T14:35:52.064664726Z` the task came up on a build with
   the affected binding removed (`Streaming change events (all 3 bindings are backfilled)`, down
   from 4). No errors after that point.

Offered to save eng time, not a spec. The fix is eng's call.

1. Customer-immediate stopgap (in place, ops action): remove the affected binding. Cost: that
   object stops capturing entirely, and deletions for it during the gap are lost. Re-enabling
   before a fix will crash again, since the deletions cursor has not advanced past the offending
   audit row.
2. Data-preserving option: none available today at the config level - the deletions subtask cannot
   be disabled independently of its binding, so there is no way to keep the object's inserts and
   updates flowing while skipping deletions.
3. Platform / code direction: two separable problems.
   - `parse_object_key` (`source-sage-intacct/source_sage_intacct/models.py`) models two `OBJECTKEY`
     shapes; `AUDITHISTORY` appears to return a third. Because `RECORDNO` is the document key,
     accepting the text key needs it resolved to a `RECORDNO` (a lookup on the object by its ID
     field), otherwise the emitted `op: d` will not join to the row it is meant to delete.
   - Independently, one bad record from one subtask takes down every binding on the task
     (`_fetch_records_generic` -> `cls.model_validate` -> uncaught). Worth deciding whether a
     record-level validation failure should be fatal to the whole connector.

Any code fix is drafted for human review, not self-merged or published - changes affect live tenants.

## 7. Definition of done
- Acceptance: a deletion whose `AUDITHISTORY.OBJECTKEY` is a text business key is either captured
  as an `op: d` document keyed to the correct `RECORDNO`, or explicitly skipped with a warning;
  the task runs a full interval with no `shard failed` and the deletions cursor advances past the
  offending `ACCESSTIME`.
- Regression test: exercise `parse_object_key` / `DeletionRecord` with all three shapes -
  `"12345"`, `"12345--REC"`, and a text key such as `"V10001"` - asserting the first two are
  unchanged and the third no longer raises.
- Constraints / do not break: do not regress the two currently-accepted `OBJECTKEY` shapes; do not
  silently drop deletions (a dropped `op: d` is undetectable data divergence in the destination);
  do not require a re-backfill of unaffected bindings; if the text key cannot be resolved to a
  `RECORDNO`, fail loudly on that record rather than emitting a mis-keyed delete.

## 8. References
- Slack thread: pending, to be backfilled

nicolaslazo:
Fixes #5207

**Description:**

The connector learns of deletions from Sage's audit log. Each delete event
carries an `OBJECTKEY`, which the connector copied into the tombstone's
`RECORDNO`, accepting only an integer or an integer suffixed `--REC`. That
assumption was never true in general. Sage defines `OBJECTKEY` as the object's
"unique name field, which differs based on the object type". For a vendor it is
`VENDORID`; for a customer, `CUSTOMERID`. The first vendor deletion on a tenant
therefore failed to parse and the cursor never moved.

Such an event cannot be turned into a delete. Flow matches tombstones by the
collection key alone, and the collection is keyed on `RECORDNO`. The record is
already gone, so its number cannot be looked up. The connector now writes one
debug line and moves on.

The audit row and the tombstone are now separate models. `DeletionEvent` holds
the row (`OBJECTKEY`, `ACCESSTIME`, `ID`) and is what the walker iterates.
`DeletionRecord.from_event` builds the tombstone and raises
`ObjectKeyNotRecordNo` when it cannot. The document on the wire is unchanged:
`{"_meta": {"op": "d"}, "RECORDNO": …, "WHENMODIFIED": …}`.

**Notes for reviewers:**

No test credentials were available for testing.

**Checklist:**

- [ ] If fixing a regression, I have linked the PR that introduced the regression: not a regression
- [x] If there are user-visible changes, `CHANGELOG.md` has been updated
- [x] If there are user-facing behavior changes, documentation has been updated: `docs/reference/Connectors/capture-connectors/sage-intacct.md`


source-sage-intacct/source_sage_intacct/models.py
```diff
@@ -235,56 +234,51 @@ def parse_backfill_record(raw: dict) -> "IncrementalResource | CreationRecord":
...
-                raise ValueError(
-                    f"RECORDNO in OBJECTKEY must be positive, got {record_no}"
-                )
-            return record_no
-        except ValueError:
-            raise ValueError(
-                f"RECORDNO in OBJECTKEY must be an integer, got {parts[0]}"
-            )
+    @classmethod
+    def try_from_event(cls, event: DeletionEvent) -> "DeletionRecord":
```
Alex-Bair: nit: the `try`/`except` is outside this method, so I think naming it `from_event` is clearer than `try_from_event`.

Alex-Bair (dismissed):
LGTM % one small nit.

source-sage-intacct/source_sage_intacct/models.py
nicolaslazo: Old Rust habits, hehe. It's been addressed

Alex-Bair (approved):
LGTM!

PR #5221, opened 2026-09-11
Docs: Reformat connector lists in reference docs

aeluce:
**Description:**

We've been using the same list style for our capture and materialization connector landing pages for a while now and it no longer really works for us.

One issue is that each connector includes the image name and tag. Whenever the default connector tag gets updated, the list needs to be updated as well. The detail doesn't really make sense to include on these pages; it's not presented in the order someone using the product would actually need it, and connector pages should generally be providing this as part of the spec example.

The other issue is that the list, as-is, is mostly redundant. The sidebar also lists all the available connectors. The landing page should provide unique information that a user would find helpful.

To address these issues, I've:
* Streamlined the connector lists: removing the extraneous image/tag info as well as removing deprecated connectors from the list for a cleaner experience.
* Grouped connectors by category. So if someone is particularly interested in database connectors, they can specifically look at that section, or commerce/payment SaaS connectors or whatever. This provides an at-a-glance spread of our different integration options.
* There was also a slight side-quest to add spec examples to Dremio and Bauplan pages to ensure they had their image names and tags.

For example, currently-served version is at the left, proposed changes on the right:

connector-landing-page-lists

Docs-only change; no code updates.

**Notes for reviewers:**

Looking for feedback from the integrations team. Does this seem helpful? Any reservations about maintaining category lists? And please let me know if you have any thoughts about the categories themselves, or if you think any connectors should be categorized elsewhere.

Thanks for reviewing!

### DIFF
--- a/docs/reference/Connectors/capture-connectors/README.md
+++ b/docs/reference/Connectors/capture-connectors/README.md
@@ -18,303 +18,105 @@ These connectors are created by Estuary. We prioritize high-scale technology sys
 
 All Estuary connectors capture data in real time, as it appears in the source system.
 
-- Ada
-  - [Configuration](./ada.md)
-  - Package - ghcr.io/estuary/source-ada:v1
-- Airtable
-  - [Configuration](./airtable-native.md)
-  - Package - ghcr.io/estuary/source-airtable-native:v1
-- AlloyDB
-  - [Configuration](./alloydb.md)
-  - Package - ghcr.io/estuary/source-alloydb:v3
-- Alpaca
-  - [Configuration](./alpaca.md)
-  - Package - ghcr.io/estuary/source-alpaca:v1
-- Amazon DocumentDB
-  - [Configuration](./MongoDB/amazon-documentdb.md)
-  - Package - ghcr.io/estuary/source-amazon-documentdb:v3
-- Amazon Dynamodb
-  - [Configuration](./amazon-dynamodb.md)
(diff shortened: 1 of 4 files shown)


docs/reference/Connectors/capture-connectors/README.md
```diff
@@ -4,335 +4,161 @@ description: Browse Estuary's list of capture connectors for databases, SaaS app
...
+- [Azure Blob Storage](./azure-blob-storage.md)
+- [Dropbox](./dropbox.md)
+- [Google Cloud Storage](./gcs.md)
+- [Google Drive](./google-drive.md)
+- [Google Sheets](./google-sheets.md)
+- [HTTP file](./http-file.md)
+- [OneDrive](./onedrive.md)
+- [SFTP](./sftp.md)
+- [SharePoint](./sharepoint.md)
+- [Smartsheet](./smartsheet.md)
```
Alex-Bair: I'd move Smartsheet into a SaaS connector category (maybe the Other Applications bucket?). It's built with a different framework than these other `filesource` capture connectors.

Alex-Bair (dismissed):
LGTM % moving the Smartsheet capture to a different category.

Thanks for doing this! I think these categorizations make more sense than what we had before.

aeluce:
Got it; thanks! Moved Smartsheet (and added PostHog, which was missing from the list).

Alex-Bair (approved):
LGTM, thanks!

PR #5228, opened 2026-09-14
source-postgres: Ignore stray STREAM_ABORT messages

willdonnelly:
**Description:**

We use logical replication protocol v1 (which doesn't support streaming of in-progress transactions) and also we never request the feature. Nonetheless, we have seen databases in the wild sending us unsolicited `LOGICAL_REP_MSG_STREAM_ABORT` messages.

This appears to be a bug in PostgreSQL 18, specifically around how subtransactions are tracked so that streaming aborts can be sent, which causes a subtransaction to be incorrectly marked as streaming even though we're not even using the feature, so then when it's aborted that abort gets sent to us without any further checks.

The easiest solution here is to simply parse and discard those messages, so that's what we'll do.

**Checklist:**


Alex-Bair (approved):
LGTM, thanks for taking care of this!

PR #5233, opened 2026-09-15
materialize-iceberg: variant columns behind the variant_columns feature flag

### LINKED ISSUE #5226: materialize-iceberg: VARIANT columns as a cutoff-gated feature flag
## Problem Statement

materialize-iceberg stores every object, array, multi-type field, and the root document as a JSON string column. Users querying those tables from Spark, Snowflake, Dremio, Athena, or DuckDB have to `parse_json` on every read, lose column statistics and file pruning on the nested data, and cannot use engine-native semi-structured operators. materialize-s3-iceberg already offers Iceberg format v3 `variant` columns for these shapes; materialize-iceberg (and its Dremio and Bauplan image variants) does not, so users of a REST, Glue, S3 Tables, Dremio, or Bauplan catalog have no way to get natively typed semi-structured columns.

## Solution

New materialize-iceberg tasks create tables as Iceberg format v3 and materialize objects, arrays, multi-type fields, and the root document as `variant` columns. Existing tasks keep their JSON string columns unless support opts them in. The behavior is a cutoff-gated feature flag, `variant_columns`, resolved from the task's creation date and recorded into the task's config, following the mechanism introduced by PR #4646. Because the Spark job runs on the customer's own EMR Serverless application, the connector checks at publish time that the application runs Spark 4 (`emr-spark-8.0.0` or later) and fails with an actionable message otherwise. Support can override per task with `no_variant_columns`, and existing tables convert in either direction through the connector's data-preserving column migration.

## User Stories

1. As a data engineer creating a new materialize-iceberg task, I want object and array fields written as `variant` columns, so that I can query nested data natively without `parse_json` on every read.
2. As a data engineer creating a new task, I want the root document (`flow_document`) written as `variant`, so that ad-hoc exploration of the full record uses native operators.
3. As a data engineer creating a new task, I want multi-type fields written as `variant`, so that a field that is sometimes a string and sometimes an object is preserved with its type per row.
4. As a data engineer, I want tables created by a new task to be Iceberg format v3, so that `variant` columns are valid and v3-capable engines can read them.
5. As a data engineer, I want collection key fields to remain string columns even with variant enabled, so that joins and filters on keys work across all engines and the Iceberg identifier-field rule is respected.
6. As a data engineer, I want the per-field `castToString` option to force a JSON string column even when variant is on, so that I can keep a specific field readable by a v2-only consumer.
7. As a data engineer, I want string-encoded numbers (`string` plus `integer`/`number` with a matching `format`) to keep their numeric column, so that variant does not regress numeric typing.
8. As a data engineer, I want a top-level `format: date-time` string value inside a variant field stored as a variant timestamp, so that it is typed the same way as it would be as a column of its own.
9. As a data engineer, I want a top-level `format: date` value inside a variant stored as a variant date, and base64 binary content stored as variant binary, for the same reason.
10. As a data engineer, I want a value that fails to parse as its annotated type stored as a plain string inside the variant, so that a bad value never fails the merge.
11. As a data engineer, I want non-string values in a format-annotated multi-type field left as their JSON type inside the variant, so that typing only applies where it makes sense.
12. As a data engineer, I want values nested inside objects, arrays, and the root document stored as plain JSON types inside the variant, matching materialize-s3-iceberg, so that both Iceberg connectors behave the same.
13. As the owner of an existing materialize-iceberg task, I want my tables and columns to stay exactly as they are after upgrading the connector, so that downstream readers and dashboards are undisturbed.
14. As the owner of an existing task, I want to be able to opt in by setting `variant_columns` in the task's feature flags, so that I can adopt variant when my readers are ready.
15. As the owner of an existing task opting in, I want existing string columns converted to `variant` with their existing rows preserved, so that I do not have to backfill.
16. As the owner of an existing v2 table opting in, I want the table upgraded to format v3 in the same atomic commit as the schema change, so that the table is never in an invalid intermediate state.
17. As the owner of a new task whose downstream engine cannot read variant, I want to set `no_variant_columns` and have the task use JSON strings, so that I can keep my current readers.
18. As the owner of a task that has variant columns, I want to be able to disable the flag and have those columns converted back to JSON strings with rows preserved, so that there is an exit path.
19. As a user pointing a new task at a table that already exists with string columns, I want the connector to migrate them to variant rather than reject the binding, so that existing tables can be reused.
20. As a user creating a new task against an EMR Serverless application on an EMR 7.x release, I want publish-time validation to fail with a message telling me to upgrade the application to `emr-spark-8.0.0` or set `no_variant_columns`, so that I do not discover the problem hours later as a failed background merge.
21. As a user whose EMR execution role lacks `emr-serverless:GetApplication`, I want validation to log a warning and proceed rather than fail, so that a missing read-only permission does not block me, matching how the connector already handles `TagResource`.
22. As a user of a catalog that rejects format v3 or variant, I want the table creation or alteration error to name `no_variant_columns` as the remedy, so that I know how to unblock the task without contacting support.
23. As a user who set the `format-version` table property to something other than 3 on a binding, I want validation to reject the conflict with variant explicitly, so that the connector never silently overrides my property.
24. As a user materializing with delta updates, I want variant columns to work the same way as in standard updates, so that both modes are consistent.
25. As a user loading documents for standard updates, I want the load path to return the stored document as JSON text, so that the runtime reduces it correctly regardless of the column type.
26. As a user relying on hard deletes, I want the delete sentinel detection to keep working when the document column is variant, so that deletions are applied.
27. As a support engineer, I want a task's resolved `variant_columns` value written into its config at Open, so that I can read a task's behavior off its spec instead of inferring it from its creation date.
28. As a support engineer, I want an explicit `variant_columns` or `no_variant_columns` in `/advanced/feature_flags` to take priority over the cutoff, so that I can override either way per task.
29. As a Dremio Cloud or Bauplan connector user, I want the same behavior as materialize-iceberg, so that the image variants stay identical to their parent.
30. As a connector developer, I want the local docker-compose stack to run both a Spark 3.5 and a Spark 4.0 set against a shared Polaris and RustFS, so that the frozen string path and the new variant path are both exercised by CI.
31. As a connector developer, I want the existing materialize, apply, and migrate scenarios run against both Spark sets, so that a change to shared SQL generation cannot silently break one of them.
32. As a connector developer, I want variant-only scenarios (v3 creation, both migration directions, typed variant values, load through JSON) to run against the Spark 4.0 set only, so that the suite does not attempt the impossible on Spark 3.5.
33. As a connector developer, I want the test harness to snapshot table contents through the local Spark daemon, so that variant data can be verified rather than only the protocol interaction.
34. As a connector developer, I want the cloud test configurations pointed at an `emr-spark-8.0.0` application, with one configuration left on the 7.x application with `no_variant_columns` pinned, so that both the AWS-side variant write path and the frozen path have cloud coverage.
35. As a connector developer, I want proof that Iceberg 1.10.1 on Spark 4.0.2 (the exact EMR 8.0.0 pairing) can write variant through `MERGE INTO` before any connector code is written, so that the feature is not built on an unshippable dependency.
36. As a release manager, I want the cutoff date set to the day the connector image ships, so that tasks created on that day get variant from birth and none created earlier are touched.
37. As a reader of the changelog, I want the variant behavior, the Spark 4 requirement, the override, the migration semantics, and the divergence from materialize-s3-iceberg spelled out, so that I can decide whether to opt in or out.

### Feature flag

- `variant_columns` is a materialization feature flag, not an advanced config field. Its default becomes `common.FlagEnabledForTasksCreatedAfter(<ship date>)` once PR #4646 (automatic feature flags setting for materializations) has merged. Until then the branch is built on main with a fixed disabled default so it builds and tests in isolation; the final commit, after rebasing onto the merged mechanism, switches the entry to the cutoff form with the date set to the day the image ships.
- Resolution stays a pure function of the feature flag string and the task's `created_at`. The connector never downgrades or auto-detects at runtime; the `configUpdate` record emitted at Open is therefore always truthful.
- The flag applies uniformly to the Dremio and Bauplan image variants; the defaults map is not keyed on connector name.

### Type mapping

- With the flag on: object, array (including single-type arrays, retiring the existing "arrays as strings" TODO), multi-type, and the root document projections map to `variant`. Collection key projections keep their current mapping. A projection with `castToString` maps to string. String-encoded numbers keep their numeric mapping. Everything else is unchanged.
- With the flag off the mapping is byte-for-byte what it is today.
- The mapped type records an optional variant format hint, one of `date-time`, `date`, or `binary`, taken from the projection's string inference (format, or base64 content encoding with the binary media type) when the projection maps to variant. Object, array, and root-document projections carry no hint.
- Allowed migrations gain string→variant and variant→string. Because the connector already migrates through a temporary column populated by a Spark `UPDATE` and then swapped in atomically, both directions preserve existing rows. This is a deliberate improvement over materialize-s3-iceberg, which drops and re-adds the column.

### Catalog

- Table creation with any variant column sets the `format-version` table property to 3. A user-supplied `format-version` other than 3 in the binding's additional table properties is rejected at validation when variant is on, mirroring materialize-s3-iceberg.
- The REST catalog client gains an `upgrade-format-version` table update. When an update introduces a variant column into a table whose metadata is below v3, the upgrade is included in the same `UpdateTable` request as the add-schema and set-current-schema updates, under the existing assert-current-schema-id requirement.
- Schema serialization goes through iceberg-go, whose `VariantType` is available at the pinned version. The custom REST client does not use iceberg-go's `UpdateSchema` builder, so the placeholder workaround materialize-s3-iceberg needs does not apply.
- Errors from `CreateTable` and `UpdateTable` are wrapped with a hint naming `no_variant_columns` whenever the request carried format version 3 or a variant column. No catalog capability probing is performed.

### EMR preflight

- Validate calls `emr-serverless:GetApplication` for the configured application when the flag resolves on and the compute type is EMR Serverless. A release label below `emr-spark-8` fails validation with a message instructing the user to move the application to `emr-spark-8.0.0` or later, or to set `no_variant_columns`. An access-denied response logs a warning and proceeds, matching the existing `TagResource` fallback. The local Spark Standalone compute type performs no preflight.
- Validate resolves the flag from the last materialization spec's `created_at`; a brand new task has none and resolves on, so the preflight runs for every new task.

### Spark job

- The staged CSV format is unchanged: variant columns are carried as JSON text exactly as string columns are today.
- The Python `NestedField` contract gains an optional `variant_format` field. The Go side emits `type: "variant"` plus the hint for each variant column in the per-binding column and key lists.
- In `read_csv_opts`, a variant column is read as a Spark string. Before the temp view is created, the job replaces each variant column with a variant expression:

```python
v = f"parse_json({c})"
s = f"try_variant_get({v}, '$', 'string')"
{
  None:        v,
  "date-time": f"coalesce(cast(try_to_timestamp({s}) as variant), {v})",
  "date":      f"coalesce(cast(try_to_date({s}) as variant), {v})",
  "binary":    f"coalesce(cast(unbase64({s}) as variant), {v})",  # applied only when s matches a base64 regexp
}
```

  `try_variant_get` is null for non-string values and `try_to_*` is null on parse failure, so `coalesce` falls back to the untyped variant. This reproduces materialize-s3-iceberg's typing rule (typed when it parses, string otherwise, non-strings untouched, top level only) without touching the SQL templates. Spark 4 casts primitives to variant directly. Timestamps are microsecond precision, consistent with this connector's `timestamptz` columns; there is no nanosecond option here.
- The merge template is unchanged: the right-hand side view already yields variant, and the delete sentinel comparison is against the raw CSV document text on the right-hand side, which stays a string.
- The load template wraps the document column in `to_json` when it is variant, so loaded documents return as JSON text.
- The migration cast expression becomes `parse_json` for string→variant and `to_json` for variant→string.
- The load and merge scripts remain valid on Spark 3.5 whenever no variant column is present; no Spark 4 syntax is introduced into the non-variant path.

### Local test stack

- The compose file keeps the existing Spark 3.5.5 + Iceberg 1.6.1 master, worker, and daemon, and adds a second master, worker, and daemon on Spark 4.0.2 + `iceberg-spark-runtime-4.0` 1.10.1 with a distinct daemon port, both sharing Polaris and RustFS. Two local configs are rendered from templates. The PySpark dependency in the Python project moves to 4.0.
- The 3.5 config pins `no_variant_columns`; the 4.0 config leaves the flag to resolve, which for test specs without `created_at` means on.
- The test-only Spark daemon gains the ability to return rows for a query, so the connector's `SnapshotTestResource` can be implemented for the local compute type by selecting the table through the daemon with variant columns rendered via `to_json`.

### Cloud test configuration

- A new EMR Serverless application on `emr-spark-8.0.0` is created in the test account by a human; the cloud configs are repointed to it with `sops set`. One config stays on the existing 7.x application with `no_variant_columns` pinned so the frozen path retains cloud coverage. Connector code and local-stack work do not depend on this.

### Step zero

- Before connector code: bring up the Spark 4.0.2 + Iceberg 1.10.1 set and run a hand-written `MERGE INTO` with `parse_json` into a v3 table with a variant column, then read it back. If Iceberg 1.10.1 cannot write variant from Spark, the feature is blocked on AWS shipping a newer Iceberg and work stops with that finding recorded on this issue.

## Testing Decisions

- A good test exercises the connector through the materialization protocol and observes the destination table, never internal structs. Table contents and Iceberg schema, including format version and column types, are the external behavior.
- Primary seam: the boilerplate materialization harness driven by `flow.yaml` specs under testdata, run against both local daemons. `RunMaterializationTest` and `RunApplyTest` cover v3 creation with variant columns, merges, loads, and deletes. `RunFeatureFlagMigrationTest` (prior art: materialize-bigquery's JSON-toggle test) covers string→variant and variant→string across phases, snapshotting the apply action and the resulting table so row preservation is verified. Typed variant values are verified through a spec with a format-annotated multi-type field and a base64 field.
- The one enabling test-only change is the daemon returning rows so that `SnapshotTestResource` works for the local compute type.
- Existing scenarios run on both Spark sets; variant scenarios on 4.0 only. Snapshots are kept per Spark set where the output differs.
- Unit seam: `mapProjection` table tests for the mapping rule with the flag on and off, key fields, `castToString`, string-encoded numbers, and the emitted format hint. Prior art: the existing type mapping tests in this connector and materialize-s3-iceberg's projection tests.
- Unit seam: the release-label preflight against a fake EMR client covering below-8, at-or-above-8, and access-denied.
- Cloud runs, when executed with the full matrix flag, exercise the EMR 8 application and the pinned 7.x application.

## Out of Scope

- Nanosecond timestamps; this connector has no such option.
- Typing values nested inside objects, arrays, or the root document; they remain plain JSON types inside the variant, as in materialize-s3-iceberg.
- Converting materialize-s3-iceberg's `variant_columns` advanced config option into a cutoff feature flag.
- Shredded variant writes or any tuning of Spark's variant shredding.
- Backfilling existing rows of a table that was created before opt-in; the migration preserves rows, so no backfill is needed.
- Catalog capability detection or an allowlist of catalogs known to support v3.
- Per-image defaults for the Dremio and Bauplan variants.
- CI coverage of the Spark 3.5 path on AWS beyond the single pinned cloud config.

## Further Notes

- EMR `emr-spark-8.0.0` ships Spark 4.0.2 with Iceberg 1.10.1. Several Iceberg variant-on-Spark-4 items landed in 1.11.0 (a Spark round-trip test, a `VariantLogicalType` NPE fix, `ParquetTypeVisitor` support), which is why step zero is a hard gate.
- Catalog support as of this writing: Glue REST supports v3; S3 Tables supports variant since July 2026 in 15 regions; Dremio Cloud since April 2026; Bauplan unverified. Polaris 1.4 in the local stack must accept `format-version: 3` at create and `upgrade-format-version` at update; this is verified as part of step zero.
- PR #4646 notes that release of the cutoff mechanism must wait until every enabled task carries `created_at`; that constraint governs when the final flip commit can ship, not when this branch can be developed.
- The changelog entry states the divergence from materialize-s3-iceberg: migrations preserve rows here, and typed variant values are produced by the Spark job rather than the Parquet writer.
- Grilling session and decisions recorded on 2026-09-13.

mdibaiee:
**Description:**

Closes #5226. Objects, arrays, multi-type fields, and the root document materialize as Iceberg format v3 `variant` columns when the `variant_columns` feature flag is on. Tables are created as (or upgraded in-place to) v3, columns migrate in either direction preserving rows, and Validate checks that the EMR Serverless application runs Spark 4.

The flag defaults to off; a follow-up commit switches it to a cutoff default once #4646 has merged.

**Workflow steps:**

Set `variant_columns` in the endpoint's advanced feature flags (or `no_variant_columns` to keep JSON strings). The EMR application must run release `emr-spark-8.0.0` or later.

**Notes for reviewers:**

- Any column may migrate to variant (not just string): a scalar field evolving to multi-type would otherwise become INCOMPATIBLE, a regression from the existing any→string path.
- The `format-version` property conflict is rejected at Apply rather than Validate, since Validate has no per-binding hook that sees the resource config with resolved flags.
- The local stack now runs a second Spark 4.0.2 + Iceberg 1.10.1 set; existing scenarios run against both, variant scenarios against Spark 4 only. Cloud configs are untouched until an EMR 8 application exists.

**Checklist:**


materialize-boilerplate/type_mapping.go
```diff
@@ -129,6 +129,12 @@ func (m TypeMigrations[T]) CanMigrate(from string, to T) bool {
 	return false
 }
 
+// MapProjection lifts a pf.Projection into a Projection with its FlatType
+// resolved, for connector unit tests of their type mapping.
+func MapProjection(p pf.Projection, fc FieldConfiger) Projection {
+	return mapProjection(p, fc)
+}
+
```
mdibaiee: Instead of this, just expose `mapProjection` as public

materialize-iceberg/type_mapping.go
```diff
@@ -45,48 +53,63 @@ func (m mapped) CanMigrate(existing boilerplate.ExistingField) bool {
...
+func jsonColumnType(variant bool) iceberg.Type {
+	if variant {
+		return iceberg.VariantType{}
+	}
+	return iceberg.StringType{}
+}
+
+// variantFormatHint is the type a top-level string value of a variant column
+// is stored as, from the projection's string inference: the same resolution a
+// string column of its own would get, limited to the types a variant can hold.
```
mdibaiee:
```suggestion
// variantFormatHint is the type a top-level string value of a variant column
// is stored as, from the projection's string inference.
```

materialize-iceberg/type_mapping.go
```diff
@@ -45,48 +53,63 @@ func (m mapped) CanMigrate(existing boilerplate.ExistingField) bool {
...
+	"decimal(38, 0)": {iceberg.Float64Type{}},
+	"string":         {iceberg.BinaryType{}},
+	// Any column can become a string (JSON text) or a variant: with
+	// variant_columns on, variant takes the role string has as the
+	// catch-all for fields that evolve to multiple types.
+	boilerplate.AnyExistingType: {iceberg.StringType{}, iceberg.VariantType{}},
+}
+
+// jsonColumnType is the column type of a JSON-shaped projection: variant with
+// variant_columns on, otherwise a string holding JSON text.
```
mdibaiee:
```suggestion
```

materialize-iceberg/type_mapping.go
```diff
@@ -45,48 +53,63 @@ func (m mapped) CanMigrate(existing boilerplate.ExistingField) bool {
...
-	"long":                      {iceberg.DecimalTypeOf(38, 0), iceberg.Float64Type{}},
-	"decimal(38, 0)":            {iceberg.Float64Type{}},
-	"string":                    {iceberg.BinaryType{}},
-	boilerplate.AnyExistingType: {iceberg.StringType{}},
+	"long":           {iceberg.DecimalTypeOf(38, 0), iceberg.Float64Type{}},
+	"decimal(38, 0)": {iceberg.Float64Type{}},
+	"string":         {iceberg.BinaryType{}},
+	// Any column can become a string (JSON text) or a variant: with
+	// variant_columns on, variant takes the role string has as the
+	// catch-all for fields that evolve to multiple types.
```
mdibaiee:
```suggestion
```

materialize-iceberg/type_mapping.go
```diff
@@ -30,8 +31,15 @@ type mapped struct {
 	// nullable even if the collection schema says it's always present and
 	// non-null. Primary key fields are always required regardless of this flag.
 	Nullable bool
+	// VariantFormat hints how a top-level string value written into a variant
+	// column is typed by the Spark job: "date-time", "date", or "binary". It
+	// is empty for columns that are not variant and for variants whose string
+	// values carry no such annotation.
```
mdibaiee:
```suggestion
	// VariantFormat hints how a top-level string key:value written into a variant
	// column is typed by the Spark job.
```

materialize-iceberg/type_mapping.go
```diff
@@ -45,48 +53,63 @@ func (m mapped) CanMigrate(existing boilerplate.ExistingField) bool {
...
 var migrateFieldSuffix = "_flow_tmp"
 
-func mapProjection(p boilerplate.Projection, translateField boilerplate.TranslateFieldFn) (mapped, boilerplate.ElementConverter) {
+// mapProjection maps a projection to its Iceberg column type. With
+// variantColumns set, JSON-shaped projections (objects, arrays, multi-type
+// fields, and the root document) map to variant instead of a JSON string.
+// Collection keys keep their string mapping regardless: Iceberg forbids
+// variant as an identifier field, and keys are joined and filtered on across
+// engines. A castToString field config arrives here already flattened to a
+// plain string type, and string-encoded numbers keep their numeric mapping.
```
mdibaiee:
```suggestion
```

materialize-iceberg/driver.go
```diff
@@ -416,6 +430,38 @@ func (d *materialization) CreateResource(ctx context.Context, res boilerplate.Ma
...
+	}
+	return nil
+}
+
+func withVariantHint(err error, hasVariant bool) error {
+	if err == nil || !hasVariant {
+		return err
+	}
+	return fmt.Errorf("%w (%s)", err, variantRemedyHint)
+}
```
mdibaiee: This doesn't check that the error is actually concerned with a variant column. If there is an error with column X which is not a variant, and we show variantRemedyHint, we are confusing the user.

materialize-iceberg/catalog/catalog.go
```diff
@@ -146,6 +146,24 @@ func SetCurrentSchemaUpdate(id int) TableUpdate {
...
+	baseUpdate
+	FormatVersion int `json:"format-version"`
+}
+
+func (upgradeFormatVersionUpdateReq) isTableUpdate() {}
+
+// UpgradeFormatVersionUpdate raises the table's Iceberg format version. It is
+// committed in the same request as a schema change that needs the newer
+// version (a variant column needs v3), so the table is never left at a
+// version that cannot describe its own schema.
```
mdibaiee:
```suggestion
// UpgradeFormatVersionUpdate raises the table's Iceberg format version.
```

materialize-iceberg/driver.go
```diff
@@ -416,6 +430,50 @@ func (d *materialization) CreateResource(ctx context.Context, res boilerplate.Ma
 	}, nil
 }
 
+const formatVersionProperty = "format-version"
+
+// variantRemedy tells the user how to fall back to JSON string columns; it is
+// appended to every error a task can only hit because variant columns are on.
```
mdibaiee:
```suggestion
```

materialize-iceberg/driver.go
```diff
@@ -416,6 +430,50 @@ func (d *materialization) CreateResource(ctx context.Context, res boilerplate.Ma
...
+const variantRemedy = "set the 'no_variant_columns' feature flag in the endpoint's advanced configuration to materialize these fields as JSON strings instead"
+
+// variantRemedyHint is appended to catalog errors from requests that carried a
+// variant column or format version 3, which not every catalog accepts.
```
mdibaiee:
```suggestion
```

materialize-iceberg/driver.go
```diff
@@ -416,6 +430,50 @@ func (d *materialization) CreateResource(ctx context.Context, res boilerplate.Ma
...
+const variantRemedyHint = "the catalog may not support Iceberg format v3 or variant columns; " + variantRemedy
+
+func schemaHasVariant(schema *iceberg.Schema) bool {
+	return slices.ContainsFunc(schema.Fields(), func(f iceberg.NestedField) bool {
+		return f.Type.Equals(iceberg.VariantType{})
+	})
+}
+
+// checkFormatVersionProperty rejects a user-supplied format-version other than
+// 3 on a binding that needs variant columns, rather than silently overriding it.
```
mdibaiee:
```suggestion
```

materialize-iceberg/driver.go
```diff
@@ -416,6 +430,50 @@ func (d *materialization) CreateResource(ctx context.Context, res boilerplate.Ma
...
+	}
+	return nil
+}
+
+// withVariantHint appends the remedy to a catalog error from a request that
+// carried a variant column, when the error itself is about the variant type
+// or the table's format version; other errors on such requests are left as
+// they are. The markers match Iceberg Java's messages, which Polaris returns
+// verbatim ("variant is not supported until v3", "Unsupported format
+// version: v3"); other catalog implementations' wording is unverified.
```
mdibaiee:
```suggestion
// they are.
```

materialize-iceberg/type_mapping.go
```diff
@@ -48,45 +51,33 @@ var allowedMigrations = boilerplate.TypeMigrations[iceberg.Type]{
...
+	}
+	return iceberg.StringType{}
 }
 
 var migrateFieldSuffix = "_flow_tmp"
 
-func mapProjection(p boilerplate.Projection, translateField boilerplate.TranslateFieldFn) (mapped, boilerplate.ElementConverter) {
+// mapProjection maps a projection to its Iceberg column type. With
+// variantColumns set, JSON-shaped projections (objects, arrays, multi-type
+// fields, and the root document) map to variant instead of a JSON string.
```
mdibaiee:
```suggestion
```

materialize-iceberg/type_mapping.go
```diff
@@ -132,9 +123,24 @@ func mapProjection(p boilerplate.Projection, translateField boilerplate.Translat
...
+		converter = variantConverter
+	}
+
 	return m, converter
 }
 
+// variantConverter makes every staged value of a variant column valid JSON
+// text for parse_json. Objects, arrays, and the root document already arrive
+// as JSON, and numbers and booleans are written as their JSON literals, but a
+// string value of a multi-type field would be written bare.
```
mdibaiee:
```suggestion
```

materialize-iceberg/docker-compose.yaml
```diff
@@ -211,13 +211,117 @@ services:
       - /bin/bash
       - -lc
       - /opt/spark/bin/spark-submit --master spark://spark-master:7077 /opt/jobs/daemon.py
+    # The daemon answers 503 on /health until its SparkSession is up, which
+    # takes several seconds; failures inside start_period do not count toward
+    # retries, so `up --wait` doesn't declare it unhealthy while it boots.
```
mdibaiee:
```suggestion
```

materialize-iceberg/docker-compose.yaml
```diff
@@ -211,13 +211,117 @@ services:
...
+      - ./python:/opt/jobs:ro
+      - ./testdata/.local:/shared:ro
+    working_dir: /opt/jobs
+    command:
+      - /bin/bash
+      - -lc
+      - /opt/spark/bin/spark-submit --master spark://spark4-master:7077 /opt/jobs/daemon.py
+    # The daemon answers 503 on /health until its SparkSession is up, which
+    # takes several seconds; failures inside start_period do not count toward
+    # retries, so `up --wait` doesn't declare it unhealthy while it boots.
```
mdibaiee:
```suggestion
```

materialize-boilerplate/type_mapping.go
```diff
@@ -129,6 +129,12 @@ func (m TypeMigrations[T]) CanMigrate(from string, to T) bool {
...
 func mapProjection(p pf.Projection, fc FieldConfiger) Projection {
```
jacobmarble: Why not export the existing func `mapProjection`?

materialize-iceberg/driver.go
```diff
@@ -416,6 +430,38 @@ func (d *materialization) CreateResource(ctx context.Context, res boilerplate.Ma
...
+// checkFormatVersionProperty rejects a user-supplied format-version other than
+// 3 on a binding that needs variant columns, rather than silently overriding it.
+func checkFormatVersionProperty(props map[string]string) error {
+	if v, ok := props[formatVersionProperty]; ok && v != "3" {
+		return fmt.Errorf("table property %s=%q conflicts with variant columns, which require format version 3: remove the property, or %s", formatVersionProperty, v, variantRemedy)
+	}
+	return nil
+}
+
+func withVariantHint(err error, hasVariant bool) error {
```
jacobmarble: Claude seems to name error-checking or error-wrapping functions without "error". For this wrapper, maybe `errorWithVariantHint()`?

jacobmarble (dismissed):
Nice work trimming comments.

jacobmarble (approved):

PR #5244, opened 2026-09-16
source-hubspot-native: include objects whose scopes are optional in the `properties` binding

Alex-Bair:
**Description:**

I'm working on moving some of `source-hubspot-native`'s required scopes to be optional, and this PR is some prep work for that upcoming change. It also fixes a bug where the `properties` snapshot wasn't trying to fetch properties for all of the different CRM objects we support.

This PR's scope includes:
- Allow the `properties` binding to gracefully handle the case where the provided credentials don't have the proper scopes to request an object's properties.
- Start requesting properties for all CRM objects we support, including the optional ones, in the `properties` snapshot.
- Various clean up, mostly consolidating references to CRM objects and their names to a single list in `resources.py` instead of maintaining two lists like before.

**Notes for reviewers:**

Best reviewed commit-by-commit. It looks like a lot of code changes, but the individual commits do a good job of breaking them into smaller chunks.

**Checklist:**

- [ ] If fixing a regression, I have linked the PR that introduced the regression: [insert link]
- [x] If there are user-visible changes, `CHANGELOG.md` has been updated (see [CONTRIBUTING.md](../CONTRIBUTING.md#changelog-entries))
- [ ] If there are user-facing behavior changes, documentation has been updated (see [documentation conventions](https://docs.google.com/document/d/1SRC9VS9zyCzWl3n4HXHbc4wPB1eLxJHkA2rtu9ZNokM/edit?usp=sharing)), or the docs PR is linked here: [insert link]

PR #5245, opened 2026-09-16
source-hubspot-native: make some required scopes optional

Alex-Bair:
**Description:**

Some users aren't able to authenticate their captures with OAuth because they lack some required permissions to do so. An example screen of what that looks like is below:
image

Some permissions, like "App Marketplace Access", are required to install an app & go through the OAuth process regardless of what scopes the connector requests. But other permissions listed in that screen seem to be related to the required scopes the connector requests, although that's still a bit unclear. Regardless, moving what scopes we can from the required bucket to the optional one is good practice & widens how many users can install our app. So, that's what this PR does.

The PR's scope includes:
- tolerating when the provided credentials don't have the necessary permissions to fetch associated entities for an object
- making the `forms`, `e-commerce`, and `tickets` scopes optional instead of required.

**Notes for reviewers:**
This is a follow up PR after https://github.com/estuary/connectors/pull/5244 was merged to make the `properties` binding resilient to optional scopes being missing.

Best reviewed commit-by-commit.

Our production OAuth app for `source-hubspot-native` needs the following scopes moved from required to optional once this PR is merged & the changes are live:
- `forms`
- `e-commerce`
- `tickets`

I will do that after merging this PR, then I'll also create a new capture to ensure the OAuth process still completes like it did before merging this PR.

**Checklist:**


source-hubspot-native/source_hubspot_native/api/object_with_associations.py
```diff
@@ -66,6 +67,55 @@
...
+        url = f"{HUB}/crm/v4/associations/{object_name}/{associated_entity}/batch/read"
+
+        try:
+            # A well-formed read for an ID that won't exist, so the probe reaches
+            # HubSpot's scope check without depending on the account holding any
+            # particular record.
+            await http.request(log, url, method="POST", json={"inputs": [{"id": "0"}]})
+        except HTTPError as err:
+            if is_missing_scope_error(log, err):
+                return False
```
nicolaslazo, line 95:
```suggestion
        except HTTPError as err:
            if not is_missing_scope_error(log, err):
                raise
            return False
```

I'd add the guard we use on the other missing scopes checks so we raise on any non-403s

nicolaslazo (commented):
Left a small suggestion

PR #5246, opened 2026-09-16
source-hubspot: migrate off deprecated v1 OAuth API

midianyoseph:
Refs #4620. Same migration as #5213, for the non-native connector.

- Token exchange and refresh endpoints → `/oauth/2026-09/token`
- `GET /oauth/v1/access-tokens/{token}` → `POST /oauth/2026-09/token/introspect`, which takes the token and app credentials as form fields instead of the token in the URL path. `get_granted_scopes` now takes `credentials` for that reason.

Tests: 105 passed, including live `test_discover` and `test_capture` against HubSpot.


Alex-Bair (approved):
LGTM, thank you!

PR #5253, opened 2026-09-17
materialize-s3-json: new connector writing gzipped NDJSON

SeanWhelan:
**Description:**

Adds `materialize-s3-json`, a third object-store file format alongside the existing CSV and Parquet sinks. Each line of a materialized file is one JSON document, gzip compressed and written as `.json.gz`. Object and array fields are written as nested JSON rather than encoded strings.

The connector is materialize-s3-csv with the writer and file extension swapped. The new `filesink.NewJsonStreamWriter` wraps `writer.JsonWriter`, which already satisfies `filesink.StreamWriter` and is what the Redshift, Databricks, BigQuery, MotherDuck and Snowflake stagers use, so there is no new serialization code here.

**Workflow steps:**

Same as the CSV and Parquet S3 connectors: configure bucket, region, credentials, optional prefix and upload interval, then one binding per collection with a `path`. The only format-specific option is `jsonConfig.skipNulls`, which omits null-valued fields instead of writing them as null.

**Notes for reviewers:**

The file extension is `.json.gz` rather than `.jsonl.gz`. The contents are newline-delimited, but `.json` is what materialize-motherduck already stages NDJSON as, and it is what Athena, Glue crawlers and Snowflake key off. Easy to change if you'd rather be literal.

`writer.WithJsonSkipNulls` had no test coverage anywhere, so `TestJsonWriterSkipNulls` is added next to the existing writer tests. Nothing was added to `filesink`, which does not test its other format constructors.

Requested by a customer who asked for JSON output from the S3 materialization. Still needs a control-plane connectors row, a UI tile, and the `go.estuary.dev/materialize-s3-json` short link, none of which live in this repo.

**Checklist:**

- [x] Not a regression.
- [x] No `CHANGELOG.md`: the connector is new, so there is no prior behaviour to describe a change to. materialize-s3-csv likewise has none.
- [x] Docs added in this PR: new `amazon-s3-json.md` page plus the materialization-connectors index entry.


williamhbaker (approved):
LGTM

materialize-s3-json/VERSION
```diff
@@ -0,0 +1 @@
+v1
```
danielnelson, line 1: Nit: can you add a trailing EOL.

danielnelson (approved):

PR #5258, opened 2026-09-18
source-tpc-ds: capture emitting the TPC-DS dataset via a forked dsdgen

### LINKED ISSUE #5256: source-tpc-ds: minimal capture connector emitting the exact TPC-DS dataset via a forked dsdgen
TL;DR for humans:
 
1. We use the existing tpc-ds C implementation for generating the documents (+ some patches to make it resumable).
4. There is a tpcds SQL file which creates all the tables for all the “bindings”: we use `go gen` to generate a file that maps the tables, their columns and each column’s type for JSONSchema
5. The Dockerfile pulls in the tpc-ds binary from a docker image generated by the tpc-ds fork’s CI. The Go connector shells out to the binary.
6. There is a `scale` factor that tells the capture connector how many documents it should emit. Supported scales are normally 1, 3, 10, 30, 100, etc., but 1 is already pretty big for a small smoke test / snapshot testing in the repo, so we add support for fractional scales by essentially applying a proportional limit to number of rows generated (e.g. 0.1 scale means 10% of rows generated, and the rest are discarded)
7. A fixed worker pool with one worker per enabled binding run concurrently.
8. Some tables are “children” of other tables. For example, there is a `store_returns` table, which is the child of `store_sales`. Documents of `store_returns` are written out as part of the `store_sales` generator, so if `store_sales` binding is disabled, we still have to run its generator to be able to get `store_returns`.7.
9. Once all documents of all streams are generated an exhausted (each `scale` has a limit, although very high numbers are practically infinite), the connector just idles.
10. In order to split the work for each binding into chunks so that we can resume them, the only mechanism available in tpc-ds's C library is to chunk the whole space of documents by "N" splits. So we get the row counter for each binding, split it according to our chunk size, and pass that as `-PARALLEL N` and when we want to resume from the `k`th part, we add `-CHILD k`

## Problem Statement

We have no way to verify materializations end to end against a well-established, externally defined benchmark. Existing synthetic sources (source-hello-world) emit trivial documents with no relational structure, so they cannot show whether a destination holds correct data after a full load. TPC-DS provides 24 related fact and dimension tables, a deterministic generator (dsdgen), 99 reference queries and reference answers, but nothing in Flow can produce that dataset as a capture.

## Solution

A minimal Go capture connector, `source-tpc-ds`, that emits the exact rows dsdgen produces for a chosen scale factor, one binding per TPC-DS table, typed so that SQL destinations create the column types the benchmark expects. After a full load, the TPC-DS reference queries can be run against the destination and compared to the published answers. Generation streams row by row through a patched dsdgen binary shipped from an Estuary fork, so the connector runs within a 1GB memory budget and uses no disk at any scale factor.

## User Stories

1. As a Flow engineer, I want a capture that emits the TPC-DS dataset, so that I can run the TPC-DS reference queries against a materialized destination and compare results to the published answers.
2. As a Flow engineer, I want the emitted rows to be byte-for-byte what dsdgen produces at the same scale factor, so that reference answers are valid without tolerance fudging.
3. As a Flow engineer, I want to choose the scale factor when creating the task, so that I can size the benchmark from smoke test to terabyte load.
4. As a Flow engineer, I want one binding per TPC-DS table, so that each destination table matches one benchmark entity and the reference queries run unmodified.
5. As a Flow engineer, I want all 24 tables discovered and enabled by default, so that a new task produces the complete schema without manual binding setup.
6. As a Flow engineer, I want to disable bindings I do not need, so that I can benchmark a subset of tables.
7. As a Flow engineer, I want each binding keyed on the spec's primary key, so that non-delta materializations upsert correctly and re-emission after a restart cannot create duplicates.
8. As a Flow engineer, I want integer and identifier columns emitted as JSON integers, so that destinations create integer columns.
9. As a Flow engineer, I want decimal columns emitted exactly, so that aggregate query results match the reference answers rather than drifting through floating point.
10. As a Flow engineer, I want date columns emitted with a date format, so that destinations create DATE columns and date arithmetic in the queries works.
11. As a Flow engineer, I want NULL values represented as absent fields with a nullable schema, so that destinations store SQL NULL.
12. As a Flow engineer, I want the connector to stay within a 1GB memory budget at any scale factor, so that it runs in a standard Flow task allocation.
13. As a Flow engineer, I want the connector to use no persistent disk, so that terabyte scale factors do not depend on task storage.
14. As a Flow engineer, I want throughput of at least 50MB/s of emitted documents, so that large scale factors load in reasonable time.
15. As a Flow engineer, I want all enabled tables to generate concurrently, so that wall-clock time is bounded by the largest table and not the sum of all tables.
16. As a Flow engineer, I want the connector to checkpoint progress regularly, so that a restart resumes rather than starting over.
17. As a Flow engineer, I want restart to reproduce exactly the rows that would have followed the checkpoint, so that the destination ends up identical to an uninterrupted run.
18. As a Flow engineer, I want restart cost to be bounded to a small fraction of a table, so that a restart at high scale factor does not regenerate hours of data.
19. As a Flow engineer, I want the connector to idle once every binding has emitted its full dataset, so that the destination stays fixed and the task stays healthy.
20. As a Flow engineer, I want a restart after completion to emit nothing further, so that the completed dataset is never re-sent.
21. As a Flow engineer, I want changing the scale factor on an existing task to fail with a clear error directing me to backfill, so that the destination never holds rows from two different datasets.
22. As a Flow engineer, I want fractional scale factors below 1, so that I can run a smoke test or demo with a proportionally smaller dataset (0.1 is a tenth of what scale 1 produces).
23. As a Flow engineer, I want fractional scale factors to follow DuckDB's tpcds semantics exactly, so that DuckDB's shipped reference answers for sf=0.01 validate a small end-to-end run.
24. As a Flow engineer, I want a returns binding to work even when its sales parent is disabled, so that I can benchmark returns alone.
25. As a Flow engineer, I want the dsdgen patches maintained in an Estuary fork rather than applied at image build time, so that the provenance of the generator is reviewable and reusable.
26. As a Flow engineer, I want the fork to publish a container image of the generator, so that the connector build pins a tag and never compiles C.
27. As a Flow engineer, I want the connector's table definitions generated from the TPC-DS DDL, so that column names, types, nullability and keys cannot drift from the spec by hand-typing.
28. As a Flow engineer, I want the connector's configuration knobs to survive catalog sanitization, so that sanitized task specs remain reproducible.
29. As a Flow engineer, I want a README with purpose, knobs, fork provenance and how to run the reference queries, so that I can use the connector without reading the code.
30. As a Flow engineer, I want the connector built and tested in CI like other Go connectors, so that regressions are caught on every change.
31. As a reviewer, I want a snapshot test of discovered schemas and a small captured dataset, so that changes to typing or discovery are visible in the diff.
32. As a reviewer, I want unit tests of the row decoder without the generator binary, so that typing and routing bugs are isolated quickly.

### Generator: patched upstream dsdgen via an Estuary fork

- Rows are produced by the upstream C dsdgen (TPC-DS tools v2.10.0), run as a subprocess with its stdout mode, streaming pipe-delimited rows. It uses about 2MB of memory and no disk regardless of scale factor.
- Embedding DuckDB's tpcds extension was rejected: its dsdgen materializes all 24 tables into a database before any row can be read (3.6GB peak at sf=1, linear in scale factor), which violates the memory and disk constraints. A Go port and the Java (Trino) port were rejected as disproportionate or wrong language.
- The generator lives in a fork `estuary/tpcds-kit` of gregrahn/tpcds-kit, carrying five patches:
  1. Stdout mode fix: the option is registered as `_FILTER` but checked as `FILTER`, so stdout mode never engaged.
  2. Stdout mode fix: after selecting stdout, the output handle was overwritten by the table's null file pointer, producing "Failed to open output file".
  3. Chunking threshold removal: upstream only honours `-PARALLEL`/`-CHILD` for tables over 1M rows and silently emits nothing for other children of smaller tables. With `-PARALLEL` set explicitly, every table chunks.
  4. Row-count query: a hidden flag that prints the row count of `-TABLE` at `-SCALE` and exits, following dsdgen's hidden-flag convention. The connector cannot compute row counts itself because dimension tables scale through dsdgen's internal scaling bands.
  5. Build fix: compiles cleanly under current gcc/clang without relaxed C standard flags.
  6. Fractional scale: a port of DuckDB's tpcds extension change. `SCALE` becomes a string option read as a double; for scale below 1, every table starts from its sf=1 row count and is multiplied by the fraction with a floor of one row, except tables the rowcounts distribution marks static with no multiplier (date_dim, time_dim, catalog_page, ship_mode, income_band and similar), which keep their fixed size. call_center and inventory read the scale as a double. Scale index and scale-band lookups treat any fraction as band 1. Acceptance: the fork's output at sf=0.01 is byte-identical to DuckDB's `dsdgen(sf=0.01)` for every table.
- The fork keeps TPC's legal notice in its README. A workflow in the fork builds and publishes `ghcr.io/estuary/dsdgen`, containing the `dsdgen` binary and its `tpcds.idx` distributions file, tagged by commit.
- The connector Dockerfile copies the binary and distributions file from that image at a pinned tag into both the build stage (so tests have it) and the runtime stage. The connector passes the distributions file path explicitly on every invocation.
- Verified locally: chunked output concatenates byte-identically to unchunked output for store_sales, customer, inventory and small dimension tables; a chunk of 1/100 of store_sales at sf=1 generates in 0.13s; output is deterministic across runs.

### Connector shape

- Go, in the captures context, modelled directly on source-hello-world: a `driver` implementing the boilerplate `Connector` interface with Spec, Validate, Discover, Apply and Pull.
- Endpoint config: `scale` only (number, default 1, must be positive). Values of 1 and above should be the spec's official factors (1, 3, 10, 30, 100, 300, 1000, 3000, 10000, 30000, 100000) and the description says so; values below 1 produce a proportionally smaller dataset under the fork's fractional-scale semantics and are intended for smoke tests, CI and demos. No concurrency knob and no row limit. The key is added to the obfuscation allowlist.
- Resource config: the table name. Resource path pointer is the table name. Discovery recommends the table name as the binding name.
- Bindings: one per TPC-DS table, 24 total (7 fact: store_sales, store_returns, catalog_sales, catalog_returns, web_sales, web_returns, inventory; 17 dimension). `dbgen_version` is excluded. All discovered bindings are enabled by default. Keys are the spec's primary keys.
- Document field names are the spec's column names unchanged (for example `ss_sold_date_sk`).

### Table definitions

- A `go generate` step parses the fork's `tpcds.sql` DDL (25 CREATE TABLE statements with column types, `not null` and `primary key` clauses) and writes a checked-in Go file holding, per table: name, ordered columns with kind (integer, decimal, date, string), nullability, primary key columns, parent table (for returns tables) and field count.
- The DDL file is copied into the connector directory from the fork at the pinned version. CI verifies the generated file is up to date.

### Typing

- integer and identifier columns: JSON integer, schema `type: integer`.
- decimal(p,s): JSON string, schema `type: string, format: number`, so materialize-sql creates DECIMAL/NUMERIC and values stay exact.
- date: JSON string `YYYY-MM-DD`, schema `type: string, format: date`.
- char(n), varchar(n): JSON string.
- NULL (empty field in dsdgen output): field omitted; such columns are not `required` in the schema. Key columns are `not null` in the DDL and are always present.
- Discovered document schemas are generated from the table definitions, not from Go structs.

### Execution model

- Work item: `(parent table, chunk k of N)`. Parent tables are the 21 tables dsdgen can generate directly; the three returns tables are emitted by their sales parent's process, interleaved on stdout, and routed to their own binding by field count (store_sales 23 fields vs store_returns 20; catalog 34 vs 27; web 34 vs 24, before the trailing delimiter).
- N per table is derived from the fork's row-count flag at the configured scale: about 1M rows per chunk, at least 2 for any chunked table, no chunking for tables under the target. The scale is passed to dsdgen verbatim as the configured decimal string.
- A fixed worker pool with one worker per enabled parent binding runs concurrently; chunks within a table run in order. If only a returns binding is enabled, its parent's process still runs and parent rows are discarded.
- Each work item runs one subprocess: `dsdgen -SCALE <sf> -TABLE <parent> -PARALLEL <N> -CHILD <k> -_FILTER Y -DISTRIBUTIONS <path>` (unchunked tables omit `-PARALLEL`/`-CHILD`). Lines are decoded and emitted as documents to the owning binding as they stream.
- Per-binding checkpoint state: the highest contiguous completed chunk, a map of in-flight chunk to lines consumed from that chunk's stdout (shared by a parent and its returns binding since they read one stream), a done flag, and the scale factor. Checkpoints are emitted every 10k documents and at chunk completion.
- On resume, in-flight chunks are rerun and their first `linesConsumed` lines discarded; completed chunks are skipped. Determinism guarantees identical rows.
- On open, if the state's scale differs from the configured scale, Pull fails with an error telling the user to backfill the bindings. If state carries no scale (fresh), it is recorded.
- When all enabled bindings are done, the connector logs completion and blocks on context cancellation. A restart in that state emits nothing.
- Emitted documents carry no `_meta`; the dataset is a fixed load, not a change stream.

### Repo integration

- Directory `source-tpc-ds` with VERSION, Dockerfile, CHANGELOG entry, README (purpose, knobs, fork provenance and patch list, running reference queries), and a CI matrix entry. `DocumentationUrl` points at the README on GitHub. No docs-site page.

## Testing Decisions

A good test observes the connector through its protocol surface or a pure function's inputs and outputs, and never asserts on subprocess handling, goroutine structure or internal state layout.

- Seam 1 (existing): the boilerplate capture test harness (`CaptureSpec` in source-boilerplate/testing), as used by source-hello-world's test. One snapshot test runs Spec, Discover and a full Pull with all 24 bindings at a fractional scale (0.001), using the real dsdgen binary present in the Dockerfile build stage. Because the static tables (date_dim 73k rows, time_dim 86k rows, catalog_page 11.7k rows) do not shrink with scale, the snapshot records per binding the document count, a digest of all documents in emission order, and the first few documents verbatim, plus the final checkpoints, rather than every document. Tests needing the binary skip when it is not found locally. Snapshots are updated with `UPDATE_SNAPSHOTS=true` per repo convention.
- Seam 2 (new): the row decoder, a pure function from table definition plus one pipe-delimited line to a document, with the field-count router in front. Unit tests cover each column kind, empty-field NULL omission, trailing delimiter handling, routing of parent versus returns lines, and rejection of lines with an unexpected field count. No binary required.
- The DDL generator has no dedicated seam; its checked-in output is exercised by both seams and a CI check that regeneration produces no diff.
- Prior art: source-hello-world's `TestCapture` for the harness usage and cupaloy snapshotting.

## Out of Scope

- TPC-DS data maintenance (refresh) sets: inserts and deletes against fact tables via dsdgen's update mode. The state model does not preclude adding this later.
- A resume test that interrupts and restarts a capture.
- Running the 99 reference queries or comparing answers; that belongs to the benchmark harness that consumes the destination.
- A docs-site page or customer-facing availability.
- Exposing dsdgen's arbitrary start-row skip as a separate flag; chunking covers resume cost for now.
- Persisting generated data outside the connector.

## Further Notes

- DuckDB ships TPC-DS reference answers for sf 0.01, 1, 10 and 100. Matching DuckDB's fractional semantics means a full end-to-end verification (capture, materialize, run the 99 queries, compare) is possible at sf=0.01 in minutes: about 200k rows, dominated by the static date_dim and time_dim tables.
- Measured at sf=1: 19.6M rows across all tables, about 1GB pipe-delimited, about 4GB as JSON. One dsdgen process generates store_sales at about 220k rows/s, roughly 30MB/s raw and 115MB/s as JSON, so the Go decoding and Flow runtime, not dsdgen, are the throughput ceiling.
- Restart waste per binding without resume would have been 13s at sf=1, 22 minutes at sf=100 and 3.6 hours at sf=1000; chunking bounds it to one chunk.
- The dsdgen source is distributed under TPC's legal notice, not an OSI licence. DuckDB and Trino redistribute it; the fork does the same with the notice intact.
- Local build note for the fork: upstream fails under current clang with implicit-int errors in date.c; the fork's build fix addresses this rather than passing `-Wno-implicit-int -std=gnu89`.

mdibaiee:
Closes #5256.

A minimal Go capture that emits the exact TPC-DS dataset for a configurable scale factor, one binding per table.

Rows stream from the stdout of a patched `dsdgen` published by [estuary/tpcds-kit](https://github.com/estuary/tpcds-kit) (`ghcr.io/estuary/dsdgen:a0d4ada`, from estuary/tpcds-kit#2), so the connector uses no disk and a few megabytes per table at any scale.

## Design

See https://github.com/estuary/connectors/issues/5256

## Tests

Decoder and router unit tests without the binary. Harness tests at scale 0.01 (all 24 bindings, ~278k documents, a few seconds) snapshot per-binding counts, an order-sensitive digest and the first three documents, and check that chunked generation, an interrupted-then-resumed run, and a returns-only capture reproduce the same digests. They run inside the Dockerfile build stage against the real generator; `go test` skips them locally unless `dsdgen` is on `$PATH` or named by `$DSDGEN`.

Verified locally: image builds for linux/amd64 with all tests passing, `tests/smoke.sh` passes, and the Linux gcc build's output equals the macOS clang build's for the whole sf=0.01 dataset.


source-tpc-ds/dsdgen.go
```diff
@@ -0,0 +1,107 @@
...
+	n, err := strconv.ParseInt(strings.TrimSpace(string(out)), 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("dsdgen row count for %s: unexpected output %q", table, out)
+	}
+	return n, nil
+}
+
+func (g *generator) stream(ctx context.Context, table string, chunks, chunk int, fn func(line []byte) error) error {
+	var args = g.args(table, "-_FILTER", "Y")
+	if chunks > 1 {
```
mdibaiee, line 69:
```suggestion
// In order to split the work for each binding into chunks so that we can resume them, the only mechanism available in tpc-ds's C library is to chunk the whole space of documents by "N" splits. So we get the row counter for each binding, split it according to our chunk size, and pass that as `-PARALLEL N` and when we want to resume from the `k`th part, we add `-CHILD k`
	if chunks > 1 {
```

danielnelson (approved):

Alex-Bair (approved):
LGTM as an internal tool for testing materializations.

PR #5265, opened 2026-09-18
source-iterable-native: detect campaign changes that don't bump updatedAt

### LINKED ISSUE #5259: source-iterable-native: campaigns cursor misses changes that don't bump updatedAt
## 1. Priority

Medium — silent staleness, no error. Campaign rows freeze at a pre-send state (e.g. `Scheduled`) indefinitely while the task reports healthy and keeps ingesting. No data loss, but the destination is quietly wrong and there is nothing to alert on.

## 2. Scope and prevalence

Every `source-iterable-native` capture with the `campaigns` binding enabled. Observed at one tenant on 2026-09-18, with a stale row roughly 15 days old. Prevalence is unmeasured — the task stays OK and the stream keeps writing, so this is likely under-reported.

## 3. Trigger (testable)

A campaign whose `campaignState` changes without `updatedAt` advancing past the stored cursor.

A campaign was scheduled, sent, and marked Finished in Iterable, but its updatedAt never moved off the time it was scheduled. Because the code only compares updatedAt, the connector never saw the campaign as changed and never re-read it. The source shows Finished; the destination still shows Scheduled, and will keep showing that.

## 4. Root cause + confidence

Likely. `fetch_campaigns` filters solely on `updatedAt`, `break`s at the first campaign at or below the cursor, and advances a single high-water mark with no lookback:

```python
async for campaign in _paginate_through_campaigns(http, log, sort_param="-updatedAt"):
    if campaign.updatedAt <= last_checkpointed_ts:
        break

    most_recent_ts = max(most_recent_ts, campaign.updatedAt)
    yield campaign
```

https://github.com/estuary/connectors/blob/main/source-iterable-native/source_iterable_native/api.py — `fetch_campaigns`

`campaignState` is an ordinary field on the document, so a state transition is not separately detectable — it rides along when the document is yielded, or it never arrives.

The connector already models this gap elsewhere. `Campaigns.most_recent_ts` returns `max(updatedAt, endedAt)` and is used by `_should_fetch_metrics` for final-state campaigns; it is dead code unless `endedAt` can exceed `updatedAt`. `fetch_campaigns` does not call it. `fetch_campaign_metrics` additionally carries a 15-day `CAMPAIGN_METRICS_LOOKBACK_WINDOW`, while `fetch_campaigns` has no lookback at all.

Iterable's published sample payload for `GET /api/campaigns` shows a `Finished` campaign whose `endedAt` is roughly 25 hours after its `updatedAt`, which is consistent with `updatedAt` not tracking completion.

Not verified: whether Iterable failed to bump `updatedAt` on completion, or bumped it to a value already below the checkpointed maximum. The stored document cannot distinguish these, since it reflects what the connector last captured either way. Both are permitted by the code above and both are addressed by the same fix.

## 5. Investigation steps

Observed on a healthy capture — status OK, all bindings backfilled, `campaigns` writing 12–30 docs/day throughout the period. Only the affected campaign was stale, so this is a skipped record rather than a stall.

The stored document for the affected campaign had `updatedAt` about four hours *before* its own `startAt`, and no `endedAt` key at all — i.e. the last version captured predates the send. The collection wrote 17 documents that day and 12 the next, so the cursor moved past that campaign's `updatedAt` almost immediately and the descending scan never reached it again.

A binding backfill corrects the data, since `backfill_campaigns` has no watermark and re-emits every campaign at or below the backfill cutoff. That is a one-time correction, not a fix — the next campaign to finish will drift the same way.

## 6. Definition of done

- A campaign that transitions to a final state is re-emitted with its current `campaignState` on the next poll, without a backfill.
- Regression test: a fixture where `campaignState` changes and `endedAt` is set while `updatedAt` stays fixed; assert the document is re-emitted.
- Suggested fix: key the cursor on `most_recent_ts`, or give `campaigns` a lookback window. `fetch_campaign_metrics` already re-paginates the full campaign list on every run, so a periodic re-snapshot of `campaigns` costs no additional API calls.
- Constraints / do not break: existing cursor state must not force a full re-backfill on upgrade.

## 7. References

- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1789739841869859
- Iterable API endpoints and sample payloads — https://support.iterable.com/hc/en-us/articles/204780579-Iterable-API-Endpoints-and-Sample-Payloads

Alex-Bair:
**Description:**

Iterable leaves a campaign's `updatedAt` at its last edit when the campaign starts or finishes, so the `-updatedAt` watermark scan in `fetch_campaigns` never re-read a campaign after it sent; destination rows stayed frozen at `Scheduled` with no endedAt (#5259).

`fetch_campaigns` now scans every campaign sorted by the immutable `id` and emits those with `updatedAt`, `startAt`, or `endedAt` inside `(cursor, horizon]`, where the horizon is fixed at scan start and trails now by one minute.

Discover snapshot changes are expected due to the addition of the `startAt` field to the Pydantic model for the `campaigns` stream.

**Checklist:**


nicolaslazo (approved):
LGTM

PR #5266, opened 2026-09-18
source-mysql{,-batch}: log when traffic may travel unencrypted by def...

nicolaslazo:
**Description:**

ault.

**Workflow steps:**

**Notes for reviewers:**

**Checklist:**


source-mysql-batch/main.go
```diff
@@ -96,6 +96,9 @@ func (c *Config) Validate() error {
 // falling back to an unencrypted connection.
 func (c *Config) sslSettings() mysqltls.Settings {
 	var mode = c.Advanced.SSLMode
+	// TODO: We eventually want to make `ModeRequired` the global default.
+	// We'll have to set up a feature flag so we don't break any existing
+	// captures that do not support SSL
```
Alex-Bair, line 101: note: instead of setting a feature flag that determines what the default is, I think we could just set the mode for each capture directly, couldn't we?

Alex-Bair (approved):
LGTM

source-mysql-batch/main.go
nicolaslazo: Of course, I only assumed the capture count would make that impractical. `source-mysql-batch` is not that popular, though...

PR #5269, opened 2026-09-21
source-sftp: verify the server's host key against configured known_hosts

nicolaslazo:
**Description:**

This PR adds an optional top-level field, `knownHosts` ("SSH Known Hosts"). Set it, and the connector checks the server's host key against it and refuses to connect on a mismatch. Leave it empty, and the connector behaves as it did before and logs a warning.

The field's content goes to `golang.org/x/crypto/ssh/knownhosts` as it stands. That buys the whole `known_hosts` grammar: hashed hosts, wildcards, negation, several keys for one host, `@cert-authority` and `@revoked`.

**Workflow steps:**

Run `ssh-keyscan -p 2222 myserver.com` from a machine you trust and paste the output into the new field. The host column must name the address the connector dials: the bare hostname on port 22, `[host]:port` on any other. `ssh-keyscan -p` prints it that way already.

Two other routes work. Pin an `@cert-authority` line and every host certificate that authority signs will verify, so a rebuilt server does not break the capture. Or pin several keys at once to cover both sides of a rotation.

If your provider publishes a fingerprint and no key, as AWS Transfer Family does, scan the server and check what you scanned against what they publish:

```
ssh-keyscan -p 2222 myserver.com | ssh-keygen -lf -
```

**Notes for reviewers:**

Two decisions deserve your attention.

The algorithm list comes from the pins. Go prefers ECDSA, then RSA, then DSA, then ed25519. A server holding both an RSA and an ed25519 key would therefore offer the RSA one, and a correct ed25519 pin would fail as a mismatch. So the connector offers only the pinned key types, in the order configured, and a type nobody pinned can never be offered. RSA keys expand to `rsa-sha2-512`, `rsa-sha2-256` and `ssh-rsa`, per RFC 8332, because modern servers refuse the SHA-1 name. `SetDefaults()` runs after this and touches only ciphers, key exchanges and MACs.

The presented key is echoed for an unknown host and nowhere else. A user who has no working pin must get one somewhere, so that error prints a paste-ready line and the SHA-256 fingerprint, and it names the entry to write if the same key is pinned under another host column. On a mismatch, a revoked key or a failed certificate check the error names the key type and stops. Printing the key we were just offered would turn the error message into a "trust me" button for the attacker who sent it.

I ran six scenarios by hand, through `flowctl raw preview-next`, against five SFTP servers:

| #   | Scenario                                                                                                               | Expected                                                                                                                                | Result                                           |
| --- | ---------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------ |
| 1   | The server's own ed25519 key, pinned. The server also holds RSA and ECDSA keys.                                        | Succeeds. Shows that the pinned type is the one negotiated.                                                                             | 30 documents                                     |
| 2   | `knownHosts` empty.                                                                                                    | Succeeds, and logs `no SSH Known Hosts are configured, so the identity of the SFTP server is not verified`.                             | 30 documents, and the warning                    |
| 3   | Another server's ed25519 key, pinned for this host.                                                                    | Refused as a mismatch. The key the server offered must not appear in the error.                                                         | Refused.                                         |
| 4   | The right key, written without the port.                                                                               | Refused as an unknown host, naming `[sftp-hk-a.flow-test]:2222` as the entry to write and echoing the offered key with its fingerprint. | Refused, hint and echo both present              |
| 5   | A server holding a certificate signed by a private authority. One `@cert-authority` line pinned, no host key anywhere. | Succeeds.                                                                                                                               | 30 documents                                     |
| 6   | A good pin, then a line of rubbish.                                                                                    | Rejected as the config is validated, by line number, before the connector dials.                                                        | `knownHosts line 2: not a known_hosts entry ...` |

What I considered and dropped:

- **Typed `SHA256:` fingerprints as a second accepted form.** Tempting for providers that publish nothing else. But it is a second parser in a path where parsing is the security, it cannot carry `@cert-authority` or `@revoked`, and it is paste-ready in neither direction. The docs send those users through `ssh-keygen -lf -` instead: the same check, one format fewer to keep.
- **MD5 or SHA-1 fingerprints.** x/crypto has no SHA-1 helper, and MD5 as a trust anchor inside a fix for a high CWE-295 finding will not survive review.
- **The field under `advanced`.** That object tells users they "should not typically need to modify these", which is the wrong advice for a security control. It sits at the top level, after `address`.
- **Marking the field `secret`.** Not `secret`: the UI makes users retype encrypted fields at every edit, and a public host key is not a secret.

**Checklist:**


source-sftp/knownhosts.go
```diff
@@ -0,0 +1,267 @@
...
+		"Verify this is the server's genuine host key, then add that line to the SSH Known Hosts configuration",
+		address, hint, key.Type(), knownhosts.Line([]string{address}, key), ssh.FingerprintSHA256(key))
+}
+
+// callback builds the ssh.HostKeyCallback verifying the server at `address`
+// (the configured host:port, which is also what ssh.Dial is given) against the
+// pins.
+func (p *knownHosts) callback(address string) ssh.HostKeyCallback {
+	return func(hostname string, remote net.Addr, key ssh.PublicKey) error {
+		err := p.verify(hostname, remote, key)
```
Alex-Bair, line 177: Revoked host keys remain usable through certificates.

With a trusted `@cert-authority` and an `@revoked` entry for a host key, the callback still accepts a valid certificate containing that revoked key. The dependency checks revocation of the certificate and CA, but omits the underlying host key. A compromised key therefore remains usable with its previously issued certificate. Check `cert.Key` against revoked entries before accepting.

Alex-Bair (commented):
Thanks for working on this! Astra found one item around `@revoked` host keys that should be addressed before merging, but it looks good otherwise!

source-sftp/knownhosts.go
nicolaslazo: Great find. It's fixed 💪

Alex-Bair (approved):
LGTM, thanks!

PR #5272, opened 2026-09-21
materialize-kafka: register flow_published_at as an Avro timestamp for new materializations

### LINKED ISSUE #5162: materialize-kafka: flow_published_at emitted as Avro string instead of timestamp-micros
## 1. Priority
Low - one customer, data flows correctly and only the schema type annotation is wrong.

## 2. Scope and prevalence
Reproduces in this repo's own test suite, so it affects every `materialize-kafka` binding using Avro, not one tenant's configuration. Reported by one customer so far, on 2026-09-01, running `ghcr.io/estuary/materialize-kafka:v1`. Affects the synthetic projections only (`flow_published_at`, `flow_document`); regular collection fields are not affected. Dekaf, which encodes Avro from the same shared crate, emits `flow_published_at` as a timestamp.

## 3. Problem description
- Expected: `flow_published_at` is emitted as an Avro `long` with `logicalType: timestamp-micros`, matching how it is materialized elsewhere (`timestamptz` on Postgres, `TIMESTAMP` on Snowflake, a `timestamp` column on S3 Iceberg) and matching Dekaf's Avro output for the same field.
- Observed: it is emitted as a plain Avro `string` with no logical type. The registered value schema contains:

```
{"name":"flow_published_at","type":"string"}
```

  The value itself is RFC 3339 shaped, so only the type annotation differs. A customer consumer that reads the field as a `long` cannot use it.

- Steps to reproduce: run the existing `materialize-kafka` test suite and read the committed snapshot at `materialize-kafka/tests/snapshots/test__materialization.snap`, topic `test_topic_avro`. The decoded value is:

```
"flow_published_at":"1970-01-01T01:00:00.000000000Z"
```

  The snapshot decodes through `avro_to_json` in `materialize-kafka/tests/test.rs`, which renders `AvroValue::TimestampMicros` via RFC 3339 and emits no fractional digits for a whole-second value. The nine fractional digits above are produced by the `AvroValue::String` arm. The same field in `crates/dekaf/tests/snapshots/dekaf_integration_test__fields_not_required-value-0.snap` in `estuary/flow` is redacted as a timestamp type.

Note on compatibility: correcting the type changes the registered Avro schema for existing topics. The reporting customer raised this themselves and said opt-in behaviour would be acceptable to them.

## 4. References
- Slack thread: https://estuaryworkspace.slack.com/archives/C03Q2NRFKDL/p1788242254382919
- Related prior art: #5101

jacobmarble:
**Description:**

Fixes #5162. In Avro topics, `flow_published_at` was registered as a plain `string` even though it holds a timestamp, so a consumer that wanted a real timestamp had to parse it themselves. New materializations now register it (and any other field with `format: date-time`) as a `long` with `logicalType: timestamp-micros`, which is what Dekaf already does for the same field.

Two things caused the bug:

1. When the connector built the Avro schema for `flow_published_at`, it copied the field's `format` into the wrong slot (`content_encoding`), so the schema builder never saw `date-time`.
2. The connector registered schemas in Avro's "Parsing Canonical Form". That form strips `logicalType` by design, so even correctly typed fields lost their timestamp/decimal labels on the way to the registry.

Changing a field's type on an existing topic would break consumers that expect a string, so existing materializations keep the old encoding. The connector decides once per task, at the first Apply: a task with no previously applied spec is new and gets logical types; any other task stays on the old encoding. The decision is stored in the connector's persisted state and never changes afterwards.

**Workflow steps:**

- New Avro materializations get the timestamp encoding automatically. Nothing to configure.
- Existing materializations are unchanged. To adopt the new encoding, create a new materialization.

**Notes for reviewers:**

- The decision relies on the runtime persisting the state returned from Apply and passing the previously applied spec. The v2 runtime does both. The v1 runtime does neither, so on v1 every task stays on the old encoding. That is safe but means new tasks only get timestamps once they run on v2.
- Commits are in TDD pairs: a snapshot-only commit that fails, then the fix that makes it pass. The second pair adds an `avro-legacy` test task seeded with `--initial-state` to cover the old path.
- Subject names are a hash of the registered schema text, so every existing Avro binding that adopts the new encoding gets new registry subjects. That is the same thing that happens on any schema change today.

**Checklist:**


danielnelson (commented):
I feel like we need to add a feature flag to override the automatic behavior too.  This way you don't have to recreate your existing task or you can keep the old behavior on a new task to match others using a string.

jacobmarble:
> I feel like we need to add a feature flag to override the automatic behavior too. This way you don't have to recreate your existing task or you can keep the old behavior on a new task to match others using a string.

The Avro schema change is enabled only when both (1) connector is newly created and (2) `Apply.state_json` and `Applied.state` are handled (runtime v2).

I've added a flag that *explicitly disables* the schema change, so if a customer has any trouble, they have an escape hatch.

materialize-kafka/src/transactor.rs
danielnelson, line 100: I think this might clear the connector state?  Also below in Acknowledge.

danielnelson (commented):
It would be nice if we could enable this flag on an old task, this would require the task to be recreated from scratch.  I'm also uneasy with storing this information into the connector state, since it doesn't have very good visibility and maybe it can get cleared by accident.

materialize-kafka/src/transactor.rs
jacobmarble: `state: None` means "no update" not "clear the state". `Some(...)` touches the state.

jacobmarble:
> It would be nice if we could enable this flag on an old task, this would require the task to be recreated from scratch. I'm also uneasy with storing this information into the connector state, since it doesn't have very good visibility and maybe it can get cleared by accident.

I'll try something simpler with a regular config parameter. My goal is that new tasks to use the new Avro schema, and existing tasks to stay with the current schema.

Claude suggests we can do this by setting a JSON schema field default value to `true`, and the corresponding serde field default value to `false`.

danielnelson (approved):

PR #5274, opened 2026-09-21
source-sftp: add a hidden explicit host key verification opt-out

Alex-Bair:
**Description:**

source-sftp: add a hidden explicit host key verification opt-out

Leaving `knownHosts` empty is the only way to connect without verifying the server, and the connector cannot tell an empty field from an absent one. That means we can't change the meaning of an absent or empty `knownHosts` until there is a way to pin the current behavior for existing tasks.

This commit adds a `skipHostKeyVerification` boolean directly under `knownHosts` in the UI so existing tasks' current behavior can be pinned before functionality for `skipHostKeyVerification` is added.

The connector does not act on `skipHostKeyVerification` yet. An empty `knownHosts` still connects unverified with the flag set or not. Setting both the flag and `knownHosts` is a configuration error.

The new field is hidden in the UI and undocumented since it doesn't functionally do anything yet. Functionality will be added in a future PR after setting an explicit value for all existing tasks.

**Notes for reviewers:**

This is part of the effort to address the finding in https://github.com/estuary/security/issues/845. Essentially, we need to opt-out all existing tasks from host key verification before changing the connector's default behavior. `skipHostKeyVerification` is how we'll opt-out existing tasks.

The overall plan to eventually have `skipHostKeyVerification` affect behavior is:
1. Introduce the `skipHostKeyVerification` field without it changing any functionality.
2. Pin all existing tasks' specs to specify `skipHostKeyVerification: true`.
3. Update the connector to actually consider `skipHostKeyVerification` when deciding whether host key verification should be performed.

No changelog or doc changes in this PR yet - none of the changes are user facing. Changelog entries and documentation changes will be made as part of step 3 above.

**Checklist:**


nicolaslazo (approved):
LGTM

Issue #5275, opened 2026-09-22
SEO/AEO: deprecated connector pages compete with their replacements, two live pages lack descriptions


jwhartley:
## Goal

Stop deprecated connector pages competing in search with the connectors that replaced them. Each page in `docs/reference/Connectors` is a landing page for a high-intent query, so a deprecated page and its successor split authority for the same term. The deprecated URL is usually older with more inbound links, so it often outranks the page we want people to land on, and an AI assistant retrieving it will describe a connector we no longer want anyone to configure.

## Status

#5277 applies `noindex, follow` to all 19 deprecated pages and drops them from `sitemap.xml`, while keeping them reachable for customers still running those connectors.

## Notes

`estuary/docs` aggregates this directory as a submodule, so changes here reach docs.estuary.dev on the next deploy.

jwhartley:
#5277 applies `noindex, follow` to all 19 deprecated pages. Verified on a full build: 327 sitemap URLs on the base commit, 308 with the change, exactly 19 fewer and nothing else lost. All 19 still serve.

Note for anyone doing this again: `noindex: true` in front matter looks like it should work but silently does nothing. Docusaurus accepts unknown front matter keys and ignores them, so the page stays indexed and the build still passes.

Instead, put a `<head>` block with the robots meta tag in the page body. Docusaurus treats it like any other head tag, and the sitemap builder reads those same tags when deciding what to leave out. So one block both tells crawlers not to index the page and removes it from `sitemap.xml`, with no theme changes.

PR #5277, opened 2026-09-22
docs: stop deprecated connector pages competing with their replacements

jwhartley:
## What changed

19 deprecated connector pages now carry a `<head>` block with `<meta name="robots" content="noindex, follow" />`, plus a `description` where one was missing or did not mention the deprecation.

## Why

Deprecated connector pages compete in search with the connector that replaced them, and the deprecated URL is usually the older one with more inbound links, so it can outrank the page we want people to land on. The pages must stay reachable, because customers still run these connectors in production tasks.

`follow` is deliberate, so crawlers keep walking the in-body link to the successor and pass the deprecated page's authority to it.

A `<head>` block needs no theme code. The mdx-loader rewrites a lowercase `<head>` to the `Head` component (`mdx-loader/lib/remark/head/index.js`), and `Head` is a thin `react-helmet-async` wrapper (`core/lib/client/exports/Head.js`). That is the same Helmet instance the sitemap plugin reads when it decides what to exclude, so these pages drop out of `sitemap.xml` automatically. Setting the tag any other way would leave the URL advertised in the sitemap while telling crawlers to ignore it.

This replaces an earlier approach that used `noindex: true` front matter. Docusaurus has no such key (the docs front matter schema ends in `.unknown()`, so it validates and does nothing), so that route needed a swizzled `DocItem/Metadata` in estuary/docs. The `<head>` block reaches the same place with no theme file to own across Docusaurus upgrades.

## How it was verified

Full site build against this branch:

- build passes
- 19 pages render `<meta data-rh="true" name="robots" content="noindex, follow">`, all with the identical directive
- all 19 dropped out of `sitemap.xml`: 327 URLs on the base commit, 308 here, exactly 19 fewer and nothing else lost
- successor pages stay in: `HubSpot-real-time`, `Salesforce/salesforce-native`, `shopify-native`, `stripe-realtime`, `amazon-sqs-native`, `mailchimp-native`
- the `/Salesforce/` hub page correctly stays indexed
- all 19 still serve a non-empty page
- `data-rh="true"` confirms Helmet emitted the tag, which is what the sitemap exclusion keys off

## Notes

No companion change is needed. estuary/docs#77 was the swizzle for the front-matter approach and can be closed.

`amazon-sqs.md` and `mailchimp.md` were missing from the audit table in #5275 and so from the first pass here. Both carry a `:::deprecated` admonition and a named native successor, and both had descriptions that read as active connectors, so they are included now. That brings this to all 19 deprecated pages in the tree.

Two pre-existing items, both out of scope:

- The Algolia crawler still needs its own deprecated-page exclusion, configured in the Algolia dashboard rather than in a repo. Kapa already excludes them via `theme-admonition-deprecated`. A robots noindex reaches neither.
- `/reference/organizing-catalogs/` already sets `unlisted: true`, which emits `noindex, nofollow`. Unrelated to this change, left alone.

Not addressed here, tracked in #5275: this treats all 19 as `noindex`, where that issue ranks a redirect higher for pages whose successor fully replaces them, to be chosen per page against Search Console impressions. `Rockset.md` and `google-analytics.md` also still need their keep-or-remove decision.


aeluce (approved):
Descriptions were originally deliberately left off deprecated connectors. Adding `noindex` addresses concerns about that, though updating metadata in this way seems to require explicitly stating other metadata, like page title, rather than Docusaurus automatically picking it up: the sidebar displayed the raw file name like `salesforce-historical-data`. I added a commit to address that.

(Separately, do we have a list somewhere of everything that should happen when a connector gets deprecated? We're amassing additional differences to keep in mind: add a `:::deprecated` note, update the header to include `(Deprecated)`, update the meta description, add the meta title, add `noindex`…)

Also would be a huge fan of removing some of these deprecated connectors completely if there are no live tasks using them anymore.

